PR #13256, opened 2024-05-26
feat(Analysis/Convex): define half-spaces

Jun2M:
This is the first part of the proof of the main theorem of polytopes done in https://github.com/Jun2M/Main-theorem-of-polytopes

It introduces Halfspace type, its SetLike instance and a few simple lemmas.

---

### DIFF
--- a/Mathlib/Analysis/Convex/Halfspace.lean
+++ b/Mathlib/Analysis/Convex/Halfspace.lean
@@ -1,0 +1,109 @@
+/-
+Copyright (c) 2024 Hyeokjun Kwon. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Hyeokjun Kwon
+-/
+import Mathlib.Analysis.Convex.Independent
+import Mathlib.Analysis.InnerProductSpace.Dual
+import Mathlib.Analysis.NormedSpace.AddTorsorBases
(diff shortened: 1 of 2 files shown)


grunweg:
Thanks for your PR! Did you mean to make your pull request as ready for review? If so, please add the label `awaiting-review`.

I see you have a few linter failures - don't hestitate to ask if the error messages are unclear. Once you have these and you get a green checkmark, the PR will appear in the review queue - where it's more likely to receive attention.

Mathlib/Analysis/Convex/Halfspace.lean
```diff
@@ -0,0 +1,126 @@
...
+map of the direction vector.
+-/
+
+open LinearMap Set
+open scoped BigOperators Convex Pointwise
+
+
+/-- Halfspace is a set of points that lie on one side of a hyperplane.
+  or a set of points that their inner product with the direction vector is less than the bound -/
+structure Halfspace (E : Type) [NormedAddCommGroup E] [InnerProductSpace ℝ E] where
```
urkud, line 27: Do we need `InnerProductSpace` for this? Note that, e.g., `Real × Real` is not an inner product space. In any space, we can take a dual vector. Also, please use `Type*`, not `Type`.

PR #13257, opened 2024-05-26
[Merged by Bors] - feat: add `List.iterate_add`

eric-wieser:
---

### DIFF
--- a/Mathlib/Data/List/Iterate.lean
+++ b/Mathlib/Data/List/Iterate.lean
@@ -45,4 +45,10 @@ theorem range_map_iterate (n : ℕ) (f : α → α) (a : α) :
     (List.range n).map (f^[·] a) = List.iterate f a n := by
   apply List.ext_get <;> simp
 
+theorem iterate_add (f : α → α) (a : α) (m n : ℕ) :
+    iterate f a (m + n) = iterate f a m ++ iterate f (f^[m] a) n := by
+  induction m generalizing a with
+  | zero => simp
+  | succ n ih => rw [iterate, add_right_comm, iterate, ih, Nat.iterate, cons_append]
+
 end List
--- a/Mathlib/Data/List/Iterate.lean
+++ b/Mathlib/Data/List/Iterate.lean
@@ -4,7 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Miyahara Kō
 -/
 
-import Mathlib.Data.List.Defs
+import Mathlib.Data.List.Range
 import Mathlib.Algebra.Order.Ring.Nat
 
 /-!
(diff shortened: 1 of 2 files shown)


Mathlib/Data/List/Iterate.lean
```diff
@@ -17,6 +17,14 @@ variable {α : Type*}
 
 namespace List
 
+theorem take_range (m n : ℕ) : take m (range n) = range (min m n) := by
+  apply List.ext_get
+  · simp
+  · simp only [length_take, length_range, lt_min_iff, get_range]
+    intro i h _
+    rw [← get_take _ _ h.1, get_range]
+    simpa using h.2
```
Komyyy, line 26:
```suggestion
theorem take_range (m n : ℕ) : take m (range n) = range (min m n) := by
  apply List.ext_get <;> simp (config := { contextual := true }) [← get_take]
```
also, I think this theorem should be moved to `Mathlib.Data.List.Range`.

Komyyy (changes requested):

Mathlib/Data/List/Iterate.lean
eric-wieser: I think `lt_min_iff` is not available in that file

Komyyy: Use `Nat.lt_min` in `Init.Data.Nat.MinMax` instead.

eric-wieser: Nice, thanks!

Komyyy (approved):
Thank you!

Mathlib/Data/List/Range.lean
```diff
@@ -84,6 +84,11 @@ theorem pairwise_le_range (n : ℕ) : Pairwise (· ≤ ·) (range n) :=
   Pairwise.imp (@le_of_lt ℕ _) (pairwise_lt_range _)
 #align list.pairwise_le_range List.pairwise_le_range
 
+theorem take_range (m n : ℕ) : take m (range n) = range (min m n) := by
```
urkud, line 87: Should it be a `simp` lemma?

Mathlib/Data/List/Iterate.lean
```diff
@@ -45,4 +45,14 @@ theorem range_map_iterate (n : ℕ) (f : α → α) (a : α) :
     (List.range n).map (f^[·] a) = List.iterate f a n := by
   apply List.ext_get <;> simp
 
+theorem iterate_add (f : α → α) (a : α) (m n : ℕ) :
+    iterate f a (m + n) = iterate f a m ++ iterate f (f^[m] a) n := by
+  induction m generalizing a with
+  | zero => simp
+  | succ n ih => rw [iterate, add_right_comm, iterate, ih, Nat.iterate, cons_append]
+
+theorem take_iterate (f : α → α) (a : α) (m n : ℕ) :
```
urkud, line 54: Should it be a `simp` lemma?

eric-wieser (commented):
@urkud: my instinct was to leave these as not `simp` lemmas, since I'm not all that familiar with what's actually useful for `simp` to do when working on lists. Presumably at some point these will end up upstream somewhere, and I'll let the upstreamers decide whether `simp` is useful.

urkud:
Thanks! :tada:
bors merge

PR #13258, opened 2024-05-26
[Merged by Bors] - feat(Algebra/Lie/Weights/Chain): More API for chains.

erdOne:
---

### DIFF
--- a/Mathlib/Algebra/Lie/Weights/Chain.lean
+++ b/Mathlib/Algebra/Lie/Weights/Chain.lean
@@ -295,12 +295,12 @@ lemma weightSpace_neg_zsmul_add_ne_bot {n : ℕ} (hn : n ≤ chainBotCoeff α β
     weightSpace M ((-n : ℤ) • α + β : L → R) ≠ ⊥ := by
   apply weightSpace_zsmul_add_ne_bot α β <;> omega
 
-/-- This is the first weight in an `α`-chain through `β`. -/
+/-- The last weight in an `α`-chain through `β`. -/
 noncomputable
 def chainTop (α : L → R) (β : Weight R L M) : Weight R L M :=
   ⟨chainTopCoeff α β • α + β, weightSpace_nsmul_add_ne_bot_of_le α β le_rfl⟩
 
-/-- This is the first weight in an `α`-chain through `β`. -/
+/-- The first weight in an `α`-chain through `β`. -/
 noncomputable
 def chainBot (α : L → R) (β : Weight R L M) : Weight R L M :=
   ⟨(- chainBotCoeff α β : ℤ) • α + β, weightSpace_neg_zsmul_add_ne_bot α β le_rfl⟩
--- a/Mathlib/Algebra/Lie/Weights/Chain.lean
+++ b/Mathlib/Algebra/Lie/Weights/Chain.lean
@@ -217,4 +217,124 @@ end IsCartanSubalgebra
 
 end LieSubalgebra
 
+section
+
+variable {M}
+variable [LieAlgebra.IsNilpotent R L]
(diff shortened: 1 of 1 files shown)


Mathlib/Algebra/Lie/Weights/Chain.lean
```diff
@@ -217,4 +217,124 @@ end IsCartanSubalgebra
...
+lemma weightSpace_neg_zsmul_add_ne_bot {n : ℕ} (hn : n ≤ chainBotCoeff α β) :
+    weightSpace M ((-n : ℤ) • α + β : L → R) ≠ ⊥ := by
+  apply weightSpace_zsmul_add_ne_bot α β <;> omega
+
+/-- This is the first weight in an `α`-chain through `β`. -/
+noncomputable
+def chainTop (α : L → R) (β : Weight R L M) : Weight R L M :=
+  ⟨chainTopCoeff α β • α + β, weightSpace_nsmul_add_ne_bot_of_le α β le_rfl⟩
+
+/-- This is the first weight in an `α`-chain through `β`. -/
```
jcommelin, line 303: These two docstrings are identical. I guess you want `s/first/last/` in the first docstring?

jcommelin: Also, I would start the docstrings with `The .. weight`.
```suggestion
/-- The last weight in an `α`-chain through `β`. -/
noncomputable
def chainTop (α : L → R) (β : Weight R L M) : Weight R L M :=
  ⟨chainTopCoeff α β • α + β, weightSpace_nsmul_add_ne_bot_of_le α β le_rfl⟩

/-- The first weight in an `α`-chain through `β`. -/
```

Mathlib/Algebra/Lie/Weights/Chain.lean
```diff
@@ -217,4 +217,124 @@ end IsCartanSubalgebra
...
+
+lemma chainTop_isNonZero' (hα' : weightSpace M α ≠ ⊥) :
+    (chainTop α β).IsNonZero := by
+  by_contra e
+  apply hα'
+  rw [← add_zero (α : L → R), ← e, weightSpace_add_chainTop _ _ hα]
+
+lemma chainTop_isNonZero (α β : Weight R L M) (hα : α.IsNonZero) :
+    (chainTop α β).IsNonZero :=
+  chainTop_isNonZero' α β hα α.2
```
jcommelin, line 336: Should we have analogous lemmas for `chainBot`?

erdOne: I don't think this is possible in the general case. For example `chainBot α α` will be zero if `-α` is not a weight (or will this never happen)?

Mathlib/Algebra/Lie/Weights/Chain.lean
```diff
@@ -217,4 +217,124 @@ end IsCartanSubalgebra
...
+@[simp] lemma chainBot_neg : chainBot (-α) β = chainTop α β := by ext; simp
+
+@[simp] lemma chainTop_zero : chainTop 0 β = β := by ext; simp
+@[simp] lemma chainBot_zero : chainBot 0 β = β := by ext; simp
+
+lemma weightSpace_add_chainTop :
+    weightSpace M (α + chainTop α β : L → R) = ⊥ := by
+  rw [coe_chainTop', ← add_assoc, ← succ_nsmul', weightSpace_chainTopCoeff_add_one_nsmul_add _ _ hα]
+
+lemma weightSpace_neg_add_chainTop :
```
jcommelin, line 324:
```suggestion
lemma weightSpace_neg_add_chainBot :
```

Mathlib/Algebra/Lie/Weights/Chain.lean
jcommelin: Fair enough

bors d+

erdOne:
bors merge

PR #13195, opened 2024-05-25
[Merged by Bors] - feat: the Chebyshev polynomials C and S

trivial1711:
We define the rescaled Chebyshev polynomials $C_n$ and $S_n$ (also known as the Vieta–Lucas and Vieta–Fibonacci polynomials, respectively). They are related to the Chebyshev polynomials $T_n$ and $U_n$ by the formulas $C_n(2x) = 2T_n(x)$ and $S_n(2x) = U_n(x)$. Most theorems about $T_n$ and $U_n$ have analogues involving $C_n$ and $S_n$.

We prove that $C_n$ and $S_n$ are special cases of the Dickson polynomials (though unlike general Dickson polynomials, they are defined for every integer $n$, not just natural numbers).

These polynomials are necessary to state a formula for $(r_1 r_2)^m v$, where $v \in V$ is an element of a module, $r_1, r_2 \in GL(V)$ are reflections, and $m$ is an integer. The formula will be used to define and construct reflection representations of a Coxeter group over an arbitrary commutative ring, not necessarily having an invertible 2. See #13291.

---

- [x] depends on: #13133

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13270, opened 2024-05-27
[Merged by Bors] - feat: power of product of two reflections

trivial1711:
Let $M$ be a module over a commutative ring $R$. Let $x, y \in M$ and $f, g \in M^*$ with $f(x) = g(y) = 2$. The corresponding reflections $r_1, r_2 \colon M \to M$ are given by $r_1z = z - f(z) x$ and $r_2 z = z - g(z) y$. These are linear automorphisms of $M$.

We prove that for all $z \in M$ and all $m \in \mathbb{Z}$, we have:
$$(r_1 r_2)^m z = z + S_{\left\lfloor \frac{m-2}{2} \right \rfloor}(t) (S_{\left\lfloor \frac{m-1}{2} \right \rfloor}(t) + S_{\left\lfloor \frac{m-3}{2} \right \rfloor}(t)) (g(x) f(z) y - g(z) y - f(z) x) + S_{\left\lfloor \frac{m-1}{2} \right \rfloor}(t) (S_{\left\lfloor \frac{m}{2} \right \rfloor}(t) + S_{\left\lfloor \frac{m-2}{2} \right \rfloor}(t)) (f(y) g(z) x - f(z) x - g(z) y)$$
$$(r_1 r_2)^m x = (S_{m}(t) + S_{m - 1}(t)) x - S_{m-1} g(x) y$$
$$r_2(r_1 r_2)^m x = (S_{m}(t) + S_{m - 1}(t)) x - S_{m} g(x) y$$
where $t = f(y) g(x) - 2$ and $S_n$ refers to a Chebyshev $S$-polynomial.

These formulas may look ridiculous, but they are necessary for constructing reflection representations of a Coxeter group over an arbitrary ring. See #13291 for more details.

Co-authored-by: Johan Commelin <johan@commelin.net>

---

- [x] depends on: #13195


jcommelin:
Would you mind copying your appendix to a github issue, and then linking to that issue from here?
I think that is a better location for this explanation, because then it can be linked to from multiple locations.

trivial1711:
> Would you mind copying your appendix to a github issue, and then linking to that issue from here? I think that is a better location for this explanation, because then it can be linked to from multiple locations.

Thanks for the suggestion. I have done it: #13291

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -114,6 +121,173 @@ lemma bijOn_reflection_of_mapsTo {Φ : Set M} (h : f x = 2) (h' : MapsTo (reflec
...
+$S$-polynomials (`Polynomial.Chebyshev.S`) at $t = f(y) g(x) - 2$, and they hold over any
+commutative ring. -/
+section
+
+open Int Polynomial.Chebyshev
+
+variable {x y : M} {f g : Dual R M} (hf : f x = 2) (hg : g y = 2)
+
+local notation:max "t" => f y * g x - 2
+local notation:max "S_eval_t" n:max => Polynomial.eval t (S R n)
```
jcommelin, line 151: I understand that these notations make it easier to work in this file. But outside this file they don't apply. Which means that many of the statements will become rather complicated, and not so easy to handle.

I am not completely sure what the best strategy is. But I think `S_eval_t` could maybe be a standalone definition. (But I'm not sure what the best name is for it.)

jcommelin: What do you think of the following approach? Here I just add `t` as an extra assumption, together with `ht` giving its defining formula:
```
lemma reflection_mul_reflection_pow_apply (m : ℕ) (z : M) (t : R) (ht : t = f y * g x - 2) :
    ((reflection hf * reflection hg) ^ m) z =
      z +
        ((S R ((m - 2) / 2)).eval t * ((S R ((m - 1) / 2)).eval t + (S R ((m - 3) / 2)).eval t)) •
          ((g x * f z - g z) • y - f z • x) +
        ((S R ((m - 1) / 2)).eval t * ((S R (m / 2)).eval t + (S R ((m - 2) / 2)).eval t)) •
          ((f y * g z - f z) • x - g z • y) := by
```

jcommelin: If we can convince the pretty printer to use dot-notation for `Polynomial.eval` (see [zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Polynomial.2Eeval.20dot-notation.20delaborator)), then I think this is quite readable, and doesn't need local notation.

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -114,6 +121,173 @@ lemma bijOn_reflection_of_mapsTo {Φ : Set M} (h : f x = 2) (h' : MapsTo (reflec
...
+      norm_num
+      /- Now, equate the coefficients on both sides. These linear combinations were
+      found using `polyrith`. -/
+      match_scalars
+      · rfl
+      · linear_combination (norm := ring_nf) (-g z * f y * S_eval_t k +
+            f z * S_eval_t k) * S_eval_t_sub_two (k + 1) +
+          (-g z * f y + f z) * S_eval_t_sq_add_S_eval_t_sq (k - 1)
+      · linear_combination (norm := ring_nf) g z * S_eval_t k * S_eval_t_sub_two (k + 1) +
+          g z * S_eval_t_sq_add_S_eval_t_sq (k - 1)
```
jcommelin, line 206: What do you think of the following way to finish the proof?
```suggestion
    push_cast
    set k : ℤ := m / 2
    set e : ℤ := m % 2
    rw [show m = 2 * k + e from (Int.ediv_add_emod m 2).symm]
    simp_rw [add_assoc (2 * k), add_sub_assoc (2 * k), add_comm (2 * k),
      add_mul_ediv_left _ k (by norm_num : (2 : ℤ) ≠ 0)]
    have he : e = 0 ∨ e = 1 := by omega
    clear_value e
    set t' := t; subst ht; set t := t'
    match_scalars
    · rfl
    · linear_combination (norm := skip) (-g z * f y * (S R (e - 1 + k)).eval t +
          f z * (S R (e - 1 + k)).eval t) * S_eval_t_sub_two (e + k) +
          (-g z * f y + f z) * S_eval_t_sq_add_S_eval_t_sq (k - 1)
      obtain rfl | rfl : e = 0 ∨ e = 1 := he <;> ring_nf
    · linear_combination (norm := skip)
          g z * (S R (e - 1 + k)).eval t * S_eval_t_sub_two (e + k) +
          g z * S_eval_t_sq_add_S_eval_t_sq (k - 1)
      obtain rfl | rfl : e = 0 ∨ e = 1 := he <;> ring_nf
```

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -114,6 +121,173 @@ lemma bijOn_reflection_of_mapsTo {Φ : Set M} (h : f x = 2) (h' : MapsTo (reflec
...
+
```
jcommelin, line 207: It might be useful to have an unapplied version of the lemma:
```
/-- A formula for $(r_1 r_2)^m$, where $m$ is a natural number. -/
lemma reflection_mul_reflection_pow' (m : ℕ) (t : R) (ht : t = f y * g x - 2) :
    ((reflection hf * reflection hg) ^ m) =
      LinearMap.id (R := R) (M := M) +
        (((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2))).eval t) •
          ((g x • f - g).smulRight y - f.smulRight x)) +
        (((S R ((m - 1) / 2) * (S R (m / 2) + S R ((m - 2) / 2))).eval t) •
          ((f y • g - f).smulRight x - g.smulRight y)) := by
  ext z
  simp
  rw [← reflection_mul_reflection_pow_apply hf hg m z t ht]
  rw [LinearMap.pow_apply, LinearEquiv.pow_apply]
  rfl
```

trivial1711:
Thank you so much for the feedback.

I added the hypotheses `t` and `ht` to all the lemmas in this section, as you suggested. I also gave them default values: `(t : R := f y * g x - 2) (ht : t = f y * g x - 2 := by rfl)`.

I added `reflection_mul_reflection_pow` and `reflection_mul_reflection_zpow`, which are unapplied versions of `reflection_mul_reflection_pow_apply` and `reflection_mul_reflection_zpow_apply` respectively. Currently, both `reflection_mul_reflection_pow` and `reflection_mul_reflection_zpow` are about `((reflection hf * reflection hg) ^ m).toLinearMap` instead of `(reflection hf * reflection hg) ^ m`. I could change it so that `reflection_mul_reflection_pow` is about `(reflection hf * reflection hg) ^ m` but `reflection_mul_reflection_zpow` is about `((reflection hf * reflection hg) ^ m).toLinearMap`, if you think that would be better.

Finally, I changed the proof of `reflection_mul_reflection_pow_apply` as you suggested, which greatly simplified it.

trivial1711:
Do you think it is better to state the lemmas using `((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2)).eval t` or `(S R ((m - 2) / 2)).eval t * ((S R ((m - 1) / 2)).eval t + (S R ((m - 3) / 2)).eval t)`? I think I should change all of them to be consistent.

trivial1711:
> Do you think it is better to state the lemmas using `((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2)).eval t` or `(S R ((m - 2) / 2)).eval t * ((S R ((m - 1) / 2)).eval t + (S R ((m - 3) / 2)).eval t)`? I think I should change all of them to be consistent.

I think I prefer the latter. The reason is that `((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2)).eval t` is not in simp normal form because of [Polynomial.eval_add](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Polynomial/Eval/Defs.html#Polynomial.eval_add). (It is questionable whether `Polynomial.eval_add` should be a simp lemma in the first place because it has two `x`s on the RHS and only one on the LHS, but as long as it is, `(S R ((m - 2) / 2)).eval t * ((S R ((m - 1) / 2)).eval t + (S R ((m - 3) / 2)).eval t)` should be better than `((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2)).eval t`.)

trivial1711:
But on the other hand, `((S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2)).eval t` is easier to reason with (for example, it is obviously equivalent to the statement that `t` is a root of the polynomial `(S R ((m - 2) / 2) * (S R ((m - 1) / 2) + S R ((m - 3) / 2))`)

Should I make a separate pull request that removes the @[simp] attribute from Polynomial.eval_mul and Polynomial.eval_add?

jcommelin:
Let's just stick to the current normal forms, and I agree it would then be better to have the version with 3 `eval t`s.

The fact that `Polynomial.eval_{mul,add}` is a simp lemma is consistent with the fact that for any monoid hom we have `f (x * y) =  f x * f y` as a simp lemma. So I don't think we should change that.

jcommelin (commented):
I'll leave it up to you to decide on whether you want to put the algebra inside or outside `.eval t`.
The rest LGTM.

Thanks! :tada: 

bors d+

PR #13273, opened 2024-05-27
[Merged by Bors] - feat(NumberTheory/LSeries): Riemann zeta as special case of Hurwitz

loefflerd:
We now have a general treatment of the Hurwitz zeta function. The Riemann zeta is just the Hurwitz zeta for `a = 0`, and this PR reimplements it as such. 

I have tried to carry over the `align` statements from the old treatment wherever possible (many of the intermediate lemmas from the previous iteration don't have direct analogues, but the main results do).

---

- [x] depends on: #13347


Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+def completedRiemannZeta₀ (s : ℂ) : ℂ := completedHurwitzZetaEven₀ 0 s
+#align riemann_completed_zeta₀ completedRiemannZeta₀
+
+/-- The completed Riemann zeta function, `Λ(s)`, which satisfies
+`Λ(s) = π ^ (-s / 2) Γ(s / 2) ζ(s)` (up to a minor correction at `s = 0`). -/
+def completedRiemannZeta (s : ℂ) : ℂ := completedHurwitzZetaEven 0 s
+#align riemann_completed_zeta completedRiemannZeta
+
+lemma HurwitzZeta.completedHurwitzZetaEven_zero (s : ℂ) :
+    completedHurwitzZetaEven 0 s = completedRiemannZeta s := by rfl
```
MichaelStollBayreuth, line 72: Is the `by` necessary here? (same a couple lines below)

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+    completedHurwitzZetaEven₀ 0 s = completedRiemannZeta₀ s := by rfl
+
+lemma HurwitzZeta.completedCosZeta_zero (s : ℂ) :
+    completedCosZeta 0 s = completedRiemannZeta s := by
+  rw [completedRiemannZeta, completedHurwitzZetaEven, completedCosZeta, hurwitzEvenFEPair_zero_symm]
+
+lemma HurwitzZeta.completedCosZeta₀_zero (s : ℂ) :
+    completedCosZeta₀ 0 s = completedRiemannZeta₀ s := by
+  rw [completedRiemannZeta₀,
+    completedHurwitzZetaEven₀, completedCosZeta₀, hurwitzEvenFEPair_zero_symm]
```
MichaelStollBayreuth, line 84:
```suggestion
  rw [completedRiemannZeta₀, completedHurwitzZetaEven₀, completedCosZeta₀,
    hurwitzEvenFEPair_zero_symm]
```
is the usual formatting.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+  differentiableAt_completedHurwitzZetaEven 0 (Or.inl hs) hs'
+
+/-- Riemann zeta functional equation, formulated for `Λ₀`: for any complex `s` we have
+`Λ₀(1 - s) = Λ₀ s`. -/
+theorem completedRiemannZeta₀_one_sub (s : ℂ) :
+    completedRiemannZeta₀ (1 - s) = completedRiemannZeta₀ s := by
+  rw [← completedHurwitzZetaEven₀_zero, ← completedCosZeta₀_zero, completedHurwitzZetaEven₀_one_sub]
+#align riemann_completed_zeta₀_one_sub completedRiemannZeta₀_one_sub
+
+@[deprecated completedRiemannZeta₀_one_sub (since := "2024-05-27")]
```
MichaelStollBayreuth, line 107: Why is this deprecated? Is there a replacement elsewhere?

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+  ∀ (s : ℂ) (_ : riemannZeta s = 0) (_ : ¬∃ n : ℕ, s = -2 * (n + 1)) (_ : s ≠ 1), s.re = 1 / 2
+#align riemann_hypothesis RiemannHypothesis
+
+/-!
+## Relating the Mellin transform to the Dirichlet series
+-/
+
+theorem completedZeta_eq_tsum_of_one_lt_re {s : ℂ} (hs : 1 < re s) :
+    completedRiemannZeta s = (π : ℂ) ^ (-s / 2) * Gamma (s / 2) *
+    ∑' n : ℕ, 1 / (n : ℂ) ^ s := by
```
MichaelStollBayreuth, line 183:
```suggestion
    completedRiemannZeta s =
      (π : ℂ) ^ (-s / 2) * Gamma (s / 2) * ∑' n : ℕ, 1 / (n : ℂ) ^ s := by
```
is easier to parse, I think.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+@[deprecated completedRiemannZeta₀ (since := "2024-05-27")]
+noncomputable alias riemannCompletedZeta₀ := completedRiemannZeta₀
+
+@[deprecated completedRiemannZeta (since := "2024-05-27")]
+noncomputable alias riemannCompletedZeta := completedRiemannZeta
+
+@[deprecated completedRiemannZeta_one_sub (since := "2024-05-27")]
+alias riemannCompletedZeta_one_sub := completedRiemannZeta_one_sub
+
+@[deprecated completedRiemannZeta_residue_one (since := "2024-05-27")]
```
MichaelStollBayreuth, line 243: Same question as above: why are these deprecated?

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: This is just a backward-compatibility measure: I renamed some declarations, and there's a convention that when you rename `oldName` to `newName` you add `@[deprecated] alias oldName newName` so anyone still using `oldName` in their code will get a polite message saying "oldName is deprecated, use newName", rather than everything just stopping working. (Cf. Sebastien's recent post in the PNT+ thread, about the experience of bumping the mathlib version used in the PNT+ repo.)

As for _why_ the declarations were renamed: originally we had `riemannCompletedZeta`, ported from `riemann_completed_zeta` in Mathlib3. On rewriting this stuff, I didn't like the look of `hurwitzCompletedZeta` -- it's a completion of Hurwitz' zeta function, not Hurwitz' completion of a zeta function (while both readings work for Riemann!) so went for `completedHurwitzZeta`, and changed the Riemann zeta for consistency with that.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: Same answer as above :smile:

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: (PS. Evidently you already know this, since you use the same mechanism in your `AddChar` PR; and I learned from your code there that one does not need to specify the new name in the deprecation.)

MichaelStollBayreuth: My point is that you deprecated the *new* names. This was confusing me.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+theorem completedZeta_eq_tsum_of_one_lt_re {s : ℂ} (hs : 1 < re s) :
+    completedRiemannZeta s =
+      (π : ℂ) ^ (-s / 2) * Gamma (s / 2) * ∑' n : ℕ, 1 / (n : ℂ) ^ s := by
+  have := (hasSum_nat_completedCosZeta 0 hs).tsum_eq.symm
+  simp only [QuotientAddGroup.mk_zero, completedCosZeta_zero] at this
+  simp only [this, Gammaℝ_def, mul_zero, zero_mul, Real.cos_zero, ofReal_one, mul_one, mul_one_div,
+    ← tsum_mul_left]
+  congr 1 with n
+  split_ifs with h <;>
+  simp only [h, Nat.cast_zero, zero_cpow (Complex.ne_zero_of_one_lt_re hs), div_zero, mul_zero]
```
MichaelStollBayreuth, line 187:
```suggestion
  · simp only [h, Nat.cast_zero, zero_cpow (Complex.ne_zero_of_one_lt_re hs), div_zero]
  · rfl
```
I think this is clearer (and worth the additional line). (And `mul_zero`  was apparently not necessary.)

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+    riemannZeta s = ∑' n : ℕ, 1 / (n : ℂ) ^ s := by
+  simpa only [QuotientAddGroup.mk_zero, cosZeta_zero, mul_zero, zero_mul, Real.cos_zero,
+    ofReal_one] using (hasSum_nat_cosZeta 0 hs).tsum_eq.symm
+#align zeta_eq_tsum_one_div_nat_cpow zeta_eq_tsum_one_div_nat_cpow
+
+/-- Alternate formulation of `zeta_eq_tsum_one_div_nat_cpow` with a `+ 1` (to avoid relying
+on mathlib's conventions for `0 ^ s`).  -/
+theorem zeta_eq_tsum_one_div_nat_add_one_cpow {s : ℂ} (hs : 1 < re s) :
+    riemannZeta s = ∑' n : ℕ, 1 / (n + 1 : ℂ) ^ s := by
+  have hs' : s ≠ 0 := fun h ↦ (not_lt.mpr zero_le_one) (zero_re ▸ h ▸ hs)
```
MichaelStollBayreuth, line 203:
```suggestion
  have hs' : s ≠ 0 := ne_zero_of_one_lt_re hs
```
(We had this before, but I assume the code had been written before the other discussion.)
Since `hs'` is only used once, this could perhaps be inlined (which does not save a line, though, because the `simpa` below then needs two).

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: > My point is that you deprecated the new names. This was confusing me.

No, I don't think I did, unless I'm grotesquely misunderstanding how the deprecation mechanism works. The old definitions were `riemannCompletedXXX`, they have been replaced by `completedRiemanXXX`, and I am adding back `riemannCompletedXXX` as a deprecated alias for the new declaration.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+  have := zeta_eq_tsum_one_div_nat_cpow hs
+  rw [tsum_eq_zero_add] at this
+  · simpa [Nat.cast_zero, zero_cpow hs', div_zero, zero_add, Nat.cast_add, Nat.cast_one]
```
MichaelStollBayreuth, line 206:
```suggestion
  · simpa only [one_div, Nat.cast_zero, zero_cpow <| ne_zero_of_one_lt_re hs, div_zero,
      Nat.cast_add, Nat.cast_one, zero_add]
```
(with `hs'` inlined).

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: ... but most of the lemmas in the `simpa` don't need to be explicitly given anyway.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
```diff
@@ -0,0 +1,246 @@
...
+  · refine .of_norm ?_
+    simp_rw [norm_div, norm_one, Complex.norm_eq_abs, ← ofReal_natCast,
+      abs_cpow_eq_rpow_re_of_nonneg (Nat.cast_nonneg _) (zero_lt_one.trans hs).ne',
+      summable_one_div_nat_rpow]
```
MichaelStollBayreuth: Variants of this are already present (e.g., [LSeriesSummable_of_bounded_of_one_lt_re](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/LSeries/Basic.html#LSeriesSummable_of_bounded_of_one_lt_re)), or, perhaps closer, [summable_riemannZetaSummand](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/EulerProduct/DirichletLSeries.html#summable_riemannZetaSummand)). Still, I don't think the statement is available in this precise form. It might be a consideration to add it next to [Real.summable_one_div_nat_rpow](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/PSeries.html#Real.summable_one_div_nat_rpow).

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: I've left the `simpa` unsqueezed since it's quite quick (< 0.3s), and instead removed the listed lemmas that are already simp lemmas.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: Just to be clear: you're suggesting we add something like `Complex.summable_one_div_nat_cpow`?

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -197,3 +199,57 @@ theorem hurwitzZeta_neg_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
   · exact_mod_cast hurwitzZeta_one_sub_two_mul_nat (by omega : n + 1 ≠ 0) hx
 
 end HurwitzZeta
+
+open HurwitzZeta
+
+/-- Explicit formula for `ζ (2 * k)`, for `k ∈ ℕ` with `k ≠ 0`: we have
+`ζ (2 * k) = (-1) ^ (k + 1) * 2 ^ (2 * k - 1) * π ^ (2 * k) * bernoulli (2 * k) / (2 * k)!`.
+Compare `hasSum_zeta_nat` for a version formulated explicitly as a sum, and
```
MichaelStollBayreuth, line 207: `hasSum_zeta_nat` is stated in terms of `HasSum`; I wouldn't describe that as "explicitly as a sum".

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -197,3 +199,57 @@ theorem hurwitzZeta_neg_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
...
+`riemannZeta_neg_nat_eq_bernoulli` for values at negative integers (equivalent to the above via
+the functional equation). -/
+theorem riemannZeta_two_mul_nat {k : ℕ} (hk : k ≠ 0) :
+    riemannZeta (2 * k) = (-1) ^ (k + 1) * (2 : ℂ) ^ (2 * k - 1)
+      * (π : ℂ) ^ (2 * k) * bernoulli (2 * k) / (2 * k)! := by
+  convert congr_arg ((↑) : ℝ → ℂ) (hasSum_zeta_nat hk).tsum_eq
+  · rw [← Nat.cast_two, ← Nat.cast_mul, zeta_nat_eq_tsum_of_gt_one]
+    · rw [ofReal_tsum]
+      norm_num
```
MichaelStollBayreuth, line 216:
```suggestion
    · simp only [ofReal_tsum, ofReal_div, ofReal_pow, ofReal_natCast, ofReal_one]
```
`norm_num` looks a bit out of scope here.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
loefflerd: (This docstring is the same as it was in the previous code, it's just been relocated.) 

My point is that `hasSum_zeta_nat` is *not* formulated in terms of `riemannZeta` (despite the name) but uses the defining Dirichlet series explicitly. If you prefer another formulation please let me know.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -197,3 +199,57 @@ theorem hurwitzZeta_neg_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
...
+    · refine one_lt_two.trans_le ?_
+      conv_lhs => rw [← mul_one 2]
+      rwa [mul_le_mul_left (zero_lt_two' ℕ), Nat.one_le_iff_ne_zero]
+  · norm_num
```
MichaelStollBayreuth, line 220:
```suggestion
  · rw [← Nat.cast_two, ← Nat.cast_mul, zeta_nat_eq_tsum_of_gt_one (by omega)]
    simp only [ofReal_tsum, ofReal_div, ofReal_pow, ofReal_natCast, ofReal_one]
  · norm_cast
```
`norm_num` looks out of scope (no numerical computation), and inequalities of natural numbers or integers can be (efficiently) dealt with using `omega`.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -197,3 +199,57 @@ theorem hurwitzZeta_neg_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
...
+#align riemann_zeta_two_mul_nat riemannZeta_two_mul_nat
+
+theorem riemannZeta_two : riemannZeta 2 = (π : ℂ) ^ 2 / 6 := by
+  convert congr_arg ((↑) : ℝ → ℂ) hasSum_zeta_two.tsum_eq
+  · rw [← Nat.cast_two, zeta_nat_eq_tsum_of_gt_one one_lt_two, ofReal_tsum]
+    norm_num
```
MichaelStollBayreuth, line 226:
```suggestion
    simp only [ofReal_div, ofReal_one, ofReal_pow, ofReal_natCast]
```

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -197,3 +199,57 @@ theorem hurwitzZeta_neg_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
...
+  · norm_num
```
MichaelStollBayreuth, line 227:
```suggestion
  · norm_cast
```

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
MichaelStollBayreuth: My understanding is that when you write `@[deprecated <name>]`, then this deprecates `<name>` (and `<name>` was `completedRiemann...`.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
MichaelStollBayreuth: Ah, OK. I was reading "explicitly as a sum" as somehow differentiating from the equality and wondering, since in the `HasSum` version, no actual sum appears. Maybe write "... in terms of a sum over `1/n^(2*k)`" or so.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
acmepjz: > My understanding is that when you write `@[deprecated <name>]`, then this deprecates `<name>` (and `<name>` was `completedRiemann...`.

No, the `<name>` is the replacement.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
loefflerd: Good point, I think I must have got `norm_num` and `norm_cast` muddled.

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
MichaelStollBayreuth: Thanks for clarifying (and sorry for the noise!).

Mathlib/NumberTheory/LSeries/RiemannZeta.lean
loefflerd: Done. I made a new file for it in order to avoid massively increasing the imports of `Analysis.PSeries`.

MichaelStollBayreuth (approved):
LGTM!

MichaelStollBayreuth:
maintainer merge

PR #13244, opened 2024-05-26
[Merged by Bors] - chore: Split `Algebra.Function.Support`/`Algebra.Function.Indicator`

YaelDillies:
... in a file about `Group` and another one about `GroupWithZero`. Credit Yury for https://github.com/leanprover-community/mathlib/pull/2340 and Zhouhang for https://github.com/leanprover-community/mathlib/pull/1949.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13243, opened 2024-05-26
[Merged by Bors] - chore: Do not import algebra in `Data.Finset.Lattice`

YaelDillies:
Move the `OrderBot Nat` instance to a new file `Order.Nat` and the `LinearOrderedSemiring` lemmas from `Data.Finset.Lattice` to a new file `Algebra.Order.Ring.Finset`. I credit Eric for https://github.com/leanprover-community/mathlib/pull/12912.

---

Note that @Ruben-VandeVelde will move more instances to `Order.Nat` soon.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13277, opened 2024-05-27
[Merged by Bors] - chore: Split `Algebra.BigOperators.Basic`

YaelDillies:
... into two files:
* `Algebra.BigOperators.Group.Finset` for lemmas about group-like objects.
* `Algebra.BigOperators.GroupWithZero.Finset` for lemmas about group-with-zero-like objects. Credit Johannes for https://github.com/leanprover-community/mathlib/commit/edfbf3cf.

---

- [x] depends on: #13243
- [x] depends on: #13244


dupuisf (commented):
Thanks!

bors r+

PR #13278, opened 2024-05-27
chore(GradedAlgebra.HomogeneousLocalization): do not compute `smul` on every pass of `Algebra`

mattrobball:
We make a separate `SMul (HomogeneousLocalization A x) (Localization x)` to avoid Lean computing `smul` uneccessarily.

---


mattrobball:
!bench

fpvandoorn:
Marking as `awaiting-author`. Please replace that label with `awaiting-review` if appropriate.

PR #13279, opened 2024-05-27
[Merged by Bors] - style(Mathlib/Tactic/*): rename wrongly capitalized filenames

Komyyy:
`Mathlib.Tactic.Congrm`
→ `Mathlib.Tactic.CongrM`

`Mathlib.Tactic.Existsi`
→ `Mathlib.Tactic.ExistsI`

`Mathlib.Tactic.Widget.Congrm`
→ `Mathlib.Tactic.Widget.CongrM`

`Mathlib.Tactic.Widget.Gcongr`
→ `Mathlib.Tactic.Widget.GCongr`

Also, we rename the namespaces and subroutines in these files.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #12757, opened 2024-05-08
[Merged by Bors] - chore(Mathlib/Tactic/Clear!): rename `Clear!.lean` to `ClearExclamation.lean` to avoid the illegal character `!`

Komyyy:
---

Resolves #1212

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13280, opened 2024-05-27
[Merged by Bors] - style(Mathlib/Tactic/Congr!): rename `Congr!.lean` to `CongrExclamation.lean` to avoid the illegal character `!`

Komyyy:
I renamed `Clear!.lean` to `ClearExclamation.lean` to avoid the illegal character in nixOS: `!` in #12757.
However, there is another file which contains `!`.

---

- [x] depends on: #13279


YaelDillies:
Looks reasonable.

maintainer merge

PR #13282, opened 2024-05-27
dev: refine replace linter dev

adomani:
---


adomani:
The refine vs refine' linter is in mathlib, so this PR with the script to automatically substitute `refine'` with `refine` is no longer needed.

PR #13286, opened 2024-05-27
[Merged by Bors] - chore: remove most of the material on BitVec

PatrickMassot:
This PR removes most of the material on BitVec. While it is being actively developed in Lean, we do not want conflicting, but essentially unused, material here causing conflicts.

If downstream users of Mathlib find that they need some of this material is being used, they are advised to either:
* restore the material in their downstream repository
* ask someone from the FRO, e.g. @semorrison, about making a PR adding it to the Lean library.

---


Mathlib/Data/BitVec/Lemmas.lean
```diff
@@ -1,199 +0,0 @@
...
-proof_wanted toFin_intCast (z : ℤ) : toFin (z : BitVec w) = z
-
-/-!
-## Ring
--/
-
--- TODO: generalize to `CommRing` after `ofFin_intCast` is proven
-instance : CommSemiring (BitVec w) :=
-  toFin_injective.commSemiring _ toFin_zero toFin_one toFin_add toFin_mul toFin_nsmul
-    toFin_pow toFin_natCast
```
alexkeizer, line 197: While most of these files are indeed results that should be in core after the move, this particular instance is about `CommSemiring`, a Mathlib typeclass, and thus should remain in Mathlib.

bollu: Moreover, we in fact have a [`CommRing` instance](https://github.com/opencompl/ssa/blob/ec6e647cd91936509e372d1615e95aab7415fe8c/SSA/Projects/InstCombine/ForMathlib.lean#L51-L54) which we use for [proof automation](https://github.com/opencompl/ssa/blob/ec6e647cd91936509e372d1615e95aab7415fe8c/SSA/Projects/InstCombine/TacticAuto.lean#L70-L83) that we would eventually like to re-home into `mathlib`.

alexkeizer: In fact, most proofs that this instance depends on are just def-eqs, so we can inline them if we really don't want those results in mathlib.

Here's a relatively minimal file that retains the `CommSemiring` instance:
```lean4
/-
Copyright (c) 2020 Simon Hudon. All rights reserved.
Authors: Simon Hudon, Harun Khan, Alex Keizer
-/
import Mathlib.Algebra.Ring.InjSurj
import Mathlib.Data.ZMod.Defs

/-!
# Basic Theorems About Bitvectors

This file contains theorems about bitvectors.
-/

namespace BitVec

variable {w v : Nat}

theorem toFin_injective {n : Nat} : Function.Injective (toFin : BitVec n → _)
  | ⟨_, _⟩, ⟨_, _⟩, rfl => rfl

/-!
## Ring
-/

instance : SMul ℕ (BitVec w) := ⟨fun x y => ofFin <| x • y.toFin⟩
instance : SMul ℤ (BitVec w) := ⟨fun x y => ofFin <| x • y.toFin⟩
instance : Pow (BitVec w) ℕ  := ⟨fun x n => ofFin <| x.toFin ^ n⟩

instance : CommSemiring (BitVec w) :=
  toFin_injective.commSemiring _ rfl rfl toFin_add toFin_mul (fun _ _ => rfl)
    (fun _ _ => rfl) (fun _ => rfl)

end BitVec
```

semorrison: @alexkeizer, do you want to just push that to this PR?

Mathlib/Data/BitVec/Lemmas.lean
```diff
@@ -1,199 +0,0 @@
...
-variable {w v : Nat}
-
-theorem toFin_injective {n : Nat} : Function.Injective (toFin : BitVec n → _)
-  | ⟨_, _⟩, ⟨_, _⟩, rfl => rfl
-
-theorem toFin_inj {x y : BitVec w} : x.toFin = y.toFin ↔ x = y :=
-  toFin_injective.eq_iff
-
-theorem toNat_injective {n : Nat} : Function.Injective (BitVec.toNat : BitVec n → _) :=
-  Fin.val_injective.comp toFin_injective
```
eric-wieser, line 31: I think anything about `Function.Injective` should be kept, since we can't _state_ those lemmas any further upstream

Mathlib/Data/BitVec/Lemmas.lean
eric-wieser: > In fact, most proofs that this instance depends on are just def-eqs, so we can inline them if we really don't want those results in mathlib.

Dropping the defeq lemmas seems like a bad idea to me. If mathlib is providing the definitions (`Pow`/`SMul` etc), then it should provide the defeq lemmas as well.

The defeq lemmas about things like `^^^` which are themselves defined upstream should probably migrate upstream.

alexkeizer: I've now pushed a version of the file I commented which keeps `toFin_nsmul`, `toFin_zsmul` and `toFin_pow`, plus the results about `Function.Injective`

Mathlib/Data/BitVec/Defs.lean
```diff
@@ -1,146 +0,0 @@
...
-/-- The bitvector representing `1`.
-    That is, the bitvector with least-significant bit `1` and all other bits `0` -/
-@[simp] abbrev one (w : ℕ) : BitVec w := 1
-#align bitvec.one BitVec.one
-
-#align bitvec.cong BitVec.cast
-#align bitvec.append BitVec.append
-#align bitvec.shl BitVec.shiftLeft
-#align bitvec.ushr BitVec.ushiftRight
-#align bitvec.sshr BitVec.sshiftRight
```
eric-wieser, line 48: Repeating a comment that was made privately; we should keep the `#align`s around, at least as long as we keep any `align`s around.

Mathlib/Data/BitVec/Defs.lean
```diff
@@ -1,146 +0,0 @@
...
-
-/-- Signed greater than for bitvectors. -/
-protected def sgt (x y : BitVec w) : Bool := BitVec.slt y x
-#align bitvec.sgt BitVec.sgt
-
-#align bitvec.sle BitVec.sle
-
-/-- Signed greater than or equal to for bitvectors. -/
-protected def sge (x y : BitVec w) : Bool := BitVec.sle y x
-#align bitvec.sge BitVec.sge
```
eric-wieser, line 104: Arguably these `def` should become `@[deprecated] alias`es before we remove them outright, but perhaps no one cares.

Mathlib/Data/BitVec/Lemmas.lean
```diff
@@ -1,199 +0,0 @@
...
-
-/-!
-### `IntCast`
--/
-
--- Either of these follows trivially from the other. Which one to
--- prove is not yet clear.
-proof_wanted ofFin_intCast (z : ℤ) : ofFin (z : Fin (2^w)) = z
-
-proof_wanted toFin_intCast (z : ℤ) : toFin (z : BitVec w) = z
```
eric-wieser, line 188: We still want these `proof_wanted`s, right?

semorrison: No.

Mathlib/Data/BitVec/Lemmas.lean
semorrison: I've merged https://github.com/leanprover/lean4/pull/4298

Mathlib/Data/BitVec/Lemmas.lean
PatrickMassot: That should be stated downstream of Mathlib, not upstream.

Mathlib/Data/BitVec/Lemmas.lean
semorrison: Thanks for the push.

Mathlib/Data/BitVec/Defs.lean
semorrison: I've restored the #aligns, turning them into #noalign when something is being deleted.

Mathlib/Data/BitVec/Lemmas.lean
semorrison: I think the decision is that these can just be done downstream for now.

semorrison:
bors merge

Mathlib/Data/BitVec/Defs.lean
eric-wieser: Thanks!

PR #13290, opened 2024-05-27
[Merged by Bors] - feat(RingTheory/RootsOfUnity): product representation of the order of a primitive root of unity

MichaelStollBayreuth:
This
* weakens the assumptions in [X_pow_sub_C_eq_prod](https://leanprover-community.github.io/mathlib4_docs/Mathlib/FieldTheory/KummerExtension.html#X_pow_sub_C_eq_prod) on the coefficient ring from "field" to "integral domain", and
* adds results on representing the order `n` of a primitive root of unity `μ` in `R` as a product involving terms like `μ^k - 1`. (In a new file `RingTheory.RootsOfUnity.Lemmas` because of fairly heavy imports.)

This will  be useful for results on Jacobi sums (to be added eventually).

---


Mathlib/RingTheory/RootsOfUnity/More.lean
```diff
@@ -0,0 +1,83 @@
...
+
+variable {R : Type*} [CommRing R] [IsDomain R]
+
+namespace IsPrimitiveRoot
+
+open Finset Polynomial BigOperators
+
+/-- If `μ` is a primitive `n`th root of unity in `R`, then `∏(1≤k<n) (1-μ^k) = n`. -/
+lemma prod_one_sub_pow_eq_order {n : ℕ} (hn : n ≠ 0) {μ : R} (hμ : IsPrimitiveRoot μ n) :
+    ∏ k ∈ range (n - 1), (1 - μ ^ (k + 1)) = n := by
```
Ruben-VandeVelde, line 37: Is it worth substituting n := n - 1?

MichaelStollBayreuth: Looks like it.

Mathlib.lean
```diff
@@ -3599,6 +3599,7 @@ import Mathlib.RingTheory.RingInvo
 import Mathlib.RingTheory.RootsOfUnity.Basic
 import Mathlib.RingTheory.RootsOfUnity.Complex
 import Mathlib.RingTheory.RootsOfUnity.Minpoly
+import Mathlib.RingTheory.RootsOfUnity.More
```
riccardobrasca, line 3602: I think we usually call such files `lemmas`, or something like that (of course it would be better to find a more precise name, but I understand it is not always possible).

Mathlib/RingTheory/RootsOfUnity/More.lean
```diff
@@ -0,0 +1,79 @@
...
+Assume that `μ` is a primitive `n`th root of unity in an integral domain `R`. Then
+$$ \prod_{k=1}^{n-1} (1 - \mu^k) = n \,; $$
+see `IsPrimitiveRoot.prod_one_sub_pow_eq_order` and its variant
+`IsPrimitiveRoot.prod_pow_sub_one_eq_order` in terms of `∏ (μ^k - 1)`.
+
+We use this to deduce that `n` is divisible by `(μ - 1)^k` in `ℤ[μ] ⊆ R` when `k < n`.
+-/
+
+/-- If `A` is an `R`-algebra and `μ : A`, then `μ^k-1` is divisible by `μ-1` in `R[μ] ⊆ A`. -/
+lemma Algebra.adjoin.sub_one_dvd_pow_sub_one (R : Type*) {A : Type*} [CommRing R] [CommRing A]
```
riccardobrasca, line 23: Can you use `∣` explicitly here?

Mathlib/RingTheory/RootsOfUnity/More.lean
```diff
@@ -0,0 +1,79 @@
...
+  exact Subalgebra.sum_mem _ fun m _ ↦ Subalgebra.pow_mem _ (self_mem_adjoin_singleton _ μ) _
+
+variable {R : Type*} [CommRing R] [IsDomain R]
+
+namespace IsPrimitiveRoot
+
+open Finset Polynomial BigOperators
+
+/-- If `μ` is a primitive `n`th root of unity in `R`, then `∏(1≤k<n) (1-μ^k) = n`.
+(Stated with `n+1` in place of `n` to avoid the condition `n ≠ 0`.) -/
```
riccardobrasca, line 36: I am not sure using `n+1` is a good idea here. I can be better in principle, but looking at `Mathlib.RingTheory.RootsOfUnity.Basic` I see a bunch of `0<n` and `n : ℕ+`: this is already not great, but I am afraid that adding yet another way of saying the same thing is not great. Of course the real solution is to choose once and for all what to do...

Mathlib/RingTheory/RootsOfUnity/More.lean
```diff
@@ -0,0 +1,79 @@
...
+/-- If `μ` is a primitive `n`th root of unity in `R`, then `(-1)^(n-1) * ∏(1≤k<n) (μ^k-1) = n`.
+(Stated with `n+1` in place of `n` to avoid the condition `n ≠ 0`.) -/
+lemma prod_pow_sub_one_eq_order {n : ℕ} {μ : R} (hμ : IsPrimitiveRoot μ (n + 1)) :
+    (-1) ^ n * ∏ k ∈ range n, (μ ^ (k + 1) - 1) = n + 1 := by
+  have : (-1 : R) ^ n = ∏ k ∈ range n, -1 := by rw [prod_const, card_range]
+  simp only [this, ← prod_mul_distrib, neg_one_mul, neg_sub, ← prod_one_sub_pow_eq_order hμ]
+
+/-- If `μ` is a primitive `n`th root of unity in `R` and `k < n`, then `n` is divisible
+by `(μ-1)^k` in `ℤ[μ] ⊆ R`. -/
+lemma order_eq_mul_self_sub_one_pow {k n : ℕ} (hn : k < n) {μ : R} (hμ : IsPrimitiveRoot μ n) :
```
riccardobrasca, line 55: Again we can use `∣` (and `dvd` in the name).

riccardobrasca (approved):

Mathlib/RingTheory/RootsOfUnity/More.lean
MichaelStollBayreuth: The use of `n+1` to denote a positive natural is, I think, fairly common in Mathlib (and certainly better than using `PNat` in my opinion). The question is probably which version is easier to use. I have no strong opinion here (but hope that merging this will not be delayed because of a lack of consesus among the reviewers on this point :smile:).

Mathlib/RingTheory/RootsOfUnity/More.lean
MichaelStollBayreuth: See above. Using "dvd" in the name may still make sense, even though the statement does not mention `Dvd.dvd` explicitly.

Mathlib/RingTheory/RootsOfUnity/More.lean
MichaelStollBayreuth: I was also thinking of inlining this in the proof of `order_eq_mul_self_sub_one_pow`, as it is a fairly specialized statement (with a fairly short proof) whose usefulness outside `order_eq_mul_self_sub_one_pow` is perhaps questionable.

Mathlib/RingTheory/RootsOfUnity/More.lean
MichaelStollBayreuth: I have plans to eventually refactor roots of unity to avoid `PNat`, but this will have to wait until teaching is over.

Mathlib/RingTheory/RootsOfUnity/More.lean
MichaelStollBayreuth: I have now inlined `Algebra.adjoin.sub_one_dvd_pow_sub_one`.

Mathlib.lean
MichaelStollBayreuth: Changed the file name to "Lemmas.lean".

Mathlib/RingTheory/RootsOfUnity/More.lean
riccardobrasca: I think it is a good thing, we can maybe discuss it on zulip (for example, do we want to do the same in `IsCyclotomicExtension`? I think so).

riccardobrasca:
Thanks!

bors merge

riccardobrasca:
bors merge

PR #13205, opened 2024-05-25
[Merged by Bors] - chore: Rename more `coe_nat`/`coe_int` to `natCast`/`intCast`

YaelDillies:
---

- [x] depends on: #13197

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13292, opened 2024-05-27
copy of chore: Rename more `coe_nat`/`coe_int` to `natCast`/`intCast`

adomani:
A copy of #13205: the only difference is that I replaced a `refine'` by `refine` in `Mathlib.Order.Atoms.Finite`.  The reason is that the cache appears to be corrupted and forcing `lake` to rebuild `Mathlib.Order.Atoms.Finite` (or possibly a downstream file) makes the PR pass CI.

Authored-by: [YaelDillies](https://github.com/YaelDillies)

---


adomani:
maintainer merge

:smile:

sgouezel:
Now that #13205 has been merged, this can be closed, right?

adomani:
Yes, indeed!  Closing it.

PR #13190, opened 2024-05-24
[Merged by Bors] - feat: lint `attribute [instance] instName in`

grunweg:
Contrary to what users would expect, this defines a global instance.

This linter would have caught #13144 and #13170.
[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/ConcreteCategory.2EinstFunLike.20became.20a.20.20global.20instance)

Co-authored by: @adomani

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #7678, opened 2023-10-14
[Merged by Bors] - fix: `attribute [simp] ... in` -> `attribute [local simp] ... in`

collares:
Mathlib.Logic.Unique contains the line `attribute [simp] eq_iff_true_of_subsingleton in ...`:

https://github.com/leanprover-community/mathlib4/blob/96a11c7aac574c00370c2b3dab483cb676405c5d/Mathlib/Logic/Unique.lean#L255-L256

 Despite what the `in` part may imply, this adds the lemma to the simp set "globally", including for downstream files; it is likely that `attribute [local simp] eq_iff_true_of_subsingleton in ...` was meant instead (or maybe `scoped simp`, but I think "scoped" refers to the current namespace). Indeed, the relevant lemma is not marked with `@[simp]` for possible slowness: https://github.com/leanprover/std4/blob/846e9e1d6bb534774d1acd2dc430e70987da3c18/Std/Logic.lean#L749. Adding it to the simp set causes the example at https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Regression.20in.20simp to slow down.

This PR changes this and fixes the relevant downstream `simp`s. There was also one ocurrence of `attribute [simp] FullSubcategory.comp_def FullSubcategory.id_def in` in Mathlib.CategoryTheory.Monoidal.Subcategory but that was much easier to fix.

https://github.com/leanprover-community/mathlib4/blob/bc49eb9ba756a233370b4b68bcdedd60402f71ed/Mathlib/CategoryTheory/Monoidal/Subcategory.lean#L118-L119

---
There are other instances of `attribute ... in` in the codebase:

```
Topology/Category/TopCat/Basic.lean:attribute [instance] ConcreteCategory.funLike in
LinearAlgebra/BilinearForm/TensorProduct.lean:attribute [ext] TensorProduct.ext in
SetTheory/Game/Short.lean:attribute [-instance] subsingleton_short in
RingTheory/Trace.lean:attribute [-instance] Field.toEuclideanDomain
RingTheory/Polynomial/Basic.lean:attribute [-instance] Ring.toNonAssocRing in
RingTheory/Polynomial/Basic.lean:attribute [-instance] Ring.toSemiring in
RingTheory/Polynomial/Basic.lean:attribute [-instance] Polynomial.semiring Polynomial.commSemiring in
```

I don't know if those are OK. Also, does the `in` in `attribute [local/scoped simp] ... in` do anything, or do they operate on the current file/current namespace (respectively) regardless?

[...]

PR #13293, opened 2024-05-27
[Merged by Bors] - fix: extend attributeInstanceIn linter to lint all global attributes declared with `in`

adomani:
Extend the `AttributeInstanceIn` linter in two ways:
- allow for a list of attributes and flag the ones that are not `scoped` or `local`
- also lint any other attribute declared in an `attribute in` line which is neither `scoped` nor `local` (such as `ext` or `simp`).

This uncovered a global `TensorProduct.ext` attribute in `Mathlib.LinearAlgebra.BilinearForm.TensorProduct`.
It would have caught #7678 as well.

See #13190, especially the final comments by [Eric](https://github.com/leanprover-community/mathlib4/pull/13190#discussion_r1615652567) and [Mauricio](https://github.com/leanprover-community/mathlib4/pull/13190#issuecomment-2133615950).

```lean
theorem what : False := sorry

attribute [simp] what in
#guard true

-- the `simp` attribute on `what` persists
example : False := by simp  -- `simp` finds `what`

theorem who {x y : Nat} : x = y := sorry

attribute [ext] who in
#guard true

-- the `ext` attribute on `who` persists
example {x y : Nat} : x = y := by ext
```

---


adomani:
@eric-wieser, @collares, @grunweg: I am pinging you since you were also involved in the previous version of the linter.  This update makes the linter flag *all* attributes used with `in` and *not* `local` or `scoped`.  This is because all the examples that I tried actually made them global.

grunweg:
Thanks for acting so quickly. Extending the linter was on my list, but I would have needed a moment to understand `SepArray` etc. - I'm glad you took this on. The new tests look good; I didn't wrap my head around the extended linter yet. (Obviously, docs still need some adjustment.)

Should this PR also apply to `attribute [-instance] Ring.toNonAssocRing in` (found in `Mathlib/RingTheory/Polynomial/Basic.lean`)? I did not test this; in any case, a comment seems helpful.

test/attributeInstanceIn.lean
```diff
@@ -53,24 +51,44 @@ attribute [instance] Int.add in
...
+warning: Despite the `in`, the attribute 'simp' is added globally to 'foo'
+please remove the `in` or make this a `local instance` instead
+note: this linter can be disabled with `set_option linter.attributeInstanceIn false`
+---
+warning: Despite the `in`, the attribute 'ext' is added globally to 'foo'
+please remove the `in` or make this a `local instance` instead
+note: this linter can be disabled with `set_option linter.attributeInstanceIn false`
+-/
 #guard_msgs in
 -- TODO: where's the docBlame linter defined? what do I need to import?
```
grunweg, line 86: In `Batteries.Tactic.Lint.Misc`

grunweg: I agree that adding it would be a good test.

adomani: Unfortunately, I am not sure what should be done here with the `docBlame` linter... :upside_down_face:

adomani:
Michael, thanks for your comments!  I think that `eraseAttr` behaves as expected, in that its effect is limited to the scope implied by `in`.  I added a comment about this in the doc-module string.

test/attributeInstanceIn.lean
eric-wieser: I don't understand why `[nolint docBlame]` is something we care about.

eric-wieser (approved):
I pushed some cleanups, otherwise this looks good!

adomani:
Thanks!  Eric, your cleanups look great!

Mathlib/Tactic/Linter/AttributeInstanceIn.lean
```diff
@@ -41,6 +41,34 @@ attribute [ext] who in
 
 example {x y : Nat} : x = y := by ext
 ```
+
+For *removing* instances, the `in` seems to work as expected.
```
eric-wieser, line 45: I think this wording is overly strong; `attribute [-foo] in` very explicitly uses a feature of scoped attributes to locally remove the attribute within the current section.

eric-wieser:
```suggestion
For *removing* instances, the `in` works as expected.
```

eric-wieser:
maintainer merge

Mathlib/Tactic/Linter/AttributeInstanceIn.lean
```diff
@@ -41,28 +88,37 @@ namespace attributeInstanceLinter
...
-    Linter.logLint linter.attributeInstanceIn stx[0] m!
-      "`attribute [instance] ... in` declarations are surprising:\n\
-      they are **not** limited to the subsequent declaration, but define global instances\n\
-      please remove the `in` or make this a `local instance` instead"
+  for s in stx.topDown do
+    if let .some (id, nonScopedNorLocal) := getAttrInstance? s then
+      for attr in nonScopedNorLocal do
+        Linter.logLint linter.attributeInstanceIn attr m!
+          "Despite the `in`, the attribute '{attr}' is added globally to '{id}'\n\
+          please remove the `in` or make this a `local instance` instead"
```
fpvandoorn, line 121:
```suggestion
          please remove the `in` or make this a `local {attr}` instead"
```

Mathlib/Tactic/Linter/AttributeInstanceIn.lean
```diff
@@ -41,28 +88,37 @@ namespace attributeInstanceLinter
...
+      | `(Parser.Term.attrInstance| scoped $_attr:attr) => none
+      | `(attr| $a) => some a
+    (id, xs)
+  | _ => default
 
 /-- The `attributeInstanceLinter` linter flags any occurrence of `attribute [instance] name in`,
 including scoped or with instance priority: despite the `in`, these define *global* instances,
 which can be rather misleading.
-Instead, remove the `in` or make this a `local instance` instead.
+Instead, remove the `in` or make this a `local instance`.
```
fpvandoorn, line 109: please update the doc

fpvandoorn (commented):
LGTM, two comments

bors d+

test/attributeInstanceIn.lean
```diff
@@ -53,24 +51,44 @@ attribute [instance] Int.add in
...
 --attribute [nolint docBlame] foo in
```
grunweg, line 87:
```suggestion

File: test/attributeInstanceIn.lean [hunk 6]
```diff
@@ -53,24 +51,44 @@ attribute [instance] Int.add in
...
-attribute [simp] foo in
+set_option linter.attributeInstanceIn true in
+attribute [simp, local simp, ext, scoped instance, -simp, -ext] foo in
```
grunweg, line 89: *If* this was what you wanted to do: import `Batteries.Tactic.Lint.Misc`, adding this
```suggestion
attribute [simp, local simp, ext, nolint docBlame, scoped instance, -simp, -ext] foo in
```
and updating the #guard_msgs in. I don't think that's worth testing, though.

grunweg:
Thanks for the improvements, Eric, and for the review, Floris!

I incorporated the comments and made a pass over the entire PR, renaming it to the "globalAttributeIn" linter: it has grown in scope now, and is not limited to instance attributes any more. From my point, this looks good to go now.

Mathlib/Tactic/Linter/GlobalAttributeIn.lean
```diff
@@ -0,0 +1,130 @@
...
+      | `(Parser.Term.attrInstance| local $_attr:attr) => none
+      | `(Parser.Term.attrInstance| scoped $_attr:attr) => none
+      | `(attr| $a) => some a
+    (id, xs)
+  | _ => default
+
+/-- The `nonlocalAttributeInLinter` linter flags any occurrence of `attribute [instance] name in`,
+including scoped or with instance priority: despite the `in`, these define *global* instances,
+which can be rather misleading.
+Instead, remove the `in` or make this a `local instance`.
```
fpvandoorn, line 112: this docstring is still outdated

grunweg: Indeed, thanks for re-checking. Updated.

Mathlib/Tactic/Linter/GlobalAttributeIn.lean
```diff
@@ -0,0 +1,131 @@
...
+
+theorem who {x y : Nat} : x = y := sorry
+
+attribute [ext] who in
+#guard true
+
+-- the `ext` attribute persists
+example {x y : Nat} : x = y := by ext
+```
+Therefore, we lint against this patter on all instance.
```
adomani, line 46:
```suggestion
Therefore, we lint against this pattern on all instances.
```

adomani:
I'll wait for CI to finish to push the change, since interrupting a build seems correlated to getting cache failures.

adomani:
bors r+


I decided to take advantage of a rare passing of the CI and merged it with the typos!

PR #13294, opened 2024-05-27
[Merged by Bors] - feat: add pointwise actions on ideals

eric-wieser:
I've copied the API from `Subring`, but some proofs from there don't work here due to `Ideal.map` including an extra `Ideal.span`.

---


Mathlib/RingTheory/Ideal/Pointwise.lean
```diff
@@ -0,0 +1,115 @@
...
+
+This file provides the action `Ideal.pointwiseMulAction` which morally matches the action of
+`mulActionSet` (though here an extra `Ideal.span` is inserted).
+
+This actions is available in the `Pointwise` locale.
+
+## Implementation notes
+
+This file is similar (but not identical) to `RingTheory/Subring/Pointwise.lean`.
+Where possible, try to keep them in sync.
```
kbuzzard, line 19: This file doesn't seem to exist. Do you mean `Algebra/Ring/Subsemiring/Pointwise.lean`? I was going to say "you might want to put a note in that file suggesting that it be kept in sync with this one", but then I looked at that file and there's already a note saying that it should be kept in sync with a third file, which also doesn't seem to exist.

eric-wieser: I think @YaelDillies move all the files around, but the comments were left with the old names.

YaelDillies: My regex assumed files were referred to as `A.B.C`, not `A/B/C` 😐

kbuzzard: OK but just to be clear: purely for the purposes of reviewing this PR, which file am I supposed to be checking that it's similar to?

YaelDillies: Wherever you see a missing file, try replacing `RingTheory` by `Algebra.Ring`

kbuzzard: And `GroupTheory/Submonoid/Pointwise.lean` (mentioned in the subsemiring file) -> `Algebra/Group/Submonoid/Pointwise.lean` presumably?

kbuzzard: #13469

kbuzzard: So should the file in this PR be called `Mathlib/Algebra/Ring/Ideal/Pointwise.lean`?

Mathlib/RingTheory/Ideal/Pointwise.lean
```diff
@@ -0,0 +1,115 @@
...
+
+section Monoid
+
+variable [Monoid M] [CommRing R] [MulSemiringAction M R]
+
+/-- The action on an ideal corresponding to applying the action to every element.
+
+This is available as an instance in the `Pointwise` locale. -/
+protected def pointwiseMulSemiringAction : MulSemiringAction M (Ideal R) where
+  smul a I := Ideal.map (MulSemiringAction.toRingHom _ _ a) I
```
kbuzzard, line 38:
```suggestion
  smul a I := I.map (MulSemiringAction.toRingHom _ _ a)
```
is there a reason you went for the more longwinded spelling?

Mathlib/RingTheory/Ideal/Pointwise.lean
```diff
@@ -0,0 +1,115 @@
...
+  one_smul I :=
+    congr_arg (I.map ·) (RingHom.ext <| one_smul M) |>.trans I.map_id
+  mul_smul _a₁ _a₂ I :=
```
kbuzzard, line 41: Is there a reason you used `_a₁` and not just `_`?

Mathlib/RingTheory/Ideal/Pointwise.lean
```diff
@@ -0,0 +1,115 @@
...
+
+scoped[Pointwise] attribute [instance] Ideal.pointwiseMulSemiringAction
+
+open Pointwise
+
+theorem pointwise_smul_def {a : M} (S : Ideal R) :
+    a • S = S.map (MulSemiringAction.toRingHom _ _ a) :=
+  rfl
+
+-- note: unlike with `Subring`, `pointwise_smul_toAddSubgroup` wouldn't be true
```
kbuzzard, line 56: It's true for group actions though, right? Should it be in the `Group` part of the file? Similar comment for `coe_pointwise_smul`.

This observation is perhaps an indication that the correct definition for monoids is to use comap and let them act on the right? But I'm not advocating for this!

kbuzzard (approved):

Mathlib/RingTheory/Ideal/Pointwise.lean
kbuzzard: Furthermore, using comap on a^{-1} for the definition works much better definitionally in the case where M is a group. I'm just moving from my own implementation to this and noticing this annoyance. Monoids can act by endomorphisms; groups have to act via automorphisms. We have to sacrifice nice definitional properties just because we want some generalization to monoids which in practice mathematicians may never use? :-(

eric-wieser: > Furthermore, using comap on a^{-1} for the definition works much better definitionally in the case where M is a group.

The same could be said for pretty much all the pointwise actions - I took the map approach here mainly for consistency.

> We have to sacrifice nice definitional properties just because we want some generalization to monoids which in practice mathematicians may never use? :-(

This already happened with the definition of `Disjoint`, and surprisingly I haven't really noticed anyone complaining...

Mathlib/RingTheory/Ideal/Pointwise.lean
eric-wieser: Yes, to make the goal view easier to understand. It seems the linter no longer requires the `_`s though.

Mathlib/RingTheory/Ideal/Pointwise.lean
eric-wieser: I added a weaker version.

Mathlib/RingTheory/Ideal/Pointwise.lean
eric-wieser: > just because we want some generalization to monoids which in practice mathematicians may never use?

Maybe `MulSemiringAction` is to blame here; are there any examples of such actions which are a monoid action and not a group action?

kbuzzard: Probably the endomorphisms of a commutative ring act on the ideals of the ring and I could imagine that someone might need that.

kbuzzard:
As far as I am concerned this is ready, modulo the fact that Yael moved various files from RingTheory into Alegebra/Ring, so my one remaining question is whether the file in this PR should be called `Mathlib/Algebra/Ring/Ideal/Pointwise.lean`.

Mathlib/RingTheory/Ideal/Pointwise.lean
eric-wieser: Perhaps it should be called that in a future PR, but for now the important thing is that it live next to `RingTheory/Ideal/Basic.lean`, which is where I have it now. If @YaelDillies  wants to move all of the files somewhere else in a later PR I'm happy to stay out of their way!

kbuzzard:
Thanks!

bors merge

PR #13295, opened 2024-05-28
[Merged by Bors] - refactor: explicit args in `Finset.cons_induction` etc

urkud:
This way we don't need `@cons` in `induction` calls.

---


sgouezel:
bors r+
Thanks!

YaelDillies:
I'm not convinced. Isn't the better option to have `{a}` syntax for `induction`, like we do for pattern matching?

PR #13313, opened 2024-05-28
[Merged by Bors] - feat(CategoryTheory/Monoidal): convolution monoid

kim-em:
I'll use this shortly to prove basic properties of hopf monoids.

- [x] depends on: #13316


Mathlib/CategoryTheory/Monoidal/Conv.lean
```diff
@@ -0,0 +1,62 @@
...
+instance : One (Conv M N) where
+  one := M.counit ≫ N.one
+
+@[simp, nolint simpNF] -- This simpNF incorrectly claims this simp lemma can not be applied.
+theorem one_eq : (1 : Conv M N) = M.counit ≫ N.one := rfl
+
+instance : Mul (Conv M N) where
+  mul := fun f g => M.comul ≫ f ▷ M.X ≫ N.X ◁ g ≫ N.mul
+
+@[simp] theorem mul_eq (f g : Conv M N) : f * g = M.comul ≫ f ▷ M.X ≫ N.X ◁ g ≫ N.mul := rfl
```
joelriou, line 37: Would not it be slightly better not to declare this as a global simp? (And similarly for `one_eq`.) Otherwise, when we do computations in this monoid, everything would be `dsimp`lified into these formulas, which would make the use of algebraic relations (associativity, etc) difficult. I would think that the user should explicitly trigger these lemmas when needed.

semorrison: Thank you, I agree.

dagurtomas (approved):
LGTM

joelriou:
Thanks!

bors merge

PR #13315, opened 2024-05-28
[Merged by Bors] - feat(Algebra/Category/CoalgebraCat): minor changes

kim-em:
It turns out @101damnations and I had independently written `CoalgebraCat`. She got in first, and I'm happy to use that version but want one more instance.

PR #13312, opened 2024-05-28
[Merged by Bors] - chore(CategoryTheory/Monoidal): oplax functors and comonoids

kim-em:
Oplax monoidal functors take comonoids to comonoids. Also, generalize constructions of limits in `Mon_` so limits of a particular shape in `C` are enough to give limits of that shape in `Mon_ C`.

- [x] depends on: #13310
- [x] depends on: #13316

PR #13317, opened 2024-05-28
[Merged by Bors] - feat(CategoryTheory/Monoidal): Hopf monoids

kim-em:
Defines Hopf monoids in a braided category.

Proves two facts (antipode is an antihomomorphism, morphisms intertwine antipode) that are not yet proved for unbundled Hopf algebras. Once we prove `Hopf_ (ModuleCat R)` is equivalent to `HopfAlgebraCat` (having defined that!), we will get those two facts for free.

We should also check later that `Hopf_ C` in a cartesian monoidal category is equivalent `GroupObject_ C` (being defined independently elsewhere).

- [x] depends on: #13313
- [x] depends on: #13312 (probably doesn't actually depend, but I don't want to sort out the git history now)
- [x] depends on: #13315


Mathlib/CategoryTheory/Monoidal/Hopf_.lean
```diff
@@ -0,0 +1,465 @@
...
+  slice_lhs 3 5 =>
+    rw [← associator_naturality_right, ← Category.assoc, ← tensorHom_def]
+  slice_lhs 3 9 =>
+    erw [Bimon_.compatibility]
+  slice_lhs 1 3 =>
+    erw [A.antipode_left]
+  simp
+
+/--
+This calculation calls for some ASCII art out of This Week's Finds.
```
dagurtomas, line 130:
```suggestion
/--
Auxiliary calculation for `antipode_comul`.
This calculation calls for some ASCII art out of This Week's Finds.
```

dagurtomas (approved):
I admit that I didn't read the proofs closely, but the statements look good, the file is structured nicely and is reasonably fast. I just had one minor comment about a docstring, otherwise LGTM. Maybe it's best also to merge master before this is sent to bors, since it's been sitting for a while.

dagurtomas:
Thanks!

maintainer merge

dagurtomas:
Let's try again:

maintainer merge

jcommelin:
Thanks :tada:

bors merge

PR #13318, opened 2024-05-28
[Merged by Bors] - feat(CategoryTheory/Sites): more properties of locally surjective morphisms of presheaves

joelriou:
`IsLocallySurjective` is made a type-class, and basic properties (similar to properties of injective/surjective maps) are obtained for locally injective/surjective morphisms of presheaves on a category equipped with a Grothendieck topology.

---

- [x] depends on: #11719

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13324, opened 2024-05-28
[Merged by Bors] - feat(CategoryTheory/Sites): locally bijective morphisms of presheaves

joelriou:
Let `C` a category equipped with a Grothendieck topology `J`. Let `A` be a concrete category. In this file, we introduce a type-class `J.WEqualsLocallyBijective A` which says that the class `J.W` (of morphisms of presheaves which become isomorphisms after sheafification) is the class of morphisms that are both locally injective and locally surjective (i.e. locally bijective). We prove that this holds iff for any presheaf `P : Cᵒᵖ ⥤ A`, the sheafification map `toSheafify J P` is locally bijective. We show that this holds under certain universe assumptions.

---

- [x] depends on: #13318


Mathlib/CategoryTheory/Sites/LocallyBijective.lean
```diff
@@ -0,0 +1,152 @@
...
+namespace Sheaf
+
+section
+
+variable {F G : Sheaf J (Type w)} (f : F ⟶ G)
+
+/-- A morphism of sheaves of types is locally bijective iff it is an isomorphism.
+(This is generalized below as `isLocallyBijective_iff_isIso`.) -/
+private lemma isLocallyBijective_iff_isIso' :
+    IsLocallyInjective f ∧ IsLocallySurjective f ↔ IsIso f := by
```
erdOne, line 38: Doesn't this follow from `IsLocallyInjective f ↔ mono f` and `IsLocallyInjective f ↔ epi f` and that `Sheaf J (Type w)` is balanced?

joelriou: This lemma does not use `HasSheafify J (Type w)`, but `IsLocallySurjective f ↔ epi f` does, so that I would prefer keeping the statement of this lemma as is.

Mathlib/CategoryTheory/Sites/LocallyBijective.lean
```diff
@@ -0,0 +1,152 @@
...
+      intro Y₁ Y₂ W g₁ g₂ f₁ f₂ hf₁ hf₂ w
+      apply h₁
+      have eq₁ := FunctorToTypes.naturality _ _ f.val g₁.op (t f₁ hf₁)
+      have eq₂ := FunctorToTypes.naturality _ _ f.val g₂.op (t f₂ hf₂)
+      have eq₃ := congr_arg (G.val.map g₁.op) (Presheaf.app_localPreimage f.val s _ hf₁)
+      have eq₄ := congr_arg (G.val.map g₂.op) (Presheaf.app_localPreimage f.val s _ hf₂)
+      refine' eq₁.trans (eq₃.trans (Eq.trans _ (eq₄.symm.trans eq₂.symm)))
+      erw [← FunctorToTypes.map_comp_apply, ← FunctorToTypes.map_comp_apply]
+      simp only [← op_comp, w]
+    refine' ⟨H.amalgamate t ht, _⟩
```
dagurtomas, line 61:
```suggestion
    refine ⟨H.amalgamate t ht, ?_⟩
```

Mathlib/CategoryTheory/Sites/LocallyBijective.lean
```diff
@@ -0,0 +1,152 @@
...
+    let t : Presieve.FamilyOfElements F.val (Presheaf.imageSieve f.val s).arrows :=
+      fun Y g hg => Presheaf.localPreimage f.val s g hg
+    have ht : t.Compatible := by
+      intro Y₁ Y₂ W g₁ g₂ f₁ f₂ hf₁ hf₂ w
+      apply h₁
+      have eq₁ := FunctorToTypes.naturality _ _ f.val g₁.op (t f₁ hf₁)
+      have eq₂ := FunctorToTypes.naturality _ _ f.val g₂.op (t f₂ hf₂)
+      have eq₃ := congr_arg (G.val.map g₁.op) (Presheaf.app_localPreimage f.val s _ hf₁)
+      have eq₄ := congr_arg (G.val.map g₂.op) (Presheaf.app_localPreimage f.val s _ hf₂)
+      refine' eq₁.trans (eq₃.trans (Eq.trans _ (eq₄.symm.trans eq₂.symm)))
```
dagurtomas, line 58:
```suggestion
      refine eq₁.trans (eq₃.trans (Eq.trans ?_ (eq₄.symm.trans eq₂.symm)))
```

dagurtomas (approved):
LGTM! I need #13396 (which depends on this) for applications in condensed mathematics.

riccardobrasca:
Thanks!

bors merge

joelriou:
> How often will this be used? Is it worth adding a new abbrev `Presheaf.IsLocallyBijective J f := Presheaf.IsLocallyInjective J f ∧ Presheaf.IsLocallySurjective J f `?

We could do it, but as `IsLocallyInjective/Surjective` are type-classes, I would rather not introduce this abbreviation. Lemmas typically use the local injectivitity or the local surjectivity, but not that many use both assumptions.

PR #13329, opened 2024-05-29
[Merged by Bors] - chore(Combinatorics/Additive/PluenneckeRuzsa): Use semantic lemma names

YaelDillies:
I, as the main author of those theorems, can never remember their names. Symbol-reading as a naming convention fails here because there are too many symbols to mention to
* distinguish the series of results from other results in Mathlib
* distinguish the different results within that series

Using semantic names solves both issues quite nicely.

---

### DIFF
--- a/Mathlib/Combinatorics/Additive/PluenneckeRuzsa.lean
+++ b/Mathlib/Combinatorics/Additive/PluenneckeRuzsa.lean
@@ -19,10 +19,10 @@ inequality.
 
 ## Main declarations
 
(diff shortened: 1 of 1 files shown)


semorrison:
Deprecations?

(I won't insist here if you're sure no one is using these downstream yet. But I wouldn't want anyone to think I was shirking on deprecations. :-)

bors d+

YaelDillies:
There's one person using them downstream: me 😄. I am very sure of that because the current statements are a bit weird and anyone trying to use them would have complained to me.

I don't have a huge lot of time to add deprecations that certainly no one will use, so I'm going to declare this fine.

bors merge

PR #13331, opened 2024-05-29
[Merged by Bors] - feat(CategoryTheory/MorphismProperty): the factorization axiom

joelriou:
In this PR, we formalize the factorization axiom which asserts that a morphism in a category can be factored as a morphism in a certain class `W₁` followed by a morphism in a class `W₂`. A functorial version of the axiom is also  defined.

---


Mathlib/CategoryTheory/MorphismProperty/Factorization.lean
```diff
@@ -0,0 +1,203 @@
...
+attribute [reassoc (attr := simp)] MapFactorizationData.fac
+
+/-- The data of a term in `MapFactorizationData W₁ W₂ f` for any morphism `f`. -/
+abbrev FactorizationData := ∀ {X Y : C} (f : X ⟶ Y), MapFactorizationData W₁ W₂ f
+
+/-- The factorization axiom for two classes of morphisms `W₁` and `W₂` in a category `C`. It
+asserts that any morphism can be factored as a morphism in `W₁` followed by a morphism
+in `W₂`. -/
+class HasFactorization : Prop where
+  nonempty_mapFactorizationData {X Y : C} (f : X ⟶ Y) : Nonempty (MapFactorizationData W₁ W₂ f)
```
erdOne, line 59: Maybe we should make `fun {X Y : C} (f : X ⟶ Y) => Nonempty (MapFactorizationData W₁ W₂ f)` into a `MorphismProperty`? It will be useful for say immersions and quasi-projective morphisms (on Hartshorne).

Looking at the material below, I think this definition is quite orthogonal to them and will not shorten any proof though. What do you think?

joelriou: Thanks for the suggestion. I have added `MorphismProperty.comp` and a lemme `comp_eq_top_iff`. It also seems to me that it is essentially independent of the rest of the code though.

erdOne: Yeah I was expecting the answer to be "the idea might have potential but perhaps in a separate PR" but I brought it up just in case.

erdOne:
Thanks!
maintainer merge

jcommelin (approved):
Thanks :tada:

bors merge

PR #13319, opened 2024-05-28
[Merged by Bors] - refactor(LightProfinite): redefine light profinite spaces as second countable profinite spaces

dagurtomas:
Changes the definition of `LightProfinite` to the category of second countable totally disconnected compact Hausdorff spaces. The old `LightProfinite` is renamed to `LightDiagram` and we give an equivalence of categories between them. 

This changes `LightProfinite.Basic` to match `Profinite.Basic` more closely. It also contains the equivalence of categories with `LightDiagram` and an `EssentiallySmall` instance. 

The `Limits` and `EffectiveEpi` files now match their counterparts for `Profinite` more closely as well. 

This PR also adds a new feature: `LightProfinite` has countable limits and the functor to `Profinite` creates countable limits.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13333, opened 2024-05-29
[Merged by Bors] - feat(LightProfinite): light profinite sets as sequential limits with surjective transition maps

dagurtomas:
Adds the file `LightProfinite/AsLimit` which mirrors `Profinite/AsLimit`.

---
- [x] depends on: #13319 


Mathlib/CategoryTheory/Limits/Cones.lean
```diff
@@ -335,6 +335,14 @@ theorem cone_iso_of_hom_iso {K : J ⥤ C} {c d : Cone K} (f : c ⟶ d) [i : IsIs
         w := fun j => (asIso f.hom).inv_comp_eq.2 (f.w j).symm }, by aesop_cat⟩⟩
 #align category_theory.limits.cones.cone_iso_of_hom_iso CategoryTheory.Limits.Cones.cone_iso_of_hom_iso
 
+/-- Given an isomorphism of cones, produce an isomorphism of their points. -/
+@[simps]
+def ptIsoOfIso  {K : J ⥤ C} {c d : Cone K} (i : c ≅ d) : c.pt ≅ d.pt where
```
semorrison, line 340: I'm slightly cautious about expanding the API with this one. (i.e. If we introduce it, what lemmas should we be writing, or feel we been later, relating it to other existing things. e.g. if you start with an iso between the points of two limits cones, obtain the iso of cones, and then apply this you are back where you started.) Similarly -- what about the dual?

If you can think of a way to avoid introducing this I'd be perfectly happy with the rest of the PR.

dagurtomas: One should probably use the "forgetful" functor from cones to the target category instead. We have this, right? I might have time to fix this later today.

dagurtomas: Ok, `ptIsoOfIso` is gone

semorrison:
bors merge

PR #13336, opened 2024-05-29
[Merged by Bors] - chore: deprecate Nat.div_mod_eq_mod_mul_div and Nat.mul_div_mul_comm_of_dvd_dvd

dwrensha:
* `div_mod_eq_mod_mul_div` is a flipped version of [`mod_mul_right_div_self`](https://github.com/leanprover/lean4/blob/91334702431b8a7bf07f3f9e5696603de7b11fb6/src/Init/Data/Nat/Mod.lean#L50).
* `mul_div_mul_comm_of_dvd_dvd` is a flipped version of [`div_mul_div_comm`](https://github.com/leanprover-community/mathlib4/blob/b6cb80af68d53667dbbe8267f20ca075d5ffa953/Mathlib/Data/Nat/Defs.lean#L583).

These were found by [tryAtEachStep](https://github.com/dwrensha/tryAtEachStep).

---

### DIFF
--- a/Mathlib/Algebra/Ring/Parity.lean
+++ b/Mathlib/Algebra/Ring/Parity.lean
@@ -335,7 +335,7 @@ lemma Odd.of_mul_right (h : Odd (m * n)) : Odd n :=
 #align nat.odd.of_mul_right Nat.Odd.of_mul_right
 
 lemma even_div : Even (m / n) ↔ m % (2 * n) / n = 0 := by
-  rw [even_iff_two_dvd, dvd_iff_mod_eq_zero, ←Nat.mod_mul_right_div_self, mul_comm]
+  rw [even_iff_two_dvd, dvd_iff_mod_eq_zero, ← Nat.mod_mul_right_div_self, mul_comm]
 #align nat.even_div Nat.even_div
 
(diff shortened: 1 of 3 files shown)


grunweg (commented):
Thanks! This raises the question whether having both lemmas is useful - but I think this can also happen in a follow-up.
maintainer merge

sgouezel:
I'd rather remove the lemmas (and deprecate them properly).

dwrensha:
Deprecated in https://github.com/leanprover-community/mathlib4/pull/13336/commits/671100db12b1e4dde3d988629c969001a9d7d358.

I wanted to replace the lemmas with `alias`, because I don't think it's possible because then there's no way to apply the `symm`.

riccardobrasca:
Thanks!

bors merge

YaelDillies:
`Nat.mul_div_mul_comm_of_dvd_dvd`/`Nat.div_mul_div_comm` is supposed to mimic `mul_div_mul_comm`/`div_mul_div_comm`. Can you revert? (or maybe reintroduce `Nat.mul_div_mul_comm` instead of `Nat.mul_div_mul_comm_of_dvd_dvd` since I agree the name is strange)

dwrensha:
https://github.com/leanprover-community/mathlib4/pull/13833

PR #13201, opened 2024-05-25
[Merged by Bors] - refactor(Order/OmegaCompletePartialOrder): Unify concepts of Scott Continuity

mans0954:
We currently have two related but different concepts of Scott Continuity in Mathlib - `ScottContinuous` in `Order/Bounds/Basic` and `OmegaCompletePartialOrder.Continuous'` in `Order/OmegaCompletePartialOrder`. We previously showed (#6831) that `ScottContinuous` implies `OmegaCompletePartialOrder.Continuous'` (in general the converse isn't true).

This PR gives a more general definition of Scott Continuity, `ScottContinuousOn` which specialises to `ScottContinuous` and something equivalent to  `OmegaCompletePartialOrder.Continuous'` to give a more satisfactory unification of the concepts.

---

- [x] depends on: #13337
- [x] depends on: #13338

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13338, opened 2024-05-29
[Merged by Bors] - chore(Order): More simp lemmas

YaelDillies:
These lemmas are needed in #13201

---

### DIFF
--- a/Mathlib/Order/CompleteLattice.lean
+++ b/Mathlib/Order/CompleteLattice.lean
@@ -1738,11 +1738,13 @@ instance Pi.instCompleteLattice {α : Type*} {β : α → Type*} [∀ i, Complet
   le_sInf _ _ hf := fun i => le_iInf fun g => hf g g.2 i
 #align pi.complete_lattice Pi.instCompleteLattice
 
+@[simp]
 theorem sSup_apply {α : Type*} {β : α → Type*} [∀ i, SupSet (β i)] {s : Set (∀ a, β a)} {a : α} :
     (sSup s) a = ⨆ f : s, (f : ∀ a, β a) a :=
   rfl
 #align Sup_apply sSup_apply
 
+@[simp]
 theorem sInf_apply {α : Type*} {β : α → Type*} [∀ i, InfSet (β i)] {s : Set (∀ a, β a)} {a : α} :
     sInf s a = ⨅ f : s, (f : ∀ a, β a) a :=
   rfl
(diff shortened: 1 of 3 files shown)


Mathlib/Order/CompleteLattice.lean
```diff
@@ -1787,15 +1789,33 @@ theorem binary_relation_sInf_iff {α β : Type*} (s : Set (α → β → Prop))
...
+protected lemma Antitone.sSup (hs : ∀ f ∈ s, Antitone f) : Antitone (sSup s) :=
+  fun _ _ h ↦ iSup_mono fun f ↦ hs f f.2 h
+
+protected lemma Antitone.sInf (hs : ∀ f ∈ s, Antitone f) : Antitone (sInf s) :=
+  fun _ _ h ↦ iInf_mono fun f ↦ hs f f.2 h
 
-theorem monotone_sSup_of_monotone {s : Set (α → β)} (m_s : ∀ f ∈ s, Monotone f) :
-    Monotone (sSup s) := fun _ _ h => iSup_mono fun f => m_s f f.2 h
-#align monotone_Sup_of_monotone monotone_sSup_of_monotone
+@[deprecated (since := "2024-05-29")] alias monotone_sSup_of_monotone := Monotone.sSup
```
b-mehta, line 1808: Since dot notation isn't actually useful here, I'm not sure deprecation is really deserved.

Mathlib/Order/Hom/Basic.lean
```diff
@@ -344,6 +344,9 @@ theorem comp_mono ⦃g₁ g₂ : β →o γ⦄ (hg : g₁ ≤ g₂) ⦃f₁ f₂
     g₁.comp f₁ ≤ g₂.comp f₂ := fun _ => (hg _).trans (g₂.mono <| hf _)
 #align order_hom.comp_mono OrderHom.comp_mono
 
+@[simp] lemma mk_comp_mk (g : β → γ) (f : α → β) (hg hf) :
```
b-mehta, line 347:
```suggestion
@[simp] lemma mk_comp_mk {g : β → γ} {f : α → β} (hg hf) :
```

YaelDillies: Not sure I agree. `f` AND `hf` are explicit arguments to `mk`. Further, we expect `f` to be short and `hf` to be long and `hf` is inferrable from the rewrite (assuming we rewrite forward), so it's easier to specify `f` than `hf`.

Mathlib/Order/CompleteLattice.lean
YaelDillies: It is useful thanks to the new expected type dot notation. `.sSup _ : Monotone _` should* elaborate to `Monotone.sSup`

* seems like there currently is a bug due to `Monotone` being defined as a forall. Not sure it's known

b-mehta: Let's not deprecate until the bug is reported then.

b-mehta:
bors d+

Awaiting the report of the bug above.

Mathlib/Order/CompleteLattice.lean
YaelDillies: https://github.com/leanprover/lean4/issues/4761

YaelDillies:
bors merge

PR #13245, opened 2024-05-26
feat: rewrite `update-style-exceptions.sh` as a Lean executable

grunweg:
This allows running both the Python and Lean style linters.
As a user, replace `./scripts/update-style-exceptions.sh` by `lake exe update_style_exceptions`.

The Lean version is slightly slower (1s vs 0.5), presumably due to some overhead of spawning processes/using `xargs` the way I do. I'm not inclined to worry.

-------

- [x] depends on: #13339 to remove the copy-pasted code from `mk_all.lean`

PR #11853, opened 2024-04-02
[Merged by Bors] - feat: `lake exe mk_all` as a Lean executable

adomani:
Running
```bash
lake exe mk_all --git
```
should have the same effect as running
```bash
./script/mk_all.sh
```

that is, it creates the files `Mathlib.lean`, `Mathlib/Tactic.lean`, `Archive.lean`, `Counterexamples.lean`.

It does *not* create analogous files for `Cache` and `LongestPole`.

```bash
lake exe mk_all
```
is similar, but uses all the `.lean` files in `Mathlib`, not just the Git-managed ones.

See #11874 for using the script in CI.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13339, opened 2024-05-29
[Merged by Bors] - refactor: tweak and move functionality for getting all modules out of `scripts`

grunweg:
With #13245, two scripts in `scripts` would like to use very similar information:
move it to a central place instead (a file in `Mathlib/Util`). Suggestions for better locations are welcome!

Also, split the `getAll` function in two pieces, getting the file names and transforming this into module names.
While #11853 only requires the latter, #13245 needs the former (which is computed anyway).
Only accessible the converted module names and converting them back to files would be absurd.

- [x] depends on: #11853 which adds these utility functions

---

### DIFF
--- a/Mathlib/Util/GetAllModules.lean
+++ b/Mathlib/Util/GetAllModules.lean
@@ -41,12 +41,12 @@ def getAllFiles (git : Bool) (ml : String) : IO (Array System.FilePath) := do
       let all ← walkDir ml
       return all.filter (·.extension == some "lean"))
   let files := (allModules.erase ml.lean).qsort (·.toString < ·.toString)
-  let withImport ← files.mapM fun f => do
+  let existingFiles ← files.mapM fun f => do
     -- this check is helpful in case the `git` option is on and a local file has been removed
     if ← pathExists f then
       return f
     else return ""
-  return withImport.filter (· != "")
+  return existingFiles.filter (· != "")
 
 /-- Like `getAllFiles`, but return an array of *module* names instead,
 i.e. names of the form `"Mathlib.Algebra.Algebra.Basic"`. -/
--- a/Mathlib/Util/GetAllModules.lean
+++ b/Mathlib/Util/GetAllModules.lean
@@ -1,0 +1,72 @@
+/-
+Copyright (c) 2024 Damiano Testa. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro, Kim Morrison, Damiano Testa
+-/
+
+import Lean.Util.Path
+import Lake.CLI.Main
+
+
+/-!
+# Utility functions for finding all `.lean` files or modules in a project.
+
+TODO:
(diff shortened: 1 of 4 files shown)


Mathlib/Util/GetAllModules.lean
```diff
@@ -0,0 +1,72 @@
...
+  let allModules : Array System.FilePath ← (do
+    if git then
+      let mlDir := ml.push pathSeparator   -- for example, `Mathlib/`
+      let allLean ← IO.Process.run { cmd := "git", args := #["ls-files", mlDir ++ "*.lean"] }
+      return (((allLean.dropRightWhile (· == '\n')).splitOn "\n").map (⟨·⟩)).toArray
+    else do
+      let all ← walkDir ml
+      return all.filter (·.extension == some "lean"))
+  let files := (allModules.erase ml.lean).qsort (·.toString < ·.toString)
+  let withImport ← files.mapM fun f => do
```
adomani, line 44:
```suggestion
  let withImport ← files.mapM fun f => do
```
The entries in the array no longer have `import` in them, so maybe it can be renamed to `existingFiles` or something similar?

adomani (approved):

adomani:
This looks good to me!  The rename of the variable should have really happened when I refactored the code, by I only noticed it now.

Thanks!

(I won't be at a computer until Monday, so I can't really test this code right now.)

Mathlib/Util/GetAllModules.lean
grunweg: Good point, changed (and tested locally).

grunweg:
For the future, `resolveDefaultRootModules` in leanprover-community/batteries#811seems related, but different.

Mathlib/Util/GetAllModules.lean
```diff
@@ -0,0 +1,72 @@
...
+If the current project is `mathlib`, then it excludes the libraries `Cache` and `LongestPole` and
+it includes `Mathlib/Tactic`. -/
+def getLeanLibs : IO (Array String) := do
+  let (elanInstall?, leanInstall?, lakeInstall?) ← findInstall?
+  let config ← MonadError.runEIO <| mkLoadConfig { elanInstall?, leanInstall?, lakeInstall? }
+  let ws ← MonadError.runEIO (MainM.runLogIO (loadWorkspace config)).toEIO
+  let package := ws.root
+  let libs := (package.leanLibs.map (·.name)).map (·.toString)
+  return if package.name == `mathlib then
+    libs.erase "Cache" |>.erase "LongestPole" |>.push ("Mathlib".push pathSeparator ++ "Tactic")
```
semorrison, line 70: I wonder if this deserves a note in the lakefile saying that if you add more `lean_exe`s they probably should be added here?

semorrison:
bors d+

Mathlib/Util/GetAllModules.lean
grunweg: Good point! I presume you mean `lean_lib`s, right - added a comment about this.

PR #13343, opened 2024-05-29
[Merged by Bors] - fix(`nolints.yml`): correct the main step of the `update-nolints` workflow

grunweg:
This fixes the main step of the `update_nolints` workflow.
As I see it, this was quite impressive: about three bugs in four lines of code.
- updating `nolints.json` now works using the `update` option: no need to run the linter and move some file (presumably, that file wouldn't exist either)
- running `git diff` right after generating the exceptions didn't work: no git repository was configured yet
- wrapping things in `problem-matcher-wrap` was a bad idea, as that only supports one command: instead, we remove the matcher, as we don't need this here

Since this workflow was originally cargo-culted from the mathlib one, I view this also as a cautionary tale about cargo cult programming.

---------

The new incantation was tested in #13346, so I'm confident this one actually works.

### DIFF
--- a/.github/workflows/nolints.yml
+++ b/.github/workflows/nolints.yml
@@ -70,7 +70,7 @@ jobs:
           # github using a different username.
           git config --unset http.https://github.com/.extraheader
 
-      - name: file a PR with the updated nolints.json and style-exceptions.txt
+      - name: file a new PR to update nolints.json and style-exceptions.txt
         run: ./scripts/update_nolints_CI.sh
         env:
           DEPLOY_GITHUB_TOKEN: ${{ secrets.UPDATE_NOLINTS_TOKEN }}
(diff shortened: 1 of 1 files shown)


Ruben-VandeVelde (approved):
Let's try

maintainer merge

.github/workflows/nolints.yml
```diff
@@ -75,7 +70,7 @@ jobs:
           # github using a different username.
           git config --unset http.https://github.com/.extraheader
 
-      - name: update nolints.json and style-exceptions.txt
+      - name: file a PR with the updated nolints.json and style-exceptions.txt
```
grunweg, line 73:
```suggestion
      - name: file a new PR to update nolints.json and style-exceptions.txt

semorrison:
bors merge

PR #13346, opened 2024-05-29
CI test: nolint update workflow

grunweg:
Test my theory about the nolint workflow failing: let's run my proposed fixed logic in the normal lint job and see how it fares.

This should be noisy, but hopefully the good kind of noise :-)

---


grunweg:
Ok: it seems the issue is not resolved yet. Help welcome...

grunweg:
I've found the incantation which works: this PR has served its purpose and can be closed.

PR #13347, opened 2024-05-29
[Merged by Bors] - feat(NumberTheory/LSeries): special values of Hurwitz zeta

loefflerd:
Formulae for Hurwitz zeta functions at negative integers in terms of Bernoulli polynomials.

---

### DIFF
--- a/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
+++ b/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
@@ -102,9 +102,9 @@ theorem cosZeta_two_mul_nat' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
     rw [(by { norm_cast; omega } : 2 * (k : ℂ) = ↑(2 * k - 1) + 1), Complex.Gamma_nat_eq_factorial,
       ← Nat.cast_add_one, ← Nat.cast_mul, ← Nat.factorial_succ, Nat.sub_add_cancel (by omega)]
   simp_rw [this, Gammaℂ, cpow_neg, ← div_div, div_inv_eq_mul, div_mul_eq_mul_div, div_div]
-  norm_num
-  rw [(by simp : 2 * (k : ℂ) = ↑(2 * k)), cpow_natCast]
-  ring
+  congr 2
+  · rw [(by simp : 2 * (k : ℂ) = ↑(2 * k)), cpow_natCast]
+  · ring
 
 /-- Reformulation of `sinZeta_two_mul_nat_add_one` using `Gammaℂ`. -/
 theorem sinZeta_two_mul_nat_add_one' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
--- a/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
+++ b/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
@@ -31,7 +31,7 @@ sums in the convergence range.)
 * Extend to cover Dirichlet L-functions.
 * The formulae are correct for `s = 0` as well, but we do not prove this case, since this requires
   Fourier series which are only conditionally convergent, which is difficult to approach using the
-  methods in the library at the present time.
+  methods in the library at the present time (May 2024).
 -/
 
 open Complex Real Set
--- a/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
+++ b/Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
@@ -44,7 +44,7 @@ variable {k : ℕ} {x : ℝ}
 of the Bernoulli polynomial. -/
 theorem cosZeta_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc 0 1) :
     cosZeta x (2 * k) = (-1) ^ (k + 1) * (2 * π) ^ (2 * k) / 2 / (2 * k)! *
-      ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval ↑x := by
+      ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval (x : ℂ) := by
   rw [← (hasSum_nat_cosZeta x (?_ : 1 < re (2 * k))).tsum_eq]
   refine Eq.trans ?_ <| (congr_arg ofReal' (hasSum_one_div_nat_pow_mul_cos hk hx).tsum_eq).trans ?_
   · rw [ofReal_tsum]
@@ -95,7 +95,7 @@ theorem sinZeta_two_mul_nat_add_one (hk : k ≠ 0) (hx : x ∈ Icc 0 1) :
 /-- Reformulation of `cosZeta_two_mul_nat` using `Gammaℂ`. -/
 theorem cosZeta_two_mul_nat' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
     cosZeta x (2 * k) = (-1) ^ (k + 1) / (2 * k) / Gammaℂ (2 * k) *
-      ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval ↑x := by
+      ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval (x : ℂ) := by
   rw [cosZeta_two_mul_nat hk hx]
   congr 1
   have : (2 * k)! = (2 * k) * Complex.Gamma (2 * k) := by
@@ -108,7 +108,7 @@ theorem cosZeta_two_mul_nat' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
 /-- Reformulation of `sinZeta_two_mul_nat_add_one` using `Gammaℂ`. -/
 theorem sinZeta_two_mul_nat_add_one' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
     sinZeta x (2 * k + 1) = (-1) ^ (k + 1) / (2 * k + 1) / Gammaℂ (2 * k + 1) *
-      ((Polynomial.bernoulli (2 * k + 1)).map (algebraMap ℚ ℂ)).eval ↑x := by
+      ((Polynomial.bernoulli (2 * k + 1)).map (algebraMap ℚ ℂ)).eval (x : ℂ) := by
   rw [sinZeta_two_mul_nat_add_one hk hx]
   congr 1
   have : (2 * k + 1)! = (2 * k + 1) * Complex.Gamma (2 * k + 1) := by
@@ -121,7 +121,7 @@ theorem sinZeta_two_mul_nat_add_one' (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1)
 
 theorem hurwitzZetaEven_one_sub_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
     hurwitzZetaEven x (1 - 2 * k) =
-      -1 / (2 * k) * ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval ↑x := by
+      -1 / (2 * k) * ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval (x : ℂ) := by
   have h1 (n : ℕ) : (2 * k : ℂ) ≠ -n := by
     rw [← Int.cast_ofNat, ← Int.cast_natCast, ← Int.cast_mul, ← Int.cast_natCast n, ← Int.cast_neg,
       Ne, Int.cast_inj, ← Ne]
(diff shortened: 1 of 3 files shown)


Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,197 @@
...
+
+## TODO
+
+* Derive the special case of the Riemann zeta function from these results (rather than proving it
+  separately). This will come in a future PR which will re-implement Riemann zeta as a special
+  case of Hurwitz zeta.
+* Extend to cover Dirichlet L-functions.
+* The formulae are correct for `s = 0` as well, but we do not prove this case, since this requires
+  Fourier series which are only conditionally convergent, which is difficult to approach using the
+  methods in the library at the present time.
```
faenuccio, line 34: Do you want to add a date to make "present time" clearer?

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,197 @@
...
+
+open Complex Real Set
+
+open scoped Nat
+
+variable {k : ℕ} {x : ℝ}
+
+/-- Express the value of `cosZeta` at a positive even integer as a value
+of the Bernoulli polynomial. -/
+theorem cosZeta_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
```
faenuccio, line 45: This and most of the lemmas below express an "easy"  LHS in terms of a "complicated" RHS. I understand that this is closer to the mathematician's idea that we are "expressing the Hurwitz zeta in terms of..." but I wonder if the opposite would not be more natural given the library design. Do you already have applications showing that these directions are more useful than the opposite ones?

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,197 @@
...
+
+-- private because it is superseded by `hurwitzZeta_neg_nat` below
+private lemma hurwitzZeta_one_sub_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
+    hurwitzZeta x (1 - 2 * k) =
+      -1 / (2 * k) * ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval ↑x := by
+  suffices hurwitzZetaOdd x (1 - 2 * k) = 0 by
+    rw [hurwitzZeta, this, add_zero, hurwitzZetaEven_one_sub_two_mul_nat hk hx]
+  obtain ⟨k, rfl⟩ := Nat.exists_eq_succ_of_ne_zero hk
+  rw [Nat.cast_succ, show (1 : ℂ) - 2 * (k + 1) = - 2 * k - 1 by ring]
+  apply hurwitzZetaOdd_neg_two_mul_nat_sub_one
```
faenuccio, line 177:
```suggestion
  rw [Nat.cast_succ, show (1 : ℂ) - 2 * (k + 1) = - 2 * k - 1 by ring,
    hurwitzZetaOdd_neg_two_mul_nat_sub_one]
```

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,197 @@
...
+  rw [sinZeta_two_mul_nat_add_one hk hx]
+  congr 1
+  have : (2 * k + 1)! = (2 * k + 1) * Complex.Gamma (2 * k + 1) := by
+    rw [(by simp : Complex.Gamma (2 * k + 1) = Complex.Gamma (↑(2 * k) + 1)),
+       Complex.Gamma_nat_eq_factorial, ← Nat.cast_ofNat (R := ℂ), ← Nat.cast_mul,
+      ← Nat.cast_add_one, ← Nat.cast_mul, ← Nat.factorial_succ]
+  simp_rw [this, Gammaℂ, cpow_neg, ← div_div, div_inv_eq_mul, div_mul_eq_mul_div, div_div]
+  congr 2
+  · rw [(by simp : 2 * (k : ℂ) + 1 = ↑(2 * k + 1)), cpow_natCast]
+  · ring
```
faenuccio, line 122:
```suggestion
  rw [(by simp : 2 * (k : ℂ) + 1 = ↑(2 * k + 1)), cpow_natCast]
  ring
```

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,197 @@
...
+      ((Polynomial.bernoulli (2 * k)).map (algebraMap ℚ ℂ)).eval ↑x := by
+  rw [cosZeta_two_mul_nat hk hx]
+  congr 1
+  have : (2 * k)! = (2 * k) * Complex.Gamma (2 * k) := by
+    rw [(by { norm_cast; omega } : 2 * (k : ℂ) = ↑(2 * k - 1) + 1), Complex.Gamma_nat_eq_factorial,
+      ← Nat.cast_add_one, ← Nat.cast_mul, ← Nat.factorial_succ, Nat.sub_add_cancel (by omega)]
+  simp_rw [this, Gammaℂ, cpow_neg, ← div_div, div_inv_eq_mul, div_mul_eq_mul_div, div_div]
+  congr 2
+  · rw [(by simp : 2 * (k : ℂ) = ↑(2 * k)), cpow_natCast]
+  · ring
```
faenuccio, line 107:
```suggestion
  norm_num
  rw [(by simp : 2 * (k : ℂ) = ↑(2 * k)), cpow_natCast]
  ring
```

faenuccio (commented):
Nice work!

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
loefflerd: That’s an Interesting point, but I think I disagree about which side is “complicated”. The RHS is lengthy to write down but all the terms are elementary in some sense (particularly for the lemmas with `s < 0`, where the RHS is a rational number and algorithmically computable, while the LHS takes literally thousands of lines of code to even define.) E.g. in “zeta(-1) = 1/12” I think it’s clear which side is the easy one!

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
loefflerd: I tried applying this change, but it didn't compile. Maybe there was a typo? Could you take another look and let me know what you intended here?

loefflerd:
Thanks for the review, Filippo! I've applied all the changes you suggested, except one which didn't seem to compile.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
faenuccio: I am sorry, it was a bad copy-pasting. The following compiles for me:

```suggestion
  norm_cast
  ring_nf
```

loefflerd: Thanks, that works indeed!

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
faenuccio: Yes, I see what you mean and I also agree that the direction you chose is the best one. I was more thinking about "is a mathematician's and a mathlib-ician's definition of "easy" the same one"? One can argue that "simple things" are goal states you are happy with, and "hard ones" are goal states you are unhappy with. If the above lemma is typically used to transform the RHS into a value of the cosZeta function, that it shuold be changed, but if it is meant to unwind coZeta(2k) than this is the right direction (same idea as for simp). For the code at hand I do not want to suggest modifications, after all, I was just opening the discussion ;-)

faenuccio:
> Thanks for the review, Filippo! I've applied all the changes you suggested, except one which didn't seem to compile.

I have proposed something new that seems to compile.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
loefflerd: Indeed, it's meant as a way to "unwind" cosZeta(2k) in terms of more elementary things (as a step towards formulae like zeta(2) = pi^2 / 6).

loefflerd:
> > Thanks for the review, Filippo! I've applied all the changes you suggested, except one which didn't seem to compile.
> 
> I have proposed something new that seems to compile.

Thanks, it builds fine now with that fix. Let me know if you're happy with the code now, if so I'll ping one of the maintainers to merge it.

faenuccio:
> > > Thanks for the review, Filippo! I've applied all the changes you suggested, except one which didn't seem to compile.
> > 
> > 
> > I have proposed something new that seems to compile.
> 
> Thanks, it builds fine now with that fix. Let me know if you're happy with the code now, if so I'll ping one of the maintainers to merge it.

Yes, I am perfectly happy with the code.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,195 @@
...
+
+open Complex Real Set
+
+open scoped Nat
+
+variable {k : ℕ} {x : ℝ}
+
+/-- Express the value of `cosZeta` at a positive even integer as a value
+of the Bernoulli polynomial. -/
+theorem cosZeta_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
```
riccardobrasca, line 45:
```suggestion
theorem cosZeta_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc 0 1) :
```

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,195 @@
...
+/--
+Express the value of `sinZeta` at an odd integer `> 1` as a value of the Bernoulli polynomial.
+
+Note that this formula is also correct for `k = 0` (i.e. for the value at `s = 1`), but we do not
+prove it in this case, owing to the additional difficulty of working with series that are only
+conditionally convergent.
+-/
+theorem sinZeta_two_mul_nat_add_one (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
+    sinZeta x (2 * k + 1) = (-1) ^ (k + 1) * (2 * π) ^ (2 * k + 1) / 2 / (2 * k + 1)! *
+      ((Polynomial.bernoulli (2 * k + 1)).map (algebraMap ℚ ℂ)).eval ↑x := by
```
riccardobrasca, line 74:
```suggestion
theorem sinZeta_two_mul_nat_add_one (hk : k ≠ 0) (hx : x ∈ Icc 0 1) :
    sinZeta x (2 * k + 1) = (-1) ^ (k + 1) * (2 * π) ^ (2 * k + 1) / 2 / (2 * k + 1)! *
      ((Polynomial.bernoulli (2 * k + 1)).map (algebraMap ℚ ℂ)).eval (x : ℂ) := by
```
I think `(x : ℂ)` is more readable, but as you prefer.

Mathlib/NumberTheory/LSeries/HurwitzZetaValues.lean
```diff
@@ -0,0 +1,195 @@
...
+  congr 1
+  have : (2 * k + 1)! = (2 * k + 1) * Complex.Gamma (2 * k + 1) := by
+    rw [(by simp : Complex.Gamma (2 * k + 1) = Complex.Gamma (↑(2 * k) + 1)),
+       Complex.Gamma_nat_eq_factorial, ← Nat.cast_ofNat (R := ℂ), ← Nat.cast_mul,
+      ← Nat.cast_add_one, ← Nat.cast_mul, ← Nat.factorial_succ]
+  simp_rw [this, Gammaℂ, cpow_neg, ← div_div, div_inv_eq_mul, div_mul_eq_mul_div, div_div]
+  rw [(by simp : 2 * (k : ℂ) + 1 = ↑(2 * k + 1)), cpow_natCast]
+  ring
+
+theorem hurwitzZetaEven_one_sub_two_mul_nat (hk : k ≠ 0) (hx : x ∈ Icc (0 : ℝ) 1) :
```
riccardobrasca, line 122: Same comments as above (and also in the declarations below).

riccardobrasca (commented):
I think this results should go in some namespace, but otherwise LGTM, thanks!

bors d+

loefflerd:
Hi Riccardo,

Thanks for the review! 

Re namespaces: all of the results currently in Mathlib about the Hurwitz and Riemann zeta functions are currently in the root namespace. I'm happy to discuss moving them into a separate namespace of their own, but I feel that should be orthogonal to this PR; it clearly wouldn't make sense to namespace just the new lemmas from this PR without doing the same to the already-extant ones. So if it's OK with you, I'm going to merge this PR once CI passes.

Best wishes, David

riccardobrasca:
> Hi Riccardo,
> 
> Thanks for the review!
> 
> Re namespaces: all of the results currently in Mathlib about the Hurwitz and Riemann zeta functions are currently in the root namespace. I'm happy to discuss moving them into a separate namespace of their own, but I feel that should be orthogonal to this PR; it clearly wouldn't make sense to namespace just the new lemmas from this PR without doing the same to the already-extant ones. So if it's OK with you, I'm going to merge this PR once CI passes.
> 
> Best wishes, David

Yes, this can be done in another PR.

faenuccio:
> Hi Riccardo,
> 
> Thanks for the review!
> 
> Re namespaces: all of the results currently in Mathlib about the Hurwitz and Riemann zeta functions are currently in the root namespace. I'm happy to discuss moving them into a separate namespace of their own, but I feel that should be orthogonal to this PR; it clearly wouldn't make sense to namespace just the new lemmas from this PR without doing the same to the already-extant ones. So if it's OK with you, I'm going to merge this PR once CI passes.
> 
> Best wishes, David

Yes, putting all these results in a namespace could be useful (for instance, thinking ahead at when we will have more general Dedekind zera, something like `residue_zeta` might be in the `RiemannZeta` or in `DedekindZeta` namespaces). For the next PR...

PR #13348, opened 2024-05-29
[Merged by Bors] - feat(GroupTheory/OrderOfElement): equivalences preserve orders

MichaelStollBayreuth:
We add the statement that (multiplicative and additive) equivalences preserve the order of an element.

---

### DIFF
--- a/Mathlib/GroupTheory/OrderOfElement.lean
+++ b/Mathlib/GroupTheory/OrderOfElement.lean
@@ -1299,25 +1299,10 @@ end Prod
 
 section MulEquiv
 
-/-- An injective homomorphism of monoids preserves orders of elements. -/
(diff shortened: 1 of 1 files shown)


erdOne:
I think it is enough to assume that `e` is injective?

erdOne:
Wait I just found `orderOf_injective`

MichaelStollBayreuth:
OK, that simplifies things. (I guess I didn't think of looking for the injective version...)
I still think it's good to have an explicit `MulEuiv` version so that one can write `e.orderOf_eq m` when `e` is an equivalence.

erdOne:
I think it should go under the injective counter part, and maybe made a simp lemma?

riccardobrasca:
Thanks!

bors merge

PR #13349, opened 2024-05-29
[Merged by Bors] - feat: TendstoUniformlyOn for tprod's

CBirkbeck:
Give conditions for prods to converge uniformly to a tprod.

---

- [x] depends on: #16543
- [x] depends on: #16546
- [x] depends on: #15398


CBirkbeck:
I have this result `le_of_TendstoUniformlyOn_le_right` which I expect is true more generally/ has an easier proof, but I'm not sure of a good way to do this, any ideas?

j-loreaux (commented):
Does this review get you off the ground? It seems like there's a lot of legwork left yet before it's ready for primetime. If you need more suggestions, ping me again.

CBirkbeck:
> Does this review get you off the ground? It seems like there's a lot of legwork left yet before it's ready for primetime. If you need more suggestions, ping me again.

Yes this is really useful, thank you!

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,104 @@
...
+open Filter Function Complex Real
+
+open scoped Interval Topology BigOperators Nat Classical Complex
+
+variable {α β ι : Type*}
+
+/--There is likely a way more general proof of this, what should it be? -/
+lemma le_of_TendstoUniformlyOn_le_right (f : ι → α → ℝ) {p : Filter ι}
+    {g : α → ℝ} {K : Set α} {T : ℝ} (hf : TendstoUniformlyOn f g p K) (hg : ∀ x : K, g x ≤ T) :
+    ∀ᶠ (n : ι) in p, ∀ x : K, f n x ≤ (T + 1) := by
```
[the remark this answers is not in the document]
CBirkbeck: So maybe I'm being stupid, but I'm struggling to meaningfully generalise this. One can do something like: 
```
 /--There is likely a way more general proof of this, what should it be? -/
lemma TendstoUniformlyOn_Eventually_le_const (f : ι → α → β) [NormedAddGroup β] {p : Filter ι}
    {g : α → β} {K : Set α} {T V : ℝ} (htv : T < V)
    (hf : TendstoUniformlyOn f g p K) (hg : ∀ x, x ∈ K → ‖g x‖ ≤ T) :
    ∀ᶠ (n : ι) in p, ∀ x, x ∈ K → ‖f n x‖ ≤ V := by
```
But this is more generall but also a bit weaker than what we have. I also don't quite know how to in effect remove the metric space condition and replace it with something like the hyp of [Filter.Tendsto.eventually_le_const](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Order/OrderClosed.html#Filter.Tendsto.eventually_le_const)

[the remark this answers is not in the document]
j-loreaux: I had originally been thinking of something like this (but perhaps it's too much to hope for):
```lean
import Mathlib

theorem foo {ι : Type z} {α : Type u} {γ : Type w} [UniformSpace α] [LinearOrder α]
    [ClosedIciTopology α] {l : Filter ι} {f : ι → γ → α} {g : γ → α} {u v : α} (hv : v < u) (s : Set γ)
    (hg : ∀ x ∈ s, g x ≤ v) (h : TendstoUniformlyOn f g l s) :
    ∀ᶠ (i : ι) in l, (∀ x ∈ s, f i x ≤ u) := by
  sorry
```
Maybe @ADedecker  can specify what is the right generalization of `Filter.Tendsto.eventually_le_const` to `TendstoUniformlyOn`.

Mathlib/Topology/UniformSpace/UniformConvergence.lean
```diff
@@ -137,6 +137,10 @@ theorem tendstoUniformlyOn_iff_tendstoUniformly_comp_coe :
     TendstoUniformlyOn F f p s ↔ TendstoUniformly (fun i (x : s) => F i x) (f ∘ (↑)) p :=
   forall₂_congr fun u _ => by simp
 
+lemma tendstouniformlyOn_iff_restrict {K : Set α} : TendstoUniformlyOn F f p K ↔
```
ADedecker, line 140:
```suggestion
lemma tendstoUniformlyOn_iff_restrict {K : Set α} : TendstoUniformlyOn F f p K ↔
```

I'm surprised we don't have this. Does `apply?` give anything?

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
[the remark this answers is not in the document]
j-loreaux: Yes, it also reminded me of our discussion of Dini's theorem.

> ordered uniform groups

What do you mean by this exactly @ADedecker ? `[UniformGroup G] [LinearOrder G] [OrderTopology G]`, or something else?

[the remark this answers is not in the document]
ADedecker: Yes sorry, I forgot the very important linearity hypothesis, which cannot be avoided for `<` to be something meaningful. I think here we cannot do any generalization towards [NormedLatticeAddCommGroup](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Order/Lattice.html#NormedLatticeAddCommGroup) precisely because of `<`.

j-loreaux:
@CBirkbeck are you clear how to proceed?

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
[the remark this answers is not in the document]
CBirkbeck: Thanks this is really helpful, I'll try these suggestions!

CBirkbeck:
> @CBirkbeck are you clear how to proceed?

Yes I think so, thank you! (Sorry I only just saw all these messages/emails in my inbox).

Mathlib/Topology/UniformSpace/UniformConvergence.lean
CBirkbeck: unfortunately not :(

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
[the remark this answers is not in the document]
CBirkbeck: Sorry for the delay, teaching and grants have had me busy! I've had another go at the lemmas. I have left a couple of comments to see if anyone has a good idea as to what to name the lemma and also if there is an easier way to get the le version.

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,135 @@
...
+
+variable {α β ι : Type*}
+
+@[to_additive] --what do I call this?
+lemma TendstoUniformlyOn_Eventually_lt_mul_const [UniformSpace β] [Group β]
+    [UniformGroup β] [LinearOrder β] [OrderTopology β] [MulLeftStrictMono β] [MulRightMono β]
+    (f : ι → α → β)  {p : Filter ι} {g : α → β} {K : Set α} {T V : β} (htv : T < V)
+    (hf : TendstoUniformlyOn f g p K) (hg : ∀ x, x ∈ K → g x ≤ T) :
+    ∀ᶠ (n : ι) in p, ∀ x, x ∈ K → f n x < V := by
+  rw [@tendstoUniformlyOn_iff_tendsto] at hf
```
loefflerd, line 28:
```suggestion
  rw [tendstoUniformlyOn_iff_tendsto] at hf
```

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,135 @@
...
+  simp only [uniformity_eq_comap_inv_mul_nhds_one, tendsto_iff_eventually, eventually_comap,
+    Prod.forall, uniformContinuous_def, mem_comap, forall_exists_index, and_imp] at *
+  rw [Subtype.forall'] at hf
+  have hf2 := hf ⟨(fun x : β × β => x.1⁻¹ * x.2 < T⁻¹ * V), by
+    rw [eventually_iff_exists_mem]
+    refine ⟨{ x : β | x < T⁻¹ * V }, IsOpen.mem_nhds (isOpen_gt' (T⁻¹ * V)) (by simp [htv]) ,
+      fun y hy a b hab => ?_⟩
+    simpa only [Subtype.forall, lt_inv_mul_iff_mul_lt, Set.mem_setOf_eq, hab] using hy⟩
```
loefflerd, line 36: Here we have a ⟨ ⟩ pattern nested inside another, can it be restructured to avoid that? Also the `simpa only` is terminal so it doesn't need to be squeezed, which gets you a few golf points.

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,135 @@
...
+@[to_additive] --what do I call this? and is there an easier way to get this from the above?
+lemma TendstoUniformlyOn_Eventually_le_mul_const [UniformSpace β] [Group β]
+    [UniformGroup β] [LinearOrder β] [OrderTopology β] [MulLeftStrictMono β] [MulRightMono β]
+    (f : ι → α → β)  {p : Filter ι} {g : α → β} {K : Set α} {T V : β} (htv : T < V)
+    (hf : TendstoUniformlyOn f g p K) (hg : ∀ x, x ∈ K → g x ≤ T) :
+    ∀ᶠ (n : ι) in p, ∀ x, x ∈ K → f n x ≤ V := by
+    have := TendstoUniformlyOn_Eventually_lt_mul_const f htv hf hg
+    rw [eventually_iff_exists_mem] at *
+    obtain ⟨v, hv, H⟩ := this
+    refine ⟨v, hv, fun n hn x hx => (H n hn x hx).le⟩
```
loefflerd, line 51:
```suggestion
  filter_upwards [hf.eventually_lt_mul_const htv hg] with i hi x hx using (hi x hx).le
```

Mathlib/Analysis/SpecialFunctions/Log/Summable.lean
loefflerd, line 1: There is a TODO in the heading of this file which looks like it may now have been fixed.

loefflerd (commented):
This is as far as I got before stopping for dinner.

Mathlib/Topology/Algebra/InfiniteSum/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,135 @@
...
+open scoped Interval Topology BigOperators Nat Complex
+
+variable {α β ι : Type*}
+
+@[to_additive] --what do I call this?
+lemma TendstoUniformlyOn_Eventually_lt_mul_const [UniformSpace β] [Group β]
+    [UniformGroup β] [LinearOrder β] [OrderTopology β] [MulLeftStrictMono β] [MulRightMono β]
+    (f : ι → α → β)  {p : Filter ι} {g : α → β} {K : Set α} {T V : β} (htv : T < V)
+    (hf : TendstoUniformlyOn f g p K) (hg : ∀ x, x ∈ K → g x ≤ T) :
+    ∀ᶠ (n : ι) in p, ∀ x, x ∈ K → f n x < V := by
```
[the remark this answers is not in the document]
loefflerd: See [#maths > Ordered uniform spaces?](https://leanprover.zulipchat.com/#narrow/channel/116395-maths/topic/Ordered.20uniform.20spaces.3F) for discussion on this point.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+    forall_apply_eq_imp_iff₂] at hT
+  have h2 := TendstoUniformlyOn.eventually_forall_le (fun n x => (f n x).re) (lt_add_one T)
+    hf.re hT
+  have w2 := tendstoUniformlyOn_univ.mpr <| UniformContinuousOn.comp_tendstoUniformly_eventually
+    {x : ℂ | x.re ≤ T + 1} (fun a => K.restrict (f (a))) (fun b => g b) (by simpa using h2) ?_
+      (UniformContinuousOn.cexp (T + 1)) ((tendstouniformlyOn_iff_restrict).mp hf)
+  · rw [tendstouniformlyOn_iff_restrict, ← tendstoUniformlyOn_univ]
+    exact w2
+  · simp only [Set.mem_setOf_eq, Subtype.forall]
+    exact fun a ha => le_trans (hT a ha) (by aesop)
```
loefflerd, line 57: This proof seems very convoluted for no good reason, e.g. you rewrite with `← tendstoUniformlyOn_univ` and then close the goal with `exact w2`, but the proof of `w2` begins with `tendstoUniformlyOn_univ.mpr`, so these two steps just cancel each other out. Please, try to write proofs a little more straightforwardly!

A somewhat more direct proof is the following:
```suggestion
lemma tendstoUniformlyOn_comp_cexp {p : Filter ι} {f : ι → α → ℂ} {g : α → ℂ}
    {K : Set α} (hf : TendstoUniformlyOn f g p K) (hg : BddAbove <| (fun x ↦ (g x).re) '' K) :
    TendstoUniformlyOn (cexp ∘ f ·) (cexp ∘ g) p K := by
  obtain ⟨T, hT⟩ : ∃ T, ∀ x ∈ K, (g x).re ≤ T := hg.imp fun _ h ↦ by simpa [mem_upperBounds] using h
  have : ∀ᶠ i in p, ∀ x ∈ K, (f i x).re ≤ T + 1 := hf.re.eventually_forall_le (lt_add_one T) hT
  rw [tendstouniformlyOn_iff_restrict]
  refine (UniformContinuousOn.cexp _).comp_tendstoUniformly_eventually _ _ _ (by simpa) ?_ 
    (tendstouniformlyOn_iff_restrict.mp hf)
  simpa using fun a ha ↦ (hT a ha).trans (lt_add_one T).le
```
I'm giving the types explicitly in the `obtain` and `have` for readability (in your version it is very non-obvious to the reader what `hT` actually says after that `simp only` call). 

The proof would be substantially simpler if we had a lemma `comp_tendstoUniformlyOn_eventually` which would save flipping back+forth with `tendstouniformlyOn_iff_restrict`.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+
+lemma tendstoUniformlyOn_tprod_of_clog {f : ι → α → ℂ} {K : Set α}
+    (h : ∀ x, x ∈ K → Summable fun n => log (f n x))
+    (hf : TendstoUniformlyOn (fun n a => ∑ i ∈ n, log (f i a))
+    (fun a : α => ∑' n : ι, log (f n a)) atTop K) (hfn : ∀ x, x ∈ K → ∀ n : ι, f n x ≠ 0)
+    (hg : BddAbove ((fun x => (∑' n : ι, log (f n x)).re) '' K)) :
+    TendstoUniformlyOn (fun n a => ∏ i ∈ n, (f i a)) (fun a => ∏' i, (f i a)) atTop K := by
```
loefflerd, line 64: Please get into the habit of, whenever you see a pair of brackets, trying deleting them to check if they are needed.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+      (UniformContinuousOn.cexp (T + 1)) ((tendstouniformlyOn_iff_restrict).mp hf)
+  · rw [tendstouniformlyOn_iff_restrict, ← tendstoUniformlyOn_univ]
+    exact w2
+  · simp only [Set.mem_setOf_eq, Subtype.forall]
+    exact fun a ha => le_trans (hT a ha) (by aesop)
+
+lemma tendstoUniformlyOn_tprod_of_clog {f : ι → α → ℂ} {K : Set α}
+    (h : ∀ x, x ∈ K → Summable fun n => log (f n x))
+    (hf : TendstoUniformlyOn (fun n a => ∑ i ∈ n, log (f i a))
+    (fun a : α => ∑' n : ι, log (f n a)) atTop K) (hfn : ∀ x, x ∈ K → ∀ n : ι, f n x ≠ 0)
```
loefflerd, line 62: ... and the same for explicit type-ascriptions (`a` can't be anything else since it's the second argument to `f`).

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+    {x : ℂ | x.re ≤ T + 1} (fun a => K.restrict (f (a))) (fun b => g b) (by simpa using h2) ?_
+      (UniformContinuousOn.cexp (T + 1)) ((tendstouniformlyOn_iff_restrict).mp hf)
+  · rw [tendstouniformlyOn_iff_restrict, ← tendstoUniformlyOn_univ]
+    exact w2
+  · simp only [Set.mem_setOf_eq, Subtype.forall]
+    exact fun a ha => le_trans (hT a ha) (by aesop)
+
+lemma tendstoUniformlyOn_tprod_of_clog {f : ι → α → ℂ} {K : Set α}
+    (h : ∀ x, x ∈ K → Summable fun n => log (f n x))
+    (hf : TendstoUniformlyOn (fun n a => ∑ i ∈ n, log (f i a))
```
loefflerd, line 61: In the same declaration you once use `n` for a generic term of `ι` and then for a generic `Finset ι`. Neither is an ideal choice and the combination is worse (the reader will naturally expect the same letter has the same type throughout the statement).

loefflerd (commented):
Hi Chris,

I hate to say this, but I think it would ease the task of refereeing your PR's a great deal if you could internalize the Mathlib coding conventions a bit more, and pay more attention to proof-reading your code before submission to simplify and tighten up the arguments. Please get into the habit of thinking, "is this the best name for this variable? are these brackets, type-ascriptions, etc, needed?"

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+
+open Filter Function Complex
+
+variable {α β ι : Type*}
+
+lemma TendstoUniformlyOn.eventually_forall_lt [UniformSpace β] [AddGroup β]
+    [UniformAddGroup β] [LinearOrder β] [OrderTopology β] [AddLeftStrictMono β] [AddRightMono β]
+    {f : ι → α → β} {p : Filter ι} {g : α → β} {K : Set α} {T V : β} (htv : T < V)
+    (hf : TendstoUniformlyOn f g p K) (hg : ∀ x, x ∈ K → g x ≤ T) :
+    ∀ᶠ (n : ι) in p, ∀ x, x ∈ K → f n x < V := by
```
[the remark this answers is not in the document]
CBirkbeck: I guess I stuggle to see the consistency elsewhere in mathlib, so I'm never quite sure about some things. I agree with what you are saying though, but for example, you say 'K' is fine for sets, but the library style guides say 's, t, ... for sets' but we use upper case for sets all the time.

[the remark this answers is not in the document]
loefflerd: The general rule is "capital letters for types, lowercase letters for terms"; but sets are occasionally tolerated as an exception to this rule – using `x` for a generic element of the set `X` is so obviously natural that the rules are bent to allow it. So the alternative convention "Greek for types, capital for sets, lowercase for elements" is common, particularly in analysis / topology. But I do agree this rule is less clear-cut than the other ones.

"Seeing the consistency in mathlib" is indeed not easy, since the coding conventions have shifted over time and code is not always retrospectively re-styled to adapt to changes (e.g. there is still a lot of suboptimal formatting hanging around from the lean3 port). Also, the review process is not perfect; the fact that one reviewer once approved something does not mean it is always allowed in all subsequent contributions.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+    (h : ∀ x, x ∈ K → Summable fun n => log (f n x))
+    (hf : TendstoUniformlyOn (fun n a => ∑ i ∈ n, log (f i a)) (fun a : α => ∑' n : ℕ, log (f n a))
+    atTop K) (hfn : ∀ x, x ∈ K → ∀ n : ℕ, f n x ≠ 0)
+    (hg : BddAbove ((fun x => (∑' n : ℕ, log (f n x)).re) '' K)) :
+    TendstoUniformlyOn (fun n : ℕ => fun a : α => ∏ i ∈ Finset.range n, (f i a))
+    (fun a => ∏' i, (f i a)) atTop K :=
+  fun v hv => tendsto_finset_range.eventually (tendstoUniformlyOn_tprod_of_clog h hf hfn hg v hv)
+
+/-- This is the version for infinite products of with terms of the form `1 + f n x`. -/
+lemma tendstoUniformlyOn_tprod_nat [TopologicalSpace α] {f : ℕ → α → ℂ} {K : Set α}
```
loefflerd, line 85: "The version" of what? Please specify. Moreover, why do you require here that the bound on `f` holds for _all_ `n`, not just eventually as in the previous lemma? This complicates the proof while also weakening the statement – you lose both ways.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+      apply ContinuousOn.log
+      · apply ContinuousOn.comp continuous_norm.continuousOn
+          (ContinuousOn.add continuousOn_const (hcts c)) (Set.mapsTo_image (fun x ↦ 1 + f c x) K)
+      · intro z hz
+        simpa using hfn z hz c
+    exact IsCompact.bddAbove_image hK H
+  · intro x hx
+    exact Complex.summable_log_one_add_of_summable (summable_norm_iff.mp
+      (Summable.of_nonneg_of_le (fun b ↦ norm_nonneg (f b ↑x)) (fun _ => h _ _ hx) hu))
+  · exact Complex.tendstoUniformlyOn_tsum_log_one_add K hu (Filter.Eventually.of_forall h)
```
loefflerd, line 107:
```suggestion
lemma tendstoUniformlyOn_tprod_nat [TopologicalSpace α] {f : ℕ → α → ℂ} {K : Set α}
    (hK : IsCompact K) {u : ℕ → ℝ} (hu : Summable u) (h : ∀ n x, x ∈ K → ‖f n x‖ ≤ u n)
    (hfn : ∀ x ∈ K, ∀ n, 1 + f n x ≠ 0) (hcts : ∀ n, ContinuousOn (f n) K) :
    TendstoUniformlyOn (fun n a ↦ ∏ i ∈ Finset.range n, (1 + f i a))
    (fun a ↦ ∏' i, (1 + f i a)) atTop K := by
  refine tendstoUniformlyOn_tprod_nat_of_clog ?_ ?_ hfn (hK.bddAbove_image ?_)
  · exact fun x hx ↦ summable_log_one_add_of_summable (summable_norm_iff.mp
      (.of_nonneg_of_le (fun b ↦ norm_nonneg (f b x)) (fun _ ↦ h _ _ hx) hu))
  · exact tendstoUniformlyOn_tsum_log_one_add K hu (Eventually.of_forall h)
  · apply (tendstoUniformlyOn_tsum_nat_log_one_add K hu (.of_forall h)).re.continuousOn
    filter_upwards with n
    simp only [re_sum, log_re]
    refine continuousOn_finset_sum _ fun c _ ↦ ?_
    exact (continuousOn_const.add (hcts c)).norm.log (fun z hz ↦ by simpa using hfn z hz c)
```

When you have a `have` block and then the proof is a one-liner from that, you can almost always re-order it to avoid needing the `have` at all, as with `have H : ...` here. Also, I think it's very confusing to use `apply` with some *but not all* arguments specified as explicit blanks `?_`, because the resulting goals come out in the wrong order! Arguably this is an `apply` bug, but I think it's best to use `refine` as soon as you have one `?_`.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+    {f : ℕ → α → ℂ} {K : Set α} (hK : IsOpen K) {u : ℕ → ℝ} (hu : Summable u)
+    (h : ∀ n x, x ∈ K → ‖f n x‖ ≤ u n)(hfn : ∀ x, x ∈ K → ∀ n : ℕ, 1 + f n x ≠ 0)
+    (hcts : ∀ n, ContinuousOn (fun x => (f n x)) K) :
+    TendstoLocallyUniformlyOn (fun n : ℕ => fun a : α => ∏ i ∈ Finset.range n, (1 + (f i a)))
+    (fun a => ∏' i, (1 + (f i a))) atTop K := by
+  rw [tendstoLocallyUniformlyOn_iff_forall_isCompact hK]
+  intro S hS hS2
+  apply tendstoUniformlyOn_tprod_nat hS2 hu
+  · exact fun n x hx ↦ h n x (hS hx)
+  · exact fun x hx n => hfn x (hS hx) n
```
loefflerd, line 120: Both of these goals can be closed by `tauto`.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,121 @@
...
+  rw [tendstoUniformlyOn_iff_tendsto] at hf
+  simp only [uniformity_eq_comap_neg_add_nhds_zero, tendsto_iff_eventually, eventually_comap,
+    Prod.forall] at *
+  conv at hf => enter [2]; rw [eventually_iff_exists_mem]
+  have hf2 := hf (fun x : β × β => -x.1 + x.2 < -T + V) ⟨{x : β | x < -T + V},
+      IsOpen.mem_nhds (isOpen_gt' (-T + V)) (by simp [htv]),
+      fun y hy a b hab => by simpa [hab] using hy⟩
+  rw [eventually_prod_principal_iff, eventually_iff_exists_mem] at *
+  obtain ⟨v, hv, H⟩ := hf2
+  refine ⟨v, hv, fun y hy x hx => by simpa using (add_lt_add_of_le_of_lt (hg x hx) (H y hy x hx))⟩
```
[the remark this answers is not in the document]
CBirkbeck: Ok so this is a bit confusing to me. This looks nicer but is slower in terms of heartbeats, it goes from 5815 to 6515, which isn't much, but I'm normally prioritising speed over other things, what's the rule of thumb here?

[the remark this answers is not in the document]
loefflerd: The difference here, 700 heartbeats, translates to something like 0.2 sec on my machine.

I think there is general consensus that readability / maintainability should be prioritized over speed. E.g. I refer you to the discussion on Zulip not long back introducing a new rule that terminal simps should _never_ be squeezed (not the same specific issue as here, but the same general trade-off, prioritizing other goals over speed). 

[Just to clarify, here I'm talking about "localized" speed issues, affecting a single proof, rather than things like typeclass-synthesis issues that can be a performance drag across big swathes of the library.]

Mathlib/Topology/UniformSpace/UniformConvergenceTopology.lean
```diff
@@ -1124,35 +1124,43 @@ section UniformComposition
...
+theorem UniformContinuousOn.comp_tendstoUniformly {s : Set β} {F : ι → α → β} {f : α → β}
+    (hF : ∀ i x, F i x ∈ s) (hf : ∀ x, f x ∈ s) {g : β → γ} (hg : UniformContinuousOn g s)
+    (h : TendstoUniformly F f p) :
     TendstoUniformly (fun i x => g (F i x)) (fun x => g (f x)) p := by
   rw [uniformContinuousOn_iff_restrict] at hg
   lift F to ι → α → s using hF with F' hF'
   lift f to α → s using hf with f' hf'
   rw [tendstoUniformly_iff_tendsto] at h
-  have : Tendsto (fun q : ι × α ↦ (f' q.2, (F' q.1 q.2))) (p ×ˢ ⊤) (𝓤 s) :=
+  have : Tendsto (fun q ↦ (f' q.2, (F' q.1 q.2))) (p ×ˢ ⊤) (𝓤 s) :=
```
loefflerd, line 1135:
```suggestion
  have : Tendsto (fun q ↦ (f' q.2, F' q.1 q.2)) (p ×ˢ ⊤) (𝓤 s) :=
```

Mathlib/Topology/UniformSpace/UniformConvergenceTopology.lean
```diff
@@ -1124,35 +1124,43 @@ section UniformComposition
...
     h.of_tendsto_comp isUniformEmbedding_subtype_val.comap_uniformity.le
```
loefflerd, line 1136: There is a step that could be factored out here, something like `tendstoUniformly.of_comp`: if `r` is a `UniformEmbedding`, then a sequence of functions converges uniformly iff its composition with `r` does. If we had that, we could prove this lemma more tidily without having to use `tendstoUniformly_iff_tendsto`. But perhaps that is a matter for another PR.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
loefflerd, line 1: This file mixes general lemmas about uniform convergence (which have relatively few pre-reqs) with lemmas which require `Complex.exp` or `Complex.log` (which need much more imports). Can the lemmas be re-homed somehow? I feel the lemmas requiring `Complex.log` should be in the `Analysis` hierarchy, not the `Topology` one.

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,88 @@
...
+    TendstoUniformlyOn (fun s a => ∏ i ∈ s, f i a) (fun a => ∏' i, f i a) atTop K := by
+  suffices H : TendstoUniformlyOn (fun s a => ∏ i ∈ s, f i a)
+       (cexp ∘ fun a ↦ ∑' i, log (f i a)) atTop K by
+        apply TendstoUniformlyOn.congr_right H
+        exact fun x hx ↦ (cexp_tsum_eq_tprod (fun n ↦ f n x) (hfn x hx) (h x hx))
+  apply TendstoUniformlyOn.congr (hf.comp_cexp hg)
+  filter_upwards with s i hi using by simp [exp_sum, fun y ↦ exp_log (hfn i hi y)]
+
+/-- This is the version of `tendstoUniformlyOn_tprod_of_clog` for infinite products of with terms
+of the form `1 + f n x`. -/
```
loefflerd, line 59: This does not seem to be an accurate description of the contents of the lemma. It also changes many things at once relative to the previous lemma (the shape of the terms has changed, the uniform bound `u` has made an appearance, the index set is `ℕ` rather than an arbitrary `ι`, etc); it might be better to try to make these steps one at a time, rather than all at once, in order to offer the maximum flexibility to others who want to use these lemmas in future

Mathlib/Topology/UniformSpace/MultipliableTendstoUniformly.lean
```diff
@@ -0,0 +1,88 @@
...
+    filter_upwards [Nat.cofinite_eq_atTop ▸ h] with n hn using (hn x hx)
+  · exact tendstoUniformlyOn_tsum_log_one_add K hu (Nat.cofinite_eq_atTop ▸ h)
+  · apply (tendstoUniformlyOn_tsum_nat_log_one_add K hu h).re.continuousOn
+    filter_upwards with n
+    simp only [re_sum, log_re]
+    refine continuousOn_finset_sum _ fun c _ ↦ ?_
+    exact (continuousOn_const.add (hcts c)).norm.log (fun z hz ↦ by simpa using hfn z hz c)
+
+/-- This is the locally version of `tendstoUniformlyOn_tprod_nat_one_add`
+for infinite products of with terms of the form `1 + f n x`. -/
```
loefflerd, line 78: typo ("products of with")

loefflerd (commented):
I'm sorry this is taking so long to (ahem) converge. It's a real challenge to find the right setup here. I wonder whether it might be better to start by "multiplicativizing" the existing file `Analysis/NormedSpace/FunctionSeries.lean` (using `to_additive` as much as possible) and then doing the things which specifically rely on `Complex.log` / `Complex.exp` right at the end.

Mathlib/Topology/UniformSpace/UniformConvergenceTopology.lean
CBirkbeck: I think I'll leave this for another PR as this one is already dragging on.

CBirkbeck:
Ok I've updated this to use the `MultipliableUniformlyOn` stuff. I think it'll be lots of work to multiplicativise the funtion series stuff, since it relies on the Weierstrass M-test, for which I don't know if there is a version that would play well with `to_additive`.  Also `TendstoUniformlyOn.eventually_forall_lt` feels like it should go elsewhere but find_home cant suggest anything and places I've looked would require more imports which maybe we dont want? But suggestions on where to put this are welcome.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,125 @@
...
+    (hf : SummableUniformlyOn (fun i a => log (f i a)) 𝔖) (hfn : ∀ K ∈ 𝔖, ∀ x ∈ K, ∀ i, f i x ≠ 0)
+    (hg : ∀ K ∈ 𝔖, BddAbove <| (fun x => (∑' n : ι, log (f n x)).re) '' K) :
+    HasProdUniformlyOn f (fun a => ∏' i, f i a) 𝔖 := by
+  simp only [hasProdUniformlyOn_iff_tendstoUniformlyOn, Set.mem_singleton_iff, forall_eq]
+  obtain ⟨r, hr⟩ := hf.exists
+  intro K hK
+  suffices H : TendstoUniformlyOn (fun s a => ∏ i ∈ s, f i a) (cexp ∘ r) atTop K by
+        apply TendstoUniformlyOn.congr_right H
+        apply Set.EqOn.trans (Set.EqOn.comp_left (hr.tsum_eqOn hK)).symm
+        exact fun x hx ↦ (cexp_tsum_eq_tprod (hfn K hK x hx) (hf.summable hK hx))
```
j-loreaux, line 81:
```suggestion
  apply TendstoUniformlyOn.congr_right H
  apply Set.EqOn.trans (Set.EqOn.comp_left (hr.tsum_eqOn hK)).symm
  exact fun x hx ↦ (cexp_tsum_eq_tprod (hfn K hK x hx) (hf.summable hK hx))
```

j-loreaux (commented):
@loefflerd are you happy with this?

loefflerd:
I should have some time to review this in detail later today. Sorry for not getting around to this earlier. One quick observation: the naming conventions are slightly off, e.g. `MultipliableUniformlyOn_nat_one_add` should be `multipliableUniformlyOn_xxx` – please fix this.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
loefflerd: FWIW, I think the correct indent is halfway between the original and Jireh's suggestion, but I may be wrong (it's hard to get whitespace right in the github suggestion view, since everything shows in a non-monospaced font).

Mathlib/Topology/UniformSpace/UniformConvergenceTopology.lean
```diff
@@ -1120,35 +1120,41 @@ section UniformComposition
...
   rw [eventuallyEq_iff_exists_mem]
   refine ⟨s', hs', fun i hi => by aesop⟩
 
+theorem UniformContinuousOn.comp_tendstoUniformlyOn_eventually {s : Set β} {F : ι → α → β}
+    {f : α → β} {t : Set α} (hF : ∀ᶠ i in p, ∀ x ∈ t, F i x ∈ s) (hf : ∀ x ∈ t, f x ∈ s)
+    {g : β → γ} (hg : UniformContinuousOn g s) (h : TendstoUniformlyOn F f p t) :
+    TendstoUniformlyOn (fun i x ↦ g (F i x)) (fun x => g (f x)) p t := by
+  rw [tendstouniformlyOn_iff_restrict]
+  apply UniformContinuousOn.comp_tendstoUniformly_eventually (by simpa using hF )
+    (by simpa using hf) hg (by rw [tendstouniformlyOn_iff_restrict] at h; exact h)
```
loefflerd, line 1158:
```suggestion
    (by simpa using hf) hg (tendstouniformlyOn_iff_restrict.mp h)
```

loefflerd (commented):
Here are some comments on the files in `Topology/`. Will look at the analysis files later.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+    ∀ᶠ i in p, ∀ x ∈ K, f i x ≤ v := by
+  filter_upwards [hf.eventually_forall_lt huv hg] with i hi x hx using (hi x hx).le
+
+lemma TendstoUniformlyOn.comp_cexp {p : Filter ι} {f : ι → α → ℂ} {g : α → ℂ}
+    {K : Set α} (hf : TendstoUniformlyOn f g p K) (hg : BddAbove <| (fun x ↦ (g x).re) '' K) :
+    TendstoUniformlyOn (cexp ∘ f ·) (cexp ∘ g) p K := by
+  obtain ⟨v, hv⟩ : ∃ v, ∀ x ∈ K, (g x).re ≤ v := hg.imp fun _ h ↦ by simpa [mem_upperBounds] using h
+  have : ∀ᶠ i in p, ∀ x ∈ K, (f i x).re ≤ v + 1 := hf.re.eventually_forall_le (lt_add_one v) hv
+  refine (UniformContinuousOn.cexp _).comp_tendstoUniformlyOn_eventually (by simpa) ?_ hf
+  simpa using fun a ha ↦ (hv a ha).trans (lt_add_one v).le
```
loefflerd, line 49:
```suggestion
  exact fun a ha ↦ (hv a ha).trans (lt_add_one v).le
```
(We should have a linter that checks automatically if `simpa using foo` can become `exact foo`.)

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+
+lemma Complex.hasSumUniformlyOn_log_one_add {α : Type*} {f : ι → α → ℂ} (K : Set α)
```
loefflerd, line 51: Maybe put this & next lemma in the `Summable` namespace so they can be dot-notated?

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+    {u : ι → ℝ} (hu : Summable u) (h : ∀ᶠ n in cofinite, ∀ x ∈ K, ‖f n x‖ ≤ u n) :
+    HasSumUniformlyOn (fun i x ↦ log (1 + f i x)) (fun x ↦ ∑' i, log (1 + f i x)) {K} := by
+  simp only [hasSumUniformlyOn_iff_tendstoUniformlyOn, Set.mem_singleton_iff, forall_eq]
+  apply tendstoUniformlyOn_tsum_of_cofinite_eventually <| hu.mul_left (3 / 2)
```
loefflerd, line 55: This seems to be using a lemma from `Analysis.NormedSpace.FunctionSeries` which isn't formulated in terms of `HasSumUniformlyOn`, but could be. Rewriting that is probably out of scope for the present PR, but maybe put a TODO in the code of `FunctionSeries.lean` suggesting this?

loefflerd (commented):
I started writing a review and then realised that I had forgotten to update to your latest commits first. Here is what I had up to that point.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+  have h1 := hr.tsum_eqOn hK
+  simp only [hasSumUniformlyOn_iff_tendstoUniformlyOn, Set.mem_singleton_iff, forall_eq] at hr
+  apply TendstoUniformlyOn.congr ((hr K hK).comp_cexp ?_)
+  · filter_upwards with s i hi using by simp [exp_sum, fun y ↦ exp_log (hfn K hK i hi y)]
+  · convert hg K hK
+    simp_all only [h1 _, ne_eq]
+
+lemma multipliableUniformlyOn_of_clog {f : ι → α → ℂ} {𝔖 : Set (Set α)}
+    (hf : SummableUniformlyOn (fun i x ↦ log (f i x)) 𝔖) (hfn : ∀ K ∈ 𝔖, ∀ x ∈ K, ∀ i, f i x ≠ 0)
+    (hg : ∀ K ∈ 𝔖, BddAbove <| (fun x ↦ (∑' n : ι, log (f n x)).re) '' K) :
```
loefflerd, line 91:
```suggestion
    (hg : ∀ K ∈ 𝔖, BddAbove <| (fun x ↦ (∑' i, log (f n x)).re) '' K) :
```

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+    MultipliableUniformlyOn f 𝔖 :=
+    ⟨fun x ↦ ∏' i, f i x, hasProdUniformlyOn_of_clog  hf hfn hg⟩
```
loefflerd, line 93:
```suggestion
  ⟨_, hasProdUniformlyOn_of_clog hf hfn hg⟩
```

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
CBirkbeck: Yes so I started making this change here #25096, but it looks like the naive changing imports was maybe something think more carefully about so I left it out for now.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+
+lemma hasProdUniformlyOn_nat_one_add [TopologicalSpace α] {f : ℕ → α → ℂ} {K : Set α}
+    (hK : IsCompact K) {u : ℕ → ℝ} (hu : Summable u) (h : ∀ᶠ n in atTop, ∀ x ∈ K, ‖f n x‖ ≤ u n)
+    (hfn : ∀ x ∈ K, ∀ n, 1 + f n x ≠ 0) (hcts : ∀ n, ContinuousOn (f n) K) :
+    HasProdUniformlyOn (fun n x ↦ 1 + f n x) (fun x ↦ ∏' i, (1 + f i x)) {K} := by
+  refine hasProdUniformlyOn_of_clog ?_ (by simpa using hfn) ?_
+  · apply HasSumUniformlyOn.summableUniformlyOn (g := fun x ↦ ∑' i, log (1 + f i x))
+    apply Complex.hasSumUniformlyOn_log_one_add K hu (Nat.cofinite_eq_atTop ▸ h)
```
loefflerd, line 101:
```suggestion
  · exact (hu.hasSumUniformlyOn_log_one_add K (Nat.cofinite_eq_atTop ▸ h)).summableUniformlyOn
```

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+  · convert hg K hK
+    simp_all only [h1 _, ne_eq]
+
+lemma multipliableUniformlyOn_of_clog {f : ι → α → ℂ} {𝔖 : Set (Set α)}
+    (hf : SummableUniformlyOn (fun i x ↦ log (f i x)) 𝔖) (hfn : ∀ K ∈ 𝔖, ∀ x ∈ K, ∀ i, f i x ≠ 0)
+    (hg : ∀ K ∈ 𝔖, BddAbove <| (fun x ↦ (∑' n : ι, log (f n x)).re) '' K) :
+    MultipliableUniformlyOn f 𝔖 :=
+    ⟨fun x ↦ ∏' i, f i x, hasProdUniformlyOn_of_clog  hf hfn hg⟩
+
+lemma hasProdUniformlyOn_nat_one_add [TopologicalSpace α] {f : ℕ → α → ℂ} {K : Set α}
```
loefflerd, line 95: Can we have a version of this with a general index set instead of Nat? It should also probably be in the `Summable` namespace. Ditto for the remaining 3 lemmas below.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,130 @@
...
+    (hK : IsCompact K) {u : ℕ → ℝ} (hu : Summable u) (h : ∀ᶠ n in atTop, ∀ x ∈ K, ‖f n x‖ ≤ u n)
+    (hfn : ∀ x ∈ K, ∀ n, 1 + f n x ≠ 0) (hcts : ∀ n, ContinuousOn (f n) K) :
```
loefflerd, line 97: Is the hypothesis `hfn` really necessary for this lemma? My feeling is that it probably shouldn't be; having zero terms makes it easier, not harder, for the product to converge.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,151 @@
...
+  by_cases hKe : K = ∅
+  · simp [TendstoUniformly, hKe]
+  · haveI hCK : CompactSpace K := isCompact_iff_compactSpace.mp hK
+    haveI hne : Nonempty K := by rwa [Set.nonempty_coe_sort, Set.nonempty_iff_ne_empty]
+    let f' n : C(K, R) := ⟨_, continuousOn_iff_continuous_restrict.mp (hcts n)⟩
+    have hf'_bd : ∀ᶠ (n : ι) in cofinite, ‖f' n‖ ≤ u n := by
+      simp_rw [ContinuousMap.norm_le_of_nonempty]
+      filter_upwards [h] with n hn using fun x ↦ hn x x.2
+    have hM : Multipliable fun i ↦ 1 + f' i := by
+      apply _root_.multipliable_one_add_of_summable
```
loefflerd, line 98: We should add the `protected` flag to `Complex.multipliable_one_add_of_summable` so we don't need `_root_` here.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,151 @@
...
+      apply hu.of_norm_bounded_eventually
+      simpa only [norm_norm] using hf'_bd
+    convert ContinuousMap.tendsto_iff_tendstoUniformly.mp hM.hasProd
+    · aesop
+    · exact funext fun k ↦ ContinuousMap.tprod_apply hM k
+
+lemma multipliableUniformlyOn_one_add (hK : IsCompact K) (hu : Summable u)
+    (h : ∀ᶠ n in cofinite, ∀ x ∈ K, ‖f n x‖ ≤ u n) (hcts : ∀ n, ContinuousOn (f n) K) :
+    MultipliableUniformlyOn (fun n x ↦ 1 + f n x) {K} :=
+    ⟨_, hasProdUniformlyOn_one_add hK hu h hcts⟩
```
loefflerd, line 108:
```suggestion
  ⟨_, hasProdUniformlyOn_one_add hK hu h hcts⟩
```

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,151 @@
...
+    HasProdUniformlyOn (fun n x ↦ 1 + f n x) (fun x ↦ ∏' i, (1 + f i x)) {K} :=
+  hasProdUniformlyOn_one_add hK hu (Nat.cofinite_eq_atTop ▸ h) hcts
+
+lemma multipliableUniformlyOn_nat_one_add {f : ℕ → α → R} (hK : IsCompact K)
+    {u : ℕ → ℝ} (hu : Summable u) (h : ∀ᶠ n in atTop, ∀ x ∈ K, ‖f n x‖ ≤ u n)
+    (hcts : ∀ n, ContinuousOn (f n) K) :
+    MultipliableUniformlyOn (fun n x ↦ 1 + f n x) {K} :=
+  ⟨_, hasProdUniformlyOn_nat_one_add hK hu h hcts⟩
+
+lemma hasProdLocallyUniformlyOn_one_add [LocallyCompactSpace α] (hK : IsOpen K) (hu : Summable u)
```
loefflerd, line 122: Do we have a lemma saying that `HasProdUniformlyOn` implies `HasProdLocallyUniformlyOn`? If not, then we probably should; and then it ought to reduce this proof, and its nat version, to one-liners.

We should also wrap these lemmas in a section & promote the `LocallyCompactSpace` assumption to a section-level variable.

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
```diff
@@ -0,0 +1,151 @@
...
+    TendstoUniformlyOn (fun n x ↦ ∑ i ∈ Finset.range n, log (1 + f i x))
+    (fun x ↦ ∑' i : ℕ, Complex.log (1 + f i x)) atTop K := by
+  rw [← Nat.cofinite_eq_atTop] at h
+  exact (hu.hasSumUniformlyOn_log_one_add h).tendstoUniformlyOn_finset_range rfl
+
+end summable
+
+section UniformlyOn
+
+/-Note this is false without hfn. -/
```
loefflerd, line 54: Rather more detail needed, I think. Here's a suggestion; please reformat this so it fits in the line width, & maybe write similarly detailed docstrings for some of the other lemmas.

```suggestion
/- If `x ↦ ∑' i, log (f i x)` is uniformly convergent on `𝔖`, its sum has bounded-above real part on each set in `𝔖`, and the functions `f i x` have no zeroes, then  `∏' i, f i x` is uniformly convergent on `𝔖`. 

Note that the non-vanishing assumption is really needed here: if this assumption is dropped then one obtains a counterexample by taking `ι = α = ℕ` and `f i x` is 0 if `i = x` and `1` otherwise. -/
```

loefflerd (commented):
Thanks for pushing this through!

Mathlib/Analysis/NormedSpace/MultipliableUniformlyOn.lean
loefflerd: Oh, sorry, you're right of course – I forgot `K` was not necessarily compact here.

jcommelin (commented):
Thanks :tada:

bors merge

PR #13351, opened 2024-05-29
[Merged by Bors] - chore: clean up `nolints.yml` workflow

grunweg:
Synchronise the early steps with the main build workflow: the differences were minor and not essential, but there is little point in being different. Best reviewed by `diff`ing against `build.yml`.

It would be nice to avoid the copy-pasting of e.g. the build mathlib jobs: that pre-existing issue is left for future PRs.

---


.github/workflows/nolints.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           # github using a different username.
           git config --unset http.https://github.com/.extraheader
 
-      - name: update nolints.json and style-exceptions.txt
+      - name: file a PR to update nolints.json and style-exceptions.txt
```
YaelDillies, line 78: Something which is not really clear from the description is what happens if a PR already exists. Reading `./scripts/update_nolints_CI.sh`, it seems that simply nothing happens, but personally I would have expected the PR to be updated.

YaelDillies (approved):
Yep, I agree with all of this.

maintainer merge

.github/workflows/nolints.yml
grunweg: Sure, let's make the wording more precise. :-) I'd rather ensure these PRs get merged quickly than spend time on improving the workflow for a scenario that hopefully never happens.
```suggestion
     - name: file a new PR to update nolints.json and style-exceptions.txt

File: .github/workflows/nolints.yml [hunk 2]
```diff
@@ -47,15 +47,15 @@ jobs:
...
-        uses: liskin/gh-problem-matcher-wrap@v2
+        uses: liskin/gh-problem-matcher-wrap@v3
         with:
           linters: gcc
           run: |
-            bash -o pipefail -c "env LEAN_ABORT_ON_PANIC=1 lake build -KCI | tee stdout.log"
+            bash -o pipefail -c "env LEAN_ABORT_ON_PANIC=1 lake build --wfail -KCI"
 
-      - name: lint mathlib
+      - name: lint mathlib and update style exceptions files
```
grunweg, line 56:
```suggestion
      - name: update nolints.json and style-exceptions.txt

File: .github/workflows/nolints.yml [hunk 3]
```diff
@@ -47,15 +47,15 @@ jobs:
...
-        uses: liskin/gh-problem-matcher-wrap@v2
+        uses: liskin/gh-problem-matcher-wrap@v3
         with:
           linters: gcc
           run: |
-            bash -o pipefail -c "env LEAN_ABORT_ON_PANIC=1 lake build -KCI | tee stdout.log"
+            bash -o pipefail -c "env LEAN_ABORT_ON_PANIC=1 lake build --wfail -KCI"
 
-      - name: lint mathlib
+      - name: update nolints.json and style-exceptions.txt
```
YaelDillies, line 56:
```suggestion
      - name: annotate the PR, update nolints.json and style-exceptions.txt
```

YaelDillies: Actually, no, your description is complete wrong: it does not update any file, it only recreates `nolints.json` and `style-exceptions.txt` for the purpose of diffing them with the existing ones.
```suggestion
      - name: annotate the PR with linting and style errors
```

grunweg: After #13343, this will be correct: let's just land that PR first.

grunweg: That PR has landed, so this hunk is now removed.

jcommelin:
Thanks :tada:

bors merge

PR #13353, opened 2024-05-29
[Merged by Bors] - feat(Analysis/Fourier): discrete Fourier transform on `ZMod N`

loefflerd:
We write down basic definitions for the discrete Fourier transform, and compute the Fourier transforms of primitive Dirichlet characters (a minor reformulation of a result from an earlier PR). 

---

- [x] depends on: #13579
- [x] depends on: #13872
- [x] depends on: #13878
- [x] depends on: #13937


Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,125 @@
...
+lemma stdAddChar_apply {N : ℕ+} (j : ZMod N) :
+    stdAddChar j = ↑(toCircle j) :=
+  rfl
+
+section fourier
+
+open MeasureTheory
+
+/-- The discrete measurable space structure (every set is measurable). -/
+local instance instMeasurableSpaceZMod (N : ℕ+) : MeasurableSpace (ZMod N) := ⊤
```
jcommelin, line 87: Should this use any `n : Nat` together with an assumption `[NeZero n]`? It would apply to `ZMod n` even if `n` a `PNat` coerced to `Nat`.

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,125 @@
...
+    (val_injective N).eq_iff] at hxy <;>
+  exact ⟨by positivity, by simpa only [zero_add, div_lt_one this, Nat.cast_lt] using val_lt _⟩
+
+/-- The additive character from `ZMod N` to the unit circle in `ℂ`, sending `j mod N` to
+`exp (2 * π * I * j / N)`. -/
+noncomputable def toCircle {N : ℕ+} : AddChar (ZMod N) circle where
+  toFun := fun j ↦ (toAddCircle j).toCircle
+  map_add_mul' a b := by simp_rw [map_add, AddCircle.toCircle_add]
+  map_zero_one' := by simp_rw [map_zero, AddCircle.toCircle, ← QuotientAddGroup.mk_zero,
+    Function.Periodic.lift_coe, mul_zero, expMapCircle_zero]
```
YaelDillies, line 55:
```suggestion
noncomputable def toCircle {N : ℕ+} : AddChar (ZMod N) circle :=
  AddChar.toAddMonoidHomEquiv.symm toAddCircle 
```

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,125 @@
...
+
+section fourier
+
+open MeasureTheory
+
+/-- The discrete measurable space structure (every set is measurable). -/
+local instance instMeasurableSpaceZMod (N : ℕ+) : MeasurableSpace (ZMod N) := ⊤
+
+/-- The discrete Fourier transform on `ℤ / Nℤ`. -/
+noncomputable def discreteFourierTransform {N : ℕ+} (Φ : ZMod N → ℂ) (k : ZMod N) : ℂ :=
```
YaelDillies, line 90: This is an unbearably long name to include in lemma names.

Mathlib/Analysis/Fourier/ZMod.lean
YaelDillies, line 87: This should just be a global instance

loefflerd: Aha, that's a really good idea! In the next phase of the project I was constantly getting tangled in things like (1 : ℕ+) not being the same as `(1 : ℕ)` and I was sure there had to be a better way. This is clearly the right approach, thanks!

Mathlib/Analysis/Fourier/ZMod.lean
loefflerd: This does not work. 

Firstly `AddChar.toAddMonoidHomEquiv` has some arguments; you probably meant
`(AddChar.toAddMonoidHomEquiv _ _).symm toAddCircle`. But that just exposes the more serious problem that `UnitAddCircle` is not defined as the additivization of `circle`.

Mathlib/Analysis/Fourier/ZMod.lean
loefflerd: Happy to rename it as `ZMod.dft` if you prefer.

Mathlib/Analysis/Fourier/ZMod.lean
YaelDillies: First problem is solved in #13579. Second problem, okay sure

Mathlib/Analysis/Fourier/ZMod.lean
YaelDillies: Yeah, that would be great

loefflerd:
Thanks for the reviews guys! I'm going to review #13579 and then merge it here since it's clearly relevant.

YaelDillies (commented):
Following the Zulip discussion, what is your plan between the following two?
* Copy over the API
* Make the user unfold `dft` every time they need a `fourierIntegral` lemma?

loefflerd:
I'm thinking of a middle ground. If applying some general lemma from the Fourier transform API "just works", then I don't see there's a pressing need to create a copy. So e.g. compatibility with scalar-multiplication works out of the box, but addition needs some side conditions:
```lean
example {N : ℕ} [NeZero N] (Φ : ZMod N → ℂ) (c : ℂ) : 𝓕 (c • Φ) = c • 𝓕 Φ := 
  VectorFourier.fourierIntegral_smul_const .. -- no need for a special DFT lemma

example {N : ℕ} [NeZero N] (Φ₁ Φ₂ : ZMod N → ℂ) : 𝓕 (Φ₁ + Φ₂) = 𝓕 Φ₁ + 𝓕 Φ₂ := by
  rw [Eq.comm] -- doh!
  apply VectorFourier.fourierIntegral_add 
  · continuity
  · continuity
  · measurability
  · measurability
```
(This example also reveals that (a) `VectorFourier.fourierIntegral_add` is clearly the wrong way around, and (b) we need a `TopologicalSpace` instance for `ZMod` as well as a `MeasurableSpace` one.)

YaelDillies:
I'm not sure I like this middle-ground since you lose the ability to rewrite. Personally, I heavily rely on the rewriting behavior and can't afford to lose it.

On the other hand, if you are happy for my discrete Fourier transform API (for general finite abelian groups) to coexist (at least temporarily) with the content of this PR, then I don't really care about the technicalities of this PR and am basically happy with it.

loefflerd:
> I'm not sure I like this middle-ground since you lose the ability to rewrite. Personally, I heavily rely on the rewriting behavior and can't afford to lose it.
> 
> On the other hand, if you are happy for my discrete Fourier transform API (for general finite abelian groups) to coexist (at least temporarily) with the content of this PR, then I don't really care about the technicalities of this PR and am basically happy with it.

I'm completely happy for multiple approaches to discrete Fourier theory to coexist.

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,131 @@
...
+
+open MeasureTheory
+
+/-- The discrete measurable space structure (every set is measurable). -/
+instance : MeasurableSpace (ZMod N) := ⊤
+
+/-- The discrete topology (every set is open). -/
+instance : TopologicalSpace (ZMod N) := ⊥
+
+instance : DiscreteTopology (ZMod N) := ⟨rfl⟩
```
YaelDillies, line 92: Can all of these move to earlier files?

YaelDillies: [Fin.instMeasurableSpace](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/MeasurableSpace/Basic.html#Fin.instMeasurableSpace) is in `MeasureTheory.MeasurableSpace.Basic`, so I guess the `MeasurableSpace (ZMod n)` instance could go there?

loefflerd: That file's already 2200 lines long; we should be moving stuff out of it, not adding more.

loefflerd: See [#13937](https://github.com/leanprover-community/mathlib4/pull/13937)

YaelDillies: Okay, what about now?

loefflerd: Give me a few hours, I do have a day job too 😄

loefflerd:
Ready for review once more

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,67 @@
...
+
+/-!
+# Fourier theory on `ZMod N`
+
+Basic definitions and properties of the discrete Fourier transform for functions on `ZMod N`.
+
+### Main definitions and results
+
+* `ZMod.dft`: the Fourier transform with respect to the standard additive character
+  `ZMod.stdAddChar`(mapping `j mod N` to `exp (2 * π * I * j / N)`). The notation `𝓕`, scoped in
```
YaelDillies, line 18:
```suggestion
  `ZMod.stdAddChar` (mapping `j mod N` to `exp (2 * π * I * j / N)`). The notation `𝓕`, scoped in
```

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
```diff
@@ -235,3 +236,43 @@ open AddCircle
...
+
+/-- The additive character from `ZMod N` to the unit circle in `ℂ`, sending `j mod N` to
+`exp (2 * π * I * j / N)`. -/
+noncomputable def toCircle : AddChar (ZMod N) circle where
+  toFun := fun j ↦ (toAddCircle j).toCircle
+  map_add_eq_mul' a b := by simp_rw [map_add, AddCircle.toCircle_add]
+  map_zero_eq_one' := by simp_rw [map_zero, AddCircle.toCircle, ← QuotientAddGroup.mk_zero,
+    Function.Periodic.lift_coe, mul_zero, expMapCircle_zero]
+
+lemma toCircle_coe (j : ℤ) :
```
YaelDillies, line 258:
```suggestion
lemma toCircle_intCast (j : ℤ) :
```

Mathlib/MeasureTheory/MeasurableSpace/Instances.lean
YaelDillies, line 1: You're missing `DiscreteMeasurableSpace (ZMod N)`

loefflerd: It's not needed, since the following works:
```lean
example (n : ℕ) : DiscreteMeasurableSpace (ZMod n) := inferInstance
```

YaelDillies: It's putting a bit more trust in looping instances than I am comfortable with, but sure

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
```diff
@@ -5,6 +5,7 @@ Authors: Yury G. Kudryashov
 -/
 import Mathlib.Analysis.Complex.Circle
 import Mathlib.Analysis.SpecialFunctions.Complex.Log
+import Mathlib.Algebra.Group.AddChar
```
YaelDillies, line 8:
```suggestion
import Mathlib.Algebra.Group.AddChar
import Mathlib.Analysis.Complex.Circle
import Mathlib.Analysis.SpecialFunctions.Complex.Log
```

Mathlib/Topology/Instances/AddCircle.lean
```diff
@@ -693,3 +693,36 @@ end ZeroBased
...
+namespace ZMod
+
+variable {N : ℕ} [NeZero N]
+
+/-- The `AddMonoidHom` from `ZMod N` to `ℝ / ℤ` sending `j mod N` to `j / N mod 1`. -/
+noncomputable def toAddCircle : ZMod N →+ UnitAddCircle :=
+  lift N ⟨AddMonoidHom.mk' (fun j ↦ ↑(j / N : ℝ)) (by simp [add_div]),
+    by simp [div_self (NeZero.ne _)]⟩
+
+lemma toAddCircle_coe (j : ℤ) :
```
YaelDillies, line 706:
```suggestion
lemma toAddCircle_intCast (j : ℤ) :
```
and you're missing `toAddCircle_natCast`

Mathlib/Topology/Instances/AddCircle.lean
```diff
@@ -693,3 +693,36 @@ end ZeroBased
...
+variable (N) in
+lemma toAddCircle_injective : Function.Injective (toAddCircle : ZMod N → _) := by
+  intro x y hxy
+  have : (0 : ℝ) < N := Nat.cast_pos.mpr (NeZero.pos _)
+  rwa [toAddCircle_apply, toAddCircle_apply, AddCircle.coe_eq_coe_iff_of_mem_Ico
+    (hp := Real.fact_zero_lt_one) (a := 0), div_left_inj' this.ne', Nat.cast_inj,
+    (val_injective N).eq_iff] at hxy <;>
+  exact ⟨by positivity, by simpa only [zero_add, div_lt_one this, Nat.cast_lt] using val_lt _⟩
+
+lemma toAddCircle_eq_zero {j : ZMod N} :
```
YaelDillies, line 724: You're missing `toAddCircle_inj`, and
```suggestion
@[simp] lemma toAddCircle_eq_zero {j : ZMod N} :
```

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
loefflerd: This is a no-op, right? Or is the order of `import` statements significant somehow?

loefflerd:
Thanks for the (re-)review! Good point about `natCast` as well as `intCast`, that actually simplifies several other arguments since it meshes better with `ZMod.val` being `Nat`-valued. I added the corresponding lemma for `Circle` as well as `AddCircle`.

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
YaelDillies: We are semitrying to keep it alphabetical. And no it's not a no-op, although exceptions are hard to spot

loefflerd: In case it wasn't clear, I have already applied your suggestion; thank you for explaining the reasoning behind it. Are you content with this PR now?

YaelDillies: Sorry, I'm reviewing every time I am commuting (the point of my github status being to let you know about this) and I have very little time on a computer with internet connection, so I can only do superficial reviews. Hopefully I will have time tonight to send this PR off, but no promise.

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -0,0 +1,67 @@
...
+open ZMod
+
+namespace DirichletCharacter
+
+variable {N : ℕ} [NeZero N] (χ : DirichletCharacter ℂ N)
+
+lemma fourierTransform_eq_gaussSum_mulShift (k : ZMod N) :
+    𝓕 χ k = gaussSum χ (stdAddChar.mulShift (-k)) := by
+  simp only [dft_def]
+  refine congr_arg Finset.univ.sum (funext fun j ↦ ?_)
```
YaelDillies, line 57:
```suggestion
  refine Fintype.sum_congr rfl fun j ↦ ?_
```

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
```diff
@@ -235,3 +236,47 @@ open AddCircle
...
+lemma toCircle_intCast (j : ℤ) :
+    toCircle (j : ZMod N) = Complex.exp (2 * π * Complex.I * j / N) := by
+  rw [toCircle, AddChar.coe_mk, AddCircle.toCircle, toAddCircle_intCast,
+    Function.Periodic.lift_coe, expMapCircle_apply]
+  push_cast
+  ring_nf
+
+lemma toCircle_natCast (j : ℕ) :
+    toCircle (j : ZMod N) = Complex.exp (2 * π * Complex.I * j / N) := by
+  rw [← Int.cast_natCast, toCircle_intCast, Int.cast_natCast]
```
YaelDillies, line 267:
```suggestion
  simpa using toCircle_intCast j
```
is a more robust proof

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
```diff
@@ -235,3 +236,47 @@ open AddCircle
...
+
+lemma toCircle_apply (j : ZMod N) :
```
YaelDillies, line 269: I would add a docstring warning that this lemma is "evil" in that it refers to `ZMod.val` which is an unmathematical function that could put you in a cul-de-sac (as a matter of fact, it happened to me several times already) and that the user shouild instead lift `j` to `Int` and use `toCircle_intCast`

YaelDillies (commented):
Rest looks good, except for the fact that your new definition has no API (but apparently that's a conscious design decision on your part)

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
loefflerd: Makes sense. I had noticed the "cul-de-sac" issue which is why I didn't tag this lemma as simp. I will add a warning to the docstring as you suggest.

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
loefflerd: The "more robust proof" does not quite work because it cannot infer `N` in the argument. I went for `simpa using toCircle_intCast (N := N) j`.

Mathlib/Analysis/SpecialFunctions/Complex/Circle.lean
loefflerd: PS. The docstring would be more useful if we had an induction principle for ZMod. Surprisingly this seems to be absent.

Mathlib/Analysis/Fourier/ZMod.lean
loefflerd: I'm afraid your suggestion does not work. (It took me a while to spot since there's a lot of library to compile in between `Topology.Instances.AddCircle` and here.) 

I guess you meant either `Finset.sum_congr rfl fun j _ ↦ ?_` or `Fintype.sum_congr _ _ fun j ↦ ?_`. However, while playing with this I spotted that `congr 1 with j` also works, and IMHO that's nicer than any of these other possibilities, so I am going with that.

loefflerd:
Thanks Yael for the comments – I have incorporated them. I promise there will be more API in follow-up PRs; but this PR is a bit of a pain to work on at present (because it involves edits to widely-separated files in the import hierarchy), so I'd rather get it in as it stands.

YaelDillies:
Okay then

maintainer merge

Ruben-VandeVelde:
Odd.

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13355, opened 2024-05-29
[Merged by Bors] - feat(Topology): some lemmas about continuity of maps out of one point compactifications

dagurtomas:
We prove that a map out of the one point compactification of `X` is continuous if and only if its restriction to `X` satisfies the two properties
1) is continuous
2) is continuous "at infinity", namely that it tends to the value at infinity for the `coclosedCompact` filter on `X`

We deduce that continuous functions out of the one point compactification of the natural numbers correspond to convergent sequences. 

We also prove that the one point compactification of a discrete space is totally separated. 

Co-authored-by: Jireh Loreaux

---
This is all in preparation for the definition of `ℕ ∪ {∞}` as a light profinite set, which is important in light condensed mathematics.


Mathlib/Topology/Compactification/OnePoint.lean
```diff
@@ -72,6 +72,14 @@ scoped notation "∞" => OnePoint.infty
 /-- Coercion from `X` to `OnePoint X`. -/
 @[coe, match_pattern] def some : X → OnePoint X := Option.some
 
+protected lemma «forall» {p : OnePoint X → Prop} :
```
faenuccio, line 75: The `@coe` tag in the previous def does not seem to fire, why cannot we simply write `x : OnePoint X` to mean `some x : OnePoint X`?

Mathlib/Topology/Compactification/OnePoint.lean
```diff
@@ -383,6 +391,21 @@ theorem continuousAt_coe {Y : Type*} [TopologicalSpace Y] {f : OnePoint X → Y}
   rw [ContinuousAt, nhds_coe_eq, tendsto_map'_iff, ContinuousAt]; rfl
 #align alexandroff.continuous_at_coe OnePoint.continuousAt_coe
 
+lemma continuous_iff {X Y : Type*} [TopologicalSpace X]
+    [TopologicalSpace Y] (f : OnePoint X → Y) : Continuous f ↔
+    Tendsto (fun x : X ↦ f x) (coclosedCompact X) (𝓝 (f ∞)) ∧ Continuous (fun x : X ↦ f x) := by
+  simp_rw [continuous_iff_continuousAt, OnePoint.forall, continuousAt_coe, continuousAt_infty']
+  rfl
```
faenuccio, line 398:
```suggestion
  simp only [continuous_iff_continuousAt, OnePoint.forall, continuousAt_coe, continuousAt_infty']
  rfl
```

Mathlib/Topology/Compactification/OnePoint.lean
```diff
@@ -383,6 +391,21 @@ theorem continuousAt_coe {Y : Type*} [TopologicalSpace Y] {f : OnePoint X → Y}
...
+
+lemma continuous_iff_of_discrete {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    [DiscreteTopology X] (f : OnePoint X → Y) :
+    Continuous f ↔ Tendsto (fun x : X ↦ f x) cofinite (𝓝 (f ∞)) := by
+  simp [continuous_iff, cocompact_eq_cofinite, continuous_of_discreteTopology]
+
+lemma continuous_iff_of_nat {Y : Type*} [TopologicalSpace Y] (f : OnePoint ℕ → Y) :
```
faenuccio, line 405: Not sure that `of_nat` is the right name. What about `_from_nat`?

Mathlib/Topology/Compactification/OnePoint.lean
dagurtomas: It works after the `CoeTC`

Mathlib/Topology/Compactification/OnePoint.lean
dagurtomas: Well we have [docs#continuous_of_discreteTopology](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Order.html#continuous_of_discreteTopology), so this is consistent with that. But I agree that `from` would be clearer

riccardobrasca:
Can you add a `def` that allows to extend a continuous function from `X` to a continuous from `OnePoint X` (given the relevant condition)? We probably also want unicity.

More generally, is there a universal property I think (I don't know in which category), sooner or later we want that.

j-loreaux:
The universal property should be something like `CocompactMap`s on `X` extend uniquely to `ContinuousMap`s on `OnePoint X`, if I'm not mistaken. (probably you need Hausdorff because currently `CocompactMap` only uses the `cocompact` filter, not the `coclosedCompact` filter in its definition; this may be an oversight that we should change, I'm not sure.)

dagurtomas:
I added some constructors which take a continuous map and a "limit value" and produce a continuous map from `OnePoint`. For the universal property suggested by Jireh we would either need a new structure `CoclosedcompactMap` or something like that, or refactor the definition of `CocompactMap`

Mathlib/Topology/Compactification/OnePoint.lean
```diff
@@ -383,6 +391,87 @@ theorem continuousAt_coe {Y : Type*} [TopologicalSpace Y] {f : OnePoint X → Y}
...
+
+variable (X) in
+/--
+Continuous maps out of the one point compactification of an infinite discrete space to a Hausdorff
+space correspond bijectively to "convergent" maps out of the discrete space.
+-/
+noncomputable def continuousMapDiscreteEquiv (Y : Type*) [DiscreteTopology X] [TopologicalSpace Y]
+    [T2Space Y] [Infinite X] :
+    C(OnePoint X, Y) ≃ { f : X → Y // ∃ L, Tendsto (fun x : X ↦ f x) cofinite (𝓝 L) } where
+  toFun f := ⟨fun x ↦ f x, ⟨f ∞, continuous_iff_from_discrete _ |>.mp (map_continuous f)⟩⟩
```
faenuccio, line 434:
```suggestion
  toFun f := ⟨(f ·), ⟨f ∞, continuous_iff_from_discrete _ |>.mp (map_continuous f)⟩⟩
```

Mathlib/Topology/Compactification/OnePoint.lean
```diff
@@ -383,6 +391,87 @@ theorem continuousAt_coe {Y : Type*} [TopologicalSpace Y] {f : OnePoint X → Y}
...
+    (f : ℕ → Y) (y : Y) (h : Tendsto f atTop (𝓝 y)) :
+    C(OnePoint ℕ, Y) :=
+  continuousMapMkDiscrete f y (by rwa [Nat.cofinite_eq_atTop])
+
+/--
+Continuous maps out of the one point compactification of `ℕ` to a Hausdorff space `Y` correspond
+bijectively to convergent sequences in `Y`.
+-/
+noncomputable def continuousMapNatEquiv (Y : Type*) [TopologicalSpace Y] [T2Space Y] :
+    C(OnePoint ℕ, Y) ≃ { f : ℕ → Y // ∃ L, Tendsto (fun x : ℕ ↦ f x) atTop (𝓝 L) } := by
```
faenuccio, line 468:
```suggestion
    C(OnePoint ℕ, Y) ≃ { f : ℕ → Y // ∃ L, Tendsto (f ·) atTop (𝓝 L) } := by
```

faenuccio:
Thanks!

faenuccio:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13356, opened 2024-05-29
[Merged by Bors] - feat(AlgebraicGeometry/EllipticCurve/Affine): add further map lemmas

Multramate:
... and make definition and lemma names consistent with downstream files.

---


Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -770,180 +775,252 @@ section BaseChange
 
 /-! ### Maps and base changes -/
 
-variable {A : Type v} [CommRing A] (φ : R →+* A)
```
grhkm21, line 773: Just wondering, is there any "real" reason to rename all the $\varphi$ to $f$?

grhkm21 (approved):
There's a lot of noise and chore changes but the `map` lemmas look good to me.

grhkm21:
Do I have to change any labels after reviewing? It still says "awaiting-review"

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: It doesn't make a difference in this file, but I plan to use `f` for all maps going forward (e.g. in the upcoming `Projective`, `Jacobian`, and `DivisionPolynomial` files), because `\phi` and `\psi` will be reserved for division polynomials.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -770,180 +775,252 @@ section BaseChange
...
+        map_negY, slope_of_Y_ne hx hy]
+      map_simp
+  · rw [slope_of_X_ne fun h => hx <| f.injective h, slope_of_X_ne hx]
+    map_simp
 #align weierstrass_curve.base_change_slope WeierstrassCurve.Affine.map_slope
 
 variable {R : Type r} [CommRing R] (W : Affine R) {S : Type s} [CommRing S] [Algebra R S]
   {A : Type u} [CommRing A] [Algebra R A] [Algebra S A] [IsScalarTower R S A]
-  {B : Type v} [CommRing B] [Algebra R B] [Algebra S B] [IsScalarTower R S B] (ψ : A →ₐ[S] B)
+  {B : Type v} [CommRing B] [Algebra R B] [Algebra S B] [IsScalarTower R S B] (φ : A →ₐ[S] B)
```
alreadydone, line 877: Still using φ here?

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -770,180 +775,252 @@ section BaseChange
...
 where `W` is defined over a subring of a ring `S`, and `F` and `K` are field extensions of `S`. -/
 def map : W⟮F⟯ →+ W⟮K⟯ where
-  toFun := mapFun W ψ
+  toFun := mapFun W φ
   map_zero' := rfl
   map_add' := by
     rintro (_ | @⟨x₁, y₁, _⟩) (_ | @⟨x₂, y₂, _⟩)
     any_goals rfl
     by_cases hx : x₁ = x₂
     · by_cases hy : y₁ = (W.baseChange F).toAffine.negY x₂ y₂
```
alreadydone, line 995: Note that it's possible to merge two cases in the [definition of add](https://github.com/leanprover-community/mathlib4/commit/022e4c6e420719b0408fe3faa211a06e80e196b0#diff-4acfb4ffccff493de337a21876fa44aed15d232b58d810d3498b488e53c86b2cR729-R733)
and [do one less case split](https://github.com/leanprover-community/mathlib4/commit/022e4c6e420719b0408fe3faa211a06e80e196b0#diff-4acfb4ffccff493de337a21876fa44aed15d232b58d810d3498b488e53c86b2cR956-R963).

alreadydone (commented):
LGTM otherwise

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: Oops.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: Thanks.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: Seems like I can't reuse the variable `f`, because otherwise I can't make it implicit when I want it to be, thanks to the bug!

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: This is far too noisy. I'll change this in a separate PR.

alreadydone (approved):
Thanks 🎉 
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

Ruben-VandeVelde:
It seems like this renamed a number of declarations; is there a reason we didn't add deprecated aliases?

alreadydone (commented):
It's my oversight and I agree it's better to add the `deprecated` declarations. Even though the file isn't used outside of EllipticCurve within mathlib, it may be used outside of mathlib, for example [this paper](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.ITP.2023.7), which cites our paper, states intention to use our results.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -708,257 +713,332 @@ noncomputable instance instAddZeroClassPoint : AddZeroClass W.Point :=
...
 
-/-! ### Maps and base changes -/
+lemma map_polynomial : (W.map f).toAffine.polynomial = W.polynomial.map (mapRingHom f) := by
+  simp only [polynomial]
+  map_simp
 
-variable {A : Type v} [CommRing A] (φ : R →+* A)
+lemma map_eval_polynomial (x : R[X]) (y : R) :
+    ((W.map f).toAffine.polynomial.eval <| x.map f).eval (f y) =
+      f ((W.polynomial.eval x).eval y) := by
```
alreadydone, line 786: I think I'll rename this to `map_polynomial_eval_eval` in my affine formulas PR, and I'll also add the single `eval` version).

PR #13287, opened 2024-05-27
[Merged by Bors] - chore: more `refine'` to `refine`

adomani:
A PR analogous to #13166.  The reason some of these `refine'` are now caught is that my previous script missed all instances.  I am not sure why it is still missing other `refine'`.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #11890, opened 2024-04-04
feat: the terminal refine linter

adomani:
A linter that warns on usages of `refine` and `refine'` as a finishing tactic.

See this [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Usage.20of.20refine').

###  Conclusion of the experiment
Systematic replacements of terminal `refine` with `exact` leads to an overall slow-down.

---

- [ ] depends on: #15616 (disable the linter in downstream projects)

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13357, opened 2024-05-29
[Merged by Bors] - chore: final batch of mass replacements `refine'` to `refine`

adomani:
These replacements have been performed completely mechanically by a script and, in the branch with the script, have passed CI with no extra input (except fixing a long line and one indentation issue).

The script exclusively removes `'` from `refine'` and add `?` in front of `_` as appropriate.

Assuming that I did not leave out any cases, the only `refine'`s that could still be replaced by a simple-minded translation `refine' ...` to ``refine [sprinkle some `?`]`` are the 52 that involve at least 5 holes.

---


Mathlib/Algebra/Homology/HomologySequence.lean
```diff
@@ -185,13 +185,13 @@ lemma opcycles_right_exact (S : ShortComplex (HomologicalComplex C c)) (hS : S.E
   have : Epi (ShortComplex.map S (eval C c i)).g := by dsimp; infer_instance
   have hj := (hS.map (HomologicalComplex.eval C c i)).gIsCokernel
   apply ShortComplex.exact_of_g_is_cokernel
-  refine' CokernelCofork.IsColimit.ofπ' _ _  (fun {A} k hk => by
+  refine CokernelCofork.IsColimit.ofπ' _ _  (fun {A} k hk => by
```
rosborn, line 188: I think it would be a good idea to replace the `refine`s with `exact`s where possible. (I can help with that if needed). Also, a `refine` within a `refine` seems like code smell.

adomani: Replacing terminal `refine`s with `exact` turned out to result in slowing mathlib.  There is a PR somewhere with timings and all.

Nested `refine`s: I agree, but I think that it is a separate issue to replacing `refine'` with `refine`.

jcommelin:
!bench

Mathlib/Algebra/Polynomial/UnitTrinomial.lean
```diff
@@ -377,3 +377,4 @@ theorem irreducible_of_coprime' (hp : IsUnitTrinomial p)
 end IsUnitTrinomial
 
 end Polynomial
+
```
sgouezel, line 380: Is there a reason why the script is sometimes adding empty lines to the end of files ?

adomani: I am glad that you asked!  I noticed that an earlier version of the script was careless with ending character, so I fixed it (I think), but some of the earlier modifications were not modified again by the updated script.

I'll fix the endings of the files.

riccardobrasca:
You have a merge conflict... I think we should merge this asap

adomani:
Riccardo, with the last commit, I should have fixed the merge conflict.

riccardobrasca:
Does this include the modification in #13287?

adomani:
No, the two PRs touch independent file (or, if they touch the same file, it is a coincidence).

riccardobrasca:
OK, let's start with merging the other one.

adomani:
Sure, I am waiting for CI to finish and for the `merge-conflict` label to be removed (I fixed the conflict locally).

adomani:
There were conflicts, but I resolved them all locally and I just pushed the resolved branch.

Mathlib/Algebra/Homology/HomologySequence.lean
riccardobrasca: I think both issues can be deal with in separate PRs.

jcommelin:
Thanks :tada: 

bors d+

Mathlib/Algebra/Homology/HomologySequence.lean
adomani: For completeness, #11890 is the PR that systematically replaced every terminal `refine` with `exact`.  You can see there that the modification made mathlib slower overall.  I then selected the files that were *improved* by the modification and PRed those alone.

So, unless there is the desire for a further "terminal `refine` audit", I would be skeptical of mass-replacing finishing `refine`s by `exact`.

rosborn: Interesting. I would not have expected `exact` to be slower than `refine`.

adomani: I was also surprised by this, but luckily the PR was benchmarked before deciding to merge!

riccardobrasca:
I still count more than 600 `refine'`. Do you have plans to remove those too?

riccardobrasca:
In branch `RB/refine` I removed a couple of those, that didn't create any problem. I can of course keep going, but maybe it is worth to understand why your script didn't catch them.

PR #13358, opened 2024-05-29
[Merged by Bors] - feat(LightProfinite): define `ℕ∪{∞}` as a light profinite set

dagurtomas:
- [x] depends on: #13319 
- [x] depends on: #13355 

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #13365, opened 2024-05-30
[Merged by Bors] - feat: instances of the non-unital continuous functional calculus for C⋆-algebras

j-loreaux:
This adds instances of the non-unital continuous functional calculus (for normal, selfadjoint and nonnegative elements) for non-unital C⋆-algebras.

---

I'm sorry the diff on this is so bad. No existing results have been deleted (and one proof was only minimally modified), but this is I think the correct ordering of results in this file.

### DIFF
--- a/Mathlib/Analysis/NormedSpace/Star/ContinuousFunctionalCalculus/Instances.lean
+++ b/Mathlib/Analysis/NormedSpace/Star/ContinuousFunctionalCalculus/Instances.lean
@@ -66,20 +66,20 @@ calc
   _             ≃⋆[𝕜] C(σ 𝕜 (↑a : A⁺¹), 𝕜) := Homeomorph.compStarAlgEquiv'
   _             →⋆ₐ[𝕜] A⁺¹ := cfcHom
 ```
-This range of this map is contained in the range of `(↑) : A → A⁺¹` (see `nucfcAux_mem_range_inr`),
+This range of this map is contained in the range of `(↑) : A → A⁺¹` (see `cfcₙAux_mem_range_inr`),
 and so we may restrict it to `A` to get the necessary homomorphism for the non-unital continuous
 functional calculus.
 -/
-noncomputable def nucfcAux : C(σₙ 𝕜 a, 𝕜)₀ →⋆ₙₐ[𝕜] A⁺¹ :=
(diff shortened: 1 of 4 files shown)


Mathlib/Analysis/NormedSpace/Star/ContinuousFunctionalCalculus/Instances.lean
```diff
@@ -32,19 +33,115 @@ continuous functional calculus, normal, selfadjoint
...
+  C(σₙ 𝕜 a, 𝕜)₀ →⋆ₙₐ[𝕜] C(σₙ 𝕜 a, 𝕜) := ContinuousMapZero.toContinuousMapHom
+  _             ≃⋆[𝕜] C(σ 𝕜 (↑a : A⁺¹), 𝕜) := Homeomorph.compStarAlgEquiv'
+  _             →⋆ₐ[𝕜] A⁺¹ := cfcHom
+```
+This range of this map is contained in the range of `(↑) : A → A⁺¹` (see `nucfcAux_mem_range_inr`),
+and so we may restrict it to `A` to get the necessary homomorphism for the non-unital continuous
+functional calculus.
+-/
+noncomputable def nucfcAux : C(σₙ 𝕜 a, 𝕜)₀ →⋆ₙₐ[𝕜] A⁺¹ :=
+  (cfcHom (R := 𝕜) (hp₁.mpr ha) : C(σ 𝕜 (a : A⁺¹), 𝕜) →⋆ₙₐ[𝕜] A⁺¹) |>.comp
```
dupuisf, line 74: Shouldn't we stick with `cfcₙ` throughout, rather than have a few random `nucfc`?

Mathlib/Analysis/NormedSpace/Star/ContinuousFunctionalCalculus/Instances.lean
```diff
@@ -355,4 +633,45 @@ lemma cfc_nnreal_eq_real {a : A} (f : ℝ≥0 → ℝ≥0) (ha : 0 ≤ a := by c
 
 end NNRealEqReal
 
+section Unitary -- TODO: move to a new file
+
+/-!
+### conditions on unitary elements imposed by the continuous functional calculus
+-/
```
dupuisf, line 640:
```suggestion
/-!
### Conditions on unitary elements imposed by the continuous functional calculus
-/
```

dupuisf (commented):
Looks good!

bors d+

PR #11486, opened 2024-03-18
[Merged by Bors] - chore: Rename `nat_cast`/`int_cast`/`rat_cast` to `natCast`/`intCast`/`ratCast`

YaelDillies:
Now that I am defining `NNRat.cast`, I want a definitive answer to this naming issue. Plenty of lemmas in mathlib already use `natCast`/`intCast`/`ratCast` over `nat_cast`/`int_cast`/`rat_cast`, and this matches with the general expectation that underscore-separated name parts correspond to a single declaration.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #10185, opened 2024-02-02
[Merged by Bors] - chore: rename op_norm to opNorm

sgouezel:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13368, opened 2024-05-30
[Merged by Bors] - chore: add deprecation dates for `nat_cast` and `int_cast` lemmas

sgouezel:
Deprecations have been forgotten in #11486. We add them in this PR.

Methodology: start from the last commit of #11486, use the script by @adomani in https://github.com/leanprover-community/mathlib4/pull/10185#issuecomment-1925481325 to generate deprecation aliases (I am also adding an updated version of the script in a comment below) and commit. Then cherry pick this last commit on master and fix whatever has changed in between. This is not perfect as declarations which have been moved since #11486 won't get their deprecated alias, but this should be good enough.

---

### DIFF
--- a/Mathlib/Algebra/CharP/ExpChar.lean
+++ b/Mathlib/Algebra/CharP/ExpChar.lean
@@ -401,6 +401,9 @@ theorem frobenius_natCast (n : ℕ) : frobenius R p n = n :=
   map_natCast (frobenius R p) n
 #align frobenius_nat_cast frobenius_natCast
 
+@[deprecated (since := "2024-04-17")]
+alias frobenius_nat_cast := frobenius_natCast
+
 variable {R}
 
 theorem list_sum_pow_char (l : List R) : l.sum ^ p = (l.map (· ^ p : R → R)).sum :=
--- a/Mathlib/Algebra/CharZero/Lemmas.lean
+++ b/Mathlib/Algebra/CharZero/Lemmas.lean
@@ -67,9 +67,15 @@ lemma support_natCast (hn : n ≠ 0) : support (n : α → M) = univ :=
   support_const <| Nat.cast_ne_zero.2 hn
 #align function.support_nat_cast Function.support_natCast
 
+@[deprecated (since := "2024-04-17")]
+alias support_nat_cast := support_natCast
+
 lemma mulSupport_natCast (hn : n ≠ 1) : mulSupport (n : α → M) = univ :=
   mulSupport_const <| Nat.cast_ne_one.2 hn
 #align function.mul_support_nat_cast Function.mulSupport_natCast
+
+@[deprecated (since := "2024-04-17")]
+alias mulSupport_nat_cast := mulSupport_natCast
 
 end Function
 end AddMonoidWithOne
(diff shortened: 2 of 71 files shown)


sgouezel:
The script I have used, due to @adomani in #10185 and barely adapted:
```
#! /bin/bash

if [ -z "${1}" ]
then
  commit="$( git merge-base master "$( git rev-parse --abbrev-ref HEAD )" )"
  read -p $'Type a commit number such that all diff lines containing `theorem/def`
correspond to deprecated declarations (use '"'${commit}'"' otherwise):
' comm
  [ -n "${comm}" ] && commit=${comm}
else
  commit="${1}"
fi

git cat-file -e "${commit}" || exit 1

##  `mkDeclAndDepr <file>` outputs a single line of the form
##  `@[deprecated (since := "yyyy-mm-dd")]||||alias yyy := xxx@@@`
##  for each modified declaration in `<file>`.
##  The separators `@@@` delimit different declarations.
##  The separators `||||` are later replaced by line breaks.
## To use a specific date, replace $( date +%Y-%m-%d ) with 2024-04-17 for instance

mkDeclAndDepr () {
  git diff --unified=0 "${commit}" "${1}" |
    awk -v date="$( date +%Y-%m-%d )" 'function depr(ol,ne) {
      return sprintf("@[deprecated (since := \"%s\")]||||alias %s := %s", date, ol, ne)
    }
    /^-[^+-]*(theorem|lemma)/ {
      for(i=1; i<=NF; i++) {
        if (($i ~ /theorem$/) || ($i ~ /lemma$/)) { old=$(i+1) }
      }
    }
    /^+[^+-]*(theorem|lemma)/ {
      for(i=1; i<=NF; i++) {
        if (($i ~ /theorem$/) || ($i ~ /lemma$/)) {
          sub(/^+/, "", $i)
          printf("%s %s ,%s@@@", $i, $(i+1), depr(old, $(i+1)))
        }
      }
    }'
}

##  `addDeprecations <file>` adds the deprecation statements to `<file>`,
##  using the first new line after the start of each declaration as position
addDeprecations () {
  awk -v data="$( mkDeclAndDepr "${1}" )" 'BEGIN{
    found=0
    split(data, pairs, "@@@")  ## we setup the data:
    for(i in pairs) {
      if (pairs[i] ~ ",") {
        split(pairs[i], declDepr, ",")
        lines[i]=declDepr[1]   ## `lines` contains `theorem/lemma name`s
        deprs[i]=declDepr[2]   ## `deprs` contains the deprecation statements
      }
    }
    currDep=""
    ## scanning the file, if we find an entries of `lines`, the we assign `currDep`
  } /theorem|lemma/ { for(l in lines) { if ($0 ~ lines[l]) { found=1; currDep=deprs[l] } } } {
    ## when we find the next empty line, we print the deprecation statement in `currDep`
    if ((found == 1) && (NF == 0)) {
      found=0
      printf("\n%s\n", currDep)
    }  ## we print all the lines anyway
    print $0 }
   END{ # in case the statement to deprecate is the last of the file
    if (found == 1) { printf("\n%s\n", currDep) } }' "${1}" |
  sed 's=||||=\n=g'
}

##  loops through the changed files and runs `addDeprecations` on each one of them
new="i_am_a_file_name_with_no_clashes"
while [ -f "${new}" ]; do new=${new}0; done
for fil in $( git diff --name-only ${commit} ); do
  printf $'Processing %s\n' "${fil}"
  addDeprecations "${fil}" > "${new}" ; mv "${new}" "${fil}"
done
```

jcommelin (approved):
Thanks :tada:

bors merge

adomani:
With the "experience" of `move-decls`, this script could probably be upgraded and made more stable.  Is this a common enough situation that it would be desirable, or is the current status ok?

PR #13369, opened 2024-05-30
[Merged by Bors] - fix(Mathlib/Tactic/CC/Addition): `cc` tactic throws error for literals

Komyyy:
```lean
example : nat_lit 0 = nat_lit 0 := by
  cc
-- failed

example : "Miyahara Kō" = "Miyahara Kō" := by
  cc (config := { values := false })
-- failed
```
The cause of this error is that code on `Mathlib.Tactic.CC.CCM.internalizeAppLit` from L1222 onwards doesn't assume that the input is `Expr.lit`.
This error can be fixed simply by adding `unless e.isApp do return`.
Also, we add a test for this bug.

p.s. This is an new error in Lean 4 because Lean 3 has no `Expr.lit` so we doesn't have to fix code in Lean 3.

---

### DIFF
--- a/Mathlib/Tactic/CC/Addition.lean
+++ b/Mathlib/Tactic/CC/Addition.lean
@@ -1219,6 +1219,7 @@ partial def internalizeAppLit (e : Expr) : CCM Unit := do
   else
     mkEntry e false
     if (← get).values && isValue e then return -- we treat values as atomic symbols
+  unless e.isApp do return
   if let some (_, lhs, rhs) ← e.relSidesIfSymm? then
     internalizeCore lhs (some e)
     internalizeCore rhs (some e)
--- a/Mathlib/Tactic/CC/Addition.lean
+++ b/Mathlib/Tactic/CC/Addition.lean
@@ -1219,6 +1219,7 @@ partial def internalizeAppLit (e : Expr) : CCM Unit := do
   else
     mkEntry e false
     if (← get).values && isValue e then return -- we treat values as atomic symbols
+  -- At this point we should have handled a literal; otherwise we fail.
   unless e.isApp do return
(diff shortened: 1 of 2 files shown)


Mathlib/Tactic/CC/Addition.lean
```diff
@@ -1219,6 +1219,7 @@ partial def internalizeAppLit (e : Expr) : CCM Unit := do
   else
     mkEntry e false
     if (← get).values && isValue e then return -- we treat values as atomic symbols
+  unless e.isApp do return
```
Vierkantor, line 1222:
```suggestion
  -- At this point we should have handled a literal; otherwise we fail.
  unless e.isApp do return
```

Vierkantor (commented):
Unfortunately the following doesn't seem to work:

```lean
example : nat_lit 0 = nat_lit 0 := by
  cc
```

I think the correct fix is to check `e.isRawNatLit` and if so, replace `e` with an expression shaped like `(OfNat.ofNat _ e)`.

Komyyy:
> Unfortunately the following doesn't seem to work:
> 
> ```lean
> example : nat_lit 0 = nat_lit 0 := by
>   cc
> ```
This works, as you can see in `test/cc.lean`.

Did you mean this?
```lean
example : nat_lit 0 = 0 := by
  cc
```
I think this is another issue which should be solved in another PR.

Komyyy:
Even if we fix `nat_lit` problem, this bug occurs in this form:
```lean
example : "Anne Baanen" = "Anne Baanen" := by
  cc (config := { values := false })
-- failed
```

Komyyy:
I changed the title and the description, and added the new test.

Vierkantor (approved):
I see. Okay, then let's merge this fix.

bors r+

PR #13379, opened 2024-05-30
[Merged by Bors] - feat: simp lemma replacing MulEquiv.toMonoidHom with coercion

kim-em:
I was surprised to encounter post-simp goals that included both `x.toMonoidHom` and the coercion.

I think it would be good to follow up by avoiding `toMonoidHom` in favour of the coercion in the first place.


Ruben-VandeVelde (approved):
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #7721, opened 2023-10-17
[Merged by Bors] - feat(CategoryTheory/Adhesive): Sheaf toposes are finitary extensive and adhesive

erdOne:
Port of https://github.com/leanprover-community/mathlib/pull/17637

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13386, opened 2024-05-30
[Merged by Bors] - feat(CategoryTheory): surjective-injective factorizations in concrete categories

joelriou:
In this PR, we formalize the notion of surjective/injective factorizations in concrete categories. We show that this obviously holds in the category of types. In future PRs, it shall be promoted to locally-surjective/locally-injective factorizations in categories of (pre)sheaves, and this shall be used in order to relate locally surjective morphisms and epimorphisms in categories of sheaves (beyond the case of sheaves of types which was obtained in #7721 and #13318).

---

- [x] depends on: #13331


erdOne:
Should the spelling be surjective-injective or epi-mono?

joelriou:
This API (and the downstream applications I intend to do) is very much a surjective/injective thing. (In the future, we may deduce this property from epi/mono factorizations, for categories where we can compare epi/surjective and mono/injective, but I feel it would be ackward otherwise.)

erdOne:
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #12978, opened 2024-05-17
[Merged by Bors] - feat(FiberedCategory/HomLift): definition of IsHomLift

callesonne:
We define the notion `IsHomLift` which aims to provide API for working with equalities `p(\phi) = f` for a functor `p` and morphisms `\phi`, `f`. This API is extensively used in an upcoming PR defining the notion of fibered categories. See the repository [Stacks-project](https://github.com/Paul-Lez/Stacks-project) for WIP using this notion.

Co-authored-by: Paul Lezau <paul.lezeau@gmail.com>

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13388, opened 2024-05-30
feat(FiberedCategory/BasedCategory): add bicategory of based categories

callesonne:
We define the strict bicategory of based categories. Its objects are categories equipped with a functor to some fixed base-category. The strict bicategory of fibered categories will later be built on top of this one.

---

- [ ] depends on: #12978


Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,259 @@
...
+universe u₁ v₁
+
+open CategoryTheory Functor Category NatTrans IsHomLift
+
+variable {𝒮 : Type u₁} [Category.{v₁} 𝒮]
+
+/-- A based category over `𝒮` is a category `𝒳` together with a functor `p : 𝒳 ⥤ 𝒮` -/
+structure BasedCategory (𝒮 : Type u₁) [Category.{v₁} 𝒮] where
+  /-- The type of objects in a `BasedCategory`-/
+  cat : Type u₁
```
joelriou, line 36: I think it is important to allow different universe parameters for the categories above and below.

callesonne: This is what I had originally, but if I let `cat` have type `u₂` then the linter complains that `u₂` and `u₁` always appear together.

joelriou: I believe this is a case where it is ok to insert `@[nolint checkUnivs]` before the declaration. Then, the `universe` declaration at the beginning of the file should be `universe v₂ u₂ v₁ u₁` (with `cat : Type u₂` and `[... : Category.{v₂} cat]`. Then, we may use `BasedCategory.{v₂, u₂} 𝒮` in order to refer explicitly to the category of based caetgory with certain universe parameters.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,259 @@
...
+  /-- The underlying category of a `BasedCategory` -/
+  isCat : Category.{v₁} cat
```
joelriou, line 38: I would rather rename `cat` as `obj` and `isCat` as `category`. A slightly better syntax would be `[category : Category.{v₂} obj]`. You should also define a constructor `BasedCategory.of` which would take a functor as input.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,259 @@
...
+  /-- The functor to the base -/
+  p : cat ⥤ 𝒮
+
+instance (𝒳 : BasedCategory 𝒮) : Category 𝒳.cat := 𝒳.isCat
+
+/-- A functor between based categories is a functor between the underlying categories that commutes
+with the projections. -/
+structure BasedFunctor (𝒳 𝒴 : BasedCategory 𝒮) extends CategoryTheory.Functor 𝒳.cat 𝒴.cat where
```
joelriou, line 46: I am not sure using the `extends` syntax is better compared to just having a field `functor : 𝒳.obj ⥤ 𝒴.obj`.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,259 @@
...
+def whiskerRight {𝒵 : BasedCategory 𝒮} {F G : BasedFunctor 𝒳 𝒴} (α : F ⟶ G)
+    (H : BasedFunctor 𝒴 𝒵) : BasedFunctor.comp F H ⟶ BasedFunctor.comp G H where
+  toNatTrans := CategoryTheory.whiskerRight α.toNatTrans H.toFunctor
+  aboveId' := fun a => by apply BasedFunctor.pres_IsHomLift
+
+end
+
+/-- `BasedCategory 𝒮` forms a bicategory -/
+instance bicategory : Bicategory (BasedCategory 𝒮) where
+  Hom 𝒳 𝒴 := BasedFunctor 𝒳 𝒴
```
joelriou, line 243: I have not worked with bicategories in mathlib, but I think you could define the `CategoryStruct` instance on `BasedCategory 𝒮` first, prove a few lemmas/definitions using this instance (which allow you to use the notations `≫` and `𝟙`), and finally only defining the `Bicategory` instance.

callesonne:
Okay this merge with mathlib went awfully wrong for some reason. I think I will just remake this PR

PR #13387, opened 2024-05-30
[Merged by Bors] - chore(Algebra/Lie/Submodule): the lattice of Lie submodules is atomic

jcommelin:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13265, opened 2024-05-27
[Merged by Bors] - feat(Algebra/Lie): Killing Lie algebras are semisimple

jcommelin:
---

- [x] depends on: #13217
- [x] depends on: #13366
- [x] depends on: #13367
- [x] depends on: #13370
- [x] depends on: #13384
- [x] depends on: #13387
- [x] depends on: #13391

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13391, opened 2024-05-30
[Merged by Bors] - feat(Algebra/Lie/InvariantForm): Lie algebras with a nondegenerate invariant form are semisimple

jcommelin:
- [x] depends on: #13387

---

### DIFF
--- a/Mathlib/Algebra/Lie/InvariantForm.lean
+++ b/Mathlib/Algebra/Lie/InvariantForm.lean
@@ -1,0 +1,207 @@
+/-
+Copyright (c) 2024 Johan Commelin. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Johan Commelin
+-/
+import Mathlib.Algebra.Lie.Semisimple.Defs
+import Mathlib.LinearAlgebra.BilinearForm.Orthogonal
+
+/-!
+# Lie algebras with non-degenerate invariant bilinear forms are semisimple
+
+In this file we prove that a finite-dimensional Lie algebra over a field is semisimple
+if it does not have non-trivial abelian ideals and it admits a
+non-degenerate reflexive invariant bilinear form.
+Here a form is *invariant* if it is compatible with the Lie bracket: `Φ ⁅x, y⁆ z = Φ x ⁅y, z⁆`.
+
+## Main results
+
+* `LieAlgebra.InvariantForm.orthogonal`: given a Lie submodule `N` of a Lie module `M`,
+  we define its orthogonal complement with respect to a non-degenerate invariant bilinear form `Φ`
+  as the Lie ideal of elements `x` such that `Φ n x = 0` for all `n ∈ N`.
+* `LieAlgebra.InvariantForm.isSemisimple_of_nondegenerate`: the main result of this file;
+  a finite-dimensional Lie algebra over a field is semisimple
+  if it does not have non-trivial abelian ideals and admits
+  a non-degenerate invariant reflexive bilinear form.
+
+## References
+
+We follow the short and excellent paper [dieudonne1953].
+-/
+
+namespace LieAlgebra
+
+namespace InvariantForm
+
+section ring
+
+variable {R L M : Type*}
+variable [CommRing R] [LieRing L] [LieAlgebra R L]
+variable [AddCommGroup M] [Module R M] [LieRingModule L M]
+
+variable (Φ : LinearMap.BilinForm R M) (hΦ_nondeg : Φ.Nondegenerate)
+
+variable (L) in
+/--
+A bilinear form on a Lie module `M` of a Lie algebra `L` is *invariant* if
+for all `x : L` and `y z : M` the condition `Φ ⁅x, y⁆ z = -Φ y ⁅x, z⁆` holds.
+-/
+def _root_.LinearMap.BilinForm.lieInvariant : Prop :=
+  ∀ (x : L) (y z : M), Φ ⁅x, y⁆ z = -Φ y ⁅x, z⁆
+
+lemma _root_.LinearMap.BilinForm.lieInvariant_iff [LieModule R L M] :
+    Φ.lieInvariant L ↔ Φ ∈ LieModule.maxTrivSubmodule R L (LinearMap.BilinForm R M) := by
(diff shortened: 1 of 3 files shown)


jcommelin:
* **#13265** <a href="https://app.graphite.dev/github/pr/leanprover-community/mathlib4/13265?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13391** <a href="https://app.graphite.dev/github/pr/leanprover-community/mathlib4/13391?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @jcommelin and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Mathlib/Algebra/Lie/Submodule.lean
```diff
@@ -1545,3 +1552,29 @@ def LieModuleEquiv.ofTop : (⊤ : LieSubmodule R L M) ≃ₗ⁅R,L⁆ M :=
...
+termination_by N => finrank K N
+decreasing_by exact finrank_lt_finrank_of_lt hbN
+
+instance [FiniteDimensional K V] : IsAtomic (LieSubmodule K L V) := by
+  constructor
+  intro N
+  rw [or_iff_not_imp_left]
+  exact exists_atom_le_of_finite N
+
+end LieSubmodule
```
ocfnash: This should be gone, right?

Mathlib/Algebra/Lie/InvariantForm.lean
```diff
@@ -0,0 +1,207 @@
...
+import Mathlib.Algebra.Lie.Semisimple.Defs
+import Mathlib.LinearAlgebra.BilinearForm.Orthogonal
+
+/-!
+# Lie algebras with non-degenerate invariant bilinear forms are semisimple
+
+In this file we prove that a finite-dimensional Lie algebra over a field is semisimple
+if it does not have non-trivial abelian ideals and it admits a
+non-degenerate reflexive invariant bilinear form.
+Here a form is *invariant* if it is compatible with the Lie bracket: `Φ ⁅x, y⁆ z = Φ x ⁅y, z⁆`.
```
ocfnash, line 15: I think I'd say something like:
```suggestion
Here a form is *invariant* if it invariant under the Lie bracket in the sense that ⁅x, Φ⁆ = 0 for all `x` or equivalently, `Φ ⁅x, y⁆ z = Φ x ⁅y, z⁆`.
```
to justify the terminology a bit

Mathlib/Algebra/Lie/InvariantForm.lean
```diff
@@ -0,0 +1,207 @@
...
+variable [AddCommGroup M] [Module K M] [LieRingModule L M]
+
+variable [Module.Finite K L]
+variable (Φ : LinearMap.BilinForm K L) (hΦ_nondeg : Φ.Nondegenerate)
+variable (hΦ_inv : Φ.lieInvariant L) (hΦ_refl : Φ.IsRefl)
+-- TODO: replace the following assumption by a typeclass assumption `[HasNonAbelianAtoms]`
+variable (hL : ∀ I : LieIdeal K L, IsAtom I → ¬IsLieAbelian I)
+
+open FiniteDimensional Submodule in
+lemma orthogonal_isCompl_coe_submodule (I : LieIdeal K L) (hI : IsAtom I) :
```
ocfnash, line 124: I don't think I'd bother with this lemma: it's the one below that we really want, right?

Hopefully using `LieSubmodule..isCompl_iff_coe_toSubmodule` it should be possible to survive with just `orthogonal_isCompl`?

Mathlib/Algebra/Lie/InvariantForm.lean
```diff
@@ -0,0 +1,207 @@
...
+
+lemma restrict_orthogonal_nondegenerate (I : LieIdeal K L) (hI : IsAtom I) :
+    (Φ.restrict (orthogonal Φ hΦ_inv I)).Nondegenerate := by
+  rw [LinearMap.BilinForm.restrict_nondegenerate_iff_isCompl_orthogonal hΦ_refl]
+  simp only [LieIdeal.coe_to_lieSubalgebra_to_submodule, orthogonal_toSubmodule,
+    LinearMap.BilinForm.orthogonal_orthogonal hΦ_nondeg hΦ_refl]
+  exact (orthogonal_isCompl_coe_submodule Φ hΦ_nondeg hΦ_inv hΦ_refl hL I hI).symm
+
+open FiniteDimensional Submodule in
+lemma atomistic : ∀ I : LieIdeal K L, sSup {J : LieIdeal K L | IsAtom J ∧ J ≤ I} = I := by
```
ocfnash, line 149: I can't help wondering if this might hold if we just have `[IsArtinian R L]` but I really don't think you should bother trying!

ocfnash (approved):
Very nice!

Mathlib/Algebra/Lie/InvariantForm.lean
jcommelin: But this version is used multiple times below. So I would rather keep it.

Mathlib/Algebra/Lie/InvariantForm.lean
jcommelin: I thought about it for a bit, but we need to prove `J ⊔ J' = ⊤`, and I don't see how to do that if we're note over a field.

ocfnash:
Incidentally, I wonder if there are more general for `LieSubmodule R L M` rather than just `LieIdeal R L`.

I guess we'd have to start back at `lie_eq_self_of_isAtom_of_nonabelian`. I haven't thought about it but mindless generalisation would say that `lie_eq_self_of_isAtom_of_nonabelian` would have to make some statement of the form `⁅I, N⁆ = N` given `⁅I, N⁆ ≠ ⊥` (where now `N` is an atom rather than `I`). 

Anyway, I thought I had already delegated this!

bors merge

jcommelin:
I will just record that I also spent some time thinking about *that* generalization, and it didn't lead anywhere. It seems that the equality $$[I, I] = I$$ is pretty hard-coded into the current proof, and a version for submodules of a general Lie algebra presentation doesn't really seem to work.

PR #13393, opened 2024-05-30
[Merged by Bors] - feat(FiberedCategory/Cartesian): define cartesian morphisms

callesonne:
We define cartesian morphisms with respect to a functor `p : 𝒳 ⥤ 𝒮`, as in SGA 1 VI 5.1. We also provide some basic API for working with the universal property.

---

- [x] depends on: #12978


callesonne:
Here is one thing that I'm not sure about this PR: based on the `BasedCategory` PR, this could all be rewritten to take based categories as parameters instead of functors `p : 𝒳 ⥤ 𝒮`. I think this would look cleaner, but maybe there is not much benefit in doing this since we are not talking about functors in this file anyways.

Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
```diff
@@ -0,0 +1,119 @@
...
+lifting `𝟙 R`, then `inducedMap f φ φ'` is the morphism `a' ⟶ a` obtained from the universal
+property of `φ`. -/
+noncomputable def inducedMap : a' ⟶ a :=
+  Classical.choose <| IsCartesian.universal_property (p:=p) (f:=f) (φ:=φ) φ'
+
+instance inducedMap_isHomLift : IsHomLift p (𝟙 R) (inducedMap p f φ φ') :=
+  (Classical.choose_spec <| IsCartesian.universal_property (p:=p) (f:=f) (φ:=φ) φ').1.1
+
+@[simp]
+lemma inducedMap_comp : (inducedMap p f φ φ') ≫ φ = φ' :=
```
joelriou, line 56:
```suggestion
@[reassoc (attr := simp]
lemma inducedMap_comp : inducedMap p f φ φ' ≫ φ = φ' :=
```

Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
```diff
@@ -0,0 +1,119 @@
...
+variable {R S : 𝒮} {a b : 𝒳} (f : R ⟶ S) (φ : a ⟶ b) [IsCartesian p f φ]
+
+section
+
+variable {a' : 𝒳} (φ' : a' ⟶ b) [IsHomLift p f φ']
+
+/-- Given a cartesian arrow `φ : a ⟶ b` lying over `f : R ⟶ S` in `𝒳`, a morphism `φ' : a' ⟶ b`
+lifting `𝟙 R`, then `inducedMap f φ φ'` is the morphism `a' ⟶ a` obtained from the universal
+property of `φ`. -/
+noncomputable def inducedMap : a' ⟶ a :=
```
joelriou, line 49: I am not sure about the namespaces here. I would suggest placing `inducedMap` (and some other lemmas and definitions) in the `Functor` namespace, as `Functor.liftOfIsCartesian`. Below, the factorization could be `fac_of_isCartesian`. (This way, the names would kind of parallel the names in the limits api.)

callesonne: Good suggestion!

callesonne: I ended up trying to mirror the `IsLimit` names like your suggestion, but I'm not sure about the `Functor` namespace. Partly because this leads to especially long names later on like `liftOfIsStronglyCartesian`, and also because it doesn't feel like these are properties of the functor but rather the morphism (with respect to a given functor, of course). Also, like I said in the comment above, I'm not sure if Cartesian arrows should be directly formulated in terms of functors, or if they should be formulated in terms of `BasedCategory` as in another one of my PRs.

joelriou: Another problem is that "cartesian" is a terminology that is used beyond the case of fibered categories. Then, I think that `Functor` is better.

callesonne: Okay that's a good point. Is it maybe possible to put it in a namespace `Functor.IsCartesian`, so that these short names can be used? I just feel it will be annoying to write `ofIsStronglyCartesian` everywhere (if I make the same changes in the following PR).

callesonne:
I'm not sure why the check isn't succeeding. The build succeeds though.

joelriou:
Thanks!

bors merge

PR #13395, opened 2024-05-30
[Merged by Bors] - feat(Cache): add success message to `lake exe cache get`

### LINKED ISSUE #13276: `lake exe cache get` should print a better "success" message.
Currently it ends with something like
```
Decompressing 4542 file(s)
unpacked in 2097 ms
```
which is true, but for a beginner not obviously meaning "Success!"

(I noticed more than one person asking if this meant success during an install party this morning.)

Rida-Hamadani:
closes #13276 
---

### DIFF
--- a/Cache/IO.lean
+++ b/Cache/IO.lean
@@ -368,7 +368,7 @@ def unpackCache (hashMap : HashMap) (force : Bool) : CacheM Unit := do
     let exitCode ← child.wait
     if exitCode != 0 then throw <| IO.userError s!"leantar failed with error code {exitCode}"
     IO.println s!"Unpacked in {(← IO.monoMsNow) - now} ms"
-    IO.println s!"Completed successfully!"
+    IO.println "Completed successfully!"
   else IO.println "No cache files to decompress"
 
 instance : Ord FilePath where
--- a/Cache/IO.lean
+++ b/Cache/IO.lean
@@ -367,7 +367,8 @@ def unpackCache (hashMap : HashMap) (force : Bool) : CacheM Unit := do
     stdin.putStr <| Lean.Json.compress <| .arr config
     let exitCode ← child.wait
     if exitCode != 0 then throw <| IO.userError s!"leantar failed with error code {exitCode}"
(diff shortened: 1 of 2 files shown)


Cache/IO.lean
```diff
@@ -367,7 +367,8 @@ def unpackCache (hashMap : HashMap) (force : Bool) : CacheM Unit := do
     stdin.putStr <| Lean.Json.compress <| .arr config
     let exitCode ← child.wait
     if exitCode != 0 then throw <| IO.userError s!"leantar failed with error code {exitCode}"
-    IO.println s!"unpacked in {(← IO.monoMsNow) - now} ms"
+    IO.println s!"Unpacked in {(← IO.monoMsNow) - now} ms"
+    IO.println s!"Completed successfully!"
```
grunweg, line 371: Since you're just printing a simple string, you don't need the `s!` macro.
```suggestion
    IO.println "Completed successfully!"
```
In fact, I wonder if it's nicer to combine these two lines, as in `IO.println s!"Unpacked in {(← IO.monoMsNow) - now} ms\nCompleted successfully!"`.

On a higher level: is this the best place to put this message? I would rather expect this line to be wherever the main logic of the `cache get` command is... (so, if some other step errors, you never get a "success" message from this later step, by design).

grunweg (commented):
Thanks for the PR. I have one high-level and one minor comment.

Cache/IO.lean
Rida-Hamadani: Thank you for the comments!
I assumed that errors cause the function to halt, then `unpackCache` won't be called. I'm not sure how the existence of errors is being checked, for instance at `getFiles` in `Cache/Requests.lean` or `main` in `main.lean`. Do  you have any hints?

grunweg: Good question! I don't know myself, so I took a look at the code.
The two previous functions exit the process (with exit code 1) if there's any error, so your assumption seems correct.
I added some documentation to this effect, and also a success message if only decompression was wanted.

semorrison:
bors merge

PR #13396, opened 2024-05-30
[Merged by Bors] - feat(CategoryTheory/Sites): epi/mono factorizations in categories of sheaves

joelriou:
In this PR, we show that if a concrete category `A` is such that morphisms can be factored functorially as surjective maps followed by injective maps, then morphisms of sheaves with values in `A` can be factored functorially as locally surjective maps followed by locally injective maps (and this is also an epi/mono factorization).

Then, it is shown that a morphism of sheaves is epi iff it is locally surjective.

---

- [x] depends on: #13386
- [x] depends on: #13324


dagurtomas:
It's not immediately clear that `isLocallySurjective_iff_epi'` applies to, say, sheaves of abelian groups, or `R`-modules, right? We would need to add an instance `HasFunctorialSurjectiveInjectiveFactorization` for the relavant category? Maybe I'm missing something

joelriou:
> It's not immediately clear that `isLocallySurjective_iff_epi'` applies to, say, sheaves of abelian groups, or `R`-modules, right? We would need to add an instance `HasFunctorialSurjectiveInjectiveFactorization` for the relavant category? Maybe I'm missing something

Yes, we will have to obtain this instance for categories of abelian groups, `R`-modules (which all may fit in the case of abelian categories where surjective = epi and injective = mono). But, it is quite obvious that one way or another, we will obtain this!

dagurtomas:
I can confirm that this together with #13411 works right away to prove that `IsLocallySurjective f ↔ Epi f` in the category of condensed `R`-modules for a ring `R`. See [this branch](https://github.com/leanprover-community/mathlib4/blob/dagur/CondensedEpi/Mathlib/Condensed/Epi.lean)

dagurtomas (approved):

Mathlib/CategoryTheory/Sites/LocallyInjective.lean
```diff
@@ -229,6 +232,21 @@ lemma isLocallyInjective_iff_injective :
...
 
+lemma mono_of_injective
+    (hφ : ∀ (X : Cᵒᵖ), Function.Injective (φ.val.app X)) : Mono φ where
+  right_cancellation := by
+    intro H f₁ f₂ h
+    ext Z x
+    have eq := congr_fun ((forget D).congr_map (congr_app (congr_arg Sheaf.Hom.val h) Z)) x
+    dsimp at eq
+    simp only [FunctorToTypes.map_comp_apply] at eq
+    exact hφ Z eq
```
erdOne, line 243:
```suggestion
    (hφ : ∀ (X : Cᵒᵖ), Function.Injective (φ.val.app X)) : Mono φ :=
  have := fun X ↦ ConcreteCategory.mono_of_injective _ (hφ X)
  (sheafToPresheaf _ _).mono_of_mono_map (NatTrans.mono_of_mono_app φ.1)
```

joelriou: Thanks!

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #6703, opened 2023-08-21
[Merged by Bors] - feat(AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic): define division polynomials

Multramate:
---

- [x] depends on: #10814 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13057, opened 2024-05-20
feat(NumberTheory): characterize elliptic divisibility sequences

alreadydone:
Main results:

+ Every `normEDS` is an elliptic divisibility sequence (EDS). The key proof is `rel₄_of_anti_oddRec_evenRec`, based on my original argument first published on [MathSE](https://math.stackexchange.com/a/4903422/12932)

+ Conversely, every EDS is equal to some `normEDS` (assuming that the first two terms are not zero divisors)

---

- [ ] depends on: #13155
- [x] depends on: #13153
- [x] depends on: #10843

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13399, opened 2024-05-30
[Merged by Bors] - feat(AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic): add maps for division polynomials

Multramate:
---

- [x] depends on: #6703 
- [x] depends on: #13356 


Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -564,21 +570,20 @@ lemma addPolynomial_slope {x₁ x₂ y₁ y₂ : F} (h₁ : W.Equation x₁ y₁
...
   rw [neg_eq_zero, sub_self, mul_zero]
-#align weierstrass_curve.equation_add' WeierstrassCurve.Affine.equation_add'
+#align weierstrass_curve.equation_add' WeierstrassCurve.Affine.equation_negAdd
 
 /-- The addition of two affine points in `W` on a sloped line lies in `W`. -/
 lemma equation_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Equation x₁ y₁) (h₂ : W.Equation x₂ y₂)
     (hxy : x₁ = x₂ → y₁ ≠ W.negY x₂ y₂) :
     W.Equation (W.addX x₁ x₂ <| W.slope x₁ x₂ y₁ y₂) (W.addY x₁ x₂ y₁ <| W.slope x₁ x₂ y₁ y₂) :=
-  equation_neg <| equation_add' h₁ h₂ hxy
+  equation_neg <| equation_negAdd h₁ h₂ hxy
```
alreadydone, line 586: What do you think of making these dot notations? Same for `equation_neg` and `nonsingular` lemmas.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial.lean
```diff
@@ -0,0 +1,504 @@
...
+    rw [C_Ψ₂Sq_eq]
+    ring1
+  · have hm1 : Even (m + 1) := Nat.even_add_one.mpr hm
+    have hm2 : ¬Even (m + 2) := fun h => Nat.even_add_one.mp h hm1
+    have hm3 : Even (m + 3) := Nat.even_add_one.mpr hm2
+    have hm4 : ¬Even (m + 4) := fun h => Nat.even_add_one.mp h hm3
+    rw [if_neg hm, if_neg hm, if_neg hm4, if_neg hm2, if_pos hm1, if_pos hm3, if_neg hm]
+    C_simp
+    rw [C_Ψ₂Sq_eq]
+    ring1
```
alreadydone, line 316: You can probably golf the same way as I do [here](https://github.com/leanprover-community/mathlib4/pull/13057/files#diff-02d4668be95cd41b094c2ec5859acf1cf92c74833b9181a533bcce3a23c00961R765-R770).

Multramate: This belongs to #6703 so I will change it there.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: Good idea, but since we also need to make the changes in `Projective` and `Jacobian` I'll do them in a separate PR.

alreadydone (commented):
The diff is large (haven't merged #13356 yet, but probably we want to merge after #6703 is in) so I'm not sure I've seen all the diffs, but the parts in EllipticDivisibilitySequence.lean and DivisionPolynomial.lean looks good.

alreadydone (approved):
LGTM, just one question

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -290,11 +290,11 @@ noncomputable def normEDSRec {P : ℕ → Sort u}
...
-  | zero => exact map_zero f
-  | one => exact map_one f
-  | two => exact map_one f
-  | three => rfl
-  | four => rfl
+  | zero => rw [preNormEDS'_zero, map_zero, preNormEDS'_zero]
+  | one => rw [preNormEDS'_one, map_one, preNormEDS'_one]
+  | two => rw [preNormEDS'_two, map_one, preNormEDS'_two]
+  | three => rw [preNormEDS'_three, preNormEDS'_three]
+  | four => rw [preNormEDS'_four, preNormEDS'_four]
```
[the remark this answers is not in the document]
Multramate: It's not... but unfortunately we now have to do this due to #12969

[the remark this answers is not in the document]
alreadydone: Oh interesting, thanks for letting me know. Is that change for performance reasons?

maintainer merge

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -42,10 +42,10 @@ Some examples of EDSs include
 ## Implementation notes
 
 The normalised EDS `normEDS b c d n` is defined in terms of the auxiliary sequence
-`preNormEDS (b ^ 4) c d n`, which are equal when `n` is odd, and which differ by a factor of `b`
-when `n` is even. This coincides with the definition in the references since both agree for
-`normEDS b c d 2` and for `normEDS b c d 4`, and the correct factors of `b` are removed in
-`normEDS b c d (2 * (m + 2) + 1)` and in `normEDS b c d (2 * (m + 3))`.
+`preNormEDS (b ^ 4) c d n`, which are equal when `n` is odd, and which differ by a factor of an
+extraneous parameter `b` when `n` is even. This coincides with the definition in the references
```
alreadydone, line 46: I think the "extraneous parameter" here is `b^4` rather than `b`; I suggest you keep the original "differ by a factor of `b`".

I'm not sure "extraneous" has the intended meaning; how about just "parameter" without an adjective?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -38,16 +38,16 @@ defined as its image under the associated universal morphism $\mathbb{Z}[A_i][X,
 
 Now, in the coordinate ring $R[W]$, note that $\psi_2^2$ is congruent to the polynomial
 $\Psi_2^{[2]} := 4X^3 + b_2X^2 + 2b_4X + b_6 \in R[X]$. As such, in $R[W]$, the recurrences
-associated to a normalised elliptic divisibility sequence show that $\psi_n / \psi_2$ are congruent
-to certain polynomials in $R[X]$. In particular, define $\tilde{\Psi}_n \in R[X]$ as the auxiliary
-sequence for a normalised elliptic divisibility sequence with initial values
+associated to a normalised EDS show that $\psi_n / \psi_2$ are congruent to certain polynomials in
+$R[X]$. In particular, define $\tilde{\Psi}_n \in R[X]$ as the auxiliary sequence for a normalised
+EDS with extraneous parameter $\Psi_2^{[2]}$ and initial values
```
alreadydone, line 44: The parameter equals $(\Psi_2^{[2]})^2$ here, right?

alreadydone (changes requested):

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
alreadydone: Even "extra" is better IMO, though you probably then want to say "initial values ... and an extra parameter ...".

Multramate: Sorry, sometimes my English short-circuits.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -42,10 +42,10 @@ Some examples of EDSs include
 ## Implementation notes
 
 The normalised EDS `normEDS b c d n` is defined in terms of the auxiliary sequence
-`preNormEDS (b ^ 4) c d n`, which are equal when `n` is odd, and which differ by a factor of `b`
-when `n` is even. This coincides with the definition in the references since both agree for
-`normEDS b c d 2` and for `normEDS b c d 4`, and the correct factors of `b` are removed in
-`normEDS b c d (2 * (m + 2) + 1)` and in `normEDS b c d (2 * (m + 3))`.
+`preNormEDS (b ^ 4) c d n`, which are equal when `n` is odd, and which differ by a factor of an
+extra parameter `b` when `n` is even. This coincides with the definition in the references since
```
alreadydone, line 46: I was suggesting this, since "extra parameter" is associated to a preNormEDS, and it's `b ^ 4` in this case.
```suggestion
`preNormEDS (b ^ 4) c d n`, which are equal when `n` is odd, and which differ by a factor of 
`b` when `n` is even. This coincides with the definition in the references since
```

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -102,8 +102,8 @@ lemma IsDivSequence.smul (h : IsDivSequence W) (x : R) : IsDivSequence (x • W)
 lemma IsEllDivSequence.smul (h : IsEllDivSequence W) (x : R) : IsEllDivSequence (x • W) :=
   ⟨h.left.smul x, h.right.smul x⟩
 
-/-- The auxiliary sequence for a normalised EDS `W : ℕ → R`,
-with initial values `W(0) = 0`, `W(1) = 1`, `W(2) = 1`, `W(3) = c`, and `W(4) = d`. -/
+/-- The auxiliary sequence for a normalised EDS `W : ℕ → R`, with extra parameter `b` and initial
+values `W(0) = 0`, `W(1) = 1`, `W(2) = 1`, `W(3) = c`, and `W(4) = d`. -/
```
alreadydone, line 106: I was suggesting this, since it's probably better to first list "main parameters" (the initial values), and then the extra one.
```suggestion
/-- The auxiliary sequence for a normalised EDS `W : ℕ → R`, with initial values 
`W(0) = 0`, `W(1) = 1`, `W(2) = 1`, `W(3) = c`, and `W(4) = d` and extra parameter `b`. -/
```

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -159,8 +159,8 @@ lemma preNormEDS'_even (m : ℕ) : preNormEDS' b c d (2 * (m + 3)) =
   simp only [Nat.mul_add_div two_pos]
   rfl
 
-/-- The auxiliary sequence for a normalised EDS `W : ℤ → R`,
-with initial values `W(0) = 0`, `W(1) = 1`, `W(2) = 1`, `W(3) = c`, and `W(4) = d`.
+/-- The auxiliary sequence for a normalised EDS `W : ℤ → R`, with extra parameter `b` and initial
+values `W(0) = 0`, `W(1) = 1`, `W(2) = 1`, `W(3) = c`, and `W(4) = d`.
```
alreadydone, line 163: ditto here

alreadydone (approved):
Thanks 🎉 
This is now good for merging again.

jcommelin:
Does it make sense to put this in a new file? Because the current import adds 300 new transitive imports. Or do you think that everything importing `.Basic` will also want access to the new results from this PR anyways?

Multramate:
> Does it make sense to put this in a new file? Because the current import adds 300 new transitive imports. Or do you think that everything importing `.Basic` will also want access to the new results from this PR anyways?

Very understandable. I do think that users of \Psi and \Phi should have access to their equivalence with \psi and \phi in the coordinate ring, because that's the original motivation for having separate definitions. This only requires the definition of the coordinate ring and some basic properties, hence only requires importing `AdjoinRoot` and `Norm`. Unfortunately the file defining the coordinate ring is interlaced with the proof of the group law that imports `ClassGroup`, which presumably is the majority of the 300+ transitive imports. How about this: let's merge this PR, and I'll open another PR immediately that splits off the stuff on the coordinate ring into another file, and this file can just import that new file.

alreadydone:
The central result about division polynomials is the [formula](https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Defining.20n-division.20polynomials/near/440803664) for zsmul on the group of rational points in terms of division polynomials, and the group structure already require `CoordinateRing`, `ClassGroup` and `Norm`,  so I don't think it makes much sense to avoid them.

Multramate:
Yes, but `ZSMul.lean` or `Group.lean` will be a different file from `Basic.lean`. Maybe there are uses of division polynomials without needing the fact that the points form a group, e.g. if you just need an explicit duplication/triplication formula?

I think it's probably good to split off a file just for the coordinate ring anyway - this can include all the API for the universal elliptic curve, the evaluation and specialisation maps, and whatever else that are universal constructions.

Now that I think of it, we don't actually need the two definitions to agree until we actually compute the structure of the n-torsion. Maybe I'll leave the lemmas here for now, and the move it out eventually once I get to that.

alreadydone:
It seems our consensus is to merge this PR first, so let's do that and worry about splitting later? If you put your proposed splitting in some branch I can check whether it's reasonable ...

alreadydone:
In fact, I think it's pretty rare that people would want to talk about elliptic curves without knowing the rational points form a group. Do you have such examples in mind?

YaelDillies:
I tend to agree that the split can be done later, if at all

Multramate:
> In fact, I think it's pretty rare that people would want to talk about elliptic curves without knowing the rational points form a group. Do you have such examples in mind?

Not specifically the theory of division polynomials, but my primary example would be moduli problems (e.g. the moduli stack of genus one curves, or just modular curves X/X_0/X_1), where the group law is not relevant? Obviously this is very far away and proving any useful theorems about moduli spaces would require some form of Riemann-Roch, in which case the group law is pretty much automatic, but maybe there's a nice description of the Weierstrass coefficients in terms of the moduli space.

Another argument for avoiding the import could be if the user prefers to write a different group law (e.g. the one induced by complex uniformisation), which might have properties easier to work with, but that's probably a weak argument.

alreadydone:
Well, the modular curves seem pretty closely connected to the group structure:
image
You might be able to define modular curves through their own equations but as soon as you connect them to elliptic curves you'd be talking about torsion points.

Vierkantor (approved):
The code looks good to me, and the consensus seems to be that the imports are okay or can be split later on. So let's get this merged.

bors r+

PR #13400, opened 2024-05-30
[Merged by Bors] - feat: add `Polynomial.coeff_list_sum_map`

BoltonBailey:
Adds lemmas about transposing a `coeff` with a sum of a list, analogous to `finset_sum_coeff` (although in this case I will not mark `simp` because the expression on the RHS is visually more complicated).

---


Ruben-VandeVelde (approved):
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13401, opened 2024-05-30
feat(RingTheory/Congruence): interpret `RingCon` as two sided ideals

jjaassoonn:
In this PR, we set up an instance `SetLike (RingCon R) R` for any `NonUnitalNonAssocRing R`. This enables us to interpret `RingCon R` as two-sided-ideals.

Co-authored-by: Eric Wieser <@eric-wieser>

---

- [x] depends on: #13610
- [x] depends on: #13730
- [x] depends on: #13902 
- [ ] depends on: #14453


Mathlib/RingTheory/Congruence.lean
```diff
@@ -111,6 +112,24 @@ protected theorem mul {w x y z} : c w x → c y z → c (w * y) (x * z) :=
...
 
+lemma sum {ι S : Type*} [AddCommMonoid S] [Mul S] (t : RingCon S) (s : Finset ι)
+    {f g : ι → S} (h : ∀ i ∈ s, t (f i) (g i)) :
+    t (s.sum f) (s.sum g) := by
+  classical
+  induction s using Finset.induction_on with
+  | empty => simpa using t.refl 0
+  | @insert i s hi ih =>
+    rw [Finset.sum_insert hi, Finset.sum_insert hi]
+    exact t.add (h _ (by simp)) <| ih fun i hi ↦ h _ (by aesop)
```
eric-wieser, line 123: Can you prove this for `Con` and `prod`, and then get the result for free?

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,125 @@
...
+-/
+
+section ideal
+
+namespace RingCon
+
+variable {R : Type*} [Ring R]
+
+instance : SetLike (RingCon R) R where
+  coe t := {r | t r 0}
```
eric-wieser, line 25: Can you add a (non-`simp`) lemma `mem_iff` that expands membership to `t r 0`?

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,125 @@
...
+-/
+
+import Mathlib.RingTheory.Congruence
+import Mathlib.Algebra.Module.LinearMap.Basic
+import Mathlib.Tactic.Abel
+
+/-!
+# Two Sided Ideals
+
+In this file, for any `Ring R`, we reinterpret `I : RingCon R` as a two sided ideal of a ring.
```
eric-wieser, line 14: A short discussion about the relation with subbimodules (which we have elsewhere, but are more heavyweight) would be good to have here.

It's perhaps also worth noting something about how `RingCon S` is more general in some sense as it works for semirings, but that in practice no one cares.

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,125 @@
...
+  smul n x := ⟨n • x.1, n.rec (fun a ↦ a.rec (by simpa using I.zero_mem)
+    (fun a ha ↦ by
+      simp only [Nat.succ_eq_add_one, Int.ofNat_eq_coe, Nat.cast_add, Nat.cast_one,
+        Int.cast_add, Int.cast_natCast, Int.cast_one, add_smul, one_smul]
+      exact I.add_mem ha x.2))
+    (fun a ↦ a.rec (by simpa using I.neg_mem x.2)
+    (fun a ha ↦ by
+      simp only [Nat.succ_eq_add_one, negSucc_zsmul, nsmul_eq_mul, Nat.cast_add, Nat.cast_one,
+        add_mul, one_mul, neg_add_rev] at ha ⊢
+      exact I.add_mem (I.neg_mem x.2) ha))⟩
```
eric-wieser, line 89: Can't you get here much faster by converting the smul into a cast/mul, and then using `mul_mem_left`? Same for `nsmul`.

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,125 @@
...
+  smul_zero x := by
+    ext; show x * 0 = 0; simp
+  smul_add x y z := by
+    ext; show x • (y.1 + z.1) = x • y.1 + x • z.1; simp [mul_add]
+  add_smul x y z := by
+    ext; show (x + y) • z.1 = x • z.1 + y • z.1; simp [add_mul]
+  zero_smul x := by
+    ext; show 0 * x.1 = 0; simp
+
+instance : Module Rᵐᵒᵖ I where
```
eric-wieser, line 109: You should add the `SMulCommClass R Rᵐᵒᵖ I` instance that makes this truly a bimodule.

eric-wieser (approved):
This is a pretty unusual use of `SetLike`, but I don't forsee it causing much trouble.

It might be good to get the connection to `Ideal` in this version; in particular that `(I.con : Set R) = I` or whatever the map from ideals to congruence relations is actually called.

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: hmm. If I understand it correctly, I need subtraction when proving that `RingCon S` is `SetLike`, so it means that `RingCon S` as two-sided-Ideals won't work for semirings I think.

jjaassoonn:
Thanks for the quick response!

eric-wieser:
Defining `RingCon.toIdeal` might allow you to further golf some of the proofs / instances.

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: done in here [cfe1bbd](https://github.com/leanprover-community/mathlib4/pull/13401/commits/cfe1bbdc7aca2906bf8022bdca025714576617ca)

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: Added a short description

Mathlib/GroupTheory/Congruence.lean
```diff
@@ -1257,6 +1258,18 @@ protected theorem div : ∀ {w x y z}, c w x → c y z → c (w / y) (x / z) :=
 #align con.div Con.div
 #align add_con.sub AddCon.sub
 
+/-- Multiplicative congruence relations preserve finite product. -/
+@[to_additive "Additive congruence relations preserve finite sum."]
+protected theorem finset_prod {ι M : Type*} [CommMonoid M] (c : Con M) (s : Finset ι)
+    {f g : ι → M} (h : ∀ i ∈ s, c (f i) (g i)) :
+    c (s.prod f) (s.prod g) := by
+  classical
+  induction s using Finset.induction_on with
```
eric-wieser, line 1267:
```suggestion
  induction s using Finset.cons_induction_on with
```
with a small tweak to the proof below.

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: Since I just suddenly decided that I care about `NonUnitalNonAssocRing R`, define `n • x` as `(n : R) * x` does not make sense anymore xd since we don't have a one.

Mathlib/GroupTheory/Congruence.lean
jjaassoonn: Done in [99fc44a](https://github.com/leanprover-community/mathlib4/pull/13401/commits/99fc44a9a0d81f339a69757d5edce5cb673fb936).

Why is `Finset.cons_induction_on` preferred over `Finset.induction_on` btw?

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: done in [3b5e4cf](https://github.com/leanprover-community/mathlib4/pull/13401/commits/3b5e4cf0b9c7ee12e780c79f0523509245eaaaa5)

Mathlib/RingTheory/TwoSidedIdeal.lean
eric-wieser: I found a golf

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,137 @@
...
+theorem nsmul (n : ℕ) {x y : R} (hx : I x y) : I (n • x) (n • y) := I.toAddCon.nsmul n hx
+
+instance : SMul ℕ I where
+  smul n x := ⟨n • x.1, by simpa using I.nsmul _ x.2⟩
+
+instance : Neg I where neg x := ⟨-x.1, I.neg_mem x.2⟩
+
+instance : Sub I where sub x y := ⟨x.1 - y.1, I.sub_mem x.2 y.2⟩
+
+theorem zsmul (z : ℤ) {x y : R} (hx : I x y) : I (z • x) (z • y) := I.toAddCon.zsmul z hx
```
eric-wieser, line 86: I'm afraid I was lazy; can you move this upstream to the `Congruence.lean` file, and the `nsmul` one too?

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,137 @@
...
+
+instance : SMul ℤ I where
+  smul n x := ⟨n • x.1, by simpa using I.zsmul n x.2⟩
```
eric-wieser, line 89: The proof here should perhaps exist as `zsmul_mem` alongside all the others.

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,125 @@
...
+lemma zero_mem : 0 ∈ I := I.refl 0
+
+lemma add_mem {x y} (hx : x ∈ I) (hy : y ∈ I) : x + y ∈ I := by
+  simpa using I.add hx hy
+
+lemma neg_mem {x} (hx : x ∈ I) : -x ∈ I := by
+  simpa using I.neg hx
+
+lemma sub_mem {x y} (hx : x ∈ I) (hy : y ∈ I) : x - y ∈ I := by
+  simpa using I.sub hx hy
```
eric-wieser, line 65: Can you create an `AddSubgroupClass` instance?

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: done in [77d72f8](https://github.com/leanprover-community/mathlib4/pull/13401/commits/77d72f8ba31ad70b324f2dbcb80db13cd6845b02). Thanks for the golf!!

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: Nice!

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: done in [dad9307](https://github.com/leanprover-community/mathlib4/pull/13401/commits/dad9307184fa06a43ac198127ee6d469ed107516)

jjaassoonn:
> This is a pretty unusual use of `SetLike`, but I don't forsee it causing much trouble.
> 
> It might be good to get the connection to `Ideal` in this version; in particular that `(I.con : Set R) = I` or whatever the map from ideals to congruence relations is actually called.

I have added `toIdeal` and `fromIdeal` and proved that for any `I : RingCon R`, `from I.toIdeal = I` and for any `J : Ideal R`, `J \le (fromIdeal J).toIdeal`

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+  intro h I hI
+  refine sInf_le (α := RingCon R) ?_ h
+  intro x y hxy
+  specialize hI hxy
+  rwa [SetLike.mem_coe, ← rel_iff] at hI
+
+/--
+Every left ideal generates a two sided ideal.
+-/
+def fromIdeal (I : Ideal R) : RingCon R := span I
```
eric-wieser, line 210: This exists already, it's `Ideal.Quotient.ringCon`

jjaassoonn: `Ideal.Quotient.ringCon` needs the commutativity assumption. Not sure if it really need it or the commutativity assumption was added by accident

jjaassoonn:
```lean
protected def ringCon (I : Ideal R) : RingCon R :=
  { QuotientAddGroup.con I.toAddSubgroup with
    mul' := fun {a₁ b₁ a₂ b₂} h₁ h₂ => by
      rw [Submodule.quotientRel_r_def] at h₁ h₂ ⊢
      have F := I.add_mem (I.mul_mem_left a₂ h₁) (I.mul_mem_right b₁ h₂)
      have : a₁ * a₂ - b₁ * b₂ = a₂ * (a₁ - b₁) + (a₂ - b₂) * b₁ := by
        rw [mul_sub, sub_mul, sub_add_sub_cancel, mul_comm, mul_comm b₁]
      rwa [← this] at F }
```
seems the proof actually used `mul_comm`

jjaassoonn:
If people like the idea of this PR but find it too big to review, I will split it up.

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: If I'm not mistaken, `Ideal.Quoteint.ringCon` will ensure that `I : Ideal R` when seen as a two-sided-ideal is exactly the same as `I` (because `R` is commutative). But `fromIdeal I` might be bigger than `I` because we are multiplying things on the right side of the left idea I?

Mathlib/RingTheory/Congruence.lean
```diff
@@ -111,6 +112,26 @@ protected theorem mul {w x y z} : c w x → c y z → c (w * y) (x * z) :=
...
+protected theorem neg {S : Type*} [AddGroup S] [Mul S] (t : RingCon S)
+    {a b} (h : t a b) : t (-a) (-b) :=
+  t.toAddCon.neg h
+
+protected theorem nsmul {S : Type*} [AddGroup S] [Mul S] (t : RingCon S)
+    (m : ℕ) {x y : S} (hx : t x y) : t (m • x) (m • y) := t.toAddCon.nsmul m hx
+
+protected theorem zsmul {S : Type*} [AddGroup S] [Mul S] (t : RingCon S)
+    (z : ℤ) {x y : S} (hx : t x y) : t (z • x) (z • y) := t.toAddCon.zsmul z hx
+
```
alreadydone, line 133:
```suggestion
```

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+variable {R : Type*} [NonUnitalNonAssocRing R] (I : RingCon R)
+
+instance : SetLike (RingCon R) R where
+  coe t := {r | t r 0}
+  coe_injective' t₁ t₂ (h : {x | _} = {x | _}) := by
+    refine RingCon.ext fun a b ↦ ⟨fun H ↦ ?_, fun H ↦ ?_⟩
+    · have H' : a - b ∈ {x | t₁ x 0} := sub_self b ▸  t₁.sub H (t₁.refl b)
+      rw [h] at H'
+      convert t₂.add H' (t₂.refl b) using 1 <;> abel
+    · have H' : a - b ∈ {x | t₂ x 0} := sub_self b ▸  t₂.sub H (t₂.refl b)
```
alreadydone, line 38:
```suggestion
    · have H' : a - b ∈ {x | t₁ x 0} := sub_self b ▸ t₁.sub H (t₁.refl b)
      rw [h] at H'
      convert t₂.add H' (t₂.refl b) using 1 <;> abel
    · have H' : a - b ∈ {x | t₂ x 0} := sub_self b ▸ t₂.sub H (t₂.refl b)
```

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+      rw [← h] at H'
+      convert t₁.add H' (t₁.refl b) using 1 <;> abel
+
+lemma mem_iff (x : R) : x ∈ I ↔ I x 0 := Iff.rfl
+
+lemma rel_iff (x y : R) : I x y ↔ (x - y) ∈ I := by
```
alreadydone, line 44:
```suggestion
lemma rel_iff (x y : R) : I x y ↔ x - y ∈ I := by
```

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+lemma nsmul_mem {x} (n : ℕ) (hx : x ∈ I) : n • x ∈ I := by
+  simpa using I.nsmul _ hx
+
+lemma zsmul_mem {x} (n : ℤ) (hx : x ∈ I) : n • x ∈ I := by
+  simpa using I.zsmul _ hx
+
+instance : AddSubgroupClass (RingCon R) R where
+  zero_mem := zero_mem
+  add_mem := @add_mem _ _
+  neg_mem := @neg_mem _ _
```
alreadydone, line 89: Can you add two [SMulMemClass](https://leanprover-community.github.io/mathlib4_docs/Mathlib/GroupTheory/GroupAction/SubMulAction.html#SMulMemClass) instances with R and R^mop respectively?

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+
+lemma mem_iff (x : R) : x ∈ I ↔ I x 0 := Iff.rfl
+
+lemma rel_iff (x y : R) : I x y ↔ (x - y) ∈ I := by
+  rw [mem_iff]
+  constructor
+  · intro h; convert I.sub h (I.refl y); abel
+  · intro h; convert I.add h (I.refl y) <;> abel
+
+lemma le_iff (I J : RingCon R) : I ≤ J ↔ (I : Set R) ⊆ (J : Set R) := by
```
alreadydone, line 50: This says the coercion to Set R is an [OrderEmbedding](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Hom/Basic.html#OrderEmbedding); not sure how useful it is though

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,227 @@
...
+  add_mem' := I.add_mem
+  zero_mem' := I.zero_mem
+  smul_mem' := I.mul_mem_left
+
+lemma mem_toIdeal {x} : x ∈ I.toIdeal ↔ x ∈ I := Iff.rfl
+
+/--
+Every two-sided ideal is also a right ideal.
+-/
+def toIdealMop : Ideal Rᵐᵒᵖ := LinearMap.range I.subtypeMop
```
erdOne, line 175: This gives bad defeqs. Maybe first show `RingCon Rᵐᵒᵖ = RinCon R` and then use `toIdeal`.

jjaassoonn: I don’t think I understand. Isn’t type equality even more evil.

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: done in here [ee508d5](https://github.com/leanprover-community/mathlib4/pull/13401/commits/ee508d56d6c46d16bf283fc368962d9ba7dddc34)

Mathlib/RingTheory/TwoSidedIdeal.lean
erdOne: Sorry I meant equivs

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: added in [8629ec0](https://github.com/leanprover-community/mathlib4/pull/13401/commits/8629ec0d2f63d7c8ae524ec968b64f2de63ea10c)

Mathlib/RingTheory/TwoSidedIdeal.lean
jjaassoonn: I have made a change, but I still can't get
```
lemma mem_toIdealMop {x} : x ∈ I.toIdealMop ↔ x.unop ∈ I := by
```
to be true by Iff.rfl

jjaassoonn:
There are still a lot of missing lemmas like
```lean

lemma mem_inf {I J : RingCon R} {x} : x ∈ I ⊓ J ↔ x ∈ I ∧ x ∈ J := by simp [mem_iff]

lemma mem_sup_left {I J : RingCon R} {x} : x ∈ I → x ∈ I ⊔ J := @le_sup_left _ _ I J x 0

lemma mem_sup_right {I J : RingCon R} {x} : x ∈ J → x ∈ I ⊔ J := @le_sup_right _ _ I J x 0
```
But I guess I will wait for another PR

Mathlib/RingTheory/TwoSidedIdeal.lean
```diff
@@ -0,0 +1,294 @@
...
+  rw [mem_iff]
+  constructor
+  · intro h; convert I.sub h (I.refl y); abel
+  · intro h; convert I.add h (I.refl y) <;> abel
+
+/--
+Any two-sided-ideal in `A` corresponds to a two-sided-ideal in `Aᵒᵖ`.
+-/
+@[simps]
+def toMop (rel : RingCon R) : (RingCon Rᵐᵒᵖ) :=
```
erdOne, line 68: Following other structures, I'd call this `RingCon.op` and the next construction `RingCon.unop`. They probably below to `Mathlib.RingTheory.Congruence` too.

jjaassoonn: done in [d107a35](https://github.com/leanprover-community/mathlib4/pull/13401/commits/d107a356f2cc703f2eb856fe937bb99d4540285f)

Mathlib/GroupTheory/Congruence.lean
```diff
@@ -6,6 +6,8 @@ Authors: Amelia Livingston
 import Mathlib.Algebra.Group.Equiv.Basic
 import Mathlib.Algebra.Group.Submonoid.Operations
 import Mathlib.Data.Setoid.Basic
+import Mathlib.Algebra.BigOperators.Basic
+import Mathlib.Algebra.Opposites
```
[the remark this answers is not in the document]
kbuzzard: How about a new PR where Mathlib/GroupTheory/Congruence.lean (which is getting really big!) becomes Mathlib/GroupTheory/Congruence/Basic.lean and then there are two new files Mathlib/GroupTheory/Congruence/Finset.lean and Mathlib/GroupTheory/Congruence/Opposites.lean which include the two new lemmas? That could be reviewed independent of this PR (which will make this PR shorter!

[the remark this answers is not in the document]
YaelDillies: This new import is a bit suspicious in that it brings in finiteness. Also note that `Algebra.BigOperators.Basic` is now `Algebra.BigOperators.Group.Finset` on master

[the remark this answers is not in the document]
jjaassoonn: I have splitted the file. Thanks for the suggestions!

Mathlib/GroupTheory/Congruence/BigOperators.lean
```diff
@@ -6,24 +6,42 @@ Authors: Amelia Livingston
 
 import Mathlib.GroupTheory.Congruence.Basic
 import Mathlib.Algebra.BigOperators.Group.Finset
+import Mathlib.Algebra.BigOperators.Group.Multiset
+import Mathlib.Algebra.BigOperators.Group.List
```
YaelDillies, line 10: The first imports the other two
```suggestion
import Mathlib.Algebra.BigOperators.Group.Finset
```

Mathlib/GroupTheory/Congruence/BigOperators.lean
```diff
@@ -6,24 +6,42 @@ Authors: Amelia Livingston
...
 
 /-!
 # Interactions between `∑, ∏` and `(Add)Con`
 
 -/
 
+open BigOperators
+
```
YaelDillies, line 18: This is now a no-op
```suggestion
```

j-loreaux:
Is there any chance we can have an `abbrev TwoSidedIdeal := RingCon`? I really hate the current spelling (although I love the implementation! Thanks!)

Mathlib/RingTheory/TwoSidedIdeal.lean
j-loreaux: Can we also have a bundling as an order isomorphism when `R` is commutative, please?

jjaassoonn:
Superseded by https://github.com/leanprover-community/mathlib4/pull/14453

PR #13402, opened 2024-05-30
[Merged by Bors] - chore(test/linarith): remove stray variables from guard_msgs

eric-wieser:
Some of the `#guard_msgs` testing for errors in this file capture a pile of random unrelated variables. This scopes these variables to `section`s to avoid this behavior.

I also uncomment a test that doesn't seem to be problematic any more, and remove `pp.mvars` which wasn't actually having any useful effect.

---

### DIFF
--- a/test/linarith.lean
+++ b/test/linarith.lean
@@ -626,9 +626,8 @@ example (a b c d e : ℚ)
     e = 3 := by
   linarith
 
-set_option autoImplicit true in
 /-- https://github.com/leanprover-community/mathlib4/issues/2717 -/
-example :
+example {x1 x2 x3 x4 x5 x6 x7 x8 : ℚ} :
     (3 * x4 - x3 - x2 - x1 : ℚ) < 0 →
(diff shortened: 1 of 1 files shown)


test/linarith.lean
```diff
@@ -634,11 +646,6 @@ example :
 /--
 error: linarith failed to find a contradiction
 case h1.h
-E : Type _
-inst✝¹ : AddGroup E
-R : Type _
-inst✝ : Ring R
-abs : R → ℚ
```
eric-wieser, line 641: Removing these was the main motivation

grunweg (commented):
Looks good to me, thanks for doing this.

The remaining `autoImplicit` calls are not necessarily pretty, but removing these in a follow-up is fine with me.
maintainer merge

grunweg:
Actually: I just pushed a commit removing the added `autoImplicit true`; reordering variables since I saw no point in the old order.

fpvandoorn:
bors merge

PR #13403, opened 2024-05-31
[Merged by Bors] - fix: `cc` tactic can't unify def-eq instances in `OfNat.ofNat`

### LINKED ISSUE #13362: `cc` tactic can't unify def-eq instances in `OfNat.ofNat`
We can't recover `cc` here in `bernoulliPowerSeries_mul_exp_sub_one` on `Mathlib.NumberTheory.Bernoulli`:
```
A : Type u_1
inst✝¹ : CommRing A
inst✝ : Algebra ℚ A
n : ℕ
hfact : ∀ (m : ℕ), ↑m ! ≠ 0
hite2 : (if n.succ = 0 then 1 else 0) = 0
x : ℕ × ℕ
h : x.1 + x.2 = n.succ
⊢ (↑x.2! * (↑x.2 + 1) = (↑x.2 + 1) * ↑x.2! ∨ (x.1 + x.2)! = 0) ∨ bernoulli x.1 = 0
```
As a porter of `cc`, I'm trying to fix this regression.

Komyyy:
This issue is consisted of one problem and one bug.

**Problem: `cc` treats an `OfNat.ofNat` term as an atomic value in the default config**
In the default `CCConfig`, the `values` field is set to `true` so `cc` treats an `OfNat.ofNat` term as an atomic and it doesn't unify `OfNat` instances.
This problem can be solved simply by setting `values := false` by default.

**Bug: `cc` tries to disprove `@OfNat α (nat_lit n) i₁ ≠ @OfNat α (nat_lit n) i₂` where `α` is `ℕ` or `ℤ` and `i₁` & `i₂` is not alpha-equivalent but def_eq**
`cc` tactic has a mechanism to disprove equality between two different values of `ℕ`, `ℤ`, `Char` & `String`, in the function `propagateEqUp`. This mechanism is trigered when the root representations of two values are not alpha-equivalent, so this example fails.
```lean
local instance instOfNatNat' (n : ℕ) : OfNat ℕ n where
  ofNat := n

example : @OfNat.ofNat ℕ (nat_lit 0) (instOfNatNat _) =
    @OfNat.ofNat ℕ (nat_lit 0) (instOfNatNat' _) := by
  cc
/-
application type mismatch
  (bne_iff_ne 0 0).1 (Eq.refl true)
argument has type
  true = true
but function has type
  (0 != 0) = true → 0 ≠ 0
-/
```
So we strengthen the condition in `propagateEqUp`. For the detail, refer to docs in this function.

---

Resolves #13362

### DIFF
--- a/Mathlib/Tactic/CC/Addition.lean
+++ b/Mathlib/Tactic/CC/Addition.lean
@@ -1449,7 +1449,25 @@ partial def propagateEqUp (e : Expr) : CCM Unit := do
   let rb ← getRoot b
   if ra != rb then
     let mut raNeRb : Option Expr := none
-    if ← isInterpretedValue ra <&&> isInterpretedValue rb then
+    /-
+    We disprove inequality for interpreted values here.
+    The possible types of interpreted values are in `{String, Char, Int, Nat}`.
+    1- `String`
+      `ra` & `rb` are string literals, so if `ra != rb`, `ra.int?.isNone` is `true` and we can
+      prove `$ra ≠ $rb`.
+    2- `Char`
+      `ra` & `rb` are the form of `Char.ofNat (nat_lit n)`, so if `ra != rb`, `ra.int?.isNone` is
+      `true` and we can prove `$ra ≠ $rb` (assuming that `n` is not pathological value, i.e.
+      `n.isValidChar`).
+    3- `Int`, `Nat`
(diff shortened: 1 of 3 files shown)


Vierkantor (approved):
Looks good to me, thanks for the detailed explanations!

bors r+

PR #13404, opened 2024-05-31
[Merged by Bors] - chore: restore `cc` tactic

Komyyy:
The `cc` tactic is a high-cost tactic, so I restored `cc` tactics only if a proof can be reduced significantly.
I made sure that `cc` tactic works on in the all changed proofs except when the other specification changes in Lean 4 changes a goal.

---

- [x] depends on: #13403


grunweg:
!bench

Mathlib/Tactic/CC/Datatypes.lean
```diff
@@ -130,7 +130,7 @@ structure CCConfig where
   /-- If `true`, then use excluded middle -/
   em : Bool := true
   /-- If `true`, we treat values as atomic symbols -/
-  values : Bool := true
+  values : Bool := false
```
joneugster, line 133: Can you explain here quickly why you decided to change this?

joneugster (approved):
Looks good to me.

Mathlib/Tactic/CC/Datatypes.lean
Komyyy: This is a change in the dependent PR; I merge to master now...

Komyyy: @joneugster Merged!

Ruben-VandeVelde (approved):
This seems reasonable

maintainer merge

semorrison:
bors merge

PR #13407, opened 2024-05-31
[Merged by Bors] - chore: Added namespace Mathlib around Vector defs

Shreyas4991:
This PR puts `Mathlib.Data.Vector`'s `Vector` type into the namespace `Mathlib`. This will help ensure that users of `Mathlib` and `Batteries` can use this Vector type and Batteries' upcoming Array based Vector type without name conflicts.

Tasks:
- [x] Put all Vector definitions, currently all of which are inside Mathlib/Data/Vector into the namespace `Mathlib`.
~~- [x] Add deprecation aliases for all `def`s, `theorem`s and `lemmas` thus namespaced.~~ NOTE : This is  not possible because `alias` simply reintroduces naming conflicts. See the linked zulip discussion
- [x] Fix breakages within Mathlib

Discussion in [Zulip topic](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Mathlib.20Vector.20namespace/near/441643042)

---


Shreyas4991:
Note at this commit : When opening the namespace Mathlib inside the namespace Vector in Topology.List, a number of Vector functions were inferred as List functions and this created errors. I correctly namespaced them

PR #13408, opened 2024-05-31
[Merged by Bors] - chore(CategoryTheory/Sites): remove unused variable

dagurtomas:
Removes an old assumption `HasFiniteLimits` on an abelian category (this is automatic and doesn't need to be stated)

---


erdOne:
Thanks!
maintainer merge

PR #13410, opened 2024-05-31
[Merged by Bors] - feat(FiberedCategory/Cartesian): define strongly cartesian morphisms

callesonne:
We define strongly cartesian morphisms with respect to a functor `p : 𝒳 ⥤ 𝒮`, as in [02XK](https://stacks.math.columbia.edu/tag/02XK). We provide some basic API, and prove some basic properties.

---

- [x] depends on: #13393
- [x] depends on: #12978


Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
```diff
@@ -115,4 +143,200 @@ instance of_comp_iso {b' : 𝒳} (φ' : b ≅ b') [IsHomLift p (𝟙 S) φ'.hom]
...
+```
+such that `φ` is a cartesian, and a morphism `φ' : a' ⟶ b`. Then `map` is the map `a' ⟶ a`
+lying over `g` obtained from the universal property of `φ`. -/
+noncomputable def map : a' ⟶ a :=
+  Classical.choose <| universal_property p f φ _ _ hf' φ'
+
+instance map_isHomLift : IsHomLift p g (map p f φ hf' φ') :=
+  (Classical.choose_spec <| universal_property p f φ _ _ hf' φ').1.1
+
+@[simp]
```
joelriou, line 187:
```suggestion
@[reassoc (attr := simp)]
```

Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
```diff
@@ -115,4 +143,200 @@ instance of_comp_iso {b' : 𝒳} (φ' : b ≅ b') [IsHomLift p (𝟙 S) φ'.hom]
...
+/-- Given a diagram
+```
+a'        a --φ--> b
+|         |        |
+v         v        v
+R' --g--> R --f--> S
+```
+such that `φ` is cartesian, and morphisms `φ' : a' ⟶ b`, `ψ ψ' : a' ⟶ a` such that
+`g ≫ ψ = φ' = g ≫ ψ'`. Then `ψ = ψ'`. -/
+protected lemma eq_of_fac {ψ ψ' : a' ⟶ a} [IsHomLift p g ψ] [IsHomLift p g ψ']
```
joelriou, line 212: Is it possible to formulate this lemma without referring to `φ'` so that the assumption become `ψ ≫ φ = ψ' ≫ φ`? If so, the lemma could be renamed `ext`?

callesonne: Great suggestion!

Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
```diff
@@ -115,6 +144,197 @@ instance of_comp_iso {b' : 𝒳} (φ' : b ≅ b') [IsHomLift p (𝟙 S) φ'.hom]
...
+
+end
+
+/-- The canonical isomorphism between the domains of two strongly cartesian morphisms lying over
+isomorphic objects. -/
+noncomputable def domainIsoOfBaseIso {R R' S : 𝒮} {a a' b : 𝒳} {f : R ⟶ S} {f' : R' ⟶ S}
+  {g : R' ≅ R} (h : f' = g.hom ≫ f) (φ : a ⟶ b) (φ' : a' ⟶ b) [IsStronglyCartesian p f φ]
+    [IsStronglyCartesian p f' φ'] : a' ≅ a where
+  hom := map p f φ h φ'
+  inv := @map _ _ _ _ p _ _ _ _ f' φ' _ _ _ _ _ (congrArg (g.inv ≫ ·) h.symm) φ
```
joelriou, line 335: Is there a way to avoid the `@map` syntax here?

joelriou:
Thanks!

bors d+

Mathlib/CategoryTheory/FiberedCategory/Cartesian.lean
callesonne: Yes, fixed!

PR #13413, opened 2024-05-31
[Merged by Bors] - fix(Tactic/Elementwise): type inference in elementwise

rmbellovin:
added an optional type argument to SimpType; changed type inference in elementwise tactic

Co-authored-by: Kim Morrison <kim@tqft.net>

---

### DIFF
--- a/Mathlib/Lean/Meta/Simp.lean
+++ b/Mathlib/Lean/Meta/Simp.lean
@@ -157,7 +157,11 @@ def simpOnlyNames (lemmas : List Name) (e : Expr) (config : Simp.Config := {}) :
 /--
 Given a simplifier `S : Expr → MetaM Simp.Result`,
 and an expression `e : Expr`, run `S` on the type of `e`, and then
-convert `e` into that simplified type, using a combination of type hints and `Eq.mp`.
+convert `e` into that simplified type,
+using a combination of type hints as well as casting if the proof is not definitional `Eq.mp`.
+
+The optional argument `type?`, if present, must be definitionally equal to the type of `e`.
+When it is specified we simplify this type rather than the inferred type of `e`.
 -/
 def simpType (S : Expr → MetaM Simp.Result) (e : Expr) (type? : Option Expr := none) :
     MetaM Expr := do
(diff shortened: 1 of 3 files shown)


semorrison:
I'll add a test to this before we merge.

This mistake in elementwise was causing it to produces lots of incorrect lemmas. The problem was that we were running `inferType` on the proof, which often disagrees with the expected type signature of the lemma we are making an auxiliary copy of.

semorrison:
I've added a test.

Mathlib/Lean/Meta/Simp.lean
```diff
@@ -159,8 +159,10 @@ Given a simplifier `S : Expr → MetaM Simp.Result`,
 and an expression `e : Expr`, run `S` on the type of `e`, and then
 convert `e` into that simplified type, using a combination of type hints and `Eq.mp`.
 -/
-def simpType (S : Expr → MetaM Simp.Result) (e : Expr) : MetaM Expr := do
-  match ← S (← inferType e) with
+def simpType (S : Expr → MetaM Simp.Result) (e : Expr) (typeHint : Option Expr := none) :
+    MetaM Expr := do
+  let type ← typeHint.getDM  (inferType e)
```
kmill, line 164: Rather than saying it's a hint, how about saying it's the (optional) type? Using a question mark like this is fairly standard in Lean APIs. Could you also add to the docstring that `type?` should be defeq to the type of `e`?
```suggestion
def simpType (S : Expr → MetaM Simp.Result) (e : Expr) (type? : Option Expr := none) :
    MetaM Expr := do
  let type ← type?.getDM (inferType e)
```

kmill (commented):
I guess the `simpType` change is independent of the change to `@[elementwise]`? I'm not sure `simpType` needs to be changed (a user can use `mkExpectedTypeHint`) but it doesn't hurt being able to pass in the type to save a type hint.

The change to `@[elementwise]` looks good to me. Once you address the comment about `simpType`, feel free to merge.

bors d+

semorrison:
Sorry, yes, we made the change in `simpType` when at first we thought we needed a deeper fix. I'll make the suggested changes.

semorrison:
bors merge

PR #13417, opened 2024-05-31
[Merged by Bors] - feat(Topology): fiberwise summation

jtristan:
Fiberwise summation for extended non-negative reals.

---

### DIFF
--- a/Mathlib/Topology/Instances/ENNReal.lean
+++ b/Mathlib/Topology/Instances/ENNReal.lean
@@ -1084,19 +1084,13 @@ theorem finset_card_const_le_le_of_tsum_le {ι : Type*} {a : ι → ℝ≥0∞} 
 #align ennreal.finset_card_const_le_le_of_tsum_le ENNReal.finset_card_const_le_le_of_tsum_le
 
 theorem HasSum_fiberwise {f : β → ENNReal} {a : ENNReal} (hf : HasSum f a) (g : β → γ) :
-    HasSum (fun (c : γ) ↦ ∑' (b : (g ⁻¹' {c})), f b) a := by
-  have B := @Equiv.hasSum_iff ENNReal β ((y : γ) × { x // g x = y }) _ _ f a (Equiv.sigmaFiberEquiv g)
-  replace B := B.2 hf
-  have C := @HasSum.sigma ENNReal γ _ _ _ _ (fun y : γ => { x // g x = y }) (f ∘ ⇑(Equiv.sigmaFiberEquiv g)) (fun c => ∑' (b : ↑(g ⁻¹' {c})), f ↑b) a B
-  apply C
-  intro b
-  have F := @Summable.hasSum_iff ENNReal _ _ _ (fun c => (f ∘ ⇑(Equiv.sigmaFiberEquiv g)) { fst := b, snd := c }) ((fun c => ∑' (b : ↑(g ⁻¹' {c})), f ↑b) b) _
-  apply (F _).2 rfl
-  apply ENNReal.summable
+    HasSum (fun (c : γ) ↦ ∑' (b : (g ⁻¹' {c})), f b) a :=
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/Instances/ENNReal.lean
```diff
@@ -1083,6 +1083,12 @@ theorem finset_card_const_le_le_of_tsum_le {ι : Type*} {a : ι → ℝ≥0∞}
     _ ≤ c := tsum_le_c
 #align ennreal.finset_card_const_le_le_of_tsum_le ENNReal.finset_card_const_le_le_of_tsum_le
 
+theorem tsum_fiberwise (f : β → ENNReal) (g : β → γ) :
```
sgouezel, line 1086:
```suggestion
theorem tsum_fiberwise (f : β → ℝ≥0∞) (g : β → γ) :
```
We use the notation for ENNReal most of the time.

Mathlib/Topology/Instances/ENNReal.lean
```diff
@@ -1083,6 +1083,12 @@ theorem finset_card_const_le_le_of_tsum_le {ι : Type*} {a : ι → ℝ≥0∞}
...
+    ∑' (x : γ), ∑' (b : (g ⁻¹' {x})), f b = ∑' i : β, f i :=
+  HasSum.tsum_eq <|
+  (((Equiv.sigmaFiberEquiv g).hasSum_iff).mpr (Summable.hasSum ENNReal.summable)).sigma <|
+  (fun _ ↦ ((ENNReal.summable).hasSum_iff).mpr rfl)
```
sgouezel, line 1090: Term mode proofs tend to be much harder to read than tactic proofs, where it is easier to read the order of arguments and the elementary steps. So, as a general rule, we avoid term mode for proofs which are longer than one line, although this is not an absolute rule. Do you think you could try to convert your proof to tactic mode?

sgouezel (commented):
Nitpicking comments, but this looks good!

Mathlib/Topology/Instances/ENNReal.lean
jtristan: Will do!

Mathlib/Topology/Instances/ENNReal.lean
```diff
@@ -1083,6 +1083,13 @@ theorem finset_card_const_le_le_of_tsum_le {ι : Type*} {a : ι → ℝ≥0∞}
     _ ≤ c := tsum_le_c
 #align ennreal.finset_card_const_le_le_of_tsum_le ENNReal.finset_card_const_le_le_of_tsum_le
 
+theorem tsum_fiberwise (f : β → ℝ≥0∞) (g : β → γ) :
+    ∑' (x : γ), ∑' (b : (g ⁻¹' {x})), f b = ∑' i : β, f i := by
+  apply HasSum.tsum_eq
+  let equiv := Equiv.sigmaFiberEquiv g
+  apply ((equiv.hasSum_iff).mpr (Summable.hasSum ENNReal.summable)).sigma
+  apply (fun _ ↦ ((ENNReal.summable).hasSum_iff).mpr rfl)
```
sgouezel, line 1091:
```suggestion
    ∑' x, ∑' b : g ⁻¹' {x}, f b = ∑' i, f i := by
  apply HasSum.tsum_eq
  let equiv := Equiv.sigmaFiberEquiv g
  apply (equiv.hasSum_iff.mpr ENNReal.summable.hasSum).sigma
  exact fun _ ↦ ENNReal.summable.hasSum_iff.mpr rfl
```
Thanks a lot, this looks great! I have removed a few useless parentheses (this tends to increase readability) and type annotations, and changed the last `apply` to `exact` (because this indicates this is the end of the proof). I have also used dot notation. This is just to show you some tricks, your proof was already perfectly acceptable!

sgouezel (commented):
bors d+
Thanks!

PR #13202, opened 2024-05-25
[Merged by Bors] - style: remove last isolated where

grunweg:
Discovered when tweaking the rewritten style linter in #13199.

---

Continuing the work started by @Komyyy.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #12991, opened 2024-05-17
[Merged by Bors] - style: remove all isolated `where`

Komyyy:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13420, opened 2024-05-31
[Merged by Bors] - feat: lint on isolated where and leading by

grunweg:
Extend the style linter with two tiny features:
- lint on "isolated where": all occurrences were already fixed in #12991 and #13202.
- lint on "leading by": if one line starts with `by ` but the previous line ends with `:=`, according to the style guide the `by` should move to the previous line. For now, we only lint if the `by` fits on the previous line (fixed in #13618 and predecessors): there about other 28 cases in mathlib which don't have an obvious fix.

---


robertylewis:
This looks reasonable to me, thanks! Delegating in case there are any errors introduced since the last update.

bors d+

grunweg:
Thank you for the review! There was one new occurrence, which I have just auto-fixed.
bors r+

PR #13421, opened 2024-05-31
[Merged by Bors] - feat (RingTheory/HahnSeries) SMulZeroClass instances

ScottCarnahan:
This PR adds instances for `SMulZeroClass R (HahnSeries Γ R)` and `SMulZeroClass R (HahnModule Γ R V)`.  Also, some minor variable cleanup.

---


Mathlib/RingTheory/HahnSeries/Multiplication.lean
```diff
@@ -80,25 +80,26 @@ def HahnModule (Γ R V : Type*) [PartialOrder Γ] [Zero V] [SMul R V] :=
 
 namespace HahnModule
 
-section
-
-variable {Γ R V : Type*} [PartialOrder Γ] [Zero V] [SMul R V]
```
jcommelin, line 85: What is the reason for inlining these typeclass assumptions?

jcommelin (commented):
Looks good to me. Minor question. Feel free to address it however you seem fit.

bors d+

Mathlib/RingTheory/HahnSeries/Multiplication.lean
ScottCarnahan: I cannot remember any particular reason.  I have restored the variable declaration inside a small section.

PR #13422, opened 2024-05-31
[Merged by Bors] - fix (Algebra/Vertex/HVertexOperator): rm defeq abuse in coeff function

ScottCarnahan:
The coefficient function for heterogeneous vertex operators currently uses the HahnSeries coefficient function on a Hahn module, which is a type alias.  This PR makes the use of the equivalence explicit. Also, we add a `HVertexOperator` namespace so various standard functions have cleaner names.

---


Ruben-VandeVelde:
If you're changing the full names for these, please add deprecated aliases

semorrison:
bors merge

PR #13423, opened 2024-05-31
[Merged by Bors] - feat(GroupTheory/SpecificGroups/Cyclic): add material on groups with given generator

MichaelStollBayreuth:
This adds some material to `Mathlib.GroupTheory.SpecificGroups.Cyclic` for dealing with groups that have a specified generator, mostly on existence and uniqueness of homomorphisms and isomorphisms in terms of the image of the given generator.

---


Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -787,3 +787,106 @@ noncomputable def mulEquivOfPrimeCardEq {p : ℕ} [Fintype G] [Fintype H] [Group
...
+
+The main statements are about the existence and uniqueness of homomorphisms and isomorphisms
+specified by the image of the given generator.
+-/
+
+variable {G G' : Type*} [Group G] [Group G'] {g : G} (hg : ∀ x, x ∈ Subgroup.zpowers g) {g' : G'}
+
+@[to_additive]
+lemma orderOf_generator_eq_natCard : orderOf g = Nat.card G :=
+  Nat.card_zpowers g ▸ (Nat.card_congr <| Equiv.subtypeUnivEquiv hg)
```
erdOne, line 807: I think this should go below `orderOf_eq_card_of_forall_mem_zpowers` (or maybe pull that lemma here)

erdOne: And maybe other lemmas mentioning `∀ x : α, x ∈ zpowers a` should be here as well?

MichaelStollBayreuth: OK; moving the lemma up (the other lemma is needed for proofs in between, so cannot be as easily moved).

MichaelStollBayreuth: > And maybe other lemmas mentioning `∀ x : α, x ∈ zpowers a` should be here as well?

Which place is "here"? The bottom part of the file where I put the new stuff?

erdOne: Yes. I assume grouping similar lemmas boost visibility? But don't bother if it isn't possible.

MichaelStollBayreuth: And I guess, based on the other lemmas using this condition, I should rename "generates" to "forall_mem_zpowers".

erdOne: Yeah I think so. Maybe we should define `IsGenerator x := ∀ x : α, x ∈ zpowers a` to enable dot products and better names some time in the future.

MichaelStollBayreuth: > Yes. I assume grouping similar lemmas boost visibility? But don't bother if it isn't possible.

The other lemmas are distrbiuted all over the file, so grouping them would be a fair amount of work, which I think is a bit out of scope for this PR.

> Yeah I think so. Maybe we should define `IsGenerator x := ∀ x : α, x ∈ zpowers a` to enable dot products and better names some time in the future.

It certainly would make sense to either define such a predicate or a `Generator` structure  bundling an element with the predicate (or both?), but this would be a larger refactor of the material and needs more effort to be done right.

erdOne: Fair enough. This is still a great step forwards nonetheless.

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -787,3 +791,105 @@ noncomputable def mulEquivOfPrimeCardEq {p : ℕ} [Fintype G] [Fintype H] [Group
...
+
+variable (hg' : orderOf g' ∣ orderOf (g : G))
+
+/-- If `g` generates the group `G` and `g'` is an element of another group `G'` whose order
+divides that of `g`, then there is a homomorphism `G →* G'` mapping `g` to `g'`. -/
+@[to_additive
+   "If `g` generates the additive group `G` and `g'` is an element of another additive group `G'`
+   whose order divides that of `g`, then there is a homomorphism `G →+ G'` mapping `g` to `g'`."]
+noncomputable
+def monoidHom_of_forall_mem_zpowers : G →* G' where
```
erdOne, line 821: This should be in snake case as this is a def.

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -787,3 +791,105 @@ noncomputable def mulEquivOfPrimeCardEq {p : ℕ} [Fintype G] [Fintype H] [Group
...
+                (Int.natCast_dvd_natCast.mpr hg').trans <| orderOf_dvd_iff_zpow_eq_one.mpr <|
+                Classical.choose_spec <| mem_zpowers_iff.mp <| hg 1
+  map_mul' x y := by
+    simp only [← zpow_add, zpow_eq_zpow_iff_modEq]
+    apply Int.ModEq.of_dvd (Int.natCast_dvd_natCast.mpr hg')
+    rw [← zpow_eq_zpow_iff_modEq, zpow_add]
+    simp only [fun x ↦ Classical.choose_spec <| mem_zpowers_iff.mp <| hg x]
+
+@[to_additive]
+lemma monoidHom_of_forall_mem_zpowers_apply_gen :
```
erdOne, line 833: should this be a simp lemma?

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -787,3 +791,105 @@ noncomputable def mulEquivOfPrimeCardEq {p : ℕ} [Fintype G] [Fintype H] [Group
...
+
+@[to_additive]
+lemma mulEquiv_of_orderOf_eq_symm_apply_gen : (mulEquiv_of_orderOf_eq hg hg' h).symm g' = g :=
+  monoidHom_of_forall_mem_zpowers_apply_gen hg' h.dvd
+
+@[to_additive]
+lemma mulEquiv_of_orderOf_eq_symm :
+    (mulEquiv_of_orderOf_eq hg hg' h).symm = mulEquiv_of_orderOf_eq hg' hg h.symm := by
+  rw [MulEquiv.eq_iff_eq_on_generator hg', mulEquiv_of_orderOf_eq_symm_apply_gen,
+    mulEquiv_of_orderOf_eq_apply_gen]
```
erdOne, line 891:
```suggestion
@[to_additive]
lemma mulEquiv_of_orderOf_eq_symm :
    (mulEquiv_of_orderOf_eq hg hg' h).symm = mulEquiv_of_orderOf_eq hg' hg h.symm := rfl
```

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -787,3 +791,105 @@ noncomputable def mulEquivOfPrimeCardEq {p : ℕ} [Fintype G] [Fintype H] [Group
...
+@[to_additive]
+lemma mulEquiv_of_orderOf_eq_apply_gen : mulEquiv_of_orderOf_eq hg hg' h g = g' :=
+  monoidHom_of_forall_mem_zpowers_apply_gen hg h.symm.dvd
+
+@[to_additive]
+lemma mulEquiv_of_orderOf_eq_symm_apply_gen : (mulEquiv_of_orderOf_eq hg hg' h).symm g' = g :=
+  monoidHom_of_forall_mem_zpowers_apply_gen hg' h.dvd
+
+@[to_additive]
+lemma mulEquiv_of_orderOf_eq_symm :
```
erdOne, line 888: should these be simp lemmas?

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
erdOne: Same for the other defs

MichaelStollBayreuth: Lower camel case, you mean? 
```text
  4. All other terms of Types (basically anything else) are in lowerCamelCase.
```
(because it is actually snake case). I'll change that.

erdOne: Oh yes. Sorry about that.

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
MichaelStollBayreuth: I can make them (all four you flagged) `simp` lemmas, but I'm doubtful that `simp` will fire often due to the generating conditions.

erdOne: They are present in the LHS so I believe simp can find it.

MichaelStollBayreuth: I take that back -- the conditions are explicit in the terms...

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13424, opened 2024-05-31
[Merged by Bors] - feat: add `NormedSpace.Core` structure to easily define normed vector spaces

dupuisf:
This PR introduces a `NormedSpace.Core` structure loosely based on `InnerProductSpace.Core`, which encapsulates the minimal axioms needed to show that a type forms a normed vector space, as found in textbooks. We then provide functions that take a `NormedSpace.Core` and produce instances of `NormedAddCommGroup` and `NormedSpace`.

Note that as it stands, defining a normed vector space from scratch (i.e. on a type without a preexisting notion of distance) is surprisingly painful, we have to keep converting between norm and distance, and there are many redundant proofs to provide.

Update (2024-07-25): This PR has now been significantly expanded to include tools to add a norm structure to a type that already has a uniformity or bornology. This involves "replacing" the uniformity/bornology induced by the norm by the preexisting one, while ensuring that the uniformity/bornology induced by the norm is a equal to the old one. (This is fairly common for e.g. matrices, where all reasonable norms induce the product uniformity.)

---


Mathlib/Analysis/NormedSpace/Basic.lean
```diff
@@ -514,3 +514,46 @@ def NormedAlgebra.restrictScalars : NormedAlgebra 𝕜 E :=
...
+
+section Core
+
+/-- A structure encapsulating minimal axioms needed to defined a normed vector space, as found
+in textbooks. This is meant to be used to easily define `NormedAddCommGroup E` and `NormedSpace E`
+instances from scratch on a type with no preexisting distance. -/
+structure NormedSpace.Core (𝕜 : Type*) (E : Type*) [NormedField 𝕜] [AddCommGroup E] [Module 𝕜 E]
+    [Norm E] : Prop where
+  norm_nonneg (x : E) : 0 ≤ ‖x‖
+  norm_eq_zero_iff (x : E) : ‖x‖ = 0 ↔ x = 0
```
j-loreaux, line 526: Can we drop this condition from `.Core`? Then we can have a `SeminormedAddCommGroup.ofCore` and a `NormedAddCommGroup.ofCore` where the latter requires additionally this hypothesis.

Mathlib/Analysis/NormedSpace/Basic.lean
```diff
@@ -514,3 +514,46 @@ def NormedAlgebra.restrictScalars : NormedAlgebra 𝕜 E :=
...
+    rw [this]
+    exact core.norm_triangle _ _
+  edist_dist x y := by exact (ENNReal.ofReal_eq_coe_nnreal _).symm
+  eq_of_dist_eq_zero {x} {y} h := by
+    rw [← sub_eq_zero, ← core.norm_eq_zero_iff]
+    exact h
+
+/-- Produces a `NormedSpace 𝕜 E` instance from a `NormedSpace.Core`. -/
+def NormedSpace.ofCore {𝕜 : Type*} {E : Type*} [NormedField 𝕜] [NormedAddCommGroup E] [Module 𝕜 E]
+    (core : NormedSpace.Core 𝕜 E) : NormedSpace 𝕜 E where
```
j-loreaux, line 556: You should just be able to assume `SeminormedAddCommGroup` here.

Also, you should instruct the user in the docstring that this is only intended for use when they have used one of the `.ofCore` declarations already (otherwise it's just silly).

Mathlib/Analysis/NormedSpace/Basic.lean
dupuisf: I just did this. It does feel a bit silly to have to have module stuff in something called `SeminormedAddCommGroup.Core` but I don't have a better naming idea. Also, I don't see how to avoid doing the proofs twice, is there a clean way to reuse the seminormed structure in the normed one?

j-loreaux (approved):
This looks pretty nice, thanks! I think all the `def`s should be marked `@[reducible]` since the point is that we will use them as instances.

bors d+

PR #13431, opened 2024-06-01
[Merged by Bors] - refactor(GroupTheory/Coset): Replace `Fintype.card` with `Nat.card`

tb65536:
A lot of the group theory library still needs to be switched over from `Fintype.card` to `Nat.card`. This PR switches over `GroupTheory/Coset.lean`. I had to add a little noise to a few other files, but these will get cleaned up once those files are switched over later.

---


Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
```diff
@@ -137,7 +137,7 @@ theorem Subgroup.eq_bot_or_eq_top_of_prime_card {G : Type*} [Group G] {_ : Finty
     (H : Subgroup G) [hp : Fact (Fintype.card G).Prime] : H = ⊥ ∨ H = ⊤ := by
   classical
   have := card_subgroup_dvd_card H
-  rwa [Nat.dvd_prime hp.1, ← Nat.card_eq_fintype_card, ← Nat.card_eq_fintype_card,
+  rwa [Nat.card_eq_fintype_card (α := G), Nat.dvd_prime hp.1, ← Nat.card_eq_fintype_card,
```
alreadydone, line 140: Not sure why this is necessary? Looks like you're just exchanging the order of two rewrites.

alreadydone (approved):
Thanks!

Mathlib/GroupTheory/SpecificGroups/Cyclic.lean
tb65536: It's necessary because `this` is in terms of `Nat.card` and `hp` is in terms of `Fintype.card`, so we have to switch `this` over to `Fintype.card` before apply `Nat.dvd_prime`.

tb65536: (This will all get cleaned up once the file is switched over to `Nat.card`).

alreadydone: I see, I missed that you removed an arrow.

maintainer merge

urkud:
Thanks! :tada:
bors merge

PR #13434, opened 2024-06-01
[Merged by Bors] - feat(NumberTheory/Harmonic): explicit formula for `ζ 1`

loefflerd:
Mathematically Riemann zeta is not defined at 1, but our construction ascribes a value to `ζ 1`. This PR calculates the value, and shows that it is not zero.

---

### DIFF
--- a/Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
+++ b/Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
@@ -409,8 +409,9 @@ lemma _root_.completedRiemannZeta₀_one :
 /-- With Mathlib's particular conventions, we have `ζ 1 ≠ 0`. -/
 lemma _root_.riemannZeta_one_ne_zero : riemannZeta 1 ≠ 0 := by
   -- This one's for you, Kevin.
-  rw [riemannZeta_one, ← ofReal_ofNat, ← ofReal_mul, ← ofReal_log (by positivity),
-    ← ofReal_sub, ← ofReal_ofNat, ← ofReal_div, ofReal_ne_zero]
+  suffices (γ - (4 * π).log) / 2 ≠ 0 by
+    simpa only [riemannZeta_one, ← ofReal_ne_zero, ofReal_log (by positivity : 0 ≤ 4 * π),
+      push_cast]
   refine div_ne_zero (sub_lt_zero.mpr (lt_trans ?_ ?_ (b := 1))).ne two_ne_zero
   · exact Real.eulerMascheroniConstant_lt_two_thirds.trans (by norm_num)
   · rw [lt_log_iff_exp_lt (by positivity)]
--- a/Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
+++ b/Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
@@ -361,7 +361,7 @@ lemma tendsto_Gamma_term_aux : Tendsto (fun s ↦ 1 / (s - 1) - 1 / Gammaℝ s /
   have h := hasDerivAt_Gammaℝ_one
   rw [hasDerivAt_iff_tendsto_slope, slope_fun_def_field, Gammaℝ_one] at h
   have := h.div (hasDerivAt_Gammaℝ_one.continuousAt.tendsto.mono_left nhdsWithin_le_nhds)
-    (by rw [Gammaℝ_one]; exact one_ne_zero)
+    (Gammaℝ_one.trans_ne one_ne_zero)
   rw [Gammaℝ_one, div_one] at this
   refine this.congr' ?_
   have : {z | 0 < re z} ∈ 𝓝 (1 : ℂ) := by
@@ -377,7 +377,7 @@ lemma tendsto_riemannZeta_sub_one_div_Gammaℝ :
     (𝓝 ((γ - Complex.log (4 * ↑π)) / 2)) := by
   have := tendsto_riemannZeta_sub_one_div.add tendsto_Gamma_term_aux
   simp_rw [sub_add_sub_cancel] at this
-  convert this using 1
+  convert this using 2
   ring_nf
 
 /-- Formula for `ζ 1`. Note that mathematically `ζ 1` is undefined, but our construction ascribes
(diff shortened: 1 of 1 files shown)


MichaelStollBayreuth:
I'm wondering what this is good for (apart perhaps from `completedRiemannZeta₀_one`), as it seems that we are talking about "junk values" here that should be irrelevant in applications. @loefflerd, can you explain what purpose you have in mind with this?

Parcly-Taxel:
> I'm wondering what this is good for (apart perhaps from `completedRiemannZeta₀_one`), as it seems that we are talking about "junk values" here that should be irrelevant in applications. @loefflerd, can you explain what purpose you have in mind with this?

`completedRiemannZeta₀ 1 = (γ - Complex.log (4 * ↑π)) / 2 + 1 =` [OEIS A074760](https://oeis.org/A074760).

loefflerd:
When I initially added the statement of the Riemann hypothesis to mathlib, I formalised it as "if s is not a negative even integer, then `ζ s ≠ 0`". Kevin pointed out that this statement might very well be false if the value for `ζ 1` was not well chosen, and took me to task on his public twitter feed for my sloppiness. 

So the main motivation of this PR is to clear up this mess by demonstrating that the value we assigned for `ζ 1` was nonzero all along.

More generally, there is a convention that "junk" values should be explicit; e.g. we have a lemma `div_zero` saying that `x / 0 = 0`, we don't just assign a random value but never disclose what it is. (For this reason I seem to recall one of the reviewers when I first added the zeta function – not sure who, maybe Sebastien? – writing that it would be good if we had a lemma explaining precisely what `ζ 1` is.)

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
```diff
@@ -347,4 +352,71 @@ lemma _root_.isBigO_riemannZeta_sub_one_div {F : Type*} [Norm F] [One F] [NormOn
...
+section val_at_one
+
+open Complex
+
+lemma tendsto_Gamma_term_aux : Tendsto (fun s ↦ 1 / (s - 1) - 1 / Gammaℝ s / (s - 1)) (𝓝[≠] 1)
+    (𝓝 (-(γ + Complex.log (4 * ↑π)) / 2)) := by
+  have h := hasDerivAt_Gammaℝ_one
+  rw [hasDerivAt_iff_tendsto_slope, slope_fun_def_field, Gammaℝ_one] at h
+  have := h.div (hasDerivAt_Gammaℝ_one.continuousAt.tendsto.mono_left nhdsWithin_le_nhds)
+    (by rw [Gammaℝ_one]; exact one_ne_zero)
```
MichaelStollBayreuth, line 364:
```suggestion
    (Gammaℝ_one.trans_ne one_ne_zero)
```

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
```diff
@@ -347,4 +352,71 @@ lemma _root_.isBigO_riemannZeta_sub_one_div {F : Type*} [Norm F] [One F] [NormOn
...
+  filter_upwards [this] with a ha _
+  rw [Pi.div_apply, ← sub_div, div_right_comm, sub_div' _ _ _ (Gammaℝ_ne_zero_of_re_pos ha),
+    one_mul]
+
+lemma tendsto_riemannZeta_sub_one_div_Gammaℝ :
+    Tendsto (fun s ↦ riemannZeta s - 1 / Gammaℝ s / (s - 1)) (𝓝[≠] 1)
+    (𝓝 ((γ - Complex.log (4 * ↑π)) / 2)) := by
+  have := tendsto_riemannZeta_sub_one_div.add tendsto_Gamma_term_aux
+  simp_rw [sub_add_sub_cancel] at this
+  convert this using 1
```
MichaelStollBayreuth, line 380:
```suggestion
  convert this using 2
```
speeds it up by a third :smile:

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
```diff
@@ -347,4 +352,71 @@ lemma _root_.isBigO_riemannZeta_sub_one_div {F : Type*} [Norm F] [One F] [NormOn
...
+    completedRiemannZeta₀ 1 = (γ - Complex.log (4 * ↑π)) / 2 + 1 := by
+  have := completedRiemannZeta_eq 1
+  rw [sub_self, div_zero, div_one, sub_zero, eq_sub_iff_add_eq] at this
+  rw [← this, completedRiemannZeta_one]
+
+/-- With Mathlib's particular conventions, we have `ζ 1 ≠ 0`. -/
+lemma _root_.riemannZeta_one_ne_zero : riemannZeta 1 ≠ 0 := by
+  -- This one's for you, Kevin.
+  rw [riemannZeta_one, ← ofReal_ofNat, ← ofReal_mul, ← ofReal_log (by positivity),
+    ← ofReal_sub, ← ofReal_ofNat, ← ofReal_div, ofReal_ne_zero]
```
MichaelStollBayreuth, line 413: This looks like `norm_cast` should be able to do it (after `rw [riemannZeta_one]`).
I think the problem is `ofReal_log` that needs a positivity assumption (`ofReal_log` is also not tagged `norm_cast`, but trying to add the attribute locally does not help). It would be nice if `norm_cast` could do this, but this is another topic...

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
```diff
@@ -347,4 +352,71 @@ lemma _root_.isBigO_riemannZeta_sub_one_div {F : Type*} [Norm F] [One F] [NormOn
...
+  refine div_ne_zero (sub_lt_zero.mpr (lt_trans ?_ ?_ (b := 1))).ne two_ne_zero
+  · apply Real.eulerMascheroniConstant_lt_two_thirds.trans (by norm_num)
```
MichaelStollBayreuth, line 415:
```suggestion
  · exact Real.eulerMascheroniConstant_lt_two_thirds.trans (by norm_num)
```

MichaelStollBayreuth:
OK. It certainly does not hurt to have this, and the value of `Λ₀` at 1 could be useful (and it needs the other ones anyway).
I won't "maintainer merge" (yet), so that Kevin has an opportunity to have a look.

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
loefflerd: That's a handy lemma, I didn't know about that one :smile:

Mathlib/NumberTheory/Harmonic/ZetaAsymp.lean
MichaelStollBayreuth: I think it is a bit clearer than doing all the explicit rewrites with `ofReal` lemmas, so yes.

loefflerd: Gesagt, getan.

MichaelStollBayreuth (approved):
LGTM!
@kbuzzard are you going to have a look?

MichaelStollBayreuth:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

kbuzzard:
LGTM. Thanks David!

PR #13439, opened 2024-06-01
[Merged by Bors] - fix(`gitpod.yml`): force elan to self-update in gitpod

zeramorphic:
Ensure that `elan` is up-to-date before trying to run `lake exe cache get` etc in gitpod containers.

---

Because this change is in `gitpod.yml` and not in the `Dockerfile`, this will be executed on startup of the container, and not cached. This means that we'll never be running an out-of-date `elan`.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,4 +6,6 @@ vscode:
     - leanprover.lean4
 
 tasks:
-  - init: lake exe cache get
+  - init: |
+      elan self update
+      lake exe cache get


YaelDillies (approved):
Thank you very much!

bors merge

PS: For anyone wondering, we do want the docker image to contain *any* version of elan, as installing elan takes much longer than updating it.

YaelDillies:
Whoop sorry I meant

maintainer merge

semorrison:
bors merge

PR #13440, opened 2024-06-01
[Merged by Bors] - feat(Order/Interval): conversion Icc <-> range

rwst:
Add lemma `range_eq_Icc_zero_sub_one`, as direct conversion between Icc and range is missing

---

### DIFF
--- a/Mathlib/Order/Interval/Finset/Nat.lean
+++ b/Mathlib/Order/Interval/Finset/Nat.lean
@@ -67,6 +67,11 @@ theorem Iio_eq_range : Iio = range := by
 theorem Ico_zero_eq_range : Ico 0 = range := by rw [← Nat.bot_eq_zero, ← Iio_eq_Ico, Iio_eq_range]
 #align nat.Ico_zero_eq_range Nat.Ico_zero_eq_range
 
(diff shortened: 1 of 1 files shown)


Mathlib/Order/Interval/Finset/Nat.lean
```diff
@@ -67,6 +67,11 @@ theorem Iio_eq_range : Iio = range := by
 theorem Ico_zero_eq_range : Ico 0 = range := by rw [← Nat.bot_eq_zero, ← Iio_eq_Ico, Iio_eq_range]
 #align nat.Ico_zero_eq_range Nat.Ico_zero_eq_range
 
+lemma range_eq_Icc_zero_minus (n : ℕ) (hn : n > 0): range n = Icc 0 (n - 1) := by
```
urkud, line 70: Please use `sub_one` or `pred` in the name. Also, please assume `n ≠ 0`, not `n > 0`. Otherwise LGTM. Thanks!

rwst: Done, thanks.

urkud:
Thanks! :tada:
bors merge

PR #13444, opened 2024-06-01
[Merged by Bors] - chore: Make `Subgroup.coe_set_mk`/`Subsemigroup.coe_set_mk` a `norm_cast` lemma

YaelDillies:
Those are the only eligible lemmas of their series as all other `coe_set_mk` lemmas have a bundled set coercion on their RHS.

---


urkud:
Thanks! :tada:
bors merge

PR #13448, opened 2024-06-01
[Merged by Bors] - fix(AlgebraicGeometry/EllipticCurve): high prio for point notation

kbuzzard:
This makes things like `#synth AddCommGroup W⟮ℚ⟯` parse, for `W : WeierstrassCurve ℚ` (before, you had to write `#synth AddCommGroup (W⟮ℚ⟯)`).

---

I don't really understand all this `notation3` stuff, so someone should check what I'm doing, but I was inspired by
```
@[inherit_doc] scoped[Polynomial] notation:9000 R "[X]" => Polynomial R
```


Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -634,7 +634,7 @@ inductive Point
 #align weierstrass_curve.point WeierstrassCurve.Affine.Point
 
 /-- For an algebraic extension `S` of `R`, the type of nonsingular `S`-rational points on `W`. -/
-scoped[WeierstrassCurve] notation3 W "⟮" S "⟯" => Affine.Point <| baseChange W S
+scoped[WeierstrassCurve] notation3:9000 W "⟮" S "⟯" => Affine.Point <| baseChange W S
```
Multramate, line 637: I believe `notation3` uses the Lean 3 `notation` delaborator, which happens to be more flexible? I don't know what the arbitrary number `9000` is, but I would just do `notation3:max` like other places in the repo:
```suggestion
scoped[WeierstrassCurve] notation3:max W "⟮" S "⟯" => Affine.Point <| baseChange W S
```
I did something similar for `R[X][Y]` in #13356 (which is currently stuck in a merge but I've just fixed it).

Multramate (approved):

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -639,7 +639,7 @@ inductive Point
 #align weierstrass_curve.point WeierstrassCurve.Affine.Point
 
 /-- For an algebraic extension `S` of `R`, the type of nonsingular `S`-rational points on `W`. -/
-scoped notation3 W "⟮" S "⟯" => Affine.Point <| baseChange W S
+scoped[WeierstrassCurve] notation3:max W "⟮" S "⟯" => Affine.Point <| baseChange W S
```
Multramate, line 642: Seems like there's no need for `[WeierstrassCurve]` since it's already in the namespace:
```suggestion
scoped notation3:max W "⟮" S "⟯" => Affine.Point <| baseChange W S
```

alreadydone:
Thanks :)
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13450, opened 2024-06-01
[Merged by Bors] - fix (RingTheory/HahnSeries/Multiplication) : rm defeq abuse in SMul instance

ScottCarnahan:
This PR replaces `y` with `((of R).symm y)` in 6 places, so that functions that take Hahn series are given Hahn series instead of Hahn modules (which are defeq).

---


robertylewis:
bors merge

PR #13447, opened 2024-06-01
[Merged by Bors] - feat: restriction of scalars for (pre)sheaves of modules

joelriou:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13441, opened 2024-06-01
[Merged by Bors] - feat: the sheafification functor for presheaves of modules

joelriou:
In this PR, we construct the sheafification functor for presheaves of modules, and show that it is a left adjoint functor.

---

- [x] depends on: #11555
- [x] depends on: #13447 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #11555, opened 2024-03-21
[Merged by Bors] - feat: the associated sheaf of a presheaf of modules

joelriou:
Let `C` be a category equipped with a Grothendieck topology `J`. Let `α : R₀ ⟶ R.val` be a morphism, with `R₀` a presheaf of rings and `R` a sheaf of rings. We assume that `α` is a sheafification map (i.e. it is both locally injective and locally surjective) . Let `M₀` be a presheaf of modules over `R₀`. Let `A` be the associated sheaf of the underlying presheaf of abelian groups of `M₀`. Then, it is shown in this PR that there is a canonical structure of sheaf of `R`-modules on `A`.

---

- [x] depends on: #6845
- [x] depends on: #11479
- [x] depends on: #13324

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13455, opened 2024-06-02
[Merged by Bors] - feat: the category of sheaves of modules is abelian

joelriou:
---

- [x] depends on: #13441
- [x] depends on: #11555
- [x] depends on: #13447 


Mathlib/Algebra/Category/ModuleCat/Presheaf/ChangeOfRings.lean
```diff
@@ -54,4 +54,9 @@ noncomputable def restrictScalars (α : R ⟶ R') :
     { hom := φ.hom
       map_smul := fun X r ↦ φ.map_smul X (α.app _ r) }
 
+instance (α : R ⟶ R') : (restrictScalars.{v} α).Additive where
+
+instance : (restrictScalars (𝟙 R)).Full := inferInstanceAs (𝟭 _).Full
+instance : (restrictScalars (𝟙 R)).Faithful := inferInstanceAs (𝟭 _).Faithful
```
erdOne, line 60: Faithful-ness should hold for all `α` (see the code on the previous PR).

Mathlib/CategoryTheory/Functor/ReflectsIso.lean
```diff
@@ -71,6 +71,13 @@ instance reflectsIsomorphisms_of_comp (F : C ⥤ D) (G : D ⥤ E)
     haveI := isIso_of_reflects_iso (F.map f) G
     exact isIso_of_reflects_iso f F⟩
 
+lemma reflectsIsomorphisms_of_comp' (F : C ⥤ D) (G : D ⥤ E)
```
erdOne, line 74: arguably the previous lemma should be `reflectsIsomorphisms_comp` and this should be unprimed.

Mathlib/CategoryTheory/Sites/Sheaf.lean
```diff
@@ -386,6 +386,9 @@ instance : (sheafToPresheaf J A).Full :=
 instance : (sheafToPresheaf J A).Faithful :=
   (fullyFaithfulSheafToPresheaf J A).faithful
 
+instance : (sheafToPresheaf J A).ReflectsIsomorphisms :=
+  (fullyFaithfulSheafToPresheaf J A).reflectsIsomorphisms
```
erdOne, line 390: Do we not have the instance `[Full F] [Faithful F] : ReflectsIsomorphisms F`? We should add one.

Mathlib/Algebra/Category/ModuleCat/Sheaf/Abelian.lean
```diff
@@ -0,0 +1,29 @@
...
+universe v v' u u'
+
+open CategoryTheory Limits
+
+variable {C : Type u'} [Category.{v'} C] {J : GrothendieckTopology C}
+
+namespace SheafOfModules
+
+variable (R : Sheaf J RingCat.{u}) [HasSheafify J AddCommGroupCat.{v}]
+  [J.WEqualsLocallyBijective AddCommGroupCat.{v}]
```
erdOne, line 23: Can you add a comment on what to import and what universes constraints there are for typical applications?

Mathlib/Algebra/Category/ModuleCat/Presheaf/ChangeOfRings.lean
joelriou: Yes, this is indeed very easy!

Mathlib/CategoryTheory/Functor/ReflectsIso.lean
joelriou: Yes, indeed.

Mathlib/CategoryTheory/Sites/Sheaf.lean
joelriou: At some point working on this PR, this instance was not found automatically, but now, it seems it works :-)

Mathlib/AlgebraicGeometry/Modules/Sheaf.lean
```diff
@@ -4,24 +4,28 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+open CategoryTheory
+
 namespace AlgebraicGeometry.Scheme
 
 variable (X : Scheme.{u})
 
 /-- The category of sheaves of modules over a scheme. -/
 abbrev Modules := SheafOfModules.{u} X.ringCatSheaf
 
+noncomputable instance : Abelian X.Modules := inferInstance
```
erdOne: 🎉

Mathlib/Algebra/Category/ModuleCat/Sheaf/Abelian.lean
joelriou: Thanks for the suggestion. I have added the remark that if one imports `Algebra.Category.GroupCat.FilteredColimits`, then the `Abelian` instance is found automatically in the case of small categories.

Mathlib/AlgebraicGeometry/Modules/Sheaf.lean
joelriou: Actually, I have removed it because I am not completely sure what happens when two (defeq) `instance` declarations are made. But, at least, if `AlgebraicGeometry.Modules.Sheaf` and `ModuleCat.Sheaf.Abelian` are imported, this will be found!

erdOne: I believe it doesn't have bad effects as `inferInstance` is an `abbrev`. In fact it even helps because typeclass inference no longer needs to jump through hoops.

joelriou: Ok, I have added it back...

erdOne (commented):
Great work!
maintainer merge

riccardobrasca:
This is great, thanks!

bors merge

PR #12386, opened 2024-04-24
[Merged by Bors] - perf(NumberTheory.NumberField.Basic): make `RingOfIntegers` a Type.

riccardobrasca:
We make `𝓞 K` a type rather then a `Subalgebra` to speed up instances search.

Co-authored-by: Anne Baanen <[Vierkantor](https://github.com/Vierkantor)>
Co-authored-by: Yaël Dillies <[YaelDillies](https://github.com/YaelDillies)>
Co-authored-by: Matthew Ballard <[mattrobball](https://github.com/mattrobball)>

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13456, opened 2024-06-02
perf(Geometry/Manifold/Instances/Sphere): speed up

grunweg:
Speed up the file by about 34% by doing the following
- add a few shortcut instances
- squeeze simps where the full form is short and a clear win, also when it's clearly worth it (use `says` in this case)
- "squeeze" three really slow `field_simp` calls: replace them by the equivalent `simp` invocation and squeeze that
- avoid some expensive `convert`s
- squeeze two really slow `nlinarith` calls (they're still slow, but *much* better)
- replace one `congr!` by `congr`

This file is still very slow: some `nlinarith` calls and a few slow `convert`s are the main offenders.
Help fixing any of these is welcome (but also perfect for a subsequent PR).

---

This file still contains a number of notes about slow parts: if this PR looks good in principle, I'll happily take those out.
If you see them and feel like scratching that itch, be my guest!


grunweg:
!bench

grunweg:
!bench

grunweg:
The latest commit should speed up this file even more; #13459 will benchmark this also.

grunweg:
!bench

grunweg:
Marking as awaiting-author to
- file proper tracking issues about the field_simp hacks
- fix the instances properly: `making Submodule.span return a type a'la #12386 might help`

grunweg:
I've split out #15488 for a hopefully less controversial subset of these changes.

PR #13458, opened 2024-06-02
perf(Manifold/ContMDiff/NormedSpace): add shortcut instances

grunweg:
This speeds up the file by 10%.

---------
The effect on the profiling output seems mixed (some "typeclass synthesis took 100ms" like appear after this change, but not before)... but the benchmarking seems clear.

---


grunweg:
!bench

grunweg:
Let's try again, now that CI definitely passed.
!bench

grunweg:
!bench

grunweg:
Feedback from @mattroball on zulip: `The first local instance in #13458 is almost surely tripping synthPendingDepth. Maybe the second also`
awaiting-author

grunweg:
Closing this PR: broader actions are more useful here.

PR #13461, opened 2024-06-02
[Merged by Bors] - feat: the free (pre)sheaf of modules of rank 1

joelriou:
---


Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -416,3 +416,56 @@ lemma restrictionApp_toPresheafOfModules {X Y : Cᵒᵖ} (f : X ⟶ Y) :
...
+  map {X Y} f := by
+    exact
+      { toFun := (R.map f).toFun
+        map_add' := by simp
+        map_smul' := by simp }
+
+/-- The obvious free presheaf of modules of rank `1`. -/
+abbrev unit : PresheafOfModules R := (unitCore R).toPresheafOfModules
+
+lemma unit_map_one {X Y : Cᵒᵖ} (f : X ⟶ Y) : (unit R).map f (1 : R.obj X) = (1 : R.obj Y) :=
```
erdOne, line 436: Doesn't this hold for any `PresheafOfModules`? Do we have this (as a simp lemma)?

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -416,3 +416,56 @@ lemma restrictionApp_toPresheafOfModules {X Y : Cᵒᵖ} (f : X ⟶ Y) :
...
+/-- The type of sections of a presheaf of modules. -/
+def sections (M : PresheafOfModules.{v} R) : Type _ := (M.presheaf ⋙ forget _).sections
+
+@[ext]
+lemma sections_ext {M : PresheafOfModules.{v} R} (s t : M.sections)
+    (h : ∀ (X : Cᵒᵖ), s.val X = t.val X) : s = t :=
+  Subtype.ext (by ext; apply h)
+
+/-- The bijection `(unit R ⟶ M) ≃ M.sections` for `M : PresheafOfModules R`. -/
+def unitHomEquiv (M : PresheafOfModules R) :
```
erdOne, line 450: Does `simps` generate good simp lemmas?

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
joelriou: No, this is specific to the ring presheaf considered as a presheaf of modules over itself.

callesonne: I don't think so, because `1` only makes sense in the underlying ring.

callesonne, line 436:
```suggestion
lemma unit_map_one {X Y : Cᵒᵖ} (f : X ⟶ Y) : (unit R).presheaf.map f (1 : R.obj X) = (1 : R.obj Y) :=
```
It might be a good idea to write the lemma like this. This makes it so that the `erw [unit_map_one]` below would become a `rw`. Then we have the `simp` lemma `PresheafOfModules.map_apply` which could be used to deduce the version without the `.presheaf`

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -416,3 +416,56 @@ lemma restrictionApp_toPresheafOfModules {X Y : Cᵒᵖ} (f : X ⟶ Y) :
...
+    (unit R ⟶ M) ≃ M.sections where
+  toFun f :=
+    { val := fun X => f.app X (1 : R.obj X)
+      property := fun {X Y} p => by
+        dsimp
+        erw [← NatTrans.naturality_apply, unit_map_one]
```
callesonne, line 457:
```suggestion
        erw [← naturality_apply, unit_map_one]
```
The lemma `PresheafOfModules.naturality_apply` seems even closer to what you want in this situation, my hope is that somehow that lemma can be used to remove the need for `erw` but I can't seem to get it to work

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
erdOne: I don't like that it makes it longer though. I think `(unit R).map` should be the preferred spelling. For the `erw` problem, maybe we should make `PresheafOfModules.map` reducible, but it might be a head symbol and it might not work. Anyway these problems are everywhere in the AG library and I believe some smart people are thinking about how to solve it (Kim, Johan, Dagur etc.) and I think we can leave this as is for now. 
Hopefully we can have a thorough discussion or brainstorm at the AG workshop.

erdOne: To clarify, I'm not stopping you from figuring this out. Another smart mind gives a greater chance in coming up with a better solution! It's just that I believe it is a bigger problem and I don't think it is possible to solve it here.

callesonne: Okay thanks for the answer! I wasn't aware of how widespread these issues were (and that some people are thinking on how to solve it!), then its fair to leave it as it is I think :)

joelriou: There is a mathematical issue here also: when studying morphisms of presheaves of modules, sometimes we want to look at the underlying morphisms of presheaves of abelian groups (this is `.presheaf`), and sometimes we want to look at the family of linear maps between modules of sections (this is `Hom.app`). I would think that `Hom.app` should be the better spelling when applicable.

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
joelriou: Thanks!

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
callesonne: But then we can at least get rid of `erw [unit_map_one]` by doing `rw [← map_apply, unit_map_one]`

callesonne: > There is a mathematical issue here also: when studying morphisms of presheaves of modules, sometimes we want to look at the underlying morphisms of presheaves of abelian groups (this is related to `.presheaf`), and sometimes we want to look at the family of linear maps between modules of sections (this is `Hom.app`). I would think that `Hom.app` should be the better spelling when applicable.

I agree with this. Do you have any thoughts on the `@[simp]` from `map_apply` in this file then, i.e. should it be removed? Because then `dsimp` would automatically try to phrase things in terms of the underlying presheaves of abelian groups, but maybe thats not a big problem?

joelriou: Thanks for the suggestion! I have slightly improved simp lemmas in this commit https://github.com/leanprover-community/mathlib4/pull/13461/commits/ed087fd3bc36f70c3944968e8ef9d828f10b8382

erdOne:
Thanks! (and also thanks Calle for reviewing!)
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13463, opened 2024-06-02
[Merged by Bors] - feat: the pushforward of (pre)sheaves of modules

joelriou:
---

- [x] depends on: #13447

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13464, opened 2024-06-02
[Merged by Bors] - feat: quasi-coherent sheaves

joelriou:
In this PR, we define the notion of quasi-coherent sheaf: a sheaf of modules `M` over a sheaf of rings `R` (on a category `C` equipped with a Grothendieck topology) is quasi-coherent if, locally, it admits a presentation involving coproducts of copies of the restriction of `R`. We also introduce the notions of sheaves of modules of finite type/presentation.

These are only definitions. No attempt is made yet to prove anything about this notion in the case of schemes!

---

- [x] depends on: #13720
- [x] depends on: #13447
- [x] depends on: #13463
- [x] depends on: #13461
- [x] depends on: #13455
- [x] depends on: #13441
- [x] depends on: #11555


Mathlib/Algebra/Category/ModuleCat/Sheaf/Quasicoherent.lean
```diff
@@ -0,0 +1,116 @@
...
+  generators i := (q.presentation i).generators
+
+end QuasicoherentData
+
+/-- A sheaf of modules is quasi-coherent if it is locally the cokernel of a
+morphism between coproducts of copies of the sheaf of rings. -/
+class IsQuasicoherent : Prop where
+  nonempty_quasicoherentData : Nonempty M.QuasicoherentData
+
+/-- A sheaf of modules is quasi-coherent if it is locally the cokernel of a
```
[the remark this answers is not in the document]
joelriou: Thanks! Indeed, too much of copy-pasting..,

Mathlib/Algebra/Category/ModuleCat/Sheaf/Generators.lean
```diff
@@ -73,4 +88,39 @@ def equivOfIso (e : M ≅ N) :
 
 end GeneratingSections
 
+variable [∀ (X : C), HasWeakSheafify (J.over X) AddCommGrp.{u}]
```
kbuzzard, line 91: Maths question: how close is `∀ (X : C), HasWeakSheafify (J.over X) AddCommGrp.{u}` to just `HasWeakSheafify J AddCommGrp.{u}`? Do you know examples where it works for J but fails for some J.over X? This is independent of the merging process of course, I was just interested.

Mathlib/Algebra/Category/ModuleCat/Sheaf/Generators.lean
```diff
@@ -73,4 +88,39 @@ def equivOfIso (e : M ≅ N) :
...
+  X : I → C
+  coversTop : J.CoversTop X
+  /-- the data of sections of `M` over `X i` which generate `M.over (X i)` -/
+  generators (i : I) : (M.over (X i)).GeneratingSections
+
+/-- A sheaf of modules is of finite type if locally, it is generated by finitely
+many sections. -/
+class IsFiniteType : Prop where
+  exists_localGeneratorsData :
+    ∃ (σ : M.LocalGeneratorsData), ∀ (i : σ.I), Finite (σ.generators i).I
```
kbuzzard, line 110: If I understood it right, then the following thing would be `IsFiniteType`: C is the category associated to some infinite discrete space like Nat, with sheaf of modules Int=Z, and attached to n you have the Z-module Z^n. Then you can cover with Nat (i.e. sigma.X=id) and you have something which I would instinctively say was "locally finite type", but *not* finite type. Is this notation standard or are we just making stuff up for mathlib here? If this is our own invention then maybe `IsLocallyFiniteType` is a better name? What do you think?

Mathlib/Algebra/Category/ModuleCat/Sheaf/PushforwardContinuous.lean
```diff
@@ -40,4 +40,9 @@ noncomputable def pushforward : SheafOfModules.{v} R ⥤ SheafOfModules.{v} S wh
   map f :=
     { val :=(PresheafOfModules.pushforward φ.val).map f.val }
 
+/-- Given `M : SheafOfModules R` and `X : D`, this is the restriction of `M`
+over the sheaf of rings `R.over X` on the category `Over X`. -/
+noncomputable abbrev over (M : SheafOfModules.{v} R) (X : D) : SheafOfModules.{v} (R.over X) :=
+  (pushforward.{v} (𝟙 _)).obj M
+
```
kbuzzard, line 47: I know this isn't part of your PR, but line 30 of this file looks pretty weird to me. Is that really what was intended?

kbuzzard (commented):
I'll try and debug those super-slow instances later, but I don't see why this should hold this PR back. I've left some comments.

Mathlib/Algebra/Category/ModuleCat/Sheaf/Generators.lean
joelriou: Your example is indeed of finite type. This is standard terminology: "finite type" means locally of finite type, see https://stacks.math.columbia.edu/tag/01B5

Here, I think that there is a parallel with relative properties of morphisms of schemes: we should think that to be a finite type sheaf of modules is a relative property "over" the site. For example, in the language of schemes, if you construct the scheme `S` which is a coproduct of countably many copies of the spectrum of the complex numbers (the underlying topological space is the discrete set `Nat`), then the scheme `X` which is a disjoint union of affine spaces of dimension `n` for all `n : Nat` over `ℂ`, then you get a morphism `X ⟶ S` in a way that the fibre over `n : Nat` (identified to `S`) is the `n`-dimensional affine space over `ℂ`. Yet, `X` is a finite type `S`-scheme, not just locally of finite type! (As a scheme over `ℂ` in an obvious way, `X` is no longer of finite type, but it is locally of finite type...)

Mathlib/Algebra/Category/ModuleCat/Sheaf/PushforwardContinuous.lean
joelriou: You must be referring to the apparently duplicate assumption `[Functor.IsContinuous.{u} F J K] [Functor.IsContinuous.{v} F J K]`. Continuous functors are defined by saying that the "candidate" pushforward functor indeed sends sheaves of types to sheaves of types, but there is a universe parameter. I need both because for the degree of generality in this file, I need to pushforward both the sheaf of rings (in `Type u`) and the sheaf of modules (in `Type v`). In the case of schemes, we shall have `v = u`.

semorrison:
bors merge

joelriou:
Retrying...

bors merge

PR #11448, opened 2024-03-17
[Merged by Bors] - refactor: Generalise special case of Jensen

YaelDillies:
`Real.pow_sum_div_card_le_sum_pow` and `NNReal.pow_sum_div_card_le_sum_pow` both state that `(∑ i in s, f i) ^ (n + 1) / s.card ^ n ≤ ∑ i in s, f i ^ (n + 1)`, but one for `f : ι → ℝ` and the other for `f : ι → ℝ≥0`.
Both lemmas can be unified by deriving the inequality for `f : ι → α` where `[LinearOrderedSemifield α] [ExistsAddOfLE α]`. This involves generalising the rearrangement inequality semirings.

From LeanAPAP

---

- [x] depends on: #13949
- [x] depends on: #15815 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13465, opened 2024-06-02
[Merged by Bors] - chore(Algebra/Order/Module/Defs): remove unused assumptions

astrainfinita:
---


urkud:
Why the linter doesn't catch them?

negiizhao:
Maybe some lemmas use it, but the argument `[OrderedAddCommGroup β]` already provides it.

urkud:
I think that these lemmas were accidentally moved from `Monoid` to `Group` section.

YaelDillies:
"unused" is not quite the correct word. They are rather "redundant". It seems this redundancy happened as I was writing the proofs and forgetting to update some statements. Anyway, the change is unequivocally good, so

maintainer merge

YaelDillies:
Turns out I had already found an even better generalisation in #11448, which is still sitting there unreviewed 😅

PR #13467, opened 2024-06-02
[Merged by Bors] - feat: create aliases and lemmas for `CovariantClass α α (· * ·) (· ≤ ·)` etc

astrainfinita:
The first step towards removing `CovariantClass` and `ContravariantClass`.

---


semorrison:
I'm sceptical about adding more ways to say the same things. Is there some previous discussion about the pros and cons here?

adomani:
I agree with Kim.

As you know, I introduced the analogous [docs#PosMulMono](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/GroupWithZero/Unbundled.html#PosMulMono) and its friends a while ago, thinking that it was a good idea.

I've come to regret this decision later on, since I think that it adds very little extra understanding to someone who does not know what `Co(ntra)variant` models and adds yet another name to have to come to grips with for everyone else.

For this reason, I would recommend not adding these `abbrev`s.

The other place where a similar "new name" for a `Co(ntra)variant`-like assumption is [docs#PosSMulMono](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/Module/Defs.html#PosSMulMono).  However, these have their justification in helping typeclass search not get derailed.

Does this sound reasonable?  :smile:

j-loreaux:
I agree with Kim and Damiano

j-loreaux:
I'm confused. Zulip said you should add these, but *remove* `CovariantClass` and `ContravariantClass`.

j-loreaux:
@semorrison and @adomani, are you on board with this now? When I commented earlier I thought we were just adding `abbrev`s, but given the discussion on Zulip, it seems this is more principled. Thoughts?

adomani:
Given the speed improvements, I am definitely more in favour of this change now, than I was before!

Did I hallucinate that this PR had been `!bench`ed?

jcommelin:
Thanks :tada:

bors merge

PR #13470, opened 2024-06-02
[Merged by Bors] - feat(Analysis/Matrix): lemmas about norms of matrix, preparation for Siegel's lemma

fbarroero:
This PR contains `norm_eq_sup_sup_nnnorm (A : Matrix m n α) : ‖A‖ = Finset.sup Finset.univ fun i ↦ Finset.sup Finset.univ fun j ↦ ‖A i j‖₊`. It is used in the proof of Siegel's lemma.

Some preliminary results used in the proof of Siegel's lemma can be found at #13487

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Siegel's.20Lemma)

Co-authored-by: Laura Capuano and Amos Turchet @amosturchet 

---


Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,118 @@
...
+
+## References
+
+See [M. Hindry and J. Silverman, Diophantine Geometry: an Introduction] .
+-/
+
+/- We set ‖⬝‖ to be Matrix.seminormedAddCommGroup  -/
+attribute [local instance] Matrix.seminormedAddCommGroup
+
+open  Finset
```
faenuccio, line 36:
```suggestion
open Finset
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,118 @@
...
+determines has a non-zero integer solution `t` with `‖t‖ ≤ ((n * ‖A‖) ^ ((m : ℝ) / (n - m)))`
+
+## Notation
+
+ - `‖_‖ ` : Matrix.seminormedAddCommGroup is the sup norm, the maximum of the absolute values of
+ the entries of the matrix
+
+## References
+
+See [M. Hindry and J. Silverman, Diophantine Geometry: an Introduction] .
```
faenuccio, line 30: You should add this book to the .bib file and then add the reference.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,118 @@
...
+# Siegel's Lemma, first version for ℤ
+
+In this file we introduce and prove Siegel's Lemma in its most basic version. This is a fundamental
+tool in diophantine approximation and transcendency and says that there exists a "small" integral
+non-zero solution of a non-trivial overdetermined system of linear equations with integer
+coefficients.
+
+## Main results
+
+- `siegels_lemma`: Given a non-singular `m × n` matrix `A` with `m < n` the linear system it
```
faenuccio, line 20: This is not yet in this file, right? And it should probably have an explicit name (and let Siegel only appear in the doc)

Mathlib/Data/Matrix/Basic.lean
```diff
@@ -1677,6 +1677,9 @@ def mulVec [Fintype n] (M : Matrix m n α) (v : n → α) : m → α
 @[inherit_doc]
 scoped infixr:73 " *ᵥ " => Matrix.mulVec
 
+lemma mulVec_def [Fintype n] (M : Matrix m n α) (v : n → α) :
```
faenuccio, line 1680: Is it really useful? Can you point at a place where this is needed.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,118 @@
...
+
+ - `‖_‖ ` : Matrix.seminormedAddCommGroup is the sup norm, the maximum of the absolute values of
+ the entries of the matrix
+
+## References
+
+See [M. Hindry and J. Silverman, Diophantine Geometry: an Introduction] .
+-/
+
+/- We set ‖⬝‖ to be Matrix.seminormedAddCommGroup  -/
```
faenuccio, line 33: Most of the results below have nothing special to do with ℕ and ℝ, but with a general cast function. Also, they are part of the matrix library, so they should be moved to a file where the sup-norm is defined.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,118 @@
...
+/-- sup commutes with casting from Nat to NNReal -/
+lemma cast_sup_eq_sup_cast_Nat_NNReal {S : Type*} (f : S → ℕ) (s : Finset S) :
+    (sup s f) = sup s fun b ↦ (f b : NNReal) :=
+  comp_sup_eq_sup_comp_of_is_total Nat.cast Nat.mono_cast
+  (by simp only [bot_eq_zero', CharP.cast_eq_zero])
+
+namespace Int.Matrix
+
+/-- The definition of ‖⬝‖ for integral matrixes -/
+lemma sup_sup_norm_def (m n : ℕ) (A : Matrix (Fin m) (Fin n) ℤ)  :
```
faenuccio, line 47: Do you need the matrix to be indexed by `Fin n` and `Fin m` or any pair of `Fintype` would do? It would be closer to the whole library design.

Mathlib/Data/Matrix/Basic.lean
fbarroero: In the first step of the proof of Siegel's lemma one needs to unfold the definition of mulVec. I thought the use of `unfold` was deprecated and one is supposed to write lemmas like that one.

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: I erased that. I will add again with the final PR.

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: Done

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: I changed it to Fintype as you suggest

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: I just got rid of cast_sup_eq_sup_cast_Nat_NNReal because I could just call comp_sup_eq_sup_comp_of_is_total where needed. I will then move these lemmas in the sup-norm folder.

faenuccio:
Here is what I suggest: that you leave in this PR only the modifications to the matrix file and that you move all the others (including the biblio) to #13487. Once this is done, I think that the main result you are adding is not only relevant to naturals vs integers, but more generally. So I would propose that you
1. Create a very general lemma about the explicit description of the L_infty norm being the sup of the sup (you might take some inspiration from what is done down in the file); this can be done in every normed commutative (semi?)group.
1. State your lemma about the existence of a natural number that coincides with the matrix norm of an integer matrix; but can you briefly explain why you need this and having a positive integer would not suffice for your application?
1. One general advice: as you are noticing, merging PR's take a long time. You have all the interest in keeping your branch updated with `master`, lest the merging, once your modifications are approved, will be painful. You can come to Zulip if you need help about how keeping the branches updated.

fbarroero:
> Here is what I suggest: that you leave in this PR only the modifications to the matrix file and that you move all the others (including the biblio) to #13487. 

This is done. 

I'll try to generalise the lemmas on matrices.

fbarroero:
> 1. Create a very general lemma about the explicit description of the L_infty norm being the sup of the sup (you might take some inspiration from what is done down in the file); this can be done in every normed commutative (semi?)group.

Done

> 2. State your lemma about the existence of a natural number that coincides with the matrix norm of an integer matrix; but can you briefly explain why you need this and having a positive integer would not suffice for your application?

Done. Actually I think it is easier to get a natural number instead of a positive integer because of the lemma NNReal.natCast_natAbs. 

> 3. One general advice: as you are noticing, merging PR's take a long time. You have all the interest in keeping your branch updated with `master`, lest the merging, once your modifications are approved, will be painful. You can come to Zulip if you need help about how keeping the branches updated.

Done.

Thanks for the help, Filippo!!

Mathlib/Analysis/Matrix.lean
```diff
@@ -185,6 +190,35 @@ theorem norm_diagonal [DecidableEq n] (v : n → α) : ‖diagonal v‖ = ‖v
...
+  rw [norm_eq_sup_sup_nnnorm, ← NNReal.coe_natCast, NNReal.coe_inj,
+    comp_sup_eq_sup_comp_of_is_total Nat.cast Nat.mono_cast
+    (by simp only [bot_eq_zero', CharP.cast_eq_zero])]
+  congr
+  rw [Function.comp_def]
+  ext; congr
+  rw [comp_sup_eq_sup_comp_of_is_total Nat.cast Nat.mono_cast
+     (by simp only [bot_eq_zero', CharP.cast_eq_zero])]
+  congr; ext; congr;
+  rw [Function.comp_apply, NNReal.natCast_natAbs]
```
faenuccio, line 209:
```suggestion
    use sup univ fun i ↦ sup univ fun j ↦ (A i j).natAbs
  simp only [norm_eq_sup_sup_nnnorm, ← NNReal.coe_natCast, bot_eq_zero', CharP.cast_eq_zero,
    comp_sup_eq_sup_comp_of_is_total Nat.cast Nat.mono_cast, Function.comp_def, NNReal.coe_inj]
  congr! with n m
  exact (NNReal.natCast_natAbs (A n m)).symm
```

Mathlib/Analysis/Matrix.lean
```diff
@@ -185,6 +190,35 @@ theorem norm_diagonal [DecidableEq n] (v : n → α) : ‖diagonal v‖ = ‖v
...
+
+/-- The norm of a non-singular integral matrix is a positive natural number-/
+lemma one_le_norm_of_nonzero (A : Matrix m n ℤ) (hA_nezero : A ≠ 0) (a : ℕ)
+    (h_norm_nat : ‖A‖ = ↑a) : 1 ≤ a := by
```
faenuccio, line 213:
```suggestion
    (h_norm_nat : ‖A‖ = a) : 1 ≤ a := by
```

Mathlib/Analysis/Matrix.lean
```diff
@@ -185,6 +190,35 @@ theorem norm_diagonal [DecidableEq n] (v : n → α) : ‖diagonal v‖ = ‖v
...
+  congr
+  rw [Function.comp_def]
+  ext; congr
+  rw [comp_sup_eq_sup_comp_of_is_total Nat.cast Nat.mono_cast
+     (by simp only [bot_eq_zero', CharP.cast_eq_zero])]
+  congr; ext; congr;
+  rw [Function.comp_apply, NNReal.natCast_natAbs]
+
+/-- The norm of a non-singular integral matrix is a positive natural number-/
+lemma one_le_norm_of_nonzero (A : Matrix m n ℤ) (hA_nezero : A ≠ 0) (a : ℕ)
```
faenuccio, line 212: Isn't it more natural to add a `def natNorm : ℕ` that is defined as what you currently have on line 199 and then restate `norm_eq_NatCast` as `norm_eq_natNorm`, so that you can then state this lemma as `one_le_natNorm` showing that the `natNorm` of a non-zero matrix is at least one?

faenuccio (commented):
Thanks for all the work!

Mathlib/Analysis/Matrix.lean
fbarroero: I implemented your suggestions. It makes more sense.

Thanks for the advice!

Mathlib/Analysis/Matrix.lean
```diff
@@ -185,6 +190,30 @@ theorem norm_diagonal [DecidableEq n] (v : n → α) : ‖diagonal v‖ = ‖v
...
+    natNorm]
+  congr! with i j
+  exact (NNReal.natCast_natAbs (A i j)).symm
+
+/-- The natNorm of a non-zero integral matrix is at least 1-/
+lemma one_le_natNorm (A : Matrix m n ℤ) (hA_nezero : A ≠ 0) : 1 ≤ natNorm A := by
+  by_contra h
+  simp only [not_le, Nat.lt_one_iff] at h
+  simp only [← norm_pos_iff', norm_eq_natNorm A, h, CharP.cast_eq_zero, lt_self_iff_false]
+    at hA_nezero
```
faenuccio, line 213:
```suggestion
  simp only [← norm_pos_iff', norm_eq_natNorm A, Nat.cast_pos] at hA_nezero
  linarith
```

faenuccio:
Thanks! Once you integrate my two pennies of golfing, I am happy with all changes. My only question is whether this file is the right one for the `Int` part, as that one has no analytic content. On the other hand, it has some "normed"  content, so it might be OK to leave it here. I'd appreciate @j-loreaux or @eric-wieser or @hrmacbeth 's opinion on this.

eric-wieser:
I'm not sure whether `natNorm` is something we want; it's starting to feel like we might want arbitrarily-valued norms (which @linesthatinterlace wanted at some point).

I'm also unsure about `norm_eq_sup_sup_nnnorm`, mixing norm and nnnorm in the statement feels a little ugly.

Can you add an explanation in the PR description of why you need `Nat`-valued norms?

j-loreaux:
Personally, I don't see why we need `natNorm` at all. Why not just a rewrite lemma and a few theorems for integer-valued matrices?

fbarroero:
> I'm also unsure about `norm_eq_sup_sup_nnnorm`, mixing norm and nnnorm in the statement feels a little ugly.

This lemma simply unfolds the definition of the elementewise norm. The nnnorm comes up because of [Pi.norm_def](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Group/Basic.html#Pi.norm_def).
 
> Can you add an explanation in the PR description of why you need `Nat`-valued norms?

Done. I hope this helps.

fbarroero:
> Personally, I don't see why we need `natNorm` at all. Why not just a rewrite lemma and a few theorems for integer-valued matrices?

`natNorm` was introduced after a suggestion of @faenuccio. Before that there used to be a lemma showing that there is a natural number whose cast is equal to the norm of the matrix.
I think now it is much better and cleaner.

faenuccio:
I think that in arithmetic context you might want to have `natNorm` pretty much for the same reason we both have `Norm` and `NNorm` in analytic ones. If you only work with the $$\mathbb{Z}$$-valued one, you constantly need to keep track of the positivity assumption, which is automatically dealt with by things like `omega` and `linarith` if all terms are in `Nat`. What @fbarroero is doing, is to go towards diophantine approximation, where you need to mix normed/geometric arguments with arithmetic ones, so I feel that this `natNorm` gadget makes some sense. But both @eric-wieser and @j-loreaux have more experience, so if there are still more objections I will withdraw my suggestion.

j-loreaux:
So, @faenuccio my issue here is that there isn't really much gain. That is, there's only a handful of lemmas, and they could easily be phrased without `natNorm`. Moreover, `natNorm` is just a bare function, it's not bundled in any way, so you don't get any benefits there.

As Eric mentioned, I think if we want something like this, we need to developed an API for distances/norms that take values in other places besides `ℝ`. There was some discussion about this (only visible to Filippo, sorry Fabrizio) regarding another PR, see [here](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/.2310680.20pseudo.20metrics.20with.20generic.20codomain/near/431317837).

In any case, I'm not going to put my foot down here. If you want to `maintainer merge` this, I think there's relatively little harm that can come from it unless the plan for the future is to develop a significant API around `natNorm`. In the latter case, we should probably follow the suggestions in that thread.

eric-wieser:
> If you want to `maintainer merge` this, I think there's relatively little harm that can come from it unless the plan for the future is to develop a significant API around `natNorm`.

I think if we do merge it it might be more appropriate in a different file nearer wherever Siegel's lemma will end up; or perhaps, combined into that PR so that reviewers can work out whether there are ways to avoid it entirely.

faenuccio:
According to @j-loreaux and @eric-wieser I would suggest that we simply keep in this PR the first lemma about the norm being equal to the sup-sup; and that you move the other results to a file closer to your result about Siegel's lemma, perhaps in a `Matrix` namespace.

fbarroero:
> According to @j-loreaux and @eric-wieser I would suggest that we simply keep in this PR the first lemma about the norm being equal to the sup-sup; and that you move the other results to a file closer to your result about Siegel's lemma, perhaps in a `Matrix` namespace.

Done! Many thanks to you and to @j-loreaux and @eric-wieser!

faenuccio:
maintainer merge

fpvandoorn:
bors merge

PR #13385, opened 2024-05-30
[Merged by Bors] - chore(*): remove `refine'`

grunweg:
Some of these changes are simple replacements, others are more involved.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13474, opened 2024-06-03
[Merged by Bors] - chore(LinearAlgebra–Topology): remove `refine'`

Parcly-Taxel:
Split from #13385; this excludes usages where `refine'` was used to replace `refine_struct` – except where the proofs for each field were already separately written out.

---


Parcly-Taxel:
!bench

Parcly-Taxel:
!bench

adomani (approved):
Looks good to me, thanks!

Here is a summary of the non-`refine'`-to-`refine` changes:
* a couple of `convert`s in `LinearAlgebra`,
* a few scattered `apply`s (I'm not a big fan of `apply` with more than one side-goal, but this is a separate issue),
* some `exact`s -- this is a potential performance issue, but `!bench` does not highlight anything in particular,
* some removed porting notes.

Overall, this is an improvement, in my opinion!

jcommelin:
Thanks :tada:

bors merge

PR #12605, opened 2024-05-02
chore: attribute [induction_eliminator]

astrainfinita:
Add attribute [induction_eliminator] to
`AdjoinRoot.induction_on`
`ENat.recTopCoe`
`ENNReal.recTopCoe`
`Finset.induction`
`Magma.AssocQuotient.induction_on`
`ManyOneDegree.ind_on`
`Module.Ray.ind` (and add `Orientation.ind` for `Orientation`, an abbrev of it. I wish that we do not need to add it in the future.)
`Multiset.induction`
`MvPolynomial.induction_on`
`OnePoint.rec`
`Opposite.rec'`
`Ordinal.limitRecOn`
`PartENat.casesOn`
`Polynomial.induction_on'`
`QuotientAddGroup.induction_on'` (and add `AddCircle.induction_on` for `AddCircle`, an abbrev of it. I wish that we do not need to add it in the future.)
`QuotientGroup.induction_on'` (doesn't actually work)
`Real.Angle.induction_on`
`SimplexCategory.rec`
`Trunc.induction_on`

---

- [x] depends on: #13476
- [x] depends on: #13264

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13476, opened 2024-06-03
[Merged by Bors] - chore: tidy induction principles for Unitization and TrivSqZeroExt

eric-wieser:
This adds `elab_as_elim`, `induction_eliminator`, and `cases_eliminator` attributes, which eliminates the need to use the `using` term of `induction`.

Also choose a better argument name, since this is used as the case name.

Split from #12605 

---


Ruben-VandeVelde (approved):
maintainer merge

PR #9857, opened 2024-01-19
[Merged by Bors] - feat: update nolints periodically

grunweg:
Add a new CI job which updates `nolints.json` and `style-exceptions.txt` periodically.

Cargo-culted from the mathlib 3 versions at
https://github.com/leanprover-community/mathlib/blob/master/scripts/update_nolints.sh (update script) and
https://github.com/leanprover-community/azure-scripts/blob/master/.github/workflows/nolints.yml (workflow).
Please review carefully!

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.239857.20update.20nolints.20periodically)

--------------
Open questions
- time interval: the mathlib3 version was once a day. That seems a bit excessive to me; I chose once a week for now (every Sunday).
- Are all the steps in the workflow necessary? (I cargo-culted them from `bors.yml`.) For instance, I can imagine the cleanup, uninstall pr-toolchain, print versions steps might not be necessary? (I don't *know* at all, though!) 
- Everything related to secrets and authentification is completely cargo-culted.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13482, opened 2024-06-03
[Merged by Bors] - fix: make `update_nolints.sh` executable

grunweg:
An oversight from #9857; perhaps this is the cause of the latest error.

---


bryangingechen:
Thanks!
bors r+

PR #13484, opened 2024-06-03
chore: remove files about core programming types

PatrickMassot:
This removes files that are used nowhere in Mathlib and are all about core programming data types. In particular those files define many instances, hence interfere with core development for all downstream users.

---


grunweg (commented):
Does what it says; the reasoning seems plausible. I don't have enough context to judge if there are hidden downsides I don't know. I do wonder, though:

is equivalent functionality present in core Lean or batteries? If not, could it be useful to file an issue and point to these files, before they're getting deleted? The answer to this could be worth recording, as additional context to whoever else comes looking for these.

Mathlib/Data/String/Basic.lean
```diff
@@ -1,200 +0,0 @@
...
-#align string.to_list_nonempty String.toList_nonempty
-
-@[simp]
-theorem head_empty : "".data.head! = default :=
-  rfl
-#align string.head_empty String.head_empty
-
-#align string.popn_empty String.drop_empty
-
-instance : LinearOrder String where
```
eric-wieser, line 154: I made this remark privately, but for transparency I'll make it here too.

There generic approach of mathlib means that many `instance`s are never actually used; but that doesn't mean that the instance is useless. In this case, the instance enables things like
```lean
import Mathlib
#eval ["hello", "world"].mergeSort (· ≤ ·)
```
work. Removing this seems like it is going to make it needlessly harder to write scripts and meta code.

grunweg:
Coming here for PR triage: this PR has a merge conflict, but is also seems that this mostly needs a decision of this PR is wanted. I'll label it awaiting-zulip accordingly.

kim-em:
We're dealing with these on a case by case basis.

kim-em:
Thanks, though!!

PR #13485, opened 2024-06-03
[Merged by Bors] - feat(Topology/Category): category of finite topological spaces

chrisflav:
The category of finite topological spaces.

---

This is needed to obtain the category of finite continuous `G`-types for a topological group `G`. This would then be
`ContAction FinTopCat G`.

### DIFF
--- a/Mathlib/Topology/Category/FinTopCat.lean
+++ b/Mathlib/Topology/Category/FinTopCat.lean
@@ -1,0 +1,77 @@
+/-
+Copyright (c) 2024 Christian Merten. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Merten
+-/
+import Mathlib.CategoryTheory.FintypeCat
+import Mathlib.Topology.Category.TopCat.Basic
+
+/-!
+# Category of finite topological spaces
+
+Definition of the category of finite topological spaces with the canonical
+forgetful functors.
+
+-/
+
+
+universe u
+
+open CategoryTheory
+
+/-- A bundled finite topological space. -/
+structure FinTopCat where
+  /-- carrier of a finite topological space. -/
+  α : Type u
+  [fintype : Fintype α]
+  [topologicalSpace : TopologicalSpace α]
+
+namespace FinTopCat
+
+instance : Inhabited FinTopCat :=
+  ⟨⟨PUnit⟩⟩
+
+instance : CoeSort FinTopCat (Type u) :=
+  ⟨FinTopCat.α⟩
+
+attribute [instance] fintype topologicalSpace
+
+instance : Category FinTopCat.{u} where
+  Hom X Y := { f : X → Y // Continuous f }
+  id X := ⟨id, by continuity⟩
+  comp f g := ⟨g.val.comp f.val, g.property.comp f.property⟩
+
+instance : ConcreteCategory FinTopCat.{u} where
+  forget := { obj := fun X ↦ X.α, map := fun f ↦ f.val }
+  forget_faithful := ⟨fun {_ _ a b} h ↦ Subtype.ext h⟩
+
+instance (X : FinTopCat) : TopologicalSpace ((forget FinTopCat).obj X) :=
+  X.topologicalSpace
+
+instance (X : FinTopCat) : Fintype ((forget FinTopCat).obj X) :=
+  X.fintype
+
+/-- Construct a bundled `FinTopCat` from the underlying type and the appropriate typeclasses. -/
+def of (X : Type u) [Fintype X] [TopologicalSpace X] : FinTopCat :=
+  ⟨X⟩
+
+@[simp]
+theorem coe_of (X : Type u) [Fintype X] [TopologicalSpace X] :
+    (of X : Type u) = X :=
+  rfl
+
+/-- The forgetful functor to `FintypeCat`. -/
+instance : HasForget₂ FinTopCat FintypeCat :=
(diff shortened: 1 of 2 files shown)


Mathlib/Topology/Category/FinTopCat.lean
```diff
@@ -0,0 +1,77 @@
...
+
+universe u
+
+open CategoryTheory
+
+/-- A bundled finite topological space. -/
+structure FinTopCat where
+  /-- carrier of a finite topological space. -/
+  α : Type u
+  [fintype : Fintype α]
```
dagurtomas, line 26: Is there a reason not to use `Finite α` instead of `Fintype`?

Mathlib/Topology/Category/FinTopCat.lean
```diff
@@ -0,0 +1,77 @@
...
+instance : Inhabited FinTopCat :=
+  ⟨⟨PUnit⟩⟩
+
+instance : CoeSort FinTopCat (Type u) :=
+  ⟨FinTopCat.α⟩
+
+attribute [instance] fintype topologicalSpace
+
+instance : Category FinTopCat.{u} where
+  Hom X Y := { f : X → Y // Continuous f }
```
dagurtomas, line 40: You probably want to use `C(X, Y)` instead. An even simpler solution would be to use `InducedCategory.category` to inherit the category instance from `TopCat`, but then you need to change the definition of `FinTopCat` so that one of the fields is an object of `TopCat`. This gives you lots of the stuff below for free, you can compare with the definitions of `CompHaus` and `Profinite` for example.

Mathlib/Topology/Category/FinTopCat.lean
chrisflav: Yes, because `FintypeCat` uses `Fintype`, so to have a computable forgetful functor to `FintypeCat`, this is needed.

dagurtomas: You can still define that functor -- you can always get a `Fintype` instance from a `Finite` instance (`Fintype.ofFinite`) -- it's just that `Finite` is a `Prop` and `Fintype` contains data. You can read the docstring [here](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Finite/Defs.html#Finite) for an explanation on when to use `Finite` and when to use `Fintype`

chrisflav: I agree, but that's what would make the forgetful functor `noncomputable`. And, as I said, for consistency I think it makes sense to have the same typeclass assumption as in `FintypeCat`.

If I am understanding correctly, if I start with an object `X` of `FintypeCat` with a topological space instance and write `(forget₂ FinTopCat FintypeCat).obj (FinTopCat.of X)`, with the `Finite` approach I would get two (possibly non def-eq) `Fintype` instances on `X` (the one I started with and the one from `Fintype.ofFinite`).

chrisflav: Note that I don't care about the `Fintype` instance, I just think that `FintypeCat` and `FinTopCat` should use the same notion of finiteness.

And I agree with you, I am in favor of using `Finite` instead of `Fintype`.

erdOne: I agree that `FinTopCat` and `FintypeCat` should have the same finiteness. Someone might want to open a Zulip discussion on whether `FintypeCat` should be changed to `FiniteCat` in case `FintypeCat` has some combinatorial applications that needs the `Fintype` instance.

Mathlib/Topology/Category/FinTopCat.lean
```diff
@@ -0,0 +1,74 @@
...
+  InducedCategory.category toTop
+
+instance : ConcreteCategory FinTopCat :=
+  InducedCategory.concreteCategory _
+
+instance (X : FinTopCat) : TopologicalSpace ((forget FinTopCat).obj X) :=
+  inferInstanceAs <| TopologicalSpace X
+
+instance (X : FinTopCat) : Fintype ((forget FinTopCat).obj X) :=
+  X.fintype
```
erdOne, line 48: Should the spelling be `(forget _).obj X` or `X` (from the `CoeSort` instance)?

Mathlib/Topology/Category/FinTopCat.lean
joelriou: I agree that unless the name/definition of `FintypeCat` is changed, it is very much ok to keep the `Fintype` assumption here.

Mathlib/Topology/Category/FinTopCat.lean
joelriou: If this does not break things, I would favour the `CoeSort` versoin.

chrisflav: The `CoeSort` version already works:
```
example (X : FinTopCat) : Fintype X := inferInstance
```
but
```
example (X : FinTopCat) : Fintype ((forget FinTopCat).obj X) := inferInstance
```
does not without the instance above and I think I'll need this one.

erdOne: It's sad that we need that. I really hope we can sort out the coercions of concrete categories one day.

chrisflav: Should I do anything about this in this case or can we merge this PR then?

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

chrisflav:
Thanks for the reviews!

PR #13486, opened 2024-06-03
[Merged by Bors] - feat(CategoryTheory): add some instances for essentially small sites

dagurtomas:
This PR proves that the properties `PreservesSheafification` and `WEqualsLocallyBijective` hold for an essentially small site if they hold for an equivalent small site.

---


joelriou:
Could you try to investigate if there are more general classes of functors between sites which preserve/reflect locally injective/surjective morphisms? For example, the postcomposition with functors like `Over.forget` must preserve locally injective and locally surjective morphisms; does this hold more generally for `CoverPreserving` functors? In the other direction, is the `IsCoverDense` condition (+ fully/faithful) sufficient? I think it would be better to prove suitable general results, and then apply them to equivalences.

dagurtomas:
Indeed, `IsContinuous` is enough to preserve locally surjectivity. I haven't managed to prove the other direction for a cover dense fully faithful functor, I'm not sure how to relate the image sieve and `coverByImage`. Here is a failed attempt which ends in something that amounts to proving that `f` is objectwise surjective on images of the functor:

```lean
lemma isLocallySurjective_of_whisker' (H : D ⥤ C) [H.IsCoverDense J] [H.Full] [H.Faithful]
    [IsLocallySurjective (H.inducedTopologyOfIsCoverDense J) (whiskerLeft H.op f)] :
      IsLocallySurjective J f where
  imageSieve_mem {X} a := by
    have := Functor.IsCoverDense.is_cover (G := H) (K := J) X
    refine GrothendieckTopology.superset_covering _ ?_ this
    intro Y g ⟨⟨Z, lift, map, fac⟩⟩
    let b := (forget A).map (G.map map.op) a
    have hh := IsLocallySurjective.imageSieve_mem (f := (whiskerLeft H.op f))
      (J := H.inducedTopologyOfIsCoverDense J) b
    change _ ∈ J.sieves _ at hh
    rw [← fac]
    apply Sieve.downward_closed
    fconstructor
    · sorry
    · change _ = b
      sorry
```

dagurtomas:
Ok, here is a more promising approach:

```lean
lemma isLocallySurjective_of_whisker' (H : D ⥤ C) [H.IsCoverDense J] [H.Full] [H.Faithful]
    [IsLocallySurjective (H.inducedTopologyOfIsCoverDense J) (whiskerLeft H.op f)] :
      IsLocallySurjective J f where
  imageSieve_mem {X} a := by
    have := Functor.IsCoverDense.is_cover (G := H) (K := J) X
    apply J.transitive this
    intro Y g ⟨⟨Z, lift, map, fac⟩⟩
    rw [← fac, Sieve.pullback_comp]
    apply J.pullback_stable

    let b := (forget A).map (G.map map.op) a
    have hh := IsLocallySurjective.imageSieve_mem (f := (whiskerLeft H.op f))
      (J := H.inducedTopologyOfIsCoverDense J) b
    change _ ∈ J.sieves _ at hh
    have : imageSieve (whiskerLeft H.op f) b =
        (Sieve.pullback map (imageSieve f a)).functorPullback H := by
      sorry
    rw [this] at hh
    exact GrothendieckTopology.superset_covering _ (Sieve.functorPullback_pushforward_le H _) hh
```

dagurtomas:
Ok, the latter approach worked. I bet it can be modified for locally injective as well. I'll finish it and clean up tomorrow.

joelriou:
> Ok, the latter approach worked. I bet it can be modified for locally injective as well. I'll finish it and clean up tomorrow.

The better statement is:
```lean
import Mathlib.CategoryTheory.Sites.LocallySurjective
import Mathlib.CategoryTheory.Sites.DenseSubsite

namespace CategoryTheory.Presheaf

attribute [local instance] ConcreteCategory.hasCoeToSort ConcreteCategory.instFunLike

variable {D C A : Type*} [Category C] [Category D] [Category A] [ConcreteCategory A]
  (H : D ⥤ C) {K : GrothendieckTopology D} {J : GrothendieckTopology C}
  (hH : CoverPreserving K J H) [H.IsCoverDense J]
  {F G : Cᵒᵖ ⥤ A} (φ : F ⟶ G) [IsLocallySurjective K (whiskerLeft H.op φ)]

lemma isLocallySurjective_of_whisker : IsLocallySurjective J φ where
  imageSieve_mem {X} s := by
    let R : ⦃Y : C⦄ → ⦃f : Y ⟶ X⦄ → Sieve.coverByImage H X f → Sieve Y :=
      fun Y f _ => imageSieve φ (G.map f.op s)
    refine J.superset_covering ?_ (J.bind_covering
      (R := R) (H.is_cover_of_isCoverDense J X) ?_)
    · rintro Y f ⟨Z, a, b, hb, ha, rfl⟩
      exact ⟨localPreimage _ _ _ ha, by simp⟩
    · rintro Y f ⟨hf⟩
      refine J.superset_covering ?_ (J.pullback_stable hf.lift
        (hH.cover_preserve (imageSieve_mem K (whiskerLeft H.op φ)
          (U := Opposite.op hf.obj) (G.map hf.map.op s))))
      intro Z a ⟨T, b, c, hb, fac⟩
      refine ⟨F.map c.op (localPreimage _ _ _ hb), ?_⟩
      replace fac := (G ⋙ forget _).congr_map (congr_arg Quiver.Hom.op (fac =≫ hf.map))
      simp only [Category.assoc, op_comp, Functor.map_comp] at fac
      conv_rhs => rw [← hf.fac]
      rw [NatTrans.naturality_apply, app_localPreimage, op_comp, G.map_comp, comp_apply]
      exact (congr_fun fac s).symm

end CategoryTheory.Presheaf

```

dagurtomas:
Nice, thanks! Actually the proof I wrote using `J.transitive` also works for that statement.

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -34,16 +34,62 @@ sufficiently small limits in the sheaf category on the essentially small site.
...
+variable (H : C ⥤ D) [ConcreteCategory A] {F G : Dᵒᵖ ⥤ A} (f : F ⟶ G)
+
+lemma isLocallyInjective_whisker_isCocontinuous [H.IsCocontinuous J K] [IsLocallyInjective K f] :
+    IsLocallyInjective J (whiskerLeft H.op f) where
+  equalizerSieve_mem x y h := H.cover_lift J K (equalizerSieve_mem K f x y h)
+
+lemma isLocallyInjective_of_whisker_coverPreserving_isCoverDense (hH : CoverPreserving J K H)
+    [H.IsCoverDense K] [IsLocallyInjective J (whiskerLeft H.op f)] : IsLocallyInjective K f where
+  equalizerSieve_mem {X} a b h := by
+    apply K.transitive (Functor.IsCoverDense.is_cover (G := H) (K := K) X.unop)
```
joelriou, line 58: Please use `H.is_cover_of_isCoverDense`. (In general, we should almost never use `TypeClass.field`, but API lemmas with suitable explicit variables and `[TypeClass]` as a parameter.)

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -34,16 +34,62 @@ sufficiently small limits in the sheaf category on the essentially small site.
...
 
 variable {C : Type*} [Category C] (J : GrothendieckTopology C)
 variable {D : Type*} [Category D] (K : GrothendieckTopology D) (e : C ≌ D)
 variable (A : Type*) [Category A]
 
+namespace Presheaf
+
+variable (H : C ⥤ D) [ConcreteCategory A] {F G : Dᵒᵖ ⥤ A} (f : F ⟶ G)
+
+lemma isLocallyInjective_whisker_isCocontinuous [H.IsCocontinuous J K] [IsLocallyInjective K f] :
```
joelriou, line 51: As `IsCocontinous` is the natural assumption, I do not think to include it in the name:
```suggestion
lemma isLocallyInjective_whisker [H.IsCocontinuous J K] [IsLocallyInjective K f] :
```

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -34,16 +34,62 @@ sufficiently small limits in the sheaf category on the essentially small site.
...
+    IsLocallyInjective J (whiskerLeft H.op f) where
+  equalizerSieve_mem x y h := H.cover_lift J K (equalizerSieve_mem K f x y h)
+
+lemma isLocallyInjective_of_whisker_coverPreserving_isCoverDense (hH : CoverPreserving J K H)
```
joelriou, line 55: Similarly:
```suggestion
lemma isLocallyInjective_of_whisker (hH : CoverPreserving J K H)
```
(And same suggestion for the next two lemmas.)

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -34,16 +34,62 @@ sufficiently small limits in the sheaf category on the essentially small site.
...
+    rw [← fac, Sieve.pullback_comp]
+    apply K.pullback_stable
+    have hh := hH.cover_preserve <|
+      imageSieve_mem J (whiskerLeft H.op f) ((forget A).map (G.map map.op) a)
+    refine K.superset_covering (Sieve.functorPullback_pushforward_le H _) ?_
+    refine K.superset_covering (Sieve.functorPushforward_monotone H _ ?_) hh
+    intro W q ⟨x, h⟩
+    simp only [Sieve.functorPullback_apply, Presieve.functorPullback_mem, Sieve.pullback_apply]
+    refine ⟨x, ?_⟩
+    simpa using h
```
joelriou, line 89:
```suggestion
    exact ⟨x, by simpa using h⟩
```

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -247,4 +382,18 @@ instance hasColimitsEssentiallySmallSite
...
+open ConcreteCategory
+
+variable [ConcreteCategory.{w} A]
+variable [HasFunctorialSurjectiveInjectiveFactorization A]
+  [((equivSmallModel C).locallyCoverDense J).inducedTopology.WEqualsLocallyBijective A]
+variable [((equivSmallModel C).locallyCoverDense J).inducedTopology.HasSheafCompose (forget A)]
+variable [Balanced (Sheaf ((equivSmallModel C).locallyCoverDense J).inducedTopology A)]
+variable {F G : Sheaf J A} (f : F ⟶ G)
+
+lemma Sheaf.isLocallySurjective_iff_epi_of_site_essentiallySmall :
```
joelriou, line 394: I would think that a better design would be to transport the `WEqualsLocallyBijective`, `HasSheafCompose` and `Balanced` instances via an equivalence. In the applications to specific sites, once these instances have been obtained, we may use `isLocallySurjective_iff_epi'`. This would also allow to use these instances for other purposes that just comparing locally surjective morphisms and epimorphisms. (This can be done in separate PRs.)

dagurtomas: You're right, I took the lazy approach.. I'll do it properly :)

joelriou: I know it is a little bit painful work, but I believe we have most of the ingredients.

dagurtomas: It wasn't too bad. However, using this approach one has to assume the existence of certain limits in the target category, so the "lazy" approach is slightly more general. I don't know if it's worth keeping that lemma though (without the limit assumption), because this only applies to concrete categories anyway.

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -34,16 +34,95 @@ sufficiently small limits in the sheaf category on the essentially small site.
...
 
 variable {C : Type*} [Category C] (J : GrothendieckTopology C)
 variable {D : Type*} [Category D] (K : GrothendieckTopology D) (e : C ≌ D)
 variable (A : Type*) [Category A]
 
+namespace Presheaf
+
+variable (H : C ⥤ D) {F G : Dᵒᵖ ⥤ A} (f : F ⟶ G)
+
+lemma _root_.CategoryTheory.GrothendieckTopology.W_iff_whisker
```
joelriou, line 51: For the transport of `WEqualsLocallyBijective`, I intended an approach where we do not rely on `HasWeakSheafify` and existence of limits, but on the very definition of `W`, taking into consideration that on both equivalent sites, we have equivalent categories of sheaves in order to test the `W` condition.

Mathlib/CategoryTheory/Sites/Equivalence.lean
joelriou: In order to proceed next with this material, I would suggest that for this PR, you keep only the general lemmas `isLocallySurjective/Injective_{/of/iff}_whisker` under quite general conditions (cocontinuous, etc) -- and these should be moved to another (maybe new) file. You may also keep the verification that equivalences are co/continuous in this PR, but the rest should very much be in differents PRs (specifically https://github.com/leanprover-community/mathlib4/pull/13486/files#diff-fd03ab2b884c2d15f411302648dd33c4f1e0db0c79f4ccaf8824354f59815bffR342-R378 and https://github.com/leanprover-community/mathlib4/pull/13486/files#diff-fd03ab2b884c2d15f411302648dd33c4f1e0db0c79f4ccaf8824354f59815bffR417-R446); actually, part of this material, especially the first chunk, will eventually be uncessary: suppose I have a site `(C,J)` which is big, and an equivalent one `(D,K)` which is small, we transport the necessary instances from `(D,K)` to `(C,J)` one by one, and then we may apply `isLocallySurjective_iff_epi'` in `(C,J)`. I think this is what you intend to do for the light profinite site, and this is what I intend to do in the future when the categories `C` and `D` are not necessarily equivalent, but at least they would have equivalent categories of sheaves (e.g. `C` is a category of schemes, and `D` is a (small) full subcategory of affines schemes).

Then, after this first PR is merged, we may proceed with transport lemmas, separately for `WEqualsLocallyBijective` (for an equivalence, I do not think we need more `HasLimitsOfShape` assumptions), `HasSheafCompose` and `Balanced`. It will be easier to work on this and review it as separate PRs.

dagurtomas:
All that's left in this PR is now some instances I added in the equivalence file, the preserving and reflecting of local injectivity and surjectivity is #13621

dagurtomas:
This PR has changed a bit, lots of the initial material has been incorporated into other PRs that have now been merged, but what's left now is still used in #13495 and #14027, so I would appreciate a second round of reviews

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -252,8 +306,8 @@ instance hasColimitsEssentiallySmallSite
 namespace GrothendieckTopology
 
 variable {A}
-variable [G.IsCoverDense J] [Functor.IsContinuous G K J] [G.Full]
-  [(G.sheafPushforwardContinuous A K J).EssSurj]
+variable [G.IsCoverDense J] [Functor.IsContinuous.{v₃} G K J]
+  [G.Full][(G.sheafPushforwardContinuous A K J).EssSurj]
```
joelriou, line 310:
```suggestion
  [G.Full] [(G.sheafPushforwardContinuous A K J).EssSurj]
```

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -64,22 +64,44 @@ theorem locallyCoverDense : LocallyCoverDense J e.inverse := by
       exact T.val.downward_closed hf _
     · simp
 
-theorem coverPreserving : CoverPreserving J (e.locallyCoverDense J).inducedTopology e.functor where
+theorem functorPushforward_eq_pullback {U : C} (S : Sieve U) :
```
joelriou, line 67: I would think that this lemma (and the next) should be moved to a more basic file. Otherwise, the PR looks good to me.

Mathlib/CategoryTheory/Sites/Sieves.lean
```diff
@@ -749,6 +749,42 @@ lemma mem_functorPushforward_inverse {X : D} {S : Sieve X} {e : C ≌ D} {f : Y
...
+  simp only [Sieve.functorPushforward_apply, Presieve.functorPushforward, exists_and_left,
+    Functor.id_obj, Functor.comp_obj, Sieve.pullback_apply]
+  constructor
+  · rintro ⟨W, g, hg, x, rfl⟩
+    rw [Category.assoc]
+    apply S.downward_closed
+    simpa using hg
+  · intro hf
+    exact ⟨_, e.unitInv.app Z ≫ f ≫ e.unitInv.app U, S.downward_closed hf _,
+      e.unit.app Z ≫ e.unit.app _, by simp⟩
```
erdOne, line 768:
```suggestion
      Sieve.pullback (e.unitInv.app U) S := by ext; simp
```

Mathlib/CategoryTheory/Sites/Sieves.lean
```diff
@@ -749,6 +749,42 @@ lemma mem_functorPushforward_inverse {X : D} {S : Sieve X} {e : C ≌ D} {f : Y
...
+  constructor
+  · intro ⟨W, g, hg, x, h⟩
+    simp only [Functor.comp_obj, Functor.id_obj, Iso.app_hom, Equivalence.inv_fun_map,
+      ← Category.assoc] at h
+    change _ ≫ (e.unitIso.app _).hom = _ ≫ (e.unitIso.app _).hom  at h
+    rw [Iso.cancel_iso_hom_right] at h
+    rw [h]
+    exact S.downward_closed hg _
+  · intro hf
+    exact ⟨_, e.unitInv.app Z ≫ f, S.downward_closed hf _, e.unit.app Z ≫ e.unit.app _, by simp⟩
```
erdOne, line 786:
```suggestion
      (Sieve.functorPushforward e.functor S)) = S := by ext; simp
```

Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -84,14 +81,26 @@ variable [e.TransportsGrothendieckTopology J K]
...
+    ext X S
+    change _ ↔ _ ∈ J.sieves _
+    simp only [symm_inverse]
+    constructor
+    · intro h
+      convert J.pullback_stable (e.unitInv.app X) h
+      exact Sieve.functorPushforward_equivalence_eq_pullback e S
+    · intro h
+      convert J.pullback_stable (e.unit.app X) h
+      exact (Sieve.pullback_functorPushforward_equivalence_eq e S).symm
```
erdOne, line 100:
```suggestion
    change _ ↔ _ ∈ J.sieves _
    simp [Sieve.functorPushforward_equivalence_eq_pullback]
```
We're probably missing simp lemmas at the `change`.

erdOne: Now that I think about it more, this is exactly the instance `instance : e.functor.IsDenseSubsite J (e.inverse.inducedTopology J)` a few line prior right? Maybe we could make `TransportsGrothendieckTopology` an abbrev or ditch it completely.

mattrobball:
!bench

Mathlib/CategoryTheory/Sites/Equivalence.lean
dagurtomas: You're right, it was very easy to just remove this. Let's see if CI finds errors elsewhere, but I think I fixed them all

joelriou:
Thanks!

bors merge

PR #11536, opened 2024-03-20
[Merged by Bors] - feat(Data/Matrix/RowCol): generalise the `Unique` indexing type in `Matrix.row` and `Matrix.col`

joneugster:
Changing `Matrix.row` and `Matrix.col` to use arbitrary `Unique` type for indexing the single column/row.

---

Compare to #11522, which presents a simpler possible solution.

For more details, see the other PR.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Why.20does.20.60Matrix.2Erow.60.20use.20.20.60Unit.60.20instead.20of.20.60Fin.201.60.3F)
 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13487, opened 2024-06-03
[Merged by Bors] - feat(NumberTheory/SiegelsLemma): Siegel's lemma for small integral non-trivial solutions of linear systems

fbarroero:
This PR contains the proof of [Siegel's Lemma](https://en.wikipedia.org/wiki/Siegel%27s_lemma).
This is a fundamental tool in diophantine approximation and transcendency and says that there exists a "small" integral
non-zero solution of a non-trivial underdetermined system of linear equations with integer coefficients.

theorem `exists_ne_zero_int_vec_norm_le`: Given a non-singular `m × n` matrix `A` with `m < n` the linear
system it determines has a non-zero integer solution `t` with
`‖t‖ ≤ ((n * ‖A‖) ^ ((m : ℝ) / (n - m)))`

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Siegel's.20Lemma)

Co-authored-by: Laura Capuano and Amos Turchet @amosturchet 

---


faenuccio:
Does this PR depend on the previous one (in the sense that it is branched from the other, or that it needs the other to be mergeable)?

fbarroero:
> Does this PR depend on the previous one (in the sense that it is branched from the other, or that it needs the other to be mergeable)?

No, it is independent.

faenuccio:
Really? But the file `Mathlib/NumberTheory/SiegelsLemma.lean` is introduced both in this PR and in #13470 . Assuming #13470 gets merged as it is, you do not want to re-modify lines 1-50, you will want to build upon the outcome of that PR. No?

fbarroero:
Yes, I just thought once #13470 is merged I would get a conflict with this and then fix it, or the other way around if this PR is merged first. I am new to this so I do not know what is the best way to proceed. If you think it is better to branch from #13470 I will just get rid of this one and make a new PR.

faenuccio:
No, one simply needs to add (as I just did) this dependency in the main message.

fbarroero:
Thanks Filippo!

docs/references.bib
```diff
@@ -1722,7 +1722,18 @@ @Article{         Higman52
...
+   title         = {Diophantine geometry},
+   series        = {Graduate Texts in Mathematics},
+   volume        = {201},
+   note          = {An introduction},
+   publisher     = {Springer-Verlag, New York},
+   year          = {2000},
+   pages         = {xiv+558}
+ }
+
+ @Unpublished{     hochsterunpublished,
```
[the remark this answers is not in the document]
riccardobrasca: This is probably a bug of the bot or something. You have to put the item in alphabetical order manually I think.

Mathlib.lean
```diff
@@ -3212,6 +3212,7 @@ import Mathlib.NumberTheory.Primorial
 import Mathlib.NumberTheory.PythagoreanTriples
 import Mathlib.NumberTheory.RamificationInertia
 import Mathlib.NumberTheory.Rayleigh
+import Mathlib.NumberTheory.SiegelsLemma
```
riccardobrasca, line 3215: I think `Siegel` is a better name for the file (judging on other names in mathlib), but this is not very important.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,242 @@
...
+attribute [local instance] Matrix.seminormedAddCommGroup
+
+namespace Matrix
+
+open Matrix Finset
+
+variable {m n : Type*} [Fintype m] [Fintype n]
+
+/-- The norm of an integral matrix as the sup of the sup of the natAbs of the entries -/
+def natNorm (A : Matrix m n ℤ) : ℕ := sup univ fun i ↦ sup univ fun j ↦ (A i j).natAbs
```
riccardobrasca, line 43: Is it really too painful to use the "standard" norm?

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,242 @@
...
+      _  = ↑((B + 1) ^ n) := by
+        rw [mul_comm, pow_mul_pow_sub]
+        simp only [cast_pow, cast_add, cast_one]
+        exact (le_of_lt hn)
+
+--end (2)
+
+end preparation
+
+theorem exists_ne_zero_int_vec_norm_le  (hA_nezero : A ≠ 0)  : ∃ (t : Fin n → ℤ), t ≠ 0 ∧
```
riccardobrasca, line 208: We are in the root namespace, right? I guess `Matrix` or something would be better.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,242 @@
...
+      rw [← Int.cast_add, Int.cast_le]
+      apply le_trans (hyT.2 i)
+      simp only [le_add_iff_nonneg_left]
+      exact hxT.1 i
+    · simp only [ tsub_le_iff_right]
+      rw [← Int.cast_add,Int.cast_le]
+      apply le_trans (hxT.2 i)
+      simp only [tsub_le_iff_right, le_add_iff_nonneg_right]
+      exact hyT.1 i
+  exact le_trans this (Nat.floor_le n_mul_norm_A_pow_e_nonneg)
```
riccardobrasca, line 242: Can you use `refine ... ?_ ...`, where the `?_` corresponds to `this` here? It would make the proof nicer.

Mathlib.lean
fbarroero: The guy proved so many very important results that I felt it was not really fair to call `Siegel` a file with only this small (but fundamental) result.

I wonder if we should create a `DiophantineApproximation` folder inside `NumberTheory` and collect this file (which could then be called `Siegel` without shame), the files `DiophantineApproximation` and `WellApproximable` and the folder `Liouville` (I might be missing something) and put them in there.

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: Done!

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: @faenuccio.. People seem not to like this `natNorm`.
After some thought I think I can get rid of it, and work with $\|| A \||$ only. It is going to be quite a bit of work and I think the code is going to look uglier, or maybe not... who knows...
What do you say? Should I try to do it?

Mathlib.lean
riccardobrasca: Yes, a folder is certainly a good idea (and you're right, a file called Siegel in `NumberTheory` is not very good...).

Mathlib/NumberTheory/SiegelsLemma.lean
riccardobrasca: The "normal" norm is a real number?

fbarroero: Yes

Mathlib.lean
fbarroero: Should I open a discussion on zulip?

Mathlib/NumberTheory/SiegelsLemma.lean
faenuccio: I think we will have the good answer only after some trials. In case you end up duplicating a lot of code (especially to prove that this norm is actually integral/positive) then it will be worth it create something, but this will become clearer only after some trial-and-error. You can start without and see how things go.

Mathlib.lean
faenuccio: I think you can freely create a folder called DiophantineApproximation, but on the other hand I believe that the lemma should not be called Siegel's lemma (the file could, if needed). It would be better if the name "tells" what is going on.

Mathlib/NumberTheory/SiegelsLemma.lean
riccardobrasca: Let me see if I can avoid it.

fbarroero: I think one can make a lemma that extracts the indices `i` and `j` that realise the sups in the norm and then work with that. It would be something like $\exists i, j :  \|| A \|| = \|| A_{ij}\||$. The RHS is by definition the cast of a `natAbs (A i j)`.

riccardobrasca: The real question is if you think this is a useful notion for others results (I guess it is?): in that case it is worth to create a new file, write the def and develop a basic API (even just the lemma you already have).

fbarroero: Let me think about it. I'll try a couple of things

faenuccio:
OK, let us know when you want a look again!

fbarroero:
> OK, let us know when you want a look again!

So I am done. I got rid of natNorm. It turns out it is not needed. One can prove the inequality of cardinalities using the real number $ \|| A \|| $.

While I was experimenting I wrote the lemma `exists_entry_norm_eq_norm_Matrix` which I think it could be useful in general . If you agree I can make another PR and add the lemma to 'Analysis/Matrix' . Otherwise I will just get rid of it.

It would be great if you guys @faenuccio and @riccardobrasca could take a look now. Thanks in advance!

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+
+lemma exists_entry_norm_eq_norm_Matrix [Nonempty m] [Nonempty n] (A : Matrix m n α) :
+    ∃ (i : m) (j : n), ‖A‖ = ‖A i j‖₊ := by
+  simp only [norm_eq_sup_sup_nnnorm]
+  rcases Finset.exists_mem_eq_sup univ univ_nonempty (fun i => univ.sup fun j => ‖A i j‖₊)
+    with ⟨i, hi⟩
+  use i
+  rcases Finset.exists_mem_eq_sup univ univ_nonempty (fun j => ‖A i j‖₊) with ⟨j, hj⟩
+  use j
+  rw [hi.2, hj.2]
```
faenuccio, line 51:
```suggestion
  simp only [norm_eq_sup_sup_nnnorm]
  obtain ⟨i, hi⟩ := Finset.exists_mem_eq_sup _ univ_nonempty (univ.sup fun j => ‖A · j‖₊)
  obtain ⟨j, hj⟩ := Finset.exists_mem_eq_sup _ univ_nonempty (‖A i ·‖₊)
  exact ⟨i, j, by rw [hi.2, hj.2]⟩
```
and yes, please move this lemma out, to where you suggested to. Than we can revise the whole Siegel part. Thanks!

MichaelStollBayreuth: I think Mathlib style is to use `↦` (\mapsto) instead of `=>` with `fun`?

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+# Step 1: ∀ v ∈  T, (A.mulVec v) ∈  S
+and
+# Step 2: S.card < T.card
+Pigeohole will give different x and y in T with A.mulVec x = A.mulVec y in S
+Their difference is the solution we are looking for
+-/
+
+-- # Step 1: ∀ v ∈  T, (A.mulVec v) ∈  S
+
+private lemma image_T_subset_S : ∀ v ∈ T, (A.mulVec v) ∈ S := by
```
MichaelStollBayreuth, line 85:
```suggestion
private lemma image_T_subset_S : ∀ v ∈ T, A.mulVec v ∈ S := by
```
There are many unnecessary parentheses. I'll flag some of them, but it is best if you try for yourself which ones are really needed.
Also, you can use notation `A *ᵥ v` for the matrix-vector product.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  intro v hv
+  rw [mem_Icc] at hv
+  rw [mem_Icc]
+  have mulVec_def : A.mulVec v =
+      fun i => Finset.sum univ fun (j : (Fin n)) => A i j * v j := by rfl
```
MichaelStollBayreuth, line 90:
```suggestion
      fun i ↦ Finset.sum univ fun j : Fin n ↦ A i j * v j := by rfl
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  rw [mulVec_def] -- unfold def of MulVec
+  refine ⟨fun i ↦ ?_, fun i ↦ ?_⟩ -- 2 goals
+  all_goals simp only [mul_neg]
+  all_goals gcongr (∑ _ : Fin n, ?_) with j _ -- get rid of sums
+  all_goals rw [← mul_comm (v j)] -- move A i j to the right of the products
+  all_goals by_cases hsign : 0 ≤ A i j   --we have to distinguish cases: we have now 4 goals
```
MichaelStollBayreuth, line 96:
```suggestion
  all_goals
    simp only [mul_neg]
    gcongr (∑ _ : Fin n, ?_) with j _ -- get rid of sums
    rw [← mul_comm (v j)] -- move A i j to the right of the products
    by_cases hsign : 0 ≤ A i j -- we have to distinguish cases: we have now 4 goals
```
I think this is a bit clearer (but I don't think there are clear style preferences on this).

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+    exact mul_nonneg (hv.1 j) hsign
+  · rw [not_le] at hsign
+    rw [negPart_eq_neg.2 (le_of_lt hsign)]
+    simp only [mul_neg, neg_neg]
+    exact mul_le_mul_of_nonpos_right (hv.2 j) (le_of_lt hsign)
+  · rw [posPart_eq_self.2  hsign]
+    exact mul_le_mul_of_nonneg_right (hv.2 j) hsign
+  · rw [not_le] at hsign
+    rw [posPart_eq_zero.2 (le_of_lt hsign)]
+    exact mul_nonpos_of_nonneg_of_nonpos (hv.1 j) (le_of_lt hsign)
```
MichaelStollBayreuth, line 107: Together wthe the `by_cases` line, this can be simplified a bit:
```lean
    rcases le_total 0 (A i j) with hsign | hsign -- we have to distinguish cases: we have now 4 goals
  · rw [negPart_eq_zero.2 hsign]
    exact mul_nonneg (hv.1 j) hsign
  · rw [negPart_eq_neg.2 hsign]
    simp only [mul_neg, neg_neg]
    exact mul_le_mul_of_nonpos_right (hv.2 j) hsign
  · rw [posPart_eq_self.2  hsign]
    exact mul_le_mul_of_nonneg_right (hv.2 j) hsign
  · rw [posPart_eq_zero.2 hsign]
    exact mul_nonpos_of_nonneg_of_nonpos (hv.1 j) hsign
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+
+--# Preparation for Step 2
+
+private lemma card_T_eq : (T).card = (B+1)^n := by
+   rw [Pi.card_Icc 0 B']
+   simp only [Pi.zero_apply, Int.card_Icc, sub_zero, Int.toNat_ofNat_add_one, prod_const, card_fin]
+
+variable  (hA : A ≠ 0 )
```
MichaelStollBayreuth, line 115:
```suggestion
variable (hA : A ≠ 0)
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+
+--This lemma is necessary to be able to apply the formula (Icc a b).card = b + 1 - a
+private lemma N_le_P_add_one : ∀ i : Fin m, N i ≤ P i + 1 := by
+  intro i
```
MichaelStollBayreuth, line 119:
```suggestion
private lemma N_le_P_add_one (i : Fin m) : N i ≤ P i + 1 := by
```
The preference is to put universally quantified variables and hypotheses left of the `:`; then one does not need to introduce them at the beginning of the proof.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+              apply Finset.sum_nonpos
+              intro j _
+              simp only [mul_neg, Left.neg_nonpos_iff]
+              exact mul_nonneg (Nat.cast_nonneg B) (negPart_nonneg (A i j))
+         _ ≤ P i + 1 := by
+              apply le_trans (Finset.sum_nonneg _) (Int.le_add_one (le_refl P i))
+              intro j _
+              exact mul_nonneg (Nat.cast_nonneg B) (posPart_nonneg (A i j))
+
+private lemma card_S_eq : (Finset.Icc N P).card = (∏ i : Fin m, (P i - N i + 1)):= by
```
MichaelStollBayreuth, line 130:
```suggestion
private lemma card_S_eq : (Finset.Icc N P).card = ∏ i : Fin m, (P i - N i + 1) := by
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  rw [Pi.card_Icc N P, Nat.cast_prod]
+  congr
+  ext i
+  rw [Int.card_Icc_of_le (N i) (P i) (N_le_P_add_one m n A i)]
+  exact add_sub_right_comm (P i) 1 (N i)
+
+/-- The sup norm of a non-zero integer matrix is at lest one  -/
+lemma one_le_norm_A_of_ne_zero : 1 ≤ ‖A‖ := by
+  by_contra h
+  apply lt_of_not_le at h
```
MichaelStollBayreuth, line 140:
```suggestion
  by_contra! h
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  apply hA
+  ext i j
+  simp only [zero_apply]
+  rw [norm_lt_iff Real.zero_lt_one] at h
+  specialize h i j
+  rw [Int.norm_eq_abs] at h
+  norm_cast at h
+  apply (Int.abs_lt_one_iff.1 h)
```
MichaelStollBayreuth, line 148:
```suggestion
  exact Int.abs_lt_one_iff.1 h
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+
+-- # Step 2: S.card < T.card
+
+open Real Nat
+
+private lemma card_S_lt_card_T : (S).card < (T).card := by
```
MichaelStollBayreuth, line 154:
```suggestion
private lemma card_S_lt_card_T : S.card < T.card := by
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+local notation3 "e" => m / ((n : ℝ) - m) --exponent
+local notation3 "B" => Nat.floor (((n : ℝ) * ‖A‖) ^ e)
+-- B' is the vector with all components = B
+local notation3 "B'" => fun _ : Fin n => (B : ℤ)
+-- T is the box [0 B]^n
+local notation3 "T" =>  Finset.Icc 0 B'
+local notation3  "P" => fun i : Fin m => (∑ j : Fin n, B * posPart (A i j))
+local notation3  "N" => fun i : Fin m =>  (∑ j : Fin n, B * (- negPart (A i j)))
+-- S is the box where the image of T goes
+local notation3  "S" => Finset.Icc N P
```
MichaelStollBayreuth, line 71:
```suggestion
local notation3 "P" => fun i : Fin m ↦ ∑ j : Fin n, B * posPart (A i j)
local notation3 "N" => fun i : Fin m ↦ ∑ j : Fin n, B * (- negPart (A i j))
-- S is the box where the image of T goes
local notation3 "S" => Finset.Icc N P
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  zify --this is necessary to use card_S_eq
+  rw [card_T_eq, card_S_eq]
+  rify --this is necessary because ‖A‖ is a real number
+  calc
+  ∏ x : Fin m, (∑ x_1 : Fin n, ↑B * ↑(A x x_1)⁺ - ∑ x_1 : Fin n, ↑B * -↑(A x x_1)⁻ + 1)
+    ≤ (n * ‖A‖ * B + 1) ^ m := by
+      rw [← Fin.prod_const m (n * ‖A‖ * B + 1)] --transform ^m into a product
+      apply Finset.prod_le_prod --two goals
+      all_goals intro i _
+      · obtain h:= N_le_P_add_one m n A i
```
MichaelStollBayreuth, line 164:
```suggestion
      · have h := N_le_P_add_one m n A i
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+        rify at h
+        linarith only [h]
+      · simp only [mul_neg, sum_neg_distrib, sub_neg_eq_add, add_le_add_iff_right]
+        have h1 : n * ‖A‖ * B = ∑ _ : Fin n, ‖A‖ * B := by
+          simp only [sum_const, card_univ, Fintype.card_fin, nsmul_eq_mul, (mul_assoc ↑n ‖A‖ ↑B)]
```
MichaelStollBayreuth, line 169:
```suggestion
          simp only [sum_const, card_univ, Fintype.card_fin, nsmul_eq_mul, mul_assoc ↑n ‖A‖ ↑B]
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+        rw [h1, ← Finset.sum_add_distrib]
+        gcongr with j _
+        rw [← mul_add, ← Int.cast_add, posPart_add_negPart (A i j), mul_comm]
+        apply mul_le_mul_of_nonneg_right _ (Nat.cast_nonneg B)
+        rw [Int.cast_abs]
+        exact norm_entry_le_entrywise_sup_norm A
+  _  ≤ (n * ‖A‖) ^ m * (B + 1) ^ m := by
+        rw [← mul_pow]
+        apply pow_le_pow_left (add_nonneg _ (zero_le_one' ℝ)) _ --two goals
+        · exact (mul_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) (Nat.cast_nonneg B))
```
MichaelStollBayreuth, line 179:
```suggestion
        · exact mul_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) (Nat.cast_nonneg B)
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+        · rw [mul_add]
+          simp only [mul_one, add_le_add_iff_left]
+          exact Left.one_le_mul_of_le_of_le
+              (by exact_mod_cast Nat.one_le_of_lt hn) (one_le_norm_A_of_ne_zero m n A hA)
+              (Nat.cast_nonneg n)
+  _ = ((n * ‖A‖) ^ ((m / ((n : ℝ) - m)))) ^ ((n : ℝ) - m)  * (B + 1) ^ m := by
```
MichaelStollBayreuth, line 185:
```suggestion
  _ = ((n * ‖A‖) ^ ((m / ((n : ℝ) - m)))) ^ ((n : ℝ) - m) * (B + 1) ^ m := by
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+        rw [mul_left_inj' (pow_ne_zero m (Nat.cast_add_one_ne_zero B)),
+          ← rpow_mul (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)), ← Real.rpow_natCast]
+        congr
+        rw [div_mul_cancel₀]
+        apply sub_ne_zero_of_ne
+        exact_mod_cast Nat.ne_of_lt' hn
```
MichaelStollBayreuth, line 191:
```suggestion
          ← rpow_mul (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)), ← Real.rpow_natCast, div_mul_cancel₀]
        exact sub_ne_zero_of_ne (mod_cast hn.ne')
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  _ < (B + 1) ^ ((n : ℝ) - m) * (B + 1) ^ m := by
+        rw [mul_lt_mul_right (pow_pos (Nat.cast_add_one_pos B) m)]
+        apply rpow_lt_rpow --three goals
+        · exact rpow_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) e
+        · exact Nat.lt_floor_add_one ((n * ‖A‖) ^ e)
+        · simp only [sub_pos, Nat.cast_lt]
+          exact hn
```
MichaelStollBayreuth, line 198:
```suggestion
        · simpa only [sub_pos, cast_lt] using hn
```
BTW, `Nat.cast_nonneg n` can be written as `n.cast_nonneg` etc. ("dot notation").

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  _ = (B + 1) ^ n := by
+        rw [← rpow_natCast, ← rpow_add (Nat.cast_add_one_pos B), ← rpow_natCast, sub_add_cancel]
+
+end preparation
+
+theorem exists_ne_zero_int_vec_norm_le  (hA_nezero : A ≠ 0)  : ∃ (t : Fin n → ℤ), t ≠ 0 ∧
```
MichaelStollBayreuth, line 204:
```suggestion
theorem exists_ne_zero_int_vec_norm_le (hA_nezero : A ≠ 0) : ∃  t : Fin n → ℤ, t ≠ 0 ∧
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+    A.mulVec t = 0 ∧ ‖t‖ ≤ ((n * ‖A‖)^((m : ℝ) / (n - m))) := by
```
MichaelStollBayreuth, line 205:
```suggestion
    A.mulVec t = 0 ∧ ‖t‖ ≤ ((n * ‖A‖) ^ ((m : ℝ) / (n - m))) := by
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  --Pigeonhole
+  rcases Finset.exists_ne_map_eq_of_card_lt_of_maps_to
+    (card_S_lt_card_T m n A hn hA_nezero) (image_T_subset_S m n A)
+    with ⟨x, hxT, y, hyT, hneq, hfeq⟩
+  use x-y
+  -- proofs that x - y ≠ 0 and x-y is a solution
+  refine ⟨sub_ne_zero.mpr hneq, (by simp only [mulVec_sub, sub_eq_zero, hfeq]), ?_⟩
```
MichaelStollBayreuth, line 212:
```suggestion
  -- proofs that x - y ≠ 0 and x-y is a solution
  refine ⟨x - y, sub_ne_zero.mpr hneq, by simp only [mulVec_sub, sub_eq_zero, hfeq], ?_⟩
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  ---Inequality
+  have n_mul_norm_A_pow_e_nonneg : 0 ≤ (n * ‖A‖) ^ e :=
+      Real.rpow_nonneg (mul_nonneg (Nat.cast_nonneg n) (le_of_lt (norm_pos_iff'.mpr hA_nezero))) e
```
MichaelStollBayreuth, line 215:
```suggestion
  have n_mul_norm_A_pow_e_nonneg : 0 ≤ (n * ‖A‖) ^ e := by positivity
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  rw [← norm_col, norm_le_iff n_mul_norm_A_pow_e_nonneg]
+  intro i j
+  simp only [col_apply, Pi.sub_apply, ge_iff_le]
```
MichaelStollBayreuth, line 218:
```suggestion
  simp only [col_apply, Pi.sub_apply]
```
`simp?` likes to include this even when it is not necessary.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,237 @@
...
+  rw [Int.norm_eq_abs, ← Int.cast_abs]
+  refine le_trans ?_ (Nat.floor_le n_mul_norm_A_pow_e_nonneg)
+  norm_cast
+  rw [abs_le]--two goals
```
MichaelStollBayreuth, line 222: It is still one goal at this point (a conjunction).

BTW, can you use a consistent style for the comments? (e.g., " -- <comment>")

MichaelStollBayreuth:
This looks quite good! My comments are mostly about minor stylistic points.
It wil be great to have Siegel's Lemma in Mathlib for further applications in Diophantine Approximation!

fbarroero:
> This looks quite good! My comments are mostly about minor stylistic points. It wil be great to have Siegel's Lemma in Mathlib for further applications in Diophantine Approximation!

Thanks for the feedback and the suggestions, Michael!
I implemented all the changes your suggested

riccardobrasca:
> > This looks quite good! My comments are mostly about minor stylistic points. It wil be great to have Siegel's Lemma in Mathlib for further applications in Diophantine Approximation!
> 
> Thanks for the feedback and the suggestions, Michael! I implemented all the changes your suggested

Can you please hit the "resolve" button on the comment you've adressed?

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,215 @@
...
+  --Pigeonhole
+  rcases Finset.exists_ne_map_eq_of_card_lt_of_maps_to
+    (card_S_lt_card_T m n A hn hA_nezero) (image_T_subset_S m n A)
+    with ⟨x, hxT, y, hyT, hneq, hfeq⟩
+  --proofs that x - y ≠ 0 and x-y is a solution
+  refine ⟨x - y, sub_ne_zero.mpr hneq, by simp only [mulVec_sub, sub_eq_zero, hfeq], ?_⟩
+  --Inequality
+  have n_mul_norm_A_pow_e_nonneg : 0 ≤ (n * ‖A‖) ^ e := by positivity
+  have : ‖x - y‖ = ‖col (Fin 1) (x - y)‖ := by exact Eq.symm (norm_col (x - y))
+  rw [this, norm_le_iff n_mul_norm_A_pow_e_nonneg]
```
fbarroero, line 194: Earlier today, when I updated Mathlib, the proof broke here.
Instead of `this` there used to be `← norm_col` but it gave the error 
`typeclass instance problem is stuck, it is often due to metavariables
  Unique ?m.174974`
I fixed it in that way but I imagine there is a better way.

riccardobrasca: You can use `← norm_col (ι := Unit)`. The statement of `norm_col` changed in #11536.

fbarroero: Thanks!

Mathlib/NumberTheory/SiegelsLemma.lean
MichaelStollBayreuth: There are still extra spaces (before `"P"`, `"N"`, `"S"` and the last `∑`.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+  rw [← norm_col (ι := Unit), norm_le_iff n_mul_norm_A_pow_e_nonneg]
+  intro i j
+  simp only [col_apply, Pi.sub_apply]
+  rw [Int.norm_eq_abs, ← Int.cast_abs]
+  refine le_trans ?_ (Nat.floor_le n_mul_norm_A_pow_e_nonneg)
+  norm_cast
+  rw [abs_le]
+  rw [Finset.mem_Icc] at hyT
+  rw [Finset.mem_Icc] at hxT
+  constructor--two goals
```
MichaelStollBayreuth, line 202:
```suggestion
  constructor --two goals
```

Mathlib/NumberTheory/SiegelsLemma.lean
fbarroero: Sorry, I missed that. It should be ok now

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+        · exact rpow_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) e
+        · exact Nat.lt_floor_add_one ((n * ‖A‖) ^ e)
+        · simpa only [sub_pos, Nat.cast_lt] using hn
+  _ = (B + 1) ^ n := by
+        rw [← rpow_natCast, ← rpow_add (Nat.cast_add_one_pos B), ← rpow_natCast, sub_add_cancel]
+
+end preparation
+
+theorem exists_ne_zero_int_vec_norm_le  (hA_nezero : A ≠ 0) : ∃ t : Fin n → ℤ, t ≠ 0 ∧
+    A *ᵥ t = 0 ∧ ‖t‖ ≤ ((n * ‖A‖) ^ ((m : ℝ) / (n - m))) := by
```
MichaelStollBayreuth, line 184:
```suggestion
    A *ᵥ t = 0 ∧ ‖t‖ ≤ (n * ‖A‖) ^ ((m : ℝ) / (n - m)) := by
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+        rw [mul_left_inj' (pow_ne_zero m (Nat.cast_add_one_ne_zero B)),
+          ← rpow_mul (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)), ← Real.rpow_natCast,
+          div_mul_cancel₀]
+        exact sub_ne_zero_of_ne (mod_cast hn.ne')
+  _ < (B + 1) ^ ((n : ℝ) - m) * (B + 1) ^ m := by
+        rw [mul_lt_mul_right (pow_pos (Nat.cast_add_one_pos B) m)]
+        apply rpow_lt_rpow --three goals
+        · exact rpow_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) e
+        · exact Nat.lt_floor_add_one ((n * ‖A‖) ^ e)
+        · simpa only [sub_pos, Nat.cast_lt] using hn
```
MichaelStollBayreuth, line 177:
```suggestion
        gcongr
        · exact sub_pos.mpr (mod_cast hn)
        · exact Nat.lt_floor_add_one ((n * ‖A‖) ^ e)
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+        · exact mul_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) (Nat.cast_nonneg B)
+        · rw [mul_add]
+          simp only [mul_one, add_le_add_iff_left]
+          exact Left.one_le_mul_of_le_of_le
+              (by exact_mod_cast Nat.one_le_of_lt hn) (one_le_norm_A_of_ne_zero m n A hA)
+              (Nat.cast_nonneg n)
+  _ = ((n * ‖A‖) ^ (m / ((n : ℝ) - m))) ^ ((n : ℝ) - m)  * (B + 1) ^ m := by
+        rw [mul_left_inj' (pow_ne_zero m (Nat.cast_add_one_ne_zero B)),
+          ← rpow_mul (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)), ← Real.rpow_natCast,
+          div_mul_cancel₀]
```
MichaelStollBayreuth, line 170:
```suggestion
        congr 1
        rw [← rpow_mul (mul_nonneg n.cast_nonneg (norm_nonneg A)), ← Real.rpow_natCast, div_mul_cancel₀]
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+        exact norm_entry_le_entrywise_sup_norm A
+  _  ≤ (n * ‖A‖) ^ m * (B + 1) ^ m := by
+        rw [← mul_pow]
+        apply pow_le_pow_left (add_nonneg _ (zero_le_one' ℝ)) _ --two goals
+        · exact mul_nonneg (mul_nonneg (Nat.cast_nonneg n) (norm_nonneg A)) (Nat.cast_nonneg B)
+        · rw [mul_add]
+          simp only [mul_one, add_le_add_iff_left]
+          exact Left.one_le_mul_of_le_of_le
+              (by exact_mod_cast Nat.one_le_of_lt hn) (one_le_norm_A_of_ne_zero m n A hA)
+              (Nat.cast_nonneg n)
```
MichaelStollBayreuth, line 166:
```suggestion
        rw [← mul_pow, mul_add, mul_one]
        gcongr
        have H : 1 ≤ (n : ℝ) := mod_cast (hm.trans hn)
        exact one_le_mul_of_one_le_of_one_le H <| one_le_norm_A_of_ne_zero m n A hA
```

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+private lemma card_S_lt_card_T : (S).card < (T).card := by
+  zify --this is necessary to use card_S_eq
+  rw [card_T_eq, card_S_eq]
+  rify --this is necessary because ‖A‖ is a real number
+  calc
+  ∏ x : Fin m, (∑ x_1 : Fin n, ↑B * ↑(A x x_1)⁺ - ∑ x_1 : Fin n, ↑B * -↑(A x x_1)⁻ + 1)
+    ≤ (n * ‖A‖ * B + 1) ^ m := by
+      rw [← Fin.prod_const m (n * ‖A‖ * B + 1)] --transform ^m into a product
+      apply Finset.prod_le_prod --two goals
+      all_goals intro i _
```
MichaelStollBayreuth, line 145:
```suggestion
      refine Finset.prod_le_prod (fun i _ ↦ ?_) (fun i _ ↦ ?_)
```
I prefer this, but it is a matter of taste.

Mathlib/NumberTheory/SiegelsLemma.lean
```diff
@@ -0,0 +1,214 @@
...
+        linarith only [h]
+      · simp only [mul_neg, sum_neg_distrib, sub_neg_eq_add, add_le_add_iff_right]
+        have h1 : n * ‖A‖ * B = ∑ _ : Fin n, ‖A‖ * B := by
+          simp only [sum_const, card_univ, Fintype.card_fin, nsmul_eq_mul, mul_assoc ↑n ‖A‖ ↑B]
+        rw [h1, ← Finset.sum_add_distrib]
+        gcongr with j _
+        rw [← mul_add, ← Int.cast_add, posPart_add_negPart (A i j), mul_comm]
+        apply mul_le_mul_of_nonneg_right _ (Nat.cast_nonneg B)
+        rw [Int.cast_abs]
+        exact norm_entry_le_entrywise_sup_norm A
```
MichaelStollBayreuth, line 157:
```suggestion
        simp_rw [h1, ← Finset.sum_add_distrib, ← mul_add, mul_comm ‖A‖, ← Int.cast_add]
        gcongr with j _
        rw [posPart_add_negPart (A i j), Int.cast_abs]
        exact norm_entry_le_entrywise_sup_norm A
```

MichaelStollBayreuth:
Sorry for the long line! I was using the web editor, which does not show the 100 character limit.

MichaelStollBayreuth (approved):
LGTM!

MichaelStollBayreuth:
maintainer merge

riccardobrasca:
Thanks!

bors merge

fbarroero:
> LGTM!

Thank you @MichaelStollBayreuth for everything. I learnt a lot from your comments!

Mathlib/NumberTheory/SiegelsLemma.lean
faenuccio: I see this comment has unfortunately not been taken into account. It is not a great deal, but apart from the golfing I believe this lemma should not stay there but be mobed to `Analysis.Matrix`.

faenuccio:
Some comments:
1. I have seen that my suggestion to (golf the proof of, and) move the first lemma towards `Analysis.Matrix` has not been taken into account. The golfing is not a big deal (although `fun a \mapsto a` is usually written `(\.)`) but the moving might be important.
2. I think you forgot an "n" in Pigeonhole
3. When you apply a `rw [foo]` to two things you can compact those with `rw [foo] at h1 h2` instead of using two lines.

But thanks for the work!

MichaelStollBayreuth:
@faenuccio I had not realized that the first lemma had not been moved before hitting "maintainer merge". Sorry about that! I can make a PR to fix this.

faenuccio:
> @faenuccio I had not realized that the first lemma had not been moved before hitting "maintainer merge". Sorry about that! I can make a PR to fix this.

Thanks! Then I would immediately hit `maintainer merge` 😄

MichaelStollBayreuth:
In fact, the lemma was just removed [here](https://github.com/leanprover-community/mathlib4/pull/13487/commits/28788163248f620e21287dd90de968b8707f8b25). @faenuccio Do you still think it should be added to `Analysis.Matrix`?

faenuccio:
Good catch, I had not noticed! Then no, I believe it is not needed (I just thought that if it were to appear somewhere, this was not the right place). So all is good @MichaelStollBayreuth , thanks!

PR #13489, opened 2024-06-04
[Merged by Bors] - chore: regularly update dependencies

kim-em:
If the update fails, it will leave the PR open, preventing the action from creating more such PRs.

Also adds a workflow action that reports on zulip if one of these PRs fails.

### DIFF
--- a/.github/workflows/update_dependencies.yml
+++ b/.github/workflows/update_dependencies.yml
@@ -24,8 +24,8 @@ jobs:
         run: |
           prs=$(gh pr list --search "author:leanprover-community-mathlib4-bot state:open" --json title,headRefName)
           for pr in $(echo "$prs" | jq -c '.[]'); do
-            pr_title=$(echo $pr | jq -r '.title')
-            pr_branch=$(echo $pr | jq -r '.headRefName')
+            pr_title=$(echo "$pr" | jq -r '.title')
+            pr_branch=$(echo "$pr" | jq -r '.headRefName')
             if [[ "$pr_title" == *"update Mathlib dependencies"* ]] || [[ "$pr_branch" == update-dependencies-* ]]; then
               echo "Existing PR found with title or branch name indicating a dependency update. Stopping."
               echo "existing_pr=true" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


YaelDillies (commented):
When does that usually happen? Every hour sounds like a lot often. Also something that's not clear from the PR name nor the code: This doesn't target `master` but `nightly-testing`, right? Can you make that clear?

semorrison:
This targets master. (I will fix the fact that it is using the NIGHTLY_TESTING token, which is wrong.)

This is usually run by me manually. It should be cheap to run, I don't see any downside to running it often.

adomani:
As always, testing scheduled CI actions is very tricky, so I'm in favour of merging and fixing whatever issue there might be when it surfaces.

maintainer merge

adomani:
maintainer merge

YaelDillies:
So it will roughly double the number of daily commits to master, right? Not sure of the implications, I just find it surprising that you would do that.

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,82 @@
+name: Update Mathlib Dependencies
+
+# This action currently uses the NIGHTLY_TESTING secret, but could be moved to a separate secret.
+# Despite this, this action is acting on the `master` branch, not the `nightly-testing` branch.
+
+on:
+  schedule:
+    - cron: '0 * * * *'  # This will run every hour
```
Parcly-Taxel, line 8: Yes. Why should this run every hour? Can't it be every week?

Parcly-Taxel (changes requested):

.github/workflows/update_dependencies.yml
Parcly-Taxel:
```suggestion
    - cron: '0 20 * * 5'  # This will run every week
```

jcommelin:
> So it will roughly double the number of daily commits to master, right? Not sure of the implications, I just find it surprising that you would do that.

But it will only commit something if there is an interesting update to make, right? If all dependencies are up to date, nothing happens. Or am I misunderstanding some of the conditionals in the script?

adomani:
Johan, that is my understanding as well.

YaelDillies:
> But it will only commit something if there is an interesting update to make, right? If all dependencies are up to date, nothing happens. Or am I misunderstanding some of the conditionals in the script?

Oh sure, it doesn't commit an empty update. But in general that means that roughly every commit to eg Batteries will get a corresponding update here. I guess that's not much of a problem since the scales of activity between the two repos are so different.

Anyway, I think my questions show that more comment would be welcome 😉

jcommelin:
Let's try this out in production.

bors merge

PR #13492, opened 2024-06-04
[Merged by Bors] - feat: The triple centralizer is equal to the centralizer

dupuisf:
This PR shows (1) that a set is included in its double centralizer, and (2) that the centralizer of a set is equal to its triple centralizer.

---


tb65536:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13493, opened 2024-06-04
[Merged by Bors] - chore(Order/Interval): use `WithBot` API

urkud:
---


eric-wieser:
bors merge

Thanks!

PR #13494, opened 2024-06-04
[Merged by Bors] - chore(Data/UInt): make instCommRing a def, and explain why

kim-em:
These instances are now in the `UIntX.CommRing` scope.

This is because the presence of these casting operations contradicts assumptions made by the
expression tree elaborator, namely that coercions do not form a cycle.
This happens in `Fin` and `BitVector` too, but those are problems for another PR.

After this change, `ring`, `norm_num`, and `rw [mul_add_one]` no longer work out of the box on these types.
To avoid further fallout (`rw [mul_one]`, `abel`, ...), `CommMonoid` and `NonUnitalCommRing` instances are provided, which fall just short of enabling the unwanted cast operators.

---

### DIFF
--- a/Mathlib/Data/UInt.lean
+++ b/Mathlib/Data/UInt.lean
@@ -12,13 +12,16 @@ import Mathlib.Data.ZMod.Defs
 /-!
 # Adds Mathlib specific instances to the `UIntX` data types.
 
-Users are advised not to use the `NatCast` and `IntCast` typeclasses, as their presence
-contradicts assumptions made by the expression tree elaborator.
-These are only available as scoped instances, requiring `open UInt32.Cast` to use.
+The `CommRing` instances (and the `NatCast` and `IntCast` instances from which they is built) are
+scoped in the `UIntX.CommRing` namespace, rather than available globally. As a result, the `ring`
+tactic will not work on `UIntX` types without `open scoped UIntX.Ring`.
 
-The `CommRing` instance is similarly scoped, and requires `open UInt32.CommRing` to use.
-(It is problematic for the same reason:
-making it available implicitly also makes the casting instances available.)
+This is because the presence of these casting operations contradicts assumptions made by the
+expression tree elaborator, namely that coercions do not form a cycle.
+
+This issue also arises for `CommRing (Fin n)`, but Mathlib has come to rely upon this. The UInt
+version also interferes more with software-verification use-cases, which is reason to be more
+cautious here.
 -/
 
 lemma UInt8.val_eq_of_lt {a : Nat} : a < UInt8.size → ((ofNat a).val : Nat) = a :=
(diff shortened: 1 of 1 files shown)


Mathlib/Data/UInt.lean
```diff
@@ -97,7 +97,14 @@ run_cmd
       @[simp] lemma mk_val_eq : ∀ (a : $typeName), mk a.val = a
       | ⟨_, _⟩ => rfl
 
-      instance : CommRing $typeName :=
+      /--
+      This is not an instance; it is not needed in Mathlib,
+      and moreover results in a `Nat → $typeName` coercion which breaks assumptions made in the
```
eric-wieser, line 102: This isn't true; the coercion is defined higher up in the file; this just proves it's lawful.

So if we go down this route there are three instances to disable for each type, not one. I think asking the user to refer to these individually by name is unreasonable, and so we should scope them.

eric-wieser (commented):
I'm a little wary of this change, even if adjusted to use a scope; the same problem exists for Fin and is far more prevalent, so what do we gain by introducing an inconsistency and fixing it only for these types?

eric-wieser:
I've pushed some changes, namely:
* Combining the `Cast` and `CommRing` scoped; I can't think of any reason you'd want one and and not both open
* Adjusting the docstring to mention `Fin`
* Adding docstrings to the three offending instances

If these look good to you, then

bors d+

Mathlib/Data/UInt.lean
```diff
@@ -9,6 +9,21 @@ import Mathlib.Algebra.GroupWithZero.Defs
...
+# Adds Mathlib specific instances to the `UIntX` data types.
+
+The `CommRing` instances (and the `NatCast` and `IntCast` instances from which they is built) are
+scoped in the `UIntX.CommRing` namespace, rather than available globally. As a result, the `ring`
+tactic will not work on `UIntX` types without `open scoped UIntX.Ring`.
+
+This is because the presence of these casting operations contradicts assumptions made by the
+expression tree elaborator, namely that coercions do not form a cycle.
+
+This issue also arises for `CommRing (Fin n)`, but Mathlib has come to rely upon this. The UInt
```
kmill, line 22:
```suggestion
The UInt
```

@eric-wieser The `Fin` problem is on my radar.

I don't see any inconsistency with fixing the problem for `UInt*` types and putting off fixing `Fin`. The `Fin` type is arguably mathematical, but the `UInt*` types clearly aren't (they're the integers modulo arbitrary powers of two).

Mentioning `Fin` at all just makes it more likely that we'll have an inconsistency in documentation once the `Fin` problem is fixed. It seems unlikely to me that anyone modifying `Fin` instances would think to look at `Mathlib.Data.UInt`'s module docs.

semorrison:
bors merge

eric-wieser:
Sorry, I thought of one more thing; we can safely add the instances which don't come with a cast, which I've done here

fpvandoorn:
@eric-wieser Can we merge this now?

eric-wieser:
Yes, sorry; I didn't merge myself to give Kim or Kyle time to object to the extra commit I pushed

semorrison:
bors merge

PR #13497, opened 2024-06-04
[Merged by Bors] - perf(Geometry/Euclidean): squeeze slow field_simp calls

grunweg:
Replace them by the equivalent `simp` call and squeeze that. We leave `field_simp` calls slower than roughly 150ms in place. A few calls cannot easily be replaced; we leave these alone as well.

Speeds up all affected files, including `Euclidean/Circumcenter` by 4%, `Inversion/Basic` by 27% and `Triangle` by 52%.

---

This is not a replacement for a field_simp rewrite, but leaving these really slow calls in place until then doesn't seem like an improvement either. I'm happy to file a tracking issue for this.


grunweg:
!bench

Mathlib/Geometry/Euclidean/Angle/Unoriented/Basic.lean
```diff
@@ -202,7 +202,8 @@ theorem sin_angle_mul_norm_mul_norm (x y : V) :
       rw [hx, inner_zero_left, zero_mul, neg_zero]
     · rw [norm_eq_zero] at hy
       rw [hy, inner_zero_right, zero_mul, neg_zero]
-  · field_simp [h]
+  · -- takes 500ms, but equivalent simp call doesn't recurse as much
+    field_simp [h]
```
semorrison, line 206: Could you expand the comment? It's not clear that you mean the equivalent simp call doesn't work!

semorrison:
Let's have a tracking issue, and link to it from all these places. I'm a bit afraid of scattering comments across so many files without a way to collect them all again when we fix the underlying problem.

grunweg:
:+1: Creating and referencing this tracking issue is on my list of "clean up loose ends once my PhD thesis is submitted" :-)

grunweg:
I have created #15486 and referenced it in all the comments. Two more slow calls can in fact replaced by simp now.

grunweg:
!bench

jcommelin:
Thanks :tada:

bors merge

PR #13498, opened 2024-06-04
[Merged by Bors] - fix(CI): make the `update-nolints` workflow work

grunweg:
- fetch the whole repository (with history) when checking it out
- fix a repeated typo mathlib -> mathlib4
- the workflow token has been switched to a bot account
- also silence a noisy `git fetch` step which was filling up the logs with unrelated branch names

---


bryangingechen:
OK, it looks like the workflow is able to successfully open PRs (and even approve them automatically with bors powers)! I *think* once we revert the "on: push" back to "schedule: cron", this is almost in good enough shape for merging. (All that is left is writing a new title / description, which I can do tomorrow, along with switching the token so that the PRs come from a bot account rather than mine.)

.github/workflows/nolints.yml
```diff
@@ -54,14 +59,15 @@ jobs:
             bash -o pipefail -c "env LEAN_ABORT_ON_PANIC=1 lake build --wfail -KCI"
 
       - name: update nolints.json and style-exceptions.txt
+        continue-on-error: true
```
bryangingechen, line 62: Actually, I'm not sure this is the best solution, as even if this allows the workflow to proceed, the job will get a red x and notify whoever modified the workflow last.  We could add `continue-on-error` to the job (as we do in the `bot_fix_style_*` workflows), but that will cover up errors in other steps of the workflow.

@grunweg Would it make sense to have `lint-style` not return an nonzero exit code when linting errors are detected when run in `--update` or `--fix` modes?

bryangingechen:
I have switched the token in the`UPDATE_NOLINTS_TOKEN` secret again. Note that the new token only has the `public_repo` scope. The token used earlier for testing here also had the "workflow" scope, which was necessary to push to this branch, which modifies workflow files. Because of this, further testing commits to this branch won't work, but the new token should work for scheduled update-nolints PRs (which should only ever touch `scripts/nolints.json` and `scripts/style-exceptions.txt` and in particular never need to touch any workflow files).

.github/workflows/nolints.yml
grunweg: I think it makes sense, yes: update or fix mode are *supposed* to fix errors (so returning zero if something changed seems morally fine); the lint-style check runs without these arguments, so is not affected

.github/workflows/nolints.yml
```diff
@@ -71,6 +76,7 @@ jobs:
           git config --unset http.https://github.com/.extraheader
 
       - name: file a new PR to update nolints.json and style-exceptions.txt
+        shell: bash
```
grunweg, line 79: Is this needed? (It probably doesn't hurt either, though.)

scripts/update_nolints_CI.sh
```diff
@@ -32,9 +32,9 @@ gh_api() {
     "https://api.github.com/$url" "$@"
 }
 
-git push "${remote_name}" "HEAD:$branch_name"
+git push -v "${remote_name}" "HEAD:$branch_name"
```
grunweg, line 35: Is this change needed? It probably doesn't hurt, though.

.github/workflows/nolints.yml
```diff
@@ -1,8 +1,10 @@
 name: update nolints
 
 on:
-  schedule:
-    - cron: "0 0 * * 0"
+  push:
```
grunweg, line 4: Reminder to change this before merging :-)

grunweg (commented):
Thanks a ton for your help. This basically looks good to me.

scripts/update_nolints_CI.sh
bryangingechen: Yeah, I think you may have added it for debugging purposes at the beginning of this branch. I'll clean it up in another commit. (I want to make the `git fetch` step less noisy since it's currently spamming the log with a line for every open branch in this repo)

.github/workflows/nolints.yml
bryangingechen: I've removed it to minimize the diff.

bryangingechen:
Thanks @grunweg for your effort on this. I think this PR will be ready as soon as #16301 is in and we fix the trivial merge conflict that might pop up here. (I did notice that I forgot to give write access to the bot account, but that is done now too.)

jcommelin (commented):
Thanks :tada:

bors merge

PR #13502, opened 2024-06-04
[Merged by Bors] - feat(Condensed): functor from light profinite sets to light condensed sets

dagurtomas:
---


Mathlib/Condensed/Light/Functors.lean
```diff
@@ -0,0 +1,45 @@
...
+  the functor that goes through `TopCat.{u+1}`.
+
+-/
+
+universe u v
+
+open CategoryTheory Limits
+
+/-- The functor from `LightProfinite.{u}` to `LightCondSet.{u}` given by the Yoneda sheaf. -/
+def lightProfiniteToLightCondSet : LightProfinite.{u} ⥤ LightCondSet.{u} where
```
joelriou, line 35: I think it would be nice to define this as a particular case of the `yoneda` functor to sheaves of types for a subcanonical Grothendieck topology. Could you add the definition which is in draft form at https://github.com/leanprover-community/mathlib4/blob/jriou_localization/Mathlib/CategoryTheory/Sites/Canonical.lean#L253 and redefine `lightProfiniteToLightCondSet`.

dagurtomas: Thanks for the suggestion. It's added.

joelriou: Great! Could you also define a `FullyFaithful` structure for this and get `Full/Faithful` instances?

joelriou:
Thanks!

bors d+

dagurtomas:
bors merge

dagurtomas:
bors merge

PR #13506, opened 2024-06-04
[Merged by Bors] - feat(CategoryTheory): transposition of commutative squares across an adjunction

emilyriehl:
In the presence of an adjunction `F ⊣ G` a square `F.map f ≫ g = h ≫ k` whose left map is in the image of the left adjoint commutes iff the transposed square `f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k` whose right map is in the image of the right adjoint commutes. The proof is by two instances of `homEquiv_naturality` and the `adj.homEquiv` bijection.

---

These short proofs could likely be simplified further. 

I defined two separate functions to be used to transpose squares in each direction, but perhaps this is better set up as an if and only if? If so I'd appreciate guidance on how to extract the functions in their current form for use in proofs.

Suggestions welcome.

### DIFF
--- a/Mathlib/CategoryTheory/Adjunction/Basic.lean
+++ b/Mathlib/CategoryTheory/Adjunction/Basic.lean
@@ -176,11 +176,13 @@ theorem homEquiv_naturality_right_symm (f : X ⟶ G.obj Y) (g : Y ⟶ Y') :
   simp only [homEquiv_naturality_right,eq_self_iff_true,Equiv.apply_symm_apply]
 #align category_theory.adjunction.hom_equiv_naturality_right_symm CategoryTheory.Adjunction.homEquiv_naturality_right_symm
 
-theorem homEquiv_naturality_left_square (f : X' ⟶ X) (g : F.obj X ⟶ Y') (h : F.obj X' ⟶ Y) (k : Y ⟶ Y') (w : F.map f ≫ g = h ≫ k) : f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k := by
+theorem homEquiv_naturality_left_square (f : X' ⟶ X) (g : F.obj X ⟶ Y') (h : F.obj X' ⟶ Y) (k : Y ⟶ Y') (w : F.map f ≫ g = h ≫ k) :
+    f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k := by
   rw [← homEquiv_naturality_left, ← homEquiv_naturality_right]
   exact congrArg (adj.homEquiv X' Y') w
 
-theorem homEquiv_naturality_right_square (f : X' ⟶ X) (g : X ⟶ G.obj Y') (h : X' ⟶ G.obj Y) (k : Y ⟶ Y') (w : f ≫ g = h ≫ G.map k) : F.map f ≫ (adj.homEquiv X Y').symm g = (adj.homEquiv X' Y).symm h ≫ k := by
+theorem homEquiv_naturality_right_square (f : X' ⟶ X) (g : X ⟶ G.obj Y') (h : X' ⟶ G.obj Y) (k : Y ⟶ Y') (w : f ≫ g = h ≫ G.map k) :
+    F.map f ≫ (adj.homEquiv X Y').symm g = (adj.homEquiv X' Y).symm h ≫ k := by
   rw [← homEquiv_naturality_left_symm, ← homEquiv_naturality_right_symm]
   exact congrArg (adj.homEquiv X' Y').symm w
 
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -176,6 +176,18 @@ theorem homEquiv_naturality_right_symm (f : X ⟶ G.obj Y) (g : Y ⟶ Y') :
   simp only [homEquiv_naturality_right,eq_self_iff_true,Equiv.apply_symm_apply]
 #align category_theory.adjunction.hom_equiv_naturality_right_symm CategoryTheory.Adjunction.homEquiv_naturality_right_symm
 
+theorem homEquiv_naturality_left_square (f : X' ⟶ X) (g : F.obj X ⟶ Y')
+    (h : F.obj X' ⟶ Y) (k : Y ⟶ Y') (w : F.map f ≫ g = h ≫ k) :
+    f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k := by
+  rw [← homEquiv_naturality_left, ← homEquiv_naturality_right]
+  exact congrArg (adj.homEquiv X' Y') w
```
joelriou, line 183:
```suggestion
  rw [← homEquiv_naturality_left, ← homEquiv_naturality_right, w]
```

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -176,6 +176,18 @@ theorem homEquiv_naturality_right_symm (f : X ⟶ G.obj Y) (g : Y ⟶ Y') :
...
+
+theorem homEquiv_naturality_right_square (f : X' ⟶ X) (g : X ⟶ G.obj Y')
+    (h : X' ⟶ G.obj Y) (k : Y ⟶ Y') (w : f ≫ g = h ≫ G.map k) :
+    F.map f ≫ (adj.homEquiv X Y').symm g = (adj.homEquiv X' Y).symm h ≫ k := by
+  rw [← homEquiv_naturality_left_symm, ← homEquiv_naturality_right_symm]
+  exact congrArg (adj.homEquiv X' Y').symm w
```
joelriou, line 189:
```suggestion
  rw [← homEquiv_naturality_left_symm, ← homEquiv_naturality_right_symm, w]
```

joelriou:
> I defined two separate functions to be used to transpose squares in each direction, but perhaps this is better set up as an if and only if? If so I'd appreciate guidance on how to extract the functions in their current form for use in proofs.

Additionally to these two lemmas, you may add two `_iff` lemmas.

Mathlib/CategoryTheory/Adjunction/Basic.lean
emilyriehl: Right, of course! I'm not used to using rewrites, and forget how powerful this tactic is. Thanks.

emilyriehl:
> > I defined two separate functions to be used to transpose squares in each direction, but perhaps this is better set up as an if and only if? If so I'd appreciate guidance on how to extract the functions in their current form for use in proofs.
> 
> Additionally to these two lemmas, you may add two `_iff` lemmas.

Added. Let me know what you think.

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -176,6 +176,41 @@ theorem homEquiv_naturality_right_symm (f : X ⟶ G.obj Y) (g : Y ⟶ Y') :
...
+theorem homEquiv_naturality_left_square_iff (f : X' ⟶ X) (g : F.obj X ⟶ Y')
+    (h : F.obj X' ⟶ Y) (k : Y ⟶ Y') :
+    (F.map f ≫ g = h ≫ k) ↔
+    (f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k) where
+  mp := homEquiv_naturality_left_square adj f g h k
+  mpr := by
+    intro w
+    have transposedw := homEquiv_naturality_right_square adj _ _ _ _ w
+    simp only [Equiv.symm_apply_apply] at transposedw
+    exact transposedw
```
joelriou, line 200:
```suggestion
    (f ≫ (adj.homEquiv X Y') g = (adj.homEquiv X' Y) h ≫ G.map k) ↔
      (F.map f ≫ g = h ≫ k) :=
  ⟨fun w ↦ by simpa only [Equiv.symm_apply_apply]
      using homEquiv_naturality_right_square adj _ _ _ _ w,
    homEquiv_naturality_left_square adj f g h k⟩
```

I suggest stating the lemma in the other direction (but unfortunately it cannot be made a `simp` lemma).

There is nothing wrong with the `where mp := ... mpr := ...` syntax, but in mathlib, we are more used to term proofs as in my suggestion when the proofs are very short, or, by using the following syntax in tactic mode otherwise:
```lean
  constructor
  · sorry -- first implication
  · sorry -- second implication
```

About, the `have transposedw := ...` syntax, when there is only one such `have`, we may use `have := ...` and use `this` instead of an explicit name. In my suggestion, I have used the tactic `simpa ... using ...`, which basically apply simp lemmas to given term until it hopefully matches the goal.

emilyriehl:
Thanks @joelriou for both the suggestion and explanation. I've implemented it for the other proof as well.

joelriou:
Thanks!

bors merge

PR #13507, opened 2024-06-04
[Merged by Bors] - feat(CategoryTheory): a sequential limit of surjections is surjective

dagurtomas:
We prove that in a concrete category whose forgetful functor preserves sequential limits, a sequential limit of surjections is surjective. Since this will be applied to `LightProfinite`, we provide the necessary instance to be able to apply this lemma as well. 

---
- [x] depends on: #13319 

This will be used in a future PR to prove that a sequential limit of epimorphisms in `LightCondSet` is an epimorphism.


joelriou:
Is not it possible to deduce this from the results in `CategoryTheory.CofilteredSystem`? (The API may have to be slightly developed though.)

dagurtomas:
> Is not it possible to deduce this from the results in `CategoryTheory.CofilteredSystem`? (The API may have to be slightly developed though.)

That file deals mostly with cofiltered systems of *finite* sets. Here I have a cofiltered system (indexed by N^op) of light profinite sets, possibly inifinite. It is surely possible, however, to deduce this from a similar statement about inverse systems in `Type _`, because the forgetful functor from `LightProfinite` to `Type _` preserves the limits in question.

Mathlib/CategoryTheory/Limits/ConcreteCategory.lean
```diff
@@ -53,16 +53,30 @@ theorem Concrete.to_product_injective_of_isLimit {D : Cone F} (hD : IsLimit D) :
...
   Concrete.isLimit_ext F (limit.isLimit _) _ _
 #align category_theory.limits.concrete.limit_ext CategoryTheory.Limits.Concrete.limit_ext
 
+section Surjective
+
+/--
+Given surjections `⋯ ⟶ Xₙ₊₁ ⟶ Xₙ ⟶ ⋯ ⟶ X₀` in a concrete category whose forgetful functor
+preserves sequential limits, the projection map `lim Xₙ ⟶ X₀` is surjective.
+-/
+lemma limit_of_surjections_surjective {C : Type u} [Category.{v} C] [ConcreteCategory.{v} C]
```
erdOne, line 72: I think the name given by the naming convention would be something like `surjective_π_app_zero_of_surjective_map`.

erdOne (approved):
LGTM modulo the nitpick on naming.

dagurtomas:
Fixed the name

erdOne:
Thanks!
maintainer merge

semorrison:
bors merge

PR #13518, opened 2024-06-04
[Merged by Bors] - chore: Scope some extremely generic notation

PatrickMassot:
In addition this notation was used exactly twice in Mathlib…

---


ocfnash:
bors merge

PR #13526, opened 2024-06-05
[Merged by Bors] - chore: bump toolchain to v4.8.0

kim-em:
---


semorrison:
Unfortunately we're getting build messages from ProofWidgets replayed during the Mathlib build. Not sure of the fix yet, it may require a new version. :-(

semorrison:
I've solved that problem just by silencing ProofWidgets, but now we're having more problems because Mathlib CI is running the `widgetJsAllTarget` in ProofWidgets, which is really shouldn't be.

semorrison:
bors merge

PR #13533, opened 2024-06-05
[Merged by Bors] - feat (RingTheory/Valuation) : add lemmas about valuations and notations for valued

jjdishere:
Add related lemmas about valuations. Add notations for integer, maximal ideal and residue field for valued field. Relocate files concerning `Valued` under a new folder Topology/Algebra/Valued.

Co-authored-by: Bichang Lei <leibichang58@ruc.edu.cn>

---

This is my first PR. Thank you!

The `IsValExtension` part is split to #14342 


grunweg:
Welcome to mathlib; thanks for your contribution! I presume you meant to mark this as ready for review, hence I have added the `awaiting-review` label to indicate so. (This label and a few related ones are explained on [this page](https://leanprover-community.github.io/contribute/index.html).)

Mathlib/Topology/Algebra/Valued/ValExtension.lean
```diff
@@ -0,0 +1,245 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jiedong Jiang, Bichang Lei
+-/
+import Mathlib.Topology.Algebra.Valued.ValuedField
+
+/-!
+# Extension of Valuation
+
+In this file, we define the typeclass for valuation extensions and prove basic facts about
+extension of valuations. Let `A` be a `R` algebra, both `R` and `A` are equipped with valued
```
erdOne, line 12:
```suggestion
extension of valuations. Let `A` be an `R` algebra, both `R` and `A` are equipped with valued
```

Mathlib/Topology/Algebra/Valued/ValExtension.lean
```diff
@@ -0,0 +1,245 @@
...
+instance. Here, extension of valuation means that the pullback of valuation on `A` is equivalent
```
erdOne, line 13:
```suggestion
instances. Here, extension of valuation means that the pullback of valuation on `A` is equivalent
```

Mathlib/RingTheory/Valuation/ValuationSubring.lean
```diff
@@ -416,6 +416,7 @@ variable {Γ Γ₁ Γ₂ : Type*} [LinearOrderedCommGroupWithZero Γ]
   (v₁ : Valuation K Γ₁) (v₂ : Valuation K Γ₂)
 
 /-- The valuation subring associated to a valuation. -/
+@[reducible]
```
erdOne, line 419: Is there a reason why this is necessary?

Mathlib/Topology/Algebra/Valued/ValExtension.lean
```diff
@@ -0,0 +1,245 @@
...
+
+/--
+The class `IsValExtension R A` states that the valuation of `A` is an extension of the valuation
+on `R`. More precisely, the valuation on `R` is equivlent to the comap of the valuation on `A`.
+-/
+class IsValExtension (R A : Type*) {ΓR ΓA : outParam Type*} [CommRing R] [Ring A]
+  [LinearOrderedCommGroupWithZero ΓR] [LinearOrderedCommGroupWithZero ΓA]
+  [Algebra R A] [vR : Valued R ΓR] [vA : Valued A ΓA] : Prop where
+  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
+  val_isEquiv_comap : vR.v.IsEquiv <| vA.v.comap (algebraMap R A)
```
erdOne, line 80: I think this (and many results below) does not actually need the topological structure and should use `Valuation` instead of `Valued` and go in `RingTheory/Valuation`.

Mathlib/Topology/Algebra/Valued/ValExtension.lean
```diff
@@ -0,0 +1,245 @@
...
+  convert val_map_le_iff hf x 1 <;>
+  simp only [_root_.map_one]
+
+theorem val_map_lt_one_iff (x : R) : v (f x) < 1 ↔ v x < 1 := by
+  convert val_map_lt_iff hf x 1 <;>
+  simp only [_root_.map_one]
+
+theorem val_map_eq_one_iff (x : R) : v (f x) = 1 ↔ v x = 1 := by
+  convert val_map_eq_iff hf x 1 <;>
+  simp only [_root_.map_one]
```
erdOne, line 68: I think most of these exists one way or another in `RingTheory/Valuation/Basic.lean` near `IsEquiv`. If there is anything missing it should be added there, and I don't think we have to restate them when one of the arguments is a restriction.

Mathlib/Topology/Algebra/Valued/ValExtension.lean
jjdishere: I agree that these results do not actually need the topological structure. The idea of choice here is that, for convenience, I want to make this into a typeclass so that Lean can infer it automatically. (Since almost all field extensions that people care about in the local case are extensions of valuation.) `Valued` is a class, but `Valuation` is not a class. What do you suggest?

Mathlib/RingTheory/Valuation/ValuationSubring.lean
jjdishere: My consideration here is to make instances equipped on `Valuation.integer` adopted to `Valuation.valuationSubring` automatically. Otherwise, these instances have to be written twice. (Similar things will happen for `Valued.integer` and `Valued.valuationSubring`, e.g. there will be different instances `Algebra vR.v.integer vA.v.integer` and `Algebra vR.v.valuationSubring vA.v.valuationSubring`) I doubt there will be any instances that should only work on `Valuation.integer` but not `Valuation.valuationSubring`.

Mathlib/Topology/Algebra/Valued/ValExtension.lean
jjdishere: Thank you! I'll move or delete these lemmas.

Mathlib/Topology/Algebra/Valued/ValExtension.lean
erdOne: What about
```
class IsValExtension {R A ΓR ΓA : Type*} [CommRing R] [Ring A]
  [LinearOrderedCommGroupWithZero ΓR] [LinearOrderedCommGroupWithZero ΓA]
  [Algebra R A] (vR : Valuation R ΓR) (vA : Valuation A ΓA) : Prop where
  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
  val_isEquiv_comap : vR.IsEquiv <| vA.comap (algebraMap R A)
```
So that `IsValExtension vR vA` still can be inferred automatically.

Mathlib/RingTheory/Valuation/ValuationSubring.lean
erdOne: Do you really need `Algebra vR.v.valuationSubring vA.v.valuationSubring` though? The main thing I'm concerning is about the performance impact.

erdOne: A few weeks ago we just changed the definition of `ringOfIntegers` from a `Subring` to a `Type*` because of performance reasons and I'm afraid this is going the opposite direction.

jjdishere: > Do you really need `Algebra vR.v.valuationSubring vA.v.valuationSubring` though? The main thing I'm concerning is about the performance impact.

Yes, I need it to create the `IsLocalRingHom vR.v.valuationSubring vA.v.valuationSubring` instance, when `R` and `A` are fields. So should I indeed write all these instances twice? That's acceptable.

erdOne: I'm not sure I understand what `IsLocalRingHom vR.v.valuationSubring vA.v.valuationSubring` means

jjdishere: Sorry, it should be`IsLocalRingHom (algebraMap vR.v.valuationSubring vA.v.valuationSubring)`.

erdOne: Why not `IsLocalRingHom (algebraMap vR.v.integer vA.v.integer)` then? And maybe put `LocalRing` instances on `vR.v.integer` instead.

jjdishere: Well, if we only use `vR.v.integer` instead of `vR.v.valuationSubring` in instances, mathematically you simply lose some information when `R` and `A` are fields. For example, in some propositions, one really needs that `vR.v.integer` is a valuation subring of `R` and ` IsScalarTower vR.v.integer vA.v.integer A` together. It would be inconvenient to do this without `IsScalarTower vR.v.valuationSubring vA.v.valuationSubring A`.

erdOne: Hmmm. Do you have an example? Mathematically $A \hookrightarrow \mathrm{Frac}(A)$ with $A$ being a valuation ring should be everything `valuationSubring` entails.

jjdishere: > Hmmm. Do you have an example? Mathematically A↪Frac(A) with A being a valuation ring should be everything `valuationSubring` entails.

Yes, you are right. What I need is exactly $$A \hookrightarrow \text{Frac}(A).$$ But for general `R`, we do not have `R = Frac vR.v.integer` unless `R` is a field. So do you think I should write instances like`LocalRing vR.v.integer` and `IsFractionRing vR.v.integer R` when `R` is a field (they fail when `R` is not a field)? If I use `vR.v.valuationSubring`, they can be inferred automatically.

erdOne: Yes I'd prefer `ValuationRing vR.v.integer` and `IsFractionRing vR.v.integer R` over putting more structure on `vR.valuationSubring` in case we would like to make `Valuation.integer` a type in the future due to performance concerns.

jjdishere: I see, I'll work on it. Thank you for your suggestion! I learned a lot : )

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,181 @@
...
+    exact mul_assoc _ _ _
+
+theorem integerAlgebra_injective {K A ΓK ΓA : Type*} [Field K] [Ring A] [Nontrivial A]
+    [LinearOrderedCommGroupWithZero ΓK] [LinearOrderedCommGroupWithZero ΓA]
+    [Algebra K A] (vK : Valuation K ΓK) (vA : Valuation A ΓA) [IsValExtension vK vA] :
+    Function.Injective (algebraMap vK.integer vA.integer) := by
+  intro x y h
+  simp only [Subtype.ext_iff, coe_algebraMap_integer] at h
+  ext
+  apply RingHom.injective (algebraMap K A) h
```
erdOne, line 160: If you state it like this, it can be an instance (and it is more general)
```lean
instance {K A ΓK ΓA : Type*} [CommRing K] [Ring A]
    [LinearOrderedCommGroupWithZero ΓK] [LinearOrderedCommGroupWithZero ΓA]
    [Algebra K A] (vK : Valuation K ΓK) (vA : Valuation A ΓA) [IsValExtension vK vA]
    [NoZeroSMulDivisors K A] :
    NoZeroSMulDivisors vK.integer vA.integer := by
  refine ⟨fun {x y} e ↦ ?_⟩
  have : (x : K) • (y : A) = 0 := by simpa [Subtype.ext_iff, Algebra.smul_def] using e
  simpa only [Subtype.ext_iff, smul_eq_zero] using this
```
(you can still provide `integerAlgebra_injective` below; and it should be called `algebraMap_injective`.)

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,181 @@
...
+    · simp [h] at hr
+    · apply Valuation.Integers.isUnit_of_one (v := vK)
+      · exact Valuation.integer.integers (v := vK)
+      · simp only [isUnit_iff_ne_zero, ne_eq]
+        exact Subtype.ext_iff.not.mp h
+      · apply Valuation.Integers.one_of_isUnit (Valuation.integer.integers (v := vL)) at hr
+        change vL (((algebraMap vK.integer vL.integer) r) : L) = 1 at hr
+        norm_cast at hr
+        rw [val_map_eq_one_iff vK] at hr
+        exact hr
```
erdOne, line 177: This works for arbitrary rings as long as the map `K -> L` is also a local ring hom.
```lean
instance instIsLocalRingHomValuationInteger {K L ΓK ΓL : Type*} [CommRing K] [CommRing L]
    [LinearOrderedCommGroupWithZero ΓK] [LinearOrderedCommGroupWithZero ΓL]
    [Algebra K L] [IsLocalRingHom (algebraMap K L)] {vK : Valuation K ΓK} {vL : Valuation L ΓL} [IsValExtension vK vL] :
    IsLocalRingHom (algebraMap vK.integer vL.integer) where
  map_nonunit r hr := by
    apply (Valuation.integer.integers (v := vK)).isUnit_of_one
    · exact (isUnit_map_iff (algebraMap K L) _).mp (hr.map (algebraMap _ L))
    · apply (Valuation.integer.integers (v := vL)).one_of_isUnit at hr
      exact (val_map_eq_one_iff vK vL _).mp hr
```
Note that we might be missing this instance somewhere
```lean
instance (priority := 100) {K L} [Field K] [CommRing L] [Nontrivial L]
    (f : K →+* L) : IsLocalRingHom f where
  map_nonunit r hr := by simpa using hr.ne_zero
```

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,181 @@
...
+      simp
+    map_add' _ _ := by
+      ext
+      simp
+    commutes' _ _ := by
+      ext
+      exact Algebra.commutes _ _
+    smul_def' _ _ := by
+      ext
+      exact Algebra.smul_def _ _
```
erdOne, line 141:
```suggestion
  smul r a := ⟨r • a,
    Algebra.smul_def r (a : A) ▸ mul_mem ((val_map_le_one_iff vR vA _).mpr r.2) a.2⟩
  __ := (algebraMap R A).restrict vR.integer vA.integer
    (by simp [Valuation.mem_integer_iff, val_map_le_one_iff vR vA])
  commutes' _ _ := Subtype.ext (Algebra.commutes _ _)
  smul_def' _ _ := Subtype.ext (Algebra.smul_def _ _)
```
Also please provide a simp lemma for smul.

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,181 @@
...
+-/
+theorem ofIntegerComap {K A ΓK ΓA : Type*} [Field K] [Ring A]
+    [LinearOrderedCommGroupWithZero ΓK] [LinearOrderedCommGroupWithZero ΓA]
+  [Algebra K A] (vK : Valuation K ΓK) (vA : Valuation A ΓA)
+    (h : vA.integer.comap (algebraMap K A) = vK.integer) : IsValExtension vK vA where
+  val_isEquiv_comap := by
+    rw [isEquiv_iff_val_le_one]
+    intro x
+    rw [← Valuation.mem_integer_iff, ← Valuation.mem_integer_iff]
+    rw [← h]
```
erdOne, line 94:
```suggestion
    rw [← Valuation.mem_integer_iff, ← Valuation.mem_integer_iff, ← h]
```

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,8 +135,68 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+        congr
+        simp only [← Subring.coe_add, h]
+        rfl
+      _ ≤ max (v a) (v b) := by
+        exact map_add v a b
+      _ < 1 := by
+        apply max_lt
+        exact lt_of_le_of_ne a.2 hab.1
+        exact lt_of_le_of_ne b.2 hab.2
+    simp only [_root_.map_one, lt_self_iff_false] at this
```
erdOne, line 169: This should be `ValuationRing v.integer`, be provided below `ValuationRing.of_integers`, using that lemma.

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,8 +135,68 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+          apply Subtype.ext_iff.not.mpr
+          simp only [ZeroMemClass.coe_zero, inv_eq_zero, this, not_false_eq_true]
+      }
+      use ⟨1, zinv⟩
+      calc
+      _ = z * z⁻¹ := by rfl
+      _ = 1 := by field_simp
+  exists_of_eq h := by
+    use 1
+    simp only [OneMemClass.coe_one, (integer.integers v).hom_inj h, one_mul]
```
erdOne, line 198: We should first show that for a valuation ring `A`, `K` is the fraction ring if the map is injective and either `x` or `x⁻¹` is in the range for all `x ≠ 0`, and then show this for `Valuation.Integers` and finally specialize to `v.nteger`.

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Thank you. I'll add the instance and rename `integerAlgebra_injective` to `algebraMap_integer_injective`

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Thank you. I will put this instance inside the file `Mathlib.RingTheory.Ideal.LocalRing`.

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Thank you! I will add `coe_smul_integer`

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,8 +135,68 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
   ⟨hv.dvd_of_le, hv.le_of_dvd⟩
 #align valuation.integers.le_iff_dvd Valuation.Integers.le_iff_dvd
 
+theorem isUnit_of_one' {x : O} (hvx : v (algebraMap O F x) = 1) :
```
erdOne, line 138: Can you add a brief docstring describing the difference between this and the unprimed version?

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Is there a danger that this theorem name without containing `integer` would make people think of this theorem? (Maybe this theorem will not get into mathlib anyway ...) 
`theorem IsValExtension.algebraMap_injective {K A ΓK ΓA : Type*} [Field K] [Ring A]
    [LinearOrderedCommGroupWithZero ΓK] [LinearOrderedCommGroupWithZero ΓA]
    [Algebra K A] (vK : Valuation K ΓK) (vA : Valuation A ΓA) [IsValExtension vK vA] [Nontrivial ΓA]
    Function.Injective K A`

And this remind me that I should add a theorem that states `Nontrivial ΓA` implies `Nontrivial A`

Mathlib/RingTheory/Valuation/Integers.lean
jjdishere: I'll move it and modify the proof.

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,150 @@
...
+instance integerAlgebra : Algebra vR.integer vA.integer where
+  smul r a := ⟨r • a,
+    Algebra.smul_def r (a : A) ▸ mul_mem ((val_map_le_one_iff vR vA _).mpr r.2) a.2⟩
+  __ := (algebraMap R A).restrict vR.integer vA.integer
+    (by simp [Valuation.mem_integer_iff, val_map_le_one_iff vR vA])
+  commutes' _ _ := Subtype.ext (Algebra.commutes _ _)
+  smul_def' _ _ := Subtype.ext (Algebra.smul_def _ _)
+
+@[simp, norm_cast]
+theorem coe_algebraMap_integer (r : vR.integer) :
```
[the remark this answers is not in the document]
jjdishere: Thank you, I'll change this name.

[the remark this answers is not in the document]
jjdishere: Also change the new smul lemma's name to `IsValExtension.val_smul`

Mathlib/RingTheory/Valuation/ValExtension.lean
erdOne: My reason is exactly because this lemma shouldn't be in mathlib. You should get the `Nontrivial A` instance and use `RingHom.injective`.

Also you should generalize `RingHom.domain_nontrivial` to `MonoidWithZero.domain_nontrivial` if you need the last result that you mention.

Mathlib/RingTheory/Valuation/Integers.lean
jjdishere: OK, I'll add some docstring for both this theorem and the unprimed version, explaining their relation and the theorem itself.

Mathlib/RingTheory/Valuation/Integers.lean
jjdishere: I'll split it into 2 instances and a theorem. Thank you.

Mathlib/RingTheory/Valuation/ValuationRing.lean
```diff
@@ -413,6 +413,65 @@ theorem of_integers : ValuationRing 𝒪 := by
...
+        | inr ha =>
+          use ⟨1,⟨a, mem_nonZeroDivisors_of_ne_zero h0⟩⟩
+          have : x ≠ 0 := by
+            intro hx
+            simp only [hx, inv_zero] at ha
+            exact h0 <| (injective_iff_map_eq_zero _).mp h.2 a ha.symm
+          field_simp [← ha]
+    · intro a b hab
+      use 1
+      simpa using h.2 hab
```
erdOne, line 468:
```suggestion
  constructor <;> intro h
  · constructor
    · intro x
      obtain (⟨a, e⟩ | ⟨a, e⟩) := (iff_isInteger_or_isInteger 𝒪 K).mp inferInstance x
      exacts [⟨a, .inl e.symm⟩, ⟨a, .inr e.symm⟩]
    · exact IsFractionRing.injective _ _
  · constructor
    · intro a
      simpa using h.2.ne_iff.mpr (nonZeroDivisors.ne_zero a.2)
    · intro x
      obtain ⟨a, ha⟩ := h.1 x
      by_cases h0 : a = 0
      · exact ⟨⟨0, 1⟩, by simpa [h0] using ha⟩
      · have : algebraMap 𝒪 K a ≠ 0 := by simpa using h.2.ne_iff.mpr h0
        rw [inv_eq_iff_eq_inv, ← one_div, eq_div_iff this] at ha
        cases ha with
        | inl ha => exact ⟨⟨a, 1⟩, by simpa⟩
        | inr ha => exact ⟨⟨1, ⟨a, mem_nonZeroDivisors_of_ne_zero h0⟩⟩, by simpa using ha⟩
    · intro a b hab
      simpa [h.2 hab] using ⟨1, one_mem _⟩
```

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,6 +138,34 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+  by_cases h : v x ≤ 1
+  · obtain ⟨a, ha⟩ := exists_of_le_one hv h
+    use a
+    exact Or.inl ha.symm
+  · have : v x⁻¹ ≤ 1 := by
+      apply le_of_lt
+      push_neg at h
+      rwa [← Valuation.one_lt_val_iff]
+      intro h0
+      simp [h0] at h
```
erdOne, line 164: We should have `Valuation.val_le_one_iff`

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,6 +138,34 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+-/
+theorem isUnit_of_one' {x : O} (hvx : v (algebraMap O F x) = 1) :
+    IsUnit x := by
+  apply isUnit_of_one hv _ hvx
+  apply IsUnit.mk0
+  rw [← v.ne_zero_iff]
+  simp only [hvx, ne_eq, one_ne_zero, not_false_eq_true]
+
+theorem eq_algebraMap_or_inv_eq_algebraMap (x : F) :
+    ∃ a : O, x = algebraMap O F a ∨ x⁻¹ = algebraMap O F a := by
```
erdOne, line 154: Can you use this to golf `Valuation.valuationSubring`?

jjdishere: Sorry, what do you mean by "golf"?

erdOne: Shorten the proof of. As in [code golf](https://en.wikipedia.org/wiki/Code_golf).

Mathlib/RingTheory/Valuation/Integers.lean
jjdishere: Yes! I'll do this.

Should we have four lemmas? `one_lt_val_iff`, `one_le_val_iff`, `val_le_one_iff`, `val_lt_one_iff`? Or only two of them?

erdOne: Hmmm. Or maybe not and we should just `rw [map_inv₀, inv_le_one₀]`

jjdishere: I added four lemmas and the following one
```
-- This theorem is not stated in the strongest form but is very useful.
theorem val_le_one_or_val_inv_le_one (v : Valuation K Γ₀) (x : K) : v x ≤ 1 ∨ v x⁻¹ ≤ 1 := by
```
and used it to golf the `Valuation.valuationSubring`

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,156 @@
...
+  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
+  val_isEquiv_comap : vR.IsEquiv <| vA.comap (algebraMap R A)
+
+namespace IsValExtension
+
+section CoeLemma
+
+variable {R A ΓR ΓA : Type*} [CommRing R] [Ring A]
+    [LinearOrderedCommMonoidWithZero ΓR] [LinearOrderedCommMonoidWithZero ΓA]
+    [Algebra R A] (vR : Valuation R ΓR) (vA : Valuation A ΓA) [IsValExtension vR vA]
```
jcommelin, line 52: Please consider moving this to L35. Then you don't need to repeat these assumptions on L40 and L100, etc.

Mathlib/Topology/Algebra/Valued/ValuedField.lean
```diff
@@ -362,3 +362,38 @@ theorem valuedCompletion_apply (x : K) : Valued.v (x : hat K) = v x :=
...
 end Valued
+
+section Notation
+
+namespace Valued
+
+/-- A `Valued` version of `Valuation.integer`, enabling the notation `𝒪[K]` for the
+valuation integers of a valued field `K`. -/
+@[reducible]
+def integer (K : Type*) [Field K] {Γ : outParam Type*} [LinearOrderedCommGroupWithZero Γ]
```
jcommelin, line 373: Please consider extracting these variables into a `variable` statement.

jcommelin (commented):
Looks good to me! Just minor comments.

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Thank you! Fixed. There is a tiny difference between `LinearOrderedCommMonoidWithZero` and `LinearOrderedCommGroupWithZero` which makes things not completely trivial.

Mathlib/Topology/Algebra/Valued/ValuedField.lean
jjdishere: Thank you! Fixed.

erdOne:
This looks good to me and I'll maintainer merge this viewing the lack of suggestions from other reviewers as a non-objection.
maintainer merge

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,6 +138,24 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+over a field. Let `v` be a valuation on some field `F` and `O` be its integers. For every element
+`x` in `O`, `x` is a unit in `O` if and only if the image of `x` in `F` has valuation 1.
+-/
+theorem isUnit_of_one' {x : O} (hvx : v (algebraMap O F x) = 1) :
+    IsUnit x := by
+  apply isUnit_of_one hv _ hvx
+  apply IsUnit.mk0
+  rw [← v.ne_zero_iff]
+  simp only [hvx, ne_eq, one_ne_zero, not_false_eq_true]
+
```
faenuccio, line 152:
```suggestion
  apply isUnit_of_one hv _ hvx
  apply IsUnit.mk0
  simp only [← v.ne_zero_iff, hvx, ne_eq, one_ne_zero, not_false_eq_true]

```

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -344,6 +344,11 @@ theorem surjective_units_map_of_local_ringHom [CommRing R] [CommRing S] (f : R 
   exact hb
 #align local_ring.surjective_units_map_of_local_ring_hom LocalRing.surjective_units_map_of_local_ringHom
 
+/-- Every ring hom `f : K → R` from a field `K` to a nontrivial ring `R` is a local ring hom. -/
```
faenuccio, line 347:
```suggestion
/-- Every ring hom `f : K →+* R` from a field `K` to a nontrivial ring `R` is a local ring hom. -/
```

Mathlib/RingTheory/Valuation/Basic.lean
```diff
@@ -349,6 +349,28 @@ theorem one_lt_val_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : 1 < v x 
   simpa using (inv_lt_inv₀ (v.ne_zero_iff.2 h) one_ne_zero).symm
 #align valuation.one_lt_val_iff Valuation.one_lt_val_iff
 
+theorem one_le_val_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : 1 ≤ v x ↔ v x⁻¹ ≤ 1 := by
+  convert (one_lt_val_iff v (inv_ne_zero h)).symm.not <;>
+  push_neg <;> simp only [inv_inv]
+
+theorem val_lt_one_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : v x < 1 ↔ 1 < v x⁻¹ := by
+  convert (one_lt_val_iff v (inv_ne_zero h)).symm
+  simp only [inv_inv]
```
faenuccio, line 358:
```suggestion
  simpa only [inv_inv] using (one_lt_val_iff v (inv_ne_zero h)).symm
```

Mathlib/RingTheory/Valuation/Basic.lean
```diff
@@ -349,6 +349,28 @@ theorem one_lt_val_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : 1 < v x 
...
+
+theorem val_eq_one_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : v x = 1 ↔ v x⁻¹ = 1 := by
+  simpa only [le_antisymm_iff, And.comm] using and_congr (one_le_val_iff v h) (val_le_one_iff v h)
+
+-- This theorem is not stated in the strongest form but is very useful.
+theorem val_le_one_or_val_inv_le_one (v : Valuation K Γ₀) (x : K) : v x ≤ 1 ∨ v x⁻¹ ≤ 1 := by
+  by_cases h : x = 0
+  · simp only [h, _root_.map_zero, zero_le', inv_zero, or_self]
+  · rw [← one_le_val_iff v h]
+    exact le_total _ _
```
faenuccio, line 372:
```suggestion
  · simp only [← one_le_val_iff v h, le_total]
```

Mathlib/RingTheory/Valuation/Basic.lean
```diff
@@ -398,10 +420,26 @@ theorem comap {S : Type*} [Ring S] (f : S →+* R) (h : v₁.IsEquiv v₂) :
     (v₁.comap f).IsEquiv (v₂.comap f) := fun r s => h (f r) (f s)
 #align valuation.is_equiv.comap Valuation.IsEquiv.comap
 
+/-- Restate the definition of `Valuation.IsEquiv` into a theorem. -/
```
faenuccio, line 423: Why is this useful?

Mathlib/RingTheory/Valuation/Basic.lean
```diff
@@ -398,10 +420,26 @@ theorem comap {S : Type*} [Ring S] (f : S →+* R) (h : v₁.IsEquiv v₂) :
...
+  h r s
+
+theorem val_lt (h : v₁.IsEquiv v₂) {r s : R} : v₁ r < v₁ s ↔ v₂ r < v₂ s := by
+  simpa only [not_le] using (h s r).not
+
 theorem val_eq (h : v₁.IsEquiv v₂) {r s : R} : v₁ r = v₁ s ↔ v₂ r = v₂ s := by
   simpa only [le_antisymm_iff] using and_congr (h r s) (h s r)
 #align valuation.is_equiv.val_eq Valuation.IsEquiv.val_eq
 
+theorem val_le_one (h : v₁.IsEquiv v₂) {r : R} : v₁ r ≤ 1 ↔ v₂ r ≤ 1 := by
```
faenuccio, line 434: Can you explain why these three results are useful?

Mathlib/RingTheory/Valuation/Basic.lean
```diff
@@ -349,6 +349,28 @@ theorem one_lt_val_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : 1 < v x 
...
+  simp only [inv_inv]
+
+theorem val_le_one_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : v x ≤ 1 ↔ 1 ≤ v x⁻¹ := by
+  convert (one_le_val_iff v (inv_ne_zero h)).symm
+  simp only [inv_inv]
+
+theorem val_eq_one_iff (v : Valuation K Γ₀) {x : K} (h : x ≠ 0) : v x = 1 ↔ v x⁻¹ = 1 := by
+  simpa only [le_antisymm_iff, And.comm] using and_congr (one_le_val_iff v h) (val_le_one_iff v h)
+
+-- This theorem is not stated in the strongest form but is very useful.
```
faenuccio, line 367: And what would be its strongest form? Why not stating it in that generality?

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,6 +138,24 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+theorem eq_algebraMap_or_inv_eq_algebraMap (x : F) :
```
faenuccio, line 153: Is this related to valuations? Why is it here?

Mathlib/RingTheory/Valuation/ValuationRing.lean
```diff
@@ -413,6 +413,39 @@ theorem of_integers : ValuationRing 𝒪 := by
...
+      obtain ⟨a, ha⟩ := h.1 x
+      by_cases h0 : a = 0
+      · exact ⟨⟨0, 1⟩, by simpa [h0] using ha⟩
+      · have : algebraMap 𝒪 K a ≠ 0 := by simpa using h.2.ne_iff.mpr h0
+        rw [inv_eq_iff_eq_inv, ← one_div, eq_div_iff this] at ha
+        cases ha with
+        | inl ha => exact ⟨⟨a, 1⟩, by simpa⟩
+        | inr ha => exact ⟨⟨1, ⟨a, mem_nonZeroDivisors_of_ne_zero h0⟩⟩, by simpa using ha⟩
+    · intro a b hab
+      exact ⟨1, by simp only [OneMemClass.coe_one, h.2 hab, one_mul]⟩
```
faenuccio, line 442:
```suggestion
  refine ⟨fun h ↦ ⟨fun x ↦ ?_, IsFractionRing.injective _ _⟩, fun h ↦ ?_⟩
  · obtain (⟨a, e⟩ | ⟨a, e⟩) := isInteger_or_isInteger 𝒪 x
    exacts [⟨a, .inl e.symm⟩, ⟨a, .inr e.symm⟩]
  · constructor
    · intro a
      simpa using h.2.ne_iff.mpr (nonZeroDivisors.ne_zero a.2)
    · intro x
      obtain ⟨a, ha⟩ := h.1 x
      by_cases h0 : a = 0
      · exact ⟨⟨0, 1⟩, by simpa [h0] using ha⟩
      · have : algebraMap 𝒪 K a ≠ 0 := by simpa using h.2.ne_iff.mpr h0
        rw [inv_eq_iff_eq_inv, ← one_div, eq_div_iff this] at ha
        cases ha with
        | inl ha => exact ⟨⟨a, 1⟩, by simpa⟩
        | inr ha => exact ⟨⟨1, ⟨a, mem_nonZeroDivisors_of_ne_zero h0⟩⟩, by simpa using ha⟩
    · intro _ _ hab
      exact ⟨1, by simp only [OneMemClass.coe_one, h.2 hab, one_mul]⟩
```

faenuccio (commented):
I made some comments improving the code and asking for some doc. I would suggest that you split this PR, first with all your additions and then another one adding only the `ValExtension` file, possibly after getting an opinion from María Inés.

Mathlib/RingTheory/Valuation/Basic.lean
jjdishere: Originally, these are lemmas for the typeclass `IsValExtension`, stating that if we have `IsValExtension R A`, we have `v (algebraMap R A r)  < v (algebraMap R A r') ↔ v r < v r'`, with a `@[simp]` label. (Same for `val_le_one`, `val_lt_one`,  `val_eq_one`) Then I was suggested that these lemmas should be stated in a more general form like this. What is your suggestion for these lemmas?

Mathlib/RingTheory/Valuation/Basic.lean
jjdishere: Same as above.

Mathlib/RingTheory/Valuation/Basic.lean
jjdishere: The strongest form is  `v x ≤ 1 ∨ v x⁻¹ < 1`. But based on my experience, many proofs just need you to discuss whether `x` or `x⁻¹` is an integer and the proof goes exactly the same afterwards (thus saves lots of lines if we use this form `v x ≤ 1 ∨ v x⁻¹ ≤ 1`). `v x ≤ 1 ∨ v x⁻¹ < 1` will make the `x⁻¹` case be treated differently with a small `le_of_lt`.

Mathlib/RingTheory/Valuation/Integers.lean
jjdishere: This is a special lemma for the design of `Valuation.Integers`. The math content here holds for all valuation subrings. However, for valuation subrings, since they are really subrings, there is no need to state something like `eq_algebraMap`. But for `Valuation.Integers`, it is a completely different ring `O` with an injective map into `F`, so we need this lemma.

jjdishere:
@faenuccio I have a technical question about splitting the PR. Should I close this PR and open new ones? Do you know some easy ways for me to select some files and discard changes in other files in a PR? Thank you!

Mathlib/RingTheory/Valuation/Basic.lean
jjdishere: This lemma and the following lemmas are used in the proof of 
```
theorem IsValExtension.val_map_le_iff (x y : R) : vA (algebraMap R A x) ≤ vA (algebraMap R A y) ↔ vR x ≤ vR y :=
  (IsEquiv.val_le val_isEquiv_comap).symm
```
and `IsValExtension.val_map_lt_iff`, `IsValExtension.val_map_le_one_iff` ... These are simp lemmas for the class `IsValExtension`.

The restating of the definition is to make the name more consistent. When there are `val_lt` and `val_le_one`, it is better to also have a `val_le`.

What do you think about these lemmas?

Mathlib/RingTheory/Valuation/Basic.lean
faenuccio: Perhaps add both lemmas, then, with the same name (and one with a `'` )

Mathlib/RingTheory/Valuation/Basic.lean
jjdishere: Then I will delete these lemmas and prove `IsValExtension.val_map_lt_iff` directly.

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -344,6 +344,11 @@ theorem surjective_units_map_of_local_ringHom [CommRing R] [CommRing S] (f : R 
   exact hb
 #align local_ring.surjective_units_map_of_local_ring_hom LocalRing.surjective_units_map_of_local_ringHom
 
+/-- Every ring hom `f : K →+* R` from a field `K` to a nontrivial ring `R` is a local ring hom. -/
+instance (priority := 100) {K R} [Field K] [CommRing R] [Nontrivial R]
```
faenuccio, line 348: Can you explain the need for the choice of priority?

Mathlib/RingTheory/Valuation/Integers.lean
```diff
@@ -133,6 +138,23 @@ theorem le_iff_dvd {x y : O} : v (algebraMap O F x) ≤ v (algebraMap O F y) ↔
...
+/--
+This is the special case of `Valuation.Integers.isUnit_of_one` when the valuation is defined
+over a field. Let `v` be a valuation on some field `F` and `O` be its integers. For every element
+`x` in `O`, `x` is a unit in `O` if and only if the image of `x` in `F` has valuation 1.
+-/
+theorem isUnit_of_one' {x : O} (hvx : v (algebraMap O F x) = 1) :
+    IsUnit x := by
+  apply isUnit_of_one hv _ hvx
+  apply IsUnit.mk0
+  simp only [← v.ne_zero_iff, hvx, ne_eq, one_ne_zero, not_false_eq_true]
```
faenuccio, line 150:
```suggestion
theorem isUnit_of_one' {x : O} (hvx : v (algebraMap O F x) = 1) : IsUnit x := by
  refine isUnit_of_one hv (IsUnit.mk0 _ ?_) hvx
  simp only [← v.ne_zero_iff, hvx, ne_eq, one_ne_zero, not_false_eq_true]
```

Mathlib/RingTheory/Valuation/ValuationRing.lean
```diff
@@ -413,6 +413,36 @@ theorem of_integers : ValuationRing 𝒪 := by
     use c; exact Or.inl hc.symm
 #align valuation_ring.of_integers ValuationRing.of_integers
 
+instance instValuationRingInteger : ValuationRing v.integer :=
+  of_integers (v := v) (Valuation.integer.integers v)
+
```
faenuccio, line 418: Can you add a reference to this result in the preamble of the file?

faenuccio (commented):
Please update master on this branch, by the way!

Mathlib/RingTheory/Ideal/LocalRing.lean
jjdishere: I think this is the case that the note [lower instance priority] will apply. This is an instance that will always apply during type-class resolution, so it should have a lower instance. I'll add the comments "-- see Note [lower instance priority]".

Mathlib/RingTheory/Valuation/ValuationRing.lean
jjdishere: I added "We also show that, given a valuation `v` on a field `K`, the ring of valuation integers is a
valuation ring and `K` is the fraction field of this ring." at the preamble.

faenuccio:
Thanks for all your work!!! 🎉

faenuccio:
maintainer merge

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -344,6 +344,12 @@ theorem surjective_units_map_of_local_ringHom [CommRing R] [CommRing S] (f : R 
   exact hb
 #align local_ring.surjective_units_map_of_local_ring_hom LocalRing.surjective_units_map_of_local_ringHom
 
+-- see Note [lower instance priority]
+/-- Every ring hom `f : K →+* R` from a field `K` to a nontrivial ring `R` is a local ring hom. -/
+instance (priority := 100) {K R} [Field K] [CommRing R] [Nontrivial R]
```
riccardobrasca, line 349: This works for `K` a division ring right?

jjdishere: You are right. Thank you!

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -344,6 +344,12 @@ theorem surjective_units_map_of_local_ringHom [CommRing R] [CommRing S] (f : R 
   exact hb
 #align local_ring.surjective_units_map_of_local_ring_hom LocalRing.surjective_units_map_of_local_ringHom
 
+-- see Note [lower instance priority]
+/-- Every ring hom `f : K →+* R` from a field `K` to a nontrivial ring `R` is a local ring hom. -/
```
riccardobrasca, line 348:
```suggestion
/-- Every ring hom `f : K →+* R` from a division ring `K` to a nontrivial ring `R` is a local ring hom. -/
```

riccardobrasca (approved):
Thanks!

bors d+

riccardobrasca:
Thanks!

bors merge

PR #13537, opened 2024-06-05
[Merged by Bors] - chore: protect `Filter.mem_nhds_iff`

j-loreaux:
This conflicts with `mem_nhds_iff`, which is already in the root namespace.

---


dupuisf (commented):
Thanks!

bors r+

PR #13538, opened 2024-06-05
[Merged by Bors] - chore: mark `IsProperMap` with `fun_prop`

j-loreaux:
This removes the `continuity` attribute from `IsProperMap.continuous`, because it is a disaster to mark it as such (it gets a `safe apply` attribute for `aesop`, but it is definitely not safe).

In addition, although `IsProperMap.continuous` was already marked as `fun_prop`, this is actually the incorrect approach unless `IsProperMap` is *also* marked as `fun_prop`, which we do here.

The reason for this is that without `IsProperMap` being marked `fun_prop`, then `IsProperMap.continuous` gets marked as an application theorem for `fun_prop` (much like wth `aesop [safe apply]`). But, when `IsProperMap` is marked as `fun_prop`, then this is marked internally by `fun_prop` as a "transition" theorem.

Consequently, this prevents `fun_prop` from getting stuck with goals of `IsProperMap f` when there is no possible way `f` could be a proper map. Instead, it will get stuck with `Continuous f`, which may be provable.

---


dupuisf (commented):
Thanks!

bors r+

PR #13534, opened 2024-06-05
[Merged by Bors] - feat: variations on the uniformity of uniform convergence given functions whose ranges cover the space

j-loreaux:
Given functions from a family `δ` of types into a type `α` whose ranges cover `α`, the uniform space of uniform convergence of functions from `α` to `β` is the infimum of the uniform spaces obtained as the comaps of precomposition with the maps `δ i → α`, for each `i`. This gives a way to determine uniform convergence on the whole space `α` by considering it on the pieces `δ i`.

Authored-by: @ADedecker 

---

Anatole authored this and I'm just PRing it for him because he is very busy at the moment.

- [x] depends on: #13536
- [x] depends on: #13537
- [x] depends on: #13538

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13541, opened 2024-06-05
[Merged by Bors] - feat: obtain a non-unital continuous functional calculus from a unital one

j-loreaux:
Co-authored by: @ADedecker 

---

- [ ] depends on: #13534


dupuisf (commented):
Thanks!

bors r+

PR #13545, opened 2024-06-05
[Merged by Bors] - feat(FiberedCategory/BasedCategory): add bicategory of based categories

callesonne:
We define the strict bicategory of based categories. Its objects are categories equipped with a functor to some fixed base-category. The strict bicategory of fibered categories will later be built on top of this one.

Co-authored-by: Paul Lezeau [paul.lezeau@gmail.com](mailto:paul.lezeau@gmail.com)

---

This is a remake of a previous PR of the same name, but some awful git accident happened and I think its just easier to remake the PR. (I by mistake added 400 commits to the PR by rebasing instead of merging with master in some weird way).


callesonne:
@joelriou I already incorporated all, except one, of your comments on the previous PR, thanks a lot for them! The suggestion about making `BasedFunctor` have `functor` as a field instead of extending `Functor` I have not changed yet. That is because this change would not let me use the notation `F.obj` for a `F : BasedFunctor`, and instead I would have to use `F.functor.obj`. Is there a way around this?

joelriou:
> @joelriou I already incorporated all, except one, of your comments on the previous PR, thanks a lot for them! The suggestion about making `BasedFunctor` have `functor` as a field instead of extending `Functor` I have not changed yet. That is because this change would not let me use the notation `F.obj` for a `F : BasedFunctor`, and instead I would have to use `F.functor.obj`. Is there a way around this?

Ok, let us keep `extends`.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,244 @@
...
+instance (𝒳 : BasedCategory.{v₂, u₂} 𝒮) : Category 𝒳.obj := 𝒳.category
+
+/-- The based category associated to a functor `p : 𝒳 ⥤ 𝒮`. -/
+def BasedCategory.ofFunctor {𝒳 : Type u₂} [Category.{v₂} 𝒳] (p : 𝒳 ⥤ 𝒮) : BasedCategory 𝒮 where
+  obj := 𝒳
+  p := p
+
+/-- A functor between based categories is a functor between the underlying categories that commutes
+with the projections. -/
+structure BasedFunctor (𝒳 𝒴 : BasedCategory.{v₂, u₂} 𝒮) extends
```
joelriou, line 52: I think we should allow `𝒳` and `𝒴` to have different universe parameters. Then, the type of based functors should be equipped with a category structure in this generality. Basically, the data involved in the definitions of the 2-category structure should be more universe polymorphic: the more restricted universe assumption should appear only in the definition of the `Bicategory` instance.

callesonne: Good point!

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,275 @@
...
+instance (F : 𝒳 ⥤ᵇ 𝒴 ) (a : 𝒳.obj) : IsHomLift 𝒴.p (𝟙 (𝒳.p.obj a)) (𝟙 (F.obj a)) :=
+  IsHomLift.id (w_obj F a)
+
+section
+
+variable (F : 𝒳 ⥤ᵇ 𝒴) {R S : 𝒮} {a b : 𝒳.obj} (f : R ⟶ S) (φ : a ⟶ b)
+
+/-- For a based functor `F : 𝒳 ⟶ 𝒴`, then whenever an arrow `φ` in `𝒳` lifts some `f` in `𝒮`,
+then `F(φ)` also lifts `f`. -/
+instance pres_IsHomLift [IsHomLift 𝒳.p f φ] : IsHomLift 𝒴.p f (F.map φ) := by
```
joelriou, line 103:
```suggestion
instance preserves_isHomLift [IsHomLift 𝒳.p f φ] : IsHomLift 𝒴.p f (F.map φ) := by
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,275 @@
...
+  apply of_fac 𝒴.p f (F.map φ) (Eq.trans (F.w_obj a) (domain_eq 𝒳.p f φ))
+    (Eq.trans (F.w_obj b) (codomain_eq 𝒳.p f φ))
+  rw [← Functor.comp_map, congr_hom F.w]
+  simpa using (fac 𝒳.p f φ)
+
+/-- For a based functor `F : 𝒳 ⟶ 𝒴`, and an arrow `φ` in `𝒳`, then `φ` lifts an arrow `f` in `𝒮`
+if `F(φ)` does. -/
+lemma IsHomLift_ofImage [IsHomLift 𝒴.p f (F.map φ)] : IsHomLift 𝒳.p f φ := by
```
joelriou, line 111:
```suggestion
lemma isHomLift_map [IsHomLift 𝒴.p f (F.map φ)] : IsHomLift 𝒳.p f φ := by
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,275 @@
...
+  apply of_fac 𝒳.p f φ  (F.w_obj a ▸ domain_eq 𝒴.p f (F.map φ))
+    (F.w_obj b ▸ codomain_eq 𝒴.p f (F.map φ))
+  simp [congr_hom F.w.symm, fac 𝒴.p f (F.map φ)]
+
+lemma IsHomLift_iff : IsHomLift 𝒴.p f (F.map φ) ↔ IsHomLift 𝒳.p f φ :=
```
joelriou, line 116:
```suggestion
lemma isHomLift_iff : IsHomLift 𝒴.p f (F.map φ) ↔ IsHomLift 𝒳.p f φ :=
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,275 @@
...
+variable {F G : 𝒳 ⥤ᵇ 𝒴} (α : BasedNatTrans F G)
+
+@[ext]
+lemma ext (β : BasedNatTrans F G) (h : α.toNatTrans = β.toNatTrans) : α = β := by
+  cases α; subst h; rfl
+
+instance app_isHomLift (a : 𝒳.obj) : IsHomLift 𝒴.p (𝟙 (𝒳.p.obj a)) (α.toNatTrans.app a) :=
+  α.aboveId' a
+
+lemma aboveId {a : 𝒳.obj} {S : 𝒮} (ha : 𝒳.p.obj a = S) :
```
joelriou, line 147: A better name should be found.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,274 @@
...
+
+/-- The based category associated to a functor `p : 𝒳 ⥤ 𝒮`. -/
+def BasedCategory.ofFunctor {𝒳 : Type u₂} [Category.{v₂} 𝒳] (p : 𝒳 ⥤ 𝒮) : BasedCategory 𝒮 where
+  obj := 𝒳
+  p := p
+
+/-- A functor between based categories is a functor between the underlying categories that commutes
+with the projections. -/
+structure BasedFunctor (𝒳 : BasedCategory.{v₂, u₂} 𝒮) (𝒴 : BasedCategory.{v₃, u₃} 𝒮) extends
+    CategoryTheory.Functor 𝒳.obj 𝒴.obj where
```
joelriou, line 55:
```suggestion
    𝒳.obj ⥤ 𝒴.obj where
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,274 @@
...
+  w : toFunctor ⋙ 𝒴.p = 𝒳.p := by aesop_cat
+
+/-- Notation for `BasedFunctor`. -/
+scoped infixr:26 " ⥤ᵇ " => BasedFunctor
+
+namespace BasedFunctor
+
+/-- The identity based functor. -/
+@[simps!]
```
joelriou, line 64: I believe it is better to generate a single simp lemma `toFunctor` rather than the projections `obj` and `map`:
```suggestion
@[simps toFunctor]
```
Same for `comp` below.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,274 @@
...
+
+lemma isHomLift {a : 𝒳.obj} {S : 𝒮} (ha : 𝒳.p.obj a = S) :
+    IsHomLift 𝒴.p (𝟙 S) (α.toNatTrans.app a) := by
+  subst ha; infer_instance
+
+end
+
+/-- The identity natural transformation is a `BasedNatTrans`. -/
+@[simps!]
+def id (F : 𝒳 ⥤ᵇ 𝒴) : BasedNatTrans F F where
```
joelriou, line 154:
```suggestion

initialize_simps_projections BasedNatTrans (+toNatTrans, -app)

/-- The identity natural transformation is a `BasedNatTrans`. -/
@[simps]
def id (F : 𝒳 ⥤ᵇ 𝒴) : BasedNatTrans F F where
```
And similarly, `simps!` has to be replaced by `simps` for `comp`. I am not a `simps` tactic guru, but what my suggestion does is that the `simps` attribute on definitions of `BasedNatTrans` will generate the `_toNatTrans` lemmas instead of the `_app` (which may be obtained using `simps!` if this were what we actually wanted).

Then, if `homCategory` is also tagged with `simps` instead of `simps!`, the two lemmas `id/comp_toNatTrans` that you added manually become unnecessary.

(You may try to make the same strategy work for `BasedFunctor`.)

joelriou: You missed the `initialize_simps_projections` part.

callesonne: Ah yes I did! So what is actually the point of this? I see that it allows me to write `@[simps]` on `id` instead of `@[simps toNatTrans]`, is that the main benefit?

joelriou: Yes. For all future declarations of based functors/nat. trans, `@[simps]` will work out of the box.

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,269 @@
...
+    isHomLift' := by
+      intro a
+      specialize isHomLift' a
+      rw [← NatIso.app_inv]
+      rw [← NatIso.app_hom] at isHomLift'
+      apply IsHomLift.lift_id_inv }
+
+/-- The inverse of a based natural transformation whose underlying natural tranformation carries an
+`IsIso` instance. -/
+noncomputable def mkNatIsIso {F G : 𝒳 ⥤ᵇ 𝒴} (α : F.toFunctor ⟶ G.toFunctor)
```
joelriou, line 216: Instead, could you prove that a based nat. trans is iso if the underlying nat. trans is. Then, `mkNatIsIso` would become basically unnecessary as it would essentially become a special of `asIso`.

callesonne: That's what I wanted to do originally, but I couldnt seem to get it to work. Both `[IsIso α.toNatTrans]` and `[IsIso (α.toNatTrans : F.toFunctor ⟶ G.toFunctor)]` give type errors (because `α.toNatTrans` is not recognized as a morphism in some category for some reason)

callesonne: Okay I figured out a trick :)

joelriou: I think it you be better to make it a lemma rather than an instance. (One approach would be to define a forget functor from based functors to functors, and phrase this by saying this functor reflects isomorphisms.)

callesonne: Why should it not be an instance? Will type-class inference not work? I will try to implement the approach with the forgetful functor now.

callesonne: I've left it there for now, but I'm happy to remove it if you think it's not useful.

joelriou: The "natural" direction is that if a based nat. trans is an iso, then the induced nat. trans is an iso: in this direction, this could certainly an instance. But, having instances in the other direction is problematic because of loops.

callesonne: Ohhhh I see! Thanks a lot! Then I will add an instance for the other direction instead

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,293 @@
...
+    (isHomLift' : ∀ a : 𝒳.obj, IsHomLift 𝒴.p (𝟙 (𝒳.p.obj a)) (α.hom.app a)) : F ≅ G where
+  hom := { toNatTrans := α.hom }
+  inv := {
+    toNatTrans := α.inv
+    isHomLift' := fun a ↦ by
+      have : 𝒴.p.IsHomLift (𝟙 (𝒳.p.obj a)) (α.app a).hom := (NatIso.app_hom α a) ▸ isHomLift' a
+      rw [← NatIso.app_inv]
+      apply IsHomLift.lift_id_inv }
+
+lemma isIso_of_toNatIsIso (α : F ⟶ G) [IsIso (X := F.toFunctor) α.toNatTrans] : IsIso α :=
```
joelriou, line 237:
```suggestion
lemma isIso_of_toNatTrans_isIso (α : F ⟶ G) [IsIso (X := F.toFunctor) α.toNatTrans] : IsIso α :=
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,293 @@
...
+
+open BasedFunctor BasedNatTrans
+
+section
+
+variable {𝒳 : BasedCategory.{v₂, u₂} 𝒮} {𝒴 : BasedCategory.{v₃, u₃} 𝒮}
+
+/-- Left-whiskering in the bicategory `BasedCategory` is given by whiskering the underlying functors
+and natural transformations. -/
+@[simps toNatTrans]
```
joelriou, line 253: In principle, this should no longer be necessary:
```suggestion
@[simps]
```

Mathlib/CategoryTheory/FiberedCategory/BasedCategory.lean
```diff
@@ -0,0 +1,293 @@
...
+def whiskerLeft {𝒵 : BasedCategory.{v₄, u₄} 𝒮} (F : 𝒳 ⥤ᵇ 𝒴) {G H : 𝒴 ⥤ᵇ 𝒵} (α : G ⟶ H) :
+    F ⋙ G ⟶ F ⋙ H where
+  toNatTrans := CategoryTheory.whiskerLeft F.toFunctor α.toNatTrans
+  isHomLift' := fun a ↦ α.isHomLift (F.w_obj a)
+
+/-- Right-whiskering in the bicategory `BasedCategory` is given by whiskering the underlying
+functors and natural transformations. -/
+@[simps toNatTrans]
```
joelriou, line 261:
```suggestion
@[simps]
```

joelriou:
Thanks!

bors merge

PR #13547, opened 2024-06-06
[Merged by Bors] - feat: order lemmas for NNRat

eric-wieser:
This copies the API for order lemmas on `Rat`

---

### DIFF
--- a/Mathlib/Data/Rat/Cast/Order.lean
+++ b/Mathlib/Data/Rat/Cast/Order.lean
@@ -6,6 +6,7 @@ Authors: Johannes Hölzl, Mario Carneiro
 import Mathlib.Algebra.Order.Ring.Rat
 import Mathlib.Data.Rat.Cast.CharZero
 import Mathlib.Tactic.Positivity.Core
+import Mathlib.Algebra.Order.Field.Basic
 
 #align_import data.rat.cast from "leanprover-community/mathlib"@"acebd8d49928f6ed8920e502a6c90674e75bd441"
 
@@ -153,10 +154,14 @@ namespace NNRat
 variable {K} [LinearOrderedSemifield K]
 
 theorem cast_strictMono : StrictMono ((↑) : ℚ≥0 → K) := fun m n h => by
-  rwa [NNRat.cast_def, NNRat.cast_def, div_lt_div_iff, ← Nat.cast_mul, ← Nat.cast_mul, Nat.cast_lt,
-    ← NNRat.lt_def]
-  -- · norm_num
-  -- · norm_num
+  rwa [NNRat.cast_def, NNRat.cast_def, div_lt_div_iff, ← Nat.cast_mul, ← Nat.cast_mul,
+    Nat.cast_lt (α := K), ← NNRat.lt_def]
+  · simp
+  · simp
+
+@[mono]
+theorem cast_mono : Monotone ((↑) : ℚ≥0 → K) :=
+  cast_strictMono.monotone
 
 /-- Coercion from `ℚ` as an order embedding. -/
 @[simps!]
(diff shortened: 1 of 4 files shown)


Mathlib/Data/Rat/Cast/Order.lean
```diff
@@ -171,4 +263,18 @@ def evalRatCast : PositivityExt where eval {u α} _zα _pα e := do
     return .nonzero q((Rat.cast_ne_zero (α := $α)).mpr $pa)
   | .none => pure .none
 
+/-- Extension for NNRat.cast. -/
+@[positivity NNRat.cast _]
+def evalNNRatCast : PositivityExt where eval {u α} _zα _pα e := do
```
YaelDillies, line 268: Can we have tests?

Mathlib/Data/Rat/Cast/Order.lean
```diff
@@ -148,6 +149,97 @@ theorem preimage_cast_uIoc (a b : ℚ) : (↑) ⁻¹' uIoc (a : K) b = uIoc a b
...
+theorem preimage_cast_Iio (a : ℚ≥0) : (↑) ⁻¹' Iio (a : K) = Iio a :=
+  castOrderEmbedding.preimage_Iio a
+
+@[simp]
+theorem preimage_cast_uIcc (a b : ℚ≥0) : (↑) ⁻¹' uIcc (a : K) b = uIcc a b :=
+  (castOrderEmbedding (K := K)).preimage_uIcc a b
+
+@[simp]
+theorem preimage_cast_uIoc (a b : ℚ≥0) : (↑) ⁻¹' uIoc (a : K) b = uIoc a b :=
+  (castOrderEmbedding (K := K)).preimage_uIoc a b
```
YaelDillies, line 239: Can we have the corresponding `image` lemmas for `NNRat -> Rat`?

Mathlib/Data/Rat/Cast/Order.lean
```diff
@@ -148,6 +149,97 @@ theorem preimage_cast_uIoc (a b : ℚ) : (↑) ⁻¹' uIoc (a : K) b = uIoc a b
...
+theorem cast_nonpos {n : ℚ≥0} : (n : K) ≤ 0 ↔ n ≤ 0 := by
+  norm_cast
+
+@[simp]
+theorem cast_pos {n : ℚ≥0} : (0 : K) < n ↔ 0 < n := by
+  norm_cast
+
+@[simp]
+theorem cast_lt_zero {n : ℚ≥0} : (n : K) < 0 ↔ n < 0 := by
+  norm_cast
```
YaelDillies, line 189: This looks useless

Mathlib/Data/Rat/Cast/Order.lean
```diff
@@ -148,6 +149,97 @@ theorem preimage_cast_uIoc (a b : ℚ) : (↑) ⁻¹' uIoc (a : K) b = uIoc a b
...
+
+@[simp, norm_cast]
+theorem cast_min {a b : ℚ≥0} : (↑(min a b) : K) = min (a : K) (b : K) :=
```
YaelDillies, line 192:
```suggestion
theorem cast_min (a b : ℚ≥0) : (↑(min a b) : K) = min (a : K) (b : K) :=
```
and below

Mathlib/Data/Rat/Cast/Order.lean
```diff
@@ -148,6 +149,97 @@ theorem preimage_cast_uIoc (a b : ℚ) : (↑) ⁻¹' uIoc (a : K) b = uIoc a b
...
+@[simp, norm_cast]
+theorem cast_lt {m n : ℚ≥0} : (m : K) < n ↔ m < n :=
+  cast_strictMono.lt_iff_lt
+
+@[simp]
+theorem cast_nonpos {n : ℚ≥0} : (n : K) ≤ 0 ↔ n ≤ 0 := by
+  norm_cast
+
+@[simp]
+theorem cast_pos {n : ℚ≥0} : (0 : K) < n ↔ 0 < n := by
```
YaelDillies, line 184: Please call elements of `Rat`/`NNRat` `p` or `q`, not `m` or `n`

eric-wieser: This is consistent with the naming in the rest of the file; I'd prefer to declare this out of scope.

Mathlib/Data/Rat/Cast/Order.lean
eric-wieser: This is consistent with the Rat versions of the lemmas; I'd prefer to declare this out of scope too.

Mathlib/Data/Rat/Cast/Order.lean
eric-wieser: I'd argue that unless the linter complains, it's not!

Mathlib/Data/Rat/Cast/Order.lean
eric-wieser: I'd claim "out of scope" here too; this PR just duplicates the existing Rat API, rather than adding new API.

Mathlib/Data/Rat/Cast/Order.lean
eric-wieser: Done

Mathlib/Data/Rat/Cast/Order.lean
YaelDillies: No but like there's no point blindly copying lemmas from one type to the other if they don't make sense for the new type. Both sides of your iff are always false, so you might as well replace it by `Not LHS`

eric-wieser: Ah, I guess the linter doesn't tell me when the RHS is further simplified.

eric-wieser: I've added your suggest but kept my original (without `simp`), because without this lemma `norm_cast` leaves some casts around.

YaelDillies: Uh okay interesting

Mathlib/Data/Rat/Cast/Order.lean
YaelDillies: I have opened #13644 to make your changes inconsistent with the Rat versions of the lemmas

eric-wieser: I'd appreciate it if you could push the corresponding changes to this PR once that lands; but it can wait till after your final exam!

eric-wieser: (or preferably, delegate this one first so that I can move on to the floorsemiring instance)

YaelDillies: Sure, I guess that's fixable later

maintainer merge

PR #13548, opened 2024-06-06
[Merged by Bors] - feat: add a FloorSemiring instance to NNRat

eric-wieser:
Also some missing `norm_cast` lemmas for inequalities of casts between different types.

---

- [x] depends on: #13547


Mathlib/Data/NNRat/Floor.lean
```diff
@@ -0,0 +1,60 @@
...
+Authors: Eric Wieser
+-/
+import Mathlib.Data.Rat.Floor
+
+/-!
+# Floor Function for Non-negative Rational Numbers
+
+## Summary
+
+We define the `FloorSemiring` instance on `ℚ≥0`, and relate it's operators to `NNRat.cast`.
```
YaelDillies, line 13:
```suggestion
We define the `FloorSemiring` instance on `ℚ≥0`, and relate its operators to `NNRat.cast`.
```

Mathlib/Data/NNRat/Floor.lean
```diff
@@ -0,0 +1,60 @@
...
+
+Note that we cannot talk about `Int.fract`, which currently only works for rings.
```
YaelDillies, line 15: Doesn't that mean we are missing `Nat.fract` with the obvious definition?

Mathlib/Data/NNRat/Floor.lean
```diff
@@ -0,0 +1,60 @@
...
+  rw [← Int.natCast_inj, NNRat.coe_floor, Rat.floor_def, Int.ofNat_ediv, den_coe, num_coe]
+
+variable {K} [LinearOrderedSemifield K] [FloorSemiring K]
+
+@[simp, norm_cast]
+theorem floor_cast (x : ℚ≥0) : ⌊(x : K)⌋₊ = ⌊x⌋₊ :=
+  (Nat.floor_eq_iff x.cast_nonneg).2 (mod_cast (Nat.floor_eq_iff x.cast_nonneg).1 (Eq.refl ⌊x⌋₊))
+
+@[simp, norm_cast]
+theorem ceil_cast (x : ℚ≥0) : ⌈(x : K)⌉₊ = ⌈x⌉₊ := by
```
YaelDillies, line 51: I think you are missing `intFloor_cast` and `intCeil_cast`

Mathlib/Data/Rat/Cast/Order.lean
YaelDillies, line 1: Can you add the same lemmas comparing `Nat.cast` and `Int.cast`? (I think they are also missing?)

Mathlib/Data/NNRat/Floor.lean
eric-wieser: Yeah, but adding new definitions seems out of scope for a PR that adds new instances.

Mathlib/Data/NNRat/Floor.lean
eric-wieser: Done

Mathlib/Data/Rat/Cast/Order.lean
eric-wieser: I want to declare this out of scope for a PR about NNRat.

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/Data/NNRat/Floor.lean
```diff
@@ -0,0 +1,88 @@
...
+namespace NNRat
+
+instance : FloorSemiring ℚ≥0 where
+  floor q := ⌊q.val⌋₊
+  ceil q := ⌈q.val⌉₊
+  floor_of_neg h := by simpa using h.trans zero_lt_one
+  gc_floor {a n} h := by rw [← NNRat.coe_le_coe, Nat.le_floor_iff] <;> norm_cast
+  gc_ceil {a b} := by rw [← NNRat.coe_le_coe, Nat.ceil_le]; norm_cast
+
+theorem floor_coe (q : ℚ≥0) : ⌊(q : ℚ)⌋₊ = ⌊q⌋₊ := rfl
```
urkud, line 31: Should these 2 lemmas be `simp`?

urkud:
Otherwise LGTM. Thanks!
bors d+

Mathlib/Data/NNRat/Floor.lean
eric-wieser: Sure, why not.

jcommelin (commented):
bors r+

sgouezel:
Can you merge master, fix the build, and send it back to bors?
bors d+

PR #13549, opened 2024-06-06
[Merged by Bors] - chore: move material on limits out of ConcreteCategory.Basic

kim-em:
This significantly reduces this imports of `ConcreteCategory.Basic` (the imports never should have been added; one day we will have CI help for this!).

There may actually be some duplication between the material I'm moving and the material in the destination file (which @joelriou added last week), but I'm going to declare deduplicating out of scope for this PR, please.


Mathlib/CategoryTheory/ConcreteCategory/Basic.lean
```diff
@@ -153,45 +150,6 @@ theorem ConcreteCategory.mono_of_injective {X Y : C} (f : X ⟶ Y) (i : Function
   (forget C).mono_of_mono_map ((mono_iff_injective f).2 i)
 #align category_theory.concrete_category.mono_of_injective CategoryTheory.ConcreteCategory.mono_of_injective
```
erdOne, line 151: Can you also move this one?
I think `import Mathlib.CategoryTheory.Functor.EpiMono` won't be needed after this.

erdOne: And you can use `assert_not_exists` to guarantee the imports aren't changed for now, but if you've got a better solution down the road feel free to ignore.

semorrison: Done, thanks for the suggestions!

joelriou:
Thanks!

bors merge

PR #13554, opened 2024-06-06
[Merged by Bors] - feat(Computability): define `Language.IsRegular`

madvorak:
This PR defines a predicate `Language.IsRegular` in a similar way `Language.IsContextFree` is defined.

---


madvorak:
Please check that the universe levels are done correctly!

erdOne:
We really need to move `IsRegular` (for regular elements) out from the root namespace...

madvorak:
> We really need to move `IsRegular` (for regular elements) out from the root namespace...

https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Regular/Basic.html#IsRegular

Yeah!

madvorak:
Can we proceed now regardless?

Mathlib/Computability/DFA.lean
```diff
@@ -28,11 +28,9 @@ Currently, there are two disjoint sets of simp lemmas: one for `DFA.eval`, and a
 - Should `mem_accepts` and `mem_acceptsFrom` be marked `@[simp]`?
 -/
 
-
-open Computability
-
 universe u v
 
+open Computability
```
YaelDillies, line 33: Actually the standard is to have
```
open ...
open scoped ...

universe ...
variable ...
```
It's important to have `open ...`/`open scoped ...` before `variable ...` because the latter might use some names or notations only accessible in a namespace, and `universe` is a very close cousin of `variable` so should stay next to it

Mathlib/Computability/DFA.lean
```diff
@@ -247,3 +245,7 @@ theorem comap_reindex (f : α' → α) (g : σ ≃ σ') :
 end Maps
 
 end DFA
+
+/-- Regular languages are defined by DFAs with finite states. -/
```
YaelDillies, line 249:
```suggestion
/-- A regular language is a language that is defined by a DFA with finite states. -/
```

Mathlib/Computability/DFA.lean
madvorak: OK, then it should probably make sense to replace
```
def Language.IsRegular {T : Type u}
```
by
```
def Language.IsRegular {T : Type*}
```
right?

madvorak: However, I am not sure the existentially-quantified type will accept `Type*` in place of `Type v`.

YaelDillies: All I'm saying is that you should revert the change I've commented on

madvorak: Did you take into account that the universes are used outside of the section as well as inside it?

Mathlib/Computability/DFA.lean
```diff
@@ -247,3 +245,7 @@ theorem comap_reindex (f : α' → α) (g : σ ≃ σ') :
 end Maps
 
 end DFA
+
+/-- A regular language is a language that is defined by a DFA with finite states. -/
+def Language.IsRegular {T : Type u} (L : Language T) : Prop :=
+  ∃ σ : Type v, ∃ _ : Fintype σ, ∃ M : DFA T σ, M.accepts = L
```
YaelDillies, line 251: This is a bad definition, in that the universe variable `v` only shows up in the body of the definition and not in the arguments. This is problematic as you then have countably many predicate `IsRegular.{0}`, `IsRegular.{1}`, `IsRegular.{2}`, ... which are all equivalent.

Please do
```suggestion
/-- A regular language is a language that is defined by a DFA with finite states. -/
def Language.IsRegular {T : Type u} (L : Language T) : Prop :=
  ∃ σ : Type, ∃ _ : Fintype σ, ∃ M : DFA T σ, M.accepts = L

lemma Language.isRegular_iff {T : Type u} {L : Language T}
    L.IsRegular ↔ ∃ σ : Type v, ∃ _ : Fintype σ, ∃ M : DFA T σ, M.accepts = L := sorry
```
which will involve using `Shrink` and importing `Mathlib.Data.Countable.Small`

madvorak: Unfortunately, I don't understand how `Shrink` works, so let's keep it at `proof_wanted` in this PR.

YaelDillies:
maintainer merge

mattrobball:
bors merge

madvorak:
FYI, I didn't have any conflict and, after the merge, the file `Mathlib/Computability/DFA.lean` still builds on my machine.

madvorak:
bors merge

mattrobball:
bors merge

PR #13556, opened 2024-06-06
[Merged by Bors] - ci: report changes to the import graph in a spoiler comment

jcommelin:
---

See https://github.com/leanprover-community/mathlib4/pull/13561 and https://github.com/leanprover-community/mathlib4/pull/13564 for an example of the output.


.github/workflows/import_count.yml
```diff
@@ -0,0 +1,57 @@
...
+        sudo apt-get install -y jq
+        # If you have additional dependencies, install them here
+
+    - name: Compute transitive imports
+      run: |
+        # Compute the counts for the HEAD of the PR
+        python ./scripts/count-trans-deps.py > head.json
+
+        # Checkout the merge base
+        git checkout $(git merge-base refs/remotes/origin/main ${{ github.sha }})
```
adomani, line 35:
```suggestion
        git checkout "$(git merge-base refs/remotes/origin/main ${{ github.sha }})"
```
I think that this is what shellcheck is saying.

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,57 @@
+name: Compute transitive imports
+
+on:
+  pull_request:
+    branches:
+      - main
```
adomani, line 6:
```suggestion
```
I am not sure, but this may not be needed/desired.

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        sudo apt-get install -y jq
+        # If you have additional dependencies, install them here
+
+    - name: Compute transitive imports
+      run: |
+        # Compute the counts for the HEAD of the PR
+        python ./scripts/count-trans-deps.py > head.json
```
adomani, line 32:
```suggestion
        python ./scripts/count-trans-deps.py <missing dir argument?> > head.json
```
Should there be a dir here, otherwise the python code seems to give an error.

jcommelin: Aah, last minute update that I forgot to fix.
```suggestion
        python ./scripts/count-trans-deps.py "Mathlib/" > head.json
```

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,57 @@
...
+
+        # Checkout the merge base
+        git checkout $(git merge-base refs/remotes/origin/main ${{ github.sha }})
+
+        # Compute the counts for the merge base
+        python ./scripts/count-trans-deps.py > base.json
```
jcommelin, line 38:
```suggestion
        python ./scripts/count-trans-deps.py "Mathlib/" > base.json
```

adomani:
### Import summary

Just trying to trick your script!  :smile:

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,69 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        sudo apt-get install -y jq
+        # If you have additional dependencies, install them here
+
+    - name: Get changed files
+      run: |
+        git fetch origin ${{ github.base_ref }}  # fetch the base branch
+        git diff --name-only origin/${{ github.base_ref }}... > changed_files.txt  # get the list of changed files
+        cat changed_files.txt
```
adomani, line 31:
```suggestion
```
Consider removing this.

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,69 @@
...
+      run: |
+        # the checkout dance, to avoid a detached head
+        git checkout master
+        git checkout -
+
+        # Compute the counts for the HEAD of the PR
+        python ./scripts/count-trans-deps.py "Mathlib/" > head.json
+
+        # Checkout the merge base
+        git checkout master # let's start easy! "$(git merge-base refs/remotes/origin/main ${{ github.sha }})"
```
adomani, line 43:
```suggestion
        git checkout "$(git merge-base master ${{ github.sha }})"
```
According to [stackoverflow](https://stackoverflow.com/questions/20808892/git-diff-between-current-branch-and-master-but-not-including-unmerged-master-com) this is the commit that `...` uses.  (I have not tested that it too much).

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,69 @@
...
+
+        # Compute the counts for the merge base
+        python ./scripts/count-trans-deps.py "Mathlib/" > base.json
+
+    - name: Compare results and comment
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        message=$(python ./scripts/import-graph-report.py base.json head.json changed_files.txt)
+        echo "${message}"
```
adomani, line 56:
```suggestion
```
This may also be unnecessary.

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,69 @@
...
+        data=$(jq -n --arg msg "$message" '{"body": $msg}')
+        method="POST"
+        if [[ -n "$message" ]]; then
+            url="https://api.github.com/repos/${GITHUB_REPOSITORY}/issues/${{ github.event.pull_request.number }}/comments"
+            headers="Authorization: token ${GITHUB_TOKEN}"
+            comments=$(curl -s -S -H "Content-Type: application/json" -H "$headers" "$url")
+            comment_id=$(echo "$comments" | jq -r '.[] | select(.body | startswith("### Import summary")) | .id')
```
adomani, line 63:
```suggestion
            comment_id=$(curl -s -S -H "Content-Type: application/json" -H "$headers" "$url" |
               jq -r '.[] | select(.body | startswith("### Import summary")) | .id')
```
Also, is `comment_id` actually a line-break-separated list of comment id's that fit the query, or is it just the first one?

scripts/import-graph-report.py
```diff
@@ -0,0 +1,70 @@
...
+    # Build the message
+    message = '### Import summary\n\n'
+    if messages:
+        message += '<details><summary>Dependency changes</summary>\n\n'
+        message += '| File | Base Count | Head Count | Change |\n'
+        message += '| --- | --- | --- | --- |\n'
+        message += '\n'.join(messages)
+        message += '\n</details>'
+    else:
+        message += 'None'
```
adomani, line 62:
```suggestion
        message += 'No changes to the import graph'
```
More explicit comment -- none could be interpreted as a failure of the tool.

bryangingechen:
I have no idea why the build failed on the latest commit. The error is:
> curl: (3) URL using bad/illegal format or missing URL

But it doesn't seem like the arguments to `curl` changed in any meaningful way?

adomani:
I think that the code fell into the trap that I had set: there were two matching id's, so `comment_id` was `id1\nid2` instead of `id1`.

I now made the script take the first matching id, hoping that since the bot posts on each pull request, its message will be the first (and likely only) beginning with `### Import summary`.

adomani (approved):
From my perspective, this PR is ready to merge!  The only potential question is whether we want it to compare to `master` or the place where it branched off.  I think that both have value and arguably, it is more interesting to know what is the import-difference with `master` than with that it was at the time of opening the PR.

I will not maintainer merge, though, since I have been quite involved in the development.

.github/workflows/import_count.yml
jcommelin: I think that `master` is actually what we should go with. For reasons that you explained in another comment.

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.8
```
Ruben-VandeVelde, line 19: Python is up to 3.12 by now; any reason to use such an old version?

scripts/count-trans-deps.py
```diff
@@ -0,0 +1,94 @@
...
+            if file in file_imports:
+                transitive_imports[file] = set(file_imports[file])
+                for import_module in file_imports[file]:
+                    if import_module in file_imports:
+                        transitive_imports[file].update(dfs(import_module))
+                    # else:
+                    #     print(f"Warning: {import_module} imported by {file} not found in file_imports")
+            else:
+                transitive_imports[file] = set()
+                # print(f"Warning: {file} not found in file_imports")
```
Ruben-VandeVelde, line 62: Do you want to keep these comments in?

.github/workflows/import_count.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         python ./scripts/count-trans-deps.py "Mathlib/" > head.json
 
         # Checkout the merge base
-        git checkout master # let's start easy! "$(git merge-base refs/remotes/origin/main ${{ github.sha }})"
+        git checkout master
```
Ruben-VandeVelde, line 42: I'm not sure this is a good idea - if an import is removed in master while the pr is open, I suspect this will blame the open PR for "adding" it back

jcommelin: Aaah, that's a good point.

.github/workflows/import_count.yml
jcommelin, line 19:
```suggestion
        python-version: 3.12
```

scripts/count-trans-deps.py
```diff
@@ -0,0 +1,94 @@
...
+    def dfs(file):
+        if file not in visited:
+            visited.add(file)
+            if file in file_imports:
+                transitive_imports[file] = set(file_imports[file])
+                for import_module in file_imports[file]:
+                    if import_module in file_imports:
+                        transitive_imports[file].update(dfs(import_module))
+                    # else:
+                    #     print(f"Warning: {import_module} imported by {file} not found in file_imports")
```
jcommelin, line 59:
```suggestion
```

scripts/count-trans-deps.py
jcommelin, line 62:
```suggestion
```

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,67 @@
...
+      run: |
+        # the checkout dance, to avoid a detached head
+        git checkout master
+        git checkout -
+
+        # Compute the counts for the HEAD of the PR
+        python ./scripts/count-trans-deps.py "Mathlib/" > head.json
+
+        # Checkout the merge base
+        git checkout master
```
jcommelin, line 42:
```suggestion
        git checkout "$(git merge-base master ${{ github.sha }})"
```

.github/workflows/import_count.yml
```diff
@@ -0,0 +1,67 @@
...
+        git checkout -
+
+        # Compute the counts for the HEAD of the PR
+        python ./scripts/count-trans-deps.py "Mathlib/" > head.json
+
+        # Checkout the merge base
+        git checkout "$(git merge-base master ${{ github.sha }})"
+        # get the files that we need from the current PR
+        # this will have to switch to `master` probably, when merging or right after
+        git checkout origin/import-count scripts/count-trans-deps.py scripts/import-graph-report.py
```
bryangingechen, line 45: We'll need a follow-up PR for this, right?

bryangingechen (approved):
LGTM. Thanks to @jcommelin for the PR and @adomani and @Ruben-VandeVelde for the reviews!

bors r+

.github/workflows/import_count.yml
adomani: I think so, though the code should still work as is, since the branch with the PR does not get deleted... I think.

In the long run, though, using `origin/master` means that the code will download the current version of the script, rather than this old one where the script originated.

PR #11806, opened 2024-03-31
[Merged by Bors] - feat: the derived category of an abelian category

joelriou:
The derived category of an abelian category is defined and it is shown that it is a triangulated category.

---
- [x] depends on: #11740
- [x] depends on: #11789
- [x] depends on: #11759
- [x] depends on: #11805
- [x] depends on: #11764
- [x] depends on: #11782
- [x] depends on: #11760
- [x] depends on: #11721
- [x] depends on: #11737
- [x] depends on: #11728
- [x] depends on: #11738
- [x] depends on: #11786 
- [x] depends on: #13558

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #11786, opened 2024-03-30
[Merged by Bors] - feat(CategoryTheory): localization of triangulated categories

joelriou:
In this PR, it is shown that the localization of a triangulated category is not only pretriangulated but also triangulated (if the class of morphisms has both calculus of left and right fractions). Pretriangulated and triangulated instances are set on the categories `W.Localization` and `W.Localization'`.

---

- [x] depends on: #11738
- [x] depends on: #11737
- [x] depends on: #11728
- [x] depends on: #13558

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13558, opened 2024-06-06
[Merged by Bors] - feat: mapComposableArrows is essentially surjective for localization functors when there is a calculus of fractions

joelriou:
If `L` is a localization functor with respect to a class of morphisms `W` that a left *or* right calculus of fractions, then the induced functors on `ComposableArrows _ n` is essentially surjective for all natural numbers `n`.

(This was initially written only for `n = 2` when there was both a calculus of left and right fractions, but Andrew Yang simplified the argument.)

Co-authored-by: Andrew Yang <the.erd.one@gmail.com>

---

This is used in the construction of the derived category #11806 and #11786


jcommelin (commented):
Thanks :tada:

bors merge

PR #13565, opened 2024-06-06
[Merged by Bors] - feat(Order/ModularLattice): alternative equality lemma

apnelson1:
We add a new lemma `eq_of_le_inf_le_of_le_sup` for proving two comparable elements are equal in a modular lattice.
It is the same as the existing `eq_of_le_inf_le_of_sup_le` but with the hypotheses in simp normal form, 
having`y ⊓ z ≤ x` and `y ≤ x ⊔ z` instead of `y ⊓ z ≤ x ⊓ z` and `y ⊔ z ≤ x ⊔ z`. The weaker hypotheses easily imply the stronger ones, and proving them in many applications involves slightly less boilerplate. 

We also golf the proof of the original lemma using the (shorter) proof of the new one. 

---

### DIFF
--- a/Mathlib/Order/ModularLattice.lean
+++ b/Mathlib/Order/ModularLattice.lean
@@ -238,7 +238,8 @@ theorem eq_of_le_of_inf_le_of_le_sup (hxy : x ≤ y) (hinf : y ⊓ z ≤ x) (hsu
 
 theorem eq_of_le_of_inf_le_of_sup_le (hxy : x ≤ y) (hinf : y ⊓ z ≤ x ⊓ z) (hsup : y ⊔ z ≤ x ⊔ z) :
     x = y :=
-  eq_of_le_of_inf_le_of_le_sup hxy (le_inf_iff.1 hinf).1 (sup_le_iff.1 hsup).1
+  eq_of_le_of_inf_le_of_le_sup hxy (hinf.trans inf_le_left) (le_sup_left.trans hsup)
+#align eq_of_le_of_inf_le_of_sup_le eq_of_le_of_inf_le_of_sup_le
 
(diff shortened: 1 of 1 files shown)


Mathlib/Order/ModularLattice.lean
```diff
@@ -230,19 +230,15 @@ theorem IsModularLattice.sup_inf_sup_assoc : (x ⊔ z) ⊓ (y ⊔ z) = (x ⊔ z)
...
-    have h : y ≤ x ⊔ z :=
-      calc
-        y ≤ y ⊔ z := le_sup_left
-        _ ≤ x ⊔ z := hsup
-    calc
-      y ≤ (x ⊔ z) ⊓ y := le_inf h le_rfl
-      _ = x ⊔ z ⊓ y := sup_inf_assoc_of_le _ hxy
-      _ ≤ x ⊔ z ⊓ x := sup_le_sup_left (by rw [inf_comm, inf_comm z]; exact hinf) _
-      _ ≤ x := sup_le le_rfl inf_le_right
-#align eq_of_le_of_inf_le_of_sup_le eq_of_le_of_inf_le_of_sup_le
```
alreadydone, line 245: The #align is getting removed ...

YaelDillies (commented):
Maybe you can simply delete the original since it is clearly strictly weaker? (you have used `le_inf_iff` and `sup_le_iff`, but you could have most directly done `hinf.trans inf_le_left`, `le_sup_left.trans hsup`)

apnelson1:
> Maybe you can simply delete the original since it is clearly strictly weaker? (you have used `le_inf_iff` and `sup_le_iff`, but you could have most directly done `hinf.trans inf_le_left`, `le_sup_left.trans hsup`)

I would prefer to leave it; it can be convenient, since inf_le_inf and sup_le_sup-type terms show up frequently when dealing with modular lattices.

YaelDillies:
Sure

maintainer merge

j-loreaux:
bors merge

PR #13571, opened 2024-06-06
[Merged by Bors] - feat: shifted morphisms in categories with a shift

joelriou:
Given a category `C` endowed with a shift by an additive monoid `M` and two objects `X` and `Y` in `C`, we consider the types `ShiftedHom X Y m` defined as `X ⟶ (Y⟦m⟧)` for all `m : M`, and the composition on these shifted hom.

---

### DIFF
--- a/Mathlib/CategoryTheory/Shift/ShiftedHom.lean
+++ b/Mathlib/CategoryTheory/Shift/ShiftedHom.lean
@@ -3,7 +3,7 @@ Copyright (c) 2024 Joël Riou. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Joël Riou
 -/
-import Mathlib.CategoryTheory.Shift.CommShift
+import Mathlib.CategoryTheory.Shift.Basic
 import Mathlib.CategoryTheory.Preadditive.AdditiveFunctor
 
 /-! Shifted morphisms
--- a/Mathlib/CategoryTheory/Shift/ShiftedHom.lean
+++ b/Mathlib/CategoryTheory/Shift/ShiftedHom.lean
@@ -1,0 +1,127 @@
+/-
+Copyright (c) 2024 Joël Riou. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joël Riou
+-/
+import Mathlib.CategoryTheory.Shift.CommShift
+import Mathlib.CategoryTheory.Preadditive.AdditiveFunctor
+
+/-! Shifted morphisms
+
+Given a category `C` endowed with a shift by an additive monoid `M` and two
+objects `X` and `Y` in `C`, we consider the types `ShiftedHom X Y m`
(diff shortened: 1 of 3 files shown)


Mathlib/CategoryTheory/Shift/ShiftedHom.lean
```diff
@@ -0,0 +1,126 @@
...
+
+/-- The bijection `(X ⟶ Y) ≃ ShiftedHom X Y m₀` when `m₀ = 0`. -/
+@[simps apply]
+noncomputable def homEquiv (m₀ : M) (hm₀ : m₀ = 0) : (X ⟶ Y) ≃ ShiftedHom X Y m₀ where
+  toFun f := mk₀ m₀ hm₀ f
+  invFun g := g ≫ (shiftFunctorZero' C m₀ hm₀).hom.app Y
+  left_inv f := by simp [mk₀]
+  right_inv g := by simp [mk₀]
+
+lemma mk₀_comp (m₀ : M) (hm₀ : m₀ = 0) (f : X ⟶ Y) {a : M} (g : ShiftedHom Y Z a) :
```
erdOne, line 73: Should this be a simp lemma?

Mathlib/CategoryTheory/Shift/ShiftedHom.lean
```diff
@@ -0,0 +1,126 @@
...
+    (mk₀ m₀ hm₀ f).comp g (by rw [hm₀, add_zero]) = f ≫ g := by
+  subst hm₀
+  simp [comp, mk₀, shiftFunctorAdd'_add_zero_inv_app, shiftFunctorZero']
+
+@[simp]
+lemma mk₀_id_comp (m₀ : M) (hm₀ : m₀ = 0) {a : M} (f : ShiftedHom X Y a) :
+    (mk₀ m₀ hm₀ (𝟙 X)).comp f (by rw [hm₀, add_zero]) = f := by
+  simp [mk₀_comp]
+
+lemma comp_mk₀ {a : M} (f : ShiftedHom X Y a) (m₀ : M) (hm₀ : m₀ = 0) (g : Y ⟶ Z) :
```
erdOne, line 83: Also this.

Mathlib/CategoryTheory/Shift/ShiftedHom.lean
joelriou: I would rather not make these simp lemmas. If I had defined `ShiftedHom` as a single field `hom` structure, it would have very much made sense to have such a simp(s) lemma `mk₀_comp_hom`, but otherwise, I prefer not to. (I will use a single field structure when I repackage this for Ext-groups #13599.)

erdOne: Fair enough.

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13577, opened 2024-06-06
feat: the infinite adele ring of a number field

smmercuri:
This PR contains the following:
- The completion of a number field at an infinite place (`NumberTheory.NumberField.Completion`).
- The proof that the completion of a number field at an infinite place is locally compact (`NumberTheory.NumberField.Completion`).
- Proofs that the completion of a number field place is isomorphic to the complex/real numbers according to whether the infinite place is complex/real (`NumberTheory.NumberField.Completion`).
- The infinite adele ring defined as the locally compact finite product of the completions of a number field over its infinite places (`NumberTheory.NumberField.InfiniteAdeleRing`).
- Helper result: the pullback of a completable topological field under a field embedding defines a completable topological field (`Topology.Algebra.UniformField`).
- Helper result: the uniform completion of a completable topological normed field is a normed field (`Analysis.NormedSpace.Completion`).

---

- [x] depends on: #15064
- [x] depends on: #15068
- [x] depends on: #15075
- [x] depends on: #15094
- [x] depends on: #16483


Mathlib/Analysis/NormedSpace/Completion.lean
```diff
@@ -111,6 +112,25 @@ instance [SeminormedCommRing A] [NormedAlgebra 𝕜 A] [UniformContinuousConstSM
...
 
+instance instNormedField [NormedField A] [CompletableTopField A] :
+    NormedField (UniformSpace.Completion A) where
+  toField := UniformSpace.Completion.instField
+  dist_eq := fun x y => by
+      refine Completion.induction_on₂ x y ?_ ?_ <;> clear x y
+      · refine isClosed_eq (Completion.uniformContinuous_extension₂ _).continuous ?_
+        exact Continuous.comp Completion.continuous_extension continuous_sub
+      · intro x y
+        rw [← Completion.coe_sub, norm_coe, Completion.dist_eq, dist_eq_norm]
```
Ruben-VandeVelde, line 123:
```suggestion
  dist_eq x y := by
    refine Completion.induction_on₂ x y ?_ ?_ <;> clear x y
    · refine isClosed_eq (Completion.uniformContinuous_extension₂ _).continuous ?_
      exact Continuous.comp Completion.continuous_extension continuous_sub
    · intro x y
      rw [← Completion.coe_sub, norm_coe, Completion.dist_eq, dist_eq_norm]
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,143 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+noncomputable def uniformSpace : UniformSpace K :=
+  v.normedField.toUniformSpace
+
+/-- The uniform additive group structure of a number field induced by an infinite place. -/
+theorem uniformAddGroup : @UniformAddGroup K v.uniformSpace _ :=
+  @SeminormedAddCommGroup.to_uniformAddGroup _ (
+    @NormedAddCommGroup.toSeminormedAddCommGroup _ (
+      @NonUnitalNormedRing.toNormedAddCommGroup _ (
+        @NormedRing.toNonUnitalNormedRing _ (
+          @NormedDivisionRing.toNormedRing _ v.normedDivisionRing))))
```
Ruben-VandeVelde, line 1035: Can this be
```suggestion
theorem uniformAddGroup : @UniformAddGroup K v.uniformSpace _ := by
  letI := v.normedDivisionRing
  infer_instance
```
?

Ruben-VandeVelde (commented):
I don't know anything about the mathematics, so just some superficial comments

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: yes, that's much nicer than chasing instances, thanks!

Mathlib/Analysis/NormedSpace/Completion.lean
smmercuri: Fixed, and also for the `norm_mul'` field

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,140 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+
+/-- The topological division ring structure of a number field induced by an infinite place. -/
+theorem topologicalDivisionRing :
+    @TopologicalDivisionRing K _ v.topologicalSpace :=
+  v.normedDivisionRing.to_topologicalDivisionRing
+
+/-- The topological ring structure of a number field induced by an infinite place. -/
+theorem topologicalRing : @TopologicalRing K v.topologicalSpace _ :=
+  @TopologicalDivisionRing.toTopologicalRing _ _ v.topologicalSpace
+    v.topologicalDivisionRing
```
erdOne, line 1046: Are these really needed?

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,140 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+
+/-- The embedding associated to an infinite place is a uniform embedding. -/
+theorem uniformEmbedding : @UniformEmbedding _ _ v.uniformSpace _ v.embedding := by
```
erdOne, line 1049:
```suggestion
theorem uniformEmbedding : letI := v.normedField; UniformEmbedding v.embedding := by
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,140 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+  @UniformInducing.uniformContinuous _ _ v.uniformSpace _ _ v.uniformInducing
+
+/-- The embedding associated to an infinite place of a number field is continuous. -/
+theorem continuous : @Continuous _ _ v.topologicalSpace _ v.embedding :=
+  @UniformContinuous.continuous _ _ v.uniformSpace _ _ v.uniformContinuous
+
+/-- The uniform structure induced by an infinite place of a number field defines a
+completable topological field. -/
+theorem completableTopField : @CompletableTopField K _ v.uniformSpace :=
+  @UniformSpace.comap_completableTopField _ _ _ _ _ _ v.normedField.instT0Space _
```
erdOne, line 1072: Similarly, I don't think these are needed. The point of type class inference is to find these for you.

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,140 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+
+/-- The completion of a number field at an infinite place. -/
+def completion := @UniformSpace.Completion K v.normedDivisionRing.toUniformSpace
```
erdOne, line 1075:
```suggestion
def completion := letI := v.normedSpace; UniformSpace.Completion K
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,140 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
       Finset.card (Finset.univ.filter <| IsUnramifiedIn K (k := k))ᶜ * (finrank k K / 2) := by
   rw [← card_isUnramified, ← card_isUnramified_compl, Finset.card_add_card_compl]
 
+variable {K : Type*} [Field K] [NumberField K] (v : InfinitePlace K)
+
+noncomputable instance : Inhabited (InfinitePlace K) :=
+  ⟨Classical.choice (instNonemptyInfinitePlaceOfNumberField K)⟩
```
erdOne, line 1013: Why do you need this (and not `Nonempty`)

erdOne (commented):
I only reviewed the first half of that file, but in general please use `letI` instead of excessive lemmas or `@`-notations.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
```diff
@@ -0,0 +1,69 @@
+/-
+Copyright (c) 2024 Salvatore Mercuri, María Inés de Frutos-Fernández. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Salvatore Mercuri, María Inés de Frutos-Fernández
```
erdOne, line 4: BTW is this PR a collaboration with María? You may add "Co-authored-by" in the PR message.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
```diff
@@ -0,0 +1,69 @@
...
+end DerivedInstances
+
+instance topologicalSpace : TopologicalSpace (infiniteAdeleRing K) := Pi.topologicalSpace
+
+instance topologicalRing : TopologicalRing (infiniteAdeleRing K) := Pi.instTopologicalRing
+
+/-- The global embedding of a number field into its infinite adele ring,
+sending `x ∈ K` to `(x)ᵥ`. -/
+def globalEmbedding : K →+* infiniteAdeleRing K :=
+  Pi.ringHom (fun (v : InfinitePlace K) => InfinitePlace.Completion.coeRingHom v)
```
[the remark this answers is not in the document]
smmercuri: Yes, have now added the algebra instance and simp lemma, also algebra instance for `v.completion`.

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: I had these in because I couldn't get the type class inferencing to work, but the `letI` approach resolves this nicely and tidies it up a lot, thanks!

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: Fixed.

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: The first four are gone now. I have left `completableTopField` separate there since it has some non-trivial proof going on in `comap_completableTopField`, which itself is included elsewhere in this PR. But this could also be removed with a `haveI` in the `NormedField v.completion` instance further down, if preferred.

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: Fixed

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: This instance is used for the `Nontrivial` instance in `InfiniteAdeleRing.lean`, which itself is needed for injectivity of the global embedding.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
smmercuri: This is not a direct collaboration as such, but María Inés originally formalised the infinite adele ring with a different approach a couple of years ago before we had the embeddings API, and has helped with advice on my current approach which is why I put the file as co-authored, but the code in this PR is all my own. Not sure what the norm for attribution in this case would be!

erdOne: Fair enough. I thought María Inés had done this a while ago and was about to ask you to collaborate with her until I saw this.

Mathlib/NumberTheory/NumberField/Embeddings.lean
erdOne: Then you should use `Pi.nontrivial_at` there instead. Inhabited is usually only used if there is a distinguished element.

Mathlib/NumberTheory/NumberField/Embeddings.lean
erdOne: `comap_completableTopField` could be an instance. Then this is not necessary.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
```diff
@@ -0,0 +1,70 @@
...
+
+instance topologicalSpace : TopologicalSpace (infiniteAdeleRing K) := Pi.topologicalSpace
+
+instance topologicalRing : TopologicalRing (infiniteAdeleRing K) := Pi.instTopologicalRing
+
+instance : Algebra K (infiniteAdeleRing K) := Pi.algebra _ _
+
+/-The global embedding of a number field into its infinite adele ring,
+sending `x ∈ K` to `(x)ᵥ`. -/
+def globalEmbedding : K →+* infiniteAdeleRing K := Algebra.toRingHom
```
erdOne, line 59:
```suggestion
/-- The global embedding of a number field into its infinite adele ring,
sending `x ∈ K` to `(x)ᵥ`. -/
abbrev globalEmbedding : K →+* infiniteAdeleRing K := algebraMap K (infiniteAdeleRing K)
```
`algebraMap` has more API support.

Mathlib/Topology/Algebra/UniformField.lean
```diff
@@ -177,6 +177,28 @@ instance : TopologicalDivisionRing (hat K) :=
 
 end Completion
 
+/-- The pullback of a completable topological field along a ring homomorphism
+is a completable topological field. -/
+theorem comap_completableTopField {α β : Type*}
+    [Field β] [Field α] {f : α →+* β} [b : UniformSpace β]
+    [@T0Space α (UniformSpace.comap f b).toTopologicalSpace] [CompletableTopField β] :
+    @CompletableTopField _ _ (UniformSpace.comap f b) := by
+  have h_ui : @UniformInducing _ _ (UniformSpace.comap f b) _ f := by
```
erdOne, line 186: You can also
`have inst := UniformSpace.comap f b` in the proof to get rid of the `@`s.

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,77 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+noncomputable instance : NormedField v.completion :=
+  letI := v.normedField; letI := v.completableTopField; UniformSpace.Completion.instNormedField K
+
+instance : CompleteSpace v.completion :=
+  letI := v.normedField; UniformSpace.Completion.completeSpace K
+
+noncomputable instance : Inhabited v.completion := ⟨0⟩
+
+noncomputable instance : Coe K v.completion :=
+  letI := v.normedField; (inferInstance : Coe K (UniformSpace.Completion K))
```
erdOne, line 1044:
```suggestion
  letI := v.normedField; inferInstanceAs (Coe K (UniformSpace.Completion K))
```
Looks nicer though it saves no chars.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
smmercuri: Fixed.

Mathlib/Topology/Algebra/UniformField.lean
smmercuri: Done.

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: Gotcha, changed that and removed `Inhabited`.

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: Makes sense, fixed that.

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,74 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+  letI := v.normedField; ⟨uniformInducing_iff_uniformSpace.2 rfl, v.embedding.injective⟩
+
+/-- The completion of a number field at an infinite place. -/
+def completion :=
+  letI := v.normedField; UniformSpace.Completion K
+
+namespace Completion
+
+instance : NormedField v.completion :=
+  letI := v.normedField; letI := UniformSpace.comap_completableTopField v.embedding;
```
erdOne, line 1030:
```suggestion
  letI := v.normedField
  letI := UniformSpace.comap_completableTopField v.embedding
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,74 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+  · rw [extensionEmbedding, UniformSpace.Completion.extensionHom, RingHom.coe_mk,
+      MonoidHom.coe_mk, OneHom.coe_mk, UniformSpace.Completion.dist_eq]
+    simp only [UniformSpace.Completion.extension_coe v.uniformEmbedding.uniformContinuous]
+    exact Isometry.dist_eq v.uniformEmbedding.to_isometry _ _
+
+variable (v)
+
+/-- The embedding `v.completion → ℂ` is a closed embedding. -/
+theorem closedEmbedding : ClosedEmbedding (extensionEmbedding v) :=
+  (Isometry.of_dist_eq extensionEmbedding_dist_eq).closedEmbedding
```
erdOne, line 1068: I think `Isometry.of_dist_eq extensionEmbedding_dist_eq` deserves its own lemma.

Mathlib/NumberTheory/NumberField/InfiniteAdeleRing.lean
```diff
@@ -0,0 +1,71 @@
...
+
+/-- The global embedding of a number field into its infinite adele ring,
+sending `x ∈ K` to `(x)ᵥ`. -/
+abbrev globalEmbedding : K →+* infiniteAdeleRing K := algebraMap K (infiniteAdeleRing K)
+
+@[simp]
+theorem globalEmbedding_apply (x : K) : globalEmbedding K x v = (x : v.completion) := rfl
+
+/-- The infinite adele ring is locally compact. -/
+theorem locallyCompactSpace : LocallyCompactSpace (infiniteAdeleRing K) :=
```
erdOne, line 66:
```suggestion
instance locallyCompactSpace : LocallyCompactSpace (infiniteAdeleRing K) :=
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,74 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+namespace Completion
+
+instance : NormedField v.completion :=
+  letI := v.normedField; letI := UniformSpace.comap_completableTopField v.embedding;
+  UniformSpace.Completion.instNormedField K
+
+instance : CompleteSpace v.completion :=
+  letI := v.normedField; UniformSpace.Completion.completeSpace K
+
+instance : Inhabited v.completion := ⟨0⟩
```
erdOne, line 1036: Is this needed?

Mathlib/Topology/Algebra/UniformField.lean
```diff
@@ -177,6 +177,22 @@ instance : TopologicalDivisionRing (hat K) :=
 
 end Completion
 
+/-- The pullback of a completable topological field along a ring homomorphism
+is a completable topological field. -/
+instance comap_completableTopField {α β : Type*}
```
erdOne, line 182: After thinking about this a bit more, this should be stated in terms of `Inducing`.

smmercuri: Do you mean in terms of `UniformInducing`?

Mathlib/NumberTheory/NumberField/Embeddings.lean
smmercuri: I don't use it specifically, but seemed like a standard instance to include in this case. Mainly, I saw that it was included for the finite completions [`IsDedekindDomain.HeightOneSpectrum.instInhabitedAdicCompletion`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/DedekindDomain/AdicValuation.html#IsDedekindDomain.HeightOneSpectrum.instInhabitedAdicCompletion)

Mathlib/Topology/Algebra/UniformField.lean
erdOne: I initially meant `Inducing` but I guess `UniformInducing` is the same thing?

smmercuri: It would need to be `UniformInducing` I think, since this guarantees the uniform structure is the pullback uniform structure, whereas `Inducing` only guarantees that the topological space is the pullback.

smmercuri: I have addressed this in the latest commit, but when I tried to make `UniformInducing.completableTopField` an instance I get the following error: 
```cannot find synthesization order for instance @completableTopField with type
  ∀ {α : Type u_3} {β : Type u_4} [inst : Field β] [b : UniformSpace β] [inst_1 : CompletableTopField β]
    [inst_2 : Field α] {f : α →+* β} [inst_3 : UniformSpace α] [inst_4 : T0Space α],
    UniformInducing ⇑f → CompletableTopField α
all remaining arguments have metavariables:
  Field ?β
  UniformSpace ?β
  @CompletableTopField ?β ?inst✝ ?b```
```
Not sure if that is avoidable, it seems unable to infer instance for `β`. For now, I have turned this into a `theorem` and kept the `comap_completableTopField` as an instance.

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,80 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+
+/-- The completion of a number field at an infinite place. -/
+def completion :=
+  letI := v.normedField; UniformSpace.Completion K
+
+namespace Completion
+
+instance : NormedField v.completion :=
+  letI := v.normedField
+  letI := UniformSpace.comap_completableTopField v.embedding
```
erdOne, line 1031:
```suggestion
```
Should be possible as it is already an instance.

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,79 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+
+variable (v)
+
+/-- The embedding `v.completion → ℂ` is an isometry. -/
+theorem isometry_extensionEmbedding :
+    Isometry (extensionEmbedding v) :=
+  Isometry.of_dist_eq extensionEmbedding_dist_eq
+
+/-- The embedding `v.completion → ℂ` is a closed embedding. -/
+theorem closed_extensionEmbedding: ClosedEmbedding (extensionEmbedding v) :=
```
erdOne, line 1072:
```suggestion
theorem closedEmbedding_extensionEmbedding: ClosedEmbedding (extensionEmbedding v) :=
```

Mathlib/NumberTheory/NumberField/Embeddings.lean
```diff
@@ -1005,6 +1007,79 @@ lemma card_eq_card_isUnramifiedIn [NumberField k] [IsGalois k K] :
...
+  (isometry_extensionEmbedding v).closedEmbedding
+
+/-- The completion of a number field at an infinite place is locally compact. -/
+instance locallyCompactSpace : LocallyCompactSpace (v.completion) :=
+  (closed_extensionEmbedding v).locallyCompactSpace
```
erdOne, line 1077:
```suggestion
  (closedEmbedding_extensionEmbedding v).locallyCompactSpace
```

erdOne (approved):
LGTM. 
Maybe @faenuccio would like to have a look?

faenuccio:
> * he uniform completion of a completable topological normed field is a normed field

I am having a look, and I am trying to understand to what extent some of the arguments are generalizable to non-archimedean completions as well and if there is room for common arguments/code.

faenuccio:
> Actually, now that I think about it, there is an argument to be made that we should define the completion of `K` at the infinite place `w` to be either `ℝ` or `ℂ` and then prove that it has the right properties. That might make things easier later on. I guess it all depends on what applications we have in mind but it is, I think, something that should be discussed. In this case, the `infiniteAdeleRing` would be essentially the same as the mixed space that appears here [NumberField.mixedEmbedding](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/NumberField/CanonicalEmbedding/Basic.html#NumberField.mixedEmbedding)

Yes, I agree. As it stands, it seems we are defining the same object twice.

mariainesdff:
> Actually, now that I think about it, there is an argument to be made that we should define the completion of `K` at the infinite place `w` to be either `ℝ` or `ℂ` and then prove that it has the right properties

This sounds reasonable to me. @kbuzzard , what do you think?

smmercuri:
> Actually, now that I think about it, there is an argument to be made that we should define the completion of `K` at the infinite place `w` to be either `ℝ` or `ℂ` and then prove that it has the right properties. That might make things easier later on. I guess it all depends on what applications we have in mind but it is, I think, something that should be discussed. In this case, the `infiniteAdeleRing` would be essentially the same as the mixed space that appears here [NumberField.mixedEmbedding](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/NumberField/CanonicalEmbedding/Basic.html#NumberField.mixedEmbedding)

I thought about doing something similar while working on this and it is certainly an option. I guess one argument for the approach in this PR is that `w.completion` inherits all the `UniformSpace.Completion` API. E.g., any uniform continuous function on `K` can then be extended to a uniform continuous function on `w.completion`, which I could imagine might be useful further down the line for functions on adele rings. But perhaps that can also be achieved with the `mixedEmbedding` approach. Anyway I agree that, if we do proceed with this PR, at some point we should show that `w.completion` is isomorphic to `ℝ` or `ℂ`, and therefore that the infinite adele ring is isomorphic to `ℝ ^ r₁ × ℂ ^ r₂`.

faenuccio:
And I think that it makes more sense to decide once and for all which of the two approaches we take, rather than using both and then show an equivalence. @xroblot , do you think that it would be possible to replace the `mixed space` approach with the infinite adeles and working with them altogether?

erdOne:
What does "define the completion of K at the infinite place w to be either ℝ or ℂ" mean?
The most naive approach
```
def NumberField.InfinitePlace.Completion : Type _ :=
  letI := Classical.propDecidable; if v.IsComplex then ℂ else ℝ
```
isn't defeq to  `ℂ` nor `ℝ` and you'll have to manually define all the instances. Hence I don't see much benefit of it.

Or do you mean that we shouldn't have this def at all and just case split on `IsComplex` and work with ℝ or ℂ directly (as in the mixed-space)?

erdOne:
Thinking about this a bit more, we should have `v.Completion` for every `v : AbsoluteValue K ℝ` and `NumberField.InfinitePlace.Completion` should just be `v.1.Completion`.

xroblot:
Yes, I think Andrew makes a good point. There is probably no nice way to define the completion as being `ℝ` or `ℂ` and working separately on real and complex places (as it's done for `mixedEmbedding`) is very messy and cumbersome.

mariainesdff:
> And I think that it makes more sense to decide once and for all which of the two approaches we take, rather than using both and then show an equivalence. @xroblot , do you think that it would be possible to replace the `mixed space` approach with the infinite adeles and working with them altogether?

Yes, I think the key point is that we should agree on a "main" approach (although I would not be opposed to showing the equivalence to the other one).

xroblot:
Here is what I would suggest:

1. Make a new file `Completion`
2. Prove everything that you can in the general case following Andrew's suggestion above
3. Specialize to the infinite places for the results that do not hold in the general case
4. Prove that the combination of the "right" combination of maps give an isomorphism between the completion at an infinite place and ℝ or ℂ (you should take a look at the proof of [NumberField.InfinitePlace.mk_eq_iff](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/NumberField/Embeddings.html#NumberField.InfinitePlace.mk_eq_iff))
5. Define the infinite Adele ring as you did and prove that it is isomorphic (using the maps in step 4) to the mixed space.

xroblot:
> And I think that it makes more sense to decide once and for all which of the two approaches we take, rather than using both and then show an equivalence. @xroblot , do you think that it would be possible to replace the `mixed space` approach with the infinite adeles and working with them altogether?

I am not sure it would be a good idea to replace the mixed space by the infinite adele ring. Most of the things we do with the mixed space is volume computations and for that it is better to have a very explicit construction for it. And defining the infinite adele ring as the mixed space does not seem such a good ideal after all too. 
At this point, I think it is better to have two distinct objects and some kind of "canonical way" to go from one to the other as I suggested above.

xroblot:
> OK, for this you are certainly more competent, so I will be happy to follow your advice, although it would have been nice to observe that the infinite adeles come with a Haar measure and one can try to use it for the (co)volume computations. 

Well, actually, it is the other way around. One should prove that the map from the infinite adeles to the mixed space is measure preserving to use the fact that volume computations are easier in the mixed space.

faenuccio:
> Thanks for the suggestions all, and @xroblot that plan makes a lot of sense. I will follow it! I assume we are OK with leaving this PR open while I do?
> 
> And yes agree with @faenuccio it would definitely be nice to have the Haar measure for the infinite adele ring, but assume this will also be able to be transferred to it from establishing an appropriate isomorphism between it and the mixed embedding.

Yes, you can add the "WIP" label, and then leave it open while you work on it.

smmercuri:
@xroblot @erdOne this is now ready for review again following the suggestions and tips

riccardobrasca:
The PR is quite big, is it possible to split it?

PR #13576, opened 2024-06-06
[Merged by Bors] - chore(AddChar): Rename `map_zero_one` to `map_zero_eq_one`

YaelDillies:
To follow the naming convention, rename
* `map_zero_one` → `map_zero_eq_one`
* `map_add_mul` → `map_add_eq_mul`
* `map_nsmul_pow` → `map_nsmul_eq_pow`
* `map_neg_inv` → `map_neg_eq_inv`
* `map_zsmul_zpow` → `map_zsmul_eq_zpow`

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13579, opened 2024-06-06
[Merged by Bors] - feat(AddChar): Various improvements

YaelDillies:
* Add a bunch of missing `simp` and `norm_cast` lemmas.
* Get rid of `IsNontrivial ψ` since it's just `ψ ≠ 1`. This simplifies proofs.
* Remove explicit arguments to `toMonoidHomEquiv`/`toAddMonoidHomEquiv`. We basically never need to provide them explicitly since unification usually does the job.

---

- [x] depends on: #13576

### DIFF
--- a/Mathlib/Algebra/Group/AddChar.lean
+++ b/Mathlib/Algebra/Group/AddChar.lean
@@ -155,6 +155,7 @@ def toAddMonoidHom (φ : AddChar A M) : A →+ Additive M where
   map_add' := φ.map_add_eq_mul'
 
 @[simp] lemma coe_toAddMonoidHom (ψ : AddChar A M) : ⇑ψ.toAddMonoidHom = Additive.ofMul ∘ ψ := rfl
+
 @[simp] lemma toAddMonoidHom_apply (ψ : AddChar A M) (a : A) :
     ψ.toAddMonoidHom a = Additive.ofMul (ψ a) := rfl
 
@@ -213,7 +214,7 @@ lemma _root_.MonoidHom.compAddChar_injective_right (f : M →* N) (hf : Injectiv
 
 /-- Composing an `AddChar` with an `AddMonoidHom` yields another `AddChar`. -/
 def compAddMonoidHom (φ : AddChar B M) (f : A →+ B) : AddChar A M :=
-  (toAddMonoidHomEquiv).symm (φ.toAddMonoidHom.comp f)
+  toAddMonoidHomEquiv.symm (φ.toAddMonoidHom.comp f)
 
 @[simp, norm_cast]
 lemma coe_compAddMonoidHom (φ : AddChar B M) (f : A →+ B) : φ.compAddMonoidHom f = φ ∘ f := rfl
@@ -227,7 +228,8 @@ lemma compAddMonoidHom_injective_left (f : A →+ B) (hf : Surjective f) :
 
 lemma compAddMonoidHom_injective_right (ψ : AddChar B M) (hψ : Injective ψ) :
     Injective fun f : A →+ B ↦ ψ.compAddMonoidHom f := by
-  rintro f g h; rw [DFunLike.ext'_iff] at h ⊢; exact hψ.comp_left h
+  rintro f g h
+  rw [DFunLike.ext'_iff] at h ⊢; exact hψ.comp_left h
 
 lemma eq_one_iff : ψ = 1 ↔ ∀ x, ψ x = 1 := DFunLike.ext_iff
 lemma ne_one_iff : ψ ≠ 1 ↔ ∃ x, ψ x ≠ 1 := DFunLike.ne_iff
--- a/Mathlib/Algebra/Group/AddChar.lean
+++ b/Mathlib/Algebra/Group/AddChar.lean
@@ -34,7 +34,7 @@ additive character
 ### Definitions related to and results on additive characters
 -/
 
-open Multiplicative
+open Function Multiplicative
 
 section AddCharDef
 
@@ -74,7 +74,7 @@ namespace AddChar
 section Basic
 -- results which don't require commutativity or inverses
 
-variable {A M : Type*} [AddMonoid A] [Monoid M]
+variable {A B M N : Type*} [AddMonoid A] [AddMonoid B] [Monoid M] [Monoid N] {ψ : AddChar A M}
 
 /-- Define coercion to a function. -/
 instance instFunLike : FunLike (AddChar A M) A M where
@@ -125,7 +125,6 @@ lemma map_nsmul_eq_pow (ψ : AddChar A M) (n : ℕ) (x : A) : ψ (n • x) = ψ 
 
 @[deprecated (since := "2024-06-06")] alias map_nsmul_pow := map_nsmul_eq_pow
 
-variable (A M) in
 /-- Additive characters `A → M` are the same thing as monoid homomorphisms from `Multiplicative A`
 to `M`. -/
 def toMonoidHomEquiv : AddChar A M ≃ (Multiplicative A →* M) where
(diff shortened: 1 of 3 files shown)


Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -74,7 +74,7 @@ namespace AddChar
 section Basic
 -- results which don't require commutativity or inverses
 
-variable {A M : Type*} [AddMonoid A] [Monoid M]
+variable {A B M N : Type*} [AddMonoid A] [AddMonoid B] [Monoid M] [Monoid N] {ψ : AddChar A M}
```
loefflerd, line 77: Seems a bit strange to declare `{ψ : AddChar A M}` as standing notation when it is only used twice much further down (most lemmas take the explicit argument)

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -137,15 +136,28 @@ def toMonoidHomEquiv : AddChar A M ≃ (Multiplicative A →* M) where
...
+
 /-- Interpret an additive character as a monoid homomorphism. -/
 def toAddMonoidHom (φ : AddChar A M) : A →+ Additive M where
   toFun := φ.toFun
   map_zero' := φ.map_zero_eq_one'
   map_add' := φ.map_add_eq_mul'
 
-@[simp] lemma toAddMonoidHom_apply (ψ : AddChar A M) (a : A) : ψ.toAddMonoidHom a = ψ a := rfl
+@[simp] lemma coe_toAddMonoidHom (ψ : AddChar A M) : ⇑ψ.toAddMonoidHom = Additive.ofMul ∘ ψ := rfl
+@[simp] lemma toAddMonoidHom_apply (ψ : AddChar A M) (a : A) :
```
loefflerd, line 158:
```suggestion
@[simp] lemma coe_toAddMonoidHom (ψ : AddChar A M) : ⇑ψ.toAddMonoidHom = Additive.ofMul ∘ ψ := rfl

@[simp] lemma toAddMonoidHom_apply (ψ : AddChar A M) (a : A) :
```

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -157,36 +169,77 @@ def toAddMonoidHomEquiv : AddChar A M ≃ (A →+ Additive M) where
...
+
+lemma _root_.MonoidHom.compAddChar_injective_right (f : M →* N) (hf : Injective f) :
+    Injective fun ψ : AddChar B M ↦ f.compAddChar ψ := by
+  rintro ψ χ h; rw [DFunLike.ext'_iff] at h ⊢; exact hf.comp_left h
+
 /-- Composing an `AddChar` with an `AddMonoidHom` yields another `AddChar`. -/
-def compAddMonoidHom {B : Type*} [AddMonoid B] (φ : AddChar B M) (f : A →+ B) : AddChar A M :=
-  (toAddMonoidHomEquiv A M).symm (φ.toAddMonoidHom.comp f)
+def compAddMonoidHom (φ : AddChar B M) (f : A →+ B) : AddChar A M :=
+  (toAddMonoidHomEquiv).symm (φ.toAddMonoidHom.comp f)
```
loefflerd, line 216:
```suggestion
  toAddMonoidHomEquiv.symm (φ.toAddMonoidHom.comp f)
```

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -157,36 +169,77 @@ def toAddMonoidHomEquiv : AddChar A M ≃ (A →+ Additive M) where
...
+
+lemma compAddMonoidHom_injective_left (f : A →+ B) (hf : Surjective f) :
+    Injective fun ψ : AddChar B M ↦ ψ.compAddMonoidHom f := by
+  rintro ψ χ h; rw [DFunLike.ext'_iff] at h ⊢; exact hf.injective_comp_right h
 
-@[simp] lemma coe_compAddMonoidHom {B : Type*} [AddMonoid B] (φ : AddChar B M) (f : A →+ B) :
-    φ.compAddMonoidHom f = φ ∘ f := rfl
+lemma compAddMonoidHom_injective_right (ψ : AddChar B M) (hψ : Injective ψ) :
+    Injective fun f : A →+ B ↦ ψ.compAddMonoidHom f := by
+  rintro f g h; rw [DFunLike.ext'_iff] at h ⊢; exact hψ.comp_left h
```
loefflerd, line 230:
```suggestion
    Injective fun (f : A →+ B) ↦ ψ.compAddMonoidHom f := by
  rintro f g h
  rw [DFunLike.ext'_iff] at h ⊢
  exact hψ.comp_left h
```
(readability)

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -157,36 +169,77 @@ def toAddMonoidHomEquiv : AddChar A M ≃ (A →+ Additive M) where
...
+
+lemma eq_one_iff : ψ = 1 ↔ ∀ x, ψ x = 1 := DFunLike.ext_iff
+lemma ne_one_iff : ψ ≠ 1 ↔ ∃ x, ψ x ≠ 1 := DFunLike.ne_iff
```
loefflerd, line 233:
```suggestion
lemma eq_one_iff : ψ = 1 ↔ ∀ x, ψ x = 1 := DFunLike.ext_iff

lemma ne_one_iff : ψ ≠ 1 ↔ ∃ x, ψ x ≠ 1 := DFunLike.ne_iff
```

Mathlib/NumberTheory/GaussSum.lean
```diff
@@ -92,7 +92,7 @@ variable {R : Type u} [Field R] [Fintype R] {R' : Type v} [CommRing R'] [IsDomai
 
 -- A helper lemma for `gaussSum_mul_gaussSum_eq_card` below
 -- Is this useful enough in other contexts to be public?
-private theorem gaussSum_mul_aux {χ : MulChar R R'} (hχ : IsNontrivial χ) (ψ : AddChar R R')
+private theorem gaussSum_mul_aux {χ : MulChar R R'} (hχ : χ.IsNontrivial) (ψ : AddChar R R')
```
loefflerd, line 95: This is a bug, it seems. Without the fix, Lean successfully deduces that we must mean `MulChar.IsNontrivial`, because `AddChar.IsNontrivial` would not fit the arguments; but it still emits the deprecation warning for the declaration it _didn't_ use!

We should clearly get rid of `MulChar.IsNontrivial` too and then this will be moot; but the bug is worth reporting anyway, I think.

Mathlib/NumberTheory/LegendreSymbol/AddCharacter.lean
```diff
@@ -56,54 +56,43 @@ lemma val_mem_rootsOfUnity (φ : AddChar R R') (a : R) (h : 0 < ringChar R) :
...
-theorem IsNontrivial.isPrimitive {F : Type u} [Field F] {ψ : AddChar F R'} (hψ : IsNontrivial ψ) :
-    IsPrimitive ψ := by
-  intro a ha
-  cases' hψ with x h
-  use a⁻¹ * x
-  rwa [mulShift_apply, mul_inv_cancel_left₀ ha]
-#align add_char.is_nontrivial.is_primitive AddChar.IsNontrivial.isPrimitive
+theorem IsPrimitive.of_ne_one {F : Type u} [Field F] {ψ : AddChar F R'} (hψ : ψ ≠ 1) :
+    IsPrimitive ψ :=
+  fun a ha h ↦ hψ $ by simpa [mulShift_mulShift, ha] using congr_arg (mulShift · a⁻¹) h
```
loefflerd, line 86: I thought `$` was deprecated in favour of `<|` in mathlib

loefflerd (commented):
This is definitely an improvement, thanks!

Mathlib/NumberTheory/LegendreSymbol/AddCharacter.lean
grunweg: This is my reading of the style guide as well: "Note: while $ is a synonym for <|, its use in mathlib is disallowed in favor of <| for consistency as well as because of the symmetry with |>."

Mathlib/Algebra/Group/AddChar.lean
YaelDillies: I'm not convinced about this one. Putting the two next to each other shows that they are completely analogous. This is done in many place throughout mathlib

Mathlib/NumberTheory/GaussSum.lean
YaelDillies: It's already known. See https://github.com/leanprover/lean4/issues/2132 (it might seem like a slightly different issue, but it is actually the same, as confirmed on Zulip by Mario)

Mathlib/Algebra/Group/AddChar.lean
YaelDillies: I think that's okay. Furthermore since [I don't have cache](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/cache.20misses.20aesop) I won't adventure myself at removing `{ψ : AddChar A M}` to put it only where needed.

Mathlib/Algebra/Group/AddChar.lean
loefflerd: Fair enough

loefflerd:
Looks good now, thanks! 

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #11203, opened 2024-03-06
[Merged by Bors] - feat: `NNRat.cast`

YaelDillies:
Define the canonical coercion from the nonnegative rationals to any division semiring.

From LeanAPAP

---

- [x] depends on: #11202
- [x] depends on: #11453
- [x] depends on: #11480
- [x] depends on: #11485
- [x] depends on: #11486
- [x] depends on: #11499
- [x] depends on: #11503
- [x] depends on: #11504
- [x] depends on: #11507
- [x] depends on: #11508
- [x] depends on: #11639
- [x] depends on: #12360

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13585, opened 2024-06-07
[Merged by Bors] - feat(RingTheory.Binomial): Add BinomialRing instance for NNRat modules.

kcaze:
This PR adds an instance of BinomialRing for NNRat modules.

---

### DIFF
--- a/Mathlib/RingTheory/Binomial.lean
+++ b/Mathlib/RingTheory/Binomial.lean
@@ -6,6 +6,7 @@ Authors: Scott Carnahan
 import Mathlib.Algebra.Polynomial.Smeval
 import Mathlib.GroupTheory.GroupAction.Ring
 import Mathlib.RingTheory.Polynomial.Pochhammer
+import Mathlib.Data.Real.Basic
 
 /-!
 # Binomial rings
@@ -175,6 +176,28 @@ instance Int.instBinomialRing : BinomialRing ℤ where
         ← Int.neg_ofNat_succ, ascPochhammer_smeval_neg_eq_descPochhammer]
 
 end Nat_Int
+
+section Real
+/-- Inductive definition of the multichoose function. -/
+noncomputable def Real.multichoose: ℝ → ℕ → ℝ
+  | _, 0 => 1
+  | r, k+1 => (r+k)/(k+1)*(Real.multichoose r k)
+
+noncomputable instance Real.instBinomialRing : BinomialRing ℝ where
+  nsmul_right_injective n hn r s hrs := by
+    simp only [nsmul_eq_mul, mul_eq_mul_left_iff, Nat.cast_eq_zero, hn, or_false] at hrs
+    exact hrs
+  multichoose := Real.multichoose
+  factorial_nsmul_multichoose r n := by
+    induction' n with n ih
+    case zero =>
+      simp only [Nat.factorial_zero, multichoose, one_smul, ascPochhammer_zero,
+        Polynomial.smeval_one, pow_zero]
+    case succ =>
+      rw [Polynomial.ascPochhammer_smeval_eq_eval] at ih
+      rw [Polynomial.ascPochhammer_smeval_eq_eval, ascPochhammer_succ_eval, ← ih, multichoose, Nat.factorial]
+      field_simp; ring_nf
+end Real
 
 section Choose
 
--- a/Mathlib/RingTheory/Binomial.lean
+++ b/Mathlib/RingTheory/Binomial.lean
@@ -195,7 +195,8 @@ noncomputable instance Real.instBinomialRing : BinomialRing ℝ where
         Polynomial.smeval_one, pow_zero]
     case succ =>
       rw [Polynomial.ascPochhammer_smeval_eq_eval] at ih
-      rw [Polynomial.ascPochhammer_smeval_eq_eval, ascPochhammer_succ_eval, ← ih, multichoose, Nat.factorial]
+      rw [Polynomial.ascPochhammer_smeval_eq_eval, ascPochhammer_succ_eval, ← ih, multichoose,
+        Nat.factorial]
       field_simp; ring_nf
 end Real
 
--- a/Mathlib/RingTheory/Binomial.lean
+++ b/Mathlib/RingTheory/Binomial.lean
@@ -176,26 +176,14 @@ instance Int.instBinomialRing : BinomialRing ℤ where
 
 end Nat_Int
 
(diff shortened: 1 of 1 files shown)


ScottCarnahan:
Your proof doesn't seem to use any special features of the reals besides the fact that `k+1` is invertible for all natural numbers `k`.  This seems to suggest you can replace real numbers with something like `{R : Type*} [AddCommMonoid R] [Module ℚ≥0 R] [Pow R ℕ]`.

kcaze:
@ScottCarnahan Thanks for the suggestion to generalize. I wasn't able to quite get my proof to work with `[AddCommMonoid R] [Module ℚ≥0 R] [Pow R ℕ]`. My latest commit requires `R` to be `[CommRing R] [CharZero R] [IsLeftCancelMulZero R] [Algebra ℚ R]` instead.

Mathlib/RingTheory/Binomial.lean
```diff
@@ -176,6 +176,35 @@ instance Int.instBinomialRing : BinomialRing ℤ where
...
 
+section Rat_Algebra
+
+variable {R : Type*} [CommRing R] [CharZero R] [IsLeftCancelMulZero R] [Algebra ℚ R]
+
+/-- Inductive definition of the multichoose function. -/
+noncomputable def multichoose: R → ℕ → R
+  | _, 0 => 1
+  | r, k+1 => ((k+1)⁻¹:ℚ)•(r+k)*(multichoose r k)
+
```
ScottCarnahan, line 187:
```suggestion
```

Mathlib/RingTheory/Binomial.lean
```diff
@@ -176,6 +176,35 @@ instance Int.instBinomialRing : BinomialRing ℤ where
...
+    case succ =>
+      rw [Polynomial.ascPochhammer_smeval_eq_eval] at ih
+      rw [Polynomial.ascPochhammer_smeval_eq_eval, ascPochhammer_succ_eval, ← ih, multichoose,
+        Nat.factorial]
+      nth_rewrite 1 [← (one_smul ℚ (multichoose r n))]
+      rw [smul_mul_smul, ← smul_assoc, ← smul_mul_smul]
+      field_simp
+      rw [Algebra.smul_def', map_natCast]
+      ring_nf
+end Rat_Algebra
```
ScottCarnahan, line 207:
```suggestion
noncomputable instance {R : Type*} [AddCommMonoid R] [Module ℚ≥0 R] [Pow R ℕ] : BinomialRing R where
  nsmul_right_injective n hn r s hrs := by
    rw [← one_smul ℚ≥0 r, ← one_smul ℚ≥0 s, show 1 = (n : ℚ≥0)⁻¹ • (n : ℚ≥0) by simp_all]
    simp_all only [smul_assoc, ← nsmul_eq_smul_cast]
  multichoose r n := (n.factorial : ℚ≥0)⁻¹ • Polynomial.smeval (ascPochhammer ℕ n) r
  factorial_nsmul_multichoose r n := by
    simp only [← smul_assoc]
    field_simp
```

Mathlib/RingTheory/Binomial.lean
ScottCarnahan: Because factorials are invertible in our setting, we can use a definition that doesn't split into cases.

Mathlib/RingTheory/Binomial.lean
ScottCarnahan: I don't think anyone will really need this additional generality, but here's a try.

kcaze:
Thanks, this looks much simpler. I think that instead of a $\mathbb{Q}_{\ge 0}$ module we should however specify a $\mathbb{Q}$ module. With `[Module ℚ≥0 R]` we cannot synthesize a BinomialRing instance for the reals. For example `def test := BinomialRing.multichoose (4:ℝ) 2` yields the error
```
failed to synthesize
  BinomialRing ℝ
```
Changing all instances of `ℚ≥0` to `ℚ` fixes this.

Arguably, the real issue is that we lack a equivalent to [NNReal.instModuleOfReal](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Real/NNReal.html#NNReal.instModuleOfReal) but for `NNRat` and `Rat`. What do you think?

ScottCarnahan:
This is curious.  I see the instance was removed from Mathlib.Algebra.Module.Basic in #11203.

ScottCarnahan:
By the way, if you don't want boldface text in the PR description, put a blank line between your text and the line with three dashes.

kcaze:
Great, thank you. I've committed your suggested version to the PR and am happy with waiting for Yaël's PR to fix synthesizing `Module ℚ≥0 ℝ` to come after this one.

ScottCarnahan (approved):
It looks good to me.  If you want, you can change the name of the `Nat_Int` section to something more general like `Basic_Instances`, and put the instance in there.

kcaze:
Great, I've changed the section to `Basic_Instances` so I think this is ready to be merged now.

kcaze:
@ScottCarnahan, sorry for the ping, but does this need a `maintainer merge` comment to move into the merging queue? Or do I just need to wait for a maintainer to take a look?

ScottCarnahan:
Sorry I don't have permission to advance it here, but I can start a discussion on Zulip.  On an unrelated note, perhaps you should edit the PR description, e.g., change "real" to "NNRat" or "rational"

Mathlib/RingTheory/Binomial.lean
```diff
@@ -174,7 +174,15 @@ instance Int.instBinomialRing : BinomialRing ℤ where
...
-end Nat_Int
+noncomputable instance {R : Type*} [AddCommMonoid R] [Module ℚ≥0 R] [Pow R ℕ] : BinomialRing R where
+  nsmul_right_injective n hn r s hrs := by
+    rw [← one_smul ℚ≥0 r, ← one_smul ℚ≥0 s, show 1 = (n : ℚ≥0)⁻¹ • (n : ℚ≥0) by simp_all]
+    simp_all only [smul_assoc, ← nsmul_eq_smul_cast]
+  multichoose r n := (n.factorial : ℚ≥0)⁻¹ • Polynomial.smeval (ascPochhammer ℕ n) r
+  factorial_nsmul_multichoose r n := by
+    simp only [← smul_assoc]
+    field_simp
+end Basic_Instances
```
jcommelin, line 185:
```suggestion

end Basic_Instances
```

jcommelin (commented):
Thanks!

bors d+

PR #13587, opened 2024-06-07
[Merged by Bors] - chore: deprecate unused Array/List function

kim-em:
---

### DIFF
--- a/Mathlib/Data/Array/Basic.lean
+++ b/Mathlib/Data/Array/Basic.lean
@@ -5,8 +5,11 @@ Authors: Mario Carneiro
 -/
 import Batteries.Tactic.Alias
 
+-- TODO 2024-06-07: This has been upstreamed in https://github.com/leanprover/lean4/pull/4389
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Array/Basic.lean
```diff
@@ -5,8 +5,11 @@ Authors: Mario Carneiro
 -/
 import Batteries.Tactic.Alias
 
+-- TODO 2024-06-07: This has been upstreamed in https://github.com/leanprover/lean4/pull/4389
```
grunweg, line 8: That PR has landed - can you reword this comment accordingly? Right now, it's not clear to me what this means.

grunweg (approved):
Looks good to me, apart from one minor comment. Thanks!

Mathlib/Data/Array/Basic.lean
urkud: IMHO, it's easy enough to follow the link.
bors merge

PR #13589, opened 2024-06-07
[Merged by Bors] - chore(Data/Bool/Basic): deprecate duplicate definitions

kim-em:
Deprecates theorems in favour of the versions in the Lean repository.

This is actually pretty close to making `Data.Bool.Basic` a leaf file. We'd just need to move the LinearOrder instance elsewhere, and `injective_iff`.

### DIFF
--- a/Mathlib/Data/Bool/Basic.lean
+++ b/Mathlib/Data/Bool/Basic.lean
@@ -21,44 +21,36 @@ bool, boolean, Bool, De Morgan
 
 namespace Bool
 
-theorem decide_True {h} : @decide True h = true :=
-  _root_.decide_eq_true True.intro
-#align bool.to_bool_true Bool.decide_True
+@[deprecated (since := "2024-06-07")] alias decide_True := decide_true_eq_true
+#align bool.to_bool_true decide_true_eq_true
 
-theorem decide_False {h} : @decide False h = false :=
-  _root_.decide_eq_false id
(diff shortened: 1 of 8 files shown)


grunweg:
(This passes CI now. Please remember to this as `awaiting-review` if you'd like this to be reviewed.)

Mathlib/Data/Bool/Basic.lean
```diff
@@ -150,8 +135,8 @@ theorem ne_not {a b : Bool} : a ≠ !b ↔ a = b :=
   not_eq_not
 #align bool.ne_bnot Bool.ne_not
 
-theorem not_ne : ∀ {a b : Bool}, (!a) ≠ b ↔ a = b := not_not_eq
-#align bool.bnot_ne Bool.not_ne
+@[deprecated (since := "2024-06-07")] alias not_ne := not_not_eq
```
urkud, line 138: Note that these lemmas aren't syntactically equal: one uses `Not (_ = _)`, another uses `Ne _ _`. Not sure if it justifies leaving the `ne` lemma.

semorrison: Given that it isn't used, I'd prefer to err on the side of more aggressively cleaning up. If someone downstream is upset by the deprecation they still have a chance to say something about it.

dupuisf:
Looks good!

bors r+

PR #13555, opened 2024-06-06
[Merged by Bors] - refactor(Mathlib/Algebra/ContinuedFractions/*): generalize determinant formula for continued fraction computation to simple continued fraction

Komyyy:
Determinant formula for continued fractions in Mathlib is specialized for continued fraction computation:
```lean
theorem GenContFract.determinant (not_terminated_at_n : ¬(of v).TerminatedAt n) :
    (of v).numerators n * (of v).denominators (n + 1) -
      (of v).denominators n * (of v).numerators (n + 1) = (-1) ^ (n + 1) :=
  determinant_aux <| Or.inr <| not_terminated_at_n
```
This formula can be generalized for arbitrary simple continued fractions, so we generalize:
```lean
theorem SimpContFract.determinant (not_terminatedAt_n : ¬(↑s : GenContFract K).TerminatedAt n) :
    (↑s : GenContFract K).nums n * (↑s : GenContFract K).dens (n + 1) -
      (↑s : GenContFract K).dens n * (↑s : GenContFract K).nums (n + 1) = (-1) ^ (n + 1)
```

Also, this PR renames `Real.exists_genContFract_convs_eq_rat` to `Real.exists_convs_eq_rat` which is forgotten to rename in #13210.

---

- [x] depends on: #13210

This PR consists [Continued fractions makes an isomorphism between irrationals and baire space
](https://github.com/users/Komyyy/projects/3/) project

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13590, opened 2024-06-07
refactor(Mathlib/Algebra/ContinuedFractions/*): change the definition of regular continuous fractions

Komyyy:
In mathlib, a regular continued fraction is defined as a simple continued fraction whose partial denominators are all positive, but no matter which literature I searched, a regular continued fraction is defined as a simple continued fraction whose head term is integer and whose partial denominators are all positive naturals. Indeed, if we redefine a regular continued fraction as this way, we can get correspondence between reals and regular continued fractions which does't finish with 1, so I redefine regular continued fractions.

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Definition.20of.20regular.20continued.20fractions)

---

- [x] depends on: #13555

This PR consists [Continued fractions makes an isomorphism between irrationals and baire space
](https://github.com/users/Komyyy/projects/3/) project


Komyyy:
❗This project is temporarily frozen, PRs are all closed for a moment. [Detail](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Definition.20of.20regular.20continued.20fractions)

PR #13596, opened 2024-06-07
[Merged by Bors] - chore: avoid instance names for DMatrix

Ruben-VandeVelde:
This should be more robust.

---


urkud:
Thanks! :tada:
bors merge

PR #13038, opened 2024-05-19
feat: Mathlib weekly reports

adomani:
This PR introduces a weekly cron job that computes a "global" report on the evolution of Mathlib in the previous week.

It consists of
* a CI workflow with a cron job that runs at minight on Sunday (`.github/workflows/mathlib_stats.yaml`);
* a Lean file extracting a categorized list of "all" the declarations in Mathlib (`scripts/count_decls.lean`);
* a bash file computing Git-diff-related information and collating the data from the Lean file (`scripts/mathlib_stats.sh`);
* a convenience CI workflow that is triggered on adding the `test-ci` label and results in posting on the PR and on Zulip the report (`.github/workflows/mathlib_stats_label.yaml`).

The second CI workflow is intended to be removed just before/right after the PR is ready to merge.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general)

[Thread for the reports](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Mathlib.20weekly.20change.20report)

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13597, opened 2024-06-07
[Merged by Bors] - ci: post a weekly report to zulip with some technical debt stats

jcommelin:
---

### DIFF
--- a/scripts/technical-debt-counters.sh
+++ b/scripts/technical-debt-counters.sh
@@ -25,6 +25,13 @@ for i in ${!titlesAndRegexes[@]}; do
   fi
 done
 
+printf '|%s | %s |\n' "$(cat scripts/nolints.json | grep -c 'docBlame')" "documentation nolint entries"
+printf '|%s | %s |\n' "$(cat scripts/style-exceptions.txt | grep -c 'ERR_MOD')" "missing module docstrings"
+printf '|%s | %s |\n' "$(cat scripts/style-exceptions.txt | grep -c 'ERR_NUM_LIN')" "large files"
+# We print the number of files, not the number of matches --- hence, the nested grep.
+printf '|%s | %s |\n' "$(git grep -c 'autoImplicit true' | grep -c -v 'test')" "non-test files with autoImplicit true"
+
+
 initFiles="$(git ls-files '**/Init/*.lean' | xargs wc -l)"
 
 printf '|%s | %s |\n\n' "$(printf '%s' "${initFiles}" | wc -l)" "\`Init\` files"
--- a/scripts/technical-debt-counters.sh
+++ b/scripts/technical-debt-counters.sh
@@ -2,40 +2,27 @@
 
 # This script quantifies some of the technical debt in mathlib4.
 
-echo "Number of porting notes:"
-git grep -i "Porting note" | wc -l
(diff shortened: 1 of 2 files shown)


grunweg:
Great idea! I'd like to suggest three further metrics, at least for the near future.
- number of documentation `nolints`s: almost all of them are genuine at this point
Computing them can be as simple as `cat scripts/nolints.json | grep -c "docBlame"`.
- number of large files (as flagged by the style linter): long-term, this metric may cease to be useful, as a few transient exceptions are presumably normal. In my opinion, we're not at this point yet.) Can be computed using `cat scripts/style-exceptions.txt | grep -c "ERR_NUM_LIN"` (currently 97); using `git grep` would also work.
- number of missing module docstrings: `cat scripts/style-exceptions.txt | grep -c "ERR_MOD"` (currently 31)
- number of files in `Mathlib/` with `autoImplicit true` : `git grep -c "autoImplicit true" | grep -c "Mathlib"`
(For all files, `git grep -c "autoImplicit true" | wc -l` will also work.)
Again, there may be a number of files where `autoImplicit true` is used sensibly throughout the entire file - but in my experience removing this from files, most files do not use this intentionally, as it's just another small piece of technical debt.)

adomani:
There is also #13038 that has more statistics.

jcommelin:
@grunweg would you like to add those to this PR? Or as a follow-up?

semorrison:
Let's iterate fast. We'll merge this and add more later.

I'd love to see graphs, too. :-)

semorrison:
bors merge

grunweg:
I just pushed a commit adding the four metrics I proposed, not having seen the approval. Apologies if this cancelled the bors push.

adomani:
Michael, I think that bors is struggling to merge a group of PRs with some conflicts among them, so even without your push this would have probably failed anyway.

scripts/technical-debt-counters.sh
```diff
@@ -0,0 +1,41 @@
...
+    then fl="-i"  # just for porting notes we ignore the case in the regex
+    else fl="--"
+    fi
+    printf '| %s | %s |\n' "$(git grep "${fl}" "${regex}" | wc -l)" "${title}"
+  fi
+done
+
+printf '|%s | %s |\n' "$(cat scripts/nolints.json | grep -c 'docBlame')" "documentation nolint entries"
+printf '|%s | %s |\n' "$(cat scripts/style-exceptions.txt | grep -c 'ERR_MOD')" "missing module docstrings"
+printf '|%s | %s |\n' "$(cat scripts/style-exceptions.txt | grep -c 'ERR_NUM_LIN')" "large files"
```
adomani, line 30:
```suggestion
printf '|%s | %s |\n' "$(grep -c 'docBlame' scripts/nolints.json)" "documentation nolint entries"
printf '|%s | %s |\n' "$(grep -c 'ERR_MOD' scripts/style-exceptions.txt)" "missing module docstrings"
printf '|%s | %s |\n' "$(grep -c 'ERR_NUM_LIN' scripts/style-exceptions.txt)" "large files"
```

grunweg: That's even nicer, thanks!

j-loreaux:
bors merge

PR #13599, opened 2024-06-07
feat: definition of Ext-groups as morphisms in the derived category

joelriou:
This is not a refactor PR (yet). In this PR, given objects `X` and `Y` in an abelian category `C`, we define a type `LargeExt X Y n` as morphisms in the derived category of `C`. In order to replace the current `Ext` API by this, we shall need more API which allow to shrink these `LargeExt` types in a smaller universes.

---

I may have found a better design: similarly as #13639, I may define a shrunk version of `ShiftedHom` similarly, and define `SmallExt` accordingly. It would make the definition of `LargeExt` unnecessary, but at the same time, we would also have bijections between `SmallExt` with morphisms in the derived category.

This PR is replaced by #13963.

- [x] depends on: #13600
- [x] depends on: #13571
- [x] depends on: #11806


Mathlib/Algebra/Homology/DerivedCategory/LargeExt.lean
```diff
@@ -0,0 +1,166 @@
...
+    dsimp
+    simp only [ofHom_comp]
+    symm
+    apply comp_assoc
+    all_goals omega
+
+variable (C)
+
+/-- The `Ext`-groups in degree `n : ℕ`, as a bifunctor `Cᵒᵖ ⥤ C ⥤ Ab.{w}`. -/
+noncomputable def LargeExtFunctor (n : ℕ) : Cᵒᵖ ⥤ C ⥤ Ab.{w} where
```
erdOne, line 143: I think `LargeExt` being a (semi-reducible) alias will let us reuse the existing API easier.
For example this is just `(singleFunctor C 0).op ⋙ preadditiveCoyoneda ⋙ (whiskeringLeft _ _ _).obj (singleFunctor C n)`.

joelriou: Actually, I have included this definition of `LargeExtFunctor` mostly to check that the API works well enough. The fact that there is a bifunctor is essentially irrelevant for me, because it involves only the composition of `LargeExt` when one of the elements is in degree `0`. The composition of elements in `LargeExt` in arbitrary degrees is a much more important structure.

(Anyway, `LargeExt` is just an implementation detail: the real API will be the future `SmallExt` which shall shrink these groups to a possibly smaller universe.)

PR #13600, opened 2024-06-07
[Merged by Bors] - feat: the single functors from the homotopy category

joelriou:
In this PR, we define the single functors from a preadditive category (with a zero object) to its homotopy category of cochain complexes and show they behave well with respect to shift functors.

---


erdOne:
Judging by the comments in the code, I have marked this `WIP`.

joelriou:
Now this is ready for review.

erdOne (commented):
I think `simps` can generate some useful lemmas?

joelriou:
> I think `simps` can generate some useful lemmas?

Actually, in most (future) situations, we do not want these definitions to be unfolded. These are almost like blackboxes which give an additional structure on these categories.

erdOne:
Gotcha. Can you record this in the docstring?

joelriou:
I have just added a comment.

Mathlib/Algebra/Homology/HomotopyCategory/SingleFunctors.lean
```diff
@@ -0,0 +1,94 @@
...
+
+variable (C : Type u) [Category.{v} C] [Preadditive C] [HasZeroObject C]
+
+namespace CochainComplex
+
+open HomologicalComplex
+
+/-- The collection of all single functors `C ⥤ CochainComplex C ℤ` along with
+their compatibilites with shifts. (This definition has purposely no `simps`
+attribute, as the generated lemmas would be very useful.) -/
```
erdOne, line 33: I don't think this is what you meant?
Or maybe even `irreducible_def` if you are 100% sure.

joelriou: I have fixed the phrasing. (`irreducible_def` would be too much.)

Mathlib/Algebra/Homology/HomotopyCategory/SingleFunctors.lean
```diff
@@ -0,0 +1,94 @@
...
+    ext
+    simp [shiftFunctorAdd_eq, XIsoOfEq]
+
+instance (n : ℤ) : ((singleFunctors C).functor n).Additive := by
+  dsimp only [singleFunctors]
+  infer_instance
+
+/-- The single functor `C ⥤ CochainComplex C ℤ` which sends `X` to the complex
+consisting of `X` in degree `n : ℤ` and zero otherwise. -/
+noncomputable abbrev singleFunctor (n : ℤ) := (singleFunctors C).functor n
```
erdOne, line 62: So this is `defeq` to `single _ _ n` but this is the main API and not `single` right?
I think it is useful to document these things. (This is what I meant above)

joelriou: Thanks for the suggestion.

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13601, opened 2024-06-07
Lean pr testing 4387

JovanGerb:
Made a fix to unification that removes some silly eta-unreduced terms from unification. Some proofs break because they used to need to some eta or beta reduction.

---


JovanGerb:
!bench

JovanGerb:
!bench

JovanGerb:
!bench

PR #13604, opened 2024-06-07
[Merged by Bors] - chore(Order/Basic): cleanup redundant type arguments

madvorak:
---

### DIFF
--- a/Mathlib/Order/Basic.lean
+++ b/Mathlib/Order/Basic.lean
@@ -65,7 +65,7 @@ open Function
 
 universe u v w
 
-variable {ι : Type*} {α : Type u} {β : Type v} {π : ι → Type*}
+variable {ι α β : Type*} {π : ι → Type*}
 
 section Preorder
 
--- a/Mathlib/Order/Basic.lean
+++ b/Mathlib/Order/Basic.lean
@@ -54,7 +54,6 @@ provide many aliases to dot notation-less lemmas. For example, `le_trans` is ali
 
 - expand module docs
 - automatic construction of dual definitions / theorems
-- remove duplicate declaration of type arguments from respective declarations
 
 ## Tags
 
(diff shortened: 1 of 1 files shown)


Mathlib/Order/Basic.lean
```diff
@@ -1503,7 +1502,7 @@ instance Prop.partialOrder : PartialOrder Prop where
 
 end «Prop»
 
-variable {s : β → β → Prop} {t : γ → γ → Prop}
+variable {γ : Type w} {s : β → β → Prop} {t : γ → γ → Prop}
```
YaelDillies, line 1505: Why not keep `γ` at the top?

Mathlib/Order/Basic.lean
```diff
@@ -66,7 +65,7 @@ open Function
 
 universe u v w
 
-variable {ι : Type*} {α : Type u} {β : Type v} {γ : Type w} {π : ι → Type*} {r : α → α → Prop}
+variable {ι : Type*} {α : Type u} {β : Type v} {π : ι → Type*}
```
YaelDillies, line 68: While you're at it
```suggestion
variable {ι α β : Type*} {π : ι → Type*}
```

Mathlib/Order/Basic.lean
madvorak: It is used here for the first time. Why to clutter the infoview for the intermediate 1500 lines?

Mathlib/Order/Basic.lean
madvorak: Do you want to get rid of `universe u v w` altogether?

Mathlib/Order/Basic.lean
YaelDillies: Type variables are usually put at the top

madvorak: OK

madvorak: Actually, the entire line 1506 was never used, lol.

Mathlib/Order/Basic.lean
madvorak: I deleted them.

Mathlib/Order/Basic.lean
YaelDillies: Welp, that solves it

YaelDillies:
maintainer merge

PR #13606, opened 2024-06-07
[Merged by Bors] - chore: desimp `compl_sdiff`

apnelson1:
We remove the `simp` attribute from `theorem compl_sdiff : (x \ y)ᶜ = x ⇨ y`, since (for sets) it turns standard notation into exotic notation. 

Given that this passed CI with no further changes, it seems likely it wasn't being used much. 

---


urkud:
Thanks! :tada:
bors merge

PR #13602, opened 2024-06-07
[Merged by Bors] - CI: extract update comment and merge import and diff summaries

adomani:
This PR extends the PR summary introduced in #13556.

* Extracts the GitHub comment creation/update to a new file;
* adds the output of `move-decls` to the automatically updated message posted by the import-automation;
* if the output of either summary exceeds a few lines, then that output is hidden behind a collapsed summary.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13608, opened 2024-06-07
[Merged by Bors] - expected fix: checkout from master for PR summary

adomani:
This PR accompanies #13602: it makes the script pull code from master, now that it is present there, rather than from the branch that introduced the code.

It is expected to fail CI until the PR on which it depends is merged.

It also fixes a copy-paste error in the labeling of `Diff of declarations`.

---

- [ ] depends on: #13602, where the code is introduced


.github/workflows/PR_summary.yml
```diff
@@ -41,10 +41,8 @@ jobs:
 
         # Checkout the merge base
         git checkout "$(git merge-base master ${{ github.sha }})"
-        # git checkout origin/master scripts/count-trans-deps.py scripts/import-graph-report.py
-        # get the files that we need from the current PR
-        # TODO ***change*** to `origin/master` in the line below, when merging or right after
-        git checkout origin/adomani/move_decls_autocomment scripts/update_PR_comment.sh
+        # get the files that we need from master
+        git checkout origin/master scripts/update_PR_comment.sh scripts/count-trans-deps.py scripts/import-graph-report.py
```
adomani, line 45:
```suggestion
```
Can these lines be removed entirely?  If a PR triggers the bot, it should have merged a version of master that contains the scripts anyway.  Will Git use an earlier master for its comparisons?

jcommelin: I think you are right.

jcommelin:
bors d+

PR #12109, opened 2024-04-13
[Merged by Bors] - feat(RingTheory/Coalgebra/*): add basics about sweedler notation

jjaassoonn:
Though the name is called sweedler notation, no new notations are actually introduced. This PR is actually about expanding tensor product.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #6524, opened 2023-08-11
[Merged by Bors] - fix(Algebra/MonoidAlgebra/Basic): resolve `ext` porting notes

eric-wieser:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13609, opened 2024-06-07
[Merged by Bors] - feat: Hopf algebra structure on monoid algebras 

101damnations:
Given a commutative semiring `R`, an `R`-bialgebra (resp. `R`-Hopf algebra) `A`, and a(n) (additive) monoid (resp. (additive) group) `G`, this PR defines the `R`-bialgebra (resp. `R`-Hopf algebra) structure on `(Add)MonoidAlgebra A G`. 

---

I've added a little API for monoid algebras that duplicates API for `Finsupp`, because I needed to go back and forth between using monoid algebra facts and `Finsupp` facts in proofs, and this seemed to be the cleanest way to do that. 

- [x] depends on: #12109 


adomani:
In case you are wondering, CI fails due to an (expected) issue that is fixed in #13608.  I will merge that PR as soon as it is green.

The underlying reason is that the failing step inspects the PR and a version of master.  When writing the PR, the version of master used did not have the new scripts, so they had to be pulled in.  Now, the branch from which they were pulled is no longer accessible, but master has the files.

101damnations:
I see, thank you very much!

adomani:
The PR causing the issue has been merged: if you merge master again, this will be a good test for whether the issue is resolved!  :smile:

Mathlib/RingTheory/Bialgebra/Basic.lean
```diff
@@ -142,6 +145,92 @@ variable {R A}
...
+
+namespace AddMonoidAlgebra
+
+open Bialgebra
+
+variable {R : Type u} [CommSemiring R] {A : Type v} [Semiring A] [Bialgebra R A]
+variable {X : Type w}
+
+variable (R A X) in
+instance instCoalgebra : Coalgebra R A[X] := Finsupp.instCoalgebra R X A
```
kbuzzard, line 199: Again presumably you don't need A to be a bialgebra until a bit later.

Mathlib/RingTheory/Coalgebra/Basic.lean
```diff
@@ -222,7 +222,7 @@ namespace Finsupp
 variable (R : Type u) (ι : Type v) (A : Type w)
 variable [CommSemiring R] [AddCommMonoid A] [Module R A] [Coalgebra R A]
 
-open LinearMap
+open LinearMap CoalgebraStruct
```
kbuzzard, line 225: Presumably this is unnecessary?

Mathlib/RingTheory/Coalgebra/Hom.lean
```diff
@@ -23,7 +23,7 @@ homomorphism.
 
 -/
 
-open TensorProduct Coalgebra
+open TensorProduct Coalgebra CoalgebraStruct
```
kbuzzard, line 26: (and this)

kbuzzard (approved):
This looks great to me, it looks like very uncontroversial additions. Maybe you could mention somewhere what the point of these results is? My understanding is that you're making G_a and G_m into R-group-schemes (and possibly other things too?) but you never mention this.

101damnations:
Thank you! As well as mentioning G_m I've added the actual Hopf algebra instance on Laurent polynomials + some API. (And also, my bad, I forgot G_a works a bit differently, so this PR doesn't cover that. But I've nearly finished the corresponding Hopf algebra instance on polynomials, so I'll open a PR for that next.)

kbuzzard (approved):
Thanks!

maintainer merge

Mathlib/Algebra/MonoidAlgebra/Basic.lean
```diff
@@ -922,6 +922,31 @@ theorem domCongr_toAlgHom (e : G ≃* H) : (domCongr k A e).toAlgHom = mapDomain
...
+variable {A : Type*} [Semiring k] [Semiring A] [Module k A]
+
+variable (k A) in
+/-- Interpret `single i` as a linear map (but with `MonoidAlgebra` in the type, rather than
+`Finsupp`). -/
+def lsingle (i : G) : A →ₗ[k] MonoidAlgebra A G :=
+  Finsupp.lsingle i
+
+lemma lsingle_def (i : G) :
+    lsingle k A i = Finsupp.lsingle i := rfl
```
eric-wieser, line 936: This lemma breaks the MonoidAlgebra / Finsupp abstraction, so probably shouldn't exist

Mathlib/Algebra/MonoidAlgebra/Basic.lean
```diff
@@ -922,6 +922,31 @@ theorem domCongr_toAlgHom (e : G ≃* H) : (domCongr k A e).toAlgHom = mapDomain
...
+
+@[simp]
+lemma lsingle_apply (i : G) (a : A) :
+    lsingle k A i a = single i a := rfl
+
+@[ext]
+theorem lhom_ext {B : Type*} [AddCommMonoid B] [Module k B]
```
eric-wieser, line 943: The `@[ext]` lemma should be the one about composition with `lsingle`

Mathlib/RingTheory/Bialgebra/Basic.lean
```diff
@@ -156,3 +158,96 @@ instance toBialgebra : Bialgebra R R where
...
+    Coalgebra.comul (single x a)
+      = (TensorProduct.map (lsingle R A x) (lsingle R A x) : _ →ₗ[R] _) (Coalgebra.comul a) :=
+  Finsupp.comul_single _ _ _ _ _
+
+end
+variable [Bialgebra R A] [Monoid X]
+
+variable (R A X) in
+instance instBialgebra : Bialgebra R (MonoidAlgebra A X) :=
+  { instCoalgebra R A X with
```
eric-wieser, line 190: nit
```suggestion
instance instBialgebra : Bialgebra R (MonoidAlgebra A X) where
  __ := instCoalgebra R A X
```
and reduce the indents below

Mathlib/RingTheory/HopfAlgebra.lean
```diff
@@ -115,3 +125,152 @@ instance toHopfAlgebra : HopfAlgebra R R where
...
+
+variable {R : Type u} {A : Type v} [CommSemiring R] [Semiring A] [HopfAlgebra R A]
+variable {G : Type w} [Group G]
+
+variable (R A G) in
+/-- The antipode in the `R`-Hopf algebra structure on `MonoidAlgebra A G`, sending
+`single g a` to `single g⁻¹ (antipode a)`. -/
+def antipode :
+    MonoidAlgebra A G →ₗ[R] MonoidAlgebra A G :=
+  Finsupp.lsum R fun g => Finsupp.lsingle g⁻¹ ∘ₗ HopfAlgebra.antipode
```
eric-wieser, line 138: This definition is arguably a bad idea, because now you have separate `HopfAlgebra.antipode` and `MonoidAlgebra.algebra` definitions, and lemmas have to pick which spelling to use.

Some options here:
* Make it clear in the docstring that `HopfAlgebra.antipode` is the preferred spelling
* Add a `simp` lemma turning this spelling into `HopfAlgebra.antipode`
* Pull out a `HopfAlgebraStruct` base class from `HopfAlgebra` that contains just the antipode (and perhaps bialgebra structure), so that you can create this data first before proving `antipode_single`.

Mathlib/RingTheory/HopfAlgebra.lean
```diff
@@ -115,3 +125,152 @@ instance toHopfAlgebra : HopfAlgebra R R where
...
+      simp only [← lsingle_apply (k := R), ← map_sum (lsingle R A (0 : G)),
+        HopfAlgebra.sum_mul_antipode_eq_smul, Algebra.algebraMap_eq_smul_one] }
+
+@[simp]
+theorem antipode_single (g : G) (a : A) :
+    HopfAlgebra.antipode (R := R) (single g a) = single (-g) (HopfAlgebra.antipode (R := R) a) :=
+  antipode_single' _ _
+
+end AddMonoidAlgebra
+namespace LaurentPolynomial
```
eric-wieser, line 209: nit:
```suggestion
end AddMonoidAlgebra

namespace LaurentPolynomial
```

Mathlib/Algebra/MonoidAlgebra/Basic.lean
urkud: See https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Polynomial/Basic.html#Polynomial.lhom_ext%27 for an example.

Mathlib/Algebra/MonoidAlgebra/Basic.lean
101damnations: I needed to use linearity of `single`, and then, afterwards, properties of the multiplication on `A[G]` in e.g. `MonoidAlgebra.instBialgebra`. And if I use `Finsupp.lsingle, Finsupp.lhom_ext` then the monoid algebra gets unfolded to `Finsupp` too early; adding `MonoidAlgebra.lsingle` was the best way I could think of to get around this

Mathlib/RingTheory/HopfAlgebra.lean
101damnations: Thanks for the review! I like the third option so I've gone with that.

Mathlib/RingTheory/HopfAlgebra.lean
```diff
@@ -115,3 +133,146 @@ instance toHopfAlgebra : HopfAlgebra R R where
...
+  antipode := Finsupp.lsum R fun g => Finsupp.lsingle g⁻¹ ∘ₗ antipode
+
+@[simp]
+lemma antipode_single (g : G) (a : A) :
+    antipode (R := R) (single g a) = single g⁻¹ (antipode (R := R) a) := by
+  simp only [MonoidAlgebra, antipode, Finsupp.coe_lsum, map_zero, Finsupp.sum_single_index]
+  rfl
+
+instance instHopfAlgebra : HopfAlgebra R (MonoidAlgebra A G) where
+  toHopfAlgebraStruct := instHopfAlgebraStruct R A G
```
eric-wieser, line 155: I think you can drop this line?

Mathlib/RingTheory/HopfAlgebra.lean
```diff
@@ -115,3 +133,146 @@ instance toHopfAlgebra : HopfAlgebra R R where
...
+
+@[simp]
+theorem counit_T (n : ℤ) :
+    Coalgebra.counit (R := R) (T (R := A) n) = 1 := by
+  erw [counit_single]
+  rw [Bialgebra.counit_one]
+
+theorem counit_C_mul_T (a : A) (n : ℤ) :
+    Coalgebra.counit (R := R) (C a * T n) = Coalgebra.counit (R := R) a := by
+  simp
```
eric-wieser, line 258: Don't these theorems belong in the file about Coalgebras? I think they only need `Coalgebra R A`).

101damnations: Thanks, done - I've moved the API for the coalgebra structure on monoid algebras to the coalgebra file too.

Mathlib/Algebra/MonoidAlgebra/Defs.lean
```diff
@@ -574,6 +574,31 @@ theorem liftNC_smul [MulOneClass G] {R : Type*} [Semiring R] (f : k →+* R) (g
...
+
+variable (k A) in
+/-- Interpret `single i` as a linear map (but with `MonoidAlgebra` in the type, rather than
+`Finsupp`). -/
+def lsingle (i : G) : A →ₗ[k] MonoidAlgebra A G :=
+  Finsupp.lsingle i
+
+lemma lsingle_def (i : G) :
+    lsingle k A i = Finsupp.lsingle i := rfl
+
```
eric-wieser, line 589: Just to re-iterate; this lemma is bad, as it unfolds across the API boundary. Given it seems you don't use it anyway, can you just delete it? (same for the `AddMonoidAlgebra` version).

Mathlib/Algebra/MonoidAlgebra/Defs.lean
```diff
@@ -574,6 +574,31 @@ theorem liftNC_smul [MulOneClass G] {R : Type*} [Semiring R] (f : k →+* R) (g
...
+@[simp]
+lemma lsingle_apply (i : G) (a : A) :
+    lsingle k A i a = single i a := rfl
+
+@[ext]
+theorem lhom_ext {B : Type*} [AddCommMonoid B] [Module k B]
+    {f g : MonoidAlgebra A G →ₗ[k] B}
+    (h : ∀ a b, f (lsingle k A a b) = g (lsingle k A a b)) : f = g :=
```
eric-wieser, line 597: This should be stated with function composition, per "partially-applied ext lemmas". I  believe the proof is then `Finsupp.lhom_ext'`

Mathlib/Algebra/MonoidAlgebra/Defs.lean
eric-wieser: To be super clear, the new `lsingle` definition is a good idea 🙂.

eric-wieser:
I think the `lsingle` stuff might overlap with #6524, which I wouldn't previously have mentioned, but @YaelDillies seems to have revived while I wasn't looking. I'd prefer to get that one in first, if that's ok.

101damnations:
Thanks for pointing that out! I've removed my `lsingle` sections.

adamtopaz (approved):
I think this PR looks good to go! I had just two very minor comments, that should be considered optional. Please feel free to either take these comments into account (or not) then merge once CI is happy!

bors d+

Mathlib/RingTheory/Bialgebra/Basic.lean
```diff
@@ -156,3 +158,73 @@ instance toBialgebra : Bialgebra R R where
...
+open AddMonoidAlgebra
+
+variable {R : Type u} [CommSemiring R] {A : Type v} [Semiring A] [Bialgebra R A]
+
+instance instBialgebra : Bialgebra R A[T;T⁻¹] :=
+  (inferInstance : Bialgebra R A[ℤ])
+
+@[simp]
+theorem comul_T (n : ℤ) :
+    Coalgebra.comul (R := R) (T (R := A) n) = T n ⊗ₜ[R] T n := by
```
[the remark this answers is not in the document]
eric-wieser: Or
```suggestion
    Coalgebra.comul (R := R) (T n : A[T;T⁻¹]) = T n ⊗ₜ[R] T n := by
```

Mathlib/RingTheory/HopfAlgebra.lean
```diff
@@ -11,6 +11,11 @@ import Mathlib.RingTheory.Bialgebra.Basic
 In this file we define `HopfAlgebra`, and provide instances for:
 
 * Commutative semirings: `CommSemiring.toHopfAlgebra`
+* The `R`-Hopf algebra instance on the group algebra `A[G]` where `G` is a group and `A` is an
+`R`-Hopf algebra: `(Add)MonoidAlgebra.instHopfAlgebra`.
+* The `R`-Hopf algebra instance on `A[t, t⁻¹]` when `A` is an `R`-Hopf algebra:
+`LaurentPolynomial.instHopfAlgebra`. When `A = R` this corresponds to the fact that `𝔾ₘ/R` is a
+group scheme.
```
eric-wieser, line 18:
```suggestion
* The `R`-Hopf algebra instance on the group algebra `A[G]` where `G` is a group and `A` is an
  `R`-Hopf algebra: `(Add)MonoidAlgebra.instHopfAlgebra`.
* The `R`-Hopf algebra instance on `A[t, t⁻¹]` when `A` is an `R`-Hopf algebra:
  `LaurentPolynomial.instHopfAlgebra`. When `A = R` this corresponds to the fact that `𝔾ₘ/R` is a
  group scheme.
```

Mathlib/RingTheory/Bialgebra/Basic.lean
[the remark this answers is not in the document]
101damnations: Thank you both for the comments, I've applied them. Very sorry for ignoring this PR for months.

Mathlib/RingTheory/Coalgebra/Basic.lean
```diff
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ali Ramsey, Eric Wieser
 -/
 import Mathlib.Algebra.Algebra.Bilinear
+import Mathlib.Algebra.Polynomial.Laurent
```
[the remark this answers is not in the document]
101damnations: Thanks! Just to be sure, do you mean all the coalgebra, bialgebra & Hopf algebra `MonoidAlgebra` stuff in one file, or 3 separate files?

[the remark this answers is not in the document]
eric-wieser: I was originally thinking 3 separate files, but either is fine. The important thing is not to overload the files that define what a hopf algebra is.

Mathlib/RingTheory/Bialgebra/MonoidAlgebra.lean
```diff
@@ -0,0 +1,94 @@
...
+  monoid and `A` is an `R`-bialgebra.
+* `LaurentPolynomial.instBialgebra`: the `R`-bialgebra structure on the Laurent polynomials
+  `A[T;T⁻¹]` when `A` is an `R`-bialgebra.
+-/
+
+suppress_compilation
+
+namespace MonoidAlgebra
+
+open Bialgebra
```
eric-wieser, line 28: I guess you could move this to the top of the file now, rather than repeating it.

Mathlib/RingTheory/Bialgebra/MonoidAlgebra.lean
```diff
@@ -0,0 +1,94 @@
...
+-/
+import Mathlib.RingTheory.Bialgebra.Basic
+import Mathlib.RingTheory.Coalgebra.MonoidAlgebra
+
+/-!
+# The bialgebra structure on monoid algebras
+
+Given a monoid `X`, a commutative semiring `R` and an `R`-bialgebra `A`, this file collects results
+about the `R`-bialgebra instance on `A[X]` defined using `Finsupp.instCoalgebra R X A` from
+`Mathlib.RingTheory.Coalgebra.Basic`.
```
eric-wieser, line 14:
```suggestion
`Mathlib/RingTheory/Coalgebra/Basic.lean`.
```

Mathlib/RingTheory/Bialgebra/MonoidAlgebra.lean
```diff
@@ -0,0 +1,94 @@
...
+Authors: Amelia Livingston
+-/
+import Mathlib.RingTheory.Bialgebra.Basic
+import Mathlib.RingTheory.Coalgebra.MonoidAlgebra
+
+/-!
+# The bialgebra structure on monoid algebras
+
+Given a monoid `X`, a commutative semiring `R` and an `R`-bialgebra `A`, this file collects results
+about the `R`-bialgebra instance on `A[X]` defined using `Finsupp.instCoalgebra R X A` from
```
eric-wieser, line 13: perhaps:
```suggestion
about the `R`-bialgebra instance on `A[X]` inherited from the corresponding structure on its coefficients from
```

Mathlib/RingTheory/HopfAlgebra/MonoidAlgebra.lean
```diff
@@ -0,0 +1,114 @@
...
+-/
+import Mathlib.RingTheory.Bialgebra.MonoidAlgebra
+import Mathlib.RingTheory.HopfAlgebra.Basic
+
+/-!
+# The Hopf algebra structure on group algebras
+
+Given a group `G`, a commutative semiring `R` and an `R`-Hopf algebra `A`, this file collects
+results about the `R`-Hopf algebra instance on `A[G]` defined using `Finsupp.instCoalgebra R G A`
+from `Mathlib.RingTheory.Coalgebra.Basic`.
```
eric-wieser, line 14:
```suggestion
results about the `R`-Hopf algebra instance on `A[G]`, building upon
results in other files about the bialgebra structure
```
or similar.

eric-wieser (approved):
bors d+

Thanks! Just some documentation tweaks

PR #13603, opened 2024-06-07
[Merged by Bors] - feat(FiberedCategory/Fiber): define the fibers of a functor

callesonne:
We define the fibers of a functor  `p : 𝒳 ⥤ 𝒮` over objects  `S : 𝒮`, and develop some basic API. Relatively little API is developed as in an upcoming PR we introduce a class `HasFibers` which should be preferred.

Co-authored-by: Paul Lezeau <paul.lezeau@gmail.com>

---

- [x] depends on: #12982
- [x] depends on: #13416
- [x] depends on: #13410
- [x] depends on: #13393

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13416, opened 2024-05-31
[Merged by Bors] - feat(FiberedCategory/Fibered): define prefibered categories

callesonne:
We define the notion of prefibered categories, as in SGA 1 VI 6.1, and add some basic API.

Co-authored-by: Paul Lezeau <paul.lezeau@gmail.com>

---

- [x] depends on: #13410
- [x] depends on: #13393
- [x] depends on: #12978

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13611, opened 2024-06-07
[Merged by Bors] - feat(FiberedCategory/HasFibers): define HasFibers class

callesonne:
We define a class `HasFibers p` for a given functor `p : 𝒳 ⥤ 𝒮`. The point of this is to give the user the option to supply their own fiber categories, as often the standard fibers will not be definitionally the same as the (equivalent) categories one is interested in.

---

This is the final PR in a sequence of PRs developing the theory of fibered categories that I will make for now. I will wait for the reviewing to catch up before I start working on more of these again, so as to not create too big of a backlog.

- [x] depends on: #13603
- [x] depends on: #12982
- [x] depends on: #13416
- [x] depends on: #13410
- [x] depends on: #13393


joelriou:
The `HasFibers` class looks good to me. However, for the future API, it would be nice to introduce also a typeclass `HasPullbacks` which would be the data of pullback functors (when be category is prefibered) between the categories given by the `HasFibers` class, along with the universal property each of these functors should satisfy. `HasPullbacks` may extend `HasFibers`.
Then, under `HasPullbacks`, one may define the composition morphisms (for the pullback by a composition of maps in the base), which are isomorphisms in case of fibered categories.
(The idea is that if we have a pseudofunctor, there will be an obvious `HasFibers` instance attached to the associated fibered category functor, and also an obvious `HasPullbacks` instance, with very good definitional properties. Then, there may not be good definitional properties for the composition (iso)morphisms, but still they shall be equal to those that are part of the structure of a pseudofunctor.)
(I hope this makes sense. I am not sure we discussed these details already.)

callesonne:
> The `HasFibers` class looks good to me. However, for the future API, it would be nice to introduce also a typeclass `HasPullbacks` which would be the data of pullback functors (when be category is prefibered) between the categories given by the `HasFibers` class, along with the universal property each of these functors should satisfy. `HasPullbacks` may extend `HasFibers`. Then, under `HasPullbacks`, one may define the composition morphisms (for the pullback by a composition of maps in the base), which are isomorphisms in case of fibered categories. (The idea is that if we have a pseudofunctor, there will be an obvious `HasFibers` instance attached to the associated fibered category functor, and also an obvious `HasPullbacks` instance, with very good definitional properties. Then, there may not be good definitional properties for the composition (iso)morphisms, but still they shall be equal to those that are part of the structure of a pseudofunctor.) (I hope this makes sense. I am not sure we discussed these details already.)

This is what a clivage is right? If so I think we have discussed it. I think the name `HasPullbacks` is better than `HasClivage` so maybe we can go for that one. I think that @Paul-Lez has started writing some stuff related to this, and we definitely want to PR this in the future!

callesonne:
I've removed awaiting-author for now, or do you mean that I should add `HasPullbacks` already to this PR? @joelriou

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+equivalent to the fibers `Fiber p S`. One would then like to use these categories `Fib S` directly,
+instead of working through this equivalence of categories. By developing an API for the `HasFibers`
+typeclass, this will be possible.
+
+Here is an example of when this typeclass is useful. Suppose we have a presheaf of types
+`F : 𝒮ᵒᵖ ⥤ Type _`. The associated fibered category then has objects `(S, a)` where `S : 𝒮` and `a`
+is an element of `F(S)`. The fiber category `Fiber p S` is then equivalent to the discrete category
+`Fib S` with objects `a` in `F(S)`. In this case, the `HasFibers` instance is given by the
+categories `F(S)` and the functor `ι` sends `a : F(S)` to `(S, a)` in the fibered category. See
+`Presheaf.lean` for more details.
```
joelriou, line 34: It seems this example is not yet in mathlib:
```suggestion
categories `F(S)` and the functor `ι` sends `a : F(S)` to `(S, a)` in the fibered category.
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+/-- HasFibers is an extrinsic notion of fibers on a functor `p : 𝒳 ⥤ 𝒮`. It is given by a
+collection of categories `Fib S` for every `S : 𝒮` (the fiber categories), each equiped with a
+functors `ι : Fib S ⥤ 𝒳` which map constantly to `S` on the base such that the induced functor
+`Fib S ⥤ Fiber p S` is an equivalence. -/
+@[nolint checkUnivs]
+class HasFibers (p : 𝒳 ⥤ 𝒮) where
+  /-- The type of objects of the category `Fib S` for each `S`. -/
+  Fib (S : 𝒮) : Type u₃
+  /-- `Fib S` is a category. -/
+  isCategory (S : 𝒮) : Category.{v₃} (Fib S) := by infer_instance
```
joelriou, line 68:
```suggestion
  category (S : 𝒮) : Category.{v₃} (Fib S) := by infer_instance
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+- `HasFibers.pullbackObj` is a version of `IsPreFibered.pullbackObj` which ensures that the object
+lies in a given fiber. The corresponding cartesian morphism is given by `HasFibers.pullbackMap`.
+- `HasFibers.inducedMap` is a version of `IsCartesian.inducedMap` which gives the corresponding
+morphism in the fiber category.
+- `fiber_factorization` is the statement that any morphism in `𝒳` can be factored as a morphism in
+some fiber followed by a pullback.
+
+-/
+
+universe v₁ u₁ v₂ u₂ v₃ u₃
```
joelriou, line 53: It would be better to put the universes `v₃` and `u₃` at the beginning. In `HasFibers`, all the other universes are determined by the functor `p`, so that it we need to specify the universes, we would do `HasFibers.{v₃, u₃} p` instead of `HasFibers.{_, _, _, _, v₃, u₃} p`.

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+  /-- The composition with the functor `p` is *equal* to the constant functor mapping to `S`. -/
+  comp_const (S : 𝒮) : (ι S) ⋙ p = (const (Fib S)).obj S
+  /-- The induced functor from `Fib S` to the fiber of `𝒳 ⥤ 𝒮` over `S` is an equivalence. -/
+  equiv (S : 𝒮) : Functor.IsEquivalence (inducedFunctor (comp_const S))
+
+namespace HasFibers
+
+/-- The `HasFibers` on `p : 𝒳 ⥤ 𝒮` given by the fibers of `p` -/
+@[default_instance]
+instance canonical (p : 𝒳 ⥤ 𝒮) : HasFibers p where
```
joelriou, line 80: I think this should not be an instance, only a `def`.

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+  Fib := Fiber p
+  ι S := @fiberInclusion _ _ _ _ p S
+  comp_const S := fiberInclusion_comp_eq_const
+  equiv S := by exact isEquivalence_of_iso (F := 𝟭 (Fiber p S)) (Iso.refl _)
+
+section
+
+variable (p : 𝒳 ⥤ 𝒮) [HasFibers p] (S : 𝒮)
+
+instance : Category (Fib p S) := isCategory S
```
joelriou, line 90:
```suggestion
attribute [instance] isCategory
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,248 @@
...
+  /-- The induced functor from `Fib S` to the fiber of `𝒳 ⥤ 𝒮` over `S` is an equivalence. -/
+  equiv (S : 𝒮) : Functor.IsEquivalence (inducedFunctor (comp_const S))
+
+namespace HasFibers
+
+/-- The `HasFibers` on `p : 𝒳 ⥤ 𝒮` given by the fibers of `p` -/
+@[default_instance]
+instance canonical (p : 𝒳 ⥤ 𝒮) : HasFibers p where
+  Fib := Fiber p
+  ι S := @fiberInclusion _ _ _ _ p S
```
joelriou, line 82:
```suggestion
  ι S := fiberInclusion
```

joelriou:
> I've removed awaiting-author for now, or do you mean that I should add `HasPullbacks` already to this PR? @joelriou

This PR is already long enough :-)

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
callesonne: Thanks for catching that! Yeah I only have that code on the stacks project.

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
callesonne: I didn't know it worked like that, thanks a lot for the explanation!

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
callesonne: Then will it still work as a "default instance"?

joelriou: I am not sure exactly what `default_instance` do. I think we should remove it.

joelriou: My opinion is that I would not like to have default instances flying around... If we need a `HasFibers` instance in order to prove a statement (which does not already involve a `HasFibers`), then we may invoke this `canonical` in order to make the argument work. Otherwise, `[HasFibers p]` should be a variable.

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,246 @@
...
+functors `ι : Fib S ⥤ 𝒳` which map constantly to `S` on the base such that the induced functor
+`Fib S ⥤ Fiber p S` is an equivalence. -/
+@[nolint checkUnivs]
+class HasFibers (p : 𝒳 ⥤ 𝒮) where
+  /-- The type of objects of the category `Fib S` for each `S`. -/
+  Fib (S : 𝒮) : Type u₃
+  /-- `Fib S` is a category. -/
+  category (S : 𝒮) : Category.{v₃} (Fib S) := by infer_instance
+  /-- The functor `ι : Fib S ⥤ 𝒳`. -/
+  ι (S : 𝒮) : (Fib S) ⥤ 𝒳
```
joelriou, line 69:
```suggestion
  ι (S : 𝒮) : Fib S ⥤ 𝒳
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,246 @@
...
+  /-- The composition with the functor `p` is *equal* to the constant functor mapping to `S`. -/
+  comp_const (S : 𝒮) : (ι S) ⋙ p = (const (Fib S)).obj S
```
joelriou, line 71:
```suggestion
  comp_const (S : 𝒮) : ι S ⋙ p = (const (Fib S)).obj S
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,246 @@
...
+
+attribute [instance] category
+
+/-- The induced functor from `Fib p S` to the standard fiber. -/
+@[simps!]
+def inducedFunctor : Fib p S ⥤ Fiber p S :=
+  Fiber.inducedFunctor (comp_const S)
+
+/-- The natural transformation `ι S ≅ (inducedFunctor p S) ⋙ (fiberInclusion p S)` -/
+def inducedFunctor.NatIso : ι S ≅ (inducedFunctor p S) ⋙ fiberInclusion :=
```
joelriou, line 96:
```suggestion
def inducedFunctor.NatIso : ι S ≅ inducedFunctor p S ⋙ fiberInclusion :=
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,246 @@
...
+  Fiber.inducedFunctorCompIsoSelf (comp_const S)
+
+lemma inducedFunctor_comp : ι S = (inducedFunctor p S) ⋙ fiberInclusion :=
+  Fiber.inducedFunctor_comp (comp_const S)
+
+instance : Functor.IsEquivalence (inducedFunctor p S) := equiv S
+
+instance : Functor.Faithful (ι (p:=p) S) :=
```
joelriou, line 104:
```suggestion
instance : Functor.Faithful (ι (p := p) S) :=
```

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
callesonne: Thanks!

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,246 @@
...
+  (inducedFunctor _ S).preimage (homMk p S φ)
+
+@[simp]
+lemma mapPreimage_eq {S : 𝒮} {a b : Fib p S} (φ : (ι S).obj a ⟶ (ι S).obj b)
+    [IsHomLift p (𝟙 S) φ] : (ι S).map (mapPreimage φ) = φ := by
+  simp [mapPreimage, congr_hom (inducedFunctor_comp p S)]
+
+/-- The lift of an isomorphism `Φ : (ι S).obj a ≅ (ι S).obj b` lying over `𝟙 S` to an isomorphism
+in `Fib S`. -/
+noncomputable def LiftIso {S : 𝒮} {a b : Fib p S}
```
joelriou, line 142: I would suggest the following, where I also move some definitions in the `Fib` namespace:
```lean
noncomputable def Fib.homMk {S : 𝒮} {a b : Fib p S} (φ : (ι S).obj a ⟶ (ι S).obj b)
    [IsHomLift p (𝟙 S) φ] : a ⟶ b :=
  (inducedFunctor _ S).preimage (Fiber.homMk p S φ)

@[simp]
lemma Fib.map_homMk {S : 𝒮} {a b : Fib p S} (φ : (ι S).obj a ⟶ (ι S).obj b)
    [IsHomLift p (𝟙 S) φ] : (ι S).map (homMk φ) = φ := by
  simp [Fib.homMk, congr_hom (inducedFunctor_comp p S)]

@[ext]
lemma Fib.hom_ext {S : 𝒮} {a b : Fib p S} {f g : a ⟶ b}
    (h : (ι S).map f = (ι S).map g) : f = g :=
  (ι S).map_injective h

@[simps]
noncomputable def Fib.isoMk {S : 𝒮} {a b : Fib p S}
    (Φ : (ι S).obj a ≅ (ι S).obj b) (hΦ : IsHomLift p (𝟙 S) Φ.hom) : a ≅ b where
  hom := Fib.homMk Φ.hom
  inv := Fib.homMk Φ.inv
```

callesonne: Thats a great suggestion, thanks! I have updated more names to agree with this naming convention as well (e.g. `objPreimage` -> `Fib.mk`)

Mathlib/CategoryTheory/FiberedCategory/HasFibers.lean
```diff
@@ -0,0 +1,238 @@
...
+b'       b --ψ--> a
+|        |        |
+v        v        v
+R ====== R --f--> S
+```
+Then the induced map τ : b' ⟶ b can be lifted to the fiber over R -/
+noncomputable def inducedMap : b ⟶ b' :=
+  Fib.homMk (IsCartesian.map p f ψ φ)
+
+lemma inducedMap_comp : (ι R).map (inducedMap f ψ φ) ≫ ψ = φ := by
```
joelriou, line 203: Maybe `simp`?
```suggestion
@[reassoc]
lemma inducedMap_comp : (ι R).map (inducedMap f ψ φ) ≫ ψ = φ := by
```

joelriou:
Thanks!

bors merge

PR #13616, opened 2024-06-07
[Merged by Bors] - feat: basic `Vector.inductionOn` API

llllvvuu:
Also make the case names slightly nicer for when using the syntax:
```lean
induction v using Vector.inductionOn with
| nil => sorry
| cons ih => sorry
```

---

### DIFF
--- a/Mathlib/Data/Vector/Basic.lean
+++ b/Mathlib/Data/Vector/Basic.lean
@@ -447,16 +447,25 @@ and `h_cons` defines the inductive step using `∀ x : α, C w → C (x ::ᵥ w)
 This can be used as `induction v using Vector.inductionOn`. -/
 @[elab_as_elim]
 def inductionOn {C : ∀ {n : ℕ}, Vector α n → Sort*} {n : ℕ} (v : Vector α n)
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Vector/Basic.lean
```diff
@@ -447,17 +447,26 @@ and `h_cons` defines the inductive step using `∀ x : α, C w → C (x ::ᵥ w)
...
-    exact h_nil
+    exact nil
   · rcases v with ⟨_ | ⟨a, v⟩, v_property⟩
     cases v_property
-    apply @h_cons n _ ⟨v, (add_left_inj 1).mp v_property⟩
-    apply ih
+    exact cons (ih ⟨v, (add_left_inj 1).mp v_property⟩)
 #align vector.induction_on Vector.inductionOn
 
+theorem inductionOn_nil {C : ∀ {n : ℕ}, Vector α n → Sort*}
```
eric-wieser, line 460: Let's make these both `simp`

eric-wieser (approved):
bors d+

PR #13617, opened 2024-06-07
[Merged by Bors] - feat: polish user experience with the check of import-only files

grunweg:
Add a new --check flag to `mk_all`: if true, only verify the files are correct,
and print an error (with instructions how to update the file) if not.
CI can use this to give an error if the files are wrong.

(Another possibility is to automatically update the files instead: this is left for future PRs.) 

---

~~It might be worth not merging this PR yet, as `mk_all` seems to not work on Windows, for reasons not related to the script.~~ IIRC, this issue is understood now - was it fixed already?


grunweg:
Follow-up to #11874: let me CC @adomani.

There was a recent new contributor (I don't remember the name or PR right now) who got stuck at the "import all files" step. I'm hoping this error would have been more helpful.

scripts/mk_all.lean
```diff
@@ -34,13 +36,19 @@ def mkAllCLI (args : Parsed) : IO UInt32 := do
     let allFiles ← getAllModules git d
     let fileContent := ("\n".intercalate (allFiles.map ("import " ++ ·)).toList).push '\n'
     if !(← pathExists fileName) then
-      IO.println s!"Creating '{fileName}'"
+      if check then
+        IO.println s!"File {fileName} does not exist"
```
adomani, line 40:
```suggestion
        IO.println s!"File '{fileName}' does not exist"
```

scripts/mk_all.lean
```diff
@@ -34,13 +36,19 @@ def mkAllCLI (args : Parsed) : IO UInt32 := do
...
+      else
+        IO.println s!"Creating '{fileName}'"
+        IO.FS.writeFile fileName fileContent
       updates := updates + 1
-      IO.FS.writeFile fileName fileContent
     else if (← IO.FS.readFile fileName) != fileContent then
-      IO.println s!"Updating '{fileName}'"
+      if check then
+        IO.println s!"The file '{fileName}' is out of date: run `lake exe mk_all` to update it"
```
adomani, line 47:
```suggestion
        IO.println s!"The file '{fileName}' is out of date: run `lake exe mk_all{if git then " --git" else ""}` to update it"
```
(untested)

adomani (approved):
I could not check it at a computer, but looks good to me, thanks!

adomani:
Btw, there is another CI step that also builds `Archive`, `Counterexamples`, `Mathlib/Tactic`: would it make sense to use `--check` there as well?  It is not clear to me how the `check` interacts with multiple libraries.

scripts/mk_all.lean
grunweg: I would prefer *not* to do this: the `--check` flag is meant to be passed in during CI only, and CI only runs on the CI output (which equals the local output). So, I'd tend to recommend the simplest thing which works, and I guess that is this one.

urkud:
Note that before a recent refactor, we were getting nice github suggestions that allowed us to click a button to add a missing line.

scripts/mk_all.lean
grunweg: ... a moment later I realised that this line really does - it doesn't seem to hurt either. (It only prints --git if the user specified this manually, in which case mentioning the flag will not be confusing.) I've included it now.

grunweg:
> Btw, there is another CI step that also builds `Archive`, `Counterexamples`, `Mathlib/Tactic`: would it make sense to use `--check` there as well? It is not clear to me how the `check` interacts with multiple libraries.

Good suggestion! I have added `--check` there as well.
This flag looks at every file this is out of date (not just the first one --- I implemented this and also tested it locally), so this seems safe.

grunweg:
> Note that before a recent refactor, we were getting nice github suggestions that allowed us to click a button to add a missing line.

I remember! But wasn't the issue that these suggestions were too often off by one, generating a non-sorted file?

semorrison:
bors d+

grunweg:
Thanks for the review (also in the other PRs)!
bors r+

PR #13419, opened 2024-05-31
[Merged by Bors] - style: remove all easy leading `by`s

grunweg:
If a line starts with `by `, the previous line ends with `:=` and the previous line is short enough,
move the "by " to the previous line to match the style guide.

----------

The first commit was automatically generated using the auto-fix linter in #13420 (the commit is cherry-picked from this PR).
The second commit are some manual adjustments: I'm happy to drop them.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13618, opened 2024-06-07
[Merged by Bors] - style: fix last easy isolated `by`s

grunweg:
Follow-up on #13419: fix three cases with an easy fix which were merged since then, and three harder cases.

---


eric-wieser (approved):
bors merge

PR #13619, opened 2024-06-07
[Merged by Bors] - chore(*): drop some long-deprecated theorems

urkud:
---


Mathlib/Data/Option/Basic.lean
```diff
@@ -319,6 +319,10 @@ theorem orElse_none' (x : Option α) : x.orElse (fun _ ↦ none) = x := by cases
 
 #align option.ne_none_iff_is_some Option.ne_none_iff_isSome
 
+@[simp]
```
grunweg, line 322: Flagging this for fellow reviewers to note: this lemma is new (but very similar to exist ones).

grunweg (commented):
Thanks for doing this!
maintainer merge

Mathlib/Data/Option/Basic.lean
urkud: Let me revert this `@[simp]` attr and move it to a new PR.
```suggestion
```

grunweg:
The last two commits also look good to me - with one question: `AddMonoidHom.map_list_sum` was only deprecated recently; should it stay for some more time? (If I remember correctly, it was used in mathlib until May - so it's not an obviously bad lemma that can be removed without second thought.)

fpvandoorn:
bors merge

PR #352, opened 2022-07-26
[Merged by Bors] - chore: port lint-style.py from mathlib

kim-em:
Some of these are no longer relevant, but I want to see what works before cleaning up.

PR #13620, opened 2024-06-07
[Merged by Bors] - feat: rewrite file length check in Lean

grunweg:
This is the first in a series of PRs rewriting most checks from `lint-style.py` in Lean.
This PR sets up the basic infrastructure and ports the check for files longer than 1500 lines.
This includes parsing `style-exceptions.txt`.

To enable a gradual conversion of the linters, this PR adds a new executable `lint_style` (which runs the rewritten linters), and has `lint-style.sh` call this one also.

---

- this currently reads the files to lint from `{Archive,Counterexamples,Mathlib}.lean`: should this traverse the file system (like `mk_all`) instead? My reasoning is that these files are ensured to be up-to-date in CI, so in practice this amounts to similar results --- and reading one file only is easier and slightly less I/O.

This changes the ordering of the output lines in `style-exceptions.txt`: the ones on the file length come last.
I think this is in fact *more* readable in practice, given that there are only about three kinds of style exceptions left.


Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+  /-- The current file was too large: this error contains the current number of lines
+  as well as a size limit (slightly larger). On future runs, this linter will allow this file
+  to grow up to this limit. -/
+  | fileTooLong (number_lines : ℕ) (new_size_limit : ℕ) : StyleError
+  deriving BEq
+
+/-- Create the underlying error message for a given `StyleError`. -/
+def StyleError.errorMessage (err : StyleError) : String := match err with
+  | StyleError.fileTooLong current_size size_limit =>
+      s!"{size_limit} file contains {current_size} lines, try to split it up"
```
grunweg, line 36: For reviewers: this matches the current output of `lint-style.py`. #13623 will add a more human-readable variant.

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+def StyleError.normalise (err : StyleError) : StyleError := match err with
+  -- NB: keep this in sync with `parse?_errorContext` below.
+  | StyleError.fileTooLong _ _ => StyleError.fileTooLong 0 0
+
+/-- Careful: we do not want to compare `ErrorContexts` exactly; we ignore some details. -/
+instance : BEq ErrorContext where
+  beq ctx ctx' :=
+      -- XXX: `lint-style.py` was calling `resolve()` on the path before before comparing them
+      -- should we also do so?
+      ctx.path == ctx'.path
```
grunweg, line 65: In my local testing, this works as-is.

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+      -- We completely ignore line numbers of errors. Not sure if this is best.
+      -- We normalise errors before comparing them.
+      && (ctx.error).normalise == (ctx'.error).normalise
+
+/-- Output the formatted error message, containing its context. -/
+def outputMessage (errctx : ErrorContext) : String :=
+  -- We are outputting for github: duplicate file path, line number and error code,
```
grunweg, line 72: #13623 will add human-readable output here - this PR preserves existing behaviour

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+  if line.startsWith "-- " then
+    return none
+  let parts := line.split (fun c ↦ c == ' ')
+  match parts with
+    | filename :: ":" :: "line" :: _line_number :: ":" :: error_code :: ":" :: error_message =>
+      -- Turn the filename into a path. XXX: is there a nicer way to do this?
+      -- Invariant: `style-exceptions.txt` always contains Unix paths
+      -- (because, for example, in practice it is updated by CI, which runs on unix).
+      -- Hence, splitting and joining on "/" is actually somewhat safe.
+      let path : System.FilePath := System.mkFilePath (filename.split (fun c ↦ c == '/'))
```
grunweg, line 91: Eyes on this are appreciated: the scenario to avoid is
- Windows user updates the style-exceptions file, to contain system paths using back-slashes
- this code tries to read the file, but doesn't split: this is still fine on its own (as the back-slashes are fine on Windows)
- the updated code gets, pushed; somebody on Mac or Linux tries to run the updated file and fails
In other words: can I split on all separators instead?

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+            match (String.toNat? limit, String.toNat? current) with
+            | (some size_limit, some current_size) =>
+              some (StyleError.fileTooLong current_size size_limit)
+            | _ => none
+          | _ => none
+        | _ => none
+      -- Omit the line number, as we don't use it anyway.
+      return (err.map fun e ↦ (ErrorContext.mk e 0 path))
+    -- XXX: print an error on any lines which don't match the particular format.
+    | _ => -- IO.println s!"Invalid input file: line {line} doesn't match any particular format"
```
grunweg, line 110: Printing an error here doesn't work because the monads don't work out. Quickly trying to fix that failed - a suggestion is welcome.

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -0,0 +1,203 @@
...
+      -- We add about 200 lines of slack to the current file size:
+      -- small PRs will be unperturbed, but sufficiently large PRs will get nudged towards
+      -- splitting up this file.
+      return some (StyleError.fileTooLong lines.size ((Nat.div lines.size 100) * 100 + 200))
+  none
+
+end
+
+/-- All text-based linters registered in this file. -/
+def allLinters : Array TextbasedLinter := Array.mk
```
grunweg, line 157: This list is currently empty (but will be filled when other linters get ported): I'm happy to remove this for now, if so desired.

.github/workflows/build_fork.yml
```diff
@@ -54,7 +54,14 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: lint
+      - name: install elan
```
grunweg, line 57: This step is necessary as the "lint style" workflow now needs lake to run `lake exe lint_style`.

Mathlib/Tactic/Linter/TextBased.lean
grunweg: I rewrote this check: the new version should be much safer now.

Mathlib/Tactic/Linter/TextBased.lean
grunweg: Looking through this file's history, this goes back to #352. I don't see why it would be useful now. I propose to simply remove the comment and the check.

Mathlib/Tactic/Linter/TextBased.lean
semorrison: Why not just return a fake `ErrorContext` here?

grunweg: Sure, I could to this - but I don't see the advantage over just returning `none` (as I'm doing now). Can you articulate why you consider that better?

Printing an error is merely for validating the style exceptions file; I'm also fine with leaving this out: the style exceptions file is usually automatically generated, after all! (And if not, we should make updating it easier, as #13623 does.)

semorrison: Okay. Do you want to just clean up the comments here, and explain the situation (maybe that it would be nice to report an error, but awkward, and hopefully because the file is automatically generated unnecessary)?

semorrison:
bors d+

Mathlib/Tactic/Linter/TextBased.lean
grunweg: Sounds good; I added such a comment.

grunweg:
Thank you for the review!
bors r+

PR #13622, opened 2024-06-07
[Merged by Bors] - feat(CategoryTheory): a conservative functor preserving epis and monos reflects the property of being balanced

dagurtomas:
---


joelriou:
Thanks!

bors merge

PR #13623, opened 2024-06-07
feat: prettier output on style errors

grunweg:
Produce prettier output on errors for the Lean style linter:
- by default, produce error messages which are human-readable and clickable (not the goop for github's matchers); output suitable for github matchers is added as an option
- allow linting a list of files (as well as everything: just specify no files)
- add an --update flag to just add exceptions for all current errors

On top of this, re-implements the `update_style_exceptions` script in Lean, as `lean exe lint_style --regenerate`. As long as there are Python style linters, this calls into the Python linters to update the output.
When the dependent PR lands, this will be split into a follow-up.
Likewise, just re-ordering the style exceptions could be split into its own PR as well.

---

This only covers the text-based linter in Lean, sorry: if you crave for better errors, you can help by reviewing my PRs to rewrite the Python lints in Lean :-)
Feedback on naming the flag is welcome: for instance, is `--ignore` (matching `shake`) better than `--update`?

This has the side effect of re-ordering the existing style exceptions: now, they are (essentially) sorted by error type first.
I'd argue this is more useful in practice. If you want me to split this re-ordering into a separate PR, I can do so.

- [ ] depends on: #13620 for porting the file length checker to Lean
- [ ] depends on: #14012 for adding human-readable output (the first part)


grunweg:
This has been implemented now (except for just linting individual files, which can be added later). Closing.

PR #13625, opened 2024-06-07
[Merged by Bors] - feat: make `Vector.inductionOn` the default `induction_eliminator`

llllvvuu:
otherwise it's `Subtype.rec`, which (IMO) is less useful for `induction` than it is for `cases`

---


eric-wieser:
bors merge

Thanks! This chips away a little more at #12605

PR #13630, opened 2024-06-08
[Merged by Bors] - refactor(GroupTheory/Index): Replace `Fintype.card` with `Nat.card`

tb65536:
A lot of the group theory library still needs to be switched over from `Fintype.card` to `Nat.card`. This PR switches over `GroupTheory/Index.lean`. I had to add a little noise to a few other files, but these will get cleaned up once those files are switched over later.

In this file there were a few `Fintype.card` lemmas whose `Nat.card` lemmas already existed, so I deleted those rather than switching them over (they are not just redundant, but strictly weaker since they assume finiteness).

---


urkud:
Could you please add some `@[deprecated (since := "2024-06-DD")] alias`es?

semorrison:
bors merge

PR #13621, opened 2024-06-07
[Merged by Bors] - feat(CategoryTheory/Sites): reflecting and preserving local injectivity and surjectivity

dagurtomas:
This PR proves that precomposition with a cocontinuous functor preserves local injectivity and surjectivity of morphisms of presheaves, and that precomposition with a cover preserving and cover dense functor reflects the same properties.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13632, opened 2024-06-08
[Merged by Bors] - feat(CategoryTheory/Sites): transfer WEqualsLocallyBijective

joelriou:
Under certains conditions on a functor between sites `(D, K)` to `(C, J)`, we express the class `J.W` of morphisms of presheaves which become iso after sheafification as the inverse image of the class `K.W`. We deduce that the typeclass `WEqualsLocallyBijective` can be transported from `(D, K)` to `(C, J)`.

---

- [x] depends on: #13621


erdOne:
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13633, opened 2024-06-08
[Merged by Bors] - feat: Order-connected sets in `ℝⁿ` are null-measurable

YaelDillies:
Prove that the frontier of an order-connected set in `ℝⁿ` (with the `∞`-metric, but it doesn't actually matter) has measure zero. As a corollary, antichains in `ℝⁿ` have measure zero.

This is not so trivial as one might think. The proof Kexing and I came up with involves the Lebesgue density theorem.

Partially forward-port leanprover-community/mathlib#16976

---

- [x] depends on: #13642

### DIFF
--- a/Mathlib/MeasureTheory/Order/UpperLower.lean
+++ b/Mathlib/MeasureTheory/Order/UpperLower.lean
@@ -1,0 +1,147 @@
+/-
+Copyright (c) 2022 Yaël Dillies, Kexing Ying. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yaël Dillies, Kexing Ying
+-/
+import Mathlib.Analysis.Normed.Order.UpperLower
+import Mathlib.Logic.Lemmas
+import Mathlib.MeasureTheory.Covering.BesicovitchVectorSpace
+import Mathlib.Topology.Order.DenselyOrdered
+
+#align_import measure_theory.order.upper_lower from "leanprover-community/mathlib"@"b1abe23ae96fef89ad30d9f4362c307f72a55010"
+
+/-!
+# Order-connected sets are null-measurable
+
+This file proves that order-connected sets in `ℝⁿ` under the pointwise order are null-measurable.
+Recall that `x ≤ y` iff `∀ i, x i ≤ y i`, and `s` is order-connected iff
+`∀ x y ∈ s, ∀ z, x ≤ z → z ≤ y → z ∈ s`.
+
+## Main declarations
+
+* `Set.OrdConnected.null_frontier`: The frontier of an order-connected set in `ℝⁿ` has measure `0`.
+
+## Notes
+
+We prove null-measurability in `ℝⁿ` with the `∞`-metric, but this transfers directly to `ℝⁿ` with
+the Euclidean metric because they have the same measurable sets.
+
+Null-measurability can't be strengthened to measurability because any antichain (and in particular
+any subset of the antidiagonal `{(x, y) | x + y = 0}`) is order-connected.
+
+## TODO
+
+Generalize so that it also applies to `ℝ × ℝ`, for example.
+-/
+
+open Filter MeasureTheory Metric Set
+open scoped Topology
+
+variable {ι : Type*} [Fintype ι] {s : Set (ι → ℝ)} {x y : ι → ℝ} {δ : ℝ}
+
+/-- If we can fit a small ball inside a set `s` intersected with any neighborhood of `x`, then the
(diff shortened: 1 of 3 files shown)


Mathlib/MeasureTheory/Order/UpperLower.lean
```diff
@@ -0,0 +1,206 @@
...
+    (isTopologicalBasis_pi fun _ ↦ isTopologicalBasis_opens).mem_nhds_iff] at hx ⊢
+  rintro t ⟨U, ⟨U, s, hU, rfl⟩, hxU, hUt⟩
+  classical
+  -- `isTopologicalBasis_pi` does not give us a product of open sets but a product of open sets on a
+  -- finset, so we need the following nonsense to make sure every factor is open.
+  let V (i) := if i ∈ s then U i else univ
+  have hV (i) : IsOpen (V i) := by
+    unfold_let; dsimp; split_ifs with hi; exacts [hU _ hi, isOpen_univ]
+  have hVU : pi s V = pi s U := Set.pi_congr rfl fun i hi ↦ if_pos hi
+  have hxV (i) : x i ∈ V i := by unfold_let; dsimp; split_ifs with hi; exacts [hxU _ hi, mem_univ _]
```
YaelDillies, line 86: To the topologists (@PatrickMassot, @ADedecker ?), why is `isTopologicalBasis_pi` giving me those open sets whose projections are not all open? This is mathematically correct but really impractical as you see here.

Mathlib/MeasureTheory/Order/UpperLower.lean
```diff
@@ -0,0 +1,206 @@
...
+
+-- lemma isOpenMap_eval (i : ι) : IsOpenMap (Function.eval i : (∀ i, π i) → π i) := by
+--   refine (isTopologicalBasis_pi fun _ ↦ isTopologicalBasis_opens).isOpenMap_iff.2 ?_
+--   rintro _ ⟨U, s, hU, rfl⟩
+--   obtain h | h := ((s : Set ι).pi U).eq_empty_or_nonempty
+--   · simp [h]
+--   by_cases hi : i ∈ s
+--   · rw [eval_image_pi (mod_cast hi) h]
+--     exact hU _ hi
+--   · sorry
```
YaelDillies, line 68: To the topologists again, I didn't end needing this but do we really not have it?? We do have `isOpenMap_fst`/`isOpenMap_snd`.

Mathlib/MeasureTheory/Order/UpperLower.lean
YaelDillies: Moved to #13642

Mathlib/MeasureTheory/Order/UpperLower.lean
YaelDillies: Moved to #13642

Mathlib/MeasureTheory/Order/UpperLower.lean
```diff
@@ -0,0 +1,147 @@
...
+
+#align_import measure_theory.order.upper_lower from "leanprover-community/mathlib"@"b1abe23ae96fef89ad30d9f4362c307f72a55010"
+
+/-!
+# Order-connected sets are null-measurable
+
+This file proves that order-connected sets in `ℝⁿ` under the pointwise order are null-measurable.
+Recall that `x ≤ y` iff `∀ i, x i ≤ y i`, and `s` is order-connected iff
+`∀ x y ∈ s, ∀ z, x ≤ z → z ≤ y → z ∈ s`.
+
```
JasonKYi, line 19: It might be useful to sketch the proof here.

Mathlib/MeasureTheory/Order/UpperLower.lean
```diff
@@ -0,0 +1,160 @@
...
+Null-measurability can't be strengthened to measurability because any antichain (and in particular
+any subset of the antidiagonal `{(x, y) | x + y = 0}`) is order-connected.
+
+## Sketch proof
+
+1. To show an order-connected set is null-measurable, it is enough to show it has null frontier.
+2. Since an order-connected set is the intersection of its upper and lower closure, it's enough to
+  show that upper and lower sets have null frontier.
+3. WLOG let's prove it for an upper set `s`.
+4. By the Lebesgue density theorem, it is enough to show that any frontier point `x` of `s` is a
```
JasonKYi, line 37:
```suggestion
4. By the Lebesgue density theorem, it is enough to show that any frontier point `x` of `s` is not a
```

JasonKYi:
Thanks!
maintainer merge

PR #13452, opened 2024-06-01
[Merged by Bors] - feat: define `RestrictGenTopology`

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13635, opened 2024-06-08
[Merged by Bors] - feat: Characterization of proper maps for Hausdorff and compactly generated spaces

EtienneC30:
A continuous map whose codomain is a Hausdorff and compactly generated space is proper if and only if the preimages of compact sets are compact.
A Hausdorff and weakly locally compact space is compactly generated.
A Hausdorff and sequential space is compactly generated.

---


Mathlib/Topology/ProperMap.lean
```diff
@@ -3,11 +3,11 @@ Copyright (c) 2023 Anatole Dedecker. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Anatole Dedecker, Etienne Marion
 -/
-
 import Mathlib.Topology.Homeomorph
 import Mathlib.Topology.StoneCech
 import Mathlib.Topology.Filter
 import Mathlib.Order.Filter.Cofinite
+import Mathlib.Topology.Defs.Sequences
```
YaelDillies, line 10: I don't think the imports need changing at all?

Mathlib/Topology/ProperMap.lean
```diff
@@ -290,8 +299,6 @@ lemma isProperMap_restr_of_proper_of_closed {U : Set X} (hf : IsProperMap f) (hU
 lemma IsProperMap.isClosed_range (hf : IsProperMap f) : IsClosed (range f) :=
   hf.isClosedMap.isClosed_range
 
-@[deprecated (since := "2024-05-08")] alias IsProperMap.closed_range := IsProperMap.isClosed_range
-
```
YaelDillies, line 294: Any particular reason you are deleting this alias? In general, please don't remove aliases in a PR called `feat: Something` unless the alias causes you troubles for some reason

Mathlib/Topology/ProperMap.lean
```diff
@@ -308,37 +315,95 @@ lemma isProperMap_iff_isClosedMap_and_tendsto_cofinite [T1Space Y] :
...
+    [WeaklyLocallyCompactSpace Y] :
+    IsProperMap f ↔ Continuous f ∧ Tendsto f (cocompact X) (cocompact Y) :=
+  _root_.isProperMap_iff_tendsto_cocompact
+    (fun _ ↦ compactlyGenerated_of_weaklyLocallyCompactSpace)
+
+/-- Version of `SequentialSpace.isProperMap_iff_isCompact_preimage` in terms of `cocompact`. -/
+lemma SequentialSpace.isProperMap_iff_tendsto_cocompact [T2Space Y] [SequentialSpace Y] :
+    IsProperMap f ↔ Continuous f ∧ Tendsto f (cocompact X) (cocompact Y) :=
+  _root_.isProperMap_iff_tendsto_cocompact
+    (fun _ ↦ compactlyGenerated_of_sequentialSpace)
```
YaelDillies, line 405: Sounds like you should define the typeclass `Topology.IsCompactlyGenerated`

Mathlib/Topology/ProperMap.lean
EtienneC30: This was not intended, I copied the file that I had in the other PR and did not see that this line had disappeared.

Mathlib/Topology/ProperMap.lean
EtienneC30: The idea of generalizing `isProperMap_iff_isCompact_preimage` to compactly generated spaces comes from @ADedecker, but he told me that they were hesitating on the definition to choose (they are many different possible, which are not equivalent). I picked one but I feel like I shouldn't be the one to choose the Mathlib definition.

Mathlib/Topology/ProperMap.lean
EtienneC30: This import is needed for the results mentioning `SequentialSpace`. This won't be necessary when `compactlyGeneratedSpace` typeclass is implemented.

Mathlib/Topology/ProperMap.lean
YaelDillies: I'll let Anatole decide here

ADedecker: Yes the point was that we have to choose the best definition among the three defined on [Wikipedia](https://en.wikipedia.org/wiki/Compactly_generated_space). I think #13452 would allow to formulate definition 1 and 3, but probably not option 2. I meant to take the time to study the question a bit more, but I didn't get to do it. I suggested this ad-hoc approach to Étienne because it contains all the mathematics so switching to a more typical setup would be straightforward.

YaelDillies:
I'll let you a week to review and then I'll review myself

urkud:
> I'll let you a week to review and then I'll review myself

This was a month ago, so I'm going to review it now.

Mathlib/Topology/ProperMap.lean
urkud: I think that the docstring should discuss this.

urkud:
@ADedecker Is there any progress with reviewing different definitions? I think that we should fix one of them, add as a typeclass, and use it here. We may need to add all three in the future. If Definition 1 is enough for our immediate needs (i.e., implies all the properties we're going to prove about these spaces in the near future), then I suggest to go with it, because this will make our theorems that assume `[CompactlyGeneratedTopology X]` stronger.

YaelDillies:
Do you need help with the git?

EtienneC30:
Yes I am completely lost. I wanted to solve the merge conflict but I messed up the solving because I added something which in fact had been moved to another file. Anyway I've been trying to revert to commit bd60edd but I can't find a way to do it.

YaelDillies:
Is there anything of value in that branch after `bd60edd`? If not you can do `git reset HEAD~N` where `N` is the number of commits you want to revert, then do `git push --force`. WARNING: You will lose the content of the commits you've reset (well, not quite lost but much more hardly accessible).

EtienneC30:
Thank you very much, that's exactly what I needed! I had tried `reset` but then I could not push, I didn't know about the `force` option.

YaelDillies:
With great power comes great responsibility. You should use `--force` sparingly as it can overwrite other people's work if you collaborate on the same branch.

EtienneC30:
Yes I looked it up and saw this could happen. Thank you for your help!

urkud:
UPD: it looks like we're going to use Definition 2 in another series of PRs. Not sure if we need more than one definition right now.

Mathlib/Topology/Maps/Proper/Basic.lean
```diff
@@ -307,37 +317,95 @@ lemma isProperMap_iff_isClosedMap_and_tendsto_cofinite [T1Space Y] :
...
+`SequentialSpace.isProperMap_iff_isCompact_preimage`. In the future those should be inferred
+by typeclass inference. -/
+theorem isProperMap_iff_isCompact_preimage [T2Space Y]
+    (compactlyGenerated : ∀ s : Set Y, IsClosed s ↔ ∀ ⦃K⦄, IsCompact K → IsClosed (s ∩ K)) :
+    IsProperMap f ↔ Continuous f ∧ ∀ ⦃K⦄, IsCompact K → IsCompact (f ⁻¹' K) :=
+  ⟨fun hf ↦ ⟨hf.continuous, fun _ ↦ hf.isCompact_preimage⟩,
+    fun ⟨hf, h⟩ ↦ isProperMap_iff_isClosedMap_and_compact_fibers.2
+    ⟨hf, fun _ hs ↦ (compactlyGenerated _).2
+    fun _ hK ↦ image_inter_preimage .. ▸ (((h hK).inter_left hs).image hf).isClosed,
+    fun _ ↦ h isCompact_singleton⟩⟩
```
YaelDillies, line 336: `Iff` is a structure so you can do
```suggestion
    IsProperMap f ↔ Continuous f ∧ ∀ ⦃K⦄, IsCompact K → IsCompact (f ⁻¹' K) where
  mp hf := ⟨hf.continuous, fun _ ↦ hf.isCompact_preimage⟩
  mpr := fun ⟨hf, h⟩ ↦ isProperMap_iff_isClosedMap_and_compact_fibers.2
    ⟨hf, fun _ hs ↦ (compactlyGenerated _).2
    fun _ hK ↦ image_inter_preimage .. ▸ (((h hK).inter_left hs).image hf).isClosed,
    fun _ ↦ h isCompact_singleton⟩
```
which I think is nicer

EtienneC30: That's nicer indeed, I didn't know about it. Thanks!

YaelDillies (commented):
Looks okay to me. Any topologist to sign off?

maintainer merge

Mathlib/Topology/Maps/Proper/Basic.lean
```diff
@@ -43,6 +45,14 @@ particular for the theory of proper group actions. That means that our terminolo
 align with that of [Stacks: Characterizing proper maps](https://stacks.math.columbia.edu/tag/005M),
 instead our definition of `IsProperMap` coincides with what they call "Bourbaki-proper".
 
+Concerning `isProperMap_iff_isCompact_preimage`, this result should be the only one needed to link
+the definition of a proper map and the criteria "preimage of compact sets are compact", however
+the notion of compactly generated space is not yet in Mathlib so it is used as an intermediate
+result to prove
```
jcommelin, line 51:
```suggestion
the notion of compactly generated space is not yet in Mathlib (TODO)
so it is used as an intermediate result to prove
```

Mathlib/Topology/Maps/Proper/Basic.lean
```diff
@@ -307,37 +317,95 @@ lemma isProperMap_iff_isClosedMap_and_tendsto_cofinite [T1Space Y] :
 theorem Continuous.isProperMap [CompactSpace X] [T2Space Y] (hf : Continuous f) : IsProperMap f :=
   isProperMap_iff_isClosedMap_and_tendsto_cofinite.2 ⟨hf, hf.isClosedMap, by simp⟩
 
+/-- If `Y` is Hausdorff and compactly generated, then proper maps `X → Y` are exactly
+continuous maps such that the preimage of any compact set is compact.
+
+This result should be the only one needed to link the definition of a proper map and
+the criteria "preimage of compact sets are compact", but the notion of compactly generated space
+is not yet in Mathlib so we use it as an intermediate result to prove
```
jcommelin, line 325:
```suggestion
is not yet in Mathlib (TODO) so we use it as an intermediate result to prove
```

Mathlib/Topology/Maps/Proper/Basic.lean
```diff
@@ -307,37 +317,95 @@ lemma isProperMap_iff_isClosedMap_and_tendsto_cofinite [T1Space Y] :
...
-        hy.mono_right (le_principal_iff.mpr hKy)
-    -- By compactness of `f ⁻¹' K`, `𝒰` converges to some `x ∈ f ⁻¹' K`.
-    rcases (H.2 hK).ultrafilter_le_nhds _ this with ⟨x, -, hx⟩
-    exact ⟨x, hx⟩
+such that the preimage of any compact set is compact.
+
+This result is a direct consequence of `isProperMap_iff_isCompact_preimage`, because any
+Hausdorff and weakly locally compact space is compactly generated.
+In the future it should be inferred by typeclass inference, however compactly generated spaces
+are not yet in Mathlib, therefore we also add this theorem. -/
```
jcommelin, line 354:
```suggestion
are not yet in Mathlib (TODO), therefore we also add this theorem. -/
```

Mathlib/Topology/Maps/Proper/Basic.lean
```diff
@@ -307,37 +317,95 @@ lemma isProperMap_iff_isClosedMap_and_tendsto_cofinite [T1Space Y] :
...
+    eventually_atTop, ge_iff_le]
+  exact ⟨0, fun n _ ↦ hu n⟩
+
+/-- If `Y` is sequential and Hausdorff, then proper maps `X → Y` are exactly continuous maps
+such that the preimage of any compact set is compact.
+
+This result is a direct consequence of `isProperMap_iff_isCompact_preimage`, because any
+Hausdorff and sequential space is compactly generated. In the future it should be inferred
+by typeclass inference, however compactly generated spaces are not yet in Mathlib, therefore
+we also add this theorem. -/
```
jcommelin, line 377:
```suggestion
by typeclass inference, however compactly generated spaces are not yet in Mathlib (TODO),
therefore we also add this theorem. -/
```

jcommelin (commented):
Thanks!

bors d+

EtienneC30:
Thanks for the review!

bors r+

PR #13638, opened 2024-06-08
[Merged by Bors] - feat(Order/Maximal): maximality/minimality with insertion/removal for sets

apnelson1:
We add a lemma showing that if `P` is a downwards-closed predicate on sets, then `s` is maximal w.r.t. `P` 
iff `s` has property `P` but no proper insertion of `s` has property `P`. We also add the dual lemma for single removals and upwards-closed predicates. 

---


YaelDillies (commented):
I believe this is more generally true for `P : α → Prop` where `α` is a distributive atomic lattice ("proof": The same as your current proof, but with `insert x s` replaced by `t = s ⊔ a` such that `s ⋖ t` and `a` is an atom),

Do you think you could generalise (and then specialise back to `Set` and `Finset`)?

apnelson1:
> I believe this is more generally true for `P : α → Prop` where `α` is a distributive atomic lattice ("proof": The same as your current proof, but with `insert x s` replaced by `t = s ⊔ a` such that `s ⋖ t` and `a` is an atom),
> 
> Do you think you could generalise (and then specialise back to `Set` and `Finset`)?

The right generality is really covers in an order that is 'strongly atomic', which probably should get its own typeclass. For instance, it should be true for distributive + atomic or for modular + complemented + atomic, neither of which implies the other. I can add the finset versions, but I don't think a lattice version make sense until it has the right generality, which doesn't belong in this PR.

YaelDillies (commented):
It is a bit problematic that you import finsets in a basic order theory file. Can you move the finset lemmas somewhere else?

apnelson1:
> It is a bit problematic that you import finsets in a basic order theory file. Can you move the finset lemmas somewhere else?

I've moved them to `Finset.Lattice` - I can't see a better place for them, and making a new file seems out of scope here.

YaelDillies (approved):
maintainer merge

semorrison:
bors merge

PR #13639, opened 2024-06-08
[Merged by Bors] - feat: small types of morphisms in the localized category

joelriou:
Given a class of morphisms `W : MorphismProperty C`, and two objects `X` and `Y`, we introduce a type-class `HasSmallLocalizedHom.{w} W X Y` which expresses that in the localized category with respect to `W`, the type of morphisms from `X` to `Y` is `w`-small for a certain universe `w`. Under this assumption, we define `SmallHom.{w} W X Y : Type w` as the shrunk type. For any localization functor `L : C ⥤ D` for `W`, we provide a bijection `SmallHom.equiv.{w} W L : SmallHom.{w} W X Y ≃ (L.obj X ⟶ L.obj Y)` that is compatible with the composition of morphisms.

This shall be used in the refactor of `Ext`-groups in abelian categories.

---

- [x] depends on: #13956


Mathlib/CategoryTheory/Localization/SmallHom.lean
```diff
@@ -0,0 +1,202 @@
...
+  {D : Type u'} [Category.{w'} D] (L : C ⥤ D) [L.IsLocalization W]
+  {D' : Type u''} [Category.{w''} D'] (L' : C ⥤ D') [L'.IsLocalization W] (X Y Z : C)
+
+namespace MorphismProperty
+
+/-- This property holds if the type of morphisms between `X` and `Y`
+in the localized category with respect to `W : MorphismProperty C`
+is small. -/
+class HasSmallLocalizedHom : Prop where
+  small : Small.{w} (W.Q.obj X ⟶ W.Q.obj Y)
```
erdOne, line 37: Should `HasSmallLocalizedHom.small` be an instance?

Edit: Or maybe `small_of_hasSmallLocalizedHom`.

Mathlib/CategoryTheory/Localization/SmallHom.lean
```diff
@@ -0,0 +1,202 @@
...
+
+variable {X Y Z}
+
+/-- Bijection between types of morphisms in two localized categories
+for the same class of morphisms `W`. -/
+noncomputable def localizationsHomEquiv :
```
erdOne, line 43: I don't think this belongs here?

Mathlib/CategoryTheory/Localization/SmallHom.lean
```diff
@@ -0,0 +1,202 @@
...
+variable [HasSmallLocalizedHom.{w} W Y X]
+
+@[simp]
+lemma mk_comp_mkInv (f : Y ⟶ X) (hf : W f) [HasSmallLocalizedHom.{w} W Y Y] :
+    (mk W f).comp (mkInv f hf) = mk W (𝟙 Y) :=
+  (equiv W W.Q).injective (by simp [equiv_comp])
+
+@[simp]
+lemma mkInv_comp_mk (f : Y ⟶ X) (hf : W f) [HasSmallLocalizedHom.{w} W X X] :
+    (mkInv f hf).comp (mk W f)= mk W (𝟙 X) :=
```
erdOne, line 193:
```suggestion
    (mkInv f hf).comp (mk W f) = mk W (𝟙 X) :=
```

Mathlib/CategoryTheory/Localization/SmallHom.lean
joelriou: Thanks. The one for `W.Q` can be an instance, and indeed, it is better doing so in order to avoid a `letI` below.

Mathlib/CategoryTheory/Localization/SmallHom.lean
joelriou: Indeed, I made generalized this construction `Localization.homEquiv` in #13956.

Mathlib/CategoryTheory/Localization/SmallHom.lean
```diff
@@ -0,0 +1,273 @@
...
+    HasSmallLocalizedHom.{w} W X Y ↔ HasSmallLocalizedHom.{w} W X' Y' := by
+  simp only [hasSmallLocalizedHom_iff W W.Q]
+  exact small_congr (Iso.homCongr (W.Q.mapIso e) (W.Q.mapIso e'))
+
+end
+
+/-- The type of morphisms from `X` to `Y` in the localized category
+with respect to `W : MorphismProperty C` that is shrunk to `Type w`
+when `HasSmallLocalizedHom.{w} W X Y` holds. -/
+def SmallHom (X Y : C) [HasSmallLocalizedHom.{w} W X Y] : Type w :=
```
erdOne, line 78: I'm thinking maybe this should be defined for all `Small (X ⟶ Y)`? And then maybe a thin wrapper (abbrev?) over that to preserve the typeclass inference capabilities we currently have. Not sure how easy the second part is though.

joelriou: No, there must we a specific API for localization, because the morphisms here are computed in a different category from `C` (in happens in the localized category), and the point of all this API is that we can relate these types `SmallHom` (which depends only on `X`, `Y` and the class `W`) by using any functor `L : C ⥤ D` which satisfies the "universal property" of the localization.

jcommelin (commented):
Thanks :tada:

bors merge

PR #13646, opened 2024-06-08
Lean pr testing 4223

JovanGerb:
I lowered the instance priorities for `monadError`.

---


JovanGerb:
!bench

PR #13649, opened 2024-06-09
chore: redefine `Nat.div2` `Nat.bodd`

astrainfinita:
The new definitions are faster than the old ones. `Nat.binaryRec` will be moved to batteries (https://github.com/leanprover-community/batteries/pull/799) or core (https://github.com/leanprover/lean4/pull/3756), so relevant contents are moved to a new file temporarily.

---

- [x] depends on: #15567
- [x] depends on: #19666


urkud:
Please describe how and why did you change the definition(s) in the commit message.

Mathlib/Data/Int/Bitwise.lean
```diff
@@ -216,12 +216,12 @@ theorem bodd_bit (b n) : bodd (bit b n) = b := by
   cases b <;> cases bodd n <;> simp [(show bodd 2 = false by rfl)]
 
 @[simp]
-theorem testBit_bit_zero (b) : ∀ n, testBit (bit b n) 0 = b
-  | (n : ℕ) => by rw [bit_coe_nat]; apply Nat.testBit_bit_zero
+theorem bit_testBit_zero (b) : ∀ n, testBit (bit b n) 0 = b
```
eric-wieser, line 219: This rename isn't consistent with `testBit_bit_succ` below

Mathlib/Data/Nat/BinaryRec.lean
```diff
@@ -0,0 +1,133 @@
...
+-/
+
+universe u
+
+namespace Nat
+
+/-- `bit b` appends the digit `b` to the little end of the binary representation of
+  its natural number input. -/
+def bit (b : Bool) (n : Nat) : Nat :=
+  cond b (n <<< 1 + 1) (n <<< 1)
```
eric-wieser, line 26: Presumably `| 1` is also faster than `+ 1`, if you're really micro-optimizing?

Mathlib/Data/Nat/BinaryRec.lean
eric-wieser, line 1: It's very hard to review the diff if you move things here and redefine them at the same time. Could you split off a PR that first just collects these results into this file?

Mathlib/Data/Int/Bitwise.lean
eric-wieser: Matching core seems sensible, but perhaps you should do the renames separately to the refactor for visibility.

eric-wieser:
@FR-vdash-bot, please solve the merge conflicts.

Mathlib/Data/Nat/BinaryRec.lean
digama0: This doesn't look like a good change to me. We have already settled on using `2 * n` and `2 * n + 1` in place of the old `bit0, bit1`, and these play better with ring theory tactics like `ring` and `linarith`. If you are doing this to improve the performance of the function, use a `@[csimp]` lemma instead.

Mathlib/Data/Nat/Bits.lean
```diff
@@ -42,10 +42,12 @@ def boddDiv2 : ℕ → Bool × ℕ
...
 
 /-- `div2 n = ⌊n/2⌋` the greatest integer smaller than `n/2` -/
-def div2 (n : ℕ) : ℕ := (boddDiv2 n).2
+def div2 (n : ℕ) : ℕ := n >>> 1
+
+theorem div2_val (n) : div2 n = n / 2 := rfl
 
 /-- `bodd n` returns `true` if `n` is odd -/
-def bodd (n : ℕ) : Bool := (boddDiv2 n).1
+def bodd (n : ℕ) : Bool := 1 &&& n != 0
```
eric-wieser, line 50: Is there any advantage to this order vs
```suggestion
def bodd (n : ℕ) : Bool := n &&& 1 != 0
```

leanprover-community-bot-assistant:

PR #13651, opened 2024-06-09
[Merged by Bors] - feat(ENat/Basic): add more `simp`/`gcongr` lemmas

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13641, opened 2024-06-08
[Merged by Bors] - feat(Finset/Lattice): add `Finset.sup_eq_top_iff` etc

urkud:
---

- [x] depends on: #13640

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13652, opened 2024-06-09
[Merged by Bors] - chore(Polynomial/TrailingDegree): use `ENat.toNat`, golf

urkud:
---

- [x] depends on: #13651
- [x] depends on: #13641
- [x] depends on: #13640


riccardobrasca:
Nice golf, thanks!

bors merge

PR #13653, opened 2024-06-09
feat: the unusedSetOptionIn linter

adomani:
---


Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
```diff
@@ -0,0 +1,65 @@
...
+/-!
+The `unusedSetOptionIn` linter warns against "most external" unused `set_option ... in`s.
+-/
+
+open Lean Elab Command
+
+/-- converts
+* `theorem x ...` to  `some (example ... , x)`,
+* `lemma x ...`   to  `some (example ... , x)`,
+* `example ...`   to ``some (example ... , `example)``,
```
grunweg, line 20: Is this correct?
```suggestion
* `example ...`   to ``some (example ... , 'example')`,

File: Mathlib/Tactic/Linter/UnusedSetOptionIn.lean [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Damiano Testa
+-/
+
+import Mathlib.Tactic.Lemma
+import Lean.Elab.Command
+import Lean.Linter.Util
+
+/-!
+The `unusedSetOptionIn` linter warns against "most external" unused `set_option ... in`s.
```
grunweg, line 12: how about "... against outer-most `set option ... in`s which are not necessary"?

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
```diff
@@ -0,0 +1,65 @@
...
+    Syntax → m (Option (Syntax × Syntax))
+  | `($dm:declModifiers theorem $did:declId $ds* : $t $dv:declVal) => do
+    return some (← `($dm:declModifiers example $ds* : $t $dv:declVal), did.raw[0])
+  | `($dm:declModifiers lemma $did:declId $ds* : $t $dv:declVal) => do
+    return some (← `($dm:declModifiers example $ds* : $t $dv:declVal), did.raw[0])
+  | `($dm:declModifiers example $ds:optDeclSig $dv:declVal) => do
+    return some (← `($dm:declModifiers example $ds $dv:declVal), mkIdent `example)
+  | _ => return none
+
+/-- The `unusedSetOptionIn` linter warns against "most external" unused `set_option ... in`s. -/
```
grunweg, line 33: ... possibly with the same phrasing

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
```diff
@@ -0,0 +1,65 @@
...
+register_option linter.unusedSetOptionIn : Bool := {
+  defValue := true
+  descr := "enable the unusedSetOptionIn linter"
+}
+
+/-- Gets the value of the `linter.unusedSetOptionIn` option. -/
+def getSetOptionIn (o : Options) : Bool := Linter.getLinterValue linter.unusedSetOptionIn o
+
+/-- reports a warning if the "first layer" `set_option ... in` is unused. -/
```
grunweg, line 42: How about this tweak?
"Report a warning if a `set_option ... in` command is unnecessary (i.e., the code elaborates fine without).
We only report the outermost `set_option ... in` (i.e., superfluous `set_option ... in`s which are nested are not linted against)."

If you'd like, you could also add
- why is not reporting on nested superfluous set_option in is not not so bad (is this just for simplicity, as the linter can be run multiple times? that's probably fine!)
- add one sentence why this is bad: for instance, might distract or confuse --- if this option was for backwards incompatibility, the fact that it can be removed now is a useful signal.

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
```diff
@@ -0,0 +1,65 @@
...
+def findSetOptionIn (cmd : CommandElab) : CommandElab := fun stx => do
+  let mut report? := (false, default)
+  let s ← get
+  match stx with
+    | .node _ ``Lean.Parser.Command.in #[
+        .node _ ``Lean.Parser.Command.set_option _,
```
grunweg, line 48: Do you want to differentiate which option is set? For instance, if I set `profiler.threshold 50 in`, I want to have this for temporary purposes - and linting on it seems more dubious than useful.
(The `set_option` linter also covers that case - written in Python, but its rewrite in Lean is awaiting review.)

grunweg (commented):
Great idea! I know this is in draft status - I took a look anyway (and hope you don't mind).
I have a few suggestions for the docs (see below). I also wonder if "unnecessarySetOptionIn" would be a better name - since you're not checking if the option *never influences* elaboration, but merely if *things work without*.

grunweg:
The linter error in `Mathlib/Order/LiminfLimsup.lean:1331` is a false positive: with removing the set_option, compilation succeeds, but with a *warning*. (By the way, this could be another nice test case: checking e.g. that in a lemma using a deprecated declaration, the `set_option linter.deprecated false in` is not considered superfluous.)

grunweg:
Another interesting puzzle: the linter does not lint the `set_option genSizeOfSpec false in` in `SetTheory/Ordinal/Notation.lean`.
Removing that and running #lint seems to work... but perhaps that's a different class of issue than what this linter is for!

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
adomani: I was planning to discriminate on which option is being set by going over what the lniter flags.  For instance, I am not sure whether this linter should handle various thresholds/heartBeats, since maybe a more dedicated one could also do some diagnostics on the side.

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
adomani: The new doc-string is (once I will push)
```
Report a warning if a `set_option ... in` command is unnecessary
(i.e., the code elaborates fine without it).
We only report the outermost `set_option ... in` (i.e., nested, superfluous `set_option ... in`s
are not linted against).

This is useful since unnecessary `set_option ... in`s are often silent left-overs of adaptations
that are no longer needed.
Cleaning them up, helps maintaining healthy, up-to-date code.

The linter runs just on the outermost `set_option ... in` mostly for simplicity.
Consider whether emitting a warning if one of the nested `set_option ... in`s can be omitted,
by removing them one at a time.
```

grunweg: Looks great to me!

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
adomani: I ended up going for your version: the reason behind what I had, was that I was intepreting the `example` as a `Name`, but it does not matter much either way -- it is a syntax node anyway!

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
adomani: Yes, I should have recycled this using `inherit_doc`.

Mathlib/Tactic/Linter/UnusedSetOptionIn.lean
adomani: Also overhauled by the new docs on the linter.

leanprover-community-bot-assistant:

PR #13654, opened 2024-06-09
[Merged by Bors] - chore: Move the inheritance `StrictOrderedSemiring α → CharZero α` instance to a less obscure file

YaelDillies:
There is little point in avoiding to import `CharZero` in `Algebra.Order.Ring.Defs` (where `StrictOrderedSemiring` is defined) since it is so basic. Exiling the titular basic instance to an obscure file meant that people had trouble finding that instance (and it just happened to Eric).

---


eric-wieser:
bors d+

This dependency does seem pretty lightweight; can you justify in the PR description why it should go here and not even earlier?

YaelDillies:
> This dependency does seem pretty lightweight; can you justify in the PR description why it should go here and not even earlier?

This is the earliest file it can go in since `StrictOrderedSemiring` is defined there!

YaelDillies:
bors merge

PR #13657, opened 2024-06-09
[Merged by Bors] - chore: Run `lake exe shake --update`

YaelDillies:
Seems like shake wants to move things around due to the `Std`/`Batteries` rename, so I'm running `lake exe shake --update` now to avoid polluting later PRs.

---


grunweg (commented):
Thanks! I checked that all lines are only moved.
maintainer merge

PR #13658, opened 2024-06-09
[Merged by Bors] - chore: add missing spaces around :

grunweg:
Created automatically; then I manually checked the diff and reverted bad changes.

---


Mathlib/CategoryTheory/Monoidal/Functor.lean
```diff
@@ -620,7 +620,7 @@ def comp : MonoidalFunctor.{v₁, v₃} C E :=
 #align category_theory.monoidal_functor.comp CategoryTheory.MonoidalFunctor.comp
 
 @[inherit_doc]
-infixr:80
+infixr : 80
```
mattrobball, line 623: I don't recall seeing spaces with `infix` or `notation` more generally

grunweg: You're right; I meant to revert that change also.
```suggestion
infixr:80

grunweg:
Thank you for the fast review.
bors r+

PR #13660, opened 2024-06-09
[Merged by Bors] - feat(lint-style): lint on the plain string 'Adaptation note:'

grunweg:
ask to use #adaptation_note instead.
This command is a unified place to annotate all commands, and avoids typos.

--------

- [x] depends on: #13665 for rewriting the existing 63 occurrences

---


scripts/lint-style.py
```diff
@@ -350,6 +351,13 @@ def left_arrow_check(lines, path):
         newlines.append((line_nr, new_line))
     return errors, newlines
 
+def adaptation_note_check(lines, path):
+    errors = []
+    for line_nr, line in lines:
+        if "Adaptation note:" in line:
```
jcommelin, line 357: Is this check safer? Because of caps and/or missing colons?
```suggestion
        if "daptation note" in line:
```

jcommelin:
bors d+

scripts/lint-style.py
grunweg: Good idea! Did this and added a comment about this.

PR #13667, opened 2024-06-09
[Merged by Bors] - chore(build*.yml): only update Mathlib.lean once

grunweg:
The current workflow updates `Mathlib.lean` twice: there's no reason to do so. Since 'check_imported' step moved into the 'build' step, any old reasons for this duplication are now obsolete.

---

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -130,12 +130,6 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      # We update `Mathlib.lean` as a convenience here,
-      # but verify that this didn't change anything in the `check_imported` job.
-      - name: update Mathlib.lean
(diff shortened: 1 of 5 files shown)


adomani:
Thanks, this looks good to me!

I suggest toalso remove the `git diff` step that follows the remaining `lake exe mk_all` step, since the exit code of `mk_all` is the number of updated/modified files, so it will fail when the "import all" files are not up to date.

semorrison:
bors d+

@grunweg please feel free to either incorporate the suggestion to remove the `git diff` step, or propose deferring it to a separate PR.

grunweg:
Thanks for the review. Good idea; I've also removed the `git diff` step. The relevant CI steps passed, so let me
bors r+

PR #13673, opened 2024-06-09
[Merged by Bors] - feat(IsSelfAdjoint): add `aesop` tags relevant for proving `IsSelfAdjoint a` automatically

dupuisf:
These are needed for the specialized tactic `cfc_tac` to work well. This tactic is used for proving side goals related to the continuous functional calculus.

Some of the `safe` attributes are not technically safe (i.e. they might turn a provable goal into an unprovable one), but I would claim that in those cases, it's not really reasonable to expect automation to work anyway.

---


j-loreaux:
bors merge

PR #13678, opened 2024-06-09
[Merged by Bors] - feat(Linter): enforce `since :=` fields in `@[deprecated]`

urkud:
---


grunweg:
Thanks for doing this - I had toyed with that idea myself (but not worked on it yet), and now you just did it! :octopus:

jcommelin (commented):
bors d+

PR #13681, opened 2024-06-10
[Merged by Bors] - chore: reduce import in NumberTheory.Divisors

Ruben-VandeVelde:
---

I'm not entirely sure this is useful, but I noticed it in review and thought I'd try it. I wouldn't mind if we just said no either.


YaelDillies (approved):
Sure, looks easy enough.

maintainer merge

I should add that the sequence `Data.Nat.Factors`  -> `NumberTheory.Divisors` -> `Algebra.IsPrimePow` is suspicious. Each is the continuation of the next, yet they are spread across three top folders.

fpvandoorn:
bors merge

PR #13685, opened 2024-06-10
feat(Combinatorics): add definitions for network flows

niklasmohrin:
Although we might not have any super-interesting results ready for mathlib yet, I feel that the definitions from this commit are already mature enough to include.

These definitions are ported from last semester's work we did at https://github.com/niklasmohrin/lean-seminar-2023 and I have put some time in to refactor our code to `Int` and now to any parameter `R`. In the linked repository, there is some more stuff that I can send in.

One obstacle without integer values is that it is not clear why a maximum flow would exist. I think the best way to tackle this would be to prove max-flow-min-cut theorem and implement the Ford-Fulkerson algorithm. It has been pointed out on Zulip that there is an [Isabelle formalization of the Edmonds Karp Algorithm](https://www.isa-afp.org/entries/EdmondsKarp_Maxflow.html).

I feel that the definitions are pretty solid and it makes sense to include them, even if just for unifying definitions across downstream projects.

---


niklasmohrin:
Ah, there is another `Flow` in mathlib already - I renamed the one from this PR to `Network.Problem.Flow` now, the current state of the PR has two styles of naming it `Pr.Flow` and `Flow Pr` - is it important we stick to one? Or do you have another suggestion?

niklasmohrin:
I am by now working on formalizing an algorithm for finding the maximum flow and noticed that the `Network.Problem` could become a bit annoying in certain contexts, in particular when flows to different source-sink pairs are considered in the same statement. For example, to switch the network behind a flow, I have `abbrev LevelFlow := Flow { s := Pr.s, t := Pr.t : FlowProblem (F.residualNetwork.levelNetwork Pr.s) }`. So maybe, using `structure Flow (N : Network V R) (s t : V) where ...` is better? The only disadvantage is that s and t will now usually be taken for variable names - but I guess, using them when there is also `Pr.s` and `Pr.t` would be confusing anyways?

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+instance : HasSSubset (Flow Pr) where
+  SSubset F₁ F₂ := F₁.f < F₂.f
+
+instance : IsStrictOrder (Flow Pr) (· ⊂ ·) where
+  irrefl F := by simp only [instHasSSubset, lt_self_iff_false, not_false_eq_true, forall_const]
+  trans F₁ F₂ F₃ h₁₂ h₂₃ := by simp_all only [instHasSSubset, lt_trans h₁₂ h₂₃]
+
+instance : IsNonstrictStrictOrder (Flow Pr) (· ⊆ ·) (· ⊂ ·) where
+  right_iff_left_not_left F₁ F₂ := by
+    constructor <;> (intro h; simp_all only [instHasSSubset, instHasSubset]; exact h)
```
YaelDillies, line 164: Do you really really care about the subset notation? If not, I would suggest using `≤`/`<` instead as subset notation is a lot of overhead to maintain for little benefit (in fact, we are planning on moving away from it for `Set` and `Finset`).

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+
+/-- The flow, which assigns zero to all edges. -/
+def Network.Problem.nullFlow (Pr : N.Problem) : Flow Pr where
+  f _ _ := 0
+  nonneg _ _ := le_rfl
+  conservation _ _ := rfl
+  capacity := N.nonneg
+
+instance {Pr : N.Problem} : Zero Pr.Flow where
+  zero := Pr.nullFlow
```
YaelDillies, line 119: Can you inline `nullFlow` inside the `Zero` instance?
```suggestion
def Network.Problem.nullFlow (Pr : N.Problem) : Flow Pr where
  f _ _ := 0
  nonneg _ _ := le_rfl
  conservation _ _ := rfl
  capacity := N.nonneg

instance {Pr : N.Problem} : Zero Pr.Flow where
  zero.f _ _ := 0
  zero.nonneg _ _ := le_rfl
  zero.conservation _ _ := rfl
  zero.capacity := N.nonneg
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+    vertices and makes some theorem statements shorter. However, since larger cycles are still
+    permitted, this usually does not help in proofs. To the contrary, defining a flow becomes more
+    cumbersome, sometimes significantly so. -/
+@[ext]
+structure Network.Problem.Flow (Pr : N.Problem) where
+  /-- The assignment of flow units to each edge. -/
+  f : V → V → R
+  nonneg : ∀ u v, 0 ≤ f u v
+  conservation : ∀ v, v ≠ Pr.s ∧ v ≠ Pr.t → flowOut f v = flowIn f v
+  capacity : ∀ u v, f u v ≤ N.cap u v
```
YaelDillies, line 109: I would suggest following the `FunLike` design
```suggestion
structure Network.Problem.Flow (Pr : N.Problem) where
  /-- The assignment of flow units to each edge.
  
  Do NOT use. Use the coercion to function instead. -/
  toFun : V → V → R
  nonneg' : ∀ u v, 0 ≤ toFun u v
  conservation' : ∀ v, v ≠ Pr.s ∧ v ≠ Pr.t → flowOut toFun v = flowIn toFun v
  capacity' : ∀ u v, toFun u v ≤ N.cap u v
  
 instance : FunLike Pr.Flow V (V → R) where
   coe := toFun
   coe_injective' f g h := by cases f; cases g; congr'
   
 -- Here reprove `nonneg`, `conservation`, `capacity` in terms of `⇑F` where `F : Pr.Flow`, rather than `F.toFun`
```
Please grep for bits of the code above to see what I am expecting

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+end FlowDefinition
+
+namespace Network.Problem.Flow
+section FlowValue
+
+variable [OrderedAddCommGroup R] {N : Network V R} {Pr : N.Problem}
+
+/-- The value of a flow is the amount of flow units that leave the source vertex and arrive at the
+    sink vertex. In particular, this is equal to the outgoing flow of the source that does not
+    return to it again. -/
```
YaelDillies, line 130:
```suggestion
/-- The value of a flow is the amount of flow units that leave the source vertex and arrive at the
sink vertex. In particular, this is equal to the outgoing flow of the source that does not
return to it again. -/
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+-/
+import Mathlib.Algebra.BigOperators.Group.Finset
+import Mathlib.Algebra.Order.Group.Defs
+import Mathlib.Combinatorics.SimpleGraph.Basic
+
+/-!
+# Network Flows
+
+In this file we introduce network flows. A flow network can be seen as a directed graph where each
+arc of the graph has a fixed capacity. A flow from a source vertex $s$ to a sink vertex $t$ is an
```
madvorak, line 14: Use `s` and `t` in the docstring.

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+## Main definitions
+
+- `Network`: A flow network assigning capacities to all pairs of vertices. Notably, capacities and
+             non-negative and the capacity from a vertex to itself is zero.
+- `Network.Problem`: A `Network` along with fixed source and sink vertices $s$ and $t$.
+- `Flow`: A valid assignment of flow values for a given `Network.Problem`.
+
+## Notation
+
+For flows `F F' : Flow Pr`, we define two relations:
```
madvorak, line 29: What is `F F` please?

YaelDillies: It's written `F F' : Flow Pr`

madvorak: Oh, my bad!

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: The existence of the subset relation itself is very useful, mostly for subtracting one flow from another. Since the resulting flow cannot have negative values for any edge, the definition in our seminar code currently is:
```lean
def Flow.sub {F₁ F₂ : Flow Pr} (h_le : F₁ ⊆ F₂) : Flow Pr where
  f := F₂.f - F₁.f
  -- ...
```

The second use case that occurred was for transferring "flow walks" between flows when decomposing a flow into paths: https://github.com/niklasmohrin/lean-seminar-2023/blob/5a53c8e067632017bded4fbf24b346b2ffddb84d/FlowEquivalentForest/Flow/Decomposition.lean#L484

---

The motivation for the subset notation was that if you imagine `F` as a collection of paths from s to t and cycles (or alternatively as a collection of edges), the subset relation would be exactly the multiset-subset relation among these collections. I suppose that in the context of mathlib however, this is exactly the opposite of how it works for `Multiset`.

I suppose an equally valid model would be to say that LE on flows is the same as LE on their underlying function `f` - this seems good to me.

The only benefit of the current system is that flows have order by their value, so we can denote a maximum flow with `IsTop F` or use `\top` to refer to a maximum flow. Is there any other existing definition for "maximal with respect to"?

Either way, I think you are right and only having LE, but defined like subset in the current version, seems to be the right move.

YaelDillies: Yes I think it's worth being consistent here and only use `≤`/`<` instead of `⊆`/`⊂`. Mind adapting the PR?

niklasmohrin:
In addition to including the changes from the reviews, I also changed the definition of `Flow` a bit. Now, there are three types `PseudoFlow`, `PreFlow` and `Flow` that extend each other. The main motivation for this change is that pre-flows exist, so that if someone wants to work on the Push Relabel algorithm some time, there won't be a need to change it then. This has the secondary advantage that the definitions of `incoming`, `outgoing` (previously `flowIn`, `flowOut`) can be on `PseudoFlow` instead of being free functions (which was needed because they are used in the definition of `Flow`).

Secondly, I got rid of `Network.Problem` for the reasons laid out in my previous comment

YaelDillies:
Please remove `awaiting-author` when this PR is ready to be reviewed again

niklasmohrin:
After [some discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/252551-graph-theory/topic/Flow.20networks.20and.20flow.20equivalent.20forest/near/466114729), I made another change so that `Network V R extends DiGraph V`. The main motivation for me is that we can then denote paths through the network as `N.Path u v` directly instead of having `N.asDigraph.Path u v` or similar.

In the process, I removed the `loopless` condition, because it was mainly used for the conversion to `SimpleGraph` which is now unneeded.

To allow for not-caring-about-graphs, the defaults for `Adj` and the newly added `cap_zero_of_not_adj` are such that the underlying network is the complete graph. It can thus still be used like

```lean
def mynet : Network ℤ ℤ where
  cap u v := max 0 <| u - v
  nonneg := sorry
```

Please let me know what you think!

niklasmohrin:
Ah, one advantage of keeping `Network.loopless` would be that once `SimpleGraph V extends DiGraph V`, we could have `UndirectedNetwork V R extends Network V R, SimpleGraph V` (introducing `loopless` only in `UndirectedNetwork` seems odd). Maybe that would be too restricting anyways though, as there could theoretically also be non-simple graphs employed with a symmetric capacity function.

Still, this made me realize that one might want a network to refer to a more specialized version of `Digraph`, which I think is not possible in the `Network extends Digraph` design.

I suppose the most flexible design to have all kinds of networks carry along all kinds of graphs would be `structure Network (G : Digraph V) (R : Type u_R)`. However, this would defeat the main purpose for the move to the `Network extends Digraph` design :thinking:

What should we do?

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+
+open BigOperators
+
+section NetworkDefinition
+
+variable (V : Type u_V) (R : Type u_R) [Zero R] [LE R]
+
+/-- A flow network assigning non-negative capacities to all pairs of vertices. -/
+structure Network extends Digraph V where
+  Adj _ _ := True
```
Shreyas4991, line 50: what is the point of adding an Adj relation that makes the graph complete. Flow problems are more general

Shreyas4991 (commented):
I am confused by adding an Adj relation that seems to make the graph a complete graph.

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: This is not what this means. I am updating the default value for the `Adj` field of `Digraph` (see https://lean-lang.org/lean4/doc/struct.html#default-field-values). Please see my last two comments in this github thread for my thoughts on this

Shreyas4991: Apologies, I forgot how lean allows new defaults for extended structures. I am happy to mark this comment resolved.

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+For flows `f f' : N.PseudoFlow`, we define `f ≤ f'` as the _subflow relation_, indicating that for
+all `u v : V`, `f u v ≤ f' u v`.
+
+## References
+
+See [goldberg_network_1989] for a thorough introduction to the topic.
+
+## Tags
+
+flow, network, max-flow, maxflow, cut, min-cut, mincut, graph, subflow
```
Shreyas4991, line 37: Do we really need two different tags `maxflow` and `max-flow`?

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Niklas Mohrin
+-/
+import Mathlib.Algebra.BigOperators.Group.Finset
+import Mathlib.Combinatorics.Digraph.Basic
+
+/-!
+# Network Flows
+
+In this file we introduce network flows. A flow network can be seen as a directed graph where each
```
Shreyas4991, line 12:
```suggestion
In this file we introduce network flows. A flow network is a directed graph where each
```

Shreyas4991: I haven't checked how this affects the character limits in each line

Shreyas4991 (approved):
Seems fine by me.

Mathlib/Combinatorics/Flow/Basic.lean
Shreyas4991: same for `mincut` and `min-cut`

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+
+section PreFlowDefinition
+
+variable [Preorder R] [AddCommMonoid R] (N : Network V R) (s t : V)
+
+/-- A `PreFlow` from `s` to `t` is a `PseudoFlow` such that all vertices except `s` and `t` have
+non-negative excess, that is, the outgoing flow is at most as much as the incoming flow. -/
+@[ext]
+structure PreFlow extends N.PseudoFlow where
+  excess_nonneg : ∀ v, v ≠ s ∧ v ≠ t → toPseudoFlow.outgoing v ≤ toPseudoFlow.incoming v
```
b-mehta, line 131:
```suggestion
  excess_nonneg : ∀ v, v ≠ s → v ≠ t → toPseudoFlow.outgoing v ≤ toPseudoFlow.incoming v
```
The usual Lean pattern is to curry constructors in this way. You'll need to change a couple of proofs for this, but it should be easy to adjust to!

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+namespace PreFlow
+
+variable {N s t}
+
+instance : Zero (N.PreFlow s t) where
+  zero := { (0 : N.PseudoFlow) with excess_nonneg := fun _ _ ↦ le_rfl }
+
+/-- The value of a flow is the amount of flow units that leave the source vertex and arrive at the
+sink vertex. In particular, this is equal to the excess of `t`. -/
+def value [Sub R] (f : N.PreFlow s t) := f.excess t
```
b-mehta, line 142: It looks to me like this definition will also work for a pseudoflow, without changes? Perhaps it's less meaningful there, but it seems perfectly well-defined.

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+section FlowDefinition
+
+variable [Preorder R] [AddCommMonoid R] (N : Network V R) (s t : V)
+
+/-- A `Flow` from `s` to `t` is a `PreFlow s t`, but instead of requiring non-negative excess, it
+requires that the excess is 0 (also called _flow conservation_). -/
+@[ext]
+structure Flow extends N.PreFlow s t where
+  conservation : ∀ v, v ≠ s ∧ v ≠ t → toPseudoFlow.outgoing v = toPseudoFlow.incoming v
+  excess_nonneg v hv := le_of_eq <| conservation v hv
```
b-mehta, line 156:
```suggestion
  excess_nonneg v hv := (conservation v hv).le
```
if you'd like!

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+section SummingLemmas
+variable [Preorder R] [AddCommGroup R] {N : Network V R}
+
+namespace PseudoFlow
+
+variable (f : N.PseudoFlow)
+
+lemma sum_outgoing_eq_sum_incoming : ∑ v, f.outgoing v = ∑ v, f.incoming v := by
+  unfold outgoing incoming
+  rw[Finset.sum_comm]
```
b-mehta, line 175:
```suggestion
  rw [Finset.sum_comm]
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+
+@[simp]
+lemma sum_excess_zero : ∑ v, f.excess v = 0 := by simp[excess, sum_outgoing_eq_sum_incoming]
```
b-mehta, line 178:
```suggestion
lemma sum_excess_zero : ∑ v, f.excess v = 0 := by simp [excess, sum_outgoing_eq_sum_incoming]
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+
+end PseudoFlow
+
+namespace Flow
+
+variable {s t : V} (f : N.Flow s t)
+
+lemma excess_s_zero_of_eq (hst : s = t) : f.excess s = 0 := by
```
b-mehta, line 186:
```suggestion
lemma excess_source_zero_of_eq (hst : s = t) : f.excess s = 0 := by
```
or maybe just `excess_zero_of_eq`. In general it's better to avoid referencing variable names in your lemma names - remember that the person using the lemma might not have used the same name as you did! Instead, giving a semantic meaning can be better

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,232 @@
...
+  rw[← f.sum_excess_zero]
```
b-mehta, line 187:
```suggestion
  rw [← f.sum_excess_zero]
```
and similarly below

b-mehta (commented):
This looks generally quite nice!

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: fair enough, `s` and `t` where ubiquitous for source and sink for me so far, but your suggestion is better. Should we go with `excess_{source, sink}_zero_of_eq` or with `excess_zero_of_eq` and `excess_zero_of_eq'`? I think I slightly prefer the first option

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: I thought about this too, but since pseudoflows don't have any associated source or sink in the current definition, there is no vertex to fix for the definition of `value`. This of course raises the question whether `PseudoFlow` should have these vertices, but I figured that if it is not needed in the definition, it doesn't really make sense (I don't think that there will be much theory regarding `PseudoFlow` directly, so it will probably be easy to change the definition either way later)

One interesting observation I had when splitting the definitions for `Flow` into `PseudoFlow` and `PreFlow` is that for `Flow`, both `-f.excess s` and `f.excess t` make perfect sense for `value` (and I find the former more intuitive), but with preflows, only `f.excess t` makes sense. Without the restrictions given by `PreFlow`, there is no vertex left with a sensible excess to pick for `value`

What do you think?

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: I don't know, @YaelDillies ?

YaelDillies: Yeah, the point of tags is for when people grep through mathlib, so it's useful to have all versions they could possibly type

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+
+Note that this definition, in contrast to what can sometimes be found in the literature, does
+not require that $f(u, v) = -f(v, u)$. This extra requirement eliminates cycles between two
+vertices and makes some theorem statements shorter. However, since larger cycles are still
+permitted, this usually does not help in proofs. To the contrary, defining a flow becomes more
+cumbersome, sometimes significantly so. -/
+@[ext]
+structure PseudoFlow (N : Network V R) where
+  /-- The assignment of flow values to each arc.
+  Do NOT use. Use the coercion to function instead. -/
```
YaelDillies, line 79: nit:
```suggestion

    Do NOT use. Use the coercion to function instead. -/
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+end PseudoFlow
+end PseudoFlowDefinition
+
+variable [Fintype V]
+
+namespace PseudoFlow
+
+variable [Preorder R] [AddCommMonoid R] {N : Network V R} (f : N.PseudoFlow) (v : V)
+
+/-- The incoming flow of a vertex. --/
```
YaelDillies, line 112: Careful! This will make your docstring end with a `-`
```suggestion
/-- The incoming flow of a vertex. -/
```
Same below

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+namespace PseudoFlow
+
+variable (f : N.PseudoFlow)
+
+lemma sum_outgoing_eq_sum_incoming : ∑ v, f.outgoing v = ∑ v, f.incoming v := by
+  unfold outgoing incoming
+  rw [Finset.sum_comm]
+
+@[simp]
+lemma sum_excess_zero : ∑ v, f.excess v = 0 := by simp [excess, sum_outgoing_eq_sum_incoming]
```
YaelDillies, line 176:
```suggestion
lemma sum_excess_eq_zero : ∑ v, f.excess v = 0 := by simp [excess, sum_outgoing_eq_sum_incoming]
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+
+end PseudoFlow
+
+namespace Flow
+
+variable {s t : V} (f : N.Flow s t)
+
+lemma excess_source_zero_of_eq (hst : s = t) : f.excess s = 0 := by
```
YaelDillies, line 184:
```suggestion
lemma excess_source_eq_zero_of_eq (hst : s = t) : f.excess s = 0 := by
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+  rw [← f.sum_excess_zero]
+  apply Eq.symm
+  apply Finset.sum_eq_single_of_mem s (Finset.mem_univ _)
+  intro v _ hs
+  simp [PseudoFlow.excess, f.conservation v hs (hst ▸ hs)]
+
+lemma value_zero_of_eq (hst : s = t) : f.value = 0 := by
```
YaelDillies, line 191:
```suggestion
lemma value_eq_zero_of_eq (hst : s = t) : f.value = 0 := by
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+  rw [← f.sum_excess_zero]
+  apply Eq.symm
+  apply Finset.sum_eq_add_of_mem s t (Finset.mem_univ _) (Finset.mem_univ _) hst
+  intro v _ hv
+  simp [PseudoFlow.excess, f.conservation v hv.left hv.right]
+
+lemma excess_sink_eq_neg_excess_source : f.excess t = -f.excess s := by
+  rw [excess_source_eq_neg_excess_sink, neg_neg]
+
+lemma value_eq_outgoing_minus_incoming_source : f.value = f.outgoing s - f.incoming s := by
```
YaelDillies, line 210: The operator is called `sub`
```suggestion
lemma value_eq_outgoing_sub_incoming_source : f.value = f.outgoing s - f.incoming s := by
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+
+section Subflows
+namespace PseudoFlow
+
+variable [PartialOrder R] [AddCommGroup R] {N : Network V R}
+
+@[simp]
+instance : PartialOrder (N.PseudoFlow) := PartialOrder.lift toFun instFunLike.coe_injective'
+
+instance : OrderBot (N.PseudoFlow) where
```
YaelDillies, line 224:
```suggestion
instance : OrderBot N.PseudoFlow where
```
Same everywhere else

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+end Flow
+end SummingLemmas
+
+section Subflows
+namespace PseudoFlow
+
+variable [PartialOrder R] [AddCommGroup R] {N : Network V R}
+
+@[simp]
+instance : PartialOrder (N.PseudoFlow) := PartialOrder.lift toFun instFunLike.coe_injective'
```
YaelDillies, line 222: nit:
```suggestion
instance : PartialOrder (N.PseudoFlow) := .lift toFun instFunLike.coe_injective'
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+    simp [f.excess_source_zero_of_eq]
+  apply eq_neg_of_add_eq_zero_left
+  rw [← f.sum_excess_zero]
+  apply Eq.symm
+  apply Finset.sum_eq_add_of_mem s t (Finset.mem_univ _) (Finset.mem_univ _) hst
+  intro v _ hv
+  simp [PseudoFlow.excess, f.conservation v hv.left hv.right]
+
+lemma excess_sink_eq_neg_excess_source : f.excess t = -f.excess s := by
+  rw [excess_source_eq_neg_excess_sink, neg_neg]
```
YaelDillies, line 208: Lemmas should be simpler on their RHS than on their LHS. If you make sure this is so, then those lemmas can be made simp and have shorter names
```suggestion
@[simp] lemma neg_excess_sink : -f.excess t = f.excess s := by
  wlog hst : s ≠ t
  · rw [not_not] at hst
    subst t
    simp [f.excess_source_zero_of_eq]
  apply eq_neg_of_add_eq_zero_left
  rw [← f.sum_excess_zero]
  apply Eq.symm
  apply Finset.sum_eq_add_of_mem s t (Finset.mem_univ _) (Finset.mem_univ _) hst
  intro v _ hv
  simp [PseudoFlow.excess, f.conservation v hv.left hv.right]

@[simp] lemma neg_excess_source : -f.excess s = f.excess t := by
  rw [← neg_excess_sink, neg_neg]
```

YaelDillies (commented):
Can you explain the `source`/`sink` naming convention in a new `## Naming convention` section of the module doc? Thanks!

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: Just to be sure: Is this about adding a newline, or about indentation? There is a similar case where there is no indent: https://github.com/leanprover-community/mathlib4/blob/97c3e004ceb529cb6e9cc64a64cd08710637603f/Mathlib/Order/Hom/Lattice.lean#L59-L62

Mathlib/Combinatorics/Flow/Basic.lean
niklasmohrin: Fair point! I wonder if "simpler" should mean "there is no unary minus anymore", or whether it should be "all f.excess terms are normalized to the same vertex" for example `f.excess s`. What do you think? I myself am not sure if either version could cause annoying `simp` behavior

b-mehta: Removing the unary minus should be pretty consistently sensible. Locally the version which normalises to `s` should also be fine, since the type of `f` determines what `s` is; but my expectation is that is much more likely to cause weird behaviour elsewhere.
All that to say, I agree with Yaël here.

Mathlib/Combinatorics/Flow/Basic.lean
b-mehta: Yup, let's keep it as is.

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+  Do NOT use. Use the coercion to function instead. -/
+  toFun : V → V → R
+  nonneg' : ∀ u v, 0 ≤ toFun u v
+  capacity' : ∀ u v, toFun u v ≤ N.cap u v
+
+namespace PseudoFlow
+
+variable {N : Network V R}
+
+instance instFunLike : FunLike N.PseudoFlow V (V → R) where
```
niklasmohrin, line 88: After this was suggested, I split `Flow` into the three structures. Now, I still would want the following lemmas to hold, but coercions fail:
```lean
lemma foo (f : N.PreFlow s t) : True := .intro
example {s t : V} (f : N.Flow s t) := foo f
example {s t : V} (f : N.Flow s t) : R := f s t
```
I fiddled around with `Coe`, `CoeOut` and `CoeFun` a bit, but couldn't get it to work. Any ideas?

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,235 @@
...
+  coe := PseudoFlow.toFun
+  coe_injective' _ _ := PseudoFlow.ext
+
+variable (f : N.PseudoFlow)
+
+lemma nonneg : ∀ u v, 0 ≤ f u v := f.nonneg'
+
+lemma capacity : ∀ u v, f u v ≤ N.cap u v := f.capacity'
```
YaelDillies, line 101:
```suggestion
lemma le_cap : ∀ u v, f u v ≤ N.cap u v := f.capacity'
```

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,235 @@
...
+
+section PreFlowDefinition
+
+variable [Preorder R] [AddCommMonoid R] (N : Network V R) (s t : V)
+
+/-- A `PreFlow` from `s` to `t` is a `PseudoFlow` such that all vertices except `s` and `t` have
+non-negative excess, that is, the outgoing flow is at most as much as the incoming flow. -/
+@[ext]
+structure PreFlow extends N.PseudoFlow where
+  excess_nonneg : ∀ v, v ≠ s → v ≠ t → toPseudoFlow.outgoing v ≤ toPseudoFlow.incoming v
```
YaelDillies, line 134: There's no `excess` in the type of this field, so it is misnamed/misstated

Mathlib/Combinatorics/Flow/Basic.lean
```diff
@@ -0,0 +1,235 @@
...
+
+section FlowDefinition
+
+variable [Preorder R] [AddCommMonoid R] (N : Network V R) (s t : V)
+
+/-- A `Flow` from `s` to `t` is a `PreFlow s t`, but instead of requiring non-negative excess, it
+requires that the excess is 0 (also called _flow conservation_). -/
+@[ext]
+structure Flow extends N.PreFlow s t where
+  conservation : ∀ v, v ≠ s → v ≠ t → toPseudoFlow.outgoing v = toPseudoFlow.incoming v
```
YaelDillies, line 158:
```suggestion
  outgoing_eq_ingoing : ∀ v, v ≠ s → v ≠ t → toPseudoFlow.outgoing v = toPseudoFlow.incoming v
```

Mathlib/Combinatorics/Flow/Basic.lean
YaelDillies: You need to write one `FunLike` instance per structure

leanprover-community-bot-assistant:

niklasmohrin:
needs updating, but to be honest I have lost interest in contributing since my Zulip chat didn't lead anywhere productive

PR #13686, opened 2024-06-10
[Merged by Bors] - feat: some finset lemmas

fpvandoorn:
* The increase in imports in `Mathlib.Data.Finset.Update` is not problematic: it is only imported in MeasureTheory files.
* Add `DepRewrite` to `Tactic.Common`.
* Change the normal form of `{x} ∪ {y}` to `{x, y}` (previously `{y, x}`).

---


Mathlib/Data/Finset/Basic.lean
```diff
@@ -1507,6 +1507,9 @@ theorem empty_union (s : Finset α) : ∅ ∪ s = s :=
   ext fun x => mem_union.trans <| by simp
 #align finset.empty_union Finset.empty_union
 
+lemma singleton_union : {a} ∪ s = insert a s := by ext; simp
+lemma union_singleton : s ∪ {a} = insert a s := by ext; simp [or_comm]
```
urkud, line 1511: For `Set`s, these are `@[simp]` lemmas.

fpvandoorn, line 1511:
```suggestion
@[simp] lemma singleton_union : {a} ∪ s = insert a s := by ext; simp
@[simp] lemma union_singleton : s ∪ {a} = insert a s := by ext; simp [or_comm]
```

Mathlib/Data/Finset/Basic.lean
```diff
@@ -1280,6 +1280,13 @@ theorem union_empty (s : Finset α) : s ∪ ∅ = s :=
 theorem empty_union (s : Finset α) : ∅ ∪ s = s :=
   ext fun x => mem_union.trans <| by simp
 
+/- Todo: make these two lemmas `simp`. If so, make sure to formulate `union_singleton` first,
+so that it has lower `simp`-priority.
+The reason is that together with `Finset.mem_insert` we don't break the natural order of the
+disjunctions. -/
+lemma union_singleton : s ∪ {a} = insert a s := by ext; simp [or_comm]
+lemma singleton_union : {a} ∪ s = insert a s := by ext; simp
```
YaelDillies, line 1288: Can you place that next to `insert_eq`?

Mathlib/Data/Finset/Basic.lean
```diff
@@ -3018,22 +3025,62 @@ def Finset.union (s t : Finset α) (h : Disjoint s t) :
...
+theorem Finset.union_inr (h : Disjoint s t) (y : t) :
     Equiv.Finset.union s t h (Sum.inr y) = ⟨y, Finset.mem_union.mpr <| Or.inr y.2⟩ :=
   rfl
 
+@[deprecated (since := "2024-06-07")] alias Finset.union_symm_inl := Finset.union_inl
+@[deprecated (since := "2024-06-07")] alias Finset.union_symm_inr := Finset.union_inr
+
+@[simp]
+theorem Finset.union_symm_left (h : Disjoint s t) {i : α} (hi : i ∈ s)
+    (hi' : i ∈ s ∪ t) : (Equiv.Finset.union s t h).symm ⟨i, hi'⟩ = Sum.inl ⟨i, hi⟩ := by
```
YaelDillies, line 3042:
```suggestion
theorem Finset.union_symm_left (h : Disjoint s t) {i : α} (hi : i ∈ s) :
    (Equiv.Finset.union s t h).symm ⟨i, subset_union_left hi⟩ = Sum.inl ⟨i, hi⟩ := by
```
for backwards rewriting. Same below

Mathlib/Data/Finset/Update.lean
```diff
@@ -61,4 +62,18 @@ theorem updateFinset_updateFinset {s t : Finset ι} (hst : Disjoint s t)
...
+    Function.update (updateFinset x s y) i z = updateFinset x (s ∪ {i})
+      ((Equiv.piFinsetUnion π <| Finset.disjoint_singleton_right.mpr hi) (y, uniqueElim z)) := by
+  rw [update_eq_updateFinset, updateFinset_updateFinset]
+
+theorem updateFinset_congr (h : s = t) :
+    updateFinset x s y = updateFinset x t (fun i ↦ y ⟨i, h ▸ i.prop⟩) := by
+  subst h; rfl
+
+theorem updateFinset_univ [Fintype ι] {y : ∀ i : Finset.univ, π i} :
+    updateFinset x .univ y = fun i : ι ↦ y ⟨i, Finset.mem_univ i⟩ := by
```
YaelDillies, line 76: Can we have an applied version of this?

Mathlib/Data/Finset/Lattice/Lemmas.lean
```diff
@@ -64,6 +64,13 @@ theorem Nonempty.inr {s t : Finset α} (h : t.Nonempty) : (s ∪ t).Nonempty :=
 theorem insert_eq (a : α) (s : Finset α) : insert a s = {a} ∪ s :=
   rfl
 
+/- Todo: make these two lemmas `simp`. If so, make sure to formulate `union_singleton` first,
+so that it has lower `simp`-priority.
+The reason is that together with `Finset.mem_insert` we don't break the natural order of the
+disjunctions. -/
+lemma union_singleton : s ∪ {a} = insert a s := by ext; simp [or_comm]
+lemma singleton_union : {a} ∪ s = insert a s := rfl
```
YaelDillies, line 72: Not so sure making those simp is a good idea, but I guess you can try in another PR

fpvandoorn: they have been made simp-lemmas in the meantime, and I reordered them

Mathlib/Data/Finset/Basic.lean
fpvandoorn: If I do that, `simp` cannot rewrite from left-to-right anymore if the proof doesn't exactly match.

Mathlib/Data/Finset/Lattice/Lemmas.lean
```diff
@@ -68,13 +68,17 @@ theorem Nonempty.inr {s t : Finset α} (h : t.Nonempty) : (s ∪ t).Nonempty :=
 theorem insert_eq (a : α) (s : Finset α) : insert a s = {a} ∪ s :=
   rfl
 
+/- We formulate `singleton_union` after `union_singleton` to ensure that `{x} ∪ {y}`
+simplifies to `{x, y}` and not `{y, x}`. -/
```
robin-carlier: Maybe usage of priority for the simp lemmas a more principled way and less brittle way to achieve this? e.g tagging `singleton_union` with `@[simp (default - 1)]` (with a comment on the line tagging it)

Mathlib/Data/Finset/Basic.lean
YaelDillies: ?! This is a serious limitation of simp! Is this documented anywhere?

Mathlib/Tactic/Common.lean
```diff
@@ -55,6 +55,7 @@ public import Mathlib.Tactic.Conv
 public import Mathlib.Tactic.Convert
 public import Mathlib.Tactic.DefEqTransformations
 public import Mathlib.Tactic.DeprecateTo
+public import Mathlib.Tactic.DepRewrite
```
YaelDillies, line 58: Maybe make this a separate PR to gather consensus

Mathlib/RingTheory/MvPolynomial/IrreducibleQuadratic.lean
```diff
@@ -220,7 +220,7 @@ theorem irreducible_sumSMulXSMulY [IsDomain R]
   · rwa [MvPolynomial.mem_support_iff, hcoeff, ← Finsupp.mem_support_iff]
   · simp [ι]
   · rw [hsupp, Finset.coe_map, ι.injective.injOn.pairwiseDisjoint_image]
-    suffices (c.support : Set n).PairwiseDisjoint fun x ↦ {Sum.inr x, Sum.inl x} by
+    suffices (c.support : Set n).PairwiseDisjoint fun x ↦ {Sum.inl x, Sum.inr x} by
```
YaelDillies, line 223: Can you explain in the PR description why this change happens?

Mathlib/Data/Finset/Lattice/Lemmas.lean
```diff
@@ -72,9 +72,13 @@ theorem insert_eq (a : α) (s : Finset α) : insert a s = {a} ∪ s :=
...
 
-@[simp, grind =]
+/- We lower the simp-priority of `union_singleton` to ensure that `{x} ∪ {y}`
+simplifies to `{x, y}` and not `{y, x}`. -/
+
+@[simp 900, grind =]
 lemma union_singleton (x : α) (s : Finset α) : s ∪ {x} = insert x s := by
-  rw [Finset.union_comm, singleton_union]
+  rw [Finset.union_comm]
+  rfl
```
YaelDillies: Why is this proof changing?

Mathlib/Data/Finset/Basic.lean
fpvandoorn: Dunno, irrelevant to this PR. I expect that there is a lean4 issue open about it.

Mathlib/Data/Finset/Lattice/Lemmas.lean
fpvandoorn: Left-over from reordering the lemmas, I'll change it back.

fpvandoorn:
```suggestion
  rw [Finset.union_comm, singleton_union]
```

Mathlib/RingTheory/MvPolynomial/IrreducibleQuadratic.lean
fpvandoorn: The normal form of `{x} ∪ {y}` changed

Mathlib/Tactic/Common.lean
fpvandoorn: https://leanprover.zulipchat.com/#narrow/channel/345428-mathlib-reviewers/topic/consensus.20gathering.3A.20DepRewrite/with/572529908

YaelDillies: Cool, can you make it a separate PR to avoid scope-creep?

fpvandoorn: no

fpvandoorn: > Cool, can you make it a separate PR to avoid scope-creep?

I've gotten unreasonably annoyed at this comment, both yesterday and today. Why do you want me to make this 1-line change into a separate PR? 

Is 61 lines too long for a PR? Is adding an import to get a working proof "scope creep" and should you always first increase the imports, and only in a follow-up PR use the lemma/tactic that required the import? Is it because the import is in `Tactic.Common`? Should I import it to `Mathlib.Data.Finset.Basic` instead, so that `rw!` is only available if you import finsets? Does Mathlib get better if the two changes are separated in the history? Is the fact that 9 reviewers have explicitly approved this change not enough, and would you like an 11th opinion?

It feels to me that you are just creating unnecessary barriers towards contributing to Mathlib here.

I should probably quit contributing Mathlib for a while.

j-loreaux (commented):
@fpvandoorn I agree, no need to make a separate PR here.

bors merge

Mathlib/Tactic/Common.lean
YaelDillies: Apologies, I did not mean to insist. I merely asked again because you hadn't explicitly said no the first time around, and I thought you might have forgotten. I also think it's fair to review a PR as new when it comes out of a year of inactivity.

The context is that I have grown more intolerant to scope-creep recently. It bothers me that important changes are drowned in unrelated PRs. This creates a small cost every time someone tries to understand why or how something happened, which is something that could happen an unbounded number of times. It also complicates searching through in-flight PRs for changes relevant to stuff you are working on. I have been more than once bitten by negative changes I did not review simply because they were hidden in PRs whose title signaled that I shouldn't be caring about them. As for this PR, sure enough it was mentioned on Zulip as the one importing `rw!` in `Tactic.Common` right after it was merged.

fpvandoorn: Thank you for responding.

I agree that it's good to defend against scope-creep if it leads to undocumented changes or lower review quality. I don't think it those concerns applied here, but I don't want to argue about it (further). I did overreact to your request.

PR #13688, opened 2024-06-10
[Merged by Bors] - feat(Data/Set/Function): simp lemmas for bij/surj/mapsTo and empty set 

apnelson1:
We add `simp` lemmas for `SurjOn`, `MapsTo` and `BijOn` in the cases where the terms force sets to be empty. 

---


urkud:
Thanks! :tada:
bors merge

PR #10864, opened 2024-02-22
feat: a deprecation tool

adomani:
###  `deprecate_to` -- a deprecation tool

Writing
```lean
deprecate_to id₁ id₂ ... idₙ 2024-02-23
theorem easy₁ : True := .intro
```
where `id₁ id₂ ... idₙ` is a sequence of identifiers adds to the environment the following aliases:
```lean
@[deprecated] alias easy₁ := id₁
@[deprecated] alias easy₂ := id₂
...
@[deprecated] alias easyₙ := idₙ
```
where `easy₁ easy₂ ... easyₙ` are the non-blacklisted declarations autogenerated by the initial theorem.

Some relevant Zulip threads:
* [deprecation automation](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/deprecation.20automation)
* [`Try this` with a sequence of commands](https://leanprover.zulipchat.com/#narrow/stream/239415-metaprogramming-.2F-tactics/topic/Try.20this.20with.20a.20sequence.20of.20commands)
* [`Mathlib` deprecation cycle](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/Mathlib.20deprecation.20cycle)

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13691, opened 2024-06-10
[Merged by Bors] - feat: a `Try this` deprecation tool

adomani:
### `deprecate to` -- a deprecation tool

Writing
```lean
deprecate to new_name new_name₂ ... new_nameₙ
theorem old_name : True := .intro
```
where `new_name new_name₂ ... new_nameₙ` is a sequence of identifiers produces the
`Try this` suggestion:
```lean
theorem new_name : True := .intro

@[deprecated (since := "YYYY-MM-DD")] alias old_name := new_name

@[deprecated (since := "YYYY-MM-DD")] alias old_name₂ := new_name₂
...

@[deprecated (since := "YYYY-MM-DD")] alias old_nameₙ := new_nameₙ
```
where `old_name old_name₂ ... old_nameₙ` are the non-blacklisted declarations
(auto)generated by the initial command.

The "YYYY-MM-DD" entry is today's date and it is automatically filled in.

(This evolved from the code of #10864, before switching to the "macro" setting.  The code here suggests a `Try this`.)

---


YaelDillies:
Can you make it use the new `(since := "2024-02-23")` syntax and let it automatically figure out the date?

YaelDillies:
If you can make your code action insert arbitrary code snippets, then you could cargo-cult my `deprecated alias` snippet (which is in `.vscode/something.snippet`

adomani:
Oh, I see, not Lean-side, but VSCode-side!

I won't have time for it today, but I'll try to do it soon!

adomani:
Yaël, I added the `since` syntax.  Currently, it does not optionally take a date: it automatically uses today's date.  This will be problematic for the test file, but I will add the optional date argument tomorrow, probably

YaelDillies:
I don't think there's a big use case for the optional date? Maybe simply don't check the precise date in the tests.

adomani:
Oh, `#guard_msgs` has some "non-exact match", right?  I'll look at what is available and will use that for the test.

adomani:
I looked a bit into the `#guard_msgs` options, but I could not find a way of making it ignore the exact date.  Thus, the command takes an optional `String` argument, but the expectation is that this will not be used outside of the test file.

Mathlib/Tactic/DeprecateMe.lean
```diff
@@ -31,17 +31,19 @@ where `d₁ d₂ ... dₙ` are the non-blacklisted declarations (auto)generated
...
 -/
 
 namespace Mathlib.Tactic.DeprecateMe
 
 open Lean Elab Term Command
 
-/-- Produce the syntax for the command `@[deprecated] alias n := id`. -/
-def mkDeprecationStx (id : TSyntax `ident) (n : Name) : CommandElabM (TSyntax `command) := do
-  let dat ← IO.Process.run { cmd := "date", args := #["-I"] }
+/-- Produce the syntax for the command `@[deprecated( since := "YYYY-MM-DD")] alias n := id`. -/
```
YaelDillies, line 41:
```suggestion
/-- Produce the syntax for the command `@[deprecated (since := "YYYY-MM-DD")] alias n := id`. -/
```

Mathlib/Tactic/DeprecateMe.lean
```diff
@@ -0,0 +1,141 @@
...
+where `id₁ id₂ ... idₙ` is a sequence of identifiers produces the `Try this` suggestion:
+```lean
+theorem easy : True := .intro
+
+@[deprecated (since := "YYYY-MM-DD")] alias d₁ := id₁
+
+@[deprecated (since := "YYYY-MM-DD")] alias d₂ := id₂
+...
+
+@[deprecated (since := "YYYY-MM-DD")] alias dₙ := idₙ
```
YaelDillies, line 27:
```suggestion
@[deprecated (since := "YYYY-MM-DD")] alias id₁ := d₁

@[deprecated (since := "YYYY-MM-DD")] alias id₂ := d₂
...

@[deprecated (since := "YYYY-MM-DD")] alias idₙ := dₙ
```
right? Consider using less obscure names than `d` and `id`, maybe `alias deprₙ := declₙ`?

adomani: I think that this is correct: the user provides the new names for all the declarations above the declaration-to-be-deprecated.  The workflow should be
* the script replaces the old name by the new name in the original declaration;
* the new declarations get automatically generated with their autogenerated names (more on this later);
* the script retrieves the old autogenerated names and creates the aliases for the (now-non-existing) old autogenerated names to be deprecated in favour of the new ones that the user provided.

The part that I am not too happy about is that you have to provide the new autogenerated names by hand, rather than autogenerating them.  The issue is not actually retrieving those names, but rather matching the new ones with what the old ones were.  This is why you would have to provide the `to_additive` name of a `to_additivized` declaration.  In the specific case of `to_additive` the matching would be easy, as there is only one non-visible declaration, but with `simps` the matching is trickier.

adomani: Maybe these names make it clearer:
```lean
/--
info: * Pairings:
#[(`new_name_mul, mul_easy_deprecated), (`new_name_add, add_easy_deprecated)]

Try this:
@[to_additive]
theorem new_name_mul : True :=
  .intro

@[deprecated (since := "2024-06-11")]
alias mul_easy_deprecated := new_name_mul

@[deprecated (since := "2024-06-11")]
alias add_easy_deprecated := new_name_add
-/
#guard_msgs in
deprecate to new_name_mul new_name_add
@[to_additive] theorem mul_easy_deprecated : True := .intro
```

adomani: Anyway, I agree with the suggestion of using meaningful names for the variable and I am going to update them!

YaelDillies:
Can you also add your new file to `Tactic.Common`?

adomani:
I just added the code with the `Try this` suggestion to `Tactic.Common`.  I'll wait for the build to finish to get `lake exe shake --update` to fix its file.

YaelDillies (approved):
Okay, let's try this in production!

maintainer merge

jcommelin:
Thanks :tada:

bors merge

PR #13517, opened 2024-06-04
[Merged by Bors] - feat: add perfectly normal (T_6) topological property

StevenClontz:
Introduces the final whole-number separation axiom: [perfectly normal](https://topology.pi-base.org/properties/P000015) / [$T_6$](https://topology.pi-base.org/properties/P000067). Includes a proof that perfectly normal implies completely normal (resp. $T_6$ implies $T_5$). Future work: show all pseudometrizable spaces are perfectly normal (resp. all metrizable spaces are $T_6$).

---

- [x] depends on: #13176
- [x] depends on: #13694

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13694, opened 2024-06-10
[Merged by Bors] - chore: reverse gdelta/separation dependency

StevenClontz:
This reversal will allow for the following definition for a [perfectly normal](https://topology.pi-base.org/properties/P15) space (with $T_1$ this is [T_6](https://topology.pi-base.org/properties/P000067), the last "whole number" separation axiom missing from mathlib), proposed at #13517.

```lean
/-- A topological space `X` is a *perfectly normal space* provided it is normal and
closed sets are Gδ. -/
class PerfectlyNormalSpace (X : Type u) [TopologicalSpace X] extends NormalSpace X : Prop where
    closed_gdelta : ∀ ⦃h : Set X⦄, IsClosed h → IsGδ h
```

---


Mathlib/Topology/GDelta.lean
```diff
@@ -4,7 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Sébastien Gouëzel, Yury Kudryashov
 -/
 import Mathlib.Topology.UniformSpace.Basic
-import Mathlib.Topology.Separation
+-- import Mathlib.Topology.Separation
```
StevenClontz, line 7:
```suggestion
```

PatrickMassot:
bors merge

PR #12293, opened 2024-04-20
[Merged by Bors] - feat(Mathlib.RingTheory.TensorProduct.MvPolynomial) :  tensor product of a (multivariate) polynomial ring

AntoineChambert-Loir:
Let `Semiring R`, `Algebra R S` and `Module R N`.

* `MvPolynomial.rTensor` gives the linear equivalence
  `MvPolynomial σ S ⊗[R] N ≃ₗ[R] (σ →₀ ℕ) →₀ (S ⊗[R] N)` characterized,
  for `p : MvPolynomial σ S`, `n : N` and `d : σ →₀ ℕ`, by
  `rTensor (p ⊗ₜ[R] n) d = (coeff d p) ⊗ₜ[R] n`
* `MvPolynomial.scalarRTensor` gives the linear equivalence
  `MvPolynomial σ R ⊗[R] N ≃ₗ[R] (σ →₀ ℕ) →₀ N`
  such that `MvPolynomial.scalarRTensor (p ⊗ₜ[R] n) d = coeff d p • n`
  for `p : MvPolynomial σ R`, `n : N` and `d : σ →₀ ℕ`, by

* `MvPolynomial.rTensorAlgHom`, the algebra morphism from the tensor product
  of a polynomial algebra by an algebra to a polynomial algebra
* `MvPolynomial.rTensorAlgEquiv`, `MvPolynomial.scalarRTensorAlgEquiv`,
  the tensor product of a polynomial algebra by an algebra
  is algebraically equivalent to a polynomial algebra

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13696, opened 2024-06-10
[Merged by Bors] - feat(RingTheory): finite type and finite presentation are stable under base change

chrisflav:
---

### DIFF
--- a/Mathlib/RingTheory/FinitePresentation.lean
+++ b/Mathlib/RingTheory/FinitePresentation.lean
@@ -6,6 +6,10 @@ Authors: Johan Commelin
 import Mathlib.RingTheory.FiniteType
 import Mathlib.RingTheory.MvPolynomial.Tower
 import Mathlib.RingTheory.Ideal.QuotientOperations
+import Mathlib.RingTheory.RingHomProperties
+import Mathlib.RingTheory.TensorProduct.Basic
+import Mathlib.RingTheory.TensorProduct.MvPolynomial
+import Mathlib.LinearAlgebra.TensorProduct.RightExactness
 
 #align_import ring_theory.finite_presentation from "leanprover-community/mathlib"@"da420a8c6dd5bdfb85c4ced85c34388f633bc6ff"
 
@@ -29,7 +33,7 @@ set_option autoImplicit true
 
 open Function (Surjective)
 
-open Polynomial
+open Polynomial TensorProduct
 
 section ModuleAndAlgebra
 
@@ -401,6 +405,26 @@ theorem ker_fG_of_surjective (f : A →ₐ[R] B) (hf : Function.Surjective f)
   simp_rw [RingHom.ker_eq_comap_bot, AlgHom.toRingHom_eq_coe, AlgHom.comp_toRingHom]
   rw [← Ideal.comap_comap, Ideal.map_comap_of_surjective (g : MvPolynomial (Fin n) R →+* A) hg]
 #align algebra.finite_presentation.ker_fg_of_surjective Algebra.FinitePresentation.ker_fG_of_surjective
+
+theorem baseChange (hfa : Algebra.FinitePresentation R A) :
+    Algebra.FinitePresentation B (B ⊗[R] A) := by
+  obtain ⟨n, f, hsurj, hfg⟩ := hfa
+  let g : B ⊗[R] MvPolynomial (Fin n) R →ₐ[B] B ⊗[R] A :=
+    Algebra.TensorProduct.map (AlgHom.id B B) f
+  have hgsurj : Function.Surjective g := Algebra.FiniteType.baseChangeAux_surj B hsurj
+  have hker_eq : RingHom.ker g = Ideal.map Algebra.TensorProduct.includeRight (RingHom.ker f) :=
+    Algebra.TensorProduct.lTensor_ker f hsurj
+  have hfgg : Ideal.FG (RingHom.ker g) := by
+    rw [hker_eq]
+    exact Ideal.FG.map hfg _
+  let g' : MvPolynomial (Fin n) B →ₐ[B] B ⊗[R] A :=
+    AlgHom.comp g (MvPolynomial.algebraTensorAlgEquiv B).symm.toAlgHom
+  refine ⟨n, g', ?_, Ideal.fg_ker_comp _ _ ?_ hfgg ?_⟩
+  · simp_all [g, g']
+  · show Ideal.FG (RingHom.ker (AlgEquiv.symm (MvPolynomial.algebraTensorAlgEquiv B)))
+    simp only [RingHom.ker_equiv]
+    exact Submodule.fg_bot
+  · simpa using EquivLike.surjective _
 
 end FinitePresentation
 
--- a/Mathlib/RingTheory/FinitePresentation.lean
+++ b/Mathlib/RingTheory/FinitePresentation.lean
@@ -6,9 +6,6 @@ Authors: Johan Commelin
 import Mathlib.RingTheory.FiniteType
 import Mathlib.RingTheory.MvPolynomial.Tower
 import Mathlib.RingTheory.Ideal.QuotientOperations
-import Mathlib.RingTheory.TensorProduct.Basic
-import Mathlib.RingTheory.TensorProduct.MvPolynomial
-import Mathlib.LinearAlgebra.TensorProduct.RightExactness
 
 #align_import ring_theory.finite_presentation from "leanprover-community/mathlib"@"da420a8c6dd5bdfb85c4ced85c34388f633bc6ff"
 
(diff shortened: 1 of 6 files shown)


Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
```diff
@@ -186,6 +186,31 @@ noncomputable def scalarRTensorAlgEquiv :
...
+    A ⊗[R] MvPolynomial σ R ≃ₐ[A] MvPolynomial σ A := AlgEquiv.ofAlgHom
+  (Algebra.TensorProduct.lift
+    (Algebra.ofId A (MvPolynomial σ A))
+    (MvPolynomial.mapAlgHom <| Algebra.ofId R A) (fun _ _ ↦ Commute.all _ _))
+  (aeval (fun s ↦ 1 ⊗ₜ X s))
+  (by ext s; simp)
+  (by ext s; simp)
+
+@[simp]
+lemma algebraTensorAlgEquiv_apply (a : A) (s : σ) :
```
erdOne, line 203: This should be called `algebraTensorAlgEquiv_tmul_X`, and I think we should have `algebraTensorAlgEquiv_tmul` as the simp lemma (`algebraTensorAlgEquiv (R := R) A (a ⊗ₜ p) = a • p.map (algebraMap R A)`)

Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
```diff
@@ -186,6 +186,31 @@ noncomputable def scalarRTensorAlgEquiv :
...
+    algebraTensorAlgEquiv (R := R) A (a ⊗ₜ X s) = a • X s := by
+  simp [algebraTensorAlgEquiv]
+  rw [Algebra.smul_def]
+  rfl
+
+@[simp]
+lemma algebraTensorAlgEquiv_symm_apply (s : σ) :
```
erdOne, line 210: Similarly this should be called `algebraTensorAlgEquiv_symm_X`. Should there be a `algebraTensorAlgEquiv_symm_monomial`? Would it be useful?

Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
chrisflav: Should I keep the `tmul_X` version? With the new lemma it is proven by `simp`.

erdOne: Hmmm maybe not. Sometimes it still useful to restate lemma provable by simp for rewriting, but this lemma is merely two rewrites.

Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
chrisflav: I added `algebraTensorAlgEquiv_symm_monomial`.

Mathlib/RingTheory/FinitePresentation.lean
```diff
@@ -402,6 +407,26 @@ theorem ker_fG_of_surjective (f : A →ₐ[R] B) (hf : Function.Surjective f)
   rw [← Ideal.comap_comap, Ideal.map_comap_of_surjective (g : MvPolynomial (Fin n) R →+* A) hg]
 #align algebra.finite_presentation.ker_fg_of_surjective Algebra.FinitePresentation.ker_fG_of_surjective
 
+theorem baseChange (hfa : Algebra.FinitePresentation R A) :
```
erdOne, line 410: This could be an instance.

Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
```diff
@@ -186,6 +186,43 @@ noncomputable def scalarRTensorAlgEquiv :
...
+  (Algebra.TensorProduct.lift
+    (Algebra.ofId A (MvPolynomial σ A))
+    (MvPolynomial.mapAlgHom <| Algebra.ofId R A) (fun _ _ ↦ Commute.all _ _))
+  (aeval (fun s ↦ 1 ⊗ₜ X s))
+  (by ext s; simp)
+  (by ext s; simp)
+
+@[simp]
+lemma algebraTensorAlgEquiv_tmul (a : A) (p : MvPolynomial σ R) :
+    algebraTensorAlgEquiv (R := R) A (a ⊗ₜ p) = a • MvPolynomial.map (algebraMap R A) p := by
```
erdOne, line 204: Is the `(R := R)` necessary here? If it is, I don't see how it could be implicit elsewhere and I'd suggest we make it explicit in the first place.

chrisflav: In this example it is indeed not necessary, in the `symm_X` lemma below it is necessary though. Should I make it explicit anyway?

erdOne: In that case I think we should make it explicit.

chrisflav: Done.

erdOne:
Thanks!
maintainer merge

fpvandoorn:
There are significantly more files imported in some files. Does it make sense to put the new material in a separate file?

chrisflav:
> There are significantly more files imported in some files. Does it make sense to put the new material in a separate file?

I made a new file.

Mathlib/RingTheory/TensorProduct/MvPolynomial.lean
```diff
@@ -186,6 +186,44 @@ noncomputable def scalarRTensorAlgEquiv :
     MvPolynomial σ R ⊗[R] N ≃ₐ[R] MvPolynomial σ N :=
   rTensorAlgEquiv.trans (mapAlgEquiv σ (Algebra.TensorProduct.lid R N))
 
+variable (R)
+variable (A : Type*) [CommRing A] [Algebra R A]
+
+/-- Tensoring `MvPolynomial σ R` on the left by an `R`-algebra `A` is algebraically
+equivalent to `M̀vPolynomial σ A`. -/
+noncomputable def algebraTensorAlgEquiv :
+    A ⊗[R] MvPolynomial σ R ≃ₐ[A] MvPolynomial σ A := AlgEquiv.ofAlgHom
```
alreadydone, line 195: Apparently this already [exists](https://github.com/leanprover-community/mathlib4/pull/12293/files#diff-fcd41314736c9eb63afb8cedfaae8800b725b0f3003d437246bf1c6caf4b3f81R185-R186) thanks to @AntoineChambert-Loir

chrisflav: I thought so as well, but I don't think that is entirely correct. First, in the equiv you cited `N` is just a module, so the equiv is only `R`-linear but I need `A`-linearity. Secondly, even if you ask `N` to be an `R`-algebra, the LHS in the equiv you cited is "the wrong way" round for my application: I think @AntoineChambert-Loir had in mind to enhance it to something `MvPolynomial σ R`-linear for which `MvPolynomial σ R` has to be on the left of the `⊗[R]`. But I need it to be `A`-linear, so `A` has to be on the left of `⊗[R]`.

chrisflav: And while I can probably massage `scalarRTensorAlgEquiv` into the thing I want by using `Algebra.TensorProduct.comm` and promoting it to an `A`-linear map, I think this would be longer and would have worse definitional properties than the direct definition.

alreadydone: `N` is a CommSemiring there (that's a prerequisite for MvPolynomial), but yes I neglected the A-linearity, which requires A on the left. Can you please generalize from Ring to Semiring though?

chrisflav: Done

alreadydone: Thanks, LGTM!

Mathlib/RingTheory/RingHom/FiniteType.lean
```diff
@@ -100,4 +101,14 @@ theorem finiteType_respectsIso : RingHom.RespectsIso @RingHom.FiniteType :=
   RingHom.finiteType_is_local.respectsIso
 #align ring_hom.finite_type_respects_iso RingHom.finiteType_respectsIso
 
+theorem finiteType_stableUnderBaseChange : StableUnderBaseChange @FiniteType := by
```
alreadydone, line 104: Why aren't you adding a `StableUnderBaseChange @FinitePresentation` theorem?

alreadydone (approved):

Mathlib/RingTheory/RingHom/FiniteType.lean
chrisflav: Because the file `RingHom/FinitePresentation.lean` does not yet exist. I am happy to follow-up on that though with another PR.

fpvandoorn:
LGTM

bors merge

chrisflav:
Thanks for the reviews!

PR #13697, opened 2024-06-10
[Merged by Bors] - feat :  define `ContinuousFunctionalCalculus` Instance for Hermitian Matrices over an RCLike Field

JonBannon:
This file contains the requisite lemmas needed to define a `ContinuousFunctionalCalculus` for Hermitian Matrices over an RCLike Field.

- [x] depends on: #13729
- [x] depends on: #13765
- [x] depends on : #13837
- [x] depends on : #13838

Co-Authored by @j-loreaux

---


Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,302 @@
...
+                    instance.
+- instContinuousFunctionalCalculus : Instance of the Continuous functional Calculus for a hermitian
+                                     matrix A over 𝕜.
+
+## Tags
+
+spectral theorem, diagonalization theorem, continuous functional calculus
+-/
+
+section ConjugateUnits
```
j-loreaux, line 30: This section, and hopefully the unitary one too, should go in `Algebra.Algebra.Spectrum`, preferably even in another PR.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,302 @@
...
+
+/-- Unitary conjugation preserves the spectrum, star on right. -/
+@[simp]
+lemma spectrum.unitary_conjugate' {a : A} {u : unitary A} :
+    spectrum R ((star u : A) * a * u) = spectrum R a := by
+  simpa using spectrum.unitary_conjugate (u := star u)
+
+end UnitaryConjugate
+
+section FiniteSpectrum
```
j-loreaux, line 75: This section as should go in the same file as the theorems used to prove these, preferably in another PR.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,302 @@
...
+theorem Matrix.finite_spectrum (A : Matrix n n R) : Set.Finite (spectrum R A) := by
+  rw [← AlgEquiv.spectrum_eq (Matrix.toLinAlgEquiv <| Pi.basisFun R n) A]
+  exact Module.End.finite_spectrum _
+
+instance Matrix.instFiniteSpectrum (A : Matrix n n R) : Finite (spectrum R A) :=
+  Set.finite_coe_iff.mpr (Matrix.finite_spectrum A)
+
+end FiniteSpectrum
+
+section SpectrumDiagonal
```
j-loreaux, line 99: This section should go in another file, although I'm not sure which is best off the top of my head. This should also constitute another PR.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,302 @@
...
+variable {𝕜 : Type*} [RCLike 𝕜] {n : Type*} [Fintype n]
+
+namespace IsHermitian
+
+variable [DecidableEq n]
+
+variable {A : Matrix n n 𝕜} (hA : IsHermitian A)
+
+/--Eigenvalues of a Hermitian Matrix, coerced, belong to the spectrum of the assoc.toEuclideanLin -/
+theorem eigenvalue_mem_toEuclideanLin_spectrum_RCLike (i : n) :
```
j-loreaux, line 160: Some of these spectrum/eigenvalue results can probably go in the `Matrix.IsHermitian` file.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,302 @@
...
+/--Eigenvalues of a hermitian matrix A are in the ℝ spectrum of A. -/
+theorem eigenvalue_mem_real : ∀ (i : n), (hA.eigenvalues) i ∈ spectrum ℝ A := by
+  intro i
+  apply spectrum.of_algebraMap_mem (S := 𝕜) (R := ℝ) (A := Matrix n n 𝕜)
+  rw [← spec_toEuclideanLin_eq_spec]
+  apply hA.eigenvalue_mem_toEuclideanLin_spectrum_RCLike i
+
+/--Definition of the StarAlgHom for the continuous functional calculus of a Hermitian matrix. -/
+@[simps]
+noncomputable def φ : StarAlgHom ℝ C(spectrum ℝ A, ℝ) (Matrix n n 𝕜) where
```
j-loreaux, line 188: This shouldn't be called `phi` it either needs to be inclined, or given a more informative name. Also, have you considered defining it on bare functions instead of continuous ones?

j-loreaux (commented):
After you merge master you can come back to my comments here.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
JonBannon: I was *wondering* if you were going to suggest this. Cool!

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
j-loreaux: I doubt it will clean up anything, but the advantage is that in the future we can write a lemma of the form `cfc f a = φ f` for *any* `f`, so long as `a` is selfadjoint. This is nice because the first argument to `cfc` is a bare function.

JonBannon: What do you mean by 'inclined' in this context? (I am inclined to give this a more informative name...)

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
JonBannon: Should I do this as part of the main PR? If so, I'd like to know why you think so, i.e. why is this different than the above sections you suggested that I PR separately?

JonBannon: I was only able to move one of these, as the others require `spectrum` so would need an import. What do you suggest?

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
JonBannon: I've changed the name to `cfc`. Now I have to look into changing this to bare functions...

JonBannon: It seems that if I replace the type of continuous functions to the type of bare functions, the instance declaration has several places where it expects a map on continuous functions.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,162 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jon Bannon, Jireh Loreaux
+-/
+
+import Mathlib.LinearAlgebra.Matrix.Spectrum
+import Mathlib.LinearAlgebra.Eigenspace.Matrix
+import Mathlib.Analysis.NormedSpace.Star.ContinuousFunctionalCalculus
+import Mathlib.Topology.ContinuousFunction.UniqueCFC
+import Mathlib.Analysis.NormedSpace.Star.Matrix
+import Mathlib.Algebra.Star.Unitary
```
j-loreaux, line 12: Surely not all of these imports are necessary?

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,162 @@
...
+
+/-!
+# Continuous Functional Calculus for Hermitian Matrices
+
+This file defines an instance of the continuous functional calculus for Hermitian matrices over an
+RCLike field 𝕜.
```
j-loreaux, line 18:
```suggestion
`RCLike` field `𝕜`.
```

j-loreaux (commented):
Can you address these and fix the build? Otherwise LGTM.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
JonBannon: You are right! I winnowed this down to four.

j-loreaux:
@JonBannon build still falling

JonBannon:
Weird let me have a look.

JonBannon:
Looks like I have to put HermitianFunctionalCalculus.lean into the Mathlib
file...because we created a new file. I'll bet that's what is happening.

J

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,160 @@
...
+  rw [← spectrum.preimage_algebraMap 𝕜]
+  exact A.finite_spectrum.preimage (NoZeroSMulDivisors.algebraMap_injective ℝ 𝕜).injOn
+
+instance : Finite (spectrum ℝ A) := A.finite_real_spectrum
+
+namespace IsHermitian
+
+variable (hA : IsHermitian A)
+
+/-- The ℝ-spectrum of a Hermitian Matrix over RCLike field is the range of the eigenvalue function-/
```
j-loreaux, line 48:
```suggestion
/-- The `ℝ`-spectrum of a Hermitian matrix over `RCLike` field is the range of the eigenvalue
function -/
```

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,160 @@
...
+theorem eigenvalues_eq_spectrum_real {a : Matrix n n 𝕜} (ha : IsHermitian a) :
+    (spectrum ℝ a) = Set.range (ha.eigenvalues) := by
+  ext x
+  conv_lhs => rw [ha.spectral_theorem, unitary.spectrum.unitary_conjugate,
+  ← spectrum.algebraMap_mem_iff 𝕜, spectrum_diagonal, RCLike.algebraMap_eq_ofReal]
+  simp
+
+/--Definition of the StarAlgHom for the functional calculus of a Hermitian matrix. -/
```
j-loreaux, line 56:
```suggestion
/-- The star algebra homomorphism underlying the instance of the continuous functional
calculus of a Hermitian matrix. This is an auxiliary definition and is not intended
for use outside of this file. -/
```

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,160 @@
...
+## Main Results
+
+- `Matrix.IsHermitian.cfc` : Realization of the functional calculus for a Hermitian matrix
+  as the triple product `U * diagonal (RCLike.ofReal ∘ f ∘ hA.eigenvalues) * star U` with
+  `U = eigenvectorUnitary hA`.
+
+- `cfc_eq` : Proof that the above agrees with the continuous functional calculus.
+
+- `instContinuousFunctionalCalculus` : Instance of the continuous functional calculus for a
+  Hermitian matrix `A` over `𝕜`.
```
j-loreaux, line 27:
```suggestion
- `Matrix.IsHermitian.instContinuousFunctionalCalculus` : Instance of the continuous functional
  calculus for a Hermitian matrix `A` over `𝕜`.
```

j-loreaux (commented):
A few more style and documentation nits.

j-loreaux:
@dupuisf I'll need you (or another reviewer, but you make the most sense I think) to take over here, because I've contributed too much to even feel comfortable putting it on the maintainer merge-queue.

Mathlib/LinearAlgebra/Matrix/HermitianFunctionalCalculus.lean
```diff
@@ -0,0 +1,163 @@
...
+instance : Finite (spectrum ℝ A) := A.finite_real_spectrum
+
+namespace IsHermitian
+
+variable (hA : IsHermitian A)
+
+/-- The `ℝ`-spectrum of a Hermitian matrix over `RCLike` field is the range of the eigenvalue
+function -/
+theorem eigenvalues_eq_spectrum_real {a : Matrix n n 𝕜} (ha : IsHermitian a) :
+    (spectrum ℝ a) = Set.range (ha.eigenvalues) := by
```
dupuisf, line 51:
```suggestion
theorem eigenvalues_eq_spectrum_real {a : Matrix n n 𝕜} (ha : IsHermitian a) :
    spectrum ℝ a = Set.range (ha.eigenvalues) := by
```

dupuisf (commented):
Looks good!

bors d+

PR #13699, opened 2024-06-10
[Merged by Bors] - perf: speed up integrable_rpow_neg_one_add_norm_sq

Ruben-VandeVelde:
This goes from about 20 seconds to instantaneous locally.

---

### DIFF
--- a/Mathlib/Analysis/SpecialFunctions/JapaneseBracket.lean
+++ b/Mathlib/Analysis/SpecialFunctions/JapaneseBracket.lean
@@ -154,7 +154,8 @@ theorem integrable_rpow_neg_one_add_norm_sq {r : ℝ} (hnr : (finrank ℝ E : �
   have hr : 0 < r := lt_of_le_of_lt (finrank ℝ E).cast_nonneg hnr
   refine ((integrable_one_add_norm hnr).const_mul <| (2 : ℝ) ^ (r / 2)).mono'
     ?_ (eventually_of_forall fun x => ?_)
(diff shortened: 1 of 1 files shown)


grunweg (commented):
Wow, great find! 20 seconds is really slow...

Can you add a comment saying that this was `measurability`, but has been expanded for performance reasons? (For extra bonus points: is there an issue about this/can you file one? Are the other such lemmas which are slow?)

I'd approve with that - consider this another `maintainer delegate` (i.e., feel free to `maintainer merge on my behalf if you feel you have addressed this - or ask on zulip if you'd like another opinion.)

Ruben-VandeVelde:
Good point, I was going to add a comment but forgot. As far as I understand it, `fun_prop` should replace `measurability` at some point (but it doesn't work yet for this proof), so I don't think anyone will look at an issue.

maintainer merge

grunweg:
Out of curiosity: how did you find this one? Do you know other `measurability` invocations which are really slow?
(If I have some free Lean time and like popping bubble wrap, fixing these could be a nice thing to do :-))

Ruben-VandeVelde:
Purely by accident - I had put a `#minimize_imports` at the end of the file and was waiting for it to run - and waiting - and waiting...

fpvandoorn:
It might also be interesting to see how `fun_prop` does.

bors merge

grunweg:
> Purely by accident - I had put a `#minimize_imports` at the end of the file and was waiting for it to run - and waiting - and waiting...

Oh wow, that was a happy accident! I'll keep an item on my long list of things to investigate.
(Current plan: `set_option profiler true` globally, build mathlib and see which tactic calls stick out. I don't have time to look at this list right now, though.)

PR #13701, opened 2024-06-10
[Merged by Bors] - feat(AlgebraicGeometry): expand api on topological scheme morphism properties

chrisflav:
Deduce properties of `MorphismProperty.topologically P` from the respective properties stated in terms of topological spaces.

---

As far as I can tell, the naming in this file deviates from the general mathlib conventions, so I tried to follow the local pattern. But maybe we should do some renames here?


erdOne:
I think they are porting errors and shouldn't be followed. I'd be great if you could fix them.

chrisflav:
> I think they are porting errors and shouldn't be followed. It'd be great if you could fix them.

Separate PR I guess? Maybe we can split this file at the same time, I have more API lemmas to add and this file is already getting big.

erdOne:
Sure. The lemmas added should follow the correct naming convention though.

erdOne:
Thanks!
maintainer merge

joelriou:
Thanks!

bors merge

PR #13703, opened 2024-06-10
feat: topology on ring of finite adeles.

kbuzzard:
Use `SubmodulesRingBasis` to put a topology on the finite adeles.

---

- [x] depends on: #13705
- [x] depends on: #13853


kbuzzard:
bleurgh, git nightmare. Making a new PR because it's easier than learning how to fix this one.

PR #13705, opened 2024-06-10
[Merged by Bors] - feat: the finite adele ring is an algebra over the finite integral adeles.

kbuzzard:
---

We change the definition of the finite adeles to a subtype, and rework the original construction to be a subalgebra with this underlying subtype. We then show that the finite adeles are an algebra over the finite integral adeles, in an attempt to demonstrate that this set-up works fine.

### DIFF
--- a/Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
+++ b/Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
@@ -313,21 +313,24 @@ open ProdAdicCompletions.IsFiniteAdele
 
 /-- The finite adèle ring of `R` is the restricted product over all maximal ideals `v` of `R`
 of `adicCompletion` with respect to `adicCompletionIntegers`. -/
-def FiniteAdeleRing : Type _ := (
-  { carrier := {x : K_hat R K | x.IsFiniteAdele}
-    mul_mem' := mul
-    one_mem' := one
-    add_mem' := add
-    zero_mem' := zero
-    algebraMap_mem' := algebraMap'
-  } : Subalgebra K (K_hat R K))
+def FiniteAdeleRing : Type _ := {x : K_hat R K // x.IsFiniteAdele}
 #align dedekind_domain.finite_adele_ring DedekindDomain.FiniteAdeleRing
 
 namespace FiniteAdeleRing
 
-instance : CommRing (FiniteAdeleRing R K) := Subalgebra.toCommRing _
+def subalgebra : Subalgebra K (K_hat R K) where
+  carrier := {x : K_hat R K | x.IsFiniteAdele}
+  mul_mem' := mul
+  one_mem' := one
+  add_mem' := add
+  zero_mem' := zero
+  algebraMap_mem' := algebraMap'
 
-instance : Algebra K (FiniteAdeleRing R K) := Subalgebra.algebra _
+instance : CommRing (FiniteAdeleRing R K) :=
+  Subalgebra.toCommRing (subalgebra R K)
+
+instance : Algebra K (FiniteAdeleRing R K) :=
+  Subalgebra.algebra (subalgebra R K)
 
 instance : Coe (FiniteAdeleRing R K) (K_hat R K) where
   coe := fun x ↦ x.1
@@ -337,25 +340,14 @@ lemma ext {a₁ a₂ : FiniteAdeleRing R K} (h : (a₁ : K_hat R K) = a₂) : a�
   Subtype.ext h
 
 @[simp]
-lemma mem_FiniteAdeleRing (x : K_hat R K) : x ∈ (
-    { carrier := {x : K_hat R K | x.IsFiniteAdele}
-      mul_mem' := mul
-      one_mem' := one
-      add_mem' := add
-      zero_mem' := zero
-      algebraMap_mem' := algebraMap'
-    } : Subalgebra K (K_hat R K)) ↔ {v | x v ∉ adicCompletionIntegers K v}.Finite := Iff.rfl
+lemma _root_.isFiniteAdele_iff (x : K_hat R K) :
+  x.IsFiniteAdele ↔ {v | x v ∉ adicCompletionIntegers K v}.Finite := Iff.rfl
 
 /-- The finite adele ring is an algebra over the finite integral adeles. -/
 instance : Algebra (R_hat R K) (FiniteAdeleRing R K) where
-  smul rhat fadele := ⟨fun v ↦ rhat v * fadele.1 v, by
-    have this := fadele.2
-    rw [mem_FiniteAdeleRing] at this ⊢
-    apply Finite.subset this (fun v hv ↦ ?_)
-    rw [mem_setOf_eq, mem_adicCompletionIntegers] at hv ⊢
-    contrapose! hv
-    rw [map_mul]
-    exact mul_le_one₀ (rhat v).2 hv
+  smul rhat fadele := ⟨fun v ↦ rhat v * fadele.1 v, Finite.subset fadele.2 <| fun v hv ↦ by
+    simp only [mem_adicCompletionIntegers, mem_compl_iff, mem_setOf_eq, map_mul] at hv ⊢
+    exact mt (mul_le_one₀ (rhat v).2) hv
     ⟩
   toFun r := ⟨r, by simp_all⟩
   map_one' := by ext; rfl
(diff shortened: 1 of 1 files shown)


Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
```diff
@@ -336,6 +336,34 @@ instance : Coe (FiniteAdeleRing R K) (K_hat R K) where
...
 
+@[simp]
+lemma mem_FiniteAdeleRing (x : K_hat R K) : x ∈ (
+    { carrier := {x : K_hat R K | x.IsFiniteAdele}
+      mul_mem' := mul
+      one_mem' := one
+      add_mem' := add
+      zero_mem' := zero
+      algebraMap_mem' := algebraMap'
+    } : Subalgebra K (K_hat R K)) ↔ {v | x v ∉ adicCompletionIntegers K v}.Finite := Iff.rfl
```
erdOne, line 347: I think the subalgebra should be separated into a new def, and `FiniteAdeleRing` could potentially be defined as `{ x : K_hat R K // x.IsFiniteAdele }` and only view it as a `subalgebra` (since they are defeq) when needed.

alreadydone: I'd just remove the coercion in FiniteAdeleRing and change its type to Subalgebra.

kbuzzard: The problem with that is that `K_hat R K`, the thing it's a subalgebra of, is a technical intermediate thing, but `FiniteAdeleRing R K` is the fundamental definition and I think it deserves to be a type. There's a problem with having coerced terms as fundamental objects, and that's that typeclass inference goes crazy trying to deal with it. Here's the issue. We need to let `FiniteAdeleRing` be a type because it has, for example, a topology on it (and it's not the subspace topology coming from K_hat). So if it's a subalgebra then the code will be full of `↥FiniteAdeleRing`, and when typeclass sees that it wants e.g. a topological space structure on this, or a coercion from this to somewhere else, it is quick to start looking looking at the auxiliary object `K_hat` for answers. This leads to timeouts (we saw it with `↥(𝓞 K)` in the theory of number fields) and this is exactly why we changed the definition of ring of integers from a term to a type.

kbuzzard: I'm going to try erdOne's suggestion.

Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
```diff
@@ -336,6 +339,24 @@ instance : Coe (FiniteAdeleRing R K) (K_hat R K) where
 lemma ext {a₁ a₂ : FiniteAdeleRing R K} (h : (a₁ : K_hat R K) = a₂) : a₁ = a₂ :=
   Subtype.ext h
 
+@[simp]
+lemma _root_.isFiniteAdele_iff (x : K_hat R K) :
+  x.IsFiniteAdele ↔ {v | x v ∉ adicCompletionIntegers K v}.Finite := Iff.rfl
```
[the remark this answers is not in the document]
erdOne: I think this lemma should go near the definition of `IsFiniteAdele`.

Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
```diff
@@ -312,22 +316,27 @@ end ProdAdicCompletions
...
-def FiniteAdeleRing : Type _ := (
-  { carrier := {x : K_hat R K | x.IsFiniteAdele}
-    mul_mem' := mul
-    one_mem' := one
-    add_mem' := add
-    zero_mem' := zero
-    algebraMap_mem' := algebraMap'
-  } : Subalgebra K (K_hat R K))
+of `adicCompletion`, with respect to `adicCompletionIntegers`. -/
+def FiniteAdeleRing : Type _ := {x : K_hat R K // x.IsFiniteAdele}
```
mariainesdff, line 320: Why do you prefer to define this as a subtype, rather than a subalgebra?

kbuzzard: If you look at the original version of this PR [here](https://github.com/leanprover-community/mathlib4/pull/13705/commits/6a5c0f781c81f21d237dc0fed160ef38be1ab868#diff-e3f60199ddcf6fe696a3eed7602eded09cf575d804ab32af60be92b53c2cabb6R339-R347) you'll see what happens if you make it a coerced subalgebra; the `simp` lemma you need to get things back under control is horrible. Now we make it as a subtype and get the algebra structure via the subalgebra.

erdOne: I think María Inés was asking why this isn't `def FiniteAdeleRing : Subalgebra K (K_hat R K)`. To quote Kevin from above:
> The problem with that is that `K_hat R K`, the thing it's a subalgebra of, is a technical intermediate thing, but `FiniteAdeleRing R K` is the fundamental definition and I think it deserves to be a type. There's a problem with having coerced terms as fundamental objects, and that's that typeclass inference goes crazy trying to deal with it. Here's the issue. We need to let `FiniteAdeleRing` be a type because it has, for example, a topology on it (and it's not the subspace topology coming from K_hat). So if it's a subalgebra then the code will be full of `↥FiniteAdeleRing`, and when typeclass sees that it wants e.g. a topological space structure on this, or a coercion from this to somewhere else, it is quick to start looking looking at the auxiliary object `K_hat` for answers. This leads to timeouts (we saw it with `↥(𝓞 K)` in the theory of number fields and people were complaining about timeouts for about a year) and this is exactly why we changed the definition of ring of integers from a term to a type (and it solved the problems!)

We should add a library note for this.

mariainesdff:
mantainer merge

Ruben-VandeVelde (approved):
maintainer merge

Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
```diff
@@ -312,22 +316,27 @@ end ProdAdicCompletions
...
 #align dedekind_domain.finite_adele_ring DedekindDomain.FiniteAdeleRing
 
 namespace FiniteAdeleRing
 
-instance : CommRing (FiniteAdeleRing R K) := Subalgebra.toCommRing _
+/-- The finite adèle ring of `R`, regarded as a `K`-subalgebra of the
+product of the local completions of `K`. -/
```
ocfnash, line 326: I think it would be good to add some remark along the following lines:
```suggestion
/-- The finite adèle ring of `R`, regarded as a `K`-subalgebra of the
product of the local completions of `K`.

Note that this definition exists to streamline the proof that the finite adèles are an algebra over `K`,
rather than to express their relationship to `K_hat R K` which is essentially a detail of their construction.
-/
```

riccardobrasca:
LGTM to me, but let's keep an eye on it if it becomes too slow (we may need to do the same as we did for the ring of integers). Thanks!

bors merge

Mathlib/RingTheory/DedekindDomain/FiniteAdeleRing.lean
```diff
@@ -312,22 +316,27 @@ end ProdAdicCompletions
...
-of `adicCompletion` with respect to `adicCompletionIntegers`. -/
-def FiniteAdeleRing : Type _ := (
-  { carrier := {x : K_hat R K | x.IsFiniteAdele}
-    mul_mem' := mul
-    one_mem' := one
-    add_mem' := add
-    zero_mem' := zero
-    algebraMap_mem' := algebraMap'
-  } : Subalgebra K (K_hat R K))
+of `adicCompletion`, with respect to `adicCompletionIntegers`. -/
```
ocfnash, line 319: Have I got this right:
```suggestion
of `adicCompletion`, with respect to `adicCompletionIntegers`.

Note that we make this a `Type` rather than a `Subtype` (e.g., a `subalgebra`) since we wish
to endow it with a finer topology than that of the subspace topology. -/
```

ocfnash (approved):
Thanks, two optional suggestions.

ocfnash:
bors d+

riccardobrasca:
bors r-

let's take Oliver's comments into account

kbuzzard:
Finally taken into account!

bors r+

PR #12058, opened 2024-04-10
[Merged by Bors] - feat(Combinatorics/SimpleGraph): Define diameter of simple graphs and provide basic lemmas related to it

Rida-Hamadani:
This defines the diameter of a simple graph, and provides a basic API for it.

Co-authored-by: Matt Diamond <matt@matt-diamond.com>

---

- [x] depends on: #11945
- [x] depends on: #15675

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13706, opened 2024-06-10
[Merged by Bors] - feat(Order): sup/inf equal to top/bot if top/bot in set

Rida-Hamadani:
The sup lemma is needed in a proof related to `SimpleGraph.ediam` in #12058 

---

### DIFF
--- a/Mathlib/Order/CompleteLattice.lean
+++ b/Mathlib/Order/CompleteLattice.lean
@@ -541,10 +541,10 @@ theorem sInf_eq_bot : sInf s = ⊥ ↔ ∀ b > ⊥, ∃ a ∈ s, a < b :=
 #align Inf_eq_bot sInf_eq_bot
 
 lemma sup_eq_top_of_top_mem (h : ⊤ ∈ s) : sSup s = ⊤ :=
-  sSup_eq_top.mpr fun _ _ => by use ⊤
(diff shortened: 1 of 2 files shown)


Mathlib/Order/CompleteLattice.lean
```diff
@@ -540,6 +540,12 @@ theorem sInf_eq_bot : sInf s = ⊥ ↔ ∀ b > ⊥, ∃ a ∈ s, a < b :=
   @sSup_eq_top αᵒᵈ _ _
 #align Inf_eq_bot sInf_eq_bot
 
+lemma sup_eq_top_of_top_mem (h : ⊤ ∈ s) : sSup s = ⊤ :=
```
urkud, line 543: This should be true in a more general case. At least, for a complete lattice. Possibly, for a conditionally complete lattice with top.

urkud: The proof for a conditionally complete lattice + `OrderTop` is `top_unique <| le_csSup (OrderTop.bddAbove s) h`

Rida-Hamadani: You are right, thank you!

urkud:
Thanks! :tada:
bors merge

PR #13707, opened 2024-06-10
[Merged by Bors] - chore(yoneda): remove unneeded universe annotations which break after leanprover/lean4#4408

kim-em:


semorrison:
It's just the `ULift.ext` lines (in which the universes would need to be swapped). But I don't think the universes I've removed here are helping with readability, so lets just remove.

Ruben-VandeVelde (approved):
maintainer merge

TwoFX (approved):
Thanks! 
bors r+

PR #13712, opened 2024-06-11
fix: adapt to rcases fix

llllvvuu:
a proof was relying on buggy behavior that is fixed in https://github.com/leanprover/lean4/pull/4422


semorrison:
This is a PR to a `lean-pr-testing-NNNN` branch. I'm merging now.

PR #13717, opened 2024-06-11
[Merged by Bors] - chore: fix junk value for `sInf` on `WithTop`

sgouezel:
As pointed out in https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Definition.20of.20.60WithTop.2EinstInfSet.60/near/443609808, the current choice of junk value for `sInf s` when `s : Set (WithTop α)` is not bounded below does not coincide with `sInf ∅`. This makes it impossible to get a `ConditionallyCompleteLinearOrder` structure on `WithTop α` as we require the junk values to coincide.

This PR fixes the choice of junk value for  `sInf s` when `s : Set (WithTop α)` is not bounded below, making it equal to `⊤ = sInf ∅`.

---


urkud:
Thanks! :tada:
bors merge

PR #13719, opened 2024-06-11
[Merged by Bors] - feat(Tactic/Relation/Trans): trans tactic for implications

joneugster:
Implications behave like transitive relations albeit not being such in mathlib for technical reasons. This extends the `trans` tactic to handle them alike. This means `trans B` turn a goal `A → C` into two goals `A → B` and `B → C`.

---

Note that this does not modify the existing functionality: So far `trans` only worked on an `Expr` of the form `Expr.app _ _` and this PR adds a case for `Expr.forallE _ _ _ _`.


joneugster:
!bench

Mathlib/Tactic/Relation/Trans.lean
```diff
@@ -109,9 +109,31 @@ that is, a relation which has a transitivity lemma tagged with the attribute [tr
...
+      -- create two new goals
+      let g₁ ← mkFreshExprMVar (some <| .forallE name binderType middle info) .synthetic
+      let g₂ ← mkFreshExprMVar (some <| .forallE name middle body info) .synthetic
+      -- close the original goal with `fun x => g₂ (g₁ x)`
+      goal.assign (.lam name binderType (.app g₂ (.app g₁ (.bvar 0))) .default)
+      pure <| [g₁.mvarId!, g₂.mvarId!] ++ if let some (_, gs') := t'? then gs' else [middle.mvarId!]
+    return
+  | _ => pure ()
+
   let (rel, x, z) ←
```
eric-wieser, line 137: I'd be tempted to put the implication handling here instead, by setting `rel` to `(. -> .)`; is that problematic in some way that I'm not seeing?

test/trans.lean
```diff
@@ -80,3 +81,18 @@ def MyLE (n m : Nat) := ∃ k, n + k = m
...
+/-- `trans` for implications. -/
+example {A B C: Prop} (h : A → B) (g : B → C) : A → C := by
+  trans B
+  · guard_target =ₛ A → B -- ensure we have `B` and not a free metavariable.
+    exact h
+  · guard_target =ₛ B → C
+    exact g
+
+/-- `trans` for arrows between types. -/
+example {A B C: Type} (h : A → B) (g : B → C) : A → C := by
```
eric-wieser, line 94: Can you add a test with the types in different universes?

Mathlib/Tactic/Relation/Trans.lean
joneugster: I looked at it and I think there are a few things that would make it impossible or less readable:

- `rel` is later expected to be an `Expr` of the form `.const` (for example in `mkAppM' rel`) and `(· → ·)` would be I guess something like `.lam x _ (.lam y _ (.forallE _ x y _) _) _`, which would require various modifications. I think that could be circumvented by creating `def Impl (a b):= (a → b)`, but I don't think that's a good idea.
- For the same reason, one couldn't put the `trans` attribute on a lemma like `{A B C : Prop} : (A → B) → (B → C) → (A → C)`, so one would need to modify that part to not search for trans lemmas in this case.
- Less importantly, it seems inefficient to search for relevant `trans` lemmas when there is only one thing to do.

However, I realised that the match statement can be moved around, so that it is one single `match tgt with`. This might indirectly address your comment. (so f223943 is big, but most of it is changing the indent)

Mathlib/Tactic/Relation/Trans.lean
```diff
@@ -109,76 +109,95 @@ that is, a relation which has a transitivity lemma tagged with the attribute [tr
...
-      s.restore
-
-  throwError m!"no applicable transitivity lemma found for {indentExpr tgt}
-"
+      catch e =>
+        trace[Tactic.trans]"failed: {e.toMessageData}"
+        s.restore
+    throwError m!"no applicable transitivity lemma found for {indentExpr tgt}"
+  | _ => throwError (m!"transitivity lemmas only apply to binary relations and " ++
+    m!"non-dependent arrows, not {indentExpr tgt}")
```
eric-wieser, line 201: One thing you could consider, to avoid repeating this similar error message in three places, is something like
```lean
inductive TransRelation
  | app (rel : Expr)
  | implies (name : Name) (bi : BinderInfo)

/-- Finds an explicit binary relation in the argument, if possible-/
def getRel (tgt : Expr) : MetaM (Option (TransRelation × Expr × Expr)) := do
  match tgt with
  | .forallE name binderType body info => return .some (.implies name info, binderType, body)
  | .app f z =>
    match (← getExplicitRelArg? tgt f z) with
    | some (rel, x) =>
      let (rel, x) ← getExplicitRelArgCore tgt rel x z
      return some (.app rel, x, z)
    | none =>
      return none
  | _ => return none

/--
`trans` applies to a goal whose target has the form `t ~ u` where `~` is a transitive relation,
that is, a relation which has a transitivity lemma tagged with the attribute [trans].

* `trans s` replaces the goal with the two subgoals `t ~ s` and `s ~ u`.
* If `s` is omitted, then a metavariable is used instead.

Additionally, `trans` also applies to a goal whose target has the form `t → u`,
in which case it replaces the goal with `t → s` and `s → u`.
-/
elab "trans" t?:(ppSpace colGt term)? : tactic => withMainContext do
  let tgt ← getMainTarget''
  let .some (rel, x, z) ← getRel tgt |
    throwError (m!"transitivity lemmas only apply to binary relations and " ++
                m!"non-dependent arrows, not {indentExpr tgt}")
  match rel with
  | .implies name info =>
    -- only consider non-dependent arrows
    if z.hasLooseBVars then
      throwError "`trans` is not implemented for dependent arrows{indentExpr tgt}"
    -- parse the intermeditate term
    let middleType ← mkFreshExprMVar none
    let t'? ← t?.mapM (elabTermWithHoles · middleType (← getMainTag))
    let middle ← (t'?.map (pure ·.1)).getD (mkFreshExprMVar middleType)
    liftMetaTactic fun goal => do
      -- create two new goals
      let g₁ ← mkFreshExprMVar (some <| .forallE name x middle info) .synthetic
      let g₂ ← mkFreshExprMVar (some <| .forallE name middle z info) .synthetic
      -- close the original goal with `fun x => g₂ (g₁ x)`
      goal.assign (.lam name x (.app g₂ (.app g₁ (.bvar 0))) .default)
      pure <| [g₁.mvarId!, g₂.mvarId!] ++ if let some (_, gs') := t'? then gs' else [middle.mvarId!]
    return
  | .app rel =>
    trace[Tactic.trans]"goal decomposed"
    trace[Tactic.trans]"rel: {indentExpr rel}"
    trace[Tactic.trans]"x: {indentExpr x}"
    trace[Tactic.trans]"z: {indentExpr z}"
    -- first trying the homogeneous case
    try
      let ty ← inferType x
      let t'? ← t?.mapM (elabTermWithHoles · ty (← getMainTag))
      let s ← saveState
      trace[Tactic.trans]"trying homogeneous case"
      let lemmas :=
        (← (transExt.getState (← getEnv)).getUnify rel transExt.config).push ``Trans.simple
      for lem in lemmas do
        trace[Tactic.trans]"trying lemma {lem}"
        try
          liftMetaTactic fun g ↦ do
            let lemTy ← inferType (← mkConstWithLevelParams lem)
            let arity ← withReducible <| forallTelescopeReducing lemTy fun es _ ↦ pure es.size
            let y ← (t'?.map (pure ·.1)).getD (mkFreshExprMVar ty)
            let g₁ ← mkFreshExprMVar (some <| ← mkAppM' rel #[x, y]) .synthetic
            let g₂ ← mkFreshExprMVar (some <| ← mkAppM' rel #[y, z]) .synthetic
            g.assign (← mkAppOptM lem (mkArray (arity - 2) none ++ #[some g₁, some g₂]))
            pure <| [g₁.mvarId!, g₂.mvarId!] ++
              if let some (_, gs') := t'? then gs' else [y.mvarId!]
          return
        catch _ => s.restore
      pure ()
    catch _ =>
    trace[Tactic.trans]"trying heterogeneous case"
    let t'? ← t?.mapM (elabTermWithHoles · none (← getMainTag))
    let s ← saveState
    for lem in (← (transExt.getState (← getEnv)).getUnify rel transExt.config).push
        ``HEq.trans |>.push ``HEq.trans do
      try
        liftMetaTactic fun g ↦ do
          trace[Tactic.trans]"trying lemma {lem}"
          let lemTy ← inferType (← mkConstWithLevelParams lem)
          let arity ← withReducible <| forallTelescopeReducing lemTy fun es _ ↦ pure es.size
          trace[Tactic.trans]"arity: {arity}"
          trace[Tactic.trans]"lemma-type: {lemTy}"
          let y ← (t'?.map (pure ·.1)).getD (mkFreshExprMVar none)
          trace[Tactic.trans]"obtained y: {y}"
          trace[Tactic.trans]"rel: {indentExpr rel}"
          trace[Tactic.trans]"x:{indentExpr x}"
          trace[Tactic.trans]"z:  {indentExpr z}"
          let g₂ ← mkFreshExprMVar (some <| ← mkAppM' rel #[y, z]) .synthetic
          trace[Tactic.trans]"obtained g₂: {g₂}"
          let g₁ ← mkFreshExprMVar (some <| ← mkAppM' rel #[x, y]) .synthetic
          trace[Tactic.trans]"obtained g₁: {g₁}"
          g.assign (← mkAppOptM lem (mkArray (arity - 2) none ++ #[some g₁, some g₂]))
          pure <| [g₁.mvarId!, g₂.mvarId!] ++ if let some (_, gs') := t'? then gs' else [y.mvarId!]
        return
      catch e =>
        trace[Tactic.trans]"failed: {e.toMessageData}"
        s.restore
    throwError m!"no applicable transitivity lemma found for {indentExpr tgt}"
```
(optional)

eric-wieser (approved):
bors d+

Thanks!

Mathlib/Tactic/Relation/Trans.lean
joneugster: That looks good to me, thanks! Would you want to look again at the updated code @eric-wieser ?

eric-wieser: Looks fine to me: but note the linter error

PR #13729, opened 2024-06-11
[Merged by Bors] - feat : added `spectrum.conjugate_units` results to `Algebra.Algebra.Spectrum`, and `spectrum.conjugate_unitary` results to `Algebra.Star.Unitary`.

JonBannon:
Included proofs that conjugation by a unit preserves the spectrum, and conjugation by a unitary preserves the spectrum.

---

### DIFF
--- a/Mathlib/Algebra/Star/Unitary.lean
+++ b/Mathlib/Algebra/Star/Unitary.lean
@@ -5,6 +5,7 @@ Authors: Shing Tak Lam, Frédéric Dupuis
 -/
 import Mathlib.Algebra.Group.Submonoid.Operations
 import Mathlib.Algebra.Star.SelfAdjoint
+import Mathlib.Algebra.Algebra.Spectrum
 
 #align_import algebra.star.unitary from "leanprover-community/mathlib"@"247a102b14f3cebfee126293341af5f6bed00237"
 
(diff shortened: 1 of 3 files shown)


Mathlib/Algebra/Star/Unitary.lean
```diff
@@ -239,4 +240,24 @@ instance : HasDistribNeg (unitary R) :=
...
 
+section UnitaryConjugate
+
+universe u
+
+variable {R A : Type*} [CommSemiring R] [Ring A] [Algebra R A] [StarMul A]
+
+/-- Unitary conjugation preserves the spectrum, star on left. -/
+@[simp]
+lemma spectrum.unitary_conjugate {a : A} {u : unitary A} :
```
dupuisf, line 251: Any reason for the name being the other way around for unitaries? (i.e. `unitary_conjugate` instead of `conjugate_unitary`)? I don't have a strong preference either way, but it would be nice to have the same order as for units.

JonBannon: I have no good reason at all, actually. I'll coordinate these. Thanks for pointing this out!

dupuisf:
Thanks!

bors r+

PR #13610, opened 2024-06-07
[Merged by Bors] - feat(GroupTheory/Congruence/*): add some missing lemma and constructions of group congruence

jjaassoonn:
1. order isomorphism between congruence on group and opposite groups
2. congruence relation is preserved by finite product/sum

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13730, opened 2024-06-11
[Merged by Bors] - feat(RingTheory/Congruence): split `RingCon` construction and add some constructions

jjaassoonn:
added results are mainly about `RingCon` on the opposite ring

---


Mathlib/RingTheory/Congruence/Basic.lean
```diff
@@ -111,6 +111,18 @@ protected theorem mul {w x y z} : c w x → c y z → c (w * y) (x * z) :=
   c.mul'
 #align ring_con.mul RingCon.mul
 
+protected theorem sub {S : Type*} [AddGroup S] [Mul S] (t : RingCon S)
+    {a b c d : S} (h : t a b) (h' : t c d) : t (a - c) (b - d) := t.toAddCon.sub h h'
```
YaelDillies, line 115: Does dot notation on `RingCon `not already work out of the box? by which I mean that without your new lemma I would expect `t.sub` to automatically expand to `t.toAddCon.sub`

Mathlib/RingTheory/Congruence/BigOperators.lean
```diff
@@ -0,0 +1,36 @@
...
+# Interactions between `∑, ∏` and `RingCon`
+
+-/
+
+open BigOperators
+
+namespace RingCon
+
+/-- Congruence relation of a ring preserves finite sum indexed by a list. -/
+protected lemma list_sum {ι S : Type*} [AddCommMonoid S] [Mul S]
```
YaelDillies, line 19:
```suggestion
protected lemma listSum {ι S : Type*} [AddCommMonoid S] [Mul S]
```
and below

Mathlib/RingTheory/Congruence/Basic.lean
YaelDillies: Ah maybe it proves `t.toAddCon (a - c) (b - d)`... Mind checking?

jjaassoonn: The following gave an error
```lean
example  {S : Type*} [AddGroup S] [Mul S] (t : RingCon S) : true := by
  have := t.sub

/-
invalid field 'sub', the environment does not contain 'RingCon.sub'
  t
has type
  RingCon S
-/
```

Mathlib/RingTheory/Congruence/BigOperators.lean
jjaassoonn: thanks

Mathlib/RingTheory/Congruence/Opposite.lean
```diff
@@ -0,0 +1,53 @@
...
+
+/--
+If `c` is a `RingCon R`, then `(a, b) ↦ c b.unop a.unop` is a `RingCon Rᵐᵒᵖ`.
+-/
+def op (c : RingCon R) : RingCon Rᵐᵒᵖ where
+  __ := c.toCon.op
+  mul' h1 h2 := c.toCon.op.mul h1 h2
+  add' h1 h2 := c.add h1 h2
+
+lemma op_iff {c : RingCon R} {x y : Rᵐᵒᵖ} : c.op x y ↔ c (y.unop) (x.unop) := Iff.rfl
```
YaelDillies, line 30:
```suggestion
lemma op_iff {c : RingCon R} {x y : Rᵐᵒᵖ} : c.op x y ↔ c y.unop x.unop := Iff.rfl
```

Mathlib/RingTheory/Congruence/Opposite.lean
```diff
@@ -0,0 +1,53 @@
...
+-/
+def unop (c : RingCon Rᵐᵒᵖ) : RingCon R where
+  __ := c.toCon.unop
+  mul' h1 h2 := c.toCon.unop.mul h1 h2
+  add' h1 h2 := c.add h1 h2
+
+lemma unop_iff {c : RingCon Rᵐᵒᵖ} {x y : R} : c.unop x y ↔ c (.op y) (.op x) := Iff.rfl
+
+/--
+The congruences of a ring `R` bijects to the congrunences of the opposite ring `Rᵐᵒᵖ`.
```
YaelDillies, line 43:
```suggestion
The congruences of a ring `R` biject to the congruences of the opposite ring `Rᵐᵒᵖ`.
```

Mathlib/RingTheory/Congruence/Opposite.lean
```diff
@@ -0,0 +1,53 @@
...
+-/
+@[simps]
+def orderIsoOp : RingCon R ≃o RingCon Rᵐᵒᵖ where
```
YaelDillies, line 46:
```suggestion
def opOrderIso : RingCon R ≃o RingCon Rᵐᵒᵖ where
```

YaelDillies (approved):
Thanks!

maintainer merge

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/GroupTheory/Congruence/BigOperators.lean
```diff
@@ -20,7 +20,7 @@ namespace Con
 
 /-- Multiplicative congruence relations preserve product indexed by a list. -/
 @[to_additive "Additive congruence relations preserve sum indexed by a list."]
-protected theorem list_prod {ι M : Type*} [CommMonoid M] (c : Con M) {l : List ι} {f g : ι → M}
+protected theorem list_prod {ι M : Type*} [Monoid M] (c : Con M) {l : List ι} {f g : ι → M}
```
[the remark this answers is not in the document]
rosborn: Gonna copy my comment from another PR which I've closed.
"Technically, [this does not need associativity]. If we really wanted to generalize, we should have api for a product over a binary tree (or free magma). I do have a feeling that mathlib is using lists and multisets for things that are morally free monoids and free commutative monoids, but that might be a refactor for another day."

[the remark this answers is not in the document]
jjaassoonn: `List.prod_cons` requires `Monoid M`. I think removing associativity here means we need to remove associativity at `List.prod_cons` as well.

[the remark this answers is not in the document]
rosborn: As `prod` is defined with `foldl`, `prod_cons` requires associativity. You would need to unfold `prod` and work with `foldl` directly to prove this without associativity. I would argue this generalization requires more work and is outside of the scope of this PR.

Mathlib/RingTheory/Congruence/Opposite.lean
```diff
@@ -0,0 +1,53 @@
...
+the opposite ring `Rᵐᵒᵖ`.
+
+-/
+
+variable {R : Type*} [Add R] [Mul R]
+
+namespace RingCon
+
+/--
+If `c` is a `RingCon R`, then `(a, b) ↦ c b.unop a.unop` is a `RingCon Rᵐᵒᵖ`.
```
eric-wieser, line 23: Any idea why we swap the argument order?

eric-wieser: Ah, I guess this question is better directed at https://github.com/leanprover-community/mathlib4/pull/13610/files#r2032230268

PR #13740, opened 2024-06-11
feat: more robust ae notation

YaelDillies:
Make sure that, when `μ : FiniteMeasure Ω`, `f =ᵐ[μ] g` elaborates to `f =ᵐ[↑μ] g` instead of complaining about `OuterMeasureClass (FiniteMeasure Ω) (Set Ω) ℝ≥0∞` not existing.

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/ae.20of.20a.20finite.20measure)

---

### DIFF
--- a/Mathlib/MeasureTheory/Integral/Bochner/L1.lean
+++ b/Mathlib/MeasureTheory/Integral/Bochner/L1.lean
@@ -341,8 +341,8 @@ lemma integral_mono {f g : α →ₛ F} (h : f ≤ᵐ[μ] g) (hf : Integrable f 
   rw [← sub_nonneg_ae] at h
   exact integral_nonneg h
 
-lemma integral_mono_measure {ν} {f : α →ₛ F} (hf : 0 ≤ᵐ[ν] f) (hμν : μ ≤ ν) (hfν : Integrable f ν) :
-    f.integral μ ≤ f.integral ν := by
+lemma integral_mono_measure {ν : Measure _} {f : α →ₛ F} (hf : 0 ≤ᵐ[ν] f) (hμν : μ ≤ ν)
+    (hfν : Integrable f ν) : f.integral μ ≤ f.integral ν := by
   simp only [integral_eq]
   apply Finset.sum_le_sum
   simp only [forall_mem_range]
--- a/Mathlib/MeasureTheory/Measure/Decomposition/IntegralRNDeriv.lean
+++ b/Mathlib/MeasureTheory/Measure/Decomposition/IntegralRNDeriv.lean
@@ -65,7 +65,7 @@ lemma mul_le_integral_rnDeriv_of_ac [IsFiniteMeasure μ] [IsFiniteMeasure ν]
   · simp [hν]
   have : NeZero ν := ⟨hν⟩
   let μ' := (ν univ)⁻¹ • μ
-  let ν' := (ν univ)⁻¹ • ν
+  let ν' : Measure α := (ν univ)⁻¹ • ν
   have : IsFiniteMeasure μ' := μ.smul_finite (by simp [hν])
   have hμν' : μ' ≪ ν' := hμν.smul _
   have h_rnDeriv_eq : μ'.rnDeriv ν' =ᵐ[ν] μ.rnDeriv ν := by
--- a/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
+++ b/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
@@ -95,27 +95,23 @@ def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM E
     let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
     elabTerm μ q(Measure $ty)
 
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+elab "elab_as_measure% " t:term : term => do elabMeasure t ‹_›
+
 /-- `f =ᵐ[μ] g` means `f` and `g` are eventually equal along the a.e. filter,
 i.e. `f=g` away from a null set.
 
 This is notation for `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
-syntax:50 (name := aeEq) term:50 " =ᵐ[" term:50 "] " term:50 : term
+macro:50 (name := aeEq) f:term:50 " =ᵐ[" μ:term:50 "] " g:term:50 : term =>
+  `(Filter.EventuallyEq (MeasureTheory.ae (elab_as_measure% $μ)) $f $g)
 
 /-- `f ≤ᵐ[μ] g` means `f` is eventually less than `g` along the a.e. filter,
 i.e. `f ≤ g` away from a null set.
 
 This is notation for `Filter.EventuallyLE (MeasureTheory.ae μ) f g`. -/
-syntax:50 (name := aeLE) term:50 " ≤ᵐ[" term:50 "] " term:50 : term
-
-/-- Elaborate almost everywhere equal notation.
-
-We elabore `f =ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
-@[term_elab aeEq]
-def elabAEEq : TermElab
-  | `($f =ᵐ[$μ] $g), expectedType? => do
-    let μ' ← elabMeasure μ expectedType?
-    elabTerm (← `(Filter.EventuallyEq (MeasureTheory.ae $(← μ'.toSyntax)) $f $g)) expectedType?
-  | _, _ => throwUnsupportedSyntax
+macro:50 (name := aeLE) f:term:50 " ≤ᵐ[" μ:term:50 "] " g:term:50 : term =>
+  `(Filter.EventuallyLE (MeasureTheory.ae (elab_as_measure% $μ)) $f $g)
 
 /-- Elaborate almost everywhere equal notation.
 
--- a/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
+++ b/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
@@ -94,13 +94,14 @@ def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM E
     let ty ← mkFreshExprMVarQ q(Type v)
     let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
     elabTerm μ q(Measure $ty)
+  | _, _ => throwUnsupportedSyntax
 
 /-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
 elaborate `μ` as `Measure ?Ω`. -/
 elab "elab_as_measure% " t:term : term => do elabMeasure t ‹_›
 
 /-- `f =ᵐ[μ] g` means `f` and `g` are eventually equal along the a.e. filter,
-i.e. `f=g` away from a null set.
+i.e. `f = g` away from a null set.
 
 This is notation for `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
 macro:50 (name := aeEq) f:term:50 " =ᵐ[" μ:term:50 "] " g:term:50 : term =>
--- a/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
+++ b/Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
@@ -79,6 +79,58 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
 /-- Notation for `Measure` with respect to a non-standard σ-algebra in the domain. -/
 scoped notation "Measure[" mα "] " α:arg => @Measure α mα
 
+end MeasureTheory
+
+namespace Mathlib.Meta
+open Lean Elab Term Meta Qq MeasureTheory
+
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM Expr := do
+  let ⟨u, T, _⟩ ← inferTypeQ (← elabTerm μ expectedType?)
+  match u, T with
+  | .succ v, ~q(OuterMeasure $α) => elabTerm μ <| .some q(OuterMeasure $α)
+  | .succ v, _ =>
+    let ty ← mkFreshExprMVarQ q(Type v)
+    let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
+    elabTerm μ q(Measure $ty)
+
+/-- `f =ᵐ[μ] g` means `f` and `g` are eventually equal along the a.e. filter,
+i.e. `f=g` away from a null set.
+
+This is notation for `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+syntax:50 (name := aeEq) term:50 " =ᵐ[" term:50 "] " term:50 : term
+
+/-- `f ≤ᵐ[μ] g` means `f` is eventually less than `g` along the a.e. filter,
+i.e. `f ≤ g` away from a null set.
+
+This is notation for `Filter.EventuallyLE (MeasureTheory.ae μ) f g`. -/
+syntax:50 (name := aeLE) term:50 " ≤ᵐ[" term:50 "] " term:50 : term
+
+/-- Elaborate almost everywhere equal notation.
+
+We elabore `f =ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+@[term_elab aeEq]
+def elabAEEq : TermElab
+  | `($f =ᵐ[$μ] $g), expectedType? => do
+    let μ' ← elabMeasure μ expectedType?
+    elabTerm (← `(Filter.EventuallyEq (MeasureTheory.ae $(← μ'.toSyntax)) $f $g)) expectedType?
+  | _, _ => throwUnsupportedSyntax
+
+/-- Elaborate almost everywhere equal notation.
+
+We elabore `f ≤ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+@[term_elab aeLE]
+def elabAELE : TermElab
+  | `($f ≤ᵐ[$μ] $g), expectedType? => do
+    let μ' ← elabMeasure μ expectedType?
+    elabTerm (← `(Filter.EventuallyLE (MeasureTheory.ae $(← μ'.toSyntax)) $f $g)) expectedType?
+  | _, _ => throwUnsupportedSyntax
+
+end Mathlib.Meta
+
+namespace MeasureTheory
+
 theorem Measure.toOuterMeasure_injective [MeasurableSpace α] :
     Injective (toOuterMeasure : Measure α → OuterMeasure α)
   | ⟨_, _, _⟩, ⟨_, _, _⟩, rfl => rfl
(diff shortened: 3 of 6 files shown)


Mathlib/MeasureTheory/OuterMeasure/AE.lean
```diff
@@ -32,18 +32,29 @@ However, we restate some lemmas specifically for `ae`.
...
-def ae (μ : F) : Filter α :=
+def MeasureTheory.ae (μ : F) : Filter α :=
   .ofCountableUnion (μ · = 0) (fun _S hSc ↦ (measure_sUnion_null_iff hSc).2) fun _t ht _s hs ↦
     measure_mono_null hs ht
 
+namespace Mathlib.Meta
+open Lean Elab Term Meta Qq
+
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasure T`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
```
urkud, line 49: You may want to try `Measure ?_` first, because otherwise Lean may decide to cast a `Measure` to an `OuterMeasure`.

YaelDillies: Sorry, you got misled by a typo in my docstring. It's supposed to be
```suggestion
/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T`. If that fails, try to
elaborate `μ` as `Measure ?Ω`. -/
```

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -74,6 +74,52 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+syntax:50 (name := aeeq) term:50 " =ᵐ[" term:50 "] " term:50 : term
+
+/-- Elaborate almost everywhere equal notation.
+
+We elabore `f =ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+@[term_elab aeeq]
+def elabAEEq : TermElab
+  | `($f =ᵐ[$μ] $g), expectedType? => do
+    let μ' ← elabMeasure μ
+    elabTerm (← `(Filter.EventuallyEq (MeasureTheory.ae $μ') $f $g)) expectedType?
```
YaelDillies, line 116: What am I doing wrong here? It's expecting `μ'` to be a `TSyntax`, but my `μ'` is a `Expr`

eric-wieser: You can use `$(μ'.toSyntax)`

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -74,6 +74,58 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+
+/-- Elaborate almost everywhere equal notation.
+
+We elabore `f =ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+@[term_elab aeEq]
+def elabAEEq : TermElab
+  | `($f =ᵐ[$μ] $g), expectedType? => do
+    let μ' ← elabMeasure μ expectedType?
+    elabTerm (← `(Filter.EventuallyEq (MeasureTheory.ae $(← μ'.toSyntax)) $f $g)) expectedType?
+  | _, _ => throwUnsupportedSyntax
```
eric-wieser, line 113: Instead of going back and forth between Expr and Syntax here, I think it would be be better to translate `$f =ᵐ[$μ] $g` to `Filter.EventuallyEq (MeasureTheory.ae (elab_as_measure% $μ)) $f $g` where `elab_as_measure` is itself an elaborator.

YaelDillies: Okay but what about my comment on Zulip? Will your solution let me use the expected type of the functions to elaborate the measure?

eric-wieser: I have no idea, try it and see!

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -90,27 +90,23 @@ def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM E
     let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
     elabTerm μ q(Measure $ty)
 
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+elab "elab_as_measure% " t:term : term => do elabMeasure t ‹_›
```
eric-wieser, line 95: I'm surprised `‹_›` works here!

It might be better to instead put `@[term_elab elab_as_measure]` on the `def` above and use
```suggestion
syntax (name := elab_as_measure) "elab_as_measure% " t:term : term
```
which also shortcuts some expected type logic you might not need.

YaelDillies: I looked at the context and saw there was an inaccessible `Option Expr` 🕵️

Mathlib/MeasureTheory/Integral/Bochner.lean
```diff
@@ -1352,7 +1352,7 @@ theorem integral_mono_of_nonneg {f g : α → ℝ} (hf : 0 ≤ᵐ[μ] f) (hgi :
   · rw [integral_non_aestronglyMeasurable hfm]
     exact integral_nonneg_of_ae (hf.trans h)
 
-theorem integral_mono_measure {f : α → ℝ} {ν} (hle : μ ≤ ν) (hf : 0 ≤ᵐ[ν] f)
+theorem integral_mono_measure {f : α → ℝ} {ν : Measure α} (hle : μ ≤ ν) (hf : 0 ≤ᵐ[ν] f)
```
YaelDillies, line 1355: Any idea why `(hf : 0 ≤ᵐ[ν] f)` was erroring with `nonexhaustive match` without that change?

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
YaelDillies:
```
@[term_elab elab_as_measure]
def elabMeasure (μ : Syntax) (expectedType? : Option Expr) : TermElabM Expr := do
```
errors with
```
unexpected type at 'Mathlib.Meta.elabMeasure', `Lean.Elab.Term.TermElab` expected
```

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -74,6 +74,54 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+namespace Mathlib.Meta
+open Lean Elab Term Meta Qq MeasureTheory
+
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM Expr := do
+  let ⟨u, T, _⟩ ← inferTypeQ (← elabTerm μ expectedType?)
+  match u, T with
+  | .succ v, ~q(OuterMeasure $α) => elabTerm μ <| .some q(OuterMeasure $α)
+  | .succ v, _ =>
```
eric-wieser, line 88: You're missing the match for `.zero` here

Mathlib/MeasureTheory/Integral/Bochner.lean
eric-wieser: Yes, because otherwise you have `v : ?T` and `T : Sort ?u`, and `?u` is a metavariable so does not participate in the match.

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
YaelDillies: `zero` I just discard, right? How should I handle metavariables? Should I assign them?

leanprover-community-bot-assistant:

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
eric-wieser: I guess I missed the `_, _` case below, indeed you don't need zero.

I think you can instead use `inferTypeQ'` which handles the metavariable case and removes the need for the universe match.

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
eric-wieser: You have to intro the `expectedType?` after the colon.

Mathlib/MeasureTheory/OuterMeasure/AE.lean
```diff
@@ -56,16 +56,16 @@ This is notation for `Filter.Frequently p (MeasureTheory.ae μ)`. -/
...
 This is notation for `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
-notation:50 f " =ᵐ[" μ:50 "] " g:50 => Filter.EventuallyEq (MeasureTheory.ae μ) f g
+local notation:50 f " =ᵐ[" μ:50 "] " g:50 => Filter.EventuallyEq (MeasureTheory.ae μ) f g
 
 /-- `f ≤ᵐ[μ] g` means `f` is eventually less than `g` along the a.e. filter,
 i.e. `f ≤ g` away from a null set.
 
 This is notation for `Filter.EventuallyLE (MeasureTheory.ae μ) f g`. -/
-notation:50 f " ≤ᵐ[" μ:50 "] " g:50 => Filter.EventuallyLE (MeasureTheory.ae μ) f g
+local notation:50 f " ≤ᵐ[" μ:50 "] " g:50 => Filter.EventuallyLE (MeasureTheory.ae μ) f g
```
eric-wieser, line 70: Why are these now `local`?

YaelDillies: because they are replaced with the general notation later on, and the general notation cannot be defined earlier due to it relying on the existence of `Measure`

eric-wieser: Can you remark on that in the docstring?

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
YaelDillies: > I guess I missed the `_, _` case below

You missed it before because I added it just now 😇

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
YaelDillies: Replacing `elabMeasure` and the `elab` with
```
syntax (name := elab_as_measure) "elab_as_measure% " term : term

/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
elaborate `μ` as `Measure ?Ω`. -/
@[term_elab elab_as_measure]
def elabMeasure : TermElab := fun μ expectedType? ↦ do
  let ⟨u, T, _⟩ ← inferTypeQ' (← elabTerm μ expectedType?)
  match T with
  | ~q(OuterMeasure $α) => elabTerm μ <| .some q(OuterMeasure $α)
  | _ =>
    let ty ← mkFreshExprMVarQ q(Type u)
    let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
    elabTerm μ q(Measure $ty)
```
makes the notation fail with `maximum recursion depth has been reached`

thorimur: The issue here is that `μ` is the syntax `elab_as_measure% $μ`, not the measure syntax itself, so this elaborator keeps trying to call itself via `elabTerm`! :)

I think it might be okay to leave things as is here, but with a comment explaining `‹_›`. (Note: you can also jettison the `do`!) Mostly, this is because I think it's nice to have the `TermElabM` API for `elabMeasure`.

The need for this exposes some friction with the syntax for handling expected types...we could do with some better ergonomics here. (Some discussion [here](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/.E2.9C.94.20Optional.20expected.20type.20in.20.60elab_rules.60/with/425791354).)

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -79,6 +79,54 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM Expr := do
+  let ⟨u, T, _⟩ ← inferTypeQ' (← elabTerm μ expectedType?)
+  match T with
+  | ~q(OuterMeasure $α) => elabTerm μ <| .some q(OuterMeasure $α)
+  | _ =>
+    let ty ← mkFreshExprMVarQ q(Type u)
+    let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
+    elabTerm μ q(Measure $ty)
```
thorimur, line 96: I'm a little wary of elaborating the same syntax multiple times! :) I have two points here:
1. We can simply `return` the expression we've elaborated if it has the correct type, right?
2. We can use something like `Term.ensureHasType` to coerce on the expression level. (I am assuming we want to coerce, correct?)

What do you think of something like
```suggestion
def elabMeasure (μ : TSyntax `term) (expectedType? : Option Expr) : TermElabM Expr := do
  let ⟨u, T, e⟩ ← inferTypeQ' (← elabTerm μ expectedType?)
  match T with
  | ~q(OuterMeasure $α) => return e
  | _ =>
    let ty ← mkFreshExprMVarQ q(Type u)
    let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
    Term.ensureHasType q(Measure $ty) e
      s!"Could not elaborate to either an `OuterMeasure` or a `Measure`.\n\n`{← ppExpr e}`"
```
?

Now, I haven't been able to test this too well, so it might not work in edge cases.

To that end: would you mind adding a test file for this notation, covering both when it should succeed and when it should error?

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -79,6 +79,54 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+
+/-- Elaborate almost everywhere equal notation.
+
+We elabore `f ≤ᵐ[μ] g` as `Filter.EventuallyEq (MeasureTheory.ae μ) f g`. -/
+@[term_elab aeLE]
+def elabAELE : TermElab
+  | `($f ≤ᵐ[$μ] $g), expectedType? => do
+    let μ' ← elabMeasure μ expectedType?
+    elabTerm (← `(Filter.EventuallyLE (MeasureTheory.ae $(← μ'.toSyntax)) $f $g)) expectedType?
+  | _, _ => throwUnsupportedSyntax
```
thorimur, line 124: This definition is now covered by the macro above and can be removed, correct?
```suggestion
```

Mathlib/MeasureTheory/Measure/MeasureSpaceDef.lean
```diff
@@ -79,6 +79,54 @@ structure Measure (α : Type*) [MeasurableSpace α] extends OuterMeasure α wher
...
+  | ~q(OuterMeasure $α) => elabTerm μ <| .some q(OuterMeasure $α)
+  | _ =>
+    let ty ← mkFreshExprMVarQ q(Type u)
+    let _ ← mkFreshExprMVarQ q(MeasurableSpace $ty)
+    elabTerm μ q(Measure $ty)
+
+/-- Try to elaborate `μ` as a term of type `T` where `OuterMeasureClass T ?Ω`. If that fails, try to
+elaborate `μ` as `Measure ?Ω`. -/
+elab "elab_as_measure% " t:term : term => do elabMeasure t ‹_›
+
```
thorimur, line 101: I think we can get rid of those `set_option quotPrecheck false`s by adding this:
```suggestion
elab (name := elabAsMeasureStx) "elab_as_measure% " t:term : term => elabMeasure t ‹_›

open Lean.Elab.Term.Quotation in
/-- `quot_precheck` for the `elab_as_measure%` syntax. Helps e.g. `notation` commands. -/
@[quot_precheck Mathlib.Meta.elabAsMeasureStx] def precheckElabAsMeasureStx : Precheck
  | `(term| elab_as_measure% $μ:term) => precheck μ
  | _ => Lean.Elab.throwUnsupportedSyntax

```
(See also `precheckFinsetStx` for a completely analogous example.)

Mathlib/Probability/Notation.lean
```diff
@@ -55,9 +55,11 @@ scoped[ProbabilityTheory] notation "𝔼[" X "]" => ∫ a, (X : _ → _) a
 scoped[ProbabilityTheory] notation P "⟦" s "|" m "⟧" =>
   MeasureTheory.condExp m P (Set.indicator s fun ω => (1 : ℝ))
 
+set_option quotPrecheck false in
 /-- `X =ₐₛ Y` if `X = Y` almost surely. -/
 scoped[ProbabilityTheory] notation:50 X " =ₐₛ " Y:50 => X =ᵐ[MeasureTheory.MeasureSpace.volume] Y
 
+set_option quotPrecheck false in
```
thorimur, line 62: (After the review suggestion above)
```suggestion
/-- `X =ₐₛ Y` if `X = Y` almost surely. -/
scoped[ProbabilityTheory] notation:50 X " =ₐₛ " Y:50 => X =ᵐ[MeasureTheory.MeasureSpace.volume] Y

```

Mathlib/MeasureTheory/Integral/Bochner/L1.lean
```diff
@@ -341,8 +341,8 @@ lemma integral_mono {f g : α →ₛ F} (h : f ≤ᵐ[μ] g) (hf : Integrable f
   rw [← sub_nonneg_ae] at h
   exact integral_nonneg h
 
-lemma integral_mono_measure {ν} {f : α →ₛ F} (hf : 0 ≤ᵐ[ν] f) (hμν : μ ≤ ν) (hfν : Integrable f ν) :
-    f.integral μ ≤ f.integral ν := by
+lemma integral_mono_measure {ν : Measure _} {f : α →ₛ F} (hf : 0 ≤ᵐ[ν] f) (hμν : μ ≤ ν)
+    (hfν : Integrable f ν) : f.integral μ ≤ f.integral ν := by
```
thorimur, line 345: Does this belong in this PR?
```suggestion
lemma integral_mono_measure {ν} {f : α →ₛ F} (hf : 0 ≤ᵐ[ν] f) (hμν : μ ≤ ν) (hfν : Integrable f ν) :
    f.integral μ ≤ f.integral ν := by
```

Mathlib/MeasureTheory/Measure/Decomposition/IntegralRNDeriv.lean
```diff
@@ -65,7 +65,7 @@ lemma mul_le_integral_rnDeriv_of_ac [IsFiniteMeasure μ] [IsFiniteMeasure ν]
   · simp [hν]
   have : NeZero ν := ⟨hν⟩
   let μ' := (ν univ)⁻¹ • μ
-  let ν' := (ν univ)⁻¹ • ν
+  let ν' : Measure α := (ν univ)⁻¹ • ν
```
thorimur, line 68: Does this belong in this PR?
```suggestion
  let ν' := (ν univ)⁻¹ • ν
```

thorimur (commented):
I've given this a review with the hopes of seeing it merged soon. :) The biggest change requested is a test file!

I'd also like to see if an approach which doesn't elaborate twice (suggested in detail below) is feasible.

urkud:
Even though I've introduced `OuterMeasureClass` some time ago, I'm not sure if this was the right call. It's a typeclass (=extra overhead) that covers exactly 2 types, and many lemmas (e.g., `ae (μ + ν)`) can't be stated for both. Should we go back to 2 separate types?

YaelDillies:
I believe we would just come back to the situation before, where Lean could disambiguate between `OuterMeasure.ae` and `Measure.ae` by itself, even when applied to a `FiniteMeasure` or a `ProbabilityMeasure`.

YaelDillies:
I believe this PR is now contingent on Yury not having time to undo his previous refactor.

YaelDillies:
Sure, although I personally don't have time either. Maybe we ask an LLM to revert Yury's PR?

thorimur:
I know there was some discussion earlier, before the holidays (also involving @urkud); has there been progress on this?

I'm going to mark it awaiting-author for the time being, even if technically what we mean is "potentially awaiting a separate PR". :) (Maybe an `awaiting-other` label would be helpful, as I've seen this sort of thing come up before.)

YaelDillies:
This has been around for too long and I do not care enough about it

PR #13747, opened 2024-06-12
[Merged by Bors] - chore(*): removing unused imports

kim-em:
These are imports of material from Batteries. Not sure why `shake` has not complained.


jcommelin (commented):
Thanks :tada:

bors merge

PR #13748, opened 2024-06-12
[Merged by Bors] - feat: positivity of `star x * x`

eric-wieser:
It is positive when `IsRegular x` (which requires weaker constraints on `R` than `IsUnit x` or `x ≠ 0`)

---

### DIFF
--- a/Mathlib/Algebra/Star/Order.lean
+++ b/Mathlib/Algebra/Star/Order.lean
@@ -5,6 +5,7 @@ Authors: Scott Morrison
 -/
 import Mathlib.Algebra.Group.Submonoid.Operations
 import Mathlib.Algebra.Star.SelfAdjoint
+import Mathlib.Algebra.Regular.Basic
 
 #align_import algebra.star.order from "leanprover-community/mathlib"@"31c24aa72e7b3e5ed97a8412470e904f82b81004"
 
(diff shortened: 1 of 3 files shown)


Mathlib/Algebra/Star/Order.lean
```diff
@@ -233,6 +234,16 @@ lemma IsSelfAdjoint.mono {x y : R} (h : x ≤ y) (hx : IsSelfAdjoint x) : IsSelf
 lemma IsSelfAdjoint.of_nonneg {x : R} (hx : 0 ≤ x) : IsSelfAdjoint x :=
   (isSelfAdjoint_zero R).mono hx
 
+theorem star_mul_self_pos [Nontrivial R] {x : R} (hx : IsRegular x) : 0 < star x * x := by
+  rw [lt_iff_le_and_ne]
+  refine ⟨star_mul_self_nonneg _, ?_⟩
```
Ruben-VandeVelde, line 239:
```suggestion
  rw [(star_mul_self_nonneg _).lt_iff_ne]
```

Ruben-VandeVelde (approved):
lgtm

j-loreaux:
This is weird to me because `IsRegular` seems *so* much stronger than `x ≠ 0`. In fact, `(hx : 0 ≤ x) (hx' : IsRegular x)` is basically the definition of *strictly positive* in a C⋆-algebra (although there are other equivalent ones). Nevertheless, I think there's nothing wrong with having these.

bors merge

eric-wieser:
> This is weird to me because `IsRegular` seems _so_ much stronger than `x ≠ 0`.

I think I stated things unclearly; what I meant is that `[Field R] (x : R) (hx : x ≠ 0)` is a much stronger requirement than `[Ring R] (x : R) (hx : IsRegular R)`.

PR #13749, opened 2024-06-12
feat(GroupTheory/Congruence/BigOperators) Generalize list_prod over Monoids

rosborn:
---

### DIFF
--- a/Mathlib/GroupTheory/Congruence/BigOperators.lean
+++ b/Mathlib/GroupTheory/Congruence/BigOperators.lean
@@ -20,7 +20,7 @@ namespace Con
 
 /-- Multiplicative congruence relations preserve product indexed by a list. -/
 @[to_additive "Additive congruence relations preserve sum indexed by a list."]
-protected theorem list_prod {ι M : Type*} [CommMonoid M] (c : Con M) {l : List ι} {f g : ι → M}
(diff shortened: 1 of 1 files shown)


Mathlib/GroupTheory/Congruence/BigOperators.lean
```diff
@@ -20,7 +20,7 @@ namespace Con
 
 /-- Multiplicative congruence relations preserve product indexed by a list. -/
 @[to_additive "Additive congruence relations preserve sum indexed by a list."]
-protected theorem list_prod {ι M : Type*} [CommMonoid M] (c : Con M) {l : List ι} {f g : ι → M}
+protected theorem list_prod {ι M : Type*} [Monoid M] (c : Con M) {l : List ι} {f g : ι → M}
```
urkud, line 23: Do we need any structure other than `[Mul M] [One M]` for this lemma?

rosborn: Technically, no. If we really wanted to generalize, we should have api for a product over a binary tree (or free magma). I do have a feeling that mathlib is using lists and multisets for things that are morally free monoids and free commutative monoids, but that might be a refactor for another day.

rosborn:
Fixed in #13730

PR #13751, opened 2024-06-12
[Merged by Bors] - chore: remove unused countLocalHypsUsed

kim-em:
This is an orphaned meta function, that used to be used in `rw?`, which has since moved to core. I wrote it, and I'm pretty sure I was the only user, so I would like to just delete it without a deprecation.


jcommelin:
Thanks :tada:

bors merge

semorrison:
bors merge

PR #13746, opened 2024-06-11
[Merged by Bors] - chore(HashMap): deprecate unused API additions

kim-em:

PR #13753, opened 2024-06-12
[Merged by Bors] - chore(HashMap): deprecated unused API additions (2/2)

kim-em:
- [x] depends on: #13752


Mathlib/Data/List/Basic.lean
```diff
@@ -1765,19 +1765,6 @@ theorem cons_get_drop_succ {l : List α} {n} :
 #align list.drop_take List.drop_take
 #align list.map_drop List.map_drop
 #align list.modify_nth_tail_eq_take_drop List.modifyNthTail_eq_take_drop
-
-@[simp]
-theorem modifyNth_nil (f : α → α) (n : ℕ) :
```
grunweg, line 1770: Let me ask for the record: are these used anywhere? Is there a particular reason these are removed; is there an obvious alternative? Should these be deprecated? (I would not demand this, but would like to understand why no deprecation is warranted if not.)

grunweg (commented):
The execution is straightforward; two questions about (documenting) the *why*.

Mathlib/Data/List/Basic.lean
semorrison: Oh, these have just moved upstream in the meantime.

grunweg:
If you like, feel free to make a similar comment tweak as in #13746. Either way,
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/Data/HashMap.lean
```diff
@@ -47,6 +47,7 @@ end Batteries.HashMap
 namespace Batteries.RBSet
 
 /-- Insert all elements of a list into an `RBSet`. -/
+@[deprecated "Cleaning up unused API additions in Mathlib." (since := "2024-06-12")]
 def insertList {cmp} (m : RBSet α cmp) (L : List α) : RBSet α cmp :=
```
eric-wieser, line 51: Should this just point to `insertMany`?

PR #13742, opened 2024-06-11
[Merged by Bors] - feat: the distinguished triangle in the derived category associated to a short exact sequence of cochain complexes

joelriou:
---

- [x] depends on: #13675
- [x] depends on: #13757

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13757, opened 2024-06-12
[Merged by Bors] - feat(CategoryTheory): more API for the commutation of functors with shifts

joelriou:
In this PR, it is shown that the composition of two functors which commute with shifts also commutes with shifts. The notion of natural transformation which commutes with the shifts is also introduced.

This shall be used in the development of the API for the derived category #13742.

---


erdOne (commented):
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13758, opened 2024-06-12
chore(RingTheory/PowerSeries): simp golfing

rwst:
Where possible:
simp --> simp only --> simp_rw --> rw
simpa --> simpa only

---


grunweg:
!bench

grunweg:
Thanks for your contribution! I'm in two minds about this PR, because there is a trade-off involved here:
- proofs compiling faster means faster turn-around on CI and makes local development more pleasant
- on the other hand, readability counts, and e.g. replacing one `simp` by a long `simp only` is strictly *less* readable

Hence, I wonder: did you have any particular reasons for removing these `simp`s? (Or were you simple "replacing all you could? This being possible is useful information, but I'm not sure if the PR should land in the current form.)

As a first step, I've sent it to the benchmarking bot: if a file gets significantly faster, that is definitely useful information.

grunweg:
By the way: did you hear about the `profiler` flag? You can locally profile a file and see which parts are slow.
If a `simp` is pretty fast, replacing it by a longer/uglier `simp only` would be a net negative in my book.
If a `simp` is very slow, but a `simp only` is much faster, the calculus might change. (This might still be an occasion to investigate `why` the simp is slow and if e.g. some simp lemmas are missing or sub-optimally stated: that is a harder task, though, which requires knowing the surrounding library a bit.)

[This PR](https://github.com/leanprover-community/leanprover-community.github.io/pull/472) contains some basic advice on how to use the profiler.

Mathlib/RingTheory/PowerSeries/Basic.lean
```diff
@@ -360,13 +360,13 @@ theorem constantCoeff_smul {S : Type*} [Semiring S] [Module R S] (φ : PowerSeri
 
 theorem smul_eq_C_mul (f : R⟦X⟧) (a : R) : a • f = C R a * f := by
   ext
-  simp
+  rw [LinearMapClass.map_smul, smul_eq_mul, coeff_C_mul]
```
grunweg, line 363: This changes feels significantly less readable for me.

Mathlib/RingTheory/PowerSeries/Basic.lean
```diff
@@ -360,13 +360,13 @@ theorem constantCoeff_smul {S : Type*} [Semiring S] [Module R S] (φ : PowerSeri
...
 set_option linter.uppercaseLean3 false in
 #align power_series.smul_eq_C_mul PowerSeries.smul_eq_C_mul
 
 @[simp]
 theorem coeff_succ_mul_X (n : ℕ) (φ : R⟦X⟧) : coeff R (n + 1) (φ * X) = coeff R n φ := by
-  simp only [coeff, Finsupp.single_add]
+  rw [coeff, Finsupp.single_add]
```
grunweg, line 369: This change is equally readable; I would say it's fine. (If it's not faster, it might not be worth spending time *making* this change in other files, but I don't see this as negative.)

Mathlib/RingTheory/PowerSeries/Basic.lean
```diff
@@ -430,7 +430,7 @@ section
 
 theorem coeff_C_mul_X_pow (x : R) (k n : ℕ) :
     coeff R n (C R x * X ^ k : R⟦X⟧) = if n = k then x else 0 := by
-  simp [X_pow_eq, coeff_monomial]
+  rw [X_pow_eq, coeff_C_mul, coeff_monomial, mul_ite, mul_one, mul_zero]
```
grunweg, line 433: This feels less readable to me: without a reason why, I would prefer not to change this.

Mathlib/RingTheory/PowerSeries/Inverse.lean
```diff
@@ -77,8 +77,8 @@ theorem coeff_inv_aux (n : ℕ) (a : R) (φ : R⟦X⟧) :
     · rfl
     refine ⟨?_, fun hh ↦ H.not_le ?_⟩
     · rintro ⟨⟩
-      simpa [Finsupp.single_eq_same] using le_of_lt H
-    · simpa [Finsupp.single_eq_same] using hh ()
+      simpa only [Finsupp.single_eq_same] using le_of_lt H
```
grunweg, line 80: These changes seem fine to me - but again, knowing if they have a performance impact or not would be worthwhile.
(Doing a readability-neutral change for *no* reason might not be worth it.)

Mathlib/RingTheory/PowerSeries/Inverse.lean
```diff
@@ -368,10 +369,11 @@ instance : NormalizationMonoid k⟦X⟧ where
     exact ((eq_divided_by_X_pow_order_Iff_Unit h₀.ne_zero).mpr h₀).symm
 
 theorem normUnit_X : normUnit (X : PowerSeries k) = 1 := by
-  simp [normUnit, ← Units.val_eq_one, Unit_of_divided_by_X_pow_order_nonzero]
+  simp only [normUnit, inv_eq_one, ← Units.val_eq_one, ne_eq, X_ne_zero, not_false_eq_true,
```
grunweg, line 372: This feels significantly less readable - I'd only like to make this change with good justification.

rwst:
Thanks for info. The aim was to replace simp's but I understand now why some replacements may not be optimal. However, I think that even if a `simp only` is long, I like to see what lemmas were used, to get a feel of what is involved. So I don't think there are bad replacements.

grunweg (commented):
To give an example, here are comments on five changes you proposed. The others feel similar to me.

grunweg:
> Thanks for info. The aim was to replace simp's but I understand now why some replacements may not be optimal. However, I think that even if a `simp only` is long, I like to see what lemmas were used, to get a feel of what is involved. So I don't think there are bad replacements.

You can do the replacement locally! The question is what should get committed in mathlib.
My personal heuristic is that all `simp` lemmas are usually mathematically obvious enough - so any "interesting" steps will either be specified as explicit arguments or in separate lemmas. So, to me, I don't "lose" much by not all `simp` being explicit.

grunweg:
> By the way: did you hear about the `profiler` flag? You can locally profile a file and see which parts are slow. If a `simp` is pretty fast, replacing it by a longer/uglier `simp only` would be a net negative in my book. If a `simp` is very slow, but a `simp only` is much faster, the calculus might change. (This might still be an occasion to investigate `why` the simp is slow and if e.g. some simp lemmas are missing or sub-optimally stated: that is a harder task, though, which requires knowing the surrounding library a bit.)
> 
> [This PR](https://github.com/leanprover-community/leanprover-community.github.io/pull/472) contains some basic advice on how to use the profiler.

Let me add: there are *many* places in mathlib which could probably be sped up by local changes like this; if you're looking for these, help would certainly be welcome.

rwst:
Well, the results from the bot are just discouraging. I did not expect this. In future, I will only submit what shows clear improvements with `profiler`, and remains readable.

PR #13767, opened 2024-06-12
[Merged by Bors] - chore: drop porting notes about nonrec

joneugster:
---

There are actually two options here: either drop the `nonrec` and write out the namespace explicitely (as done now) or just drop the porting note and accept `nonrec theorem` as the preferred style. Opinions? 


Ruben-VandeVelde (approved):
I think this is fine.

maintainer merge

fpvandoorn:
I would have chosen to just delete the porting note, but this is also fine

bors merge

PR #13771, opened 2024-06-12
[Merged by Bors] - perf(TensorAlgebra/Grading): use suppress_compilation

grunweg:
The gradedAlgebra instance takes 10 seconds to compile (on CI).
Speeds up the whole file by 75%; no downstream definitions are affected.

-----------

Is that definition being computable useful?


grunweg:
I will benchmark this (and a few other tweaks) jointly in #13774.

YaelDillies:
Ask @eric-wieser, but I believe the answer is "No"

fpvandoorn:
I doubt that the computable version is useful, so let's merge this

bors merge

eric-wieser:
Any argument for making this computable would apply just as strongly to `TensorProduct`, so for now I agree `suppress_compilation` is the right choice.

PR #13776, opened 2024-06-12
chore(Data/ZMod/Defs): add `OfNat` instance for `ZMod`

MichaelStollBayreuth:
See [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/'simp'lifying.20'match'.20expressions/near/444269587) on Zulip.

---


MichaelStollBayreuth:
This leads to incompatible double instances and so should not be pursued further.

PR #13777, opened 2024-06-12
fix: normalize pathSeparators

adomani:
This PR makes the sorting of the module names that `lake exe mk_all` uniform across platforms -- or at least it tries to!

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/automatically.20generate.20ProjectName.2Elean/near/444257002)

---


Mathlib/Util/GetAllModules.lean
```diff
@@ -40,7 +40,11 @@ def getAllFiles (git : Bool) (ml : String) : IO (Array System.FilePath) := do
     else do
       let all ← walkDir ml
       return all.filter (·.extension == some "lean"))
-  let files := (allModules.erase ml.lean).qsort (·.toString < ·.toString)
+  -- sort after replacing the `pathSeparators` by `/`, to avoid sorting issues between `\` and `/`
+  let files := (allModules.erase ml.lean).qsort fun f1 f2 =>
```
grunweg, line 44: I believe the mathlib style guide suggests this.
```suggestion
  let files := (allModules.erase ml.lean).qsort fun f1 f2 ↦

File: Mathlib/Util/GetAllModules.lean [hunk 2]
```diff
@@ -40,7 +40,11 @@ def getAllFiles (git : Bool) (ml : String) : IO (Array System.FilePath) := do
...
+    let f1 := f1.toString.replace ("".push pathSeparator) "/"
```
grunweg, line 45: Is it sufficient to just replace the "ideal" path separator, or is there need to replace all path separators? (There is [System.FilePath.pathSeparators](https://leanprover-community.github.io/mathlib4_docs/Init/System/FilePath.html#System.FilePath.pathSeparators) also.) Perhaps the APIs called before ensure to only use the ideal one, so this code works --- in any case, this subtlety seems useful to document. Yes, I would also prefer for it to not exist in the first place.

grunweg (commented):
Thank you for doing this. This mostly looks good, I have two small comments.

Mathlib/Util/GetAllModules.lean
adomani: I am not really sure how to answer this: the code works on my computer (running Linux) and on CI (running again some Unix-based system).  I think that testing it on a Mac and on a Windows computer would be desirable, to make sure that the ordering there is the same as here.

grunweg: Agreed testing on other systems is worthwhile - perhaps ask on zulip after this PR lands?
That said: I think replacing all possible path separators (use `FilePath.pathSeparators`) by '/' is sufficient. (Optional, "the extra mile": ask on zulip of walkdir makes or wants to make a guarantee that all paths returned use the "ideal" path separator.)

adomani: I looped over the pathSeparators to replace them all by `/`.  I have a feeling that `pathSeparators` should be treated like more of an implementation detail, so that it is better to try to avoid relying on what it is.  Even in this case, the script is doing a blanket replace all `pathSeparators` (or, initially, the "ideal" one), trying to be agnostic to what it was supposed to be.

grunweg:
This morning, I had an idea how to do this differently: compare the list of paths components instead. All paths you're comparing have the same separators, so this yields sane results. (The sort order is slightly different, though...) I opened #13848 with the results, so we can compare.

grunweg:
And another approach: do not guarantee that `getAllFiles` results sorted results, but do the sorting in `getAllModules`: then, we can just sort strings (which don't have path separators, but dots), so all is well.
[This branch](https://github.com/leanprover-community/mathlib4/compare/MR-sort-modules-instead) has that approach.

grunweg:
Right now, I'm leaning towards the last approach: let me know what you think!
(That branch is not fully polished: if it were to land, I'd remove the sorting guarantee from `getAllFiles` and document it on `getAllModules` instead.)

adomani:
I'm not a big fan of having the shorter string be sorted last, so I would also opt for your second alternative.

grunweg:
Sounds good: I polished and submitted that branch as #13854.

grunweg:
With #13854 landed, this PR can be closed, right? (Thanks for investigating this, though!)

adomani:
Yes, I am closing it now!

PR #13779, opened 2024-06-12
[Merged by Bors] - chore: don't import `Lake` under `Mathlib`

mattrobball:
The `Lake.CLI.Main` import was slowing down linting by ~90s/~15% on its own. We move the function `getLeanLibs` back to `scripts/mk_all.lean` to avoid the import. 

---


mattrobball:
!bench

mattrobball:
!bench

mattrobball:
!bench

grunweg:
Thanks a lot for investigating this - I would have (and still am) completely stumped!

mattrobball:
Adding the file to `nolints.json` with all linters didn't make a difference in the regression

mattrobball:
Oh wait. I'm reading this wrong. The file itself is fine.

mattrobball:
!bench

mattrobball:
Until it is added to `Mathlib.lean`, and then even blacklisting all the linters in `nolints.json` doesn't help

mattrobball:
!bench

mattrobball:
!bench

mattrobball:
!bench

lakefile.lean
```diff
@@ -35,8 +35,6 @@ require importGraph from git "https://github.com/leanprover-community/import-gra
 @[default_target]
 lean_lib Mathlib
```
mattrobball, line 37:
```suggestion

--  NB. When adding further libraries, check if they should be excluded from `getLeanLibs` in
-- `Mathlib/Util/GetAllModules.lean`.
```

lakefile.lean
```diff
@@ -35,7 +35,7 @@ require importGraph from git "https://github.com/leanprover-community/import-gra
...
 
--- NB. When adding further libraries, check if they should be excluded from `getLeanLibs` in
+--  NB. When adding further libraries, check if they should be excluded from `getLeanLibs` in
```
mattrobball, line 38:
```suggestion
-- NB. When adding further libraries, check if they should be excluded from `getLeanLibs` in
```

grunweg (approved):
Wow, thanks a lot for finding this! As a next step, I'll augment the Python lint to complain about any lake imports in Mathlib, referring to this issue :-)

This PR is just code motion, so looks fine to me. My original use-case #13245 doesn't need `getLeanLibs` anyway (and its proposed replacement won't either), so moving this function back looks good to me.

grunweg:
Actually, since you marked this as awaiting-review and CI on this was successful previously, let me go straight to
maintainer merge

robertylewis:
bors merge

PR #10853, opened 2024-02-22
[Merged by Bors] - feat: Define `AnalyticManifold` as a charted space compatible with `analyticGroupoid`

girving:
1. Change the definition of `analyticGroupoid` to mean analytic everywhere including on the boundary, using the recently added `AnalyticWithinAt` def.
2. Define `AnalyticManifold` as a charted space compatible with `analyticGroupoid`. 

In finite dimensions `AnalyticManifold` is equivalent to `SmoothManifoldWithCorners`, so an alternative option would be to leave this out and prove the necessary correspondence via Osgood's theorem. However, we don't have Osgood's theorem yet, and it's nice to have a version that works arbitrarily.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13784, opened 2024-06-12
[Merged by Bors] - feat: add a few analytic function lemmas for future AnalyticManifold use

girving:
This is a few isolated lemmas split out of
https://github.com/leanprover-community/mathlib4/pull/10853.

1. AnalyticAt.comp_of_eq
2. FormalMultilinearSeries.zero_radius
3. constFormalMultilinearSeries_zero

---

### DIFF
--- a/Mathlib/Analysis/Analytic/Basic.lean
+++ b/Mathlib/Analysis/Analytic/Basic.lean
@@ -182,6 +182,11 @@ theorem constFormalMultilinearSeries_radius {v : F} :
     (by simp [constFormalMultilinearSeries])
 #align formal_multilinear_series.const_formal_multilinear_series_radius FormalMultilinearSeries.constFormalMultilinearSeries_radius
 
+/-- `0` has infinite radius of convergence -/
+@[simp] lemma zero_radius : (0 : FormalMultilinearSeries 𝕜 E F).radius = ∞ := by
+  rw [← constFormalMultilinearSeries_zero]
+  exact constFormalMultilinearSeries_radius
+
 /-- For `r` strictly smaller than the radius of `p`, then `‖pₙ‖ rⁿ` tends to zero exponentially:
 for some `0 < a < 1`, `‖p n‖ rⁿ = o(aⁿ)`. -/
 theorem isLittleO_of_lt_radius (h : ↑r < p.radius) :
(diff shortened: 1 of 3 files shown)


Mathlib/Analysis/Analytic/Basic.lean
```diff
@@ -182,6 +182,11 @@ theorem constFormalMultilinearSeries_radius {v : F} :
     (by simp [constFormalMultilinearSeries])
 #align formal_multilinear_series.const_formal_multilinear_series_radius FormalMultilinearSeries.constFormalMultilinearSeries_radius
 
+/-- `0` has infinite radius of convergence -/
```
grunweg, line 185: Does this generalise to any constant formal series?

Mathlib/Analysis/Calculus/FormalMultilinearSeries.lean
```diff
@@ -362,6 +362,17 @@ theorem constFormalMultilinearSeries_apply [NontriviallyNormedField 𝕜] [Norme
   Nat.casesOn n (fun hn => (hn rfl).elim) (fun _ _ => rfl) hn
 #align const_formal_multilinear_series_apply constFormalMultilinearSeries_apply
 
+@[simp]
```
grunweg, line 365: Can you generalise this to any `c : F` (or am I misunderstanding how `FormatMultilinearSeries` works)?

grunweg (commented):
Thanks for splitting this out. This looks good to me - re-reviewing, I just have one question about generalisation.

Mathlib/Analysis/Analytic/Basic.lean
girving: Not really, as this is the 0 of that type, not an injection from the field 0.

Mathlib/Analysis/Calculus/FormalMultilinearSeries.lean
girving: This is saying the injection from the field commutes with Zero, so I don’t think it sensibly generalizes.

girving:
Are there changes I should make?

grunweg:
No worries, I saw your comments. Right now, I don't have the time to wrap my head around your comments in detail - but what you say sounds reasonable, so let me
maintainer merge

sgouezel:
bors r+
Thanks!

PR #6386, opened 2023-08-05
[Merged by Bors] - feat: Add analytic structure groupoid on models with corners

michaellee94:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13785, opened 2024-06-12
[Merged by Bors] - refactor: move analyticGroupoid to a new file `AnalyticManifold.lean`

girving:
It used to be in `SmoothManifoldWithCorners.lean`, but this makes no sense as nothing downstream of that file uses `analyticGroupoid`. In #10853, I will rewrite the definition of `analyticGroupoid` and define `AnalyticManifold` on top of it.

---


Mathlib/Geometry/Manifold/AnalyticManifold.lean
```diff
@@ -0,0 +1,192 @@
+/-
+Copyright (c) 2024 Geoffrey Irving. All rights reserved.
```
grunweg, line 2: If I understand correctly, mathlib practice is to put the original author(s) of the code you're moving: in this case, using git blame to find out who wrote this code. Give a moment...

grunweg (commented):
Thank you for following up so quickly! This looks good to me (it's straightforward code motion); I like the new module docstring, a happy by-product is that `SmoothManifoldWithCorners` can shrink a little (that file is quite large anyway).

I just have one comment about the copyright header.

Mathlib/Geometry/Manifold/AnalyticManifold.lean
grunweg: Turns out this was written in mathlib4, in #6386. Please update the header accordingly.

girving: Oops, fixed. Apologies: I'd copied those lines from the broader commit, where most of the new file was mine.

grunweg:
Thanks for the swift response! I still think this looks good, so let's
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13786, opened 2024-06-12
[Merged by Bors] - feat(NumberTheory/EllipticDivisibilitySequence): extend even-odd recursion to integers

Multramate:
Provide a quick proof that normalised EDSs satisfy an even-odd recursion over all integers rather than just natural numbers, without relying on heavier machinery i.e. the full EDS recurrence in #13155, and make `Int.negInduction` slightly stronger.

---

- [x] depends on: #13399 


Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -256,34 +328,48 @@ lemma normEDS_even (m : ℕ) : normEDS b c d (2 * (m + 3)) * b =
...
+lemma normEDS_even (m : ℤ) : normEDS b c d (2 * m) * b =
+    normEDS b c d (m - 1) ^ 2 * normEDS b c d m * normEDS b c d (m + 2) -
+      normEDS b c d (m - 2) * normEDS b c d m * normEDS b c d (m + 1) ^ 2 := by
+  simp only [normEDS, preNormEDS_even, if_pos <| even_two_mul m, show m + 2 = m + 1 + 1 by ring1,
+    Int.even_add_one, show m - 2 = m - 1 - 1 by ring1, Int.even_sub_one, ite_not]
+  by_cases hm : Even m
+  · simp only [if_pos hm]
+    ring1
+  · simp only [if_neg hm]
+    ring1
```
alreadydone, line 350: One way to deduce these is using my #13057, which shows every normEDS is an elliptic divisibility sequence. Also see `evenRec / oddRec` in the PR.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -208,17 +245,52 @@ lemma preNormEDS_even (m : ℕ) : preNormEDS b c d (2 * (m + 3)) =
...
+  | neg h m =>
+    rcases m with _ | m; simp
+    simp_rw [Int.natCast_add, Nat.cast_one, show 2 * -(m + 1 : ℤ) + 1 = -(2 * m + 1) by rfl,
+      show -(m + 1 : ℤ) + 2 = -(m - 1) by omega, show -(m + 1 : ℤ) - 1 = -(m + 2) by rfl,
+      show -(m + 1 : ℤ) + 1 = -m by omega, preNormEDS_neg, even_neg, Int.even_add_one, ite_not, h m]
+    ring1
+
+lemma preNormEDS_even (m : ℤ) : preNormEDS b c d (2 * m) =
+    preNormEDS b c d (m - 1) ^ 2 * preNormEDS b c d m * preNormEDS b c d (m + 2) -
+      preNormEDS b c d (m - 2) * preNormEDS b c d m * preNormEDS b c d (m + 1) ^ 2 := by
```
alreadydone, line 266: Do you plan to use these somewhere?

Multramate: Not on top of my head. It might just be nice to have a standalone proof without using your #13057

Multramate:
They're resolved in the sense that I still intend to move forward with this PR (despite #13057), but they're not resolved in the sense that @alreadydone hasn't explicitly agreed to this.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -179,16 +179,16 @@ lemma preΨ'_three : W.preΨ' 3 = W.Ψ₃ :=
 lemma preΨ'_four : W.preΨ' 4 = W.preΨ₄ :=
   preNormEDS'_four ..
 
-lemma preΨ'_odd (m : ℕ) : W.preΨ' (2 * (m + 2) + 1) =
-    W.preΨ' (m + 4) * W.preΨ' (m + 2) ^ 3 * (if Even m then W.Ψ₂Sq ^ 2 else 1) -
-      W.preΨ' (m + 1) * W.preΨ' (m + 3) ^ 3 * (if Even m then 1 else W.Ψ₂Sq ^ 2) :=
-  preNormEDS'_odd ..
-
 lemma preΨ'_even (m : ℕ) : W.preΨ' (2 * (m + 3)) =
```
faenuccio, line 182: Can you please fix in the docstring that tilde seems now to be a prime?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -272,20 +282,30 @@ lemma ΨSq_three : W.ΨSq 3 = W.Ψ₃ ^ 2 := by
 lemma ΨSq_four : W.ΨSq 4 = W.preΨ₄ ^ 2 * W.Ψ₂Sq := by
   erw [ΨSq_ofNat, preΨ'_four, if_pos <| by decide]
 
-lemma ΨSq_odd (m : ℕ) : W.ΨSq (2 * (m + 2) + 1) =
-    (W.preΨ' (m + 4) * W.preΨ' (m + 2) ^ 3 * (if Even m then W.Ψ₂Sq ^ 2 else 1) -
-      W.preΨ' (m + 1) * W.preΨ' (m + 3) ^ 3 * (if Even m then 1 else W.Ψ₂Sq ^ 2)) ^ 2 := by
-  erw [ΨSq_ofNat, preΨ'_odd, if_neg (m + 2).not_even_two_mul_add_one, mul_one]
-
-lemma ΨSq_even (m : ℕ) : W.ΨSq (2 * (m + 3)) =
+lemma ΨSq_even' (m : ℕ) : W.ΨSq (2 * (m + 3)) =
```
faenuccio, line 285: Can you add some doc expaining why this theorem has a primed name?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -322,25 +342,42 @@ lemma Ψ_three : W.Ψ 3 = C W.Ψ₃ := by
...
-        (if Even m then W.preΨ' (m + 4) * W.preΨ' (m + 2) ^ 3
-          else -W.preΨ' (m + 1) * W.preΨ' (m + 3) ^ 3) := by
-  repeat erw [Ψ_ofNat]
-  simp_rw [preΨ'_odd, if_neg (m + 2).not_even_two_mul_add_one, Nat.even_add_one, ite_not]
-  split_ifs <;> C_simp <;> rw [C_Ψ₂Sq] <;> ring1
-
-lemma Ψ_even (m : ℕ) : W.Ψ (2 * (m + 3)) * W.ψ₂ =
+lemma Ψ_even' (m : ℕ) : W.Ψ (2 * (m + 3)) * W.ψ₂ =
     W.Ψ (m + 2) ^ 2 * W.Ψ (m + 3) * W.Ψ (m + 5) - W.Ψ (m + 1) * W.Ψ (m + 3) * W.Ψ (m + 4) ^ 2 := by
   repeat erw [Ψ_ofNat]
```
faenuccio, line 347: Why is `erw` needed? Is there perhaps a bit of API missing, making `rw` work smoothly?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -426,18 +463,26 @@ lemma ψ_three : W.ψ 3 = C W.Ψ₃ :=
 lemma ψ_four : W.ψ 4 = C W.preΨ₄ * W.ψ₂ :=
   normEDS_four ..
 
-lemma ψ_odd (m : ℕ) : W.ψ (2 * (m + 2) + 1) =
-    W.ψ (m + 4) * W.ψ (m + 2) ^ 3 - W.ψ (m + 1) * W.ψ (m + 3) ^ 3 :=
-  normEDS_odd ..
-
-lemma ψ_even (m : ℕ) : W.ψ (2 * (m + 3)) * W.ψ₂ =
+lemma ψ_even' (m : ℕ) : W.ψ (2 * (m + 3)) * W.ψ₂ =
```
faenuccio, line 466: Again, some doc explaining the reason for the `'`?

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -63,9 +62,13 @@ M Ward, *Memoir on Elliptic Divisibility Sequences*
...
+universe u v
 
-variable {R : Type u} {S : Type v} [CommRing R] [CommRing S] (W : ℤ → R) (f : R →+* S)
+variable {R : Type u} [CommRing R]
+
+section IsEllDivSequence
+
+variable (W : ℤ → R)
 
 /-- The proposition that a sequence indexed by integers is an elliptic sequence. -/
```
faenuccio, line 73: I'd rather say (here and at the top) that this is the definition of being an ES, rather than being a proposition... and the same comment applies to the several `def` below.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -102,6 +105,42 @@ lemma IsDivSequence.smul (h : IsDivSequence W) (x : R) : IsDivSequence (x • W)
...
 
+end IsEllDivSequence
+
+/-- Strong recursion principle for a normalised EDS: if we have
+ * `P 0`, `P 1`, `P 2`, `P 3`, and `P 4`,
+ * for all `m : ℕ` we can prove `P (2 * (m + 3))` from `P k` for all `k < 2 * (m + 3)`, and
+ * for all `m : ℕ` we can prove `P (2 * (m + 2) + 1)` from `P k` for all `k < 2 * (m + 2) + 1`,
+then we have `P n` for all `n : ℕ`. -/
+@[elab_as_elim]
+noncomputable def normEDSRec' {P : ℕ → Sort u}
```
faenuccio, line 116: Can you provide a bit of doc explaining the diff between these two `def`s and why this is primed/

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -240,53 +312,48 @@ lemma normEDS_three : normEDS b c d 3 = c := by
...
-
-lemma normEDS_even (m : ℕ) : normEDS b c d (2 * (m + 3)) * b =
+lemma normEDS_even' (m : ℕ) : normEDS b c d (2 * (m + 3)) * b =
     normEDS b c d (m + 2) ^ 2 * normEDS b c d (m + 3) * normEDS b c d (m + 5) -
       normEDS b c d (m + 1) * normEDS b c d (m + 3) * normEDS b c d (m + 4) ^ 2 := by
   repeat erw [normEDS_ofNat]
   simp only [preNormEDS'_even, if_pos <| even_two_mul _, Nat.even_add_one, ite_not]
   split_ifs <;> ring1
 
+lemma normEDS_odd' (m : ℕ) : normEDS b c d (2 * (m + 2) + 1) =
```
faenuccio, line 322: Again, why the prime?

faenuccio (commented):
Can you please update your doc by using unicode characters for all latex codes? Although MathJax will probably render them nicely in a browser, they're still pretty unreadable in VSCode, and if you use unicode things are much more readable. Thanks!

Multramate:
@faenuccio just to be clear, would we prefer something like this instead?
```
Let `W` be a Weierstrass curve over a commutative ring `R`. The sequence of `n`-division polynomials
`ψₙ ∈ R[X, Y]` of `W` is the normalised EDS with initial values
 * `ψ₀ := 0`,
 * `ψ₁ := 1`,
 * `ψ₂ := 2Y + a₁X + a₃`,
 * `ψ₃ := 3X⁴ + b₂X³ + 3b₄X² + 3b₆X + b₈`, and
 * `ψ₄ := ψ₂⬝(2X⁶ + b₂X⁵ + 5b₄X⁴ + 10b₆X³ + 10b₈X² + (b₂b₈ - b₄b₆)X + (b₄b₈ - b₆²))`.
```

faenuccio:
> @faenuccio just to be clear, would we prefer something like this instead?
> 
> ```
> Let `W` be a Weierstrass curve over a commutative ring `R`. The sequence of `n`-division polynomials
> `ψₙ ∈ R[X, Y]` of `W` is the normalised EDS with initial values
>  * `ψ₀ := 0`,
>  * `ψ₁ := 1`,
>  * `ψ₂ := 2Y + a₁X + a₃`,
>  * `ψ₃ := 3X⁴ + b₂X³ + 3b₄X² + 3b₆X + b₈`, and
>  * `ψ₄ := ψ₂⬝(2X⁶ + b₂X⁵ + 5b₄X⁴ + 10b₆X³ + 10b₈X² + (b₂b₈ - b₄b₆)X + (b₄b₈ - b₆²))`.
> ```

If you're asking with regard to the unicode business, then yes: I find this much more readable.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: The tilde correspond to the `pre`, which I have made slightly clearer now. The primed versions are just the unprimed versions but defined over N rather than over Z. The linter does not complain for `preΨ'` because there is already a docstring, and for the lemmas `preΨ'_*` because the primes don't appear at the end? What should I do here?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: I will append `_ofNat` to these.

faenuccio: Yes, overall having a descriptive naming (like `_ofNat`) seems more appropriate than primes.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
faenuccio: I think that the solution you propose below (appending `_ofNat`) is the best.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: Some of these are just for the sake of my sanity, e.g. converting 4 - 1 to 3 by `rw [show (4 - 1 : ℤ) = 3 by rfl]` etc.

I can add these to the bottom of [this file](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Cast/Defs.html) though, what do you think?

Multramate: Wait, I have to find a different home for those lemmas, since I don't think file imports `AddGroup`...

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: I can't call them `ofNat`, because that's the name of a lemma below...

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
Multramate: The only difference is N vs Z as above, which is already mentioned in the docstring.

Again, I can't use `ofNat` because it's already the name of a lemma below.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
Multramate: I've changed this to `ofNat`.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
Multramate: These docstrings are unrelated to this PR. I wouldn't worry about them at this point, as there are downstream PRs (mostly by Junyan) that significantly modify the structure of this file.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -14,10 +14,10 @@ This file defines the type of an elliptic divisibility sequence (EDS) and a few
...
 ## Mathematical background
 
-Let $R$ be a commutative ring. An elliptic sequence is a sequence $W : \mathbb{Z} \to R$ satisfying
-$$ W(m + n)W(m - n)W(r)^2 = W(m + r)W(m - r)W(n)^2 - W(n + r)W(n - r)W(m)^2, $$
-for any $m, n, r \in \mathbb{Z}$. A divisibility sequence is a sequence $W : \mathbb{Z} \to R$
-satisfying $W(m) \mid W(n)$ for any $m, n \in \mathbb{Z}$ such that $m \mid n$.
+Let `R` be a commutative ring. An elliptic sequence is a sequence `W : ℤ → R` satisfying
+`W(m + n)W(m - n)W(r)² = W(m + r)W(m - r)W(n)² - W(n + r)W(n - r)W(m)²`
+for any `m, n, r ∈ ℤ`. A divisibility sequence is a sequence `W : ℤ → R` satisfying `W(m) ∣ W(n)`
+for any `m, n ∈ ℤ` such that `m ∣ n`.
```
faenuccio, line 20: And what is an "elliptic divisibility sequence"? One that is simultaneously elliptic and a divisibility one?

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
faenuccio: OK, but then please come up with a name so that `Int` and `Nat` show up.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -191,22 +229,64 @@ lemma preNormEDS_three : preNormEDS b c d 3 = c := by
...
 @[simp]
 lemma preNormEDS_neg (n : ℤ) : preNormEDS b c d (-n) = -preNormEDS b c d n := by
   rw [preNormEDS, Int.sign_neg, Int.cast_neg, neg_mul, Int.natAbs_neg, preNormEDS]
 
+lemma preNormEDS_even (m : ℤ) : preNormEDS b c d (2 * m) =
+    preNormEDS b c d (m - 1) ^ 2 * preNormEDS b c d m * preNormEDS b c d (m + 2) -
+      preNormEDS b c d (m - 2) * preNormEDS b c d m * preNormEDS b c d (m + 1) ^ 2 := by
+  induction m using Int.negInduction with
+  | nat m =>
+    rcases m with _ | _ | _ | m;
```
faenuccio, line 253:
```suggestion
    rcases m with _ | _ | _ | m
```
These colum is useless

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -191,22 +229,64 @@ lemma preNormEDS_three : preNormEDS b c d 3 = c := by
...
+    · simp
+    · simp
+    · simp only [Int.natCast_add, Nat.cast_one]
+      rw [Int.add_sub_cancel, show (m : ℤ) + 1 + 1 + 1 = m + 1 + 2 by rfl, Int.add_sub_cancel]
+      exact preNormEDS_even_ofNat ..
+  | neg h m =>
+    simp_rw [show 2 * -(m : ℤ) = -(2 * m) by omega, show -(m : ℤ) - 1 = -(m + 1) by omega,
+      show -(m : ℤ) + 2 = -(m - 2) by omega, show -(m : ℤ) - 2 = -(m + 2) by omega,
+      show -(m : ℤ) + 1 = -(m - 1) by omega, preNormEDS_neg, h m]
+    ring1
```
faenuccio, line 264:
```suggestion
    ring
```
`ring` works as well

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -239,53 +319,48 @@ lemma normEDS_three : normEDS b c d 3 = c := by
...
+lemma normEDS_odd_ofNat (m : ℕ) : normEDS b c d (2 * (m + 2) + 1) =
+    normEDS b c d (m + 4) * normEDS b c d (m + 2) ^ 3 -
+      normEDS b c d (m + 1) * normEDS b c d (m + 3) ^ 3 := by
+  repeat erw [normEDS_ofNat]
+  simp_rw [preNormEDS'_odd, if_neg (m + 2).not_even_two_mul_add_one, Nat.even_add_one, ite_not]
+  split_ifs <;> ring1
+
 @[simp]
 lemma normEDS_neg (n : ℤ) : normEDS b c d (-n) = -normEDS b c d n := by
   simp only [normEDS, preNormEDS_neg, neg_mul, even_neg]
```
faenuccio, line 339: Why are you removing all these recursion principles?

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -239,53 +319,48 @@ lemma normEDS_three : normEDS b c d 3 = c := by
...
-  normEDSRec' zero one two three four
-    (fun _ ih => by apply even <;> exact ih _ <| by linarith only)
-    (fun _ ih => by apply odd <;> exact ih _ <| by linarith only) n
+lemma normEDS_odd (m : ℤ) : normEDS b c d (2 * m + 1) =
+    normEDS b c d (m + 2) * normEDS b c d m ^ 3 -
+      normEDS b c d (m - 1) * normEDS b c d (m + 1) ^ 3 := by
+  simp only [normEDS, preNormEDS_odd, if_neg m.not_even_two_mul_add_one]
+  conv_lhs => rw [← @one_pow R _ 4, ← ite_pow, ← ite_pow]
+  simp only [show m + 2 = m + 1 + 1 by ring1, Int.even_add_one, Int.even_sub_one, ite_not]
+  ring1
```
faenuccio, line 357:
```suggestion
  simp only [show m + 2 = m + 1 + 1 by ring, Int.even_add_one, Int.even_sub_one, ite_not]
  ring
```
Ditto

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
```diff
@@ -239,53 +319,48 @@ lemma normEDS_three : normEDS b c d 3 = c := by
...
+lemma normEDS_even (m : ℤ) : normEDS b c d (2 * m) * b =
+    normEDS b c d (m - 1) ^ 2 * normEDS b c d m * normEDS b c d (m + 2) -
+      normEDS b c d (m - 2) * normEDS b c d m * normEDS b c d (m + 1) ^ 2 := by
+  simp only [normEDS, preNormEDS_even, if_pos <| even_two_mul m, show m + 2 = m + 1 + 1 by ring1,
+    Int.even_add_one, show m - 2 = m - 1 - 1 by ring1, Int.even_sub_one, ite_not]
+  by_cases hm : Even m
+  · simp only [if_pos hm]
+    ring1
+  · simp only [if_neg hm]
+    ring1
```
faenuccio, line 349:
```suggestion
  simp only [normEDS, preNormEDS_even, if_pos <| even_two_mul m, show m + 2 = m + 1 + 1 by ring,
    Int.even_add_one, show m - 2 = m - 1 - 1 by ring, Int.even_sub_one, ite_not]
  by_cases hm : Even m
  · simp only [if_pos hm]
    ring
  · simp only [if_neg hm]
    ring
```
Can you replace all `ring1` by `ring`?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -15,75 +15,70 @@ These are defined in terms of the auxiliary sequences for normalised elliptic di
...
+ * `ψ₃ := 3X⁴ + b₂X³ + 3b₄X² + 3b₆X + b₈`, and
+ * `ψ₄ := ψ₂ ⬝ (2X⁶ + b₂X⁵ + 5b₄X⁴ + 10b₆X³ + 10b₈X² + (b₂b₈ - b₄b₆)X + (b₄b₈ - b₆²))`.
+
+Furthermore, define the associated sequences `φₙ, ωₙ ∈ R[X, Y]` by
+ * `φₙ := Xψₙ² - ψₙ₊₁ ⬝ ψₙ₋₁`, and
+ * `ωₙ := (ψ₂ₙ / ψₙ - ψₙ ⬝ (a₁φₙ + a₃ψₙ²)) / 2`.
+
+Note that `ωₙ` is always well-defined as a polynomial in `R[X, Y]`. As a start, it can be shown by
+induction that `ψₙ` always divides `ψ₂ₙ` in `R[X, Y]`, so that `ψ₂ₙ / ψₙ` is always well-defined as
+a polynomial, while division by 2 is well-defined when `R` has characteristic different from 2. In
```
faenuccio, line 32:
```suggestion
a polynomial, while division by `2` is well-defined when `R` has characteristic different from `2`. In
```

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -15,75 +15,70 @@ These are defined in terms of the auxiliary sequences for normalised elliptic di
...
+general, it can be shown that 2 always divides the polynomial `ψ₂ₙ / ψₙ - ψₙ ⬝ (a₁φₙ + a₃ψₙ²)` in
```
faenuccio, line 33:
```suggestion
general, it can be shown that `2` always divides the polynomial `ψ₂ₙ / ψₙ - ψₙ ⬝ (a₁φₙ + a₃ψₙ²)` in
```

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -15,75 +15,70 @@ These are defined in terms of the auxiliary sequences for normalised elliptic di
...
+the characteristic 0 universal ring `𝓡[X, Y] := ℤ[A₁, A₂, A₃, A₄, A₆][X, Y]` of `W`, where the `Aᵢ`
```
faenuccio, line 34:
```suggestion
the characteristic `0` universal ring `𝓡[X, Y] := ℤ[A₁, A₂, A₃, A₄, A₆][X, Y]` of `W`, where the `Aᵢ`
```

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -15,75 +15,70 @@ These are defined in terms of the auxiliary sequences for normalised elliptic di
...
 Analogously to `Mathlib.NumberTheory.EllipticDivisibilitySequence`, the bivariate polynomials
-$\Psi_n$ are defined in terms of the univariate polynomials $\tilde{\Psi}_n$. This is done partially
-to avoid ring division, but more crucially to allow the definition of $\Psi_n^{[2]}$ and $\Phi_n$ as
-univariate polynomials without needing to work under the coordinate ring, and to allow the
-computation of their leading terms without ambiguity. Furthermore, evaluating these polynomials at a
-rational point on $W$ recovers their original definition up to linear combinations of the
-Weierstrass equation of $W$, hence also avoiding the need to work in the coordinate ring.
+`Ψₙ` are defined in terms of the univariate polynomials `preΨₙ`. This is done partially to avoid
+ring division, but more crucially to allow the definition of `ΨSqₙ` and `Φₙ` as univariate
+polynomials without needing to work under the coordinate ring, and to allow the computation of their
```
faenuccio, line 76: What do you mean by "working under the coordinate ring"?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
faenuccio: Well, can you come up with another name indicating that it is `Nat`-related?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
```diff
@@ -304,43 +319,60 @@ lemma Ψ_ofNat (n : ℕ) : W.Ψ n = C (W.preΨ' n) * if Even n then W.ψ₂ else
...
-  simp_rw [preΨ'_odd, if_neg (m + 2).not_even_two_mul_add_one, Nat.even_add_one, ite_not]
-  split_ifs <;> C_simp <;> rw [C_Ψ₂Sq] <;> ring1
-
-lemma Ψ_even (m : ℕ) : W.Ψ (2 * (m + 3)) * W.ψ₂ =
+lemma Ψ_even_ofNat (m : ℕ) : W.Ψ (2 * (m + 3)) * W.ψ₂ =
     W.Ψ (m + 2) ^ 2 * W.Ψ (m + 3) * W.Ψ (m + 5) - W.Ψ (m + 1) * W.Ψ (m + 3) * W.Ψ (m + 4) ^ 2 := by
-  repeat erw [Ψ_ofNat]
+  repeat rw_mod_cast [Ψ_ofNat]
   simp_rw [preΨ'_even, if_pos <| even_two_mul _, Nat.even_add_one, ite_not]
   split_ifs <;> C_simp <;> ring1
```
faenuccio, line 344:
```suggestion
  split_ifs <;> C_simp <;> ring
```
Can you check the other occurrences of `ring1` and leave `ring`? The first is useless here and a bit too strong for what is needed.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Degree.lean
```diff
@@ -14,29 +14,29 @@ Weierstrass curves defined in `Mathlib.AlgebraicGeometry.EllipticCurve.DivisionP
...
- * `WeierstrassCurve.leadingCoeff_preΨ`: the leading coefficient of $\tilde{\Psi}_n$ when $n \ne 0$.
- * `WeierstrassCurve.natDegree_ΨSq_le`: the expected degree of $\Psi_n^{[2]}$.
- * `WeierstrassCurve.coeff_ΨSq`: the expected leading coefficient of $\Psi_n^{[2]}$.
- * `WeierstrassCurve.natDegree_ΨSq`: the degree of $\Psi_n^{[2]}$ when $n \ne 0$.
- * `WeierstrassCurve.leadingCoeff_ΨSq`: the leading coefficient of $\Psi_n^{[2]}$ when $n \ne 0$.
- * `WeierstrassCurve.natDegree_Φ_le`: the expected degree of $\Phi_n$.
- * `WeierstrassCurve.coeff_Φ`: the expected leading coefficient of $\Phi_n$.
- * `WeierstrassCurve.natDegree_Φ`: the degree of $\Phi_n$ when $n \ne 0$.
- * `WeierstrassCurve.leadingCoeff_Φ`: the leading coefficient of $\Phi_n$ when $n \ne 0$.
+ * `WeierstrassCurve.natDegree_preΨ_le`: the expected degree of `preΨₙ`.
```
faenuccio, line 28: In the following lines of doc, instead of repeatedly saying "expected", can you say what the reult says?

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: I don't understand what you mean by `ring1` being useless/strong here. I believe `ring` is just `ring1 <|> ring_nf`, and I always prefer `ring1` because I *know* I won't need to get to `ring_nf`. There are examples of goals (which I cannot replicate right now) that are solved by `ring_nf` but not by `ring1`. I think something like this is volatile code, so closing goals with `ring1` means that I will never need to consider this scenario (i.e. it closes the goal iff it is solved by both `ring1` and `ring_nf`). In any case this is a thing I did in all the elliptic curve files, so this refactor should be done globally if at all, and hence would be outside the scope of this feat PR.

This applies to your other `ring`-related comments too.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
Multramate: I mean that they are actually bivariate polynomials, but if we quotient out by the equation of the curve (hence working in the coordinate ring rather than the full polynomial ring), there are univariate representatives that are equivalent to them. On the other hand, I can't actually define these univariate representatives in terms of `Ψₙ` because it's already bivariate, so I have to first define the univariate `preΨₙ` to define `ΨSqₙ` and `Φₙ`.

This is one of the things somewhat awkward to explain in docstrings (I wish we have the paper ready!).

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
Multramate: Yeah, I can add this.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
faenuccio: OK. I see what you mean , no problem.

Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/Basic.lean
faenuccio: OK, thanks.

Mathlib/NumberTheory/EllipticDivisibilitySequence.lean
faenuccio: Forget it, a problem with my VSCode.

faenuccio:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13793, opened 2024-06-13
feat: syntax linter for deprecated without since

adomani:
---


grunweg:
Coming here from looking at the list of open PRs: is this PR still useful? With the `env_linter` for this, it would seem to me that this PR can be closed.

PR #13797, opened 2024-06-13
chore(Algebra/Group/InjSurj): replace `@[reducible]` with `abbrev`

astrainfinita:
~~It seems `addDecl` does not work with `Lean.ReducibilityHints` correctly, so we cannot just use `to_additive`.~~ No, it works fine. But we need to add `@[reducible, inline]` ourselves.

---


negiizhao:
!bench

PR #13798, opened 2024-06-13
[Merged by Bors] - feat(Mathlib.NumberTheory.Cyclotomic.Three): add various results

riccardobrasca:
We add various results about the third cyclotomic field needed to prove flt3.

From the flt3 project at LFTCM2024.

---


Mathlib/Data/ZMod/Basic.lean
```diff
@@ -491,6 +491,13 @@ noncomputable def ringEquiv [Fintype R] (h : Fintype.card R = n) : ZMod n ≃+*
   RingEquiv.ofBijective _ (ZMod.castHom_bijective R h)
 #align zmod.ring_equiv ZMod.ringEquiv
 
+/-- The unique ring isomorphism between `ZMod p` and a ring `R` cardinality a prime `p`. -/
+noncomputable def ringEquivOfPrime [Fintype R] {p : ℕ} (hp : p.Prime) (hR : Fintype.card R = p) :
+    ZMod p ≃+* R :=
+  have : Nontrivial R := Fintype.one_lt_card_iff_nontrivial.1 (hR ▸ hp.one_lt)
+  have : CharP R p := (CharP.charP_iff_prime_eq_zero hp).2 (hR ▸ Nat.cast_card_eq_zero R)
```
TwoFX, line 498: Should `p.Prime -> Fintype.card R = p -> CharP R p` be its own lemma?

riccardobrasca: Why not, thanks!

riccardobrasca: I've added it to `Mathlib/Algebra/CharP/CharAndCard.lean`, that is not imported by this file (it seems a better place to have it). Rather then messing up the imports I think we can just keep this proof as it is.

Mathlib/Algebra/CharP/CharAndCard.lean
```diff
@@ -82,3 +82,8 @@ theorem not_isUnit_prime_of_dvd_card {R : Type*} [CommRing R] [Fintype R] (p : 
   mt (isUnit_iff_not_dvd_char R p).mp
     (Classical.not_not.mpr ((prime_dvd_char_iff_dvd_card p).mpr hp))
 #align not_is_unit_prime_of_dvd_card not_isUnit_prime_of_dvd_card
+
+lemma charP_of_card_eq_prime {R : Type*} [NonAssocRing R] [Fintype R] {p : ℕ} (hp : p.Prime)
```
TwoFX, line 86: The other lemmas in this file take `[Fact p.Prime]`, should this one do the same?

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -491,6 +491,14 @@ noncomputable def ringEquiv [Fintype R] (h : Fintype.card R = n) : ZMod n ≃+*
   RingEquiv.ofBijective _ (ZMod.castHom_bijective R h)
 #align zmod.ring_equiv ZMod.ringEquiv
 
+/-- The unique ring isomorphism between `ZMod p` and a ring `R` of cardinality a prime `p`. -/
```
faenuccio, line 494: Would it be possible to add the uniqueness statement, so that if `e` is an equivalence between two such rings then `e=ringEquivOfPrime`?

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
     have : (hζ.pow_of_coprime 2 (by decide)).toInteger = hζ.toInteger ^ 2 := by ext; simp
     simp only [this, PNat.val_ofNat, Nat.cast_ofNat, mul_neg, Int.cast_neg, ← neg_add, ←
       sub_eq_iff_eq_add.1 hx, Units.val_neg, val_pow_eq_pow_val, IsUnit.unit_spec, neg_neg]
+
+variable (x : 𝓞 K)
+
+/-- Let `(x : 𝓞 K)`. Then we have that `λ` divides one amongst `x`, `x - 1` and `x + 1`. -/
+lemma dvd_or_dvd_sub_one_or_dvd_add_one : λ ∣ x ∨ λ ∣ x - 1 ∨ λ ∣ x + 1 := by
```
faenuccio, line 120: Why not adding a `lambda` in front of the first `dvd`? Also, shouldn't this be also `private`?

Mathlib/NumberTheory/Cyclotomic/Rat.lean
```diff
@@ -182,6 +182,26 @@ noncomputable def integralPowerBasis [IsCyclotomicExtension {p ^ k} ℚ K]
...
+  refine Ideal.fintypeQuotientOfFreeOfNeBot _ (fun h ↦ ?_)
+  simp only [Ideal.span_singleton_eq_bot, sub_eq_zero, ← Subtype.coe_inj] at h
+  exact hζ.ne_one hk (RingOfIntegers.ext_iff.1 h)
+
+/-- We have that `𝓞 K ⧸ Ideal.span {ζ - 1}` has cardinality equal to the norm of `ζ - 1`.
+
+See the results below to compute this norm in various cases. -/
+lemma card_quotient_toInteger_sub_one [NumberField K] {k : ℕ+} (hk : 1 < k)
+    (hζ : IsPrimitiveRoot ζ k) :
+    letI _ := hζ.fintypeQuotienttoIntegerSubOne hk
```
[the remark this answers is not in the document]
riccardobrasca: Ah yes, this is a good idea, to avoid the `Fintype` instance. Thanks!

Mathlib/NumberTheory/Cyclotomic/Three.lean
riccardobrasca: I fixed the name. It cannot be `private` since it will be needed in the proof of flt3, in another file (I will make sure to make "public" only the results I really need in the other file, maybe only this one).

faenuccio: But if you make it private and then prepend the namespace, wouldn't this work?

riccardobrasca: Are you thinking about `protected`? AFAIK the purpose of `private` is to make lemmas "invisible" in other files (or sections, or whatever).

faenuccio: Oh right, I had `protected` in mind, and I have the same mental image of `private` as you have, so I finally agree that this should not be `private` (but hopefully in a namespace - I am not with VSCode now and cannot check).

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+/-- We have that `η ^ 2 + η + 1 = 0`. -/
+lemma eta_eq_add_eta_add_one : (η : 𝓞 K) ^ 2 + η + 1 = 0 := by
+  ext; simpa using hζ.isRoot_cyclotomic (by decide)
+
+/-- We have that `x ^ 3 - 1 = (x - 1) * (x - η) * (x - η ^ 2)`. -/
+lemma cube_sub_one_eq_mul : x ^ 3 - 1 = (x - 1) * (x - η) * (x - η ^ 2) := by
+  symm
+  calc _ = x ^ 3 - x ^ 2 * (η ^ 2 + η + 1) + x * (η ^ 2 + η + η ^ 3) - η ^ 3 := by ring
+  _ = x ^ 3 - x ^ 2 * (η ^ 2 + η + 1) + x * (η ^ 2 + η + 1) - 1 := by
+    simp [show hζ.toInteger ^ 3 = 1 from hζ.toInteger_isPrimitiveRoot.pow_eq_one]
```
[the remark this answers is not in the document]
riccardobrasca: Because `hζ.toInteger_isPrimitiveRoot` says that `hζ.toInteger ^ ↑3 = 1`, where `3 : ℕ+` but in the goal there is a `3 : ℕ`. The general statement of `toInteger_isPrimitiveRoot` is I think correct (positivity is needed in the definition of `toInteger`).

Mathlib/NumberTheory/Cyclotomic/Three.lean
riccardobrasca: Yes, everything is in a (very specific) namespace.

Mathlib/Data/ZMod/Basic.lean
riccardobrasca: Rather than duplicating all the results about `ZMod.ringEquiv` (uniqueness, reflexivity, transitivity etc), I've added a comment in the definition of `ringEquivOfPrime` to explain how to turn it into `ZMod.ringEquiv` (the results for it are already there).

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+      exact Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 h
+    · right; left
+      refine Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 ?_
+      rw [RingHom.map_sub, h, RingHom.map_one, sub_self]
+    · right; right
+      refine Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 ?_
+      rw [RingHom.map_add, h, RingHom.map_one, add_left_neg]
+  · rw [← Nat.card_eq_fintype_card, hζ.card_quotient_toInteger_sub_one (by decide),
+      hζ.norm_toInteger_sub_one_of_prime_ne_two' (by decide)]
+    simp
```
MichaelStollBayreuth, line 138:
```suggestion
  let _ : Ring (𝓞 K ⧸ Ideal.span {λ}) := CommRing.toRing -- to speed up instance synthesis
  let _ : AddGroup (𝓞 K ⧸ Ideal.span {λ}) := AddGroupWithOne.toAddGroup -- dito
  have := Finset.mem_univ (Ideal.Quotient.mk (Ideal.span {λ}) x)
  have h3 : Fintype.card (𝓞 K ⧸ Ideal.span {λ}) = 3 := by
    rw [← Nat.card_eq_fintype_card, hζ.card_quotient_toInteger_sub_one (by decide),
      hζ.norm_toInteger_sub_one_of_prime_ne_two' (by decide)]
    simp only [PNat.val_ofNat, Nat.cast_ofNat, Int.reduceAbs]
  rw [Finset.univ_of_card_eq_three h3] at this
  simp only [Finset.mem_insert, Finset.mem_singleton] at this
  rcases this with (h | h | h)
  · left
    exact Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 h
  · right; left
    refine Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 ?_
    rw [RingHom.map_sub, h, RingHom.map_one, sub_self]
  · right; right
    refine Ideal.mem_span_singleton.1 <| Ideal.Quotient.eq_zero_iff_mem.1 ?_
    rw [RingHom.map_add, h, RingHom.map_one, add_left_neg]
```
This one was quite slow. Adding the two explicit instances makes it reasonably fast.

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+
+/-- We have that `η ^ 2 + η + 1 = 0`. -/
+lemma eta_sq_add_eta_add_one : (η : 𝓞 K) ^ 2 + η + 1 = 0 := by
+  ext; simpa using hζ.isRoot_cyclotomic (by decide)
```
MichaelStollBayreuth, line 142:
```suggestion
  rw [eta_sq]
  ring
```
The private lemma `eta_sq` must be good for something...
Maybe remove the private lemma instead; it does not seem to be used.

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+
+/-- If `λ` divides `x - 1`, then `λ ^ 4` divides `x ^ 3 - 1`. -/
+lemma lambda_pow_four_dvd_cube_sub_one_of_dvd_sub_one {x : 𝓞 K} (h : λ ∣ x - 1) :
+    λ ^ 4 ∣ x ^ 3 - 1 := by
+  obtain ⟨y, hy⟩ := h
+  have : x ^ 3 - 1 = λ ^ 3 * (y * (y - 1) * (y - (η + 1))) := by
+    calc _ =  (x - 1) * (x - 1 - λ) * (x - 1 - λ * (η + 1)) := by
+          simp only [coe_eta, cube_sub_one_eq_mul hζ x]; ring
+    _ = _ := by rw [hy]; ring
+  rw [this, show λ ^ 4 = λ ^ 3 * λ by ring]
```
MichaelStollBayreuth, line 169:
```suggestion
  rw [this, pow_succ]
```

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+  exact mul_dvd_mul_left _ (lambda_dvd_mul_sub_one_mul_sub_eta_add_one hζ y)
+
+/-- If `λ` divides `x + 1`, then `λ ^ 4` divides `x ^ 3 + 1`. -/
+lemma lambda_pow_four_dvd_cube_add_one_of_dvd_add_one {x : 𝓞 K} (h : λ ∣ x + 1) :
+    λ ^ 4 ∣ x ^ 3 + 1 := by
+  replace h : λ ∣ -x - 1 := by
+    convert h.neg_right using 1
+    ring
```
MichaelStollBayreuth, line 177:
```suggestion
    exact (neg_add' x 1).symm
```

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+  obtain ⟨y, hy⟩ := lambda_pow_four_dvd_cube_sub_one_of_dvd_sub_one hζ h
+  refine ⟨-y, ?_⟩
+  rw [mul_neg, ← hy]
```
MichaelStollBayreuth, line 180:
```suggestion
  convert (lambda_pow_four_dvd_cube_sub_one_of_dvd_sub_one hζ h).neg_right using 1
```

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+
+/-- Let `(x : 𝓞 K)`. Then we have that `λ` divides one amongst `x`, `x - 1` and `x + 1`. -/
+lemma lambda_dvd_or_dvd_sub_one_or_dvd_add_one : λ ∣ x ∨ λ ∣ x - 1 ∨ λ ∣ x + 1 := by
+  classical
+  have := hζ.finite_quotient_toInteger_sub_one (by decide)
+  let _ := Fintype.ofFinite (𝓞 K ⧸ Ideal.span {λ})
+  have := Finset.mem_univ (Ideal.Quotient.mk (Ideal.span {λ}) x)
+  rw [Finset.univ_of_card_eq_three] at this
+  · simp only [Finset.mem_insert, Finset.mem_singleton] at this
+    rcases this with (h | h | h)
```
MichaelStollBayreuth, line 127: The parens around `h | h | h` are not necessary.

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,79 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+    ring
+  obtain ⟨y, hy⟩ := lambda_pow_four_dvd_cube_sub_one_of_dvd_sub_one hζ h
+  refine ⟨-y, ?_⟩
+  rw [mul_neg, ← hy]
+  ring
+
+/-- If `λ` does not divide `x`, then `λ ^ 4` divides `x ^ 3 - 1` or `x ^ 3 + 1`. -/
+lemma lambda_pow_four_dvd_cube_sub_one_or_add_one_of_lambda_not_dvd {x : 𝓞 K} (h : ¬ λ ∣ x) :
+    λ ^ 4 ∣ x ^ 3 - 1 ∨ λ ^ 4 ∣ x ^ 3 + 1 := by
+  rcases lambda_dvd_or_dvd_sub_one_or_dvd_add_one hζ x with (H | H | H)
```
MichaelStollBayreuth, line 186: The parens around `H | H | H` are not necessary.

Mathlib/NumberTheory/Cyclotomic/Three.lean
faenuccio: Is it really an advantage?

MichaelStollBayreuth: I don't insist, but my feeling is that `ring` is overkill when it is just one simple lemma.

faenuccio: I see your point, and most probably agree with you. In general I find that `ring` has the advantage of being less affected by library changes, but if it is just one lemma it looks like an overkill.

Ruben-VandeVelde (approved):
Lgtm, thanks! 

maintainer merge

Mathlib/RingTheory/Fintype.lean
```diff
@@ -12,6 +13,43 @@ import Mathlib.Data.Fintype.Units
...
+  have : Nontrivial R := by
+    refine Fintype.one_lt_card_iff_nontrivial.1 ?_
+    rw [h]
+    norm_num
+  refine (eq_of_subset_of_card_le (subset_univ _) ?_).symm
+  rw [card_univ, h, card_insert_of_not_mem, card_singleton]
+  rw [mem_singleton]
+  exact zero_ne_one
+
+lemma Finset.univ_of_card_eq_three (h : Fintype.card R = 3) :
```
jcommelin, line 31: Not sure if it will ever be useful... but I guess you can weaken the `= 3` to `\le 3`. Idem for the `= 2` above.

riccardobrasca: Why not!

Mathlib/NumberTheory/Cyclotomic/Three.lean
```diff
@@ -109,3 +113,81 @@ theorem eq_one_or_neg_one_of_unit_of_congruent (hcong : ∃ n : ℤ, λ ^ 2 ∣
...
+  _ = x ^ 3 - 1 := by rw [eta_sq_add_eta_add_one hζ]; ring
+
+/-- We have that `λ` divides `x * (x - 1) * (x - (η + 1))`. -/
+lemma lambda_dvd_mul_sub_one_mul_sub_eta_add_one : λ ∣ x * (x - 1) * (x - (η + 1)) := by
+  rcases lambda_dvd_or_dvd_sub_one_or_dvd_add_one hζ x with h | h | h
+  · exact dvd_mul_of_dvd_left (dvd_mul_of_dvd_left h _) _
+  · exact dvd_mul_of_dvd_left (dvd_mul_of_dvd_right h _) _
+  · refine dvd_mul_of_dvd_right ?_ _
+    rw [show x - (η + 1) = x + 1 - (η - 1 + 3) by ring]
+    exact dvd_sub h <| dvd_add dvd_rfl hζ.toInteger_sub_one_dvd_prime'
```
Vierkantor, line 163: More note to self than anything to do in this PR: I feel like this should be `aesop`able...

Vierkantor (approved):
LGTM with Johan's comment, thanks!

bors d+

riccardobrasca:
bors merge

PR #13800, opened 2024-06-13
[Merged by Bors] - chore: refactor LinarithConfig object so defaults are easier to find

robertylewis:
`LinarithConfig` was defined before `CertificateOracle` was instantiated, so it couldn't have a proper default. The default was chosen inline in the middle of a long function. Now that we have multiple oracles, it's important to make the default choice visible!

---

Addresses a comment at https://github.com/leanprover-community/mathlib4/pull/12819#issuecomment-2165185245


grunweg (commented):
I don't know the linarith code, but this seems straightforward: most of this is code motion; the rest seems easy enough. Please ask a more expert reviewer if you think this would be useful (I have no idea who that could be, though)!
maintainer merge

robertylewis:
Fixed. But this may have minor conflicts with #12819 which is now on the queue, so let's hold off on bors until that lands.

PatrickMassot:
bors merge

PatrickMassot:
bors merge

PR #13804, opened 2024-06-13
[Merged by Bors] - chore: fix dsimp lemmas that cause loops

jcommelin:
These dsimp lemmas had a proof subterm in their RHS that contained a
copy of the LHS, which causes dsimp to enter an infinite loop.

Nb: dsimp visits all sub-terms, even proofs: it is used to eliminate (artificial) dependencies

---

### DIFF
--- a/Mathlib/Algebra/Algebra/Equiv.lean
+++ b/Mathlib/Algebra/Algebra/Equiv.lean
@@ -360,14 +360,26 @@ theorem mk_coe' (e : A₁ ≃ₐ[R] A₂) (f h₁ h₂ h₃ h₄ h₅) :
   symm_bijective.injective <| ext fun _ => rfl
 #align alg_equiv.mk_coe' AlgEquiv.mk_coe'
 
+/-- Auxilliary definition to avoid looping in `dsimp` with `AlgEquiv.symm_mk`. -/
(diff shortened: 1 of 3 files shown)


mattrobball:
bors delegate+

Mathlib/Algebra/Algebra/Equiv.lean
```diff
@@ -360,15 +360,27 @@ theorem mk_coe' (e : A₁ ≃ₐ[R] A₂) (f h₁ h₂ h₃ h₄ h₅) :
...
 #align alg_equiv.symm_mk AlgEquiv.symm_mk
 
+-- remove before merging
+theorem symm_mk_test (f f') (h₁ h₂ h₃ h₄ h₅) :
+    (⟨⟨f, f', h₁, h₂⟩, h₃, h₄, h₅⟩ : A₁ ≃ₐ[R] A₂).symm =
+      { symm_mk.aux f f' h₁ h₂ h₃ h₄ h₅ with
+        toFun := f'
+        invFun := f } := by
+  dsimp
+
```
jcommelin, line 383:
```suggestion
```

jcommelin (commented):
bors r+

PR #13809, opened 2024-06-13
[Merged by Bors] - feat(CategoryTheory/Category/Cat): add Cat API

callesonne:
Add API for `Bicategory.whiskerRight` and `Bicategory.whiskerLeft`

---

### DIFF
--- a/Mathlib/CategoryTheory/Category/Cat.lean
+++ b/Mathlib/CategoryTheory/Category/Cat.lean
@@ -107,38 +107,32 @@ theorem comp_map {C D E : Cat} (F : C ⟶ D) (G : D ⟶ E) {X Y : C} (f : X ⟶ 
 set_option linter.uppercaseLean3 false in
 #align category_theory.Cat.comp_map CategoryTheory.Cat.comp_map
 
-@[simp, nolint simpNF]
+@[simp]
 lemma whiskerLeft_app {C D E : Cat} (F : C ⟶ D) {G H : D ⟶ E} (η : G ⟶ H) (X : C) :
     (F ◁ η).app X = η.app (F.obj X) :=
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Category/Cat.lean
```diff
@@ -105,6 +107,16 @@ theorem comp_map {C D E : Cat} (F : C ⟶ D) (G : D ⟶ E) {X Y : C} (f : X ⟶
 set_option linter.uppercaseLean3 false in
 #align category_theory.Cat.comp_map CategoryTheory.Cat.comp_map
 
+@[simp]
+lemma whiskerLeft_app {C D E : Cat} (F : C ⟶ D) {G H : D ⟶ E} (η : G ⟶ H) (X : C) :
+    (F ◁ η).app X = η.app (F.obj X) :=
+  CategoryTheory.whiskerLeft_app F η X
```
joelriou, line 113: When the proof can be `rfl`, the proof should be `rfl`.
```suggestion
    (F ◁ η).app X = η.app (F.obj X) := rfl
```
Could you do the same modification for the other lemmas (above and below this one)? Could you also add similar lemmas for the associator and unitors?

callesonne: Will we really need similar lemmas for associators and unitors? Because since `Cat` is a strict bicategory there are already simp-lemmas that these are just eqToIsos.

callesonne: (although their proofs are by aesop_cat and not rfl, so maybe thats a reason to add them here)

joelriou:
Thanks!

bors merge

PR #13810, opened 2024-06-13
[Merged by Bors] - feat(Combinatorics/SimpleGraph): add `Fintype` instance for `Path`

niklasmohrin:
---

I tried to golf the implementation a bit, I would have expected the following to work:

```lean
complete p := by
  simp
  exact ⟨p.val.length, p.prop.length_lt, p.val, p.prop, rfl⟩
```

... but for some reason only the longer version compiles. Any ideas?


Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2612,6 +2612,19 @@ instance : Decidable G.Connected := by
   rw [connected_iff, ← Finset.univ_nonempty_iff]
   infer_instance
 
+open Finset in
+instance Path.instFintype {u v : V} : Fintype (G.Path u v) where
+  elems := (range (Fintype.card V)).biUnion
+    fun n ↦ (univ (α := {p : G.Walk u v | p.IsPath ∧ p.length = n})).image
+      fun p ↦ { val := p.val, property := p.prop.left }
```
eric-wieser, line 2619: This is going to be inefficient for computation, I think; something like
```suggestion
  elems := (range (Fintype.card V)).disjiUnion
    (fun n ↦ (univ (α := {p : G.Walk u v | p.IsPath ∧ p.length = n})).map
      ⟨fun p ↦ { val := p.val, property := p.prop.left }, sorry⟩) sorry
```
would behave much better.

Of course, this has the downside that you have two proofs to fill!

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2612,6 +2612,19 @@ instance : Decidable G.Connected := by
...
+  complete p := by
+    simp only [mem_biUnion, mem_range, mem_image, mem_univ, Subtype.exists]
+    use p.val.length
+    constructor
+    · exact p.prop.length_lt
+    · use p.val
+      use ⟨p.prop, rfl⟩
```
eric-wieser, line 2626:
```suggestion
    simp only [mem_biUnion, mem_range, mem_image, mem_univ, Subtype.exists, true_and]
    exact ⟨p.val.length, p.prop.length_lt, p.val, ⟨⟨p.prop, rfl⟩, rfl⟩⟩
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
niklasmohrin: Sounds good. Out of curiosity, where would this inefficiency show? Is it just that we are avoiding the two extra deduplication steps during construction of the `Finset`?

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2612,6 +2655,15 @@ instance : Decidable G.Connected := by
...
   infer_instance
 
+open Finset in
+instance Path.instFintype {u v : V} : Fintype (G.Path u v) where
+  elems := (univ (α := { p : G.Walk u v // p.IsPath ∧ p.length < Fintype.card V })).map
+    ⟨fun p ↦ { val := p.val, property := p.prop.left },
+     fun _ _ h ↦ SetCoe.ext <| Subtype.mk.injEq .. ▸ h⟩
+  complete p := mem_map.mpr ⟨
+    ⟨p.val, ⟨p.prop, p.prop.length_lt⟩⟩,
+    ⟨@mem_univ _ (fintypeSubtypePathLengthLt ..) _, rfl⟩⟩
```
niklasmohrin, line 2665: Here it complained with just `mem_univ _`: `failed to synthesize Fintype ↑fun p ↦ p.IsPath ∧ p.length < Fintype.card V` - I guess it is fine like this?

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
eric-wieser: It would come when evaluating some computable sum over all paths, which would then perform all this deduplication at runtime. It would also appear if you evaluate something like `example : Fintype.card (G.Path _ _) = 10 = rfl` for some computable `G`.

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2554,6 +2554,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
 
 variable (G)
 
+/-- The `Finset` of walks from `u` to `v` with length less than `n`. See `finsetWalkLength`. --/
+def finsetWalkLengthLt (n : ℕ) (u v : V) : Finset (G.Walk u v) :=
+  (Finset.range n).disjiUnion
+    (fun l ↦ G.finsetWalkLength l u v)
+    (fun l _ l' _ hne _ hsl hsl' p hp ↦
+      have hl : p.length = l := (Walk.mem_finsetWalkLength_iff_length_eq p).mp (hsl hp)
```
IvanRenison, line 2562:
```suggestion
      have hl : p.length = l := p.mem_finsetWalkLength_iff_length_eq.mp (hsl hp)
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2554,6 +2554,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
...
+      have hl' : p.length = l' := (Walk.mem_finsetWalkLength_iff_length_eq p).mp (hsl' hp)
```
IvanRenison, line 2563:
```suggestion
      have hl' : p.length = l' := p.mem_finsetWalkLength_iff_length_eq.mp (hsl' hp)
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2576,12 +2599,32 @@ theorem card_set_walk_length_eq (u v : V) (n : ℕ) :
     rw [← Finset.mem_coe, coe_finsetWalkLength_eq]
 #align simple_graph.card_set_walk_length_eq SimpleGraph.card_set_walk_length_eq
 
+instance fintypeSetWalkLengthLt (u v : V) (n : ℕ) : Fintype {p : G.Walk u v | p.length < n} :=
+  Fintype.ofFinset (G.finsetWalkLengthLt n u v) fun p ↦ by
+    rw[← Finset.mem_coe, coe_finsetWalkLengthLt_eq]
```
IvanRenison, line 2604:
```suggestion
    rw [← Finset.mem_coe, coe_finsetWalkLengthLt_eq]
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2577,6 +2577,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
...
 variable (G)
 
+/-- The `Finset` of walks from `u` to `v` with length less than `n`. See `finsetWalkLength`. --/
+def finsetWalkLengthLt (n : ℕ) (u v : V) : Finset (G.Walk u v) :=
+  (Finset.range n).disjiUnion
+    (fun l ↦ G.finsetWalkLength l u v)
+    (fun l _ l' _ hne _ hsl hsl' p hp ↦
+      have hl : p.length = l := p.mem_finsetWalkLength_iff_length_eq.mp (hsl hp)
+      have hl' : p.length = l' := p.mem_finsetWalkLength_iff_length_eq.mp (hsl' hp)
+      False.elim <| hne <| hl.symm.trans hl')
```
YaelDillies, line 2587: @henkkuli, thoughts?

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2577,6 +2577,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
 
 variable (G)
 
+/-- The `Finset` of walks from `u` to `v` with length less than `n`. See `finsetWalkLength`. --/
+def finsetWalkLengthLt (n : ℕ) (u v : V) : Finset (G.Walk u v) :=
```
YaelDillies, line 2581:
```suggestion
def finsetWalkLengthLT (n : ℕ) (u v : V) : Finset (G.Walk u v) :=
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2577,6 +2577,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
...
+
+open Finset in
+theorem coe_finsetWalkLengthLt_eq (n : ℕ) (u v : V) :
+    (G.finsetWalkLengthLt n u v : Set (G.Walk u v)) = {p : G.Walk u v | p.length < n} := by
+  ext p
+  simp [finsetWalkLengthLt, mem_coe, mem_disjiUnion, Walk.mem_finsetWalkLength_iff_length_eq]
+
+variable {G}
+
+theorem Walk.mem_finsetWalkLengthLt_iff_length_lt {n : ℕ} {u v : V} (p : G.Walk u v) :
```
YaelDillies, line 2597: Iffs take implicit arguments
```suggestion
theorem mem_finsetWalkLengthLT {n : ℕ} {u v : V} {p : G.Walk u v} :
```

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
YaelDillies: I think it's fixed if you do...

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -2635,9 +2678,19 @@ instance : Decidable G.Connected := by
   rw [connected_iff, ← Finset.univ_nonempty_iff]
   infer_instance
 
+open Finset in
+instance Path.instFintype {u v : V} : Fintype (G.Path u v) where
+  elems := (univ (α := { p : G.Walk u v // p.IsPath ∧ p.length < Fintype.card V })).map
```
YaelDillies, line 2683: ...this
```suggestion
  elems := (univ (α := { p : G.Walk u v | p.IsPath ∧ p.length < Fintype.card V })).map
```

YaelDillies:
Can you please merge master?

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
niklasmohrin: Ah, I just adapted the lemma from https://github.com/leanprover-community/mathlib4/blob/bc3ab3eddef1b92b87d321f25c74899502c25239/Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean#L112 - should we change it there too then?

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
niklasmohrin: Yup, that worked! :) 

Any idea why one works better than the other?

YaelDillies: `{p : G.Walk u v | p.IsPath ∧ p.length < Fintype.card V}` is a set, `{p : G.Walk u v // p.IsPath ∧ p.length < Fintype.card V}` is a subtype. You were using the latter as a set, which doesn't really work (well, it does, but only if you spend some effort convincing Lean so)

niklasmohrin:
Ah, one advantage of keeping `Network.loopless` would be that once `SimpleGraph V extends DiGraph V`, we could have `UndirectedNetwork V R extends Network V R, SimpleGraph V` - maybe that would be too restricting though, as there could theoretically also be non-simple graphs employed with a symmetric capacity function.

I suppose the most flexible design to have all kinds of networks carry along all kinds of graphs would be `structure Network (G : Digraph V) (R : Type u_R)`. However, this would defeat the main purpose for the move to the `Network extends Digraph` design :thinking:

What should we do?

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
```diff
@@ -115,6 +115,29 @@ theorem Walk.mem_finsetWalkLength_iff_length_eq {n : ℕ} {u v : V} (p : G.Walk
 
 variable (G)
 
+/-- The `Finset` of walks from `u` to `v` with length less than `n`. See `finsetWalkLength`. --/
```
YaelDillies, line 118:
```suggestion
/-- The `Finset` of walks from `u` to `v` with length less than `n`. See `SimpleGraph.finsetWalkLength` for ???. -/
```
where `???` is something you should fill in

YaelDillies (approved):
maintainer merge

niklasmohrin:
@YaelDillies what should we do about https://github.com/leanprover-community/mathlib4/pull/13810#discussion_r1723735522 ?

YaelDillies:
Oh, Henrik must be busy for now. We will update his project once this PR has merged.

niklasmohrin:
And what about mem_finsetWalkLengthLt_iff_length_lt and the implicit arguments?

YaelDillies:
Oh yes, sorry, I overlooked that. `p` should be implicit. Can you change it?

niklasmohrin:
Just so we are on the same page because I cannot make out a clear answer to the question in my comment: I now changed _my_ definition like you said, but left the existing definition `Walk.mem_finsetWalkLength_iff_length_eq` as before (with `Walk.` namespace and explicit `p`). This means that the definitions are inconsistent with each other, although they are the same statement just for the length-eq and the length-lt cases. If that is not what you meant, please let me know

YaelDillies:
You should feel free to make `p` implicit in the existing one as well

jcommelin:
Since there's still discussion going on, I've removed the `maintainer-merge` label for now.

YaelDillies:
Okay, I think we're good?

maintainer merge

semorrison:
bors merge

kbuzzard:
I was just reading through this -- happy with the merge, but I personally am not clear about where in the library we're using `Fintype` and where we're using `Finite`. For example in finite group theory we seem to be moving to `Finite`. Why is `Fintype` the preferred option here?

niklasmohrin:
The way I understand it is that with Fintype one can actually compute the elements (because this computation is provided with `elems`) whereas Finite only contains a proof. Since some downstream algorithm may want to enumerate all paths, Fintype seems to be the right choice here. I am lost on when to use Finite vs noncomputable Fintype though

PR #13814, opened 2024-06-14
[Merged by Bors] - chore: bump to v4.9.0-rc2

kim-em:
This should not change any Lean behaviour, besides `lake`, `VS Code` interaction, and incremental compilation.

Note that to save some time I'm not moving all the other repositories to `v4.9.0-rc2` now, but can do so on an as needed basis.

---


semorrison:
bors p=10

PR #13815, opened 2024-06-14
chore: adaptations for leanprover/lean4#4400, List look-up normal form

kim-em:
This shouldn't be merged until leanprover/lean4#4400 lands in a nightly.

- [x] depends on: leanprover-community/batteries/pull/843
- [x] depends on: leanprover/lean4#4400


semorrison:
I've merged this by hand, so will close this PR. It will be replaced with a PR to `bump/v4.10.0` shortly.

PR #13817, opened 2024-06-14
chore: correct `CompletelyNormalSpace` API to match `NormalSpace` by using `SeparatedNhds`

StevenClontz:
The current behavior for CompletelyNormalSpace is to prove `Disjoint (𝓝ˢ s) (𝓝ˢ t)`, while NormalSpace uses a proof of the (equivalent) `SeparatedNhds s t`. This changes aligns both separation axioms to use `SeparatedNhds s t`.

---

### DIFF
--- a/Counterexamples/SorgenfreyLine.lean
+++ b/Counterexamples/SorgenfreyLine.lean
@@ -203,7 +203,7 @@ instance : CompletelyNormalSpace ℝₗ := by
     exists_Ico_disjoint_closed isClosed_closure (disjoint_left.1 hd₂ hx)
   choose! Y hY hYd using fun y (hy : y ∈ t) =>
     exists_Ico_disjoint_closed isClosed_closure (disjoint_right.1 hd₁ hy)
-  refine disjoint_of_disjoint_of_mem ?_
+  refine separatedNhds_iff_disjoint.mpr <| disjoint_of_disjoint_of_mem ?_
     (bUnion_mem_nhdsSet fun x hx => (isOpen_Ico x (X x)).mem_nhds <| left_mem_Ico.2 (hX x hx))
     (bUnion_mem_nhdsSet fun y hy => (isOpen_Ico y (Y y)).mem_nhds <| left_mem_Ico.2 (hY y hy))
   simp only [disjoint_iUnion_left, disjoint_iUnion_right, Ico_disjoint_Ico]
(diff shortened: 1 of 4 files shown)


JADekker (commented):
The changes look good to me, but I have the feeling that most proofs that you touched seem to become longer after this change (now requiring some form of `separatedNhds_iff_disjoint`), which makes me wonder if these proofs can be simplified if we also use `SeparatedNhds` in another place, or if it may be more natural to formulate both definitions in terms of `Disjoint` instead. Any thoughts?

StevenClontz:
That occurred to me too. I wanted to keep the PR clean, but since you bring it up, I'll take a swing at refactoring some of the lengthened results and report back.

JADekker:
> That occurred to me too. I wanted to keep the PR clean, but since you bring it up, I'll take a swing at refactoring some of the lengthened results and report back.

Doing that in a separate PR is also fine as far as I'm concerned, I just wanted to raise the issue already!

StevenClontz:
> > That occurred to me too. I wanted to keep the PR clean, but since you bring it up, I'll take a swing at refactoring some of the lengthened results and report back.
> 
> Doing that in a separate PR is also fine as far as I'm concerned, I just wanted to raise the issue already!

In that case, I'll revert back to 8497635 and open a new PR with the progress I made since.

urkud:
Note that `SeparatedNhds` were defined before the `nhdsSet` filter. I think that we should migrate in the other direction (or change the definition of `SeparatedNhds`, if we want to preserve the name).

StevenClontz:
For me, the most compelling argument to migrate everything to `Disjoint (𝓝ˢ s) (𝓝ˢ t)` is that it better matches up with Hausdorff `Disjoint (𝓝 x) (𝓝 y)` and regular `Disjoint (𝓝 x) (𝓝ˢ t)`.

Actually, I'm surprised to notice that `T2Space` is *not* defined as `Disjoint (𝓝 x) (𝓝 y)`. Need to step away at the moment but will return to this.

StevenClontz:
I think my inclination, considering https://leanprover-community.github.io/theories/topology.html#filters in particular, is that a refactor to use `Disjoint` filters for all separation properties is the approach most in line with my understanding of Mathlib style.

Also `SeparatedNhds` is a bit of a misnomer anyway, as it produces separated *open* neighborhoods.

Would a refactor that removed `SeparatedNhds` as a concept, but preserved the ability to quickly grab appropriate *open* neighborhoods when desired, be appropriate then? I'm happy to take a crack at this, but this is a bigger chore than I've done so far, so guidance from experienced community members is appreciated.

PatrickMassot:
Yes, everything that improves consistency by putting filters everywhere is welcome.

StevenClontz:
Alright. Closing this, and will open a new PR to use filters for all separation axioms.

PR #13828, opened 2024-06-14
[Merged by Bors] - feat: `a ^ (p - 1) = if a ≠ 0 then 1 else 0` in `ZMod p`

YaelDillies:
This is useful for algebraically encoding equality of indices.

From LeanCamCombi

---

### DIFF
--- a/Mathlib/FieldTheory/Finite/Basic.lean
+++ b/Mathlib/FieldTheory/Finite/Basic.lean
@@ -533,7 +533,8 @@ theorem pow_card_sub_one_eq_one {p : ℕ} [Fact p.Prime] {a : ZMod p} (ha : a �
     rwa [ZMod.card p] at h
 #align zmod.pow_card_sub_one_eq_one ZMod.pow_card_sub_one_eq_one
 
-lemma pow_card_sub_one (a : ZMod p) : a ^ (p - 1) = if a ≠ 0 then 1 else 0 := by
+lemma pow_card_sub_one {p : ℕ} [Fact p.Prime] (a : ZMod p) :
+    a ^ (p - 1) = if a ≠ 0 then 1 else 0 := by
   split_ifs with ha
   · exact pow_card_sub_one_eq_one ha
   · simp [of_not_not ha, (Fact.out : p.Prime).one_lt, tsub_eq_zero_iff_le]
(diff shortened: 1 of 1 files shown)


Mathlib/FieldTheory/Finite/Basic.lean
```diff
@@ -533,6 +533,11 @@ theorem pow_card_sub_one_eq_one {p : ℕ} [Fact p.Prime] {a : ZMod p} (ha : a 
     rwa [ZMod.card p] at h
 #align zmod.pow_card_sub_one_eq_one ZMod.pow_card_sub_one_eq_one
 
+lemma pow_card_sub_one (a : ZMod p) : a ^ (p - 1) = if a ≠ 0 then 1 else 0 := by
```
Rida-Hamadani, line 536:
```suggestion
lemma pow_card_sub_one {p : ℕ} [Fact p.Prime] (a : ZMod p) : a ^ (p - 1) = if a ≠ 0 then 1 else 0 := by
```

YaelDillies: Whoops indeed. I am editing mathlib cache-less since [this](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/cache.20misses.20aesop) happened

Ruben-VandeVelde (approved):
maintainer merge

Mathlib/FieldTheory/Finite/Basic.lean
```diff
@@ -533,6 +533,12 @@ theorem pow_card_sub_one_eq_one {p : ℕ} [Fact p.Prime] {a : ZMod p} (ha : a 
     rwa [ZMod.card p] at h
 #align zmod.pow_card_sub_one_eq_one ZMod.pow_card_sub_one_eq_one
 
+lemma pow_card_sub_one {p : ℕ} [Fact p.Prime] (a : ZMod p) :
+    a ^ (p - 1) = if a ≠ 0 then 1 else 0 := by
```
urkud, line 537: Why test on `a ≠ 0` instead of `a = 0`? Does it make sense to use `Pi.mulSingle` here?

urkud:
Otherwise LGTM. Thanks!
bors d+

Mathlib/FieldTheory/Finite/Basic.lean
YaelDillies: The point is that `if p then 1 else 0` is used in plenty of statements, eg `Finset.sum_boole`. I'm happy to add other versions later but this is precisely the one I need for now

YaelDillies:
bors merge

PR #13830, opened 2024-06-14
[Merged by Bors] - feat: `Prod.fst` as a lattice hom

YaelDillies:
From LeanCamCombi

---

### DIFF
--- a/Mathlib/Order/Hom/Lattice.lean
+++ b/Mathlib/Order/Hom/Lattice.lean
@@ -1612,7 +1612,7 @@ end LatticeHom
 
 /-! ### Pi -/
 
-namespace LatticeHom
+namespace Pi
 variable {ι : Type*} {α : ι → Type*} [∀ i, Lattice (α i)]
 
 /-- Evaluation as a lattice homomorphism. -/
@@ -1626,7 +1626,7 @@ lemma coe_evalLatticeHom (i : ι) : ⇑(evalLatticeHom (α := α) i) = Function.
 
 lemma evalLatticeHom_apply (i : ι) (f : ∀ i, α i) : evalLatticeHom i f = f i := rfl
 
-end LatticeHom
+end Pi
 
 /-! ### `WithTop`, `WithBot` -/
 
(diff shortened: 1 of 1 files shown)


Ruben-VandeVelde (approved):
maintainer merge

Mathlib/Order/Hom/Lattice.lean
```diff
@@ -1586,6 +1586,48 @@ theorem symm_dual_comp (g : BoundedLatticeHom βᵒᵈ γᵒᵈ) (f : BoundedLat
...
+
+end LatticeHom
+
+/-! ### Pi -/
+
+namespace LatticeHom
+variable {ι : Type*} {α : ι → Type*} [∀ i, Lattice (α i)]
+
+/-- Evaluation as a lattice homomorphism. -/
+def evalLatticeHom (i : ι) : LatticeHom (∀ i, α i) (α i) where
```
urkud, line 1619: Why repeat `LatticeHom` in NS and name?

YaelDillies: I think the naming is all over the place here

eric-wieser: I think we standardized on `FooHom.fst` and `Pi.evalFooHom`?

YaelDillies: Uh, really? That's kind of a terrible in-between

YaelDillies: Please just tell me what I should use and I will do it

eric-wieser: > Please just tell me what I should use and I will do it

Please rename the Prod version to `FooHom.fst`, as you had them before. The Pi version is correctly named as you have it right now.

eric-wieser: > Uh, really? That's kind of a terrible in-between

I think now that we're in Lean 4 we could reconsider the Pi naming in the face of dot notation (which works on the `.fst` definitions); but that's not something to mix into this PR.

eric-wieser (approved):
bors merge

Thanks!

PR #13831, opened 2024-06-14
[Merged by Bors] - chore(Order/Hom/Set): Use `simps!` rather than `simps`

YaelDillies:
... else the RHS of the generated lemmas involves a `∘`

---


robertylewis:
bors merge

PR #13832, opened 2024-06-14
[Merged by Bors] - chore: Rename `Nat.factors` to `Nat.primeFactorsList`

YaelDillies:
`Nat.factors` was a very generic name that didn't emphasize that
* It is a list
* It is not all factors but only the prime ones

---

### DIFF
--- a/Mathlib/Data/Nat/PrimeFin.lean
+++ b/Mathlib/Data/Nat/PrimeFin.lean
@@ -31,12 +31,12 @@ instance Primes.infinite : Infinite Primes := infinite_setOf_prime.to_subtype
 instance Primes.countable : Countable Primes := ⟨⟨coeNat.coe, coe_nat_injective⟩⟩
 
 /-- The prime factors of a natural number as a finset. -/
-def primeFactors (n : ℕ) : Finset ℕ := n.factors.toFinset
+def primeFactors (n : ℕ) : Finset ℕ := n.primeFactorsList.toFinset
 
-@[simp] lemma toFinset_factors (n : ℕ) : n.factors.toFinset = n.primeFactors := rfl
+@[simp] lemma toFinset_factors (n : ℕ) : n.primeFactorsList.toFinset = n.primeFactors := rfl
 
 @[simp] lemma mem_primeFactors : p ∈ n.primeFactors ↔ p.Prime ∧ p ∣ n ∧ n ≠ 0 := by
-  simp_rw [← toFinset_factors, List.mem_toFinset, mem_factors']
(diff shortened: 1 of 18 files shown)


Ruben-VandeVelde:
Can you limit this PR to the renaming?

YaelDillies:
What am I doing that is not the renaming?

BoltonBailey (approved):
Perhaps Ruben is confused that this changes not just `factors` but every lemma name that referenced `factors` as a substring? I don't think it makes sense to change those in a different PR though, this looks good to me.

YaelDillies:
I should say that I am not particularly attached to the name `primeFactorsList` and I would be happy to hear alternatives.

b-mehta (approved):

Mathlib/Data/Nat/PrimeFin.lean
```diff
@@ -45,7 +45,7 @@ lemma primeFactors_mono (hmn : m ∣ n) (hn : n ≠ 0) : primeFactors m ⊆ prim
   simp only [subset_iff, mem_primeFactors, and_imp]
   exact fun p hp hpm _ ↦ ⟨hp, hpm.trans hmn, hn⟩
 
-lemma mem_primeFactors_iff_mem_factors : p ∈ n.primeFactors ↔ p ∈ n.factors := by
+lemma mem_primeFactors_iff_mem_factors : p ∈ n.primeFactors ↔ p ∈ n.primeFactorsList := by
```
semorrison, line 48:
```suggestion
lemma mem_primeFactors_iff_mem_primeFactorsList : p ∈ n.primeFactors ↔ p ∈ n.primeFactorsList := by
```

Mathlib/Data/Nat/Squarefree.lean
```diff
@@ -25,15 +25,16 @@ open Finset
 
 namespace Nat
 
-theorem squarefree_iff_nodup_factors {n : ℕ} (h0 : n ≠ 0) : Squarefree n ↔ n.factors.Nodup := by
+theorem squarefree_iff_nodup_primeFactorsList {n : ℕ} (h0 : n ≠ 0) :
```
semorrison, line 28: I think we should add deprecated aliases for all of these, no?

YaelDillies: even if the `factors` appearing in the RHS is itself deprecated already?

semorrison: How will people using these lemmas know how to update their code? I think they are needed. The point is not to say that they are deprecated but to give people one click fixes.

YaelDillies:
Was a trivial fix

bors merge

YaelDillies:
bors merge

PR #13826, opened 2024-06-14
[Merged by Bors] - feat: `a * (b - 1) = a * b - a` for truncated subtraction

YaelDillies:
... along with the `Nat`-specialised versions. Also add a few simple lemmas about `div` and `dvd`, restore `Nat.mul_div_mul_comm` and fix the name `AddGroup.toHasOrderedSub` to `AddGroup.toOrderedSub`.

From LeanCamCombi

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13833, opened 2024-06-14
fix: undeprecate Nat.mul_div_mul_comm_of_dvd_dvd

dwrensha:
According to YaelDillies,

> Nat.mul_div_mul_comm_of_dvd_dvd/Nat.div_mul_div_comm is supposed to mimic mul_div_mul_comm/div_mul_div_comm. 

and therefore it should not be removed.

---


YaelDillies:
Sorry, you were too slow. I have now opened #13826 😁

dwrensha:
Made obsolete by #13826.

PR #13837, opened 2024-06-14
[Merged by Bors] - feat: the spectrum of a diagonal matrix is the range of the diagonal

JonBannon:
We add various theorems that assert the spectrum of a diagonal matrix is the range of the diagonal viewed as a function.

Co-Authored by @j-loreaux

---


Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -578,3 +578,54 @@ lemma map_add_of_iInf_genEigenspace_ne_bot_of_commute [NoZeroSMulDivisors R M]
...
+      exact (hasEigenvector_toLin'_diagonal d i).apply_eq_smul
+    have hμ_not_mem : μ ∉ Set.range d := by simpa using fun i ↦ (hμ i)
+    have := eigenspaces_independent (toLin' (diagonal d)) |>.disjoint_biSup hμ_not_mem
+    rw [h_iSup, disjoint_top] at this
+    exact h_eig this
+  · rintro ⟨i, rfl⟩
+    exact this i
+
+/-- The spectrum of the diagonal operator is the range of the diagonal viewed as a function. -/
+lemma Matrix.spectrum_diagonal (d : n → R) :
```
eric-wieser, line 624: This is probably sensible as a `simp` lemma.

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -578,3 +578,54 @@ lemma map_add_of_iInf_genEigenspace_ne_bot_of_commute [NoZeroSMulDivisors R M]
 end End
 
 end Module
+
+section SpectrumDiagonal
+
+variable {R : Type*} [Field R] {n : Type*} [DecidableEq n][Fintype n]
```
eric-wieser, line 584: style nit:
```suggestion
variable {R : Type*} [Field R] {n : Type*} [DecidableEq n] [Fintype n]
```

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -578,3 +578,54 @@ lemma map_add_of_iInf_genEigenspace_ne_bot_of_commute [NoZeroSMulDivisors R M]
...
+
+open Module.End
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma Matrix.hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
+    Module.End.HasEigenvector (Matrix.toLin' (diagonal d)) (d i) (Pi.basisFun R n i) := by
```
eric-wieser, line 590: Is this also true for something like
```suggestion
    Module.End.HasEigenvector (Matrix.toLin b b (diagonal d)) (d i) (b i) := by
```
for some `b : Basis _ _ _`? If so, it would be good to have both versions, preferably with one proof using the other.

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -578,3 +578,54 @@ lemma map_add_of_iInf_genEigenspace_ne_bot_of_commute [NoZeroSMulDivisors R M]
...
 end Module
+
+section SpectrumDiagonal
+
+variable {R : Type*} [Field R] {n : Type*} [DecidableEq n][Fintype n]
+
+open Module.End
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma Matrix.hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
```
eric-wieser, line 589: This is the first mention of `Matrix` in this file; perhaps that means this content would be better in a new `EigenSpace/Matrix.lean` file?

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
JonBannon: I'll look into this. Thanks!

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
JonBannon: Cool. Let me confer with Jireh about how best to do this.

JonBannon: @eric-wieser : Should I create this new file and move the result as part of this PR, or should I handle this in another PR? (I'm still trying to get the hang of this!)

j-loreaux: Yes, just make the new file and move this result into it as part of this PR. Generally, I would recommend basing what should constitute a separate PR in terms of semantically / mathematically distinct content (although there can be exceptions to this).

On a separate note: the `b : Basis _ _ _` from the other comment above probably should also apply here.

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,75 @@
...
+section SpectrumDiagonal
+
+variable {R : Type*} [Field R] {n : Type*} [DecidableEq n] [Fintype n] (M : Type*)
+  [AddCommGroup M] [Module R M]
+
+open Matrix
+open Module.End
+
+/-- Basis vectors are eigenvectors of associated diagonal linear operator. -/
+lemma hasEigenvector_toLin_basis_diagonal (d : n → R) (i : n) (b : Basis n R M) :
```
j-loreaux, line 30: Probably `basis` doesn't need to be in the name.

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,75 @@
...
+    all_goals simp_all
+  · exact Basis.ne_zero b i
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
+    Module.End.HasEigenvector (toLin' (diagonal d)) (d i) (Pi.basisFun R n i) :=
+  hasEigenvector_toLin_basis_diagonal (d : n → R) (i : n) (b := Pi.basisFun R n)
+
+/-- Eigenvalues of a diagonal linear operator are the diagonal entries. -/
+lemma hasEigenvalue_toLin'_diagonal_iff (d : n → R) {μ : R} :
```
j-loreaux, line 45: This should have a `toLin` generalization.

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Yeah, I wasn't too sure about this. Do you think this is a case where one of these is given a prime and the other isn't? I'm not really sure of a good name for this result...

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Ah! I just learned something new. So, `toLin` is the general basis version, and `toLin'` is the `PiLp 2` version? 

Thanks for the clarification on this. (I assume this was what your previous suggestion about basis _ _ _ was referring to?)

j-loreaux: Actually, `toLin'` is just the standard basis version, but with the max norm. `toEuclideanLin` is the `PiLp 2` version.

JonBannon:
Ah, cool. To clarify, we want the standard basis version here, right?
Should I provide versions for toEuclideanLin as well?

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+
+/-- Basis vectors are eigenvectors of associated diagonal linear operator. -/
+lemma hasEigenvector_toLin_diagonal (d : n → R) (i : n) (b : Basis n R M) :
+    Module.End.HasEigenvector (toLin b b (diagonal d)) (d i) (b i) := by
+  constructor
+  · refine mem_eigenspace_iff.mpr ?left.a
+    simp only [diagonal, Pi.basisFun_apply, toLin'_apply, mulVec_stdBasis_apply, transpose_apply,
+      of_apply, Pi.smul_apply, LinearMap.stdBasis_apply', smul_eq_mul, mul_ite, mul_one, mul_zero]
+    all_goals simp_all
+  · exact Basis.ne_zero b i
```
j-loreaux, line 37: :golfing:
```suggestion
    Module.End.HasEigenvector (toLin b b (diagonal d)) (d i) (b i) :=
  ⟨mem_eigenspace_iff.mpr <| by simp [diagonal], Basis.ne_zero b i⟩
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
+    Module.End.HasEigenvector (toLin' (diagonal d)) (d i) (Pi.basisFun R n i) :=
+  hasEigenvector_toLin_diagonal (d : n → R) (i : n) (b := Pi.basisFun R n)
```
j-loreaux, line 42: Lean can infer most of the arguments. You really only need to provide the basis.
```suggestion
  hasEigenvector_toLin_diagonal _ _ _ <| Pi.basisFun R n
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+over a field.
+
+## Tags
+eigenspace, eigenvector, eigenvalue, eigen, spectrum, matrices, matrix
+
+-/
+
+section SpectrumDiagonal
+
+variable {R : Type*} [Field R] {n : Type*} [DecidableEq n] [Fintype n] (M : Type*)
```
j-loreaux, line 23: I don't think we really want `M` explicit here. Notice also that we generally group `Type` arguments at the beginning.
```suggestion
variable {R n M : Type*} [Field R] [DecidableEq n] [Fintype n]
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+  [AddCommGroup M] [Module R M]
+
+open Matrix
+open Module.End
+
+/-- Basis vectors are eigenvectors of associated diagonal linear operator. -/
+lemma hasEigenvector_toLin_diagonal (d : n → R) (i : n) (b : Basis n R M) :
+    Module.End.HasEigenvector (toLin b b (diagonal d)) (d i) (b i) := by
```
j-loreaux, line 31: You have this namespace open
```suggestion
    HasEigenvector (toLin b b (diagonal d)) (d i) (b i) := by
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+  constructor
+  · refine mem_eigenspace_iff.mpr ?left.a
+    simp only [diagonal, Pi.basisFun_apply, toLin'_apply, mulVec_stdBasis_apply, transpose_apply,
+      of_apply, Pi.smul_apply, LinearMap.stdBasis_apply', smul_eq_mul, mul_ite, mul_one, mul_zero]
+    all_goals simp_all
+  · exact Basis.ne_zero b i
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
+    Module.End.HasEigenvector (toLin' (diagonal d)) (d i) (Pi.basisFun R n i) :=
```
j-loreaux, line 41:
```suggestion
    HasEigenvector (toLin' (diagonal d)) (d i) (Pi.basisFun R n i) :=
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+-/
+
+import Mathlib.LinearAlgebra.Eigenspace.Basic
+
+/-!
+
+# Eigenvalues, Eigenvectors and Spectrum for Matrices
+
+This file collects results about eigenvectors, eigenvalues and spectrum specific to matrices
+over a field.
```
j-loreaux, line 14: I think only the last result needs `R` to be a field. The rest hold under weaker assumptions. Please try to minimize the assumptions you need on `R` by changing them per theorem as needed.

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+
+## Tags
+eigenspace, eigenvector, eigenvalue, eigen, spectrum, matrices, matrix
```
j-loreaux, line 17: I don't think we need all of these
```suggestion
eigenspace, eigenvector, eigenvalue, spectrum, matrix
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+    rw [h_iSup, disjoint_top] at this
+    exact h_eig this
+  · rintro ⟨i, rfl⟩
+    exact this i
+
+/-- Eigenvalues of a diagonal linear operator with respect to standard basis
+    are the diagonal entries. -/
+lemma hasEigenvalue_toLin'_diagonal_iff (d : n → R) {μ : R} :
+    HasEigenvalue (toLin' (diagonal d)) μ ↔ ∃ i, d i = μ := by
+  exact hasEigenvalue_toLin_diagonal_iff (n → R) d (Pi.basisFun R n)
```
j-loreaux, line 71: No need for `by exact`, and Lean can infer the other arguments besides the basis:
```suggestion
    HasEigenvalue (toLin' (diagonal d)) μ ↔ ∃ i, d i = μ :=
  hasEigenvalue_toLin_diagonal_iff _ _ <| Pi.basisFun R n
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,81 @@
...
+
+/-- The spectrum of the diagonal operator is the range of the diagonal viewed as a function. -/
+lemma spectrum_diagonal (d : n → R) :
+    spectrum R (diagonal d) = Set.range d := by
+  ext μ
+  rw [← AlgEquiv.spectrum_eq (toLinAlgEquiv <| Pi.basisFun R n),
+    ← hasEigenvalue_iff_mem_spectrum, Set.mem_range]
```
j-loreaux, line 78: :golfing:
```suggestion
  rw [← AlgEquiv.spectrum_eq (toLinAlgEquiv <| Pi.basisFun R n), ← hasEigenvalue_iff_mem_spectrum]
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Cool. Will do this first thing tomorrow. Thanks!

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Neat. I didn't know that. Thanks!

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Ah I think I see. Both of these apply to the result below...

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,93 @@
...
+over a field.
+
+## Tags
+eigenspace, eigenvector, eigenvalue, spectrum, matrix
+
+-/
+
+section SpectrumDiagonal
+
+variable {R n M : Type*} [DecidableEq n][Fintype n]
```
j-loreaux, line 23:
```suggestion
variable {R n M : Type*} [DecidableEq n] [Fintype n]
```

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,93 @@
...
+  ⟨mem_eigenspace_iff.mpr <| by simp [diagonal], Basis.ne_zero b i⟩
+
+/--  Standard basis vectors are eigenvectors of any associated diagonal linear operator. -/
+lemma hasEigenvector_toLin'_diagonal (d : n → R) (i : n) :
+    HasEigenvector (toLin' (diagonal d)) (d i) (Pi.basisFun R n i)  :=
+  hasEigenvector_toLin_diagonal ..
+
+section NoZeroSMulDivisors
+
+variable [NoZeroSMulDivisors R M]
```
j-loreaux, line 44: I wouldn't create a section for one result, just add this hypothesis to the lemma

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
```diff
@@ -0,0 +1,93 @@
...
+lemma hasEigenvalue_toLin'_diagonal_iff (d : n → R) {μ : R} :
+    HasEigenvalue (toLin' (diagonal d)) μ ↔ (∃ i, d i = μ) :=
+  hasEigenvalue_toLin_diagonal_iff _ <| Pi.basisFun R n
+
+end NoZeroDivisors
+
+end NontrivialCommRing
+
+variable [Field R]
+
```
j-loreaux, line 85: Again, don't use `variable` for one declaration, just add the hypothesis to the lemma.

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
JonBannon: Shoot! I keep doing this!!!!

j-loreaux:
maintainer merge

Mathlib/LinearAlgebra/Eigenspace/Matrix.lean
dupuisf: Can you also adjust the docstring accordingly?

JonBannon: Certainly! I'll do this now...

PR #13840, opened 2024-06-14
[Merged by Bors] - feat (Category Theory): redefine and extend mates

emilyriehl:
The mates bijection between natural transformations inhabiting a certain dual pairs of squares (the duality in the sense of a parallel pair of adjunctions) can be defined in a more direct way: using pasting composition on 2-cells. 

This pull request:
(i) proposes a new definition of `Mates` to replace and rename the existing `transferNatTrans`
(ii) proposes a new definition of `Conjugates` to replace and rename the existing `transferNatTransSelf`
(iii) proves that the mates correspondence commutes with horizontal and vertical composition of squares, as well as some coherences relating mates to conjugates
(iv) updates the files that depends on this, involving two new proofs about exponentiation in cartesian closed categories

---

I have now refactored the rest of the library, so this pull request is ready for review.

I'd appreciate feedback on two aspects of the longer coherence proofs:

(i) Can this be done without calling "ext"? I originally tried to prove this using theorems about pasting composite of natural transformations but got stuck because I couldn't rewrite along equalities of functors of the form `1 >>> F = F`. It's somewhat unaesthetic, though, to prove this by passing to components.

(ii) The long proofs involve lots of `slice_rhs` calls. This feels fairly readable to me, but I'm curious if there is a better way. 

### DIFF
--- a/Mathlib/CategoryTheory/Adjunction/Mates.lean
+++ b/Mathlib/CategoryTheory/Adjunction/Mates.lean
@@ -75,8 +75,7 @@ Then we have a bijection between natural transformations `G ⋙ L₂ ⟶ L₁ �
 Note that if one of the transformations is an iso, it does not imply the other is an iso.
 -/
 @[simps]
-def Mates :
-    (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
+def mateEquiv : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
       toFun α :=
         whiskerLeft (R₁ ⋙ G) adj₂.unit ≫
         whiskerRight (whiskerLeft R₁ α) R₂ ≫
--- a/Mathlib/CategoryTheory/Adjunction/Mates.lean
+++ b/Mathlib/CategoryTheory/Adjunction/Mates.lean
@@ -74,44 +74,6 @@ Then we have a bijection between natural transformations `G ⋙ L₂ ⟶ L₁ �
 
 Note that if one of the transformations is an iso, it does not imply the other is an iso.
 -/
-
--- ER: The original definition of the mates equivalence.
-def transferNatTrans : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
-  toFun h :=
-    { app := fun X => adj₂.unit.app _ ≫ R₂.map (h.app _ ≫ H.map (adj₁.counit.app _))
-      naturality := fun X Y f => by
-        dsimp
-        rw [assoc, ← R₂.map_comp, assoc, ← H.map_comp, ← adj₁.counit_naturality, H.map_comp, ←
-          Functor.comp_map L₁, ← h.naturality_assoc]
-        simp }
-  invFun h :=
-    { app := fun X => L₂.map (G.map (adj₁.unit.app _) ≫ h.app _) ≫ adj₂.counit.app _
-      naturality := fun X Y f => by
-        dsimp
-        rw [← L₂.map_comp_assoc, ← G.map_comp_assoc, ← adj₁.unit_naturality, G.map_comp_assoc, ←
-          Functor.comp_map, h.naturality]
-        simp }
-  left_inv h := by
-    ext X
-    dsimp
-    simp only [L₂.map_comp, assoc, adj₂.counit_naturality, adj₂.left_triangle_components_assoc, ←
-      Functor.comp_map G L₂, h.naturality_assoc, Functor.comp_map L₁, ← H.map_comp,
-      adj₁.left_triangle_components]
-    dsimp
-    simp only [id_comp, ← Functor.comp_map, ← Functor.comp_obj, NatTrans.naturality_assoc]
-    simp only [Functor.comp_obj, Functor.comp_map, ← Functor.map_comp]
-    have : Prefunctor.map L₁.toPrefunctor (NatTrans.app adj₁.unit X) ≫
-        NatTrans.app adj₁.counit (Prefunctor.obj L₁.toPrefunctor X) = 𝟙 _ := by simp
-    simp [this]
-  -- See library note [dsimp, simp].
-  right_inv h := by
-    ext X
-    dsimp
-    simp [-Functor.comp_map, ← Functor.comp_map H, Functor.comp_map R₁, -NatTrans.naturality, ←
-      h.naturality, -Functor.map_comp, ← Functor.map_comp_assoc G, R₂.map_comp]
-#align category_theory.transfer_nat_trans CategoryTheory.transferNatTrans
-
-/-- Corresponding natural transformations under the equivalence below are called mates.-/
 @[simps]
 def Mates :
     (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
@@ -142,42 +104,12 @@ def Mates :
           ← G.map_comp, right_triangle_components, map_id, id_comp]
 #align category_theory.mates CategoryTheory.Mates
 
--- ER: Note these definitions agree.
-theorem RightMateEqualsTransferNatTrans
-    (α : G ⋙ L₂ ⟶ L₁ ⋙ H) :
-    Mates adj₁ adj₂ α = (transferNatTrans adj₁ adj₂) α := by
-  ext; unfold Mates transferNatTrans; simp
-
--- ER: Note these definitions agree.
-theorem LeftMateEqualsTransferNatTrans.symm
-    (α : R₁ ⋙ G ⟶ H ⋙ R₂) :
-    (Mates adj₁ adj₂).symm α = (transferNatTrans adj₁ adj₂).symm α := by
-  ext; unfold Mates transferNatTrans; simp
-
--- ER: Propose to cut and replace with the version below:
-theorem transferNatTrans_counit (f : G ⋙ L₂ ⟶ L₁ ⋙ H) (Y : D) :
-    L₂.map ((transferNatTrans adj₁ adj₂ f).app _) ≫ adj₂.counit.app _ =
-      f.app _ ≫ H.map (adj₁.counit.app Y) := by
-  erw [Functor.map_comp]
-  simp
-
 /-- A component of a transposed version of the mates correspondence. -/
 theorem Mates_counit (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (d : D) :
     L₂.map ((Mates adj₁ adj₂ α).app _) ≫ adj₂.counit.app _ =
       α.app _ ≫ H.map (adj₁.counit.app d) := by
   erw [Functor.map_comp]; simp
 
--- ER: Propose to cut and replace with the version below:
-theorem unit_transferNatTrans (f : G ⋙ L₂ ⟶ L₁ ⋙ H) (X : C) :
-    G.map (adj₁.unit.app X) ≫ (transferNatTrans adj₁ adj₂ f).app _ =
-      adj₂.unit.app _ ≫ R₂.map (f.app _) := by
-  dsimp [transferNatTrans]
-  rw [← adj₂.unit_naturality_assoc, ← R₂.map_comp, ← Functor.comp_map G L₂, f.naturality_assoc,
-    Functor.comp_map, ← H.map_comp]
-  dsimp; simp
-#align category_theory.unit_transfer_nat_trans CategoryTheory.unit_transferNatTrans
-
--- ER: I don't get why this is harder than the counit case.
 /- A component of a transposed version of the mates correspondence. -/
 theorem unit_Mates (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (c : C) :
     G.map (adj₁.unit.app c) ≫ (Mates adj₁ adj₂ α).app _ =
@@ -198,7 +130,6 @@ theorem unit_Mates (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (c : C) :
     }
   simp only [comp_obj, map_id, comp_id]
 
--- See library note [dsimp, simp]
 end Mates
 
 section MatesVComp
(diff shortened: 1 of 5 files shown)


Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -108,13 +111,58 @@ def transferNatTrans : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R
...
+    Mates adj₁ adj₂ α = (transferNatTrans adj₁ adj₂) α := by
+  ext; unfold Mates transferNatTrans; simp
+
+-- ER: Note these definitions agree.
+theorem LeftMateEqualsTransferNatTrans.symm
+    (α : R₁ ⋙ G ⟶ H ⋙ R₂) :
+    (Mates adj₁ adj₂).symm α = (transferNatTrans adj₁ adj₂).symm α := by
+  ext; unfold Mates transferNatTrans; simp
+
+-- ER: It's not clear to me what this theorem is for.
```
b-mehta, line 157: This theorem and the below (particularly in the Self case) are useful for proving things about the mate in applications - the self version for both are used to simplify proofs about cartesian closed-ness

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -127,6 +175,240 @@ theorem unit_transferNatTrans (f : G ⋙ L₂ ⟶ L₁ ⋙ H) (X : C) :
...
+
+def RightAdjointSquare.vcomp :
+    (R₁ ⋙ G₁ ⟶ H₁ ⋙ R₂) → (R₂ ⋙ G₂ ⟶ H₂ ⋙ R₃) →
+    (R₁ ⋙ (G₁ ⋙ G₂) ⟶ (H₁ ⋙ H₂) ⋙ R₃) := fun α β ↦
+  (whiskerRight α G₂) ≫ (whiskerLeft H₁ β)
+
+theorem Mates_vcomp
+  (α : G₁ ⋙ L₂ ⟶ L₁ ⋙ H₁)
+  (β : G₂ ⋙ L₃ ⟶ L₂ ⋙ H₂) :
+  (Mates (G := G₁ ⋙ G₂) (H := H₁ ⋙ H₂) adj₁ adj₃) (LeftAdjointSquare.vcomp α β)
```
b-mehta, line 203: Does Lean need G and H to be spelled out here to figure out what's going on? If so, maybe they should be explicit arguments in `Mates`

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -39,12 +41,11 @@ This also relates to Grothendieck's yoga of six operations, though this is not s
 mathlib: https://ncatlab.org/nlab/show/six+operations.
 -/
 
-
-universe v₁ v₂ v₃ v₄ u₁ u₂ u₃ u₄
+universe v₁ v₂ v₃ v₄ v₅ v₆ v₇ v₈ v₉ u₁ u₂ u₃ u₄ u₅ u₆ u₇ u₈ u₉
```
b-mehta, line 44: That's a lot of new universes! Can we have some of them declared in the section they're used in (I think it's only in composition?)

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -108,13 +111,58 @@ def transferNatTrans : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R
       h.naturality, -Functor.map_comp, ← Functor.map_comp_assoc G, R₂.map_comp]
 #align category_theory.transfer_nat_trans CategoryTheory.transferNatTrans
 
+-- ER: A new construction with a new name.
```
b-mehta, line 114: It's probably not worth having the construction twice, so we should pick one and use it everywhere (especially since you've already showed they're equal). It would be okay switching to yours, but that would mean you'd need to use it everywhere, which could mean changing existing proofs is a bit tricky

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -108,13 +111,58 @@ def transferNatTrans : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R
...
+def Mates :
+    (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
+      toFun := by
+        intro α
+        have R₁Gη₂ := whiskerLeft (R₁ ⋙ G) adj₂.unit
+        have R₁αR₂ := whiskerRight (whiskerLeft R₁ α) R₂
+        have ε₁HR₂ := whiskerRight adj₁.counit (H ⋙ R₂)
+        exact R₁Gη₂ ≫ R₁αR₂ ≫ ε₁HR₂
```
b-mehta, line 122:
```suggestion
      toFun := fun α => whiskerLeft (R₁ ⋙ G) adj₂.unit ≫ whiskerRight (whiskerLeft R₁ α) R₂ ≫ whiskerRight adj₁.counit (H ⋙ R₂)
```
writing it this way could save a `dsimp` when it comes to proofs - it's also closer to simp-normal form.
Similar for `invFun`.

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -108,13 +111,58 @@ def transferNatTrans : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R
       h.naturality, -Functor.map_comp, ← Functor.map_comp_assoc G, R₂.map_comp]
 #align category_theory.transfer_nat_trans CategoryTheory.transferNatTrans
 
+-- ER: A new construction with a new name.
+def Mates :
```
b-mehta, line 115: If you write `@[simps] def Mates :` here, then lean will automatically generate some lemmas describing the projections for you. (In fact, `transferNatTrans` should have been tagged simps too, but I don't think it existed when I wrote `transferNatTrans`!)

b-mehta (commented):
Thanks for this PR! Your definition certainly looks more readable, and you've proved it's equal to the old one so I'm convinced it's a good change.

Make sure to follow the lint style suggestions (more guidelines are here: https://leanprover-community.github.io/contribute/style.html).

If no-one beats me to it, I'll try to review your actual proofs in more detail soon. I think in *most* cases it's probably easiest to do `ext` and pass to components, although I agree it's not ideal. 
One thing that may help you is the reassoc_of% command, which can make rewriting an equality of morphisms inside a long chain quite a bit easier - it essentially lets you ignore associativity.

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: It seems like it needs it only sometimes, so I like keeping these implicit.

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: Yes if I have your approval, I was planning to do this.

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: Thanks I'll do this.

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: Thanks for the explanation. Would you mind if I renamed these as well (and propagated the change forward to other files?

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: I wasn't sure what the best convention is. All the universes at the top or just the necessary universes at the start of each section?

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: @b-mehta I've done this now. Let me know if you want me to change anything (eg the names).

emilyriehl:
This is now ready for review (as far as I can tell).

Mathlib/CategoryTheory/Closed/Functor.lean
```diff
@@ -107,13 +107,24 @@ theorem uncurry_expComparison (A B : C) :
...
-  rw [uncurry_natural_left, uncurry_natural_left, uncurry_expComparison, uncurry_pre,
-    prod.map_swap_assoc, ← F.map_id, expComparison_ev, ← F.map_id, ←
-    prodComparison_inv_natural_assoc, ← prodComparison_inv_natural_assoc, ← F.map_comp, ←
-    F.map_comp, prod_map_pre_app_comp_ev]
+  simp only [Functor.comp_obj, prod.functor_obj_obj, prodComparisonNatIso_inv, asIso_inv,
+    NatTrans.comp_app, whiskerLeft_app, prod.functor_map_app, NatIso.isIso_inv_app,
+    whiskerRight_app]
+  have piso := (prodComparison_inv_natural F f (𝟙 B)).symm
+  rw [← F.map_id]
+  exact piso
```
joelriou, line 127: This should work?
```suggestion
  rw [← F.map_id]
  exact (prodComparison_inv_natural F f (𝟙 B)).symm
```

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -73,199 +74,524 @@ This can be seen as a bijection of the 2-cells:
...
-  rw [← adj₂.unit_naturality_assoc, ← R₂.map_comp, ← Functor.comp_map G L₂, f.naturality_assoc,
-    Functor.comp_map, ← H.map_comp]
-  dsimp; simp
-#align category_theory.unit_transfer_nat_trans CategoryTheory.unit_transferNatTrans
-
--- See library note [dsimp, simp]
-end Square
+@[simps]
+def Mates :
+    (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
```
joelriou, line 79: Only names of types should be capitalised:
```suggestion
def mateEquiv : (G ⋙ L₂ ⟶ L₁ ⋙ H) ≃ (R₁ ⋙ G ⟶ H ⋙ R₂) where
```

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -73,199 +74,524 @@ This can be seen as a bijection of the 2-cells:
...
+      toFun := fun α ↦
```
joelriou, line 80: A slightly better syntax is:
```suggestion
      toFun α :=
```

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -73,199 +74,524 @@ This can be seen as a bijection of the 2-cells:
...
+        ext
+        unfold whiskerLeft whiskerRight
+        simp only [comp_obj, id_obj, Functor.comp_map, comp_app, map_comp, assoc,
+          unit_naturality_assoc, right_triangle_components_assoc]
+        rw [← assoc, ← Functor.comp_map, assoc, ← β.naturality, ← assoc, Functor.comp_map,
+          ← G.map_comp, right_triangle_components, map_id, id_comp]
+#align category_theory.mates CategoryTheory.Mates
+
+/-- A component of a transposed version of the mates correspondence. -/
+theorem Mates_counit (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (d : D) :
```
joelriou, line 108:
```suggestion
theorem mateEquiv_counit (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (d : D) :
```

Could you consider including similar (dual) lemmas like `(co)unit_mateEquiv_symm`?

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: Ah whoops. I'll get to work on this.

Mathlib/CategoryTheory/Adjunction/Mates.lean
emilyriehl: Done.

emilyriehl:
Note there is one (non-trivial) new definition in the latest version: `conjugateIsoEquiv`. I had to mark it non-computable since it uses asIso; can this be avoided?

This was included to use in applications such as the following:

```
namespace Over
universe v u
variable {C : Type u} [Category.{v} C]

/-- The conjugate isomorphism between pullback functors. -/
def pullbackCompIso [HasPullbacks C] {X Y Z : C} (f : X ⟶ Y) (g : Y ⟶ Z) :
    baseChange (f ≫ g) ≅ baseChange g ⋙ baseChange f :=
  conjugateIsoEquiv (mapAdjunction (f ≫ g)) ((mapAdjunction f).comp (mapAdjunction g)) (mapCompIso f g)
```

Here `mapCompIso` has type 

```
mapCompIso {X Y Z : C}(f : X ⟶ Y)(g : Y ⟶ Z) : Over.map f ⋙ Over.map g ≅ Over.map (f ≫ g)
```

For testing you could replace it with `(mathComp f g).symm` but I'm using a different isomorphism extracted from an equality between functors because it has better computational properties.

Mathlib/CategoryTheory/Adjunction/Mates.lean
```diff
@@ -73,199 +74,565 @@ This can be seen as a bijection of the 2-cells:
...
+theorem unit_mateEquiv (α : G ⋙ L₂ ⟶ L₁ ⋙ H) (c : C) :
+    G.map (adj₁.unit.app c) ≫ (mateEquiv adj₁ adj₂ α).app _ =
+      adj₂.unit.app _ ≫ R₂.map (α.app _) := by
+  dsimp [mateEquiv]
+  rw [← adj₂.unit_naturality_assoc]
+  slice_lhs 2 3 =>
+    {
+      rw [← R₂.map_comp, ← Functor.comp_map G L₂]
+      rw [α.naturality]
+    }
```
dagurtomas, line 127:
```suggestion
  slice_lhs 2 3 =>
    rw [← R₂.map_comp, ← Functor.comp_map G L₂, α.naturality]
```

Similarly below

dagurtomas (commented):
I just made a small suggestion about `slice` syntax

Mathlib/CategoryTheory/Adjunction/Mates.lean
dagurtomas: This syntax also works for multiple line tactic proofs

emilyriehl: Thanks @dagurtomas , I've propagated this suggestion throughout the file.

emilyriehl:
@jcommelin after our discussion today I realized I need to add deprecated alias tags. Should this be done for all the renamed theorems or just for the renamed definitions?

jcommelin:
@emilyriehl Concerning "deprecated": We don't have any strict rules yet. It's good practice to add these, because it will help projects that depend on mathlib when they migrate to a newer version. But if we estimate that these declarations aren't really used outside mathlib, then we might skip the process.
Otoh, I guess some of this stuff is being used in the Poly project, so the next mathlib bump in that project would benefit from having deprecated aliases in place.

emilyriehl:
@jcommelin just added for the two key definitions. We're already using this version of mates in poly so that's all set.

jcommelin (commented):
Thanks :tada:

bors merge

PR #13845, opened 2024-06-15
[Merged by Bors] - feat(EllipticCurve): affine formulas and bivariate polynomial lemmas

alreadydone:
+ Create a new file for materials about bivariate polynomials developed using `R[X][X]` rather than `MvPolynomial`. Introduce `evalEval` for evaluation at a point (x,y) on the plane, and use it in Affine.lean and Jacobian.lean.

+ Move the notations `R[X][Y]` and `Y` from `EllipticCurve/Afffine.lean` to the new file.

+ Add two lemmas in `Affine.lean` that relates `negPolynomial` and `polynomialY`, and two trivial lemmas `some_ne_zero` and `some_eq_some_iff`.

+ Prove two affine formulas `addX_eq_subX_sub` and `addY_sub_negY` which are crucial for the development of division polynomials.

+ Golf the definition of addition of rational points and subsequent lemmas, including two proofs in `Group.lean`.

+ Remove `eval_polynomial(X,Y)` lemmas in favor of bivariate polynomial lemma `map_mapRingHom_eval_map_eval`; add a lemma `Equation.map`.

---


Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -821,16 +864,16 @@ lemma map_polynomialY : (W.map f).toAffine.polynomialY = W.polynomialY.map (mapR
   simp only [polynomialY]
   map_simp
 
-lemma map_eval_polynomialY (x : R[X]) (y : R) :
+lemma map_polynomialY_eval_eval (x : R[X]) (y : R) :
     ((W.map f).toAffine.polynomialY.eval <| x.map f).eval (f y) =
       f ((W.polynomialY.eval x).eval y) := by
```
YaelDillies, line 869: Shouldn't this statement also use `evalEval`?

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -919,20 +962,20 @@ lemma baseChange_polynomialX : (W.baseChange B).toAffine.polynomialX =
     (W.baseChange A).toAffine.polynomialX.map (mapRingHom f) := by
   rw [← map_polynomialX, map_baseChange]
 
-lemma baseChange_eval_polynomialX (x : A[X]) (y : A) :
+lemma baseChange_polynomialX_eval_eval (x : A[X]) (y : A) :
     ((W.baseChange B).toAffine.polynomialX.eval <| x.map f).eval (f y) =
       f (((W.baseChange A).toAffine.polynomialX.eval x).eval y) := by
```
YaelDillies, line 967: and this one too?

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -0,0 +1,73 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Junyan Xu
+-/
+import Mathlib.Algebra.Polynomial.AlgebraMap
+
+/-!
+# Bivariate polynomials
+
+This file introduces the notation `R[X][Y]` for the polynomial ring `R[X][X]` in two variables,
+and the notation `Y` for the second variable, in the `PolynomialPolynomial` scope.
```
YaelDillies, line 12:
```suggestion
and the notation `Y` for the second variable, in the `Polynomial2` scope.
```
would be better maybe? It would match the existing `Sum3` scope which plays a similar role for `α ⊕ β ⊕ γ`.

YaelDillies (commented):
Looks fine, but what's the plan for when you start wanting trivariate polynomials? create a `PolynomialPolynomialPolynomial` scope?

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: The pattern of evalEval is `(W.polynomialY.eval (C y)).eval x` so this statement doesn't exactly match the pattern. (We do occasionally evaluate at something that is not `C y`, for example `linePolynomial`.) An `evalEval` version could be deduced from this, but I'd add it when it's actually needed.

Note: I'm not adding deprecated aliases because this was only added recently in #13356.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: ditto

Mathlib/Algebra/Polynomial/Bivariate.lean
alreadydone: I agree it's a good idea to rename (pending approval of @Multramate), though your analogy is not very exact (`Sum` only appears twice in `α ⊕ β ⊕ γ` so it could be appropriate to call it `Sum2` or `SumSum`).

alreadydone (commented):
Thanks for the review!

> Looks fine, but what's the plan for when you start wanting trivariate polynomials? create a `PolynomialPolynomialPolynomial` scope?

As you can see, we use `MvPolynomial (Fin 3) R` in the Jacobian (and projective) APIs. The choice to use `R[X][Y]` here is because the Weierstrass polynomial is a nice monic quadratic polynomial in the second variable `Y`, making it convenient to show the Weierstrass polynomial is irreducible and compute norms in the CoordinateRing, and we rely on the `AdjoinRoot` API to define the CoordinateRing, etc. (Also, division polynomials are bivariate polynomials, but many of them have equivalent univariate forms, and it's easier to convert between R[X][Y] and R[X].)

We can't exclude the possibility that someone will need `R[X][Y][Z]` in the future, so it does look saner to rename `PolynomialPolynomial` to `Polynomial2` so we can have `Polynomial3` when necessary. If @Multramate agrees I'll make the change.

It's a pity that `eval₂` is taken so we have to go with `evalEval` ... Using `eval2` would be too confusing, I think.

Mathlib/Algebra/Polynomial/Bivariate.lean
Multramate: I don't mind changing it to `Polynomial2` since `Sum3` exists. I would think it's unlikely that someone will create `R[X][Y][Z]` - I observed many months ago that working in this ring was really slow, but maybe that is a lot better now.

Multramate: Ah, I just saw your other comment - now I think that we should definitely change `evalEval` to `eval2` if we want to change `PolynomialPolynomial` to `Polynomial2`....

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +624,35 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
...
+lemma addX_eq_subX_sub :
+    W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂) = W.addX x₁ x₂ (W.slope x₁ x₂ y₁ (W.negY x₂ y₂))
+      - (y₁ - W.negY x₁ y₁) * (y₂ - W.negY x₂ y₂) / (x₂ - x₁) ^ 2 := by
+  simp_rw [slope, if_neg hx, addX, negY, ← neg_sub x₁, neg_sq]
+  field_simp [sub_ne_zero.mpr hx]; ring
+
+/-- The formula $y(P_1)(x(P_2)-x(P_3)) + y(P_2)(x(P_3)-x(P_1)) + y(P_3)(x(P_1)-x(P_2)) = 0$,
+assuming that $P_1 + P_2 + P_3 = O$. -/
+lemma cyclic_sum_Y_mul_X_sub_X :
+    letI x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
```
Multramate, line 640: `let`?

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +624,35 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
   nonsingular_neg <| nonsingular_negAdd h₁ h₂ hxy
 #align weierstrass_curve.nonsingular_add WeierstrassCurve.Affine.nonsingular_add
 
+variable {x₁ x₂ : F} (y₁ y₂ : F) (hx : x₁ ≠ x₂)
+
+/-- The formula $x(P_1 + P_2) = x(P_1 - P_2) - \psi(P_1)\psi(P_2) / (x(P_2) - x(P_1)) ^ 2$,
+where $\psi(x,y) = 2y + a_1 x + a_3$. -/
+lemma addX_eq_subX_sub :
```
Multramate, line 631: I don't like this name, because `subX` isn't a thing. We either should make that a thing, or change this to `addX_sub_addX_negY_sub` (and I don't know what to call the rest).

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -696,9 +731,8 @@ noncomputable def add : W.Point → W.Point → W.Point
   | 0, P => P
   | P, 0 => P
   | @some _ _ _ x₁ y₁ h₁, @some _ _ _ x₂ y₂ h₂ =>
-    if hx : x₁ = x₂ then
-      if hy : y₁ = W.negY x₂ y₂ then 0 else some <| nonsingular_add h₁ h₂ fun _ => hy
-    else some <| nonsingular_add h₁ h₂ fun h => (hx h).elim
+    if h : x₁ = x₂ ∧ y₁ = W.negY x₂ y₂ then 0
```
Multramate, line 734: You beat me to it! [Here is my incomplete branch](https://github.com/leanprover-community/mathlib4/compare/EllipticCurve.Chore) that does some of the golfing for affine point addition, but I haven't made a PR because I felt they were still somewhat inconsequential. Maybe you can have a look and get something out of it, and I'll update my branch with more chores in the future once this one is merged.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -797,22 +829,33 @@ lemma map_polynomial : (W.map f).toAffine.polynomial = W.polynomial.map (mapRing
...
     ((W.map f).toAffine.polynomial.eval <| x.map f).eval (f y) =
       f ((W.polynomial.eval x).eval y) := by
-  rw [map_polynomial, eval_map, ← coe_mapRingHom, eval₂_hom, coe_mapRingHom, eval_map, eval₂_hom]
+  rw [map_polynomial_eval, eval_map, eval₂_hom]
+
+lemma map_polynomial_evalEval (x y : R) :
+    (W.map f).toAffine.polynomial.evalEval (f x) (f y) = f (W.polynomial.evalEval x y) := by
+  rw [← map_polynomial_eval_eval, map_C]
+
+lemma Equation.map {x y : R} (h : W.Equation x y) : Equation (W.map f) (f x) (f y) := by
```
Multramate, line 845: Urgh, I just noticed that some of my `map` lemmas might not have the correct name: `map_*` should be a lemma of the form `f (...) = (f ...)`, but some of mine change the Weierstrass curve instead? I don't know what's the best naming convention, but maybe you can think of something that's analogous to the distinction between `Equation.map` and `map_polynomial_*`. Surely you should add the analogue of all of the new additions for `Nonsingular` too...

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -656,6 +685,12 @@ lemma zero_def : (zero : W.Point) = 0 :=
   rfl
 #align weierstrass_curve.point.zero_def WeierstrassCurve.Affine.Point.zero_def
 
+lemma some_ne_zero {x y : R} (h : W.Nonsingular x y) : some h ≠ 0 := by rintro (_|_)
+
+lemma some_eq_some_iff {x₁ x₂ y₁ y₂ : R} (h₁ : W.Nonsingular x₁ y₁) (h₂ : W.Nonsingular x₂ y₂) :
```
Multramate, line 690: Can we not get `some_eq_some_iff` from `ext` (as a tag in the structure or however else?)?

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -656,6 +685,12 @@ lemma zero_def : (zero : W.Point) = 0 :=
   rfl
 #align weierstrass_curve.point.zero_def WeierstrassCurve.Affine.Point.zero_def
 
+lemma some_ne_zero {x y : R} (h : W.Nonsingular x y) : some h ≠ 0 := by rintro (_|_)
```
Multramate, line 688: I think there might a couple of places in the files where I implicitly use `some_ne_zero`, but via the `contradiction` tactic instead. Can you check that these have all been changed?

Multramate (commented):
Thanks! Here are some preliminary suggestions.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: As I remarked [in the Jacobian lemmas PR](https://github.com/leanprover-community/mathlib4/pull/13846#discussion_r1641567888), the other naming convention is also used, e.g. in [Prod.map_injective](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Prod/Basic.html#Prod.map_injective) and many more `map_injective` lemmas.

What do you want me to add about `Nonsingular`? `Nonsingular.map` isn't necessarily true for non-injective RingHom.

Multramate: Fair enough!

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: `@[ext]` before the definition `inductive Point` produces the error
```
@[ext] attribute only applies to structures or theorems proving x = y, got Type u
```
so I guess it's not yet supported.
I seem to recall that there's some `mk_inj` stuff (in Lean 3?) but I can't find it now.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: I think I do want to inline this expression; it's just it appears multiple times so I'd like a shorthand. Without `letI` maybe there'll be issues rewriting with it, I'm not sure.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: It seems there are two appearances of `contradiction` in the `Affine` file; the appearance of `any_goals contradiction` in `map_injective` can be replaced but would make the proof longer, so I'd rather leave it as is.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: In this PR I'm just golfing the definition and fixing things, so I basically only applied forced golfs, having no intention to make the diff larger. Once this PR is accepted, your further golf of the lemmas should be easy to merge!

Mathlib/Algebra/Polynomial/Bivariate.lean
alreadydone: Don't you find `eval2` too confusing given [Polynomial.eval₂](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Polynomial/Eval.html#Polynomial.eval%E2%82%82)? Should we change `Polynomial2` back then @YaelDillies? Or should we make a poll on Zulip ...

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: I renamed this lemma and the last one in https://github.com/leanprover-community/mathlib4/pull/13845/commits/2d333ea40f58745d5aa4c5c999f24a5ba72990b8 but leave the `cyclic_sum` one unchanged.

I think I answered all questions; thanks for the review!

Mathlib/Algebra/Polynomial/Bivariate.lean
alreadydone: I guess we could also just put the notations under the Polynomial scope ...

YaelDillies:
Could you make sure it's clear from the PR title that this PR is mostly about bivariate polynomials? I'm sure most people will not look further if they think it's about elliptic curves.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: Actually, in the `universal` branch I'm now removing these lemmas in favor of bivariate polynomial lemmas
```
lemma map_mapRingHom_eval_map : (p.map <| mapRingHom f).eval (q.map f) = (p.eval q).map f

lemma map_mapRingHom_eval_map_eval (r : R) :
    ((p.map <| mapRingHom f).eval <| q.map f).eval (f r) = f ((p.eval q).eval r)

lemma map_mapRingHom_evalEval (x y : R) :
    (p.map <| mapRingHom f).evalEval (f x) (f y) = f (p.evalEval x y)
```

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -0,0 +1,96 @@
...
+
+This file introduces the notation `R[X][Y]` for the polynomial ring `R[X][X]` in two variables,
+and the notation `Y` for the second variable, in the `Polynomial` scope.
+
+It also defines `Polynomial.evalEval` for the evaluation of a bivariate polynomial at a point
+on the affine plane, which is a ring homomorphism (`Polynomial.evalEvalRingHom`), as well as
+the abbreviation `CC` to view a constant in the base ring `R` as a bivariate polynomial.
+-/
+
+/-- The notation `Y` for `X` in the `Polynomial2` scope. -/
```
YaelDillies, line 19:
```suggestion
/-- The notation `Y` for `X` in the `Polynomial` scope. -/
```
same below

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +624,35 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
...
+/-- The formula
+`ψ(P₁ + P₂) = (ψ(P₂)(x(P₁) - x(P₃)) - ψ(P₁)(x(P₂) - x(P₃))) / (x(P₂) - x(P₁))`,
+where `ψ(x,y) = 2y + a₁ x + a₃`. -/
+lemma addY_sub_negY_addY :
+    letI x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
+    letI y₃ := W.addY x₁ x₂ y₁ (W.slope x₁ x₂ y₁ y₂)
+    y₃ - W.negY x₃ y₃ =
+      ((y₂ - W.negY x₂ y₂) * (x₁ - x₃) - (y₁ - W.negY x₁ y₁) * (x₂ - x₃)) / (x₂ - x₁) := by
+  simp_rw [addY, negY, eq_div_iff (sub_ne_zero.mpr hx.symm)]
+  linear_combination 2 * W.cyclic_sum_Y_mul_X_sub_X y₁ y₂ hx
```
Multramate, line 654:
```suggestion
  linear_combination 2 * cyclic_sum_Y_mul_X_sub_X y₁ y₂ hx
```

Multramate (commented):
Here are some further comments on `Affine.lean`. I won't comment further on the changes for `Point.add` because I'll handle them in my chore PR.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +624,35 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
   nonsingular_neg <| nonsingular_negAdd h₁ h₂ hxy
 #align weierstrass_curve.nonsingular_add WeierstrassCurve.Affine.nonsingular_add
 
+variable {x₁ x₂ : F} (y₁ y₂ : F) (hx : x₁ ≠ x₂)
+
+/-- The formula `x(P₁ + P₂) = x(P₁ - P₂) - ψ(P₁)ψ(P₂) / (x(P₂) - x(P₁))²`,
+where `ψ(x,y) = 2y + a₁x + a₃`. -/
```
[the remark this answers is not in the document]
alreadydone: I find this more readable (in VS Code), so I switched from LaTeX to Unicode. I can remove the backticks if you prefer. (I know that backticks can are intended for code blocks in markdown but sometimes I just want to use it for the highlighted monospace style, not necessarily code.)

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -178,7 +173,7 @@ lemma irreducible_polynomial [IsDomain R] : Irreducible W.polynomial := by
 #align weierstrass_curve.irreducible_polynomial WeierstrassCurve.Affine.irreducible_polynomial
 
 -- Porting note (#10619): removed `@[simp]` to avoid a `simpNF` linter error
-lemma eval_polynomial (x y : R) : (W.polynomial.eval <| C y).eval x =
+lemma eval_polynomial (x y : R) : W.polynomial.evalEval x y =
```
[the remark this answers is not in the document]
alreadydone: Good idea, changed to `polynomial_evalEval_*` and added deprecation.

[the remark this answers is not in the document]
alreadydone: I won't touch the Jacobian and Projective files in this PR though, other than necessary fixes.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -293,8 +288,8 @@ lemma nonsingular_iff (x y : R) : W.Nonsingular x y ↔
 
 @[simp]
 lemma nonsingular_zero : W.Nonsingular 0 0 ↔ W.a₆ = 0 ∧ (W.a₃ ≠ 0 ∨ W.a₄ ≠ 0) := by
-  rw [Nonsingular, equation_zero, C_0, eval_polynomialX_zero, neg_ne_zero, eval_polynomialY_zero,
-    or_comm]
+  rw [Nonsingular, equation_zero, evalEval, evalEval, C_0, eval_polynomialX_zero, neg_ne_zero,
+    eval_polynomialY_zero, or_comm]
```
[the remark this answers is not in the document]
alreadydone: By the way, it's possible to directly prove `nonsingular_of_Δ_ne_zero` using the following formulas that I got from Ayad's paper:

$\Delta = 4 (36 x^3 + 9 b_2  x^2 + (42 b_4 - b_2^2) x + 27 b_6 - b_2 b_4) F + (-48 x^2 - 8 b_2 x + (b_2^2 - 32 b_4)) \phi_2 - (-12 x^3 + b_2  x^2 + 10 b_4 x + 27 b_6 - b_2 b_4) F_Y^2$
and $\phi_2 = (b_2 + 12 x) F - (2 x + a_2) F_Y^2 - a_1  F_X F_Y + F_X^2$,

where $F, F_X$ and $F_Y=\psi_2$ are W.polynomial, W.polynomialX and W.polynomialY.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: Thanks for testing! Removed the `I`

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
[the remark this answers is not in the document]
Multramate: Could you remind me why we prefer `polynomial_eval` over `eval_polynomial`? With dot notation we have `monic_polynomial` for `W.polynomial.Monic` but `polynomial_eval ...` for `W.polynomial.eval ...`.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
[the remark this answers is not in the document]
Multramate: I would argue that ultimately we want equations to be rendered best in mathlib4_docs. Personally I don't look at the code for equations, otherwise my module docstrings in e.g. `DivisionPolynomial/Basic` would be a huge mess.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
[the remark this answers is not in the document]
alreadydone: Good point. Originally I renamed `map_eval_polynomial` to `map_polynomial_eval` because the `map_*` lemmas (including auto-generated `map_a₁`) follows the inside-first, outside-last convention, so it's natural to rename `eval_polynomial` to `polynomial_eval` too. However, I've now removed `map_eval_polynomial`, so we don't have to stick to that convention when it comes to `polynomial` and it's better to follow the `monic`, `degree` lemmas etc. I'm going to rename to `evalEval_polynomial_*`.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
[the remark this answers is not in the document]
alreadydone: LaTeX won't render when you hover your mouse over the name of the lemma ...
This will still be pretty readable in the mathlib4_docs IMO, so I think it's a good compromise ...

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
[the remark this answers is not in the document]
Multramate: Thanks, and sorry for going back and forth!

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -797,40 +834,33 @@ lemma map_polynomial : (W.map f).toAffine.polynomial = W.polynomial.map (mapRing
   simp only [polynomial]
   map_simp
 
-lemma map_eval_polynomial (x : R[X]) (y : R) :
-    ((W.map f).toAffine.polynomial.eval <| x.map f).eval (f y) =
-      f ((W.polynomial.eval x).eval y) := by
-  rw [map_polynomial, eval_map, ← coe_mapRingHom, eval₂_hom, coe_mapRingHom, eval_map, eval₂_hom]
+lemma baseChange_polynomial_evalEval_X_Y :
+    (W.baseChange R[X][Y]).toAffine.polynomial.evalEval (C X) Y = W.polynomial := by
+  rw [baseChange, toAffine, map_polynomial, evalEval, eval_map, eval_C_X_eval₂_map_C_X]
```
Multramate, line 839: `baseChange_evalEval_polynomial`? Put this together with the other baseChange lemmas?

Multramate (approved):
I am happy with the changes in `AlgebraicGeometry/EllipticCurve`. I won't comment on `Algebra/Polynomial/Bivariate` as the naming is still up for discussion on Zulip.

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -0,0 +1,96 @@
...
+lemma coe_algebraMap_eq_CC [CommSemiring R] : algebraMap R R[X][Y] = CC (R := R) := rfl
+lemma coe_evalEvalRingHom [CommSemiring R] (x y : R) : evalEvalRingHom x y = evalEval x y := rfl
+
+lemma evalEvalRingHom_eq [CommSemiring R] (x : R) :
+    evalEvalRingHom x = eval₂RingHom (evalRingHom x) := by
+  ext <;> simp
+
+lemma eval₂_evalRingHom [CommSemiring R] (x y : R) (p : R[X][Y]) :
+    eval₂ (evalRingHom x) y p = evalEval x y p := by
+  rw [← coe_evalEvalRingHom, evalEvalRingHom_eq]; rfl
```
Multramate, line 50: Wouldn't this be more natural of a direction to state?
```
example (x y : R) (p : R[X][Y]) : p.evalEval x y = (p.map <| evalRingHom x).eval y
```
This says evaluating `p` at `y` then at `x` is the same as evaluating `p` at `x` then at `y`.

Multramate: Maybe this could be `evalEval_comm` or `evalEval_eq`, or `map_evalRingHom_eval` if it's reversed.

Multramate: I guess there's no harm in having both. You can replace the `rfl` with `coe_eval₂RingHom` too.

Multramate (commented):
It would be cool if we have some compatibility lemmas with `AdjoinRoot`:
```
variable {R : Type u} [CommRing R] {x y : R} {p : R[X][Y]} (h : p.evalEval x y = 0)

namespace AdjoinRoot

@[simps!]
def evalEval : AdjoinRoot p →+* R :=
  lift (evalRingHom x) y <| eval₂_evalRingHom x y p ▸ h

lemma evalEval_eq (g : R[X][Y]) : evalEval h (mk p g) = g.evalEval x y := by
  erw [AdjoinRoot.lift_mk, eval₂_evalRingHom]

end AdjoinRoot
```
Not quite sure where to add this though.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: I think `evalEval_ baseChange_polynomial_X_Y` is better ...

alreadydone:
In my refactored `universal` branch I have this for CoordinateRing:
```
variable {W} {x y : R} (eqn : W.Equation x y)

noncomputable def eval : W.CoordinateRing →+* R :=
  AdjoinRoot.lift (evalRingHom x) y <| by rwa [eval₂_evalRingHom]

lemma eval_mk (p : R[X][Y]) : eval eqn (mk _ p) = p.evalEval x y := by
  rw [← eval₂_evalRingHom]; exact AdjoinRoot.lift_mk _ p
```
Your version is more general and could fit in the Bivariate file (maybe with import AdjoinRoot added).

Mathlib/Algebra/Polynomial/Bivariate.lean
alreadydone: I added your version, which just differ by a `Polynomial.eval_map`. I consider `evalEval` to be the simp normal form so I put it on the RHS.

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: I'm not sure I should put it with other `baseChange` lemmas, since they're about general base changes but this is about a very specific one. I prefer to make it available immediately after its depency `map_polynomial` is proved.

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -0,0 +1,115 @@
...
+scoped[Polynomial] notation3:max R "[X][Y]" => Polynomial (Polynomial R)
+
+namespace Polynomial
+
+noncomputable section
+
+variable {R S : Type*}
+
+/-- `evalEval x y p` is the evaluation `p(x,y)` of a two-variable polynomial `p : R[X][Y]`. -/
+abbrev evalEval [Semiring R] (x y : R) (p : R[X][Y]) : R := eval x (eval (C y) p)
```
Multramate, line 32: Can you add this somewhere?
```
lemma evalEval_C (p : R[X]) : (C p).evalEval x y = p.eval x := by
  rw [evalEval, eval_C]
```
Thanks!

Multramate: And since we have this, why don't we add this too:
```
lemma evalEval_X : X.evalEval x y = y := by
  rw [evalEval, eval_X, eval_C]
```
We can try to mimic the API for normal `eval` and see which are relevant to `evalEval`.

Mathlib/Algebra/Polynomial/Bivariate.lean
Multramate: Thanks!

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
Multramate: Yeah, fair enough!

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +631,37 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
...
+    W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂) = W.addX x₁ x₂ (W.slope x₁ x₂ y₁ (W.negY x₂ y₂))
+      - (y₁ - W.negY x₁ y₁) * (y₂ - W.negY x₂ y₂) / (x₂ - x₁) ^ 2 := by
+  simp_rw [slope_of_X_ne hx, addX, negY, ← neg_sub x₁, neg_sq]
+  field_simp [sub_ne_zero.mpr hx]
+  ring1
+
+/-- The formula y(P₁)(x(P₂) - x(P₃)) + y(P₂)(x(P₃) - x(P₁)) + y(P₃)(x(P₁) - x(P₂)) = 0,
+assuming that P₁ + P₂ + P₃ = O. -/
+lemma cyclic_sum_Y_mul_X_sub_X :
+    let x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
```
ocfnash, line 648:
```suggestion
    letI x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
```

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
```diff
@@ -624,6 +631,37 @@ lemma nonsingular_add {x₁ x₂ y₁ y₂ : F} (h₁ : W.Nonsingular x₁ y₁)
...
+  simp_rw [slope_of_X_ne hx, negAddY, addX]
+  field_simp [sub_ne_zero.mpr hx]
+  ring1
+
+/-- The formula
+ψ(P₁ + P₂) = (ψ(P₂)(x(P₁) - x(P₃)) - ψ(P₁)(x(P₂) - x(P₃))) / (x(P₂) - x(P₁)),
+where ψ(x,y) = 2y + a₁x + a₃. -/
+lemma addY_sub_negY_addY :
+    let x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
+    let y₃ := W.addY x₁ x₂ y₁ (W.slope x₁ x₂ y₁ y₂)
```
ocfnash, line 659:
```suggestion
    letI x₃ := W.addX x₁ x₂ (W.slope x₁ x₂ y₁ y₂)
    letI y₃ := W.addY x₁ x₂ y₁ (W.slope x₁ x₂ y₁ y₂)
```

ocfnash (approved):
We seem to have a consensus that there are unsolved naming issues here but we haven't yet agreed on a solution.

However I don't think we need to hold up the whole PR until we figure this out since the current names are OK.

One minor suggestion but otherwise this LGTM.

bors d+

Mathlib/AlgebraicGeometry/EllipticCurve/Affine.lean
alreadydone: Thanks for the approval! I actually originally started with `letI`, but after @Multramate [showed](https://github.com/leanprover-community/mathlib4/pull/13845#discussion_r1644509949) that the lemmas are still good for rewrites even with `let`, I switched by his request. Do you still prefer `letI` given that `let` works?

ocfnash: I still prefer `letI` but no objection if you want to stick with `let`.

Multramate: I'm fine with `letI` (since I don't want to bother Junyan a third time), but I'm still very unconvinced about its use case. Is there a scenario where `letI` works but `let` definitely doesn't? It seems to me they're doing exactly the same thing.

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -0,0 +1,136 @@
...
+abbrev evalEval (x y : R) (p : R[X][Y]) : R := eval x (eval (C y) p)
+
+/-- A constant viewed as a polynomial in two variables. -/
+abbrev CC (r : R) : R[X][Y] := C (C r)
+
+lemma evalEval_C (x y : R) (p : R[X]) : (C p).evalEval x y = p.eval x := by
+  rw [evalEval, eval_C]
+
+lemma evalEval_X (x y : R) : X.evalEval x y = y := by
+  rw [evalEval, eval_X, eval_C]
```
Multramate, line 45: We should add the behaviour of evaluations over the usual operations in a separate PR. Here are some to begin with (because it seems that these are the only ones I need):
```
lemma evalEval_add {R : Type u} [Semiring R] (x y : R) (p q : R[X][Y]) :
    (p + q).evalEval x y = p.evalEval x y + q.evalEval x y := by
  simp only [evalEval, eval_add]

lemma evalEval_sub {R : Type u} [Ring R] (x y : R) (p q : R[X][Y]) :
    (p - q).evalEval x y = p.evalEval x y - q.evalEval x y := by
  simp only [evalEval, eval_sub]

lemma evalEval_mul {R : Type u} [CommSemiring R] (x y : R) (p q : R[X][Y]) :
    (p * q).evalEval x y = p.evalEval x y * q.evalEval x y := by
  simp only [evalEval, eval_mul]

lemma evalEval_pow {R : Type u} [CommSemiring R] (x y : R) (p : R[X][Y]) (n : ℕ) :
    (p ^ n).evalEval x y = p.evalEval x y ^ n := by
  simp only [evalEval, eval_pow]
```

alreadydone: You can also get these by rw [<- evalEvalRingHom] and then use the generic map_* lemmas ...

alreadydone: If you do want them you can submit a PR and see how it goes.

PR #13846, opened 2024-06-15
[Merged by Bors] - feat(EllipticCurve): lemmas in Jacobian coordinates

alreadydone:
+ `equiv_iff_eq_of_Z_eq`: if a point has two representations in Jacobian coordinates with the same, nonzero Z-coordinate, then the two representations are equal.

+ `nonsingular_iff_of_Z_ne_zero`: a lemma deleted in an earlier PR for no reason, now restored.

+ `addXYZ_self`: if the addition (not doubling) formula is applied to a point representative and itself, the result is (0,0,0).

+ `addXYZ_neg`: the addition formula applied to a point representative and its negation yields a representative of the point at infinity.

+ two trivial lemmas `fromAffine_ne_zero` and `comp_fin3`.

+ a series of `map` lemmas showing the addition and doubling formulas in Jacobian coordinates commute with ring homomorphisms.

---


Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1503,6 +1561,60 @@ end Point
 
 end Affine
 
+section Map
+
+lemma comp_fin3 {S} (f : R → S) (X Y Z : R) : f ∘ ![X, Y, Z] = ![f X, f Y, f Z] := by
+  ext i; fin_cases i <;> rfl
```
YaelDillies, line 1567: Can you either
* remove this lemma in favour of lemmas about `vecCons` and `vecEmpty`
* make this lemma `private`

YaelDillies (commented):
The rest looks good but I am not an expert

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: It's primarily used in a different file (`ZSMul.lean`) so I don't think it can be made private ...? We have four similar convenience lemmas starting from [WeierstrassCurve.Jacobian.fin3_def](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.html#WeierstrassCurve.Jacobian.fin3_def) and I don't think they're outrageous ...

Multramate: Yeah, I think this is sensible just for our purposes, as long we hide them in the `Jacobian` namespace.

Multramate: I would move this to right below `fin3_def` or `smul_fin3`, but it's not a huge deal. I will eventually split this file into smaller files and relocate a lot of things, so I might as well do that then.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1068,6 +1099,27 @@ lemma nonsingular_neg {P : Fin 3 → F} (hP : W.Nonsingular P) : W.Nonsingular <
   · simp only [neg_of_Z_ne_zero hPz, nonsingular_smul _ <| Ne.isUnit hPz,
       nonsingular_neg_of_Z_ne_zero hP hPz]
 
+lemma addZ_neg {P : Fin 3 → R} : addZ P (W'.neg P) = 0 := by
+  simp_rw [addZ, neg, fin3_def_ext, sub_self]
+
+lemma addX_neg {P : Fin 3 → R} (hP : W'.Equation P) : W'.addX P (W'.neg P) = W'.dblZ P ^ 2 := by
+  simp only [addX, neg, dblZ, negY, fin3_def_ext]
+  linear_combination -2 * P z ^ 2 * ((equation_iff _).mp hP)
```
Multramate, line 1107:
```suggestion
  linear_combination -2 * P z ^ 2 * (equation_iff _).mp hP
```

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -63,8 +63,8 @@ A point representative is implemented as a term `P` of type `Fin 3 → R`, which
 notation `![x, y, z]`. However, `P` is not syntactically equivalent to the expanded vector
 `![P x, P y, P z]`, so the lemmas `fin3_def` and `fin3_def_ext` can be used to convert between the
 two forms. The equivalence of two point representatives `P` and `Q` is implemented as an equivalence
-of orbits of the action of `R`, or equivalently that there is some unit `u` of `R` such that
-`P = u • Q`. However, `u • Q` is not syntactically equal to `![u ^ 2 * Q x, u ^ 3 * Q y, u * Q z]`,
+of orbits of the action of `Rˣ`, or equivalently that there is some unit `u` of `R` such that
+`P = u • Q`. However, `u • Q` is not syntactically equal to `![u² * Q x, u³ * Q y, u * Q z]`,
```
Multramate, line 67: I changed this because `²` and `³` are not Lean code.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -136,8 +136,8 @@ lemma smul_fin3_ext (P : Fin 3 → R) (u : R) :
 
 /-- The multiplicative action on a point representative. -/
 scoped instance instMulActionPoint : MulAction R <| Fin 3 → R where
-  one_smul _ := by simp only [smul_fin3, one_pow, one_mul, fin3_def]
-  mul_smul _ _ _ := by simp only [smul_fin3, mul_pow, mul_assoc, fin3_def_ext]
+  one_smul _ := by simp_rw [smul_fin3, one_pow, one_mul, fin3_def]
+  mul_smul _ _ _ := by simp_rw [smul_fin3, mul_pow, mul_assoc, fin3_def_ext]
```
Multramate, line 140: Surely `simp only` still works?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
YaelDillies: Why is my first option not viable?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1503,6 +1561,60 @@ end Point
...
+
+lemma comp_fin3 {S} (f : R → S) (X Y Z : R) : f ∘ ![X, Y, Z] = ![f X, f Y, f Z] := by
+  ext i; fin_cases i <;> rfl
+
+variable {S : Type*} [CommRing S] (f : R →+* S) (P Q : Fin 3 → R)
+
+protected lemma map_smul (u : R) : f ∘ (u • P) = f u • (f ∘ P) := by
+  ext i; fin_cases i <;> simp [smul_fin3]
+
+@[simp] lemma map_addZ : addZ (f ∘ P) (f ∘ Q) = f (addZ P Q) := by simp [addZ]
```
Multramate, line 1574: I think the direction should be `f (addZ P Q) = addZ (f ∘ P) (f ∘ Q)` based on `map` lemmas throughout mathlib, and I don't think you should make this `simp` in that case. Similarly for the other lemmas below this.

Multramate (commented):
Thanks! Here are some preliminary suggestions. I will look into the code later.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: I don't think we have to use Lean code in docstrings, and I very much prefer docstrings readable in both the editor and the online docs, so I sometimes enclose readable pseudocode rather than actual Lean code in backquotes. However I agree in this case it's probably better to use actual Lean code since you're explain a technical Lean issue, but my version could also be made into actual code by [adding an import](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/superscript.20parser/near/365151927).

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: Yeah, but `simp_rw` is shorter and allows you to see intermediate goal states, which is why I prefer it. I never tested if there are noticeable performance differences compared to `simp only`, and these two proofs are possibly too trivial for `simp_rw` to be beneficial?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: What are we supposed to do with the first option? It does seem that `comp_vecCons` and `comp_vecEmpty` are missing, but IMO that should be another PR, because I do want `comp_fin3` for precise rewrite using a single lemma, and its proofs can't be made shorter with the `comp_vecCons/Empty` lemmas.

(Lemmas like [WeierstrassCurve.Jacobian.smul_fin3](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.html#WeierstrassCurve.Jacobian.smul_fin3) can't be generalized to arbitrary vectors since different coordinate has different weighting (exponent).)

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: This is a good point; it seems I'm only using `map_addXYZ` and `map_dblXYZ` in the reverse direction, so it's probably good to reverse the direction and keep the `@[simp]`.

alreadydone: In https://github.com/leanprover-community/mathlib4/pull/13846/commits/1a5e80ac9c8729d761e69b44ab137535bb622b20 I switched the direction of the `map` lemmas except for those about `polynomial(X,Y,Z)`. Now `map_polynomial(X,Y,Z)` has `WeierstrassCurve.map` on LHS, in contrast to the other `map` lemmas, but they agree with the affine lemmas, and since they are not marked `@[simp]` now I think it's okay. I think their names are acceptable, since their are plenty of lemmas following this naming convention like [Prod.map_injective](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Prod/Basic.html#Prod.map_injective).

alreadydone: Hmm should I really have switched the directions? Now these `simp` lemmas are contrary to the `simp` direction of [WeierstrassCurve.map_a₁](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.html#WeierstrassCurve.map_a%E2%82%81) etc.

Multramate: This is why I'm hesitant on which direction simp should be. I don't know if the simp direction of `map_a*` is correct.

alreadydone: Well, those map_a* are generated by @[simps] so it would be hard to argue the direction isn't correct. I think I'll revert the commit ...

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
Multramate: Makes sense, but I'm too used to `simp only`. One of these days I'll change all my `simp only` into `simp_rw`'s.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
Multramate: Wow!

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1068,6 +1100,27 @@ lemma nonsingular_neg {P : Fin 3 → F} (hP : W.Nonsingular P) : W.Nonsingular <
...
+  linear_combination -2 * P z ^ 2 * (equation_iff _).mp hP
+
+lemma negAddY_neg {P : Fin 3 → R} (hP : W'.Equation P) :
+    W'.negAddY P (W'.neg P) = W'.dblZ P ^ 3 := by
+  simp only [negAddY, neg, dblZ, negY, fin3_def_ext]
+  linear_combination -2 * (2 * P y * P z ^ 3 + W'.a₁ * P x * P z ^ 4 + W'.a₃ * P z ^ 6)
+    * ((equation_iff _).mp hP)
+
+lemma addY_neg {P : Fin 3 → R} (hP : W'.Equation P) : W'.addY P (W'.neg P) = -W'.dblZ P ^ 3 := by
+  simp [addY, addX_neg hP, negAddY_neg hP, addZ_neg, negY, fin3_def_ext]
```
Multramate, line 1117: This reduces the need for `simp` but is two lines:
```suggestion
  rw [addY, addX_neg hP, negAddY_neg hP, negY_of_Z_eq_zero addZ_neg]
  rfl
```

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1068,6 +1100,27 @@ lemma nonsingular_neg {P : Fin 3 → F} (hP : W.Nonsingular P) : W.Nonsingular <
...
+
+lemma addXYZ_neg {P : Fin 3 → R} (hP : W'.Equation P) :
+    W'.addXYZ P (W'.neg P) = -W'.dblZ P • ![1, 1, 0] := by
+  simp_rw [addXYZ, addX_neg hP, addY_neg hP, addZ_neg, smul_fin3]
+  simp [(show Odd 3 by decide).neg_pow]
```
Multramate, line 1122: No need for `simp`:
```suggestion
  erw [addXYZ, addX_neg hP, addY_neg hP, addZ_neg, smul_fin3, neg_sq, mul_one,
    Odd.neg_pow <| by decide, mul_one, mul_zero]
```
Now that we have this, can you also add the analogous statement `add_neg` after `add_of_X_ne`?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1318,6 +1371,12 @@ lemma fromAffine_some [Nontrivial R] {X Y : R} (h : W'.toAffine.Nonsingular X Y)
     fromAffine (.some h) = ⟨(nonsingularLift_some ..).mpr h⟩ :=
   rfl
 
+lemma fromAffine_ne_zero [Nontrivial R] {X Y : R} (h : W'.toAffine.Nonsingular X Y) :
+    fromAffine (.some h) ≠ 0 := fun h0 ↦ by
+  erw [Point.ext_iff, Quotient.eq] at h0
+  obtain ⟨u, eq⟩ := h0
```
Multramate, line 1377:
```suggestion
  obtain ⟨u, eq⟩ := Quotient.eq.mp <| (Point.ext_iff ..).mp h0
```
Once we have the equivalence between affine and Jacobian coordinates, this follows from the fact that `.some` isn't `0`, but I guess you want this for non-trivial rings too?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -1503,6 +1562,53 @@ end Point
 
 end Affine
 
+section Map
+
+variable {S : Type*} [CommRing S] (f : R →+* S) (P Q : Fin 3 → R)
+
+protected lemma map_smul (u : R) : f ∘ (u • P) = f u • (f ∘ P) := by
+  ext i; fin_cases i <;> simp [smul_fin3]
```
Multramate, line 1570: Move this to below `comp_fin3` too? Since this is a general fact about `Fin 3`s.

Multramate (commented):
Some further comments.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -63,8 +63,8 @@ A point representative is implemented as a term `P` of type `Fin 3 → R`, which
 notation `![x, y, z]`. However, `P` is not syntactically equivalent to the expanded vector
 `![P x, P y, P z]`, so the lemmas `fin3_def` and `fin3_def_ext` can be used to convert between the
 two forms. The equivalence of two point representatives `P` and `Q` is implemented as an equivalence
-of orbits of the action of `R`, or equivalently that there is some unit `u` of `R` such that
-`P = u • Q`. However, `u • Q` is not syntactically equal to `![u ^ 2 * Q x, u ^ 3 * Q y, u * Q z]`,
+of orbits of the action of `Rˣ`, or equivalently that there is some unit `u` of `R` such that
+`P = u • Q`. However, `u • Q` is not syntactically equal to `![u² * Q x, u³ * Q y, u * Q z]`,
 so the lemmas `smul_fin3` and `smul_fin3_ext` can be used to convert between the two forms.
```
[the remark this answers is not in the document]
alreadydone: I like to reserve some empty spaces for occasional edits ... Do you have some automation for greedy line filling?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
eric-wieser: This lemma already exists in the reflection library

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -123,6 +123,9 @@ lemma fin3_def (P : Fin 3 → R) : ![P x, P y, P z] = P := by
 lemma fin3_def_ext (X Y Z : R) : ![X, Y, Z] x = X ∧ ![X, Y, Z] y = Y ∧ ![X, Y, Z] z = Z :=
   ⟨rfl, rfl, rfl⟩
 
+lemma comp_fin3 {S} (f : R → S) (X Y Z : R) : f ∘ ![X, Y, Z] = ![f X, f Y, f Z] := by
+  ext i; fin_cases i <;> rfl
```
eric-wieser, line 127: This is `FinVec.map_eq`

alreadydone: Thanks! It seems it can't be directly used to rewrite to the desired form, or to replace `comp_fin3` in a `simp` call, so I just changed the proof of `comp_fin3` to use it.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -548,9 +568,7 @@ lemma dblU_of_Z_eq_zero {P : Fin 3 → R} (hPz : P z = 0) : W'.dblU P = -3 * P x
 lemma dblU_ne_zero_of_Y_eq {P Q : Fin 3 → F} (hP : W.Nonsingular P) (hPz : P z ≠ 0) (hQz : Q z ≠ 0)
     (hx : P x * Q z ^ 2 = Q x * P z ^ 2) (hy : P y * Q z ^ 3 = Q y * P z ^ 3)
     (hy' : P y * Q z ^ 3 = W.negY Q * P z ^ 3) : W.dblU P ≠ 0 := by
-  rw [nonsingular_of_Z_ne_zero hPz, Affine.Nonsingular, ← equation_of_Z_ne_zero hPz,
-    ← eval_polynomialX_of_Z_ne_zero hPz, div_ne_zero_iff, and_iff_left <| pow_ne_zero 4 hPz,
-    ← eval_polynomialY_of_Z_ne_zero hPz, div_ne_zero_iff, and_iff_left <| pow_ne_zero 3 hPz,
+  rw [nonsingular_iff_of_Z_ne_zero hPz,
     show eval P W.polynomialY = P y - W.negY P by rw [negY, eval_polynomialY]; ring1,
     Y_eq_negY_of_Y_eq hQz hx hy hy', sub_self, ne_self_iff_false, or_false] at hP
   exact hP.right
```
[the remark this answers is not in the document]
alreadydone: Applied the change.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -390,6 +404,12 @@ lemma nonsingular_of_Z_ne_zero {P : Fin 3 → F} (hPz : P z ≠ 0) :
     W.Nonsingular P ↔ W.toAffine.Nonsingular (P x / P z ^ 2) (P y / P z ^ 3) :=
   (nonsingular_of_equiv <| equiv_some_of_Z_ne_zero hPz).trans <| nonsingular_some ..
 
+lemma nonsingular_iff_of_Z_ne_zero {P : Fin 3 → F} (hPz : P z ≠ 0) :
+    W.Nonsingular P ↔ W.Equation P ∧ (eval P W.polynomialX ≠ 0 ∨ eval P W.polynomialY ≠ 0) := by
+  rw [nonsingular_of_Z_ne_zero hPz, Affine.Nonsingular, ← equation_of_Z_ne_zero hPz,
+    ← eval_polynomialX_of_Z_ne_zero hPz, div_ne_zero_iff, and_iff_left <| pow_ne_zero 4 hPz,
+    ← eval_polynomialY_of_Z_ne_zero hPz, div_ne_zero_iff, and_iff_left <| pow_ne_zero 3 hPz]
```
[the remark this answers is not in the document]
Multramate: Ah sure, can you generalise it to that context then, if it's not too difficult?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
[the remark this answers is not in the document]
Multramate: Not really, it's just my habit to do it manually.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -123,8 +124,8 @@ lemma fin3_def (P : Fin 3 → R) : ![P x, P y, P z] = P := by
 lemma fin3_def_ext (X Y Z : R) : ![X, Y, Z] x = X ∧ ![X, Y, Z] y = Y ∧ ![X, Y, Z] z = Z :=
   ⟨rfl, rfl, rfl⟩
 
-lemma comp_fin3 {S} (f : R → S) (X Y Z : R) : f ∘ ![X, Y, Z] = ![f X, f Y, f Z] := by
-  ext i; fin_cases i <;> rfl
+lemma comp_fin3 {S} (f : R → S) (X Y Z : R) : f ∘ ![X, Y, Z] = ![f X, f Y, f Z] :=
+  (FinVec.map_eq _ _).symm
```
Multramate, line 128: Untested:
```suggestion
  (FinVec.map_eq ..).symm
```

Multramate: Since we already have `FinVec.map_eq`, do we still need `comp_fin3`? Otherwise, should this be `comp_fin3_ext`, and does it make sense to have an analogous `comp_fin3` for `P : Fin 3 -> R`?

Multramate (approved):
LGTM otherwise. I will likely move a few things around when I split the file, so I won't bother you further with minor golfs or rearrangements.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
alreadydone: As I [commented](https://github.com/leanprover-community/mathlib4/pull/13846#discussion_r1643750769), it can't be used to rewrite to the desired form directly.

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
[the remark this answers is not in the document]
alreadydone: I have it ready [here](https://github.com/leanprover-community/mathlib4/compare/Euler_homogeneous...EllCurve/Jacobian_homogeneous?expand=1#diff-fbad8e13903b8830a3972d6559954956a0838eb9c5729196639611aa39e181b6) but I prefer to merge this first.

jcommelin:
Thanks :tada:

bors merge

PR #13847, opened 2024-06-15
feat(EllipticCurve): the universal elliptic curve

alreadydone:
migrated to #41300

+ Define the universal Weierstrass curve (`Universal.curve`) over the polynomial ring `ℤ[A₁,A₂,A₃,A₄,A₆]`, and the universal pointed elliptic curve (`Universal.pointedCurve`) over the field of fractions (`Universal.Field`) of the universal ring `ℤ[A₁,A₂,A₃,A₄,A₆,X,Y]/⟨P⟩ = Universal.Poly/⟨P⟩` (`Universal.Ring`, where `P` is the Weierstrass polynomial) with distinguished point `(X,Y)`.

+ Given a Weierstrass curve `W` over a commutative ring `R`, we define the specialization homomorphism `W.specialize : ℤ[A₁,A₂,A₃,A₄,A₆] →+* R`. If `(x,y)` is a point on the affine plane, we define `W.polyEval x y : Universal.Poly →+* R`, which factors through `W.ringEval x y : Universal.Ring →+* R` if `(x,y)` is on `W`.

+ Introduce the cusp curve `Y² = X³`, on which lies the rational point `(1,1)`, with the nice property that `ψₙ(1,1) = n`, making it easy to prove nonvanishing of the universal `ψₙ` when `n ≠ 0` by specializing to the cusp curve, which shows that `(X,Y)` is a point of infinite order on the universal pointed elliptic curve.

---

- [x] depends on: #13845
- [x] depends on: #12883 


Multramate:
Thanks! I think I said somewhere (but I've lost it!) that I intend to combine stuff from this file with all the stuff to do with the coordinate ring, because I would think that these are "universal constructions" for any Weierstrass curve. I haven't had the time to do it so far, but can you give me some time to think about it? I'll make a subbranch (hopefully I don't mess it up again!).

alreadydone:
> Thanks! I think I said somewhere (but I've lost it!) that I intend to combine stuff from this file with all the stuff to do with the coordinate ring, because I would think that these are "universal constructions" for any Weierstrass curve. I haven't had the time to do it so far, but can you give me some time to think about it? I'll make a subbranch (hopefully I don't mess it up again!).

I think you said it [here](https://github.com/leanprover-community/mathlib4/pull/13399#issuecomment-2163491010). Yes I agree that for a Weierstrass curve W over comm. ring R you can define a "universal" Weierstrass curve over the coordinate ring R[W] with a distinguished point (X,Y) on it, such that every point (x,y) on W is the specialization of the distinguished point under the ring homomorphism R[W]->R sending X to x and Y to y. As I remarked [here](https://github.com/leanprover-community/mathlib4/blob/3d6da3dc779ad9a87bb94300bdc06f793e122d4e/Mathlib/AlgebraicGeometry/EllipticCurve/DivisionPolynomial/ZSMul.lean#L68-L72), you can almost prove the ZSMul formula using this less universal ring, but it would not be so easy to show ψₙ ≠ 0 when n ≠ 0 if the characteristic divides n, and in char 2 there's an additional issue.

I'll try to refactor my code with generalized definitions, thanks for mentioning this!

alreadydone:
@copilot resolve the merge conflicts in this pull request

alreadydone:
Superseded by #41300

PR #13848, opened 2024-06-15
fix(GetAllModules): sort file/module names in platform-independent order

grunweg:
Different platforms have different path separators, making the sorting of file names platform-dependent. To alleviate this, we simply sort their list of components instead.
This changes the sorting order slightly, but this is a detail which may change over time anyway.

---------

Alternative approach for #13777, for discussion.

---


grunweg:
Actually, as written over at #13777, I think best approach is to [not guarantee any sorting in `getAllFiles`, but not the sorting in `getAllModules` instead](https://github.com/leanprover-community/mathlib4/compare/MR-sort-modules-instead?expand=1).

grunweg:
Closing in favour of the other approach.

PR #13849, opened 2024-06-15
chore(checkYaml): remove supportInterpreter

grunweg:
It does not seem necessary.

---


grunweg:
As CI fails, it appears I was wrong. Closing.

PR #13850, opened 2024-06-15
[Merged by Bors] - fix: add `-lLake` argument to `pole` and `mk_all`

grunweg:
Since Lean 4.9, the behaviour of `supportInterpreter` executables has changed, making this necessary:
both of these executables require modules from `lake`.
In contrast, `shake` also uses `supportInterpreter`, but does not depend on `lake`.
[zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/automatically.20generate.20ProjectName.2Elean/near/444239597)

---

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -62,6 +62,7 @@ lean_exe checkYaml where
 lean_exe mk_all where
   srcDir := "scripts"
   supportInterpreter := true
+  weakLinkArgs := #["-lLake"]
 
 /-- `lake exe shake` checks files for unnecessary imports. -/
 lean_exe shake where
@@ -76,6 +77,7 @@ and then calculates the longest pole
 lean_exe pole where
   root := `LongestPole.Main
   supportInterpreter := true
+  weakLinkArgs := #["-lLake"]
 
 /--
 `lake exe test` is a thin wrapper around `lake exe batteries/test`, until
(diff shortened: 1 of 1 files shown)


grunweg:
[On zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/automatically.20generate.20ProjectName.2Elean), @Rida-Hamadani reports successfully testing this on windows.

Rida-Hamadani:
It succeeds only after deleting `.lake`, otherwise it gives an error.

lakefile.lean
```diff
@@ -62,6 +62,7 @@ lean_exe checkYaml where
 lean_exe mk_all where
   srcDir := "scripts"
   supportInterpreter := true
+  weakLinkArgs := #["-lLake"]
```
semorrison, line 65:
```suggestion
  -- Executables which import `Lake` must set `-lLake`.
  weakLinkArgs := #["-lLake"]
```

semorrison:
bors d+

lakefile.lean
```diff
@@ -76,6 +77,7 @@ and then calculates the longest pole
 lean_exe pole where
   root := `LongestPole.Main
   supportInterpreter := true
+  weakLinkArgs := #["-lLake"]
```
grunweg, line 80:
```suggestion
  -- Executables which import `Lake` must set `-lLake`.
  weakLinkArgs := #["-lLake"]

lakefile.lean
grunweg: Good idea, thanks - added.

grunweg:
Thanks for the review!
bors r+

PR #13854, opened 2024-06-15
[Merged by Bors] - fix(GetAllModules): sort module names instead of file names

grunweg:
Different platforms have different path separators, making the sorting of files by their names platform-dependent.
This meant the output `mk_all` would depend on which platform it was run on.
To avoid this, we sort the resulting module names instead --- for which the path separators have been converted to dots.

To make this difference clear, mention this in the documentation and rename to `getAllModulesSorted`.

---

### DIFF
--- a/Mathlib/Util/GetAllModules.lean
+++ b/Mathlib/Util/GetAllModules.lean
@@ -23,7 +23,7 @@ open Lean System.FilePath
 ```
 #[file₁, ..., fileₙ]
 ```
-of all their file names.
+of all their file names. These are not sorted in general.
 
 The input `git` is a `Bool`ean flag:
 * `true` means that the command uses `git ls-files` to find the relevant files;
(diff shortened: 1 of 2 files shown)


Mathlib/Util/GetAllModules.lean
```diff
@@ -39,18 +39,18 @@ def getAllFiles (git : Bool) (ml : String) : IO (Array System.FilePath) := do
     else do
       let all ← walkDir ml
       return all.filter (·.extension == some "lean"))
-  let files := (allModules.erase ml.lean).qsort (·.toString < ·.toString)
-  let existingFiles ← files.mapM fun f => do
+  let existingFiles ← (allModules.erase ml.lean).mapM fun f => do
```
adomani, line 42: Does `filterMapM` exist?  And can it be used here?

adomani (approved):

adomani:
Thanks for this: looks good to me!

I'll wait to see if there are more comments, but I think that this is an improvement!

Mathlib/Util/GetAllModules.lean
grunweg: It exists, TIL. I'll take a look!

grunweg: Used it here, thanks for the pointer!

adomani:
I'll just

maintainer merge

this, thanks!

adomani:
maintainer merge

bryangingechen:
Thanks!
bors r+

PR #13857, opened 2024-06-15
feat(AlgebraicGeometry/Curve): smooth morphisms of schemes

FMLJohn:
In this pull request, we aim to define the concept of a curve in scheme theory. Things achieved so far include:
1. given a scheme morphism `f : X ⟶ Y`, `Scheme.IsSmooth f` means that `f` is smooth.

This is a very immature draft: some declarations are not named according to the Mathlib4 naming convention; the comments for a number of definitions are not detailed enough; some proofs are much longer than necessary and look ugly; some definitions and results are not placed in appropriate files, etc. I will fix theses issues later.


joelriou:
For (formal) smoothness, I may favour more an definition which would be based on the right lifting property with respect to nilpotent immersions `Spec(B/I) ⟶ Spec(B)`. This way, almost by definition, a morphism `Spec(A) ⟶ Spec(R)` shall be formally smooth iff the `R`-algebra `A` is formally smooth. Then, we would have to show that formal smoothness is a local property, which requires more work (which may include showing the vanishing of $H^1$ of quasi-coherent sheaves over an affine scheme).

jcommelin:
@FMLJohn what's the status here? Are you planning to continue working on this PR? Would you like some help from others, or do you want to hand it over completely? (In the latter case, please label it with `please-adopt`.)

FMLJohn:
> @FMLJohn what's the status here? Are you planning to continue working on this PR? Would you like some help from others, or do you want to hand it over completely? (In the latter case, please label it with `please-adopt`.)

I wonder if the smoothness I have defined in this pull request is useful? I know that other people have already defined smoothness via the standard smoothness approach. So will my definition here be still useful? If not, then I guess this pull request is no longer meaningful and I will just close it.

Mathlib/AlgebraicGeometry/Curve/Smooth.lean
```diff
@@ -0,0 +1,78 @@
...
+
+In this file, we define when a morphism of schemes is smooth.
+-/
+
+/--
+Smooth homomorphisms between rings.
+-/
+def RingHom.IsSmooth {R A : Type _} [CommRing R] [CommRing A] (f : R →+* A) : Prop :=
+  @Algebra.Smooth R _ A _ (RingHom.toAlgebra f)
+
```
chrisflav, line 20: This should be moved to `Mathlib.RingTheory.Smooth.Basic`. Could you please also add some basic API (i.e. compositions of smooth maps are smooth, etc.)? You can take a look at https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/FinitePresentation.html#RingHom.FinitePresentation as a reference.

Mathlib/AlgebraicGeometry/Curve/Smooth.lean
```diff
@@ -0,0 +1,78 @@
...
+  val_base_closed : ClosedEmbedding f.val.base
+  epi : ∀ (x : Y.toPresheafedSpace),
+    Function.Surjective ((TopCat.Presheaf.stalkFunctor CommRingCat x).map f.val.c)
+
+/--
+Closed immersions between locally ringed spaces.
+-/
+abbrev IsClosedImmersion {X Y : Scheme} (f : X ⟶ Y) : Prop :=
+  LocallyRingedSpace.IsClosedImmersion f
+
```
chrisflav, line 39: We essentially have this as https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Morphisms/ClosedImmersion.html#AlgebraicGeometry.IsClosedImmersion and I don't think we necessarily need it for `LocallyRingedSpace`s, if you want to introduce it, then please open a separate PR for it.

Mathlib/AlgebraicGeometry/Curve/Smooth.lean
```diff
@@ -0,0 +1,78 @@
...
+section Smooth
+
+/--
+Given a scheme morphism `f : X ⟶ Y` and a point `x : X.toTopCat`, stating that `f` is smooth at
+`X`.
+-/
+class Scheme.IsSmoothAt {X Y : Scheme} (f : X ⟶ Y) (x : X.toTopCat) : Type _ where
+/- A ring whose prime spectrum is isomorphic to some open nbd of `x`. --/
+  R : CommRingCat
+/- A ring whose prime spectrum is isomorphic to some open subset of `Y` --/
```
chrisflav, line 51: I think we should instead use the "commutative algebra <-> AG" transport machinery to set this up. In particular you can use https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Morphisms/RingHomProperties.html#AlgebraicGeometry.affineLocally to produce a property of morphisms of schemes from a property on ring homomorphisms.

Also for this you can have a look at how it is done for morphisms locally of finite presentation, e.g. https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Morphisms/FinitePresentation.html#AlgebraicGeometry.LocallyOfFinitePresentation.

Note that showing https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/LocalProperties.html#RingHom.PropertyIsLocal for smooth ring maps is quite tricky and in the works by @erdOne, but it is still possible (and useful!) to make the definition and add a bit of API.

chrisflav (commented):
Good to see that there are more people interested in advancing smooth morphisms. To ensure seamless integration with the rest of the library, I added some suggestions on the code. It will probably be easier if you open separate PRs for the two points:
- adding `RingHom.IsSmooth` + basic API (see below)
- adding `AlgebraicGeometry.IsSmooth` + basic API (see below)

PR #12664, opened 2024-05-04
feat: add lemma about `degreeOf_eq_degree`

BoltonBailey:
Adds `MvPolynomial.degreeOf_eq_degree`, which states that the degree of a particular variable in a multivariate polynomial is equal to the degree of the single-variable polynomial obtained by treating the multivariable polynomial as a single variable polynomial over multivariable polynomials in the remaining variables.

needed by #11106

---

- [x] depends on: #13861 [So that we can prove `optionEquivLeft_support` analogously to `finSuccEquiv_support`]

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13861, opened 2024-06-15
feat(Data/Finsupp/Basic): `Finsupp.optionElim'`

BoltonBailey:
Similar to how `Finsupp.cons` constructs a map `Fin (n + 1) →₀ M` from a map `Fin n →₀ M`,
we define `Finsupp.optionElim'` to construct a map `Option α →₀ M` from a map `α →₀ M`, given an additional value for `none`. As a function, it behaves as `Option.elim'`, hence the name.

We prove a variety of API lemmas, based on those for `Finsupp.cons`, to bring the definitions more in line with the contents of `Data/Finsupp/Fin`. 

We also do some light refactoring for succinctness, and put this any related preexisting lemmas in a new file to not trigger the linter.

Needed by #12664

---


BoltonBailey:
This PR has been migrated to a fork-based workflow: https://github.com/leanprover-community/mathlib4/pull/25920

PR #13863, opened 2024-06-15
[Merged by Bors] - chore: use `SemilinearMapClass` in `Module.Finite.range`

urkud:
This way it automatically applies to `ContinuousLinearMap`s etc.

---


dupuisf:
Thanks!

bors r+

PR #13869, opened 2024-06-16
[Merged by Bors] - chore: distribute fun_prop tagging of measurability lemmas

grunweg:
into the respective files: essentially dissolves the `FunProp/Measurable` and `AEMeasurable` files.
(The latter is kept to note two missing statements. It is not imported anywhere any more.)

This is a pre-requisite for #13781, so fun_prop can be used without breaking existing the `assert_not_exists` statements.

---

Aside. The module docstring for Constructions/BorelSpace/Basic is outdated, `Measurable.add` is not mentioned there any more.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13770, opened 2024-06-12
[Merged by Bors] - perf(MeasureTheory,Probability): expand slow `measurabilty` calls

grunweg:
Manually expand slow `measurability` calls in various files: see individual commit messages for details.

From the benchmarking run, this seems to net -160*10⁹ instructions.
The `RadonNikodym` and `SignedMeasure` files each speed up by more than half (33 -> 18 billion resp. 100 -> 40 billion instructions). `Group/Convolution` is galvanised, 55 -> 8 billion instructions. The independence files become 10 billion instructions faster.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13871, opened 2024-06-16
[Merged by Bors] - chore/perf(Probability/MeasureTheory): use fun_prop for manual `measurability` calls

grunweg:
Replace manual `measurability` calls with `fun_prop`; replace `measurability` calls by `fun_prop` when possible.
As a by-product, this tags a few more lemmas `fun_prop` and resolves all porting notes involving `measurability`.

Follow- up to #13770.

---

- [x] depends on: #13869


grunweg:
!bench

Mathlib/MeasureTheory/MeasurableSpace/Basic.lean
```diff
@@ -772,7 +772,7 @@ theorem measurable_prod {f : α → β × γ} :
   ⟨fun hf => ⟨measurable_fst.comp hf, measurable_snd.comp hf⟩, fun h => Measurable.prod h.1 h.2⟩
 #align measurable_prod measurable_prod
 
-@[measurability]
+@[fun_prop, measurability]
```
grunweg, line 775: Should this one be tagged fun_prop, or could fun_prop infer this itself?

Mathlib/MeasureTheory/Decomposition/SignedLebesgue.lean
```diff
@@ -189,21 +189,15 @@ variable {s t : SignedMeasure α}
...
 #align measure_theory.signed_measure.measurable_rn_deriv MeasureTheory.SignedMeasure.measurable_rnDeriv
 
 theorem integrable_rnDeriv (s : SignedMeasure α) (μ : Measure α) : Integrable (rnDeriv s μ) μ := by
   refine Integrable.sub ?_ ?_ <;>
     · constructor
-      · -- NB: `measurability` proves this, but is very slow
-        -- TODO(#13864): reinstate faster automation, e.g. by making `fun_prop` work here
-        exact (((Measure.measurable_rnDeriv _ μ
-          ).ennreal_toNNReal).coe_nnreal_real).aestronglyMeasurable
+      · apply Measurable.aestronglyMeasurable
```
grunweg, line 199: Flagging this for the future: it would be nice to avoid the `Measurable.aestronglyMeasurable` lemma; `fun_prop` should infer this instead.

Mathlib/MeasureTheory/Constructions/BorelSpace/Metric.lean
```diff
@@ -237,7 +237,7 @@ section NormedAddCommGroup
 
 variable [MeasurableSpace α] [NormedAddCommGroup α] [OpensMeasurableSpace α] [MeasurableSpace β]
 
-@[measurability]
+@[fun_prop, measurability]
```
grunweg, line 240: I am not sure if `continuous_norm` should be tagged `fun_prop` instead.

grunweg (commented):
Three questions I asked myself: mainly about the `fun_prop` tagging.

Mathlib/MeasureTheory/Integral/CircleIntegral.lean
```diff
@@ -185,7 +185,7 @@ theorem continuous_circleMap (c : ℂ) (R : ℝ) : Continuous (circleMap c R) :=
   (differentiable_circleMap c R).continuous
 #align continuous_circle_map continuous_circleMap
 
-@[measurability]
+@[fun_prop, measurability]
```
grunweg, line 188: Perhaps `continuous_circleMap` should get tagged `fun_prop` instead...

grunweg:
@RemyDegenne This is the follow-up PR I promised; I guess you're interested. It does even yield some slight speed-ups.

Mathlib/MeasureTheory/Constructions/BorelSpace/Metric.lean
lecopivo: Not documented yet, I'm working on documentation. Right now, keep on pinging me to I can see what kind of questions come up and I can include them in the doc. 

Yes it can be inferred from continuity but providing specialized measurability theorem will make proofs about norm faster as `fun_prop` does not have to do as many steps.

grunweg: Thanks for the quick answer!

grunweg:
!bench

grunweg:
Comparing the two benchmarking runs, the second one is a minor improvement (save for `CircleIntegral`, which is very slightly slower - I would say this is a fair trade-off for the increased automation).

Mathlib/MeasureTheory/Decomposition/SignedLebesgue.lean
j-loreaux: Just to be clear, in order for that to work, we need to mark `AEStronglyMeasurable` (the definition) with the `fun_prop` attribute.

Mathlib/MeasureTheory/Decomposition/SignedLebesgue.lean
```diff
@@ -315,13 +309,13 @@ private theorem eq_singularPart' (t : SignedMeasure α) {f : α → ℝ} (hf : M
     JordanDecomposition.toSignedMeasure]
   congr
   -- NB: `measurability` proves this `have`, but is slow.
-  -- TODO(#13864): reinstate faster automation, e.g. by making `fun_prop` work here
+  -- XXX: fun_prop doesn't work yet...
```
j-loreaux, line 312: `XXX` is pretty useless as an indicator. Can we just leave `TODO`?

Also, flagging this for myself to understand in depth why this doesn't work.

grunweg: Sure - obviously, my preferred outcome is to make `fun_prop` work. If fixing that is for another PR, feel free to apply the following:
```suggestion
  -- TODO: make `fun_prop` able to handle this

jcommelin:
bors d+

grunweg:
Thanks for the review. Since CI passes, let me
bors r+

PR #13872, opened 2024-06-16
[Merged by Bors] - feat (MeasureTheory/Function): trivial cases of integrability / summability

loefflerd:
Any function on a finite type is multipliable / summable; a function on a discrete measurable space is integrable iff its norm is summable.

---

Excised from #13353

### DIFF
--- a/Mathlib/MeasureTheory/Function/L1Space.lean
+++ b/Mathlib/MeasureTheory/Function/L1Space.lean
@@ -713,7 +713,7 @@ lemma integrable_left_of_integrable_add_of_nonneg {f g : α → ℝ}
     (h_int : Integrable (f + g) μ) : Integrable f μ := by
   refine h_int.mono' h_meas ?_
   filter_upwards [hf, hg] with a haf hag
-  exact (Real.norm_of_nonneg haf).symm ▸ (le_add_iff_nonneg_right _).mpr hag
+  exact (Real.norm_of_nonneg haf).symm ▸ le_add_of_nonneg_right hag
 
 lemma integrable_right_of_integrable_add_of_nonneg {f g : α → ℝ}
     (h_meas : AEStronglyMeasurable f μ) (hf : 0 ≤ᵐ[μ] f) (hg : 0 ≤ᵐ[μ] g)
--- a/Mathlib/MeasureTheory/Function/L1Space.lean
+++ b/Mathlib/MeasureTheory/Function/L1Space.lean
@@ -713,7 +713,7 @@ lemma integrable_left_of_integrable_add_of_nonneg {f g : α → ℝ}
     (h_int : Integrable (f + g) μ) : Integrable f μ := by
   refine h_int.mono' h_meas ?_
   filter_upwards [hf, hg] with a haf hag
-  exact (Real.norm_of_nonneg haf).symm ▸ le_add_of_nonneg_right hag
+  exact (Real.norm_of_nonneg haf).symm ▸ (le_add_iff_nonneg_right _).mpr hag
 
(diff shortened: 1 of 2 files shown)


loefflerd:
(just realised my PR was accidentally reverting a change someone else had made to this file in the last few days. Now fixed.)

Mathlib/MeasureTheory/Function/L1Space.lean
```diff
@@ -939,6 +939,43 @@ theorem coe_toNNReal_ae_eq {f : α → ℝ≥0∞} (hf : ∀ᵐ x ∂μ, f x < 
   simp only [hx.ne, Ne, not_false_iff, coe_toNNReal]
 #align measure_theory.coe_to_nnreal_ae_eq MeasureTheory.coe_toNNReal_ae_eq
 
+section count
+
+variable [MeasurableSingletonClass α] {f : α → β}
+
+/-- A function has finite integral for the counting measure iff its norm is summable. -/
+lemma hasFiniteIntegral_count_iff :
+    HasFiniteIntegral f Measure.count ↔ Summable (‖f ·‖) := by
```
YaelDillies, line 948: Could you provide this lemma (and the ones below) for `Measure.average` as well?

Mathlib/Topology/Algebra/InfiniteSum/Basic.lean
```diff
@@ -132,6 +132,10 @@ protected theorem Set.Finite.multipliable {s : Set β} (hs : s.Finite) (f : β 
 theorem multipliable_of_finite_mulSupport (h : (mulSupport f).Finite) : Multipliable f := by
   apply multipliable_of_ne_finset_one (s := h.toFinset); simp
 
+@[to_additive]
+lemma Finite.multipliable [Finite β] (f : β → α) : Multipliable f :=
```
YaelDillies, line 136: I think
```suggestion
lemma Multipliable.of_finite [Finite β] {f : β → α} : Multipliable f :=
```
is more useful in that it lets you write `some_theorem .of_finite` to get rid of the side condition.

Mathlib/MeasureTheory/Function/L1Space.lean
loefflerd: Actually it looks like I was misunderstanding in our zulip discussion: `MeasureTheory.average` is not a measure in its own right, but an operator on functions defined as `(μ univ).toReal⁻¹ • ∫ x, f x ∂μ`. So I think there isn't an analogue.

YaelDillies: Oh I see 😢

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/MeasureTheory/Function/L1Space.lean
```diff
@@ -939,6 +939,43 @@ theorem coe_toNNReal_ae_eq {f : α → ℝ≥0∞} (hf : ∀ᵐ x ∂μ, f x < 
...
+  -- Note: this proof would be much easier if we assumed `SecondCountableTopology G`. Without
+  -- this we have to justify the claim that `f` lands a.e. in a separable subset, which is true
+  -- (because summable functions have countable range) but slightly tedious to check.
+  rw [Integrable, hasFiniteIntegral_count_iff, and_iff_right_iff_imp]
+  intro hs
+  have hs' : (Function.support f).Countable := by
+    simpa only [Ne, Pi.zero_apply, eq_comm, Function.support, norm_eq_zero]
+      using hs.countable_support
+  letI : MeasurableSpace β := borel β
+  haveI : BorelSpace β := ⟨rfl⟩
```
MichaelStollBayreuth, line 964: I don't think the "I" does make a difference here.

PR #13877, opened 2024-06-16
[Merged by Bors] - chore: Improved nth_rewrite and nth_rw docstrings

Shreyas4991:
[Zulip thread of discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/docstrings.20of.20.60nth_rewrite.60.20and.20.60nth_rw.60/near/444853793)

---

### DIFF
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -18,8 +18,8 @@ namespace Mathlib.Tactic
 open Lean Elab Tactic Meta Parser.Tactic
 
 /-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.  `nth_rewrite n [eq₁, eq₂,...]` will rewrite the `n`ᵗʰ occurrence
-of the equalities `eq₁`, `eq₂`,... etc.
+to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ occurrence
+of each of the `m` equalities `eqᵢ`.
 
 Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
 mathlib3. The translation will be handled by mathport. -/
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -17,11 +17,13 @@ namespace Mathlib.Tactic
 
 open Lean Elab Tactic Meta Parser.Tactic
 
-/-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ occurrence
-of each of the `m` equalities `eqᵢ`.
+/-- `nth_rewrite` is a variant of `rewrite` that only changes the `n`ᵗʰ _occurrence_ of the
+expression to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
+of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
+If a term `t` is introduced by rewriting with `eqᵢ`, then this instance of `t` will be counted
+as an _occurrence_ of `t`for all subsequent rewrites of `t` with `eqⱼ` for `j > i`.
 
-Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
+Note : The occurrences are counted beginning with `1` and not `0`, this is different than in
 mathlib3. The translation will be handled by mathport. -/
 syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (location)? : tactic
 
@@ -43,8 +45,11 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
 
 /--
 `nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
-`nth_rw n [eq₁, eq₂,...]` rewrites the `n`ᵗʰ occurrence of each equality `eq₁`, `eq₂`,... etc.
-Occurrences are counted beginning with `1`.
+`nth_rw n [eq₁, eq₂,..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_ of each of the `m` equalities
+`eqᵢ`in that order. Occurrences are counted beginning with `1`. If a term `t` is introduced
+by rewriting with `eqᵢ`, then this instance of `t` will be counted as an _occurrence_ of `t` for all
+subsequent rewrites of `t` with `eqⱼ` for `j > i`. Further, `nth_rw` will close the remaining goal
+with `rfl` if possible.
 -/
 macro (name := nthRwSeq) "nth_rw" c:(config)? ppSpace n:num s:rwRuleSeq l:(location)? : tactic =>
   -- Note: This is a direct copy of `nth_rw` from core.
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -43,7 +43,7 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
 
 /--
 `nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
-`nth_rw n [eq₁, eq₂,...]` rewrites the `nᵗʰ` occurrence of each equality `eq₁`, `eq₂`,... etc.
+`nth_rw n [eq₁, eq₂,...]` rewrites the `n`ᵗʰ occurrence of each equality `eq₁`, `eq₂`,... etc.
 Occurrences are counted beginning with `1`.
 -/
 macro (name := nthRwSeq) "nth_rw" c:(config)? ppSpace n:num s:rwRuleSeq l:(location)? : tactic =>
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -23,7 +23,7 @@ of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted begi
 If a term `t` is introduced by rewriting with `eqᵢ`, then this instance of `t` will be counted
 as an _occurrence_ of `t`for all subsequent rewrites of `t` with `eqⱼ` for `j > i`.
 
-Note : The occurrences are counted beginning with `1` and not `0`, this is different than in
+Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
 mathlib3. The translation will be handled by mathport. -/
 syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (location)? : tactic
 
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -18,7 +18,8 @@ namespace Mathlib.Tactic
 open Lean Elab Tactic Meta Parser.Tactic
 
 /-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.
+to be rewritten.  `nth_rewrite n [eq₁, eq₂,...]` will rewrite the `n`ᵗʰ occurrence
+of the equalities `eq₁`, `eq₂`,... etc.
 
 Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
 mathlib3. The translation will be handled by mathport. -/
@@ -41,7 +42,9 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
   | _ => throwUnsupportedSyntax
 
 /--
-`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
+`nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
+`nth_rw n [eq₁, eq₂,...]` rewrites the `nᵗʰ` occurrence of each equality `eq₁`, `eq₂`,... etc.
+Occurrences are counted beginning with `1`.
 -/
 macro (name := nthRwSeq) "nth_rw" c:(config)? ppSpace n:num s:rwRuleSeq l:(location)? : tactic =>
   -- Note: This is a direct copy of `nth_rw` from core.
--- a/Mathlib/Tactic/NthRewrite.lean
+++ b/Mathlib/Tactic/NthRewrite.lean
@@ -18,8 +18,10 @@ namespace Mathlib.Tactic
 open Lean Elab Tactic Meta Parser.Tactic
 
 /-- `nth_rewrite` is a variant of `rewrite` that only changes the `n`ᵗʰ _occurrence_ of the
-expression to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
-of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
+expression to be rewritten. `nth_rewrite n [eq₁, eq₂,..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
+of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1` in
+order of precedence.
+
 For example,
 ```lean
 example (h : a = 1) : a + a + a + a + a = 5 := by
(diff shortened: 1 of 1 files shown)


Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +42,9 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
   | _ => throwUnsupportedSyntax
 
 /--
-`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
+`nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
+`nth_rw n [eq₁, eq₂,...]` rewrites the `nᵗʰ` occurrence of each equality `eq₁`, `eq₂`,... etc.
```
eric-wieser, line 46:
```suggestion
`nth_rw n [eq₁, eq₂,...]` rewrites the `n`ᵗʰ occurrence of each equality `eq₁`, `eq₂`,... etc.
```
to match your other docstring.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -18,7 +18,8 @@ namespace Mathlib.Tactic
 open Lean Elab Tactic Meta Parser.Tactic
 
 /-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.
+to be rewritten.  `nth_rewrite n [eq₁, eq₂,...]` will rewrite the `n`ᵗʰ occurrence
+of the equalities `eq₁`, `eq₂`,... etc.
```
eric-wieser, line 22:
```suggestion
to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ occurrence
of each of the `m` equalities `eqᵢ`.
```
(which is a bit like what the `simp` docstring uses)

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,10 +17,13 @@ namespace Mathlib.Tactic
...
-/-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.
+/-- `nth_rewrite` is a variant of `rewrite` that only changes the `n`ᵗʰ _occurrence_ of the
+expression to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
+of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
+If a term `t` is introduced by rewriting with `eqᵢ`, then this instance of `t` will be counted
+as an _occurrence_ of `t`for all subsequent rewrites of `t` with `eqⱼ` for `j > i`.
 
-Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
+Note : The occurrences are counted beginning with `1` and not `0`, this is different than in
```
eric-wieser, line 26:
```suggestion
Note: The occurrences are counted beginning with `1` and not `0`, this is different than in
```

rwst (approved):
The whole change is much welcome. As a new user I can understand the text, and Grammarl y only complains about "afterward(s)".

Shreyas4991:
That's some Americanism. "Afterwards" is the correct version in English. 

See : https://dictionary.cambridge.org/thesaurus/articles/afterwards

Ruben-VandeVelde (commented):
Thanks!

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +44,12 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
   | _ => throwUnsupportedSyntax
 
 /--
-`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
+`nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
+`nth_rw n [eq₁, eq₂,..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_ of each of the `m` equalities
+`eqᵢ`in that order. Occurrences are counted beginning with `1`. If a term `t` is introduced
+by rewriting with `eqᵢ`, then this instance of `t` will be counted as an _occurrence_ of `t` for all
+subsequent rewrites of `t` with `eqⱼ` for `j > i`. Further, `nth_rw` will close the remaining goal
+with `rfl` if possible.
```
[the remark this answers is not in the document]
Shreyas4991: I think a bit of redundancy is better since the docstring is not a one-liner.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,8 +17,11 @@ namespace Mathlib.Tactic
 
 open Lean Elab Tactic Meta Parser.Tactic
 
-/-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.
+/-- `nth_rewrite` is a variant of `rewrite` that only changes the `n`ᵗʰ _occurrence_ of the
+expression to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
+of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
```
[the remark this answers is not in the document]
Shreyas4991: I added them and also some examples.

Mathlib/Tactic/NthRewrite.lean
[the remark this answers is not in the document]
Shreyas4991: I added some examples which added lines to the docstring. So the redundancy helps even more.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +78,45 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
...
+/-
+a b: ℕ
+h: a = a + b
+⊢ a + a + (a + b) + a + a = 0
+-/
+```
+This new instance of `a` also turns out to be the third _occurrence_ of `a`.  Therefore,
+the next `nth_rw` with `h` rewrites this `a`.
+
+Further, `nth_rw` will close the remaining goal with `rfl` if possible.
```
joneugster, line 119: I would suggest to `import Mathlib.Tactic.ExtendDoc` and then use `@[inherit_doc nthRewriteSeq]` on the`nth_rw` macro along with the something like the following just below it:

```
extend_docs nthRwSeq before
  "`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
  "
```
That way you could avoid writing that much documentation twice. The user will first see the sentense about `nth_rw` being essentially `nth_rewrite` followed by the doc for the latter.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,48 @@ namespace Mathlib.Tactic
 
 open Lean Elab Tactic Meta Parser.Tactic
 
-/-- `nth_rewrite` is a variant of `rewrite` that only changes the nth occurrence of the expression
-to be rewritten.
+/-- `nth_rewrite` is a variant of `rewrite` that only changes the `n`ᵗʰ _occurrence_ of the
+expression to be rewritten. `nth_rewrite n [eq₁, eq₂, ..., eqₘ]` will rewrite the `n`ᵗʰ _occurrence_
+of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
+Occurrences are counted beginning with `1`. For example,
```
joneugster, line 23:
```suggestion
of each of the `m` equalities `eqᵢ`in that order. Occurrences are counted beginning with `1`.
For example,
```
(sentence appears twice)

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,48 @@ namespace Mathlib.Tactic
...
+```lean
+example (h : a = 1) : a + a + a + a + a = 5 := by
+  nth_rewrite 2 [h]
+/-
+a: ℕ
+h: a = 1
+⊢ a + 1 + a + a + a = 5
+-/
+```
+Notice that the second occurrence of `a` from the left has been rewritten by `nth_rewrite`.
```
joneugster, line 33: "from the left" is not 100% correct in my opinion, maybe you could also add the following example to explain why it is the 2nd occurence and not the 1st:

```
example (a b c : Nat) : (a + b) + c = (b + a) + c := by
  nth_rewrite 2 [Nat.add_comm]
  rfl
```

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,48 @@ namespace Mathlib.Tactic
...
+the goal. That is, the goal state after the first rewrite looks like below
+```lean
+/-
+a b: ℕ
+h: a = a + b
+⊢ a + a + (a + b) + a + a = 0
+-/
+```
+This new instance of `a` also turns out to be the third _occurrence_ of `a`.  Therefore,
+the next `nth_rewrite` with `h` rewrites this `a`.
```
joneugster, line 57: This part I find quite hard to read, and I do find this behaviour intuitive as is, not requiring much explanation.

Maybe this entire paragraph could be shortened to something like the following?

> `nth_rewrite` with multiple arguments recalculates the occurences after each step newly, so for `(h : a = a + b)`, `nth_rewrite 3 [h, h]` changes `a + a + a + a + a = 0` to `a + a + (a + b) + a + a = 0` in the first step and then to `a + a + ((a + b) + b) + a + a = 0` in the second.

Shreyas4991: In a previous iteration I had such a short sentence. The problem is if someone is not comfortable with binder substitution and name capture issues, then it is a bit hard for them to see what is happening without a detailed example.

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: Ah yes, this is true. Technically `nth_rewrite` looks at expression trees. I am not sure how convey this intuitively without referring to the expression tree idea.

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: That sounds nice, but it doesn't adapt the examples very well. I was hoping the documentation could be entirely self-contained.

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: Fixed it with `order of precedence` and added an example to explain the importance of order of precedence.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,68 @@ namespace Mathlib.Tactic
...
+```lean
+/-
+a b: ℕ
+h: a = a + b
+⊢ a + a + (a + b) + a + a = 0
```
joneugster, line 73:
```suggestion
⊢ a + a + (a + b + b) + a + a = 0
```
Isn't it?

Mathlib/Tactic/NthRewrite.lean
joneugster: Fair enough. I also think the new iteration with the goal states is good.

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: No this is the proof state after the first rewrite.

Mathlib/Tactic/NthRewrite.lean
joneugster: I have to admit, I didn't even read the second docstring. If it is merely a `nth_rewrite --> nth_rw` renaming, I think this suggestion would make a lot of sense and wouldn't be confusing for the reader. And it would simplify maintainability (e.g. I think the last example of the second docstring is already out-of-sync again)

However, if there is more change to it might not be ideal.

Anyways, this is a completely _optional_ suggestion!

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,68 @@ namespace Mathlib.Tactic
...
+a b: ℕ
+h: a = a + b
+⊢ a + a + (a + b + b) + a + a = 0
+-/
+```
+This new instance of `a` also turns out to be the third _occurrence_ of `a`.  Therefore,
+the next `nth_rewrite` with `h` rewrites this `a`.
+
+Note: The occurrences are counted beginning with `1` and not `0`, this is different from
+mathlib3. The translation will be handled by mathport.
```
joneugster, line 80: I think the remaining question from Zulip is whether you want to remove this. I personally think it's perferctly fine to drop it.

Mathlib/Tactic/NthRewrite.lean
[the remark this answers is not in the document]
joneugster: I think I rather agree with Ruben here, but I guess that doesn't really matter either.

Mathlib/Tactic/NthRewrite.lean
j-loreaux: I think we should drop the last paragraph. No need to remind users of Lean 3 behavior a year past the completion of the port.

j-loreaux:
LGTM. Another maintainer can review the above unresolved conversations to see if they feel anything should be done. Make sure to wait for CI (probably doesn't matter since it's only docstrings, but still).

maintainer merge

Shreyas4991:
what is a `maintainer merge` and given that the bot says it is already on some merge queue, I am not sure how I wait for CI.

Mathlib/Tactic/NthRewrite.lean
semorrison: Even better, just have a single doc-string, copied via `inherit_doc` that explains both together. e.g.

`nth_rw` and `nth_rewrite` are analogues of `rw` and `rewrite` that allowing restrict which occurrence of a patttern to replace. `nth_rw` tries to close the goal after rewriting with a reducible transparency `rfl`, while `nth_rewrite` does not (just as for `rw` and `rewrite`).

.... now continue with examples just in terms of `nth_rw`.

Shreyas4991: The difference is the doc tooltip only shows you the docstring for the tactic your mouse pointer is hovering on. So if I am hovering on nth_rw, I would like to see a full explanation rather than being redirected. If inherit doc could accept parameters and format strings, then the docstrings would be correct, especially the examples. My rule of thumb for documentation examples is that they should be copy pastable with minimum modification.

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +86,54 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
   | _ => throwUnsupportedSyntax
 
 /--
-`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
+`nth_rw` is a variant of `nth_rewrite` that also tries to close the goal by trying `rfl` afterwards.
```
Vierkantor, line 89: I think users will be much more familiar with `rw` than `nth_rewrite`, so I would rephrase in terms of `rw`.
```suggestion
`nth_rw` is a variant of `rw` that only changes the `n`ᵗʰ _occurrence_ of the expression to be
rewritten. Like `rw`, unlike `nth_rewrite` it will try to close the goal by trying `rfl` afterwards.
```

Mathlib/Tactic/NthRewrite.lean
Vierkantor: I think now that we have the customized docstrings, the question is whether they will be diverging, so that one is updated and the other isn't. That's potentially more harm than copypastability breaking. My estimate is that we'll be okay with two similar docstrings, so I'm inclined to keep these as-is.

Vierkantor (approved):
@semorrison I'm okay either way, applying your suggestion to copy the docs or not. Do you want to make the final decision?

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: This is a very good idea. I'll add it in right away.

j-loreaux:
@Shreyas4991 a `maintainer merge` is when a [reviewer](https://leanprover-community.github.io/teams/reviewers.html) places a PR on a queue (really just a thread in a private channel on Zulip) for a maintainer to look at and potentially merge. Sometimes maintainers themselves use this queue so that they can get approval from another maintainer, as I have done here. You can still wait for CI as normal, but unlike delegation, a `maintainer merge` does not give anyone the authority to merge (other than maintainers themselves, but they already had such authority).

Mathlib/Tactic/NthRewrite.lean
Shreyas4991: Kim  has reacted with a (y) to @Vierkantor 's suggestion to keep these strings as is. Shall I assume that this can be marked resolved?

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -17,11 +17,56 @@ namespace Mathlib.Tactic
...
+-/
+```
+Notice that the second occurrence of `a` from the left has been rewritten by `nth_rewrite`.
+
+To understand the importance of order of precedence, consider the example below
+```lean
+example (a b c : Nat) : (a + b) + c = (b + a) + c := by
+  nth_rewrite 2 [Nat.add_comm] -- ⊢ (b + a) + c = (b + a) + c
+```
+Here, although the occurrence parameter is `2`, `(a + b)` is rewritten to `(b + a)`. This happens,
```
joneugster, line 42:
```suggestion
Here, although the occurrence parameter is `2`, `(a + b)` is rewritten to `(b + a)`. This happens
```

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +86,55 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
   | _ => throwUnsupportedSyntax
 
 /--
-`nth_rw` is like `nth_rewrite`, but also tries to close the goal by trying `rfl` afterwards.
+`nth_rw` is a variant of `rw` that only changes the `n`ᵗʰ _occurrence_ of the expression to be
+rewritten. Like `rw`, unlike `nth_rewrite` it will try to close the goal by trying `rfl` afterwards.
```
joneugster, line 90:
```suggestion
rewritten. Like `rw`, unlike `nth_rewrite`, it will try to close the goal by trying `rfl` afterwards.
```
maybe?

Mathlib/Tactic/NthRewrite.lean
```diff
@@ -41,7 +86,55 @@ syntax (name := nthRewriteSeq) "nth_rewrite" (config)? ppSpace num rwRuleSeq (lo
...
+-/
+```
+Notice that the second occurrence of `a` from the left has been rewritten by `nth_rewrite`.
+
+To understand the importance of order of precedence, consider the example below
+```lean
+example (a b c : Nat) : (a + b) + c = (b + a) + c := by
+  nth_rewrite 2 [Nat.add_comm] -- ⊢ (b + a) + c = (b + a) + c
+```
+Here, although the occurrence parameter is `2`, `(a + b)` is rewritten to `(b + a)`. This happens,
```
joneugster, line 109:
```suggestion
Here, although the occurrence parameter is `2`, `(a + b)` is rewritten to `(b + a)`. This happens
```

joneugster (approved):
I resolved the open suggestion, since I don't consider them necessary and other reviewers didn't either.

I read through the text again and it's really clear. Left some comma suggestions, but I think it should be merged with or without them

Ruben-VandeVelde (approved):
maintainer merge

j-loreaux:
bors merge

PR #13878, opened 2024-06-16
[Merged by Bors] - chore(NumberTheory/MulChar): deprecate `IsNontrivial`, fix `isPrimitive`

loefflerd:
- Deprecate `MulChar.IsNontrivial` (following the analogous change for `AddChar`'s a few days ago);
- fix capitalisation of `DirichletCharacter.isPrimitive`; 
- adjust some lemmas to take a `Nat` argument with a `NeZero` instance, rather than a `PNat`, which is more general and flexible.

---

### DIFF
--- a/Mathlib/NumberTheory/GaussSum.lean
+++ b/Mathlib/NumberTheory/GaussSum.lean
@@ -92,21 +92,20 @@ variable {R : Type u} [Field R] [Fintype R] {R' : Type v} [CommRing R'] [IsDomai
 
 -- A helper lemma for `gaussSum_mul_gaussSum_eq_card` below
 -- Is this useful enough in other contexts to be public?
-private theorem gaussSum_mul_aux {χ : MulChar R R'} (hχ : χ.IsNontrivial) (ψ : AddChar R R')
+private theorem gaussSum_mul_aux {χ : MulChar R R'} (hχ : χ ≠ 1) (ψ : AddChar R R')
     (b : R) :
     ∑ a, χ (a * b⁻¹) * ψ (a - b) = ∑ c, χ c * ψ (b * (c - 1)) := by
   rcases eq_or_ne b 0 with hb | hb
   · -- case `b = 0`
     simp only [hb, inv_zero, mul_zero, MulChar.map_zero, zero_mul,
-      Finset.sum_const_zero, map_zero_eq_one, mul_one]
-    exact (hχ.sum_eq_zero).symm
+      Finset.sum_const_zero, map_zero_eq_one, mul_one, χ.sum_eq_zero_of_ne_one hχ]
   · -- case `b ≠ 0`
     refine (Fintype.sum_bijective _ (mulLeft_bijective₀ b hb) _ _ fun x ↦ ?_).symm
     rw [mul_assoc, mul_comm x, ← mul_assoc, mul_inv_cancel hb, one_mul, mul_sub, mul_one]
 
 /-- We have `gaussSum χ ψ * gaussSum χ⁻¹ ψ⁻¹ = Fintype.card R`
 when `χ` is nontrivial and `ψ` is primitive (and `R` is a field). -/
-theorem gaussSum_mul_gaussSum_eq_card {χ : MulChar R R'} (hχ : χ.IsNontrivial) {ψ : AddChar R R'}
+theorem gaussSum_mul_gaussSum_eq_card {χ : MulChar R R'} (hχ : χ ≠ 1) {ψ : AddChar R R'}
     (hψ : IsPrimitive ψ) :
     gaussSum χ ψ * gaussSum χ⁻¹ ψ⁻¹ = Fintype.card R := by
   simp only [gaussSum, AddChar.inv_apply, Finset.sum_mul, Finset.mul_sum, MulChar.inv_apply']
@@ -124,7 +123,7 @@ theorem gaussSum_mul_gaussSum_eq_card {χ : MulChar R R'} (hχ : χ.IsNontrivial
 
 /-- When `χ` is a nontrivial quadratic character, then the square of `gaussSum χ ψ`
 is `χ(-1)` times the cardinality of `R`. -/
-theorem gaussSum_sq {χ : MulChar R R'} (hχ₁ : χ.IsNontrivial) (hχ₂ : IsQuadratic χ)
+theorem gaussSum_sq {χ : MulChar R R'} (hχ₁ : χ ≠ 1) (hχ₂ : IsQuadratic χ)
     {ψ : AddChar R R'} (hψ : IsPrimitive ψ) :
     gaussSum χ ψ ^ 2 = χ (-1) * Fintype.card R := by
   rw [pow_two, ← gaussSum_mul_gaussSum_eq_card hχ₁ hψ, hχ₂.inv, mul_rotate']
@@ -212,7 +211,7 @@ theorem Char.card_pow_char_pow {χ : MulChar R R'} (hχ : IsQuadratic χ) (ψ : 
 /-- When `F` and `F'` are finite fields and `χ : F → F'` is a nontrivial quadratic character,
 then `(χ(-1) * #F)^(#F'/2) = χ #F'`. -/
 theorem Char.card_pow_card {F : Type*} [Field F] [Fintype F] {F' : Type*} [Field F'] [Fintype F']
-    {χ : MulChar F F'} (hχ₁ : χ.IsNontrivial) (hχ₂ : IsQuadratic χ)
+    {χ : MulChar F F'} (hχ₁ : χ ≠ 1) (hχ₂ : IsQuadratic χ)
     (hch₁ : ringChar F' ≠ ringChar F) (hch₂ : ringChar F' ≠ 2) :
     (χ (-1) * Fintype.card F) ^ (Fintype.card F' / 2) = χ (Fintype.card F') := by
   obtain ⟨n, hp, hc⟩ := FiniteField.card F (ringChar F)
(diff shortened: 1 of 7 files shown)


YaelDillies:
Can you please make sure that PRs called `feat` contain new content? This is not the case of this one and, since it changes API in a trivial way (everything that could be done before can still be done after with a basically one to one correspondence), it should be `chore` (rather than `refactor`, which would involve changing definitions, refactoring the API...).

loefflerd:
Sure, sorry about that, fixed!

Mathlib/NumberTheory/GaussSum.lean
```diff
@@ -227,7 +226,7 @@ theorem Char.card_pow_card {F : Type*} [Field F] [Fintype F] {F' : Type*} [Field
   have := Fact.mk (hchar.subst hp')
   rw [Ne, ← Nat.prime_dvd_prime_iff_eq hp' hp, ← isUnit_iff_not_dvd_char, hchar] at hch₁
   exact Char.card_pow_char_pow (hχ₂.comp _) ψ.char (ringChar FF') n' hch₁ (hchar ▸ hch₂)
-       (gaussSum_sq (hχ₁.comp <| RingHom.injective _) (hχ₂.comp _) ψ.prim)
+       (gaussSum_sq ((ringHomComp_ne_one_iff (RingHom.injective _)).mpr hχ₁) (hχ₂.comp _) ψ.prim)
```
MichaelStollBayreuth, line 229: Dot notation was a benefit of `IsNontrivial` ...

Mathlib/NumberTheory/LegendreSymbol/AddCharacter.lean
```diff
@@ -193,19 +194,19 @@ theorem zmod_char_primitive_of_eq_one_only_at_zero (n : ℕ) (ψ : AddChar (ZMod
...
-  exact (ZMod.val_eq_zero a).mp (IsPrimitiveRoot.pow_inj h (ZMod.val_lt a) n.pos ha)
+  exact (ZMod.val_eq_zero a).mp (IsPrimitiveRoot.pow_inj h (ZMod.val_lt a) (NeZero.pos _) ha)
 #align add_char.zmod_char_primitive_of_primitive_root AddChar.zmodChar_primitive_of_primitive_root
 
 /-- There is a primitive additive character on `ZMod n` if the characteristic of the target
 does not divide `n` -/
 noncomputable def primitiveZModChar (n : ℕ+) (F' : Type v) [Field F'] (h : (n : F') ≠ 0) :
     PrimitiveAddChar (ZMod n) F' :=
-  haveI : NeZero ((n : ℕ) : F') := ⟨h⟩
+  haveI : NeZero (n : F') := ⟨h⟩
```
MichaelStollBayreuth, line 209:
```suggestion
  have : NeZero (n : F') := ⟨h⟩
```
The Lean4 "I" has nothing to do with instances, and without it it is actually marginally faster.

Mathlib/NumberTheory/LegendreSymbol/QuadraticChar/GaussSum.lean
```diff
@@ -51,14 +51,7 @@ theorem FiniteField.isSquare_two_iff :
...
       χ₈_nat_eq_if_mod_eight]
-    simp only [h, Nat.one_ne_zero, if_false, ite_eq_left_iff, Ne, (by decide : (-1 : ℤ) ≠ 1),
-      imp_false, Classical.not_not]
-  all_goals
-    rw [← Nat.mod_mod_of_dvd _ (by decide : 2 ∣ 8)] at h
-    have h₁ := Nat.mod_lt (Fintype.card F) (by decide : 0 < 8)
-    revert h₁ h
-    generalize Fintype.card F % 8 = n
-    intros; interval_cases n <;> simp_all -- Porting note (#11043): was `decide!`
+  all_goals omega
```
MichaelStollBayreuth, line 54: I would prefer
```lean4
  by_cases hF : ringChar F = 2
  · have h := FiniteField.even_card_of_char_two hF
    simp only [FiniteField.isSquare_of_char_two hF, true_iff_iff]
    omega
  · have h := FiniteField.odd_card_of_char_ne_two hF
    rw [← quadraticChar_one_iff_isSquare (Ring.two_ne_zero hF), quadraticChar_two hF,
      χ₈_nat_eq_if_mod_eight]
    omega
```
I think the juggling around with the goals is more confusing than having two `omega`s.

Mathlib/NumberTheory/LegendreSymbol/QuadraticChar/GaussSum.lean
```diff
@@ -81,14 +74,7 @@ theorem FiniteField.isSquare_neg_two_iff :
...
       quadraticChar_neg_two hF, χ₈'_nat_eq_if_mod_eight]
-    simp only [h, Nat.one_ne_zero, if_false, ite_eq_left_iff, Ne, (by decide : (-1 : ℤ) ≠ 1),
-      imp_false, Classical.not_not]
-  all_goals
-    rw [← Nat.mod_mod_of_dvd _ (by decide : 2 ∣ 8)] at h
-    have h₁ := Nat.mod_lt (Fintype.card F) (by decide : 0 < 8)
-    revert h₁ h
-    generalize Fintype.card F % 8 = n
-    intros; interval_cases n <;> simp_all -- Porting note (#11043): was `decide!`
+  all_goals omega
```
MichaelStollBayreuth, line 77: Same here.

Mathlib/NumberTheory/LegendreSymbol/QuadraticChar/GaussSum.lean
```diff
@@ -99,16 +85,15 @@ theorem quadraticChar_card_card [DecidableEq F] (hF : ringChar F ≠ 2) {F' : Ty
     quadraticChar F (Fintype.card F') =
     quadraticChar F' (quadraticChar F (-1) * Fintype.card F) := by
   let χ := (quadraticChar F).ringHomComp (algebraMap ℤ F')
-  have hχ₁ : χ.IsNontrivial := by
+  have hχ₁ : χ ≠ 1 := by
     obtain ⟨a, ha⟩ := quadraticChar_exists_neg_one hF
     have hu : IsUnit a := by
```
MichaelStollBayreuth, line 90: This is needed (at least) twice. Add a lemma
```lean
lemma quadraticChar_exists_neg_one' (hF : ringChar F ≠ 2) : ∃ a : Fˣ, quadraticChar F a = -1
```
in `NumberTheory.LegendreSymbol.QuadraticChar.Basic`?

Mathlib/NumberTheory/LegendreSymbol/QuadraticChar/Basic.lean
```diff
@@ -227,11 +227,17 @@ variable {F}
 
 /-- The quadratic character is nontrivial as a multiplicative character
 when the domain has odd characteristic. -/
-theorem quadraticChar_isNontrivial (hF : ringChar F ≠ 2) : (quadraticChar F).IsNontrivial := by
+theorem quadraticChar_ne_one (hF : ringChar F ≠ 2) : quadraticChar F ≠ 1 := by
   rcases quadraticChar_exists_neg_one hF with ⟨a, ha⟩
   have hu : IsUnit a := by by_contra hf; rw [MulChar.map_nonunit _ hf] at ha; omega
```
MichaelStollBayreuth, line 232: See below in `...QuadraticChar.GaussSum`.

Mathlib/NumberTheory/MulChar/Basic.lean
```diff
@@ -430,12 +430,21 @@ section nontrivial
 
 variable {R : Type*} [CommMonoid R] {R' : Type*} [CommMonoidWithZero R']
 
+lemma eq_one_iff {χ : MulChar R R'} : χ = 1 ↔ ∀ a : Rˣ, χ a = 1 := by
+  simp only [Ne, ext_iff, not_forall, one_apply_coe]
```
MichaelStollBayreuth, line 434:
```suggestion
  simp only [ext_iff, one_apply_coe]
```

Mathlib/NumberTheory/MulChar/Basic.lean
```diff
@@ -467,7 +476,23 @@ def ringHomComp (χ : MulChar R R') (f : R' →+* R'') : MulChar R R'' :=
...
+    using fun χ χ' h a ↦ hf (h a)
+
+lemma ringHomComp_eq_one_iff {f : R' →+* R''} (hf : Function.Injective f) {χ : MulChar R R'} :
+    χ.ringHomComp f = 1 ↔ χ = 1 := by
+  conv_lhs => rw [← (show  (1 : MulChar R R').ringHomComp f = 1 by ext; simp)]
+  exact (injective_ringHomComp hf).eq_iff
+
+lemma ringHomComp_ne_one_iff {f : R' →+* R''} (hf : Function.Injective f) {χ : MulChar R R'} :
+    χ.ringHomComp f ≠ 1 ↔ χ ≠ 1 := by
+  simp only [Ne, ringHomComp_eq_one_iff, hf]
```
MichaelStollBayreuth, line 491:
```suggestion
lemma ringHomComp_ne_one_iff {f : R' →+* R''} (hf : Function.Injective f) {χ : MulChar R R'} :
    χ.ringHomComp f ≠ 1 ↔ χ ≠ 1 :=
  (ringHomComp_eq_one_iff hf).not
```
Is it common to have both equivalences in Mathlib?

Mathlib/NumberTheory/LegendreSymbol/AddCharacter.lean
loefflerd: Yes, I don't really know why the porting scripts changed Lean3 `haveI` to lean4 `haveI` even though it means something different.

Mathlib/NumberTheory/MulChar/Basic.lean
loefflerd: I was following Yaël's treatment of `AddChar` which also has both. If we're using this in rewrites, it will be annoying to unfold the definition of `≠`.

MichaelStollBayreuth: I agree it's more convenient for rewriting (although one could use `rw [(ringHomComp_eq_one_iff hf).not]`).

Mathlib/NumberTheory/GaussSum.lean
loefflerd: Indeed, that's true; but there's a tradeoff against having to cook up an `IsNontrivial` type when you want to apply the lemmas. In my ongoing Dirichlet-L-series project I found `\chi \ne 1` to be an easier hypothesis to work with, and Yael seems to have found the same in the add-char context.

MichaelStollBayreuth (approved):
LGTM!

MichaelStollBayreuth:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #12661, opened 2024-05-04
chore(*): s/continuity/fun_prop/

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13880, opened 2024-06-16
[Merged by Bors] - chore: replace `continuity` by `fun_prop`, easy cases

grunweg:
Inspired by #12661: let's start with some easy instances, so we can focus on the difficult ones.

In particular, this PR does not depend on any in-progress bugfixes, not tries to change a default value `by continuity` to `by fun_prop`.

I am not at all confident about the tagging of lemmas with fun_prop: I remember this being subtle, but can't find any documentation on this. @lecopivo Can you advise on these lemmas (or simply link to the right documentation if I just overlooked it)?

---

### DIFF
--- a/Mathlib/Topology/CompactOpen.lean
+++ b/Mathlib/Topology/CompactOpen.lean
@@ -374,7 +374,7 @@ section Curry
     If `a × β` is locally compact, this is continuous. If `α` and `β` are both locally
     compact, then this is a homeomorphism, see `Homeomorph.curry`. -/
 def curry (f : C(X × Y, Z)) : C(X, C(Y, Z)) where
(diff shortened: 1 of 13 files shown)


grunweg:
CC @urkud

grunweg:
!bench

lecopivo:
Most of the theorems look fine to me. I remember there were some issues with `Sigma` so I'm not sure that all the theorems will be actually used. 

I will investigate the use of `Sigma` a bit closer but that should not block this PR.

lecopivo:
When checking you tagged the right theorem just add option
```
set_option trace.Meta.Tactic.fun_prop.attr true`
```
Every `fun_prop` attribute should produce a message. Check if the message say that the theorem is really about the function you think it should be.

PR #13881, opened 2024-06-16
[Merged by Bors] - chore: reduce use of `mono` in favour of `gcongr`

grunweg:
... as was discussed on zulip the other day; I cannot find the link quickly.
After this PR, a handful of uses remain which are harder to remove: [this branch](https://github.com/leanprover-community/mathlib4/compare/MR-mono-gcongr) comments them all

---------

Commits can be reviewed independently.

I'd like confirmation that the three lemmas I tagged with gcongr are actually fine to tag.

---


dupuisf:
Thanks! Yes I'm pretty sure those three `gcongr` lemmas are fine.

bors r+

Vierkantor:
Come on, bors!

bors merge

PR #13882, opened 2024-06-16
[Merged by Bors] - chore(ContinuousMap): add `TopologicalRing` instance

dupuisf:
This PR adds `TopologicalSemiring` and `TopologicalRing` instances on `C(α, β)`, and also gives explicit names to various related instances on `C(α, β)`.

---


dupuisf:
Sure -- I'm not a big fan of the autogenerated names even with the new algorithm but it's not a big deal. I left the one that's actually referred to by name elsewhere in mathlib though, I don't think we should rely on autogenerated names in that case.

PR #13887, opened 2024-06-17
[Merged by Bors] - feat(Geometry/Manifold): state generalized Poincaré conjecture

alreadydone:
---

### DIFF
--- a/Mathlib/Geometry/Manifold/PoincareConjecture.lean
+++ b/Mathlib/Geometry/Manifold/PoincareConjecture.lean
@@ -17,11 +17,12 @@ https://en.wikipedia.org/wiki/Generalized_Poincar%C3%A9_conjecture
 open scoped Manifold
 open Metric (sphere)
 
-variable (M : Type*) [TopologicalSpace M] [CompactSpace M] [T2Space M]
+variable (M : Type*) [TopologicalSpace M] [T2Space M]
 
 /-- The generalized topological Poincaré conjecture.
  - For n = 2 it follows from the classification of surfaces.
- - For n ≥ 5 it was proven by Stephen Smale in 1961.
+ - For n ≥ 5 it was proven by Stephen Smale in 1961 assuming M admits a smooth structure;
+   Newman (1966) and Connell (1967) proved it without the condition.
  - For n = 4 it was proven by Michael Freedman in 1982.
  - For n = 3 it was proven by Grigori Perelman in 2003. -/
 proof_wanted ContinuousMap.HomotopyEquiv.nonempty_homeomorph_sphere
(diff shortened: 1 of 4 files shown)


Mathlib/Geometry/Manifold/PoincareConjecture.lean
```diff
@@ -0,0 +1,64 @@
...
+
+/-- The 3-dimensional smooth Poincaré conjecture (proven by Perelman) -/
+proof_wanted SimplyConnectedSpace.nonempty_diffeomorph_sphere_fin4
+    [ChartedSpace (EuclideanSpace ℝ (Fin 3)) M] [SmoothManifoldWithCorners (𝓡 3) M]
+    [SimplyConnectedSpace M] [CompactSpace M] :
+    Nonempty (Diffeomorph (𝓡 3) (𝓡 3) M (sphere (0 : EuclideanSpace ℝ (Fin 4)) 1) ∞)
+
+/-- The smooth Poincaré conjecture; true for n = 1, 2, 3, 5, 6, 12, 56, and 61,
+open for n = 4, and it is conjectured that there are no other n > 4 for which it is true
+(Conjecture 1.17, https://annals.math.princeton.edu/2017/186-2/p03). -/
```
ocfnash, line 46: Huh, I had thought the prevailing 

Only a few weeks ago, I saw reports that Xu (and collaborators) had resolved the final 125-dimensional case ‼️https://www.math.princeton.edu/events/computing-differentials-adams-spectral-sequence-2024-05-30t170000 and https://bsky.app/profile/mikehill.bsky.social IIUC this resolves Conjecture 1.17 in the affirmative --- incidentally the opposite of what I was expecting / hoping 🤷‍♂️.

ocfnash (commented):
Ha, lovely!

I think it is probably a good idea to introduce:
```lean
class IsHomotopySphere (M : Type*) [TopologicalSpace M] (n : ℕ) : Prop :=
  nonempty_homotopyEquiv : Nonempty (ContinuousMap.HomotopyEquiv M S)
```
and
```lean
class IsTopologicalSphere (M : Type*) [TopologicalSpace M] (n : ℕ) : Prop :=
  nonempty_homoeomorph : Nonempty (Homeomorph M S)
```
I'll have a proper look this evening when I have some time (and try to remember if there was a reason I used structomorph rather than diffeomorph when I was over in ICERM).

Mathlib/Geometry/Manifold/PoincareConjecture.lean
alreadydone: Thanks! Extracting the definitions seem a good idea. It turns out Metric.sphere is already available in the Homotopy/Equiv file, but not in the Homeomorph file.

I think using `Structomorph` does make the statement shorter, because we're using a single model space rather than two (there's an [implementation note](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Geometry/Manifold/Diffeomorph.html#Implementation-notes) about this).

From my reading of p.12,15,31,33 of https://math.mit.edu/~mbehrens/presentations/Exotic_spheres.pdf, I think the resolution of the Kervaire invariant one problem in dimension 126 doesn't settle additional cases of smooth Poincare. Recall the smooth structures on $S^n$ are in bijection with a certain group $\Theta_n$. The existence of a 126-manifold with Kervaire invariant one means that $\Theta_{125}^{bp}$ is zero, so $\Theta_{125}$ is isomorphic to $\pi^S_{125}/J$, but we already know $\Theta_{125}\ne0$ (i.e. smooth Poincare is false for $n=125$, Proposition 1.12 in the 61-stem paper). So the odd case is already settled in the 61-stem paper.

For even $n$, $\Theta_n^{bp}$ vanishes and $\Theta_n$ is isomorphic to $\pi_^S_n/J$ or an index-2 subgroup of it, and the latter only happens if if $n\equiv 2\pmod{4}$ and Kervaire invariant one exists in dimension $n$. The new result implies it's index 2 in dimension 126, so it seems more likely to be trivial, but Isaksen and Xu already showed smooth Poincare fails ifor $n=126$ (Theorem 12.6, https://arxiv.org/abs/1708.06854v4). The same paper settles the even case in dimensions below 140, and remaining task is to show $\pi_^S_n/J$ is nontrivial for $n\ge 140$, i.e. it reduces to stable homotopy group computations.

alreadydone: I've switched to using notations and now I can just write `Nonempty (M ≃ₕ 𝕊ⁿ)` for homotopy spheres and `Nonempty (M ≃ₜ 𝕊ⁿ)` for topological spheres, and I'm not sure whether to extract these definitions at this time ...

ocfnash: Thanks for all the explanation, very much appreciated!

ocfnash:
What do you think about something like this:
```lean
def TopologicalPoincareConjecture (n : ℕ) : Prop :=
    ∀ (M : Type u) [TopologicalSpace M] [T2Space M] [ChartedSpace ℝⁿ M],
    M ≃ₕ 𝕊ⁿ → Nonempty (M ≃ₜ 𝕊ⁿ)

def SmoothPoincareConjecture (n : ℕ) : Prop :=
    ∀ (M : Type u) [TopologicalSpace M] [T2Space M] [ChartedSpace ℝⁿ M]
    [SmoothManifoldWithCorners (𝓡 n) M],
    M ≃ₕ 𝕊ⁿ → Nonempty (M ≃ₘ⟮𝓡 n, 𝓡 n⟯ 𝕊ⁿ)

proof_wanted TopologicalPoincareConjecture.proof : ∀ n, TopologicalPoincareConjecture n

proof_wanted SmoothPoincareConjecture.not_seven : ¬ SmoothPoincareConjecture 7
```
etc

semorrison:
bors merge

PR #13890, opened 2024-06-17
[Merged by Bors] - feat(Algebra/Lie/Normalizer): add `normalizer_mono` and use it to fix a porting note

grunweg:
`mono` would not apply on `monotone_normalizer`, but does apply on `normalizer_mono`.

---------

A subsequent PR will replace `mono` by `gcongr`- which needs the same change (and whose error message exposed this).

### DIFF
--- a/Mathlib/Algebra/Lie/Normalizer.lean
+++ b/Mathlib/Algebra/Lie/Normalizer.lean
@@ -72,11 +72,11 @@ theorem normalizer_inf : (N₁ ⊓ N₂).normalizer = N₁.normalizer ⊓ N₂.n
 #align lie_submodule.normalizer_inf LieSubmodule.normalizer_inf
 
 @[mono]
-theorem monotone_normalizer : Monotone (normalizer : LieSubmodule R L M → LieSubmodule R L M) := by
-  intro N₁ N₂ h m hm
+theorem monotone_normalizer (h : N₁ ≤ N₂) : normalizer N₁ ≤ normalizer N₂ := by
+  intro m hm
   rw [mem_normalizer] at hm ⊢
   exact fun x => h (hm x)
-#align lie_submodule.monotone_normalizer LieSubmodule.monotone_normalizer
+#align lie_submodule.monotone_normalizer LieSubmodule.monotone_normalizerₓ
 
 @[simp]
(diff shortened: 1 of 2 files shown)


Ruben-VandeVelde (commented):
I think we do also want the statement with `Monotone`, and I think the statement with ≤ is usually called "mono", not "monotone"

Mathlib/Algebra/Lie/Normalizer.lean
```diff
@@ -72,11 +72,11 @@ theorem normalizer_inf : (N₁ ⊓ N₂).normalizer = N₁.normalizer ⊓ N₂.n
 #align lie_submodule.normalizer_inf LieSubmodule.normalizer_inf
 
 @[mono]
-theorem monotone_normalizer : Monotone (normalizer : LieSubmodule R L M → LieSubmodule R L M) := by
-  intro N₁ N₂ h m hm
+theorem monotone_normalizer (h : N₁ ≤ N₂) : normalizer N₁ ≤ normalizer N₂ := by
```
YaelDillies, line 75: Instead of replace `monotone_normalizer`, can you add that as a new lemma called `normalizer_mono` (and tag that with `mono`)?

grunweg:
Thanks for the quick review! I have made the change you suggested.

Mathlib/Algebra/Lie/Normalizer.lean
```diff
@@ -72,10 +72,13 @@ theorem normalizer_inf : (N₁ ⊓ N₂).normalizer = N₁.normalizer ⊓ N₂.n
...
 @[mono]
-theorem monotone_normalizer : Monotone (normalizer : LieSubmodule R L M → LieSubmodule R L M) := by
-  intro N₁ N₂ h m hm
+theorem normalizer_mono (h : N₁ ≤ N₂) : normalizer N₁ ≤ normalizer N₂ := by
+  intro m hm
   rw [mem_normalizer] at hm ⊢
   exact fun x => h (hm x)
+
+theorem monotone_normalizer : Monotone (normalizer : LieSubmodule R L M → LieSubmodule R L M) :=
+  fun _ _ h ↦ normalizer_mono h
```
YaelDillies, line 81:
```suggestion
  fun _ _ ↦ normalizer_mono
```

YaelDillies (approved):
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #13891, opened 2024-06-17
chore: bump Aesop to enable precompilation

JLimperg:
---


JLimperg:
!bench

JLimperg:
!bench

Kha:
!bench

JLimperg:
So Aesop is much faster and the overall build much slower? Perhaps we should benchmark against `2c5ea133761ae723a6bce63dc236560fdcd6e17e` to minimise interference from other commits.

JLimperg:
[Here](http://speed.lean-fro.org/mathlib4/compare/8c43b26a-1ee7-4164-8aba-f42d7c2f8c75/to/562c87b2-d78e-46c8-b3e7-c3180478ead0) is the comparison between `lean-pr-testing-4478` and this PR. This should isolate the impact of turning on precompilation in Aesop. TLDR:

- Aesop -40%
- build wall-clock +7%
- interpretation +10%
- `open Mathlib` task-clock +1000%, wall-clock +225%, instructions +109%

So a loss overall. But where do these crazy numbers for `open Mathlib` come from? And why does interpretation get slower?

Kha:
> But where do these crazy numbers for open Mathlib come from?

This is likely because a no-op `lake build` suddenly takes 10x as much time on my machine. /cc @tydeu

tydeu:
@Kha I think this may be due to nature of precompile modules implying transitive import computation ([lake#132](https://github.com/leanprover/lake/pull/132), but I will investigate this further tomorrow.

JLimperg:
Closing; new attempt at #19055

PR #12930, opened 2024-05-15
[Merged by Bors] - refactor(Topology/Category): unify some definitions of subcategories of compact Hausdorff spaces

dagurtomas:
This PR unifies the definitions of the categories `CompHaus`, `Profinite`, `Stonean`, and `LightProfinite` by defining a structure `CompHausLike P` depending on a predicate `P` on topological spaces. Many of the properties of these categories of compact Hausdorff spaces are shared between them and this approach reduces duplication of boilerplate code, in particular the explicit constructions of certain limits and colimits.

---
The benefit of this refactor is demonstrated by its big negative diff, and the fact that it can be used to prove that the functors of discrete objects into light condensed sets and condensed sets are given by the sheaf of locally constant maps, at the same time (see #14027).

- [x] depends on: #13904 
- [x] depends on: #13905
- [x] depends on: #13907 
- [x] depends on: #13908 
- [x] depends on: #13909 
- [x] depends on: #13911 
- [x] depends on: #13912
- [x] depends on:  #15359
- [x] depends on: #15362 
- [x] depends on: #15363 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13893, opened 2024-06-17
feat(Condensed): the discrete functor to condensed and light condensed sets is fully faithful and given by the sheaf of locally constant maps

dagurtomas:
- [ ] depends on: #12930 

---


dagurtomas:
Continued at #14027

PR #11617, opened 2024-03-23
refactor(Finset): redefine Finset.diag, review API

urkud:
Redefine `Finset.diag` using `Finset.map`.
The new definition doesn't rely on `DecidableEq` and is much more computationally efficient.

Also review API and golf proofs here and there.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13899, opened 2024-06-17
[Merged by Bors] - feat: The diagonal of a finset

YaelDillies:
From LeanAPAP

---

### DIFF
--- a/Mathlib/Data/Finset/Pi.lean
+++ b/Mathlib/Data/Finset/Pi.lean
@@ -3,13 +3,14 @@ Copyright (c) 2018 Johannes Hölzl. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Johannes Hölzl
 -/
-import Mathlib.Data.Finset.Image
+import Mathlib.Data.Finset.Card
 import Mathlib.Data.Multiset.Pi
 
 /-!
 # The cartesian product of finsets
 -/
 
+open Function
 
 namespace Finset
 
@@ -123,6 +124,17 @@ theorem pi_disjoint_of_disjoint {δ : α → Type*} {s : Finset α} (t₁ t₂ :
     disjoint_iff_ne.1 h (f₁ a ha) (mem_pi.mp hf₁ a ha) (f₂ a ha) (mem_pi.mp hf₂ a ha) <|
       congr_fun (congr_fun eq₁₂ a) ha
 
-end Pi
+/-! ### Diagonal -/
 
+variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
+
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`. -/
+def piDiag (s : Finset α) (ι : Type*) [DecidableEq (ι → α)] : Finset (ι → α) := s.image (const ι)
+
+@[simp] lemma mem_piDiag : f ∈ s.piDiag ι ↔ ∃ a ∈ s, const ι a = f := mem_image
+
+@[simp] lemma card_piDiag (s : Finset α) (ι : Type*) [DecidableEq (ι → α)] [Nonempty ι] :
+    (s.piDiag ι).card = s.card := by rw [piDiag, card_image_of_injective _ const_injective]
+
+end Pi
 end Finset
--- a/Mathlib/Data/Finset/Pi.lean
+++ b/Mathlib/Data/Finset/Pi.lean
@@ -128,7 +128,8 @@ theorem pi_disjoint_of_disjoint {δ : α → Type*} {s : Finset α} (t₁ t₂ :
 
 variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
 
-/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`. -/
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`, namely the set of
+constant functions valued in `s`. -/
 def piDiag (s : Finset α) (ι : Type*) [DecidableEq (ι → α)] : Finset (ι → α) := s.image (const ι)
 
 @[simp] lemma mem_piDiag : f ∈ s.piDiag ι ↔ ∃ a ∈ s, const ι a = f := mem_image
--- a/Mathlib/Data/Fintype/Pi.lean
+++ b/Mathlib/Data/Fintype/Pi.lean
@@ -3,6 +3,7 @@ Copyright (c) 2017 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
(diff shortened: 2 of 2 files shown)


Mathlib/Data/Fintype/Pi.lean
```diff
@@ -163,3 +163,21 @@ theorem Finset.univ_pi_univ {α : Type*} {β : α → Type*} [DecidableEq α] [F
...
   ext; simp
 #align finset.univ_pi_univ Finset.univ_pi_univ
+
+/-! ### Diagonal -/
+
+namespace Finset
+variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
+
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`. -/
+def piDiag (s : Finset α) (ι : Type*) [DecidableEq (ι → α)] : Finset (ι → α) := s.image (const ι)
```
eric-wieser, line 173: Does `.map` work here?

YaelDillies: Subtly, no: If `\iota` is empty, then `const \iota` is not injective. Of course in that case the finset is obvious to construct, but I can't case on whether `\iota` is empty or not (unless I assume `Decidable (IsEmpty \iota)` which is.. unorthodox)

eric-wieser: Does the empty case make any sense in your application?

YaelDillies: It would be kind of strange as my application is `\iota := Fin n` and I would have to rule out `n = 0` explicitly. Not the end of the world, though.

Mathlib/Data/Fintype/Pi.lean
```diff
@@ -163,3 +163,21 @@ theorem Finset.univ_pi_univ {α : Type*} {β : α → Type*} [DecidableEq α] [F
...
+
+@[simp] lemma mem_piDiag : f ∈ s.piDiag ι ↔ ∃ a ∈ s, const ι a = f := mem_image
+
+lemma piDiag_subset_piFinset [DecidableEq ι] [Fintype ι] :
```
b-mehta, line 177: Let's also add
```
lemma piFinset_filter_const [DecidableEq ι] [Fintype ι] :
    (Fintype.piFinset fun _ => s).filter (∃ a ∈ s, const ι a = ·) = s.piDiag ι := by
  aesop
```
since it's easy to be more specific about this inclusion.

You also get a different proof of the inclusion, but that's up to you.
```
lemma piDiag_subset_piFinset [DecidableEq ι] [Fintype ι] :
    s.piDiag ι ⊆ Fintype.piFinset fun _ ↦ s := by
  simp [←piFinset_filter_const]
```

Mathlib/Data/Fintype/Pi.lean
```diff
@@ -163,3 +163,21 @@ theorem Finset.univ_pi_univ {α : Type*} {β : α → Type*} [DecidableEq α] [F
     (Finset.univ.pi fun a : α => (Finset.univ : Finset (β a))) = Finset.univ := by
   ext; simp
 #align finset.univ_pi_univ Finset.univ_pi_univ
+
+/-! ### Diagonal -/
+
+namespace Finset
+variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
+
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`. -/
```
b-mehta, line 172: I'm not sure about the placement of this *definition* - it doesn't use fintype anywhere. What do you think about putting the definition in Finset.pi instead? Of course the piFinset lemmas should remain here.

eric-wieser: Perhaps the choice of filename in #11617 is relevant here?

YaelDillies: `Data.Finset.Pi` already exists, so I have put the definition there for now

Mathlib/Data/Finset/Pi.lean
```diff
@@ -139,6 +140,17 @@ theorem pi_disjoint_of_disjoint {δ : α → Type*} {s : Finset α} (t₁ t₂ :
       congr_fun (congr_fun eq₁₂ a) ha
 #align finset.pi_disjoint_of_disjoint Finset.pi_disjoint_of_disjoint
 
-end Pi
+/-! ### Diagonal -/
+
+variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
+
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`. -/
```
Ruben-VandeVelde, line 147: Maybe this is just my ignorance, but I'm not sure what a diagonal is after reading this comment

YaelDillies: Better now?

semorrison: Perhaps foreshadow what this is for?

YaelDillies: I'm not sure my use case is a good fit for the docstring. It is used inside a proof of Almost Periodicity, where we are interested in tuples `x` and `y` differing by a constant, and we are interested in tuples because we can use a concentration inequality to approximate some random variable `X` by an average of uniform variables `x 0`, ..., `x (n - 1)` valued in some set.

You can see it for yourself here: https://github.com/search?q=repo%3AYaelDillies%2FLeanAPAP%20wideDiag&type=code (it's currently called `wideDiag` in APAP)

Mathlib/Data/Finset/Pi.lean
```diff
@@ -139,6 +140,18 @@ theorem pi_disjoint_of_disjoint {δ : α → Type*} {s : Finset α} (t₁ t₂ :
...
 #align finset.pi_disjoint_of_disjoint Finset.pi_disjoint_of_disjoint
 
-end Pi
+/-! ### Diagonal -/
+
+variable {ι : Type*} [DecidableEq (ι → α)] {s : Finset α} {f : ι → α}
+
+/-- The diagonal of a finset `s : Finset α` as a finset of functions `ι → α`, namely the set of
+constant functions valued in `s`. -/
+def piDiag (s : Finset α) (ι : Type*) [DecidableEq (ι → α)] : Finset (ι → α) := s.image (const ι)
```
semorrison, line 149: Would `constFnsWithin` be better? I think a niche notion like this can afford to have a long name, and should have a name that is clear.

YaelDillies: I see your point, but it is completely analogous to `Finset.diag`, except for `ι → α` rather than `α × α`. I think this justifies the current name (and points to the fact that we might want the `ι →₀ α` version in the future

Ruben-VandeVelde (approved):
Fine for me.

maintainer merge

riccardobrasca:
!bench

riccardobrasca:
Thanks!

bors merge

PR #13902, opened 2024-06-17
[Merged by Bors] - feat(RingTheory/TwosidedIdeal/Basic): define two-sided-ideal

jjaassoonn:
---


j-loreaux:
How do we feel about having `abbrev TwoSidedIdeal := RingCon`? I don't love the spelling `RingCon`.

jjaassoonn:
> How do we feel about having `abbrev TwoSidedIdeal := RingCon`? I don't love the spelling `RingCon`.

I have added the abbreviation `TwoSidedIdeal`.

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -0,0 +1,125 @@
...
+
+open MulOpposite
+
+section definitions
+
+/--
+A two-sided ideal of a ring `R` is a subset of `R` that contains `0` and is closed under addition,
+negation, and absorbs multiplication on both sides.
+-/
+abbrev TwoSidedIdeal (R : Type*) [NonUnitalNonAssocRing R] := RingCon R
```
jjaassoonn, line 33: Should it be `abbrev` or `def`, do we want Lean to unfold `TwoSidedIdeal`?

j-loreaux: Definitely `abbrev` if we're going to have it at all, but I'll wait for other reviewers to comment about whether we actually want the alternate spelling.

eric-wieser: I think having this alternate spelling is probably confusing; either we should have one name that functions both as a relation and a set, or we define
```lean
structure TwoSidedIdeal where
  toRingCon : RingCon R
```
to draw a clear distinction between the two APIs.

jjaassoonn: I am okay both ways, not to use a structure with a single field could save some boilerplate probably. But I don’t really mind boilerplate code.

jjaassoonn: Okay. I have switched to one-field-structure approach.

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -0,0 +1,172 @@
...
+- a proof that `x * y ∈ S` if `y ∈ S`;
+- a proof that `x * y ∈ S` if `x ∈ S`.
+-/
+def mk' (carrier : Set R)
+    (zero_mem : 0 ∈ carrier)
+    (add_mem : ∀ {x y}, x ∈ carrier → y ∈ carrier → x + y ∈ carrier)
+    (neg_mem : ∀ {x}, x ∈ carrier → -x ∈ carrier)
+    (mul_mem_left : ∀ {x y}, y ∈ carrier → x * y ∈ carrier)
+    (mul_mem_right : ∀ {x y}, x ∈ carrier → x * y ∈ carrier) : TwoSidedIdeal R where
+  ringCon :=
```
jjaassoonn, line 118: Since we are using a structure, might as well provide an alternative constructor

kbuzzard (approved):
This now looks fine to me.

j-loreaux:
@eric-wieser and @kbuzzard : sorry I didn't get to this sooner, I've been a bit busy. My issue with `structure` has to do with quotients. One of the primary reasons we need two-sided ideals in this generality is to properly talk about quotients of (non-commutative and non-unital) rings by ideals.

With the `abbrev` approach, my thinking is that this would come mostly for free, but I'm concerned that using a `structure` will significantly complicate this.

Actually, this brings me to my second concern, which is not necessarily directed at this PR: how do we plan to deal with quotients anyway? Currently, we make rather involved use of the fact that we have very convenient defeqs between quotients of `AddGroup`s and quotients of `CommRing`s. We're going to have to lose this at some point (I think). Do you have any good ideas for dealing with this?

eric-wieser:
> We're going to have to lose this at some point (I think).

I'm not convinced; I somewhat expect that we will be able to keep it. Quotients by ideals are already defined in terms of `RingCon`, without any issue.

j-loreaux:
In that case I'll just remove my objection / concern. Thanks!

eric-wieser:
> With the `abbrev` approach, my thinking is that this would come mostly for free, but I'm concerned that using a `structure` will significantly complicate this.

What this *does* apply to is the lattice structure, but I think this is just a pile of boilerplate, and not a significant complication.

jjaassoonn:
Can I gently nudge this PR in the direction of merging? perhaps @eric-wieser

kbuzzard:
maintainer merge

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -0,0 +1,174 @@
...
+import Mathlib.RingTheory.Congruence.Basic
+
+/-!
+# Two Sided Ideals
+
+In this file, for any `Ring R`, we reinterpret `I : RingCon R` as a two-sided-ideal of a ring.
+
+## Main definitions and results
+
+* `TwoSidedIdeal`: For any `NonUnitalNonAssocRing R`, `TwoSidedIdeal R` is the exactly `RingCon R`.
```
jcommelin, line 18:
```suggestion
* `TwoSidedIdeal`: For any `NonUnitalNonAssocRing R`, `TwoSidedIdeal R` is a wrapper around `RingCon R`.
```

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -0,0 +1,174 @@
...
+* `TwoSidedIdeal.setLike`: Every `I : TwoSidedIdeal R` can be interpreted as a set of `R` where
+  `x ∈ I` if and only if `I x 0`.
```
jcommelin, line 20:
```suggestion
  `x ∈ I` if and only if `I.ringCon x 0`.
```

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -0,0 +1,174 @@
...
+    (mul_mem_right : ∀ {x y}, x ∈ carrier → x * y ∈ carrier)
+    (x : R) :
+    x ∈ mk' carrier zero_mem add_mem neg_mem mul_mem_left mul_mem_right ↔ x ∈ carrier := by
+  rw [mem_iff]
+  simp [mk']
+
+instance : AddSubgroupClass (TwoSidedIdeal R) R where
+  zero_mem := zero_mem
+  add_mem := @add_mem _ _
+  neg_mem := @neg_mem _ _
```
jcommelin, line 148: I think you can move this up, just after the `neg_mem` lemma.
Then you can use it in theproofs of `sub_mem` and `nsmul_mem` etc...
Of course the current proofs are also one-liners, so this isn't very important.

jcommelin (commented):
bors d+

PR #13283, opened 2024-05-27
[Merged by Bors] - feat(NumberTheory/PrimeCounting): Lemma: The `n`th prime is greater or equal `n + 2`.

rwst:
Possibly useful lemma about the nth prime number.

Co-authored-by: Pol'tta / Miyahara Kō <52843868+Komyyy@users.noreply.github.com>
Co-authored-by: Thomas Browning <tb65536@uw.edu>

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13906, opened 2024-06-17
Revert "feat(NumberTheory/PrimeCounting): Lemma: The `n`th prime is greater or equal `n + 2`."

erdOne:
Reverts leanprover-community/mathlib4#13283

### DIFF
--- a/Mathlib/NumberTheory/PrimeCounting.lean
+++ b/Mathlib/NumberTheory/PrimeCounting.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2021 Bolton Bailey. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Bolton Bailey, Ralf Stephan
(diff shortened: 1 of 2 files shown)


kmill:
@erdOne What's your intention with reverting Kevin's mistaken merge? Do you not want it merged? Or are you trying to make sure it gets merged correctly through Bors?

erdOne:
I thought force-pushing would also cause troubles? If the mistaken merge won't cause troubles later on then we can keep it.

kmill:
Maintainers are talking about what to do here. Just to check, you're just helping with the revert process in case we need to revert it, but you're ok with https://github.com/leanprover-community/mathlib4/pull/13283 having been merged, correct?

erdOne:
Yes.

kmill:
Thanks for your quick action on this by the way. (I was writing a message to you and got distracted in the meantime.) We did end up deciding to risk Kevin's push to master given that it already went through review.

PR #13918, opened 2024-06-18
feat(Topology/Order/HullKernel): Add the Hull-Kernel Topology

mans0954:
Let `α` be a `CompleteLattice` and let `T` be a subset of `α`. The relative `Topology.lower` on `T`
takes on a particularly simple form when every element of `T` is `InfPrime` in `α`. In this case,
the relative-open sets are exactly the sets of the form `T ↓∩ (Ici a)ᶜ` for some `a` in `α`.

The pair of maps `S → ⊓ S` and `a → T ↓∩ Ici a` are often referred to as the kernel and the hull.
They form an antitone Galois connection between the subsets of `T` and `α`. When `α` can be generated
from `T` by taking infs, this becomes a Galois insertion and the topological closure coincides with
the closure arising from the Galois insertion. For this reason the relative lower topology on `T` is
often referred to as the hull-kernel topology. The names Jacobson topology and structure topology
also occur in the literature.

---


mans0954:
@YaelDillies do you have time to review this PR please? Thank you.

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+namespace PrimitiveSpectrum
+
+-- b ⊓ c ≤ a → b ≤ a ∨ c ≤ a
+
+/- The set of relative-closed sets of the form `T ↓∩ (Ici a)` for some `a` in `α` is closed under
+pairwise union. -/
+lemma ici_union_ici_eq (a b : α) :
+    (T ↓∩ (Ici a)) ∪ (T ↓∩ (Ici b)) = (T ↓∩ (Ici (a ⊓ b))) := by
+  ext p
+  constructor
```
YaelDillies, line 79:
```suggestion
  constructor <;> intro h
```

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+  · intro h
+    cases' h with h1 h3
+    · exact inf_le_of_left_le h1
+    · exact inf_le_of_right_le h3
+  · intro h
+    exact ((hT p (Subtype.coe_prop p)).2) h
```
YaelDillies, line 85:
```suggestion
    exact (hT p p.2).2 h
```

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+  exact ⟨fun a ha => by
+      simp only [mem_sInter, mem_setOf_eq, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂,
+      mem_preimage, mem_Ici] at ha
+      simp only [mem_preimage, mem_Ici, sSup_le_iff]
+      exact fun b a ↦ ha b a,
+    fun a ha =>
+      Set.preimage_val_subset_preimage_val (antitone_Ici (CompleteLattice.le_sSup S a ha))⟩
+
+/- When `α` is complete, the relative basis for the Lower topology is also closed under arbitary
+unions.-/
```
YaelDillies, line 170:
```suggestion
unions. -/
```

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+
+lemma gc_closureOperator_eq (S : Set T) : gc.closureOperator S = T ↓∩ (Ici (sInf S)) := by
+  simp only [toDual_sInf, GaloisConnection.closureOperator_apply, ofDual_sSup]
+  rw [← preimage_comp, ← OrderDual.toDual_symm_eq, Equiv.symm_comp_self, preimage_id_eq, id_eq]
+
+variable (T)
+
+/-- `T` is said to order generate `α` if, for every `a` in `α`, there exists a subset of `T` such
+that `a` is the inf of `S`. -/
+def OrderGenerate := ∀ (a : α), ∃ (S : Set T), a = sInf (S : Set α)
```
YaelDillies, line 236:
```suggestion
def OrderGenerate := ∀ (a : α), ∃ S ⊆ T, a = sInf S
```
and everywhere else.

YaelDillies, line 236: This is a very general definition. I suspect what we actually want is defining `CompleteSublattice.closure` and replacing `OrderGenerate` by `CompleteSublattice.closure T = ⊤`. What do you think?

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+open TopologicalSpace
+open Topology
+open Set
+open Set.Notation
+
+section SemilatticeInf
+
+variable [SemilatticeInf α] [TopologicalSpace α] [IsLower α]
+
+variable {T : Set α} (hT : ∀ p ∈ T, InfPrime p)
```
YaelDillies, line 68: This is suspicious... I suspect I need to read the references to understand, but I won't have time before another week.

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+
+namespace PrimitiveSpectrum
+
+-- b ⊓ c ≤ a → b ≤ a ∨ c ≤ a
```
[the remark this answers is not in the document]
mans0954: Yes, thanks.

Mathlib/Topology/Order/HullKernel.lean
```diff
@@ -0,0 +1,289 @@
...
+import Mathlib.Topology.Sets.Closeds
+
+/-!
+# Hull-Kernel Topology
+
+Let `α` be a `CompleteLattice` and let `T` be a subset of `α`. The relative `Topology.lower` on `T`
+takes on a particularly simple form when every element of `T` is `InfPrime` in `α`. In this case,
+the relative-open sets are exactly the sets of the form `T ↓∩ (Ici a)ᶜ` for some `a` in `α`.
+
+The pair of maps `S → ⊓ S` and `a → T ↓∩ (Ici a)` are often referred to as the kernel and the hull.
```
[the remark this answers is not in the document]
mans0954: A hangover from when I was working with open sets rather than closed.

Mathlib/Topology/Order/HullKernel.lean
mans0954: `S ⊆ T` gives `S : Set α` rather than `S : Set T`.

mans0954: Sounds plausible.

Mathlib/Topology/Order/HullKernel.lean
mans0954: I probably got the idea for this approach from Henriksen et al https://link.springer.com/article/10.1007/s000120050056. The hull kernel topology on the Primitive Spectrum of a C*-algebra is discussed in most introductory texts on C*-algebras. Various related concepts are used in more specialist applications.

However, one can get a long way with both the basic results and the specialist applications without reference to C*-algebras. Much of the theory depends just on the Primitive Spectrum being a dense order-generating subset of the prime elements of the lattice of norm-closed two-sided ideals of the C*-algebra. The advantage of taking this point of view is that the results then apply in more general situations like the lattice norm-closed ideals of a JB*-algebra or even the lattice of M-ideals of a Banach space. Henriksen et al also make a connection to minimal primes in algebraic lattices.

AFAIK we don't have two sided ideals in Mathlib yet, so I couldn't develop this theory specifically for C*-algebras even if I wanted to.

I don't know if there's an accepted term for a dense, order-generating subset of the prime elements of a lattice. I ended up calling such sets "primitive" in my doctoral thesis, but I was never very satisfied with that as a name.

Mathlib/Topology/Order/HullKernel.lean
YaelDillies: Yes, that's the point

Mathlib/Topology/Order/HullKernel.lean
YaelDillies: > I probably got the idea for this approach from Henriksen et al https://link.springer.com/article/10.1007/s000120050056. The hull kernel topology on the Primitive Spectrum of a C*-algebra is discussed in most introductory texts on C*-algebras. Various related concepts are used in more specialist applications.

> However, one can get a long way with both the basic results and the specialist applications without reference to C*-algebras. Much of the theory depends just on the Primitive Spectrum being a dense order-generating subset of the prime elements of the lattice of norm-closed two-sided ideals of the C*-algebra.

Maybe @j-loreaux will have an opinion on how your general machinery should interface with C*-algebras. What `T` will you take then? Shouldn't `ht : ∀ p ∈ T, InfPrime p` be automatically provided by typeclass inference then (by making that predicate a typeclass)?

> AFAIK we don't have two sided ideals in Mathlib yet, so I couldn't develop this theory specifically for C*-algebras even if I wanted to.

#13401 and #13902 are relevant.

Mathlib/Topology/Order/HullKernel.lean
mans0954: Presumably one would have `CompleteSublattice.sSupClosure` and  `CompleteSublattice.sInfClosure` by analogy with `Order/SupClosed` and `OrderGenerate` would then be  `CompleteSublattice.sInfClosure ⊤ = ⊤`?

Mathlib/Topology/Order/HullKernel.lean
mans0954: Let `A` be a Banach space and let `δₑA^*₁` be the set of extreme points of the dual unit ball. For each `x` in `δₑA^*₁` let `kₙ(ker x)` be the largest M-ideal of `A` contained in the kernel `ker x` of `x`. Then α would be the lattice of M-ideals in `A` and `T` would be `Prim A`, the set of M-ideals `{ kₙ(ker x) : x ∈ δₑA^*₁ }`.

Mathlib/Topology/Order/HullKernel.lean
YaelDillies: Actually, we need both `CompleteSublatticeSup` and `CompleteSublatticeInf` (being closed under suprema doesn't imply being closed under infima), then we can define `CompleteSublatticeSup.closure` and `CompleteSublatticeInf.closure`.

Mathlib/Topology/Order/HullKernel.lean
YaelDillies: So `ht` here is a fact that you establiush once and for all, right? I think it should therefore be a typeclass

YaelDillies:
Is this still supposed to be `awaiting-author`?

mans0954:
> Is this still supposed to be `awaiting-author`?

I figured that perhaps I should get M-ideal into Mathlib before addressing your point about `variable {T : Set α} (hT : ∀ p ∈ T, InfPrime p)` so I started working on this PR #14369. However, that needs the Krein-Šmulian theorem which I haven't had time to look at yet.

YaelDillies:
Okay, that's fair enough! I was just wondering whether you somehow forgot to remove the label.

mans0954:
This PR has been migrated to a fork-based workflow: https://github.com/leanprover-community/mathlib4/pull/26336

PR #13922, opened 2024-06-18
feat(CategoryTheory): discrete objects in sheaf categories

dagurtomas:
This PR develops the API for discrete objects in sheaf categories, and in particular shows that the notion of being discrete is invariant under equivalence of categories.

---

- [x] depends on: #13923 
- [ ] depends on: #13925 


Mathlib/Topology/Category/TopCat/Basic.lean
```diff
@@ -138,6 +138,16 @@ def discrete : Type u ⥤ TopCat.{u} where
...
+/-- The functor `TopCat.discrete` is fully faithful. -/
+def fullyFaithfulDiscrete : discrete.FullyFaithful where
+  preimage f := (f : _ → _)
+  map_preimage _ := rfl
+  preimage_map _ := rfl
+
+instance : discrete.Faithful := fullyFaithfulDiscrete.faithful
+
+instance : discrete.Full := fullyFaithfulDiscrete.full
+
```
dagurtomas, line 150:
```suggestion
```

Mathlib.lean
```diff
@@ -4103,6 +4105,7 @@ import Mathlib.Topology.Category.Stonean.EffectiveEpi
 import Mathlib.Topology.Category.Stonean.Limits
 import Mathlib.Topology.Category.TopCat.Adjunctions
 import Mathlib.Topology.Category.TopCat.Basic
+import Mathlib.Topology.Category.TopCat.DiscreteObjects
```
dagurtomas, line 4108:
```suggestion
```

Mathlib/CategoryTheory/Sites/DiscreteObjects.lean
```diff
@@ -0,0 +1,114 @@
+/-
+Copyright (c) 2024 Dagur Asgeirsson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dagur Asgeirsson
+-/
+import Mathlib.CategoryTheory.Adjunction.DiscreteObjects
+import Mathlib.CategoryTheory.Sites.ConstantSheaf
+import Mathlib.CategoryTheory.Sites.InducedTopology
```
dagurtomas, line 8:
```suggestion
import Mathlib.CategoryTheory.Sites.DenseSubsite
```

Mathlib/CategoryTheory/Sites/DiscreteObjects.lean
```diff
@@ -0,0 +1,114 @@
...
+Copyright (c) 2024 Dagur Asgeirsson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dagur Asgeirsson
+-/
+import Mathlib.CategoryTheory.Adjunction.DiscreteObjects
+import Mathlib.CategoryTheory.Sites.ConstantSheaf
+import Mathlib.CategoryTheory.Sites.DenseSubsite
+/-!
+
+# Discrete objects in sheaf categories.
```
dagurtomas, line 11:
```suggestion
# Discrete objects in sheaf categories.

TODO: prove that `sheafCompose` with a functor preserving sheafification preserves discreteness.
```

Mathlib/CategoryTheory/Adjunction/DiscreteObjects.lean
```diff
@@ -0,0 +1,104 @@
...
+-/
+
+open CategoryTheory
+
+variable {C D : Type*} [Category C] [Category D] (U : C ⥤ D)
+
+namespace CategoryTheory.Functor
+
+/-- A functor is said to have discrete objects if it has a fully faithful left adjoint. -/
+class HasDiscreteObjects : Prop where
```
joelriou, line 27: I believe that we could have the adjunction part of the class similarly as it was done in `Adjunction.Reflective`. Also, if the typeclass was attached to the left adjoint rather that to the right adjoint, it seems this would be the notion of "coreflective" functor, which is dual to reflective functors. Then, I think it would make sense to dualize (part of?) the results from `Adjunction.Reflective` into a new file `Adjunction.Coreflective` instead. This may also remove the need for #13923.

dagurtomas: I dualized some of the API from `Adjunction.Reflective` in #13948. Then I ended up not needing it for discrete sheaves, which I've defined at #13947 without using this ` DiscreteObjects` API. 

One can easily prove `Coreflective (constantSheaf J A)` if it is fully faithful and the site has a terminal object. However, I want to talk about discreteness with respect to a specific terminal object (this is because in `CompHaus` and friends, it is useful to use the terminal object `CompHaus.of PUnit` instead of the arbitrary one given by `HasTerminal`) and I do this successfully in #13497 (needs some cleanup before it's ready for review).

So I think I'll just close all these `DiscreteObjects` PRs, one should use the `Coreflective` API instead if this is desired.

PR #13923, opened 2024-06-18
feat(CategoryTheory/Adjunction): leftAdjointCongr

dagurtomas:
This PR adds an isomrphism `leftAdjointCongr`; given two isomorphic functors, one of which is a right adjoint (and hence also the other), their left adjoints are isomorphic. 

---
This is used in #13922 in the context of discrete objects with respect to a functor


joelriou:
I am not sure we need this API, see my comment on #13922

dagurtomas:
Yeah, thanks for your feedback. I'm currently working on figuring out what API I need and then I will try to implement it using your suggestion on #13922.

PR #13926, opened 2024-06-18
[Merged by Bors] - feat: small types of shifted hom in the localized category

joelriou:
If `C` is a category equipped with a shift by an additive monoid `M`, and `W : MorphismProperty C` is compatible with the shift, we define a type-class `HasSmallLocalizedShiftedHom.{w} W X Y` which says that all the types of morphisms from `X⟦a⟧` to `Y⟦b⟧` in the localized category are `w`-small for a certain universe. Then, we define types `SmallShiftedHom.{w} W X Y m : Type w` for all `m : M`, and endow these with a composition which transports the composition on the types `ShiftedHom (L.obj X) (L.obj Y) m` when `L : C ⥤ D` is any localization functor for `W`.

This shall be used in the redefinition of `Ext`-groups in abelian categories.

---

- [x] depends on: #13757
- [x] depends on: #13956
- [x] depends on: #13639


Mathlib/CategoryTheory/Iso.lean
```diff
@@ -250,6 +250,32 @@ theorem hom_eq_inv (α : X ≅ Y) (β : Y ≅ X) : α.hom = β.inv ↔ β.hom =
...
+def homFromEquiv (α : X ≅ Y) {Z : C} : (X ⟶ Z) ≃ (Y ⟶ Z) where
+  toFun f := α.inv ≫ f
+  invFun g := α.hom ≫ g
+  left_inv := by aesop_cat
+  right_inv := by aesop_cat
+
+/-- The bijection `(X ⟶ Y) ≃ (X' ⟶ Y')` that is induced by isomorphisms
+`e : X ≅ X'` and `e' : Y ≅ Y'`. -/
+@[simps]
+def homEquiv {X' Y' : C} (e : X ≅ X') (e' : Y ≅ Y') :
```
TwoFX, line 272: As noted on Zulip, this seems to be a duplicate of [Iso.homCongr](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Conj.html#CategoryTheory.Iso.homCongr).

joelriou: Thanks! This is fixed in #13956.

erdOne:
Thanks!
maintainer merge

Mathlib/CategoryTheory/Localization/SmallShiftedHom.lean
```diff
@@ -0,0 +1,187 @@
...
+
+end
+
+section
+
+variable (L : C ⥤ D) [L.IsLocalization W] [HasShift D M] [L.CommShift M]
+  {X Y Z T : C}
+
+/-- The bijection `SmallShiftedHom.{w} W X Y m ≃ ShiftedHom (L.obj X) (L.obj Y) m`
+for all `m : M`, and `X` and `Y` in `C`m when `L : C ⥤ D` is a localization functor for
```
erdOne, line 130:
```suggestion
for all `m : M`, and `X` and `Y` in `C` when `L : C ⥤ D` is a localization functor for
```
Oops missed this.

riccardobrasca:
Thanks!

bors merge

PR #13929, opened 2024-06-18
[Merged by Bors] - doc(Analysis/Convex/Slope) : fix a typo, add docstrings

LorenzoLuccioli:
- Fix a typo in the docstrings (`[x, z]` instead of `[y, z]`). Happens multiple times in the file.
- Add docstrings for `ConvexOn.secant_mono`, `StrictConvexOn.secant_strict_mono` and `StrictConcaveOn.secant_strict_mono`.
- Remove a `show` tactic. It was added during the porting, but it seems that it is no longer needed. Happens twice.

---

### DIFF
--- a/Mathlib/Analysis/Convex/Slope.lean
+++ b/Mathlib/Analysis/Convex/Slope.lean
@@ -267,6 +267,8 @@ theorem ConvexOn.secant_mono_aux3 (hf : ConvexOn 𝕜 s f) {x y z : 𝕜} (hx : 
   linarith only [hf.secant_mono_aux1 hx hz hxy hyz]
 #align convex_on.secant_mono_aux3 ConvexOn.secant_mono_aux3
 
+/-- If `f : 𝕜 → 𝕜` is convex, then for any point `a` the slope of the secant line of `f` through `a`
+and `b ≠ a` is monotone with respect to `b`. -/
 theorem ConvexOn.secant_mono (hf : ConvexOn 𝕜 s f) {a x y : 𝕜} (ha : a ∈ s) (hx : x ∈ s)
     (hy : y ∈ s) (hxa : x ≠ a) (hya : y ≠ a) (hxy : x ≤ y) :
     (f x - f a) / (x - a) ≤ (f y - f a) / (y - a) := by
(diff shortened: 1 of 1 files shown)


Mathlib/Analysis/Convex/Slope.lean
```diff
@@ -267,6 +265,8 @@ theorem ConvexOn.secant_mono_aux3 (hf : ConvexOn 𝕜 s f) {x y z : 𝕜} (hx :
   linarith only [hf.secant_mono_aux1 hx hz hxy hyz]
 #align convex_on.secant_mono_aux3 ConvexOn.secant_mono_aux3
 
+/-- If `f : 𝕜 → 𝕜` is convex, then for any point `a` the slope of the secant line of `f` through `a`
```
grunweg, line 268: This is not fully accurate: `f` is convex on `s`, and `a` and `x` are points in `s`. Can you rephrase this doc-string accordingly, please?

Mathlib/Analysis/Convex/Slope.lean
```diff
@@ -317,6 +315,8 @@ theorem StrictConvexOn.secant_strict_mono_aux3 (hf : StrictConvexOn 𝕜 s f) {x
   linarith only [hf.secant_strict_mono_aux1 hx hz hxy hyz]
 #align strict_convex_on.secant_strict_mono_aux3 StrictConvexOn.secant_strict_mono_aux3
 
+/-- If `f : 𝕜 → 𝕜` is strictly convex, then for any point `a` the slope of the secant line of `f`
```
grunweg, line 318: Similarly here, `f` is strictly convex *on s*, not globally.

Mathlib/Analysis/Convex/Slope.lean
```diff
@@ -329,6 +329,8 @@ theorem StrictConvexOn.secant_strict_mono (hf : StrictConvexOn 𝕜 s f) {a x y
   · exact hf.secant_strict_mono_aux2 ha hy hxa hxy
 #align strict_convex_on.secant_strict_mono StrictConvexOn.secant_strict_mono
 
+/-- If `f : 𝕜 → 𝕜` is strictly concave, then for any point `a` the slope of the secant line of `f`
```
grunweg, line 332: Similarly here, `f` is strictly convace *on s*, not globally.

grunweg (commented):
Thanks for the typo fixes and removed porting notes, these changes look good!
I think the docstrings have a "mathematical typo"; can you fix them, please?

Mathlib/Analysis/Convex/Slope.lean
LorenzoLuccioli: Thank you for your comment.

Actually I had the same doubt, but I decided to write it like this in order to mimic the style of the other doc-strings in the same file. I thought that the fact that everything happens in `s` had been left implicit on purpose, so that the docs are easier to read.

What do you think is better: leave everything like it is, modify only the 3 new doc-strings or modify all the doc-strings in the file?

grunweg: Good point! I had not noticed this fact about the other doc-strings. I have asked on zulip now.

YaelDillies: I'm personally happy with this small lie.

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13930, opened 2024-06-18
feat: add `mem_ConjAct` theorem for `ConjAct` in subgroup

Louis-Le-Grand:
Add a theorem stating that an element $x$ is in ``ConjAct.toConjAct g • U`` if and only if there exists a $u \in U$ such that $g u g^{-1} = x$.

---


eric-wieser:
I'm not convinced by this lemma; it follows from
```lean
simp_rw [Subgroup.mem_smul_pointwise_iff_exists, ConjAct.smul_def,
      ConjAct.ofConjAct_toConjAct]
```

I think the missing lemma here is the one that says `ConjAct.toConjAct g • u = g * u * g⁻¹`, not this one, as we don't really want to repeat the lemma for every possible action.

PR #13932, opened 2024-06-18
[Merged by Bors] - feat(MeasureTheory/Function/ConditionalExpectation): relax `integral_condexp` assumption and add `integral_condexp_indicator`

gaetanserre:
Remove the `Integrable` assumption of `integral_condexp` and add the total probability law using `condexp` as conditional probability.

---

### DIFF
--- a/Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
+++ b/Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
@@ -420,4 +420,54 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
   exact tendsto_nhds_unique_of_eventuallyEq hcond_gs hcond_fs (eventually_of_forall hn_eq)
 #align measure_theory.tendsto_condexp_unique MeasureTheory.tendsto_condexp_unique
 
+open Set
+
+set_option quotPrecheck false in
+scoped notation μ "[" X "]" => ∫ x, ↑(X x) ∂μ
+
+scoped notation μ "⟦" s "|" m "⟧" =>
+  MeasureTheory.condexp m μ (indicator s fun ω => (1 : ℝ))
+
+/-- Total probability law using `condexp` as conditional probability. -/
+theorem total_law_condexp [mF : MeasurableSpace F] {Y : α → F} (hY : Measurable Y)
+    [IsProbabilityMeasure μ] {A : Set α} (hA : MeasurableSet[mF.comap Y] A)
+    [SigmaFinite (μ.trim (measurable_iff_comap_le.mp hY))] :
+    (μ A).toReal = μ[μ⟦A | mF.comap Y⟧] := by
+  have hAm0 : MeasurableSet A := (measurable_iff_comap_le.mp hY) A hA
+  have total_exp :
+      ∫ x, (μ[A.indicator fun _ ↦ (1 : ℝ)|mF.comap Y]) x ∂μ = μ[A.indicator (fun x ↦ 1)]
+      := by
+    have hI : HasFiniteIntegral (A.indicator (fun (_ : α) ↦ (1 : ℝ))) μ := by
+      unfold HasFiniteIntegral
+      have rmv_norm : ∀ a, ‖A.indicator (fun (_ : α) ↦ (1 : ℝ)) a‖₊ =
+          A.indicator (fun (_ : α) ↦ (1 : ℝ≥0∞)) a := by
+        intro a
+        by_cases ha : a ∈ A
+        · rw [indicator_of_mem ha fun _ ↦ 1, indicator_of_mem ha fun _ ↦ 1, nnnorm_one]
+          rfl
+        rw [indicator_of_not_mem ha fun _ ↦ 1, indicator_of_not_mem ha fun _ ↦ 1,
+          _root_.nnnorm_zero]
+        rfl
+      simp_rw [rmv_norm]
+      calc ∫⁻ a, A.indicator (fun _ ↦ 1) a ∂μ = ∫⁻ _ in A, 1 ∂μ :=
+        lintegral_indicator (fun _ ↦ 1) hAm0
+      _ = μ A := set_lintegral_one A
+      _ < ⊤ := measure_lt_top μ A
+    have hAEmα : AEStronglyMeasurable (A.indicator (fun (_ : α) ↦ (1 : ℝ))) μ := by
+      suffices StronglyMeasurable (A.indicator (fun (_ : α) ↦ (1 : ℝ))) by
+        exact StronglyMeasurable.aestronglyMeasurable this
+      exact StronglyMeasurable.mono
+        (StronglyMeasurable.indicator stronglyMeasurable_const hA)
+        (measurable_iff_comap_le.mp hY)
+    exact integral_condexp (measurable_iff_comap_le.mp hY) ⟨hAEmα, hI⟩
+  rw [total_exp, integral_indicator hAm0]
+  have ae_pos : 0 ≤ᵐ[μ.restrict A] (fun _ ↦ (1 : ℝ)) := by
+      suffices 0 ≤ (fun (_ : α) ↦ (1 : ℝ)) by exact ae_of_all (μ.restrict A) this
+      intro _
+      simp only [Pi.zero_apply, zero_le_one]
+  rw [
+    integral_eq_lintegral_of_nonneg_ae ae_pos aestronglyMeasurable_const,
+    ENNReal.ofReal_one, set_lintegral_one A
+  ]
+
 end MeasureTheory
--- a/Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
+++ b/Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
@@ -435,8 +435,7 @@ theorem total_law_condexp [mF : MeasurableSpace F] {Y : α → F} (hY : Measurab
     (μ A).toReal = μ[μ⟦A | mF.comap Y⟧] := by
   have hAm0 : MeasurableSet A := (measurable_iff_comap_le.mp hY) A hA
   have total_exp :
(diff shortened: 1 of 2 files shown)


Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -420,4 +420,48 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
...
+      calc ∫⁻ a, A.indicator (fun _ ↦ 1) a ∂μ = ∫⁻ _ in A, 1 ∂μ :=
+        lintegral_indicator (fun _ ↦ 1) hAm0
+      _ = μ A := set_lintegral_one A
+      _ < ⊤ := measure_lt_top μ A
+    have hAEmα : AEStronglyMeasurable (A.indicator (fun (_ : α) ↦ (1 : ℝ))) μ := by
+      suffices StronglyMeasurable (A.indicator (fun (_ : α) ↦ (1 : ℝ))) by
+        exact StronglyMeasurable.aestronglyMeasurable this
+      exact StronglyMeasurable.mono
+        (StronglyMeasurable.indicator stronglyMeasurable_const hA)
+        (measurable_iff_comap_le.mp hY)
```
RemyDegenne, line 455: Exactly the same proof, written in a compact way:
```suggestion
    have hAEmα : AEStronglyMeasurable (A.indicator (fun _ ↦ (1 : ℝ))) μ :=
      ((stronglyMeasurable_const.indicator hA).mono hY.comap_le).aestronglyMeasurable
```

Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -420,4 +420,48 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
   exact tendsto_nhds_unique_of_eventuallyEq hcond_gs hcond_fs (eventually_of_forall hn_eq)
 #align measure_theory.tendsto_condexp_unique MeasureTheory.tendsto_condexp_unique
 
+open Set
+
```
RemyDegenne, line 424: It saves you a single `Set.` in the code. That does not seem worth it. You can event replace that `Set.indicator A` by `A.indicator`.
```suggestion
```

Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -420,4 +420,48 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
...
+/-- Total probability law using `condexp` as conditional probability. -/
+theorem total_law_condexp [mF : MeasurableSpace F] {Y : α → F} (hY : Measurable Y)
+    [IsProbabilityMeasure μ] {A : Set α} (hA : MeasurableSet[mF.comap Y] A)
```
RemyDegenne, line 427: Can you extend the result to a finite measure instead of probability measure?

RemyDegenne (commented):
Thanks!

RemyDegenne:
Actually instead of applying all the golfing suggestions above, here is a short proof of your lemma:
```
/-- Total probability law using `condexp` as conditional probability. -/
theorem total_law_condexp [mF : MeasurableSpace F] {Y : α → F} (hY : Measurable Y)
    [IsProbabilityMeasure μ] {A : Set α} (hA : MeasurableSet[mF.comap Y] A) :
    (μ A).toReal = ∫ x, (μ[(A.indicator fun _ ↦ (1 : ℝ)) | mF.comap Y]) x ∂μ := by
  have hAm0 : MeasurableSet A := hY.comap_le A hA
  have total_exp : ∫ x, (μ[A.indicator fun _ ↦ (1 : ℝ)|mF.comap Y]) x ∂μ
      = ∫ x, A.indicator (fun x ↦ 1) x ∂ μ :=
    integral_condexp (measurable_iff_comap_le.mp hY) ((integrable_const _).indicator hAm0)
  rw [total_exp, integral_indicator hAm0, setIntegral_const, smul_eq_mul, mul_one]
```
(Please still read my comments about the name and the assumptions)

JasonKYi:
This lemma seems strange to me as you assume `MeasurableSet[mF.comap Y] A` and so the conditional expectation does not do anything to the indicator. Perhaps you only want ``MeasurableSet A`?

gaetanserre:
Thank you @RemyDegenne  for all your suggestions. I tried used your shorter proof and changed the name. I have taken note of all previous suggestions for future proofs, thanks.
@JasonKYi: Indeed, I changed the hypothesis to only `MeasurableSet A`.

RemyDegenne:
Please check every comment above and mark as resolved all those that you addressed. If something remains (like my comment about switching the two sides of the equality in the result of the lemma, or my question about finite measures), please address those.

gaetanserre:
Sorry for that. I think it's done now.

JasonKYi:
`integral_condexp` can be generalized so that you don't need integrability. Namely 
```lean
lemma foo {α : Type u_1} {F' : Type u_3}
  [NormedAddCommGroup F'] [NormedSpace ℝ F'] [CompleteSpace F']
  {m : MeasurableSpace α} {m0 : MeasurableSpace α} {μ : MeasureTheory.Measure α}
  (f : α → F') (hm : m ≤ m0) [hμm : MeasureTheory.SigmaFinite (μ.trim hm)] :
∫ (x : α), MeasureTheory.condexp m μ f x ∂μ = ∫ (x : α), f x ∂μ := by
  by_cases hf : MeasureTheory.Integrable f μ
  · exact integral_condexp hm hf
  · simp only [condexp_undef hf, Pi.zero_apply, integral_zero, integral_undef hf]
```
Can you generalize `integral_condexp` and then you can require sigma finite instead of finite measure.

Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -420,4 +420,10 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
   exact tendsto_nhds_unique_of_eventuallyEq hcond_gs hcond_fs (eventually_of_forall hn_eq)
 #align measure_theory.tendsto_condexp_unique MeasureTheory.tendsto_condexp_unique
 
+/-- Total probability law using `condexp` as conditional probability. -/
+theorem integral_condexp_indicator [mF : MeasurableSpace F] {Y : α → F} (hY : Measurable Y)
+    [SigmaFinite (μ.trim (measurable_iff_comap_le.mp hY))] {A : Set α} (hA : MeasurableSet A) :
```
RemyDegenne, line 425:
```suggestion
    [SigmaFinite (μ.trim hY.comap_le)] {A : Set α} (hA : MeasurableSet A) :
```

Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -420,4 +420,10 @@ theorem tendsto_condexp_unique (fs gs : ℕ → α → F') (f g : α → F')
   exact tendsto_nhds_unique_of_eventuallyEq hcond_gs hcond_fs (eventually_of_forall hn_eq)
 #align measure_theory.tendsto_condexp_unique MeasureTheory.tendsto_condexp_unique
 
+/-- Total probability law using `condexp` as conditional probability. -/
+theorem integral_condexp_indicator [mF : MeasurableSpace F] {Y : α → F} (hY : Measurable Y)
+    [SigmaFinite (μ.trim hY.comap_le)] {A : Set α} (hA : MeasurableSet A) :
+    ∫ x, (μ[(A.indicator fun _ ↦ (1 : ℝ)) | mF.comap Y]) x ∂μ = (μ A).toReal := by
+  rw [integral_condexp, integral_indicator hA, setIntegral_const, smul_eq_mul, mul_one]
```
RemyDegenne, line 427: Can you put this right after `integral_condexp`? Since it's a consequence of that lemma, they should go together.

RemyDegenne:
> Thank you! I did it and modified few proofs consequently. However, I'm wondering if it's possible to have a "nicer" assumption than `[SigmaFinite (μ.trim (measurable_iff_comap_le.mp hY))]`.

Sigma-finite with respect to the sub-sigma-algebra is the minimal assumption under which conditional expectations exist, if I recall correctly. The book Analysis in Banach Spaces, by Hytönen et al, is a good reference.

If your measure is finite, then the typeclass inference system will infer that for you, so the fact that it is somewhat complicated does not matter much.

gaetanserre:
> Sigma-finite with respect to the sub-sigma-algebra is the minimal assumption under which conditional expectations exist, if I recall correctly. The book Analysis in Banach Spaces, by Hytönen et al, is a good reference.
Ok I see. Indeed, it seems to be a fair assumption. I'll check this reference, thanks!

> If your measure is finite, then the typeclass inference system will infer that for you, so the fact that it is somewhat complicated does not matter much.
Indeed, it should be goof for most classical usage I guess.

Mathlib/MeasureTheory/Function/ConditionalExpectation/Basic.lean
```diff
@@ -226,13 +226,21 @@ theorem setIntegral_condexp (hm : m ≤ m0) [SigmaFinite (μ.trim hm)] (hf : Int
 
 @[deprecated (since := "2024-04-17")] alias set_integral_condexp := setIntegral_condexp
 
-theorem integral_condexp (hm : m ≤ m0) [hμm : SigmaFinite (μ.trim hm)] (hf : Integrable f μ) :
+theorem integral_condexp (hm : m ≤ m0) [hμm : SigmaFinite (μ.trim hm)] :
     ∫ x, (μ[f|m]) x ∂μ = ∫ x, f x ∂μ := by
-  suffices ∫ x in Set.univ, (μ[f|m]) x ∂μ = ∫ x in Set.univ, f x ∂μ by
-    simp_rw [integral_univ] at this; exact this
-  exact setIntegral_condexp hm hf (@MeasurableSet.univ _ m)
+  by_cases hf : MeasureTheory.Integrable f μ
```
JasonKYi, line 231:
```suggestion
  by_cases hf : Integrable f μ
```

JasonKYi:
Thanks! 
maintainer merge

PR #13938, opened 2024-06-18
[Merged by Bors] - feat: Miscellaneous small ENNReal lemmas.

kkytola:
`ENNReal.*`: `.toNNReal_toReal_eq`, `.sub_add_eq_add_sub`, `.zpow_neg`, `.zpow_sub`, `.tendsto_pow_atTop_nhds_top` + instance `MeasurableSMul₂ ℝ≥0 ℝ≥0∞`

---

- [x] depends on: #14591 [the right generality of the liminf and limsup lemmas require generalizations implemented in this PR]


Mathlib/Topology/Instances/ENNReal.lean
```diff
@@ -1706,6 +1708,38 @@ lemma limsup_toReal_eq {ι : Type*} {F : Filter ι} [NeBot F] {b : ℝ≥0∞} (
   rw [key]
   rfl
 
+/-- If `xs : ι → ℝ≥0∞` is bounded, then we have `liminf (toNNReal ∘ xs) = toNNReal (liminf xs)`. -/
+lemma liminf_toNNReal_eq {b : ℝ≥0∞} (b_ne_top : b ≠ ∞)
```
sgouezel, line 1712: Isn't it enough to assume the weaker condition that `xs i` is eventually not equal to ` ∞`? If the liminf is infinity in this case, it means that the sequence is tending to infinity, so the liminf in NNReal is probably zero (what other choice of junk value could we have? I haven't checked).

Mathlib/Topology/Instances/ENNReal.lean
```diff
@@ -1706,6 +1708,38 @@ lemma limsup_toReal_eq {ι : Type*} {F : Filter ι} [NeBot F] {b : ℝ≥0∞} (
...
+    · refine ⟨b.toNNReal, ?_⟩
+      simp only [eventually_map]
+      filter_upwards [le_b] with i xs_bdd
+      exact ((toReal_le_toReal (ne_top_of_le_ne_top b_ne_top xs_bdd) b_ne_top).mpr xs_bdd).trans
+          (Preorder.le_refl b.toReal)
+    · refine ⟨0, by simp only [ge_iff_le, eventually_map, toReal_nonneg, eventually_true]⟩
+  simp [← obs, liminf_toReal_eq b_ne_top le_b]
+
+/-- If `xs : ι → ℝ≥0∞` is bounded, then we have `liminf (toNNReal ∘ xs) = toNNReal (liminf xs)`. -/
+lemma limsup_toNNReal_eq {b : ℝ≥0∞} (b_ne_top : b ≠ ∞)
```
sgouezel, line 1728: Same comment here.

Mathlib/Topology/Instances/ENNReal.lean
kkytola: True. The existing lemma `Monotone.map_liminf_of_continuousAt` used here had suboptimal assumptions, so it looks like the right course is to generalize that first. I made PR #14591 for the generalization. With it, I will return and generalize this lemma (and `liminf_toReal_eq` above, which had a similar suboptimal assumption).

kkytola:
Should I separate the generalization of the liminf / limsup lemmas from the other small ENNReal lemmas? The original PR was around 50 lines diff, but the requested generalization of the liminf lemma ended up relying on PR #14591 to generalize liminf/limsup lemmas to assume only coboundedness in one direction plus still ENNReal specific adaptations which alone are many times the size of the original simple PR.

(I obviously think it is absolutely the right thing to do to fix the generality of the ENNReal liminf lemma as well as to fix the generality of liminf and limsup lemmas in all conditionally complete linear orders with order topology, so thanks Sébastien for the careful and useful review comment!)

j-loreaux:
Yeah, it looks like you already split most of this off. Can you remove the bits from the other PR?

Mathlib/Topology/Instances/ENNReal.lean
kkytola: I moved the generalization of this lemma and the ones below to #15115.

Mathlib/Data/ENNReal/Basic.lean
```diff
@@ -223,6 +223,9 @@ theorem ofReal_eq_coe_nnreal {x : ℝ} (h : 0 ≤ x) :
 
 @[simp] theorem toReal_nonneg {a : ℝ≥0∞} : 0 ≤ a.toReal := a.toNNReal.2
 
+@[simp] theorem toNNReal_toReal_eq (z : ℝ≥0∞) : z.toReal.toNNReal = z.toNNReal := by
+  ext; simp only [Real.coe_toNNReal', ge_iff_le, toReal_nonneg, max_eq_left]; rfl
```
Ruben-VandeVelde, line 227: Why do you need the rfl here? Missing lemmas?

Ruben-VandeVelde (approved):

Mathlib/Data/ENNReal/Basic.lean
kkytola: The goal after this `simp only [...]`, before `rfl` was essentially the following lemma, which indeed seemed to be missing. However, creating it and making it a `@[simp]` lemma causes a loop in a simp later, so maybe there is a reason... On the other hand, the RHS below should definitely in my opinion be the simp normal form of this expression!

```
theorem coe_toNNReal_eq_toReal (z : ℝ≥0∞) : (z.toNNReal : ℝ) = z.toReal := rfl
```

I added it now without `@[simp]` tag, and providing this to `simp` now avoids the last `rfl`.

Mathlib/Data/ENNReal/Operations.lean
```diff
@@ -334,6 +334,12 @@ protected theorem add_sub_cancel_left (ha : a ≠ ∞) : a + b - a = b :=
 protected theorem add_sub_cancel_right (hb : b ≠ ∞) : a + b - b = a :=
   (cancel_of_ne hb).add_tsub_cancel_right
 
+protected theorem sub_add_eq_add_sub (hab : b ≤ a) (b_ne_top : b ≠ ∞) (c_ne_top : c ≠ ∞) :
+    a - b + c = a + c - b := by
```
urkud, line 338: This seems to be true for `c = ∞`: both sides are equal to `∞`.

kkytola: Indeed, thanks! I will try to drop the hypothesis.

Mathlib/Analysis/SpecificLimits/Basic.lean
```diff
@@ -243,6 +243,12 @@ protected theorem ENNReal.tendsto_pow_atTop_nhds_zero_iff {r : ℝ≥0∞} :
   norm_cast at h ⊢
   exact NNReal.tendsto_pow_atTop_nhds_zero_iff.mp h
 
+protected theorem ENNReal.tendsto_pow_atTop_nhds_top {r : ℝ≥0∞} (hr : 1 < r) :
+    Tendsto (fun n ↦ r^n) atTop (𝓝 ∞) := by
```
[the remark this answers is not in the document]
kkytola: Sorry, do you mean making `ENNReal.tendsto_pow_atTop_nhds_top` a `simp` lemma? I could do that, but I thought would be unlikely that users are ever in exactly a position where this lemma could fire in a `simp` call.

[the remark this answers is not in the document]
kkytola: Oh, do you mean making it an "iff" statement with `1<r` are the RHS simp normal form? That makes sense! Thanks!

Mathlib/MeasureTheory/Constructions/BorelSpace/Real.lean
```diff
@@ -440,6 +440,11 @@ theorem AEMeasurable.coe_ereal_ennreal {f : α → ℝ≥0∞} {μ : Measure α}
 
 namespace NNReal
 
+instance : MeasurableSMul₂ ℝ≥0 ℝ≥0∞ where
+  measurable_smul :=
+    measurable_mul.comp <|
+      Measurable.prod_mk (ENNReal.continuous_coe.measurable.comp measurable_fst) measurable_snd
```
j-loreaux, line 446: unless the attributes aren't declared yet, I think this works:
```suggestion
instance : MeasurableSMul₂ ℝ≥0 ℝ≥0∞ where
  measurable_smul := show Measurable fun r : ℝ≥0 × ℝ≥0∞ ↦ (r.1 : ℝ≥0) * r.2 by fun_prop
```

Mathlib/Data/ENNReal/Inv.lean
```diff
@@ -586,5 +586,13 @@ protected theorem zpow_add {x : ℝ≥0∞} (hx : x ≠ 0) (h'x : x ≠ ∞) (m
   replace hx : x ≠ 0 := by simpa only [Ne, coe_eq_zero] using hx
   simp only [← coe_zpow hx, zpow_add₀ hx, coe_mul]
 
+protected theorem zpow_neg {x : ℝ≥0∞} (x_ne_zero : x ≠ 0) (x_ne_top : x ≠ ⊤) (m : ℤ) :
+    x^(-m) = (x^m)⁻¹ :=
```
j-loreaux, line 590: style
```suggestion
protected theorem zpow_neg {x : ℝ≥0∞} (x_ne_zero : x ≠ 0) (x_ne_top : x ≠ ⊤) (m : ℤ) :
    x ^ (-m) = (x ^ m)⁻¹ :=
```

Mathlib/Data/ENNReal/Inv.lean
```diff
@@ -586,5 +586,13 @@ protected theorem zpow_add {x : ℝ≥0∞} (hx : x ≠ 0) (h'x : x ≠ ∞) (m
...
+  ENNReal.eq_inv_of_mul_eq_one_left (by simp [← ENNReal.zpow_add x_ne_zero x_ne_top])
+
+protected theorem zpow_sub {x : ℝ≥0∞} (x_ne_zero : x ≠ 0) (x_ne_top : x ≠ ⊤) (m n : ℤ) :
+    x^(m-n) = (x^m) * (x^n)⁻¹ := by
```
j-loreaux, line 594: arguably, because Lean doesn't need them for parsing, we should drop the parens around `x ^ m` too, but I like keeping them for readability.
```suggestion
protected theorem zpow_sub {x : ℝ≥0∞} (x_ne_zero : x ≠ 0) (x_ne_top : x ≠ ⊤) (m n : ℤ) :
    x ^ (m - n) = (x ^ m) * (x ^ n)⁻¹ := by
```

Mathlib/Data/ENNReal/Basic.lean
```diff
@@ -223,6 +223,12 @@ theorem ofReal_eq_coe_nnreal {x : ℝ} (h : 0 ≤ x) :
 
 @[simp] theorem toReal_nonneg {a : ℝ≥0∞} : 0 ≤ a.toReal := a.toNNReal.2
 
+-- Note: Tagging this with `@[simp]` causes loops, but why?
```
j-loreaux, line 226: Probably you want to tag this `@[norm_cast]`, not `@[simp]`.

j-loreaux (approved):
Assuming the `norm_cast` suggestion works and you implement the other suggestions:

bors d+

Mathlib/Data/ENNReal/Basic.lean
kkytola: Thanks! I was not (actively) aware of `@[norm_cast]` tag, but it definitely seems appropriate. I will try and see the consequences (especially no loops).

PR #13939, opened 2024-06-18
[Merged by Bors] - feat(NumberTheory/MulChar): extend API for multiplicative characters

MichaelStollBayreuth:
This adds some more API lemmas for multiplicative characters.

---


Mathlib/NumberTheory/MulChar/Lemmas.lean
```diff
@@ -0,0 +1,207 @@
...
+/-- Define the conjugation (`star`) of a multiplicative character by conjugating pointwise. -/
+@[simps!]
+def starComp [StarRing R'] (χ : MulChar R R') : MulChar R R' :=
+  χ.ringHomComp (starRingEnd R')
+
+
+variable [Fintype R] [DecidableEq R]
+
+/-- The values of a multiplicative character on `R` are `n`th roots of unity, where `n = #Rˣ`. -/
+lemma val_mem_rootsOfUnity (a : Rˣ) {χ : MulChar R R'} :
```
jcommelin, line 38: `val` is usually the name component that refers to `Subtype.val`.
I think in this case `apply` is what should be used.

Idem for the `val`s in lemmas below.

MichaelStollBayreuth: OK, thanks for pointing this out!

Mathlib/NumberTheory/MulChar/Lemmas.lean
```diff
@@ -0,0 +1,207 @@
...
+    ← coe_equivToUnitHom, ← coe_equivToUnitHom, Units.ext_iff]
+
+
+section Ring
+
+variable {R R' : Type*} [CommRing R] [CommRing R']
+
+/-- Define the conjugation (`star`) of a multiplicative character by conjugating pointwise. -/
+@[simps!]
+def starComp [StarRing R'] (χ : MulChar R R') : MulChar R R' :=
```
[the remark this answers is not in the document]
MichaelStollBayreuth: That would be `StarMul`, I think. I'm adding this (and use it to simplify some of the following statements).

jcommelin:
Thanks :tada:

bors merge

urkud:
bors merge

PR #13946, opened 2024-06-19
[Merged by Bors] - feat(MvPolynomial/PDeriv): Euler's (weighted) homogeneous identity

alreadydone:
If $f$ is a weighted homogeneous polynomial of degree $n$ with a finite set of variables $\\{X_i \mid i \in \sigma\\}$, then $n\cdot f = \sum_{i \in \sigma} w_i X_i \frac{\partial f}{\partial X_i}$, where $w_i$ are the weights.

---

- [x] depends on: #13978


alreadydone:
My current plan to fix the error is to split MvPolynomial out of the Kaehler file ... will do tomorrow.

Multramate:
Presumably (since you asked me to review this), we want to prove `Jacobian.polynomial_relation` as an analogue of `Affine.polynomial_relation`? On the other hand that can be proven directly by `linear_combination`...

alreadydone:
I don't think there's Affine.polynomial_relation, do you mean Projective? It's a bit painful to prove homogeneity, but I think it's a result we should have:
```
lemma polynomial_isWeightedHomogeneous : W'.polynomial.IsWeightedHomogeneous ![2, 3, 1] 6 := by
  rw [← mem_weightedHomogeneousSubmodule]
  refine sub_mem (add_mem (add_mem ?_ ?_) ?_) (add_mem (add_mem (add_mem ?_ ?_) ?_) ?_)
  · exact (isWeightedHomogeneous_X _ _ _).pow _
  · exact (((isWeightedHomogeneous_C ..).mul <| isWeightedHomogeneous_X ..).mul <|
      isWeightedHomogeneous_X ..).mul (isWeightedHomogeneous_X ..)
  · exact ((isWeightedHomogeneous_C ..).mul <| isWeightedHomogeneous_X ..).mul <|
      (isWeightedHomogeneous_X ..).pow _
  · exact (isWeightedHomogeneous_X _ _ _).pow _
  · exact ((isWeightedHomogeneous_C ..).mul <| (isWeightedHomogeneous_X ..).pow _).mul <|
      (isWeightedHomogeneous_X ..).pow _
  · exact ((isWeightedHomogeneous_C ..).mul <| isWeightedHomogeneous_X ..).mul <|
      (isWeightedHomogeneous_X ..).pow _
  · exact (isWeightedHomogeneous_C ..).mul <| (isWeightedHomogeneous_X ..).pow _
```
Once it's proven, `polynomial_relation` should be easy.

Mathlib/Algebra/MvPolynomial/PDeriv.lean
```diff
@@ -77,6 +78,20 @@ theorem pderiv_monomial {i : σ} :
     · simp
 #align mv_polynomial.pderiv_monomial MvPolynomial.pderiv_monomial
 
+lemma _root_.Finsupp.sub_add_single_one_cancel {m : σ →₀ ℕ} {i : σ} (hi : m i ≠ 0) :
```
semorrison, line 81: Please move to appropriate file.

Mathlib/Algebra/MvPolynomial/PDeriv.lean
```diff
@@ -5,6 +5,7 @@ Authors: Shing Tak Lam, Yury Kudryashov
 -/
 import Mathlib.Algebra.MvPolynomial.Derivation
 import Mathlib.Algebra.MvPolynomial.Variables
+import Mathlib.RingTheory.MvPolynomial.Homogeneous
```
semorrison, line 8: Please create a new file for this result, to avoid the import increase.

Mathlib/Algebra/MvPolynomial/PDeriv.lean
alreadydone: Moved.

Mathlib/Algebra/MvPolynomial/PDeriv.lean
alreadydone: Created new file, although importing PDeriv into Homogeneous instead only adds 6 trans imports.

Mathlib/Algebra/MvPolynomial/PDeriv.lean
```diff
@@ -72,6 +72,13 @@ theorem pderiv_monomial {i : σ} :
   · rw [Finsupp.not_mem_support_iff] at hi; simp [hi]
   · simp
 
+lemma X_mul_pderiv_monomial {i : σ} {m : σ →₀ ℕ} {r : R} :
```
riccardobrasca, line 75: Maybe some of the variable should be explicit?

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -105,6 +105,13 @@ theorem NonTorsionWeight.ne_zero [NonTorsionWeight w] (s : σ) :
   apply Nat.zero_ne_one.symm
   exact NonTorsionWeight.eq_zero_of_smul_eq_zero h
 
+variable {w} in
+lemma weight_sub_single_add {f : σ →₀ ℕ} {i : σ} (hi : f i ≠ 0) :
```
riccardobrasca, line 109: I think we can keep `w` explicit.

Mathlib/RingTheory/MvPolynomial/EulerIdentity.lean
```diff
@@ -0,0 +1,73 @@
...
+/-- Euler's identity for weighted homogeneous polynomials. -/
+theorem IsWeightedHomogeneous.sum_weight_X_mul_pderiv {w : σ → ℕ}
+    (h : φ.IsWeightedHomogeneous w n) : ∑ i : σ, w i • (X i * pderiv i φ) = n • φ := by
+  rw [← mem_weightedHomogeneousSubmodule, weightedHomogeneousSubmodule_eq_finsupp_supported,
+    supported_eq_span_single] at h
+  refine Submodule.span_induction ?_ ?_ (fun p q _ _ hp hq ↦ ?_) (fun r p _ h ↦ ?_) h
+  · rintro _ ⟨m, hm, rfl⟩
+    simp_rw [single_eq_monomial, X_mul_pderiv_monomial, smul_smul, ← sum_smul, mul_comm (w _)]
+    congr
+    rwa [Set.mem_setOf, weight_apply, sum_fintype] at hm
```
riccardobrasca, line 62: I didn't open VS Code, so I may be wrong, but the `rwa` tells me you have two goals here. Can you use the central dot if this is the case?

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Algebra/MvPolynomial/PDeriv.lean
alreadydone: Since each of `i` `m` and `r` appear in both side of the equality, I think this is good as is. You can see I don't need to specify any implicit arguments when rewriting with it.

Mathlib/Data/Finsupp/Weight.lean
alreadydone: That would mean `weight_sub_single_add hi` below will become `weight_sub_single_add _ hi` or `weight_sub_single_add w hi` ...

Mathlib/RingTheory/MvPolynomial/EulerIdentity.lean
alreadydone: There's a linter for multiple goals now. It's [Finsupp.sum_fintype](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/Finsupp.html#Finsupp.sum_fintype) that creates a side goal.

Mathlib/Data/Finsupp/Weight.lean
riccardobrasca: OK, no problem.

alreadydone:
Thanks!
bors r+

Mathlib/Data/Finsupp/Weight.lean
alreadydone: I assumed that you mean keeping {w} implicit is no problem, but now I realize you might mean adding an underscore is no problem ... Can you clarify and should I make a PR to change anything?

PR #13950, opened 2024-06-19
[Merged by Bors] - feat: `ExistsAddOfLE` lemmas

YaelDillies:
Add a few convenient versions of `le_iff_exists_add`.

From LeanAPAP

---

### DIFF
--- a/Mathlib/Algebra/Order/Monoid/Unbundled/ExistsOfLE.lean
+++ b/Mathlib/Algebra/Order/Monoid/Unbundled/ExistsOfLE.lean
@@ -1,8 +1,9 @@
 /-
-Copyright (c) 2016 Jeremy Avigad. All rights reserved.
+Copyright (c) 2021 Peter Nelson. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Jeremy Avigad, Leonardo de Moura, Mario Carneiro, Johannes Hölzl
+Authors: Peter Nelson, Yaël Dillies
 -/
+import Mathlib.Algebra.Order.Group.Synonym
 import Mathlib.Algebra.Order.Monoid.Unbundled.Basic
 import Mathlib.Order.MinMax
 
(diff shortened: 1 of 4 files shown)


eric-wieser:
> Interestingly, some statements generalise from `ExistsAddOfLE α` to `ExistsAddOfLEOrGE α`, which makes them dualisable.

Can you list these? There are none currently in the PR, right?

YaelDillies:
I'm not going to copy the entire diff of #11448, so cf #11448

semorrison:
This feels quite baroque to me, and I'm sceptical that such a notion belongs in a basic file.

YaelDillies:
> This feels quite baroque to me

Sorry, I'm not sure what this comment means concretely for this PR.

I can move the definition to a new file, but then the lemmas I'm generalising should move too, which is a bit weird given that this is supposed to be an in-place generalisation.

semorrison:
I'm saying that I don't believe this is a sufficiently natural notion to live in this file, and you should find a different way to achieve your application!

YaelDillies:
The only different way I see is the state at the first commit (I'm on my phone so I can't easily check this is actually the first on this PR right now, but if you don't see it I can dig up the reflog) and you can see that it is just a huge load of duplication.

Here's the basic issue distilled in one example: `a <= b -> c <= d -> a • d + b • c <=  a • c + b • d` is true both when `a b : Nat` and when `a b : OrderDual Nat` (and `c d : Real`, sya)


If that can give a bit more gravitas to this PR, I am currently writing a paper about the general strategy behind generalising inequalities using `smul`.

YaelDillies:
Update: I meant https://github.com/leanprover-community/mathlib4/pull/11448/commits/5cea0b9bd58be4113638510bc58153baefc34cdf

Other update: Seems like Matthew heard your call: #14390

YaelDillies:
As per https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2313950.20Dualisable.20notion.20of.20ExistsAddOfLE/near/465023218, I have found a way to generalise the rearrangement inequality even further and I do not need this new typeclasse anymore. Please review #15813, #15815, #11448 instead!

YaelDillies:
Actually, I need some lemmas in here, so I would reopen the PR but remove `ExistsAddOfLEOrGE` from the diff

PR #13952, opened 2024-06-19
[Merged by Bors] - feat: Product of convex hulls

YaelDillies:
From LeanCamCombi

---


Mathlib/GroupTheory/Perm/Cycle/Basic.lean
```diff
@@ -1094,8 +1094,9 @@ namespace Finset
 variable [Semiring α] [AddCommMonoid β] [Module α β] {s : Finset ι} {σ : Perm ι}
 
 theorem sum_smul_sum_eq_sum_perm (hσ : σ.IsCycleOn s) (f : ι → α) (g : ι → β) :
-    ((∑ i ∈ s, f i) • ∑ i ∈ s, g i) = ∑ k ∈ range s.card, ∑ i ∈ s, f i • g ((σ ^ k) i) := by
-  simp_rw [sum_smul_sum, product_self_eq_disjiUnion_perm hσ, sum_disjiUnion, sum_map]
+    (∑ i ∈ s, f i) • ∑ i ∈ s, g i = ∑ k ∈ range s.card, ∑ i ∈ s, f i • g ((σ ^ k) i) := by
+  rw [sum_smul_sum, ← sum_product']
+  simp_rw [product_self_eq_disjiUnion_perm hσ, sum_disjiUnion, sum_map]
```
eric-wieser, line 1099: Why did this change?

Mathlib/Analysis/Convex/Combination.lean
```diff
@@ -350,6 +351,30 @@ theorem convexHull_eq (s : Set E) : convexHull R s =
...
+  · rintro ⟨ι, _, w, z, h⟩
+    exact ⟨ι, Finset.univ, w, z, by simpa [centerMass_eq_of_sum_1 _ _ h.2.1] using h⟩
+
+/-- Universe polymorphic version of the reverse implication of `mem_convexHull_iff_exists_fintype`.
+-/
+lemma mem_convexHull_of_exists_fintype {s : Set E} {x : E} [Fintype ι] (w : ι → R) (z : ι → E)
+    (hw₀ : ∀ i, 0 ≤ w i) (hw₁ : ∑ i, w i = 1) (hz : ∀ i, z i ∈ s) (hx : ∑ i, w i • z i = x) :
+    x ∈ convexHull R s := by
+  rw [← hx, ← centerMass_eq_of_sum_1 _ _ hw₁]
+  exact centerMass_mem_convexHull _ (by simpa using hw₀) (by simp [hw₁]) (by simpa using hz)
```
eric-wieser, line 376: Why not put this before `mem_convexHull_iff_exists_fintype` so that you can use it to prove one direction?

Mathlib/Analysis/Convex/Combination.lean
```diff
@@ -590,3 +589,55 @@ Note that `AffineIndependent.convexHull_inter` should be more versatile in most
...
+      congr! with j _ g _
+      exact g.2
+    _ = ∑ j : κ i, w i j • f i j := ?_
+    _ = x i := hf _
+  congr! with j _
+  calc
+    ∑ g : {g : ∀ k, κ k // g i = j}, ∏ k, w k (g.1 k)
+      = ∑ g ∈ piFinset fun k ↦ if hk : k = i then hk ▸ {j} else univ, ∏ k, w k (g k) :=
+      Finset.sum_bij' (fun g _ ↦ g) (fun g hg ↦ ⟨g, by simpa using mem_piFinset.1 hg i⟩)
+        (by aesop) (by simp) (by simp) (by simp) (by simp)
```
eric-wieser, line 633: Is there a `fiberwise` lemma somewhere for this?

eric-wieser:
bors d+

Mathlib/GroupTheory/Perm/Cycle/Basic.lean
YaelDillies: The statement of `sum_smul_sum` changed to match `sum_mul_sum`.

Mathlib/Analysis/Convex/Combination.lean
YaelDillies: Yes I can apply `sum_fiberwise` (or one of its relatives) but there is no way for it to return me the correct RHS (unless I basically write the equality I need as a very specific lemma) so then I need to use `sum_bij` (or one of its relatives) anyway.

If you find a better way, I'm all ears.

Mathlib/Analysis/Convex/Combination.lean
YaelDillies: Doesn't actually golf anything, but okay

YaelDillies:
bors merge

PR #13957, opened 2024-06-19
[Merged by Bors] - feat(Data): Option.get!_some and Option.get!_none

javra:
Analogous to the simp lemmas for `getD`, for example.

---


semorrison:
bors merge

PR #13959, opened 2024-06-19
[Merged by Bors] - chore/perf(Analysis): replace continuity -> fun_prop

grunweg:
This is exhaustive, with the exception of one goal where the `continuity` proof is seemingly not about continuity at all...
We tag two lemmas with `fun_prop` which were continuity lemmas.

In some cases, this speeds up proofs significantly; we also resolve two porting notes about continuity not applying.

---

### DIFF
--- a/Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
+++ b/Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
@@ -40,7 +40,7 @@ noncomputable def homeomorphUnitSphereProd :
   continuous_toFun := by
     refine .prod_mk (.codRestrict (.smul (.inv₀ ?_ ?_) ?_) _) ?_
     · fun_prop
-    · continuity
+    · simp
     · fun_prop
-    · apply Continuous.subtype_mk (by fun_prop)
+    · fun_prop
   continuous_invFun := by apply Continuous.subtype_mk (by fun_prop)
--- a/Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
+++ b/Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
@@ -38,5 +38,9 @@ noncomputable def homeomorphUnitSphereProd :
     rw [mem_Ioi] at hr
     ext <;> simp [hx, norm_smul, hr.le, abs_of_pos hr, hr.ne']
   continuous_toFun := by
-    refine .prod_mk (.codRestrict (.smul (.inv₀ ?_ ?_) ?_) _) ?_ <;> continuity
-  continuous_invFun := by continuity
+    refine .prod_mk (.codRestrict (.smul (.inv₀ ?_ ?_) ?_) _) ?_
+    · fun_prop
+    · continuity
+    · fun_prop
+    · apply Continuous.subtype_mk (by fun_prop)
+  continuous_invFun := by apply Continuous.subtype_mk (by fun_prop)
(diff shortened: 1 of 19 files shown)


grunweg:
!bench

Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
```diff
@@ -38,5 +38,9 @@ noncomputable def homeomorphUnitSphereProd :
     rw [mem_Ioi] at hr
     ext <;> simp [hx, norm_smul, hr.le, abs_of_pos hr, hr.ne']
   continuous_toFun := by
-    refine .prod_mk (.codRestrict (.smul (.inv₀ ?_ ?_) ?_) _) ?_ <;> continuity
-  continuous_invFun := by continuity
+    refine .prod_mk (.codRestrict (.smul (.inv₀ ?_ ?_) ?_) _) ?_
+    · fun_prop
+    · continuity
```
grunweg, line 43: The output of continuity? looks very unrelated to continuity...

Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
```diff
@@ -38,5 +38,9 @@ noncomputable def homeomorphUnitSphereProd :
...
+    · fun_prop
+    · apply Continuous.subtype_mk (by fun_prop)
```
grunweg, line 45: It would be nice to investigate why Continuous.subtype_mk (which is tagged fun_prop) doesn't fire here.

grunweg:
!bench

Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
sgouezel, line 43:
```suggestion
    · simp
```
would make more sense.

Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
sgouezel:
```suggestion
    · fun_prop
```
works just fine, right?

Mathlib/Analysis/ODE/PicardLindelof.lean
```diff
@@ -304,8 +304,7 @@ theorem dist_next_apply_le_of_le {f₁ f₂ : FunSpace v} {n : ℕ} {d : ℝ}
       refine norm_integral_le_of_norm_le (Continuous.integrableOn_uIoc ?_) ?_
       · -- Porting note: was `continuity`
         refine .mul continuous_const <| .mul (.div_const ?_ _) continuous_const
```
sgouezel, line 306: All this block can be replaced with just `fun_prop`. Move it even to the `refine` above?

sgouezel (commented):
Looks great, thanks!

Mathlib/Analysis/NormedSpace/SphereNormEquiv.lean
grunweg: Oh, indeed it works here; changed. (It does not work on the next line.)

Mathlib/Analysis/ODE/PicardLindelof.lean
grunweg: Done, and unindented the refine.

grunweg:
Thank you for the review! I have addressed all comments.
awaiting-review

sgouezel:
bors r+
Thanks!

PR #13960, opened 2024-06-19
[Merged by Bors] - feat(Data/Matroid/Map): maps between matroids

apnelson1:
This PR creates a new file with the machinery for transferring matroid structures between types. The most important definitions are `Matroid.map`, which pushes a `M : Matroid α` along `f : α → β` to give an `N : Matroid β`, and
`Matroid.comap`,  which pulls `N : Matroid β` back along `f : α → β` to get an `M : Matroid α`. The former requires `f` to be injective on the ground set of `M`, but the latter doesn't. 

The original decision to define matroids with an embedded ground set within a type rather than just a structure on a type came with tradeoffs, and it is in places like this that they show up the most. For this reason, the file also defines several variants of `map` where the range/domain of the function in question is a coerced set rather than a type. 

There is nothing mathematically interesting going on with these definitions, and maps and comaps are not central to matroid theory in actual research, but they are very useful in making formalization painless, as they save the work of painstakingly checking matroid axioms in many cases. They will be used in future PRs to define things like direct sums and linearly representable matroids. 

---


Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,670 @@
...
+
+lemma restrictSubtype_inter_indep_iff :
+    (M.restrictSubtype X).Indep (X ↓∩ I) ↔ M.Indep (X ∩ I) := by
+  simp [restrictSubtype, Subtype.val_injective.injOn]
+
+lemma eq_of_restrictSubtype_eq (hM : M.E = E) (hN : N.E = E)
+    (h : M.restrictSubtype E = N.restrictSubtype E) : M = N := by
+  subst hM
+  refine eq_of_indep_iff_indep_forall (by rw [hN]) (fun I hI ↦ ?_)
+  rwa [← restrictSubtype_indep_iff_of_subset hI, h, restrictSubtype_indep_iff_of_subset]
```
qawbecrdtey, line 631: If the matroid `N` is only used for this lemma in section `restrictSubtype`, wouldn't it be better to declare the variable here?

Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,669 @@
...
+bipartite graph that may not correspond to a function at all [See Oxley, Thm 11.2.12].
+It would have been nice to use this more general construction as a basis for the definition
+of both `Matroid.map` and `Matroid.comap`.
+
+Unfortunately, we can't do this, because the construction doesn't extend to infinite matroids.
+Specifically, if `M₁` and `M₂` are matroids on the same type `α`,
+and `f` is the natural function from `α ⊕ α` to `α`,
+then the images under `f` of the independent sets of the direct sum `M₁ ⊕ M₂` are
+the independent sets of a matroid if and only if the union of `M₁` and `M₂` is a matroid,
+and unions do not exist for some pairs of infinite matroids: see [1].
```
jcommelin, line 89:
```suggestion
and unions do not exist for some pairs of infinite matroids: see [aignerhorev2012infinite].
```

Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,669 @@
...
+For this reason, `Matroid.map` requires injectivity to be well-defined in general.
+
+## TODO
+
+* Bundled matroid isomorphisms.
+
+## References
+
+[1] : E. Aigner-Horev, J. Carmesin, J. Fröhlic, Infinite Matroid Union, arXiv:1111.0602 [math.CO]
```
jcommelin, line 98:
```suggestion
[aignerhorev2012infinite] : E. Aigner-Horev, J. Carmesin, J. Fröhlic, Infinite Matroid Union, arXiv:1111.0602 [math.CO]
```

Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,669 @@
...
+indeed, `(M.map f _).Indep I ↔ M.Indep (f ⁻¹' I) ∧ I ⊆ range f`.
+If `f` is an equivalence we have `(M.map f _).Indep I ↔ M.Indep (f.symm '' I)`.
+In order that these stronger statements can be `@[simp]`,
+we define `mapEmbedding` and `mapEquiv` separately from `map`.
+
+## Notes
+
+For finite matroids, both maps and comaps are a special case of a construction of
+Perfect (1969) in which a matroid structure can be transported across an arbitrary
+bipartite graph that may not correspond to a function at all [See Oxley, Thm 11.2.12].
```
jcommelin, line 80: Do "Perfect (1969)" and "Oxley, ..." deserve entries in the `.bib` file?

Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,669 @@
...
+It would have been nice to use this more general construction as a basis for the definition
+of both `Matroid.map` and `Matroid.comap`.
+
+Unfortunately, we can't do this, because the construction doesn't extend to infinite matroids.
```
jcommelin, line 84: I guess you could have a `mapFinite` or something like that. But maybe that's too many variants?

Mathlib/Data/Matroid/Map.lean
```diff
@@ -0,0 +1,669 @@
...
+/-- The pullback of a matroid on `β` by a function `f : α → β` to a matroid on `α`.
+Elements with the same (nonloop) image are parallel and the ground set is `f ⁻¹' M.E`.
+The matroids `M.comap f` and `M ↾ range f` have isomorphic simplifications;
+the preimage of each nonloop of `M ↾ range f` is a parallel class. -/
+def comap (N : Matroid β) (f : α → β) : Matroid α :=
+  IndepMatroid.matroid <| IndepMatroid.mk
+  ( E := f ⁻¹' N.E )
+  ( Indep := fun I ↦ N.Indep (f '' I) ∧ InjOn f I )
+  ( indep_empty := by simp )
+  ( indep_subset := fun I J h hIJ ↦ ⟨h.1.subset (image_subset _ hIJ), InjOn.mono hIJ h.2⟩ )
```
jcommelin, line 116: Can't you use constructor notation, like
```suggestion
  { E := f ⁻¹' N.E
    Indep := fun I ↦ N.Indep (f '' I) ∧ InjOn f I
    indep_empty := by simp
    indep_subset := fun I J h hIJ ↦ ⟨h.1.subset (image_subset _ hIJ), InjOn.mono hIJ h.2⟩
```
etc...

jcommelin (commented):
Some minor questions.
Once you have `comap` up and running, the rest is very clean! Looks like a great design.

bors d+

Mathlib/Data/Matroid/Map.lean
apnelson1: I have a student working on this at the moment. It has quite a different flavour from what is currently here, since it actually has some mathematical content. I'll put it as a todo.

PR #13956, opened 2024-06-19
[Merged by Bors] - feat: bijections between morphisms in two localized categories

joelriou:
Given two localization functors `L₁ : C ⥤ D₁` and `L₂ : C ⥤ D₂` for the same class of morphisms `W : MorphismProperty C`, we define a bijection `Localization.homEquiv W L₁ L₂ : (L₁.obj X ⟶ L₁.obj Y) ≃ (L₂.obj X ⟶ L₂.obj Y)` between the types of morphisms in the two localized categories.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13876, opened 2024-06-16
[Merged by Bors] - feat: the distinguished triangle attached to a short exact sequence in an abelian category

joelriou:
Given a short exact short complex `S` in an abelian category, we construct the associated distinguished triangle in the derived category: `(singleFunctor C 0).obj S.X₁ ⟶ (singleFunctor C 0).obj S.X₂ ⟶ (singleFunctor C 0).obj S.X₃ ⟶ ...`

---

- [x] depends on: #13875
- [x] depends on: #13742
- [x] depends on: #13675
- [x] depends on: #13757

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13675, opened 2024-06-09
[Merged by Bors] - feat: the mapping cone of a monomorphism, up to a quasi-isomorphism

joelriou:
In this PR, given a short exact sequence `S` in the category of cochain complexes in an abelian category, we compare the homology sequence of `S`, and the homology sequence of the homological functor on the homotopy category applied to the distinguished triangle of the mapping cone of the monomorphism `S.f`. In particular, up to a quasi-isomorphism, the mapping cone of `S.f` identifies to `S.X₃`.

In future PRs, this shall be used in order to attach a distinguished triangle in the derived category to `S` (#13742), and compare the homology sequence of `S` and the homology sequence attached to this distinguished triangle.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13963, opened 2024-06-19
[Merged by Bors] - feat: definition of Ext-types in abelian categories

joelriou:
This PR defines Ext-types in general abelian categories by shrinking types of morphisms in the derived category. The abelian group structure shall be defined in a future PR.

(When this API is developed further, it shall replace the current definition which applies only to categories with enough projectives.)

---

- [x] depends on: #13742
- [x] depends on: #13675
- [x] depends on: #13926
- [x] depends on: #13757
- [x] depends on: #13956
- [x] depends on: #13639
- [x] depends on: #13876


Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
```diff
@@ -0,0 +1,145 @@
...
+
+@[ext]
+lemma ext {n : ℕ} {α β : Ext X Y n} (h : α.hom = β.hom) : α = β :=
+  homEquiv.injective h
+
+end
+
+-- TODO: promote `homEquiv` above as `homAddEquiv` with the same
+-- degree of generality, i.e. for any `[HasDerivedCategory.{w'} C]`
+noncomputable instance {n : ℕ} : AddCommGroup (Ext X Y n) :=
```
erdOne, line 137: I'm a bit confused. How can Lean find the right `w'` to get a `HasDerivedCategory.{w'} C` instance?
Or does Lean ignore the universe parameter and use the first `HasDerivedCategory.{*} C`? Do we need compatibility lemmas for multiple `HasDerivedCategory.{*} C` instances?

joelriou: The idea is that the definition of `Ext` groups does not depend on any `HasDerivedCategory` variable. For the definition of the abelian group structure, the best solution I have found is to define it using the constructed derived category. Then, using #14153 and some more (painful) work (wip PR #13989), I shall show that the abelian group structure is compatible with any `HasDerivedCategory` variable. At the same time, we shall also see that an exact functor induces maps on `Ext`-groups that are compatible with `Functor.mapDerivedCategory`. (By design, I believe we should not define any global instance of `HasDerivedCategory`. It is like with `Category` for which the universe is determined by the first (and hopefully unique) variable found.)

erdOne: By "constructed derived category", do you mean the "formally invert all arrows" large universe construction?
I think it would be nicer to specify which `HasDerivedCategory` is used explicitly here rather than relying on the belief that "we should not define any global instance of `HasDerivedCategory`" in order to avoid nasty diamonds, but I IIRC we don't have that construction in mathlib yet right?

joelriou: Yes, the "constructed derived category" is the large universe construction (this is the target category of `MorphismProperty.Q`), which is already used *internally* in order to define `Ext`. This is the same localized category that is provided by `HasDerivedCategory.standard`.

Providing global instances of `HasDerivedCategory` for general categories is a very bad idea, because it would mean that we would basically have to specify the universes everywhere when we use the `DerivedCategory` API...

The only other reasonable approach I see would be to define the `Add` instance on the `Ext` types using suitable binary biproducts and the basic functoriality of `Ext`. I have just removed the `AddCommGroup` instance, and we see the details in future PRs.

erdOne: Ah sorry I think I misread the original approach (I thought it took a `[HasDerivedCategory.{w} C]` instance variable) and that approach is perfectly fine IMO.

joelriou: The other approach I am suggesting just above might be slightly more convenient, I will experiment a little bit with it. The better is to keep the `AddCommGroup` instance out of this PR.

erdOne:
Thanks!
maintainer merge

semorrison:
bors merge

PR #13964, opened 2024-06-19
feat(Data/DigitExpansion): begin defining variant of reals without rationals

pechersky:
Based on a de Bruijn 1976 paper.
This file is just the basic definition of a digit expansion. Will be followed up with further constructions.

---

### DIFF
--- a/Mathlib/Data/DigitExpansion/Defs.lean
+++ b/Mathlib/Data/DigitExpansion/Defs.lean
@@ -78,7 +78,7 @@ variable {Z : Type*} [LT Z] {b : ℕ}
   [∀ (f g : DigitExpansion Z b) z, Decidable (∃ x > z, f x < g x ∧ ∀ y < x, z < y → f y ≤ g y)]
   (f g : DigitExpansion Z b)
 
-/-- Difference carry as an auxialiary function for defining subtraction. -/
+/-- Difference carry as an auxiliary function for defining subtraction. -/
 def difcar : Z → Fin (b + 1) := fun z =>
   if ∃ x > z, f x < g x ∧ ∀ y < x, z < y → f y ≤ g y then 1 else 0
 
--- a/Mathlib/Data/DigitExpansion/Defs.lean
+++ b/Mathlib/Data/DigitExpansion/Defs.lean
@@ -1,0 +1,504 @@
+/-
+Copyright (c) 2022 Yakov Pechersky. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yakov Pechersky
+-/
+import Mathlib.Order.Fin
+import Mathlib.Order.SuccPred.Basic
+import Mathlib.Tactic.Ring
+
+
+/-!
+# Defining reals without the use of rationals
+
+* [Defining reals without the use of rationals][debruijn1976]
+
+-/
+
+/-- A possibly infinite sequence of digits in `Fin (b + 1)`, which
+can represent a number system like the reals.
+It is constrained to not end in an infinite string of the top digit. -/
+structure DigitExpansion (Z : Type*) [LT Z] (b : ℕ) where
+  /-- The digit sequence, such that `b` is the base - 1. -/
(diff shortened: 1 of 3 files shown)


Mathlib/Data/DigitExpansion/Defs.lean
```diff
@@ -0,0 +1,504 @@
...
+    (z : Z) : (0 : DigitExpansion Z b) z = 0 :=
+  rfl
+
+section difcar
+
+variable {Z : Type*} [LT Z] {b : ℕ}
+  [∀ (f g : DigitExpansion Z b) z, Decidable (∃ x > z, f x < g x ∧ ∀ y < x, z < y → f y ≤ g y)]
+  (f g : DigitExpansion Z b)
+
+/-- Difference carry as an auxialiary function for defining subtraction. -/
```
semorrison, line 81:
```suggestion
/-- Difference carry as an auxiliary function for defining subtraction. -/
```

semorrison:
Two questions: 
* Can we be confident this will be finished? I wouldn't want half of the construction in Mathlib and then it to stall.
* As this is presumably a "leaf" construction, could it go in Archive or a separate repository rather than Mathlib properly? (But perhaps I am missing some consequence or application.)

pechersky:
> * Can we be confident this will be finished? I wouldn't want half of the construction in Mathlib and then it to stall.

In the PR that references this one, I have everything proved but the multiplication. I think that part ends up being straightforward after all the preliminary work I've done, since it is composition of functions.

> * As this is presumably a "leaf" construction, could it go in Archive or a separate repository rather than Mathlib properly? (But perhaps I am missing some consequence or application.)

We could put in Archive, sure. One application of this type of construction is inter-op with digit-based representations for rationals, reals, and p-adics. For example, if one wanted to construct implementations of [computable p-adics](https://arxiv.org/pdf/1701.06794) then their fidelity in representing the p-adics is often argued via their digit representations. I would then argue that computational-implementation ~ digit representation ~ Q_p -- without having to make claims about cauchy sequences, etc.

semorrison:
Okay! I'm convinced by the above that this can go in Mathlib rather than Archive. However I might then leave further review to someone with more perspective of p-adic applications. (Or just anyone. :-)

Mathlib/Data/DigitExpansion/Defs.lean
```diff
@@ -0,0 +1,504 @@
...
+-/
+
+/-- A possibly infinite sequence of digits in `Fin (b + 1)`, which
+can represent a number system like the reals.
+It is constrained to not end in an infinite string of the top digit. -/
+structure DigitExpansion (Z : Type*) [LT Z] (b : ℕ) where
+  /-- The digit sequence, such that `b` is the base - 1. -/
+  toFun : Z → Fin (b + 1)
+  /-- The constraint that the sequence does not trail end infinitely at the top digit. -/
+  bounded' : ¬∃ i₀, ∀ i > i₀, toFun i = b
```
eric-wieser, line 25: When `b = 2`, this doesn't ban an infinite series of `1` at the top digits, right?

eric-wieser: I guess I'm not clear on whether this is trying to ban `...222222121.01` or `012.12122222222...`

pechersky: 0.999999... is banned in base 10.

eric-wieser: I think then I'd argue that your digits are backwards; I'd expect digit `1` to be the tens digit, not the tenths digit (to match the order of Nat.digits)

pechersky: I was following de Bruijn's convention: 
image

pechersky: Even though he states that the sum uses `b^i`, later in the paragraph, he refers to "non-negative reals [...] a_i = 0 for all i < i_0", so that means left of the decimal point is negative.

PR #13967, opened 2024-06-19
[Merged by Bors] - chore: update `install_macos.sh` so we use official channels 

mattrobball:
The cask `elan-init` can lag well behind the latest version of `elan` and doesn't benefit from any official support.

---

### DIFF
--- a/scripts/install_macos.sh
+++ b/scripts/install_macos.sh
@@ -2,7 +2,6 @@
 
 set -exo pipefail
 
-# Install Homebrew
 if ! which brew > /dev/null; then
     # Install Homebrew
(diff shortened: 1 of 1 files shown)


scripts/install_macos.sh
```diff
@@ -2,7 +2,6 @@
 
 set -exo pipefail
 
-# Install Homebrew
 if ! which brew > /dev/null; then
```
mattrobball, line 5:
```suggestion
# Install Homebrew
if ! which brew > /dev/null; then
```

semorrison:
bors merge

PR #13971, opened 2024-06-19
[Merged by Bors] - feat: Define `AnalyticWithinAt` and `AnalyticWithinOn`

girving:
Unlike `ContinuousOn`, which allows arbitrary behavior outside of the set, `AnalyticOn` requires the function to be analytic in a neighborhood of the set. This is inconvenient for analytic manifolds with boundary, as we want analyticity for functions meaningful only up to the boundary.

Here we add `AnalyticWithinAt` and `AnalyticWithinOn` with the natural definitions. The one subtlety is that we explicitly require `ContinuousWithinAt` in `AnalyticWithinAt`, as otherwise the natural definition does not imply a nice value of the function at the point.

---

This PR is a piece of https://github.com/leanprover-community/mathlib4/pull/10853, though it also makes sense as an independent concept. There are two main annoyances:

1. `AnalyticWithinOn` is a clunky name, but unfortunately `AnalyticOn` is taken and means analytic in a neighborhood.
2. Many of the lemmas require completeness, since it is convenient to use the equivalence between `AnalyticWithinIn` and having an analytic extension, but this equivalence holds only for complete spaces. For some of the lemmas completeness could be removed by replicating more of the details of the underlying `AnalyticAt` proofs. Hopefully it is okay to not do that yet.


grunweg:
Did you meant to label this as `awaiting-review`? I presume so, hence labelling it accordingly.
awaiting-review

girving:
Oops, yep: sorry for forgetting the label. :)

girving:
@grunweg: Would you be up for a review?

urkud:
I think that we should add the definition right near the non-`within` definitions (or move the definitions to a new `Defs` file), so that we can generalize lemmas to `within` (probably, in other PRs) and deduce non-`within` version from them.

girving:
Happy to make that change, but can we agree on the types and names first? It’ll get harder to iterate once I mix things all together.

urkud:
`def`s (both names and terms) LGTM

girving:
@urkud I moved them over, and tried to adjust the docstrings reasonably. Happy to make the `Defs.lean` file if you prefer that version, but it seemed fine as is.

urkud:
I'll try to review it tomorrow, but my work week starts, so it may have to wait until next weekend (unless @sgouezel merges it first). My overall impression: the definitions are good but I would like to have another look at the proofs.

Mathlib/Analysis/Analytic/Within.lean
```diff
@@ -0,0 +1,367 @@
...
+  rcases h with ⟨r, h⟩
+  exact h.continuousWithinAt
+
+lemma AnalyticWithinAt.continuousWithinAt {f : E → F} {s : Set E} {x : E}
+    (h : AnalyticWithinAt 𝕜 f s x) : ContinuousWithinAt f s x := by
+  rcases h with ⟨p, h⟩
+  exact h.continuousWithinAt
+
+/-- `AnalyticWithinAt` is trivial if `𝓝[s] x = ⊥` -/
+lemma analyticWithinAt_of_eq_bot {f : E → F} {s : Set E} {x : E} (h : 𝓝[s] x = ⊥) :
```
sgouezel, line 78: It should be true under the weaker condition that neighborhoods of `x` within `s` are contained inside `{x}`.

sgouezel (commented):
This looks great!

Mathlib/Analysis/Analytic/Within.lean
girving: Fixed, thanks!

girving:
What's the next step for this PR?

urkud:
I'll have a look later today, and most probably merge it. I'm updating to current `master` to be sure that recent changes to the library didn't break CI.

urkud:
Thanks! :tada:
bors merge

PR #13974, opened 2024-06-19
[Merged by Bors] - feat: `Matrix.PosDef` on diagonal and numeric matrices

eric-wieser:
The main result here is `PosDef (diagonal d) ↔ ∀ i, 0 < d i`, everything else is just special or less interesting cases.

---

- [x] depends on: #13976


Mathlib/LinearAlgebra/Matrix/PosDef.lean
```diff
@@ -315,6 +333,43 @@ theorem transpose {M : Matrix n n R} (hM : M.PosDef) : Mᵀ.PosDef := by
   rw [mulVec_transpose, Matrix.dotProduct_mulVec, star_star, dotProduct_comm]
 #align matrix.pos_def.transpose Matrix.PosDef.transpose
 
+protected theorem diagonal_of_selfAdjoint [DecidableEq n] [NoZeroDivisors R] (d : n → R)
```
j-loreaux, line 336: This is misnamed and contains a `sorry`.

j-loreaux (commented):
Mostly seems fine, but I've left a few comments. I may try to golf one of these later. It's a bit sad we have to duplicate all the API from transpose to conjTranpose, but I don't really see an alternative.

Mathlib/LinearAlgebra/Matrix/PosDef.lean
eric-wieser: Whoops, I meant to unrequest your review when I split this PR, but I guess I failed to do so.

eric-wieser:
> It's a bit sad we have to duplicate all the API from transpose to conjTranpose, but I don't really see an alternative.

These are all in #13976

Mathlib/LinearAlgebra/Matrix/DotProduct.lean
```diff
@@ -101,17 +102,27 @@ section StarOrderedRing
 
 variable [PartialOrder R] [NonUnitalRing R] [StarRing R] [StarOrderedRing R] [NoZeroDivisors R]
 
+/-- Note that this applies to `ℂ` via `Complex.strictOrderedCommRing`. -/
```
j-loreaux, line 105: It looks like you copied all these docstrings from the few declarations below, but they're all wrong? I think they apply to `ℂ` via `RCLike.toStarOrderedRing`, which is available in the `ComplexOrder` scope.

Mathlib/LinearAlgebra/Matrix/PosDef.lean
```diff
@@ -48,13 +48,17 @@ def PosSemidef (M : Matrix n n R) :=
...
+      simpa only [mulVec_diagonal, ← mul_assoc] using conjugate_nonneg (h i) _⟩
+
 /-- A diagonal matrix is positive semidefinite iff its diagonal entries are nonnegative. -/
 lemma posSemidef_diagonal_iff [DecidableEq n] {d : n → R} :
     PosSemidef (diagonal d) ↔ (∀ i : n, 0 ≤ d i) := by
-  refine ⟨fun ⟨_, hP⟩ i ↦ by simpa using hP (Pi.single i 1), ?_⟩
-  refine fun hd ↦ ⟨isHermitian_diagonal_iff.2 fun i ↦ IsSelfAdjoint.of_nonneg (hd i), ?_⟩
-  refine fun x ↦ Finset.sum_nonneg fun i _ ↦ ?_
-  simpa only [mulVec_diagonal, mul_assoc] using conjugate_nonneg (hd i) _
+  refine ⟨fun ⟨_, hP⟩ i ↦ by simpa using hP (Pi.single i 1), .diagonal⟩
```
j-loreaux, line 61: I think you can remove the `by refine` here.

Mathlib/LinearAlgebra/Matrix/PosDef.lean
```diff
@@ -101,6 +105,29 @@ protected lemma one [DecidableEq n] : PosSemidef (1 : Matrix n n R) :=
...
+protected theorem intCast [DecidableEq n] (d : ℤ) (hd : 0 ≤ d) :
+    PosSemidef (d : Matrix n n R) :=
+  ⟨isHermitian_intCast _, fun x => by
+    simp only [intCast_mulVec, dotProduct_smul]
+    rw [← zsmul_eq_smul_cast]
+    refine zsmul_nonneg (dotProduct_star_self_nonneg _) hd⟩
+
+@[simp]
+protected theorem _root_.Matrix.posSemidef_intCast_iff [DecidableEq n] [Nonempty n] (d : ℤ) :
+    PosSemidef (d : Matrix n n R) ↔ 0 ≤ (d : R) :=
```
j-loreaux, line 128: Can you not make this `0 ≤ (d : ℤ)` here?

eric-wieser: `Int.cast_pos` doesn't apply in a `StarOrderedRing`; are there cases where it doesn't hold?

Mathlib/LinearAlgebra/Matrix/DotProduct.lean
eric-wieser: I think this was accurate at the time, but is no longer.

Mathlib/LinearAlgebra/Matrix/PosDef.lean
j-loreaux: No, it works (aside from the zero ring, obviously), but we don't have the `Int` stuff generalized. It all generalizes pretty trivially just like we did for `Nat` a while ago. We're also missing an instance for `StarOrderedRing`, namely `ZeroLEOneClass`.
```lean
import Mathlib

namespace Int

variable {R : Type*} [AddCommGroupWithOne R] [PartialOrder R] [CovariantClass R R (· + ·) (· ≤ ·)]
variable [ZeroLEOneClass R] [NeZero (1 : R)]

lemma cast_mono' : Monotone (Int.cast : ℤ → R) := by
  intro m n h
  rw [← sub_nonneg] at h
  lift n - m to ℕ using h with k hk
  rw [← sub_nonneg, ← cast_sub, ← hk, cast_natCast]
  exact k.cast_nonneg'

variable {m n : ℤ}

@[simp] lemma cast_nonneg' : ∀ {n : ℤ}, (0 : R) ≤ n ↔ 0 ≤ n
  | (n : ℕ) => by simp
  | -[n+1] => by
    have : -(n : R) < 1 := lt_of_le_of_lt (by simp) zero_lt_one
    simpa [(negSucc_lt_zero n).not_le, ← sub_eq_add_neg, le_neg] using this.not_le

@[simp, norm_cast] lemma cast_le' : (m : R) ≤ n ↔ m ≤ n := by
  rw [← sub_nonneg, ← cast_sub, cast_nonneg', sub_nonneg]

lemma cast_strictMono' : StrictMono (fun x : ℤ => (x : R)) :=
  strictMono_of_le_iff_le fun _ _ => cast_le'.symm

@[simp, norm_cast] lemma cast_lt' : (m : R) < n ↔ m < n := cast_strictMono'.lt_iff_lt

@[simp] lemma cast_nonpos' : (n : R) ≤ 0 ↔ n ≤ 0 := by rw [← cast_zero, cast_le']

@[simp] lemma cast_pos' : (0 : R) < n ↔ 0 < n := by rw [← cast_zero, cast_lt']

@[simp] lemma cast_lt_zero' : (n : R) < 0 ↔ n < 0 := by rw [← cast_zero, cast_lt']

end Int

variable {R : Type*} [Ring R] [StarRing R] [PartialOrder R] [StarOrderedRing R] [NeZero (1 : R)]

-- generalizes to `Semiring R`
instance : ZeroLEOneClass R where
  zero_le_one := by simpa using star_mul_self_nonneg (1 : R)

open Int
example {n : ℤ} : (0 : R) < n ↔ 0 < n := by rw [← cast_zero, cast_lt']
```

eric-wieser: Do you have time to PR the above?

j-loreaux: Maybe not for a few days. If you don't get to it before me, I'll do it.

j-loreaux: @eric-wieser see #15185 and #15186

eric-wieser: Thanks! I've simplified the lemma statements now.

j-loreaux: @eric-wieser It looks like this one is still `↔ 0 ≤ (d : R)` instead of `↔ 0 ≤ d`.

j-loreaux (commented):
other than the conversation I unresolved above with one last comment, this looks good. Thanks!

bors d+

Mathlib/LinearAlgebra/Matrix/PosDef.lean
eric-wieser:
```suggestion
    PosSemidef (d : Matrix n n R) ↔ 0 ≤ d :=
```

eric-wieser:
Thanks for filling out the StarOrderedRing API to clean it up! It's a shame how much boilerplate this needed to write some really obvious lemmas!

eric-wieser:
bors merge

PR #13975, opened 2024-06-19
[Merged by Bors] - feat: `0 < c * a * star c` in `StarOrderedRing`s

eric-wieser:
This is a prerequisite for showing when diagonal matrices are positive definite.

Missed from #13748

---

### DIFF
--- a/Mathlib/Algebra/Star/Order.lean
+++ b/Mathlib/Algebra/Star/Order.lean
@@ -246,7 +246,7 @@ theorem conjugate_lt_conjugate' {a b : R} (hab : a < b) {c : R} (hc : IsRegular 
 theorem conjugate_pos {a : R} (ha : 0 < a) {c : R} (hc : IsRegular c) : 0 < star c * a * c := by
   simpa only [mul_zero, zero_mul] using conjugate_lt_conjugate ha hc
 
-theorem conjugate_pos' {a : R} (ha : 0 < a) {c : R}  (hc : IsRegular c) : 0 < c * a * star c := by
+theorem conjugate_pos' {a : R} (ha : 0 < a) {c : R} (hc : IsRegular c) : 0 < c * a * star c := by
   simpa only [star_star] using conjugate_pos ha hc.star
 
 theorem star_mul_self_pos [Nontrivial R] {x : R} (hx : IsRegular x) : 0 < star x * x := by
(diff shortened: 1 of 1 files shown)


j-loreaux:
Oops, I reviewed the other one first, can you see my comment there relevant to this bit? (I'm on mobile so it's harder to copy)

Mathlib/Algebra/Star/Order.lean
```diff
@@ -234,11 +234,24 @@ lemma IsSelfAdjoint.mono {x y : R} (h : x ≤ y) (hx : IsSelfAdjoint x) : IsSelf
...
+    star c * a * c < star c * b * c := by
+  rw [(conjugate_le_conjugate hab.le _).lt_iff_ne, hc.right.ne_iff, hc.star.left.ne_iff]
+  exact hab.ne
+
+theorem conjugate_lt_conjugate' {a b : R} (hab : a < b) {c : R} (hc : IsRegular c) :
+    c * a * star c < c * b * star c := by
+  simpa only [star_star] using conjugate_lt_conjugate hab hc.star
+
+theorem conjugate_pos {a : R} (ha : 0 < a) {c : R} (hc : IsRegular c) : 0 < star c * a * c := by
+  simpa only [mul_zero, zero_mul] using conjugate_lt_conjugate ha hc
```
eric-wieser, line 247: > Am I right that in this generality you could get the related but different result where the hypothesis on a is nonnegativity and regularity and the conclusion is regularity?
>
> If so, do you mind adding it?

I'm not really sure how that would follow; can you give the statement you're thinking of, once you're not on mobile?

j-loreaux: The statement I was originally thinking of is trivially true, and doesn't require positivity:
```lean
import Mathlib.Algebra.Star.Order

variable {R : Type*} [NonUnitalSemiring R] [PartialOrder R] [StarRing R] [StarOrderedRing R]

example {a : R} (ha : 0 ≤ a) (ha' : IsRegular a) {c : R} (hc : IsRegular c) : IsRegular (star c * a * c) :=
  isRegular_mul_and_mul_iff.mpr ⟨(isRegular_mul_and_mul_iff.mpr ⟨hc.star, ha'⟩).1, hc⟩ |>.1
```
But I believe what I was actually wanting is something more along the lines of: if `a : R` is left-regular, then `star a * a` is regular.

However, even though it's true in C⋆-algebras (easy to see but complicated proof: represent the C⋆-algebra on a Hilbert space, notice that left-regularity corresponds to injectivity, and recall that `star a * a` and `a` have the same kernel; then `star a * a` is self-adjoint and left-regular, hence regular), I don't see how to get it in this generality. With it, one would hope to generalize my example above to `IsLeftRegular c` (and that seems hopeless too because we can't actually write `a = star b * b` because of our definition of `StarOrderedRing`), but oh well.

In any case, unless you see something I don't with the above, we'll just skip it for this PR.

j-loreaux:
bors merge

PR #13976, opened 2024-06-19
[Merged by Bors] - feat: more matrix lemmas about `37 : Matrix n n R` etc

eric-wieser:
This extends the lemmas about `zero` and `one` in this file to lemmas about `ofNat`, `natCast`, and `intCast`.

---


j-loreaux:
bors merge

PR #13980, opened 2024-06-20
[Merged by Bors] - chore: remove unnecessary @[simps]

kim-em:
This `@[simps]` is apparently unnecessary, and on a nightly generates lemmas that the linter doesn't like, so just remove it.


YaelDillies (approved):
maintainer merge

cc @101damnations just in case

fpvandoorn:
bors merge

PR #6035, opened 2023-07-20
[Merged by Bors] - feat: heterogenize TensorProduct.congr and friends

eric-wieser:
The summary is that this PR:
* generalizes a handful of results to non-commutative algebras (simply by moving them to an earlier section)
* adds heterogenous versions of `map`, `congr`, `mapBilinear`, `homTensorHom`, `leftComm`, `tensorTensorTensorComm`
* generalizes `lcurry`, `uncurry`, and `assoc` to take an extra algebra, such that the `A`s in the two different places can separately be replaced with `R`

The punchline here is really:
```lean
def tensorTensorTensorComm :
  (M ⊗[R] N) ⊗[A] (P ⊗[R] Q) ≃ₗ[A] (M ⊗[A] P) ⊗[R] (N ⊗[R] Q) :=
```
which is valuable for a base change / tensor product of bilinear forms when instantiated as
```lean
(M ⊗[R] N) ⊗[A] (Dual A M ⊗[R] Dual R N)
  ≃ₗ[A]
(M ⊗[A] Dual A M) ⊗[R] (N ⊗[R] Dual R N)
```
It would not surprise me if it's possible to introduce a third ring into `tensorTensorTensorComm`, but for now I'd prefer to assume that for that particular definition, two rings is enough for anybody!

Unfortunately, this was not the case for `lcurry`, `uncurry`, and `assoc`, which really do need to work over three rings so that the `A`s in the two different places can separately be replaced with `R`; both permutations are needed in the construction of `tensorTensorTensorComm`, and it seemed preferable to not have a plethora of primed variants.

I make no claim that the pile of `IsScalarTower` and `SMulCommClass` arguments here are in any sense optimal, besides the fact they specifically enable the use-case I have for the base change of bilinear forms.

---

- [x] depends on: https://github.com/leanprover-community/mathlib/pull/19143
- [x] depends on: #6094
- [x] depends on: #6187

[Open in Gitpod](https://gitpod.io/from-referrer/)

Extracted from https://github.com/pygae/lean-ga/pull/31. There is more I want to extract from that patch (most notably the heterobasic `BilinForm.tensorDistrib`, but the rest will come in a future PR.

PR #13987, opened 2024-06-20
[Merged by Bors] - chore(LinearAlgebra/TensorProduct/Tower): remove superfluous porting notes

joneugster:
---

The porting note was introduced in the generalisation from `def uncurry : (M →ₗ[R] N →ₗ[R] P) →ₗ[R] M ⊗[R] N →ₗ[R] P` to `def uncurry : (M →ₗ[A] N →ₗ[R] P) →ₗ[B] M ⊗[R] N →ₗ[A] P` in #6035 by @eric-wieser.

Actually I'm unsure if there is still a use for these porting notes describing the `ₓ` in `#align`. If so, they could be categorised instead...


grunweg:
I don't see why these porting notes are useful and removing them is fine. I would appreciate a second opinion... so let's
maintainer merge

Thanks for doing all these small clean-ups around porting notes!

kmill:
I could be wrong, but I believe these notes are not necessary. I'll just go ahead and merge...

bors r+

eric-wieser:
The idea was that this note might help someone using mathport to convert their project, but removing it now almost a year later is probably fine.

Generally I'd be inclined to just leave these and remove them alongside the aligns.

PR #13988, opened 2024-06-20
[Merged by Bors] - feat: Fintype instance for `Shrink`

YaelDillies:
---

### DIFF
--- a/Mathlib/Data/Fintype/Shrink.lean
+++ b/Mathlib/Data/Fintype/Shrink.lean
@@ -15,4 +15,5 @@ variable {α : Type u} [Fintype α]
 
 noncomputable instance Shrink.instFintype : Fintype (Shrink.{v} α) := .ofEquiv _ (equivShrink _)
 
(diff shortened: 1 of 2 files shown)


Mathlib/Data/Fintype/Shrink.lean
```diff
@@ -0,0 +1,18 @@
...
+/-!
+# Fintype instance for `Shrink`
+-/
+
+universe u v
+variable {α : Type u} [Fintype α]
+
+noncomputable instance Shrink.instFintype : Fintype (Shrink.{v} α) := .ofEquiv _ (equivShrink _)
+
+@[simp] lemma Fintype.card_shrink : card (Shrink.{v} α) = card α := ofEquiv_card _
```
b-mehta, line 18: This seems like a case where we should give `[Fintype (Shrink.{v} α)]` explicitly too, for the usual reasons. It's of course easy to adjust the proof.

jcommelin (commented):
Thanks :tada:

bors merge

PR #11092, opened 2024-03-01
[Merged by Bors] - fix(fun_prop):  few bug fixes for fun_prop

lecopivo:
Bug fixes
1. Fix infinite loop triggered by `(by fun_prop : ?m)`
2. Support for multiple lambda theorems. For example `ContinuousOn` has multiple versions of composition theorem.
3. Support for special constant lambda theorem. For example `IsLinearMap R fun _ => 0` should be registered as constant lambda theorem
4. Improved error messages when trying `fun_prop` on non-fun_prop goal
5. Improved error messages when missing lambda theorems

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13994, opened 2024-06-20
chore(Topology): replace continuity -> fun_prop

grunweg:
Addresses a few porting notes; I expect also a substantial speed-up of some proofs.
Change a few default values of proof fields from `continuity` to `fun_prop`; this required a few fix-ups downstream. (I checked a few instances; the failure seems related to `Continuous.subtype_mk` not always firing.)

This is almost (but not fully) exhaustive.

----------
One proof gets partially un-automated (cont. -> partial fun_prop); if desired, I can revert this change.

- [x] depends on: #11092 


Mathlib/Topology/Homotopy/Path.lean
```diff
@@ -207,15 +207,8 @@ def reparam (p : Path x₀ x₁) (f : I → I) (hf : Continuous f) (hf₀ : f 0
...
-  continuous_toFun := by
-    -- Porting note: was `continuity` in auto-param
-    refine continuous_const.path_eval ?_
-    apply Continuous.subtype_mk
-    apply Continuous.add <;> apply Continuous.mul
-    · exact continuous_induced_dom.comp (unitInterval.continuous_symm.comp continuous_fst)
-    · continuity
-    · continuity
-    · continuity
+  -- TODO: can the auto-param be converted to use `fun_prop` (instead of `continuity`)?
```
grunweg, line 210: The current auto-param uses `continuity`, which is slower... so, a good follow-up would be to make that use `fun_prop` as well. (This causes some breakage, which is why I'd like to not do so in this PR.)

Mathlib/Topology/Homotopy/HomotopyGroup.lean
```diff
@@ -447,11 +447,14 @@ def HomotopyGroup.pi0EquivZerothHomotopy : π_ 0 X x ≃ ZerothHomotopy X :=
 /-- The 1-dimensional generalized loops based at `x` are in bijection with loops at `x`. -/
 def genLoopEquivOfUnique (N) [Unique N] : Ω^ N X x ≃ Ω X x where
   toFun p :=
-    Path.mk ⟨fun t => p fun _ => t, by continuity⟩
+    Path.mk ⟨fun t => p fun _ => t, by
```
grunweg, line 450: My impression is that this version is significantly faster. That said, it would be great to reduce this to one line... perhaps mk_apply should be tagged with fun_prop?

grunweg:
CI is done (except for the linter), so let's bench this
!bench

grunweg:
!bench

MichaelStollBayreuth:
<b>General information:</b>
```text
build:                                                 -81.718 * 10⁹ (-0.0667%)
```


<b>No file got slower by at least 10⁹ instructions.</b>


<b>5 files got faster by at least 10⁹ instructions:</b>
```text
Mathlib.Topology.ContinuousFunction.StoneWeierstrass:  -31.429 * 10⁹ (-12.6%)
Mathlib.Topology.MetricSpace.Perfect:                  -21.279 * 10⁹ (-72.0%)
Mathlib.Topology.Homotopy.HomotopyGroup:               -15.318 * 10⁹ (-15.8%)
Mathlib.Topology.Homotopy.Path:                        -6.6815 * 10⁹ (-9.17%)
Mathlib.Analysis.NormedSpace.PiTensorProduct.InjectiveSeminorm:
                                                       -2.8943 * 10⁹ (-0.990%)
```


<b>3 files got faster by at least 10%:</b>
```text
Mathlib.Topology.MetricSpace.Perfect:                                 -72.0%
Mathlib.Topology.Homotopy.HomotopyGroup:                              -15.8%
Mathlib.Topology.ContinuousFunction.StoneWeierstrass:                 -12.6%
```

Mathlib/Dynamics/Flow.lean
```diff
@@ -171,9 +171,6 @@ def reverse : Flow τ α where
   map_zero' _ := by dsimp; rw [neg_zero, map_zero_apply]
 #align flow.reverse Flow.reverse
 
--- Porting note: add @continuity to Flow.toFun so that these works:
--- Porting note: Homeomorphism.continuous_toFun  : Continuous toFun  := by continuity
--- Porting note: Homeomorphism.continuous_invFun : Continuous invFun := by continuity
 @[continuity]
```
urkud, line 174: From #14040: if you add
```suggestion
@[continuity, fun_prop]
```
then `fun_prop` can prove `continuous_*` below.

grunweg: Thanks for the hint! When I tried this, I needed to call `dsimp only` also. I've pushed this so you can take a look yourself.

j-loreaux:
I would like this PR to wait for #11092. I don't think we should be making `fun_prop` a default field until some of those bugs are fixed.

grunweg:
If I find the time, I can also take out the default value changes (as the other ones are clearly good already). I'll see.

PR #13999, opened 2024-06-20
feat: a linter to flag potential confusing conventions

adomani:
Currently, the linter flags all uses of
* `a - b` where `a b : Nat`;
* `a / b` where `a b : Nat` or `a b : Int`;
* `a / 0` more or less whenever the type of `a` has a `0` and a division.

This comes up often when starting to use Lean and hopefully the flag can help mitigate initial confusions.

When the local context contains the relevant inequality/divisibility hypothesis, the linter is quiet.

A recent [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/not.20understanding.20deliberate.20error.20with.20.60ring.60.20over.20.E2.84.95/near/445879310)

---

To make sure that the linter does not produce errors, #14007 activates the linter on "all" of mathlib.  The expectation is that no error is thrown, just warnings.


Mathlib/Tactic/Linter/Papercut.lean
```diff
@@ -0,0 +1,93 @@
...
+# The papercut linter
+
+The papercut linter tries to ease new users into Lean.
+It flags certain design decisions that may be surprising or confusing.
+
+For instance,
+* `3 - 5 = 0` (`Nat`-subtraction is truncated subtraction),
+* `3 / 5 = 0` (`Nat`-division is rounded division),
+* `(3 : ℤ) / 5 = 0` (`Int.sub` is truncated subtraction),
+* `x / 0 = 0` for mot types (division by `0` equals `0`).
```
Julian, line 20:
```suggestion
* `x / 0 = 0` for most types (division by `0` equals `0`).
```

YaelDillies:
As of the Zulip discussion, I do think we should let users
1. know what they need to prove in order to ensure the inputs to the contentious function are not garbage
2. turn off the linter when they do prove it

Mathlib/Tactic/Linter/Papercut.lean
adomani: Thanks!

Mathlib/Tactic/Linter/Papercut.lean
```diff
@@ -0,0 +1,93 @@
...
+/-!
+# The papercut linter
+
+The papercut linter tries to ease new users into Lean.
+It flags certain design decisions that may be surprising or confusing.
+
+For instance,
+* `3 - 5 = 0` (`Nat`-subtraction is truncated subtraction),
+* `3 / 5 = 0` (`Nat`-division is rounded division),
+* `(3 : ℤ) / 5 = 0` (`Int.sub` is truncated subtraction),
```
eric-wieser, line 19:
```suggestion
* `(3 : ℤ) / 5 = 0` (`Int`-division is rounded division),
```

adomani: :facepalm:

adomani:
@YaelDillies, I made the warning more verbose, so now it also mentions facts that may be useful.

Mathlib/Tactic/Linter/Papercut.lean
```diff
@@ -0,0 +1,102 @@
...
+                    Division in ℤ is actually the floor of the division: e.g. `(1 : ℤ) / 2 = 0`!\n\
+                    This yields the 'expected' result only when you also prove that \
+                    '{← Meta.ppExpr b}' divides '{← Meta.ppExpr a}'"
+      else return none
+    | _ => return none
+
+/-- The "papercut" linter emits a warning on certain quirks of the basic
+definitions. -/
+register_option linter.papercut : Bool := {
+  defValue := false
```
adomani, line 65: Let me also point out that, by default, this linter is inactive!

Julian: I was going to ask about that actually. Given that this is off by default, but also seems meant to be helpful to *new* users who may not know what a linter is in the first place or that this one exists, what do you imagine is the workflow where someone ends up seeing these? They ask on Zulip and someone tells them to enable this linter?

Perhaps this is a broader topic better for elsewhere but perhaps there's room for an "easy mode / training wheel" switch which turns on this but potentially other guardrails, and new users are directed to start using Mathlib in that mode until they know they don't need it?

adomani: Yes, what I had in mind is that someone is confused once, asks on Zulip and this linter is mentioned.

It would be great if options could be set more freely, so that you could have the option on by default when you clone mathlib, but you can change the default in CI/locally if you want to.

I think that making the linter on by default would cause more noise for people who already know what they are doing, than be beneficial to new users... if you can compare these two quantities!

adomani: Actually, maybe I should run an experiment and see how many uses of `Nat`-subtraction and `Nat`-division there are in mathlib, to get a quantitative sense of what the extent of this could be.

eric-wieser:
> * `a - b` where `a b : Nat`;

ENNReal and NNReal and NNRat are also good candidates.

adomani:
Thanks, Eric!  I added warnings to `ENNReal`, `NNReal` and `NNRat` and subtraction.

If we end up adding many more of these, I may make the warning no longer tailor made to the type.

Mathlib/Tactic/Linter/Papercut.lean
```diff
@@ -0,0 +1,113 @@
...
+        let msgs := msgs.reduceOption.join
+        if let .ofTermInfo ti := info then
+          withMCtx ctx.mctx <| withLCtx ti.lctx {} do
+            if let some msg ← getPapercuts ti.expr then
+              let next := (ti.stx, msg)
+              if !(msgs.map fun d => d.1.getPos?).contains next.1.getPos? then
+                return next :: msgs
+              else return msgs
+            else return msgs
+        else return msgs
```
eric-wieser, line 108: A bit longer, but perhaps a nicer way of writing this:
```suggestion
        let msgs := msgs.reduceOption.join
        let next? ← OptionT.run do
          let .ofTermInfo ti := info | failure
          withMCtx ctx.mctx <| withLCtx ti.lctx {} do
            let msg ← OptionT.mk <| getPapercuts ti.expr
            let next := (ti.stx, msg)
            if (msgs.map fun d => d.1.getPos?).contains next.1.getPos? then
              failure
            return next
        if let .some next := next? then
          return next :: msgs
        else
          return msgs
```

eric-wieser: In fact, you might find it worth moving `getPapercuts` to `OptionT` too

adomani: Since I essentially never want the linter to produce an error, I wonder whether I should make it freely throw errors whenever something is off and then wrap the whole thing in a `try ... catch` and not worry about it.

eric-wieser: I think for debugging it's convenient to distinguish "I have nothing to do here" from "something behaved unexpectedly", especially since infotrees can be incorrectly assembled sometimes. Probably the thing to do is catch and trace on the latter errors.

adomani: Ok, so your suggestion is to do what I did, but actually conclude with `logWarning (caught error)`, instead of `return`, right?

adomani: I already know that when there are fvars in the expressions that the linter prints, this will produce errors.

adomani: If I try to `logWarning` the caught error, Lean throws this error
```lean
failed to synthesize
  MonadExcept MessageData CommandElabM
use `set_option diagnostics true` to get diagnostic information
```
on the `try`.  I have tried to lift into `TermElabM`, `CoreM`, `MetaM`, but I cannot get it to work.

Mathlib/Tactic/Linter/Papercut.lean
```diff
@@ -0,0 +1,126 @@
...
+# The papercut linter
+
+The papercut linter tries to ease new users into Lean.
+It flags certain design decisions that may be surprising or confusing.
+
+For instance,
+* `3 - 5 = 0` (`Nat`-subtraction is truncated subtraction),
+* `3 / 5 = 0` (`Nat`-division is rounded division),
+* `(3 : ℤ) / 5 = 0` (`Int`-division is rounded division),
+* `x / 0 = 0` for most types (division by `0` equals `0`).
```
YaelDillies, line 20: Sounds more like a garbage value linter than a papercut linter to me. There are other papercuts than just garbage values eg the fact that `Module` means all three of "semimodule", "module" and "vector space".

test/Papercut.lean
```diff
@@ -0,0 +1,122 @@
...
+-/
+#guard_msgs in
+set_option linter.papercut true in
+example (x y : Nat) : x / y = 0 := sorry
+
+-- if `y ∣ x` is in context, then `x / y` does not trigger the linter
+/-- warning: declaration uses 'sorry' -/
+#guard_msgs in
+set_option linter.papercut true in
+example (x y : Nat) (_ : y ∣ x): x / y = 0 := sorry
```
YaelDillies, line 87:
```suggestion
example (x y : Nat) (_ : y ∣ x) : x / y = 0 := sorry
```

test/Papercut.lean
```diff
@@ -0,0 +1,122 @@
...
+/--
+warning: declaration uses 'sorry'
+---
+warning: Division in ℕ is actually the floor of the division: e.g. `1 / 2 = 0`!
+This yields the 'expected' result only when you also prove that 'y' divides 'x'
+note: this linter can be disabled with `set_option linter.papercut false`
+-/
+#guard_msgs in
+set_option linter.papercut true in
+example (x y : Nat) : x / y = 0 := sorry
```
YaelDillies, line 81: What happens with `example (x y : Nat) : x / y = 0`? Does it tell you to prove that `y ≠ 0`?

YaelDillies:
> As of the Zulip discussion, I do think we should
> 
> 1. let users know what they need to prove in order to ensure the inputs to the contentious function are not garbage
> 2. turn off the linter when they do prove it

Thanks Damiano for implementing this! A concern I have is that the current error message doesn't explicitly say that proving the required hypothesis will silence the linter. Could it say something like "Make a proof of `x ≤ y` available in context to silence the linter"?

PR #14000, opened 2024-06-20
[Merged by Bors] - feat: The Erdős–Ginzburg–Ziv theorem

YaelDillies:
Prove the Erdős–Ginzburg–Ziv theorem as a corollary of Chevalley-Warning. This theorem states that among any (not necessarily distinct) `2 * n - 1` elements of `ZMod n`, we can find `n` elements of sum zero.

---


Mathlib/RingTheory/UniqueFactorizationDomain.lean
```diff
@@ -2150,4 +2150,13 @@ lemma _root_.induction_on_primes {P : ℕ → Prop} (h₀ : P 0) (h₁ : P 1)
   · exact fun a p _ hp ↦ h p a hp.nat_prime
 #align induction_on_primes induction_on_primes
 
+lemma prime_composite_induction {P : ℕ → Prop} (zero : P 0) (one : P 1)
```
b-mehta, line 2153: This looks like it belongs with `recOnPrimeCoprime`, and probably proved using things there.

Mathlib/Combinatorics/Additive/ErdosGinzburgZiv.lean
```diff
@@ -0,0 +1,185 @@
...
+  ∑ i : s, a i • X i ^ (p - 1)
+
+private lemma totalDegree_f₁_add_totalDegree_f₂ {a : ι → ZMod p} :
+    (f₁ s a).totalDegree + (f₂ s a).totalDegree < 2 * p - 1 := by
+  refine (add_le_add (totalDegree_finset_sum _ _) $ (totalDegree_finset_sum _ _).trans $
+    Finset.sup_mono_fun fun a _ ↦ totalDegree_smul_le _ _).trans_lt ?_
+  simp only [totalDegree_X_pow, ← two_mul]
+  refine (mul_le_mul_left' Finset.sup_const_le _).trans_lt ?_
+  rw [mul_tsub, mul_one]
+  exact tsub_lt_tsub_left_of_le ((Fact.out : p.Prime).two_le.trans $
```
b-mehta, line 47: Isn't there an instance from Fact Prime to AtLeastTwo which can help you here? If not, how about using `Prime.one_lt'` instead?

Mathlib/Combinatorics/Additive/ErdosGinzburgZiv.lean
```diff
@@ -0,0 +1,185 @@
...
+        mem_coe, ne_eq]
+      exact ⟨h𝒜disj, fun t ht _ ↦ sdiff_disjoint.mono ht₀ $ subset_biUnion_of_mem id ht⟩
+    · simp only [cons_eq_insert, mem_insert, forall_eq_or_imp, and_assoc]
+      exact ⟨ht₀.trans sdiff_subset, ht₀card, ht₀sum, h𝒜⟩
+
+/-- The **Erdős–Ginzburg–Ziv theorem** for `ℤ/nℤ`.
+
+Any sequence of at least `2 * n - 1` elements of `ZMod n` contains a subsequence of `n` elements
+whose sum is zero. -/
+theorem ZMod.erdos_ginzburg_ziv (a : ι → ZMod n) (hs : 2 * n - 1 ≤ s.card) :
```
b-mehta, line 181: Since the informal versions use the term 'sequence', let's also have a list/multiset version (of course derived from these).

YaelDillies: Does the informal "sequence" really mean the formal `Multiset`?

b-mehta: That's why I offered List instead :) In any case, my point is that there are other formulations of this statement, and we might as well provide them too.

Mathlib/Combinatorics/Additive/ErdosGinzburgZiv.lean
YaelDillies: I don't think I should be relying on obscure machinery for such simple statements. I also don't see how `(Nat.Prime;one_lt' p).out` is better than `(Fact.out : p.Prime).two_le`

Mathlib/Combinatorics/Additive/ErdosGinzburgZiv.lean
YaelDillies: Yeah but sometimes it's better to force people to use the superior statement. I've added the multiset versions but I'm unsure they are useful

kim-em (approved):
LGTM!

Mathlib/Data/Nat/Factorization/Induction.lean
```diff
@@ -74,8 +74,23 @@ def recOnMul {P : ℕ → Sort*} (h0 : P 0) (h1 : P 1) (hp : ∀ p, Prime p →
     | n + 1 => h _ _ (hp'' p n hp') (hp p hp')
   recOnPrimeCoprime h0 hp'' fun a b _ _ _ => h a b
 
-/-! ## Lemmas on multiplicative functions -/
```
b-mehta, line 77: Was this intentionally dropped?

b-mehta (approved):
bors d+

Mathlib/Data/Nat/Factorization/Induction.lean
YaelDillies: Whoops, no

YaelDillies:
bors merge

PR #14004, opened 2024-06-20
[Merged by Bors] - feat(Order/Atoms): strong atomicity typeclass

apnelson1:
A strongly atomic preorder is one in which every nontrivial interval `[a,b]` contains an element covering `a`, or equivalently an order where every closed interval is atomic. We add a new typeclass `IsStronglyAtomic` to capture this.

We provide instances of this typeclass for `SuccOrder`s, orders with `WellFoundedLT`, atomistic upper-modular lattices, complemented atomic modular lattices, and `OrdConnected` subtypes. We also show that such orders are atomic.

We also provide the dual typeclass `IsStronglyCoatomic` and dual versions of all of the above.

---


Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +356,108 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
 
+section StronglyAtomic
+
+variable {α : Type*} {a b : α} [Preorder α]
+
+/-- An order is strongly atomic if every nontrivial interval `[a,b]`
+contains an element covering `a`. -/
+@[mk_iff]
+class IsStronglyAtomic (α : Type*) [Preorder α] : Prop where
+  (exists_covBy_le_of_lt : ∀ (a b : α), a < b → ∃ x, a ⋖ x ∧ x ≤ b)
```
YaelDillies, line 367:
```suggestion
  exists_covBy_le_of_lt : ∀ (a b : α), a < b → ∃ x, a ⋖ x ∧ x ≤ b
```
or even
```suggestion
  exists_covBy_le_of_lt (a b : α) : a < b → ∃ x, a ⋖ x ∧ x ≤ b
```

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +356,108 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+
+theorem LT.lt.exists_covby_le [IsStronglyAtomic α] (h : a < b) : ∃ x, a ⋖ x ∧ x ≤ b :=
```
YaelDillies, line 369: Can you make sure this theorem is also accessible via a non-dot notation name, eg `exists_covBy_le_of_lt`?

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
       forall_congr' fun _ => isCoatomic_dual_iff_isAtomic.symm.trans Iff.rfl
 #align is_coatomic_iff_forall_is_coatomic_Ici isCoatomic_iff_forall_isCoatomic_Ici
 
+section StronglyAtomic
+
+variable {α : Type*} {a b : α} [Preorder α]
+
+/-- An order is strongly atomic if every nontrivial interval `[a,b]`
```
YaelDillies, line 362:
```suggestion
/-- An order is strongly atomic if every nontrivial interval `[a, b]`
```

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+contains an element covering `a`. -/
+@[mk_iff]
+class IsStronglyAtomic (α : Type*) [Preorder α] : Prop where
+  exists_covBy_le_of_lt' : ∀ (a b : α), a < b → ∃ x, a ⋖ x ∧ x ≤ b
+
+theorem LT.lt.exists_covby_le [IsStronglyAtomic α] (h : a < b) : ∃ x, a ⋖ x ∧ x ≤ b :=
+  IsStronglyAtomic.exists_covBy_le_of_lt' a b h
+
+theorem exists_covBy_le_of_lt [IsStronglyAtomic α] (h : a < b) : ∃ x, a ⋖ x ∧ x ≤ b :=
+  h.exists_covby_le
```
YaelDillies, line 372: Nit: The usual pattern is `lemma long_name := proof`, `alias dot_notation.name := long_name`, rather than what you have here

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+
+/-- An order is strongly coatomic if every nontrivial interval `[a,b]`
```
YaelDillies, line 374: Same here

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+contains an element covered by `b`. -/
+@[mk_iff]
+class IsStronglyCoatomic (α : Type*) [Preorder α] : Prop where
+  (exists_le_covBy_of_lt' : ∀ (a b : α), a < b → ∃ x, a ≤ x ∧ x ⋖ b)
```
YaelDillies, line 378: Same comment as for `IsStronglyAtomic`

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+
+theorem LT.lt.exists_le_covby [IsStronglyCoatomic α] (h : a < b) : ∃ x, a ≤ x ∧ x ⋖ b :=
+  IsStronglyCoatomic.exists_le_covBy_of_lt' a b h
+
+theorem exists_le_covBy_of_lt [IsStronglyCoatomic α] (h : a < b) : ∃ x, a ≤ x ∧ x ⋖ b :=
+  h.exists_le_covby
+
+@[simp] theorem isStronglyAtomic_dual_iff_is_stronglyCoatomic :
+    IsStronglyAtomic αᵒᵈ ↔ IsStronglyCoatomic α := by
```
YaelDillies, line 387: This is a bad idea as a simp lemma as it would interfere with simp's discharging of side conditions. Sorry, I'm not explaining it greatly, but I already experimented with similar simp lemmas and it was a disaster

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+  refine ⟨fun ⟨h⟩ ↦ ⟨fun a b hab ↦ ?_⟩, fun ⟨h⟩ ↦ ⟨fun a b hab ↦ ?_⟩⟩
+  · obtain ⟨x, hbx, hax⟩ := h b a hab
+    exact ⟨x, hax, hbx.toDual⟩
+  obtain ⟨x, hbx, hax⟩ := h b a hab
+  exact ⟨x, hax.toDual, hbx⟩
```
YaelDillies, line 392:
```suggestion
  simp [isStronglyAtomic_iff, isStronglyCotomic_iff, OrderDual.forall, OrderDual.exists]
```
hopefully?

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+    IsStronglyCoatomic αᵒᵈ ↔ IsStronglyAtomic α := by
+  rw [← isStronglyAtomic_dual_iff_is_stronglyCoatomic]; rfl
+
+instance [IsStronglyAtomic α] : IsStronglyCoatomic αᵒᵈ := by simpa
+
+instance [IsStronglyCoatomic α] : IsStronglyAtomic αᵒᵈ := by simpa
+
+/-- If this were an instance, it would loop in some cases. -/
+theorem IsStronglyAtomic.isAtomic (α : Type*) [PartialOrder α] [OrderBot α] [IsStronglyAtomic α] :
+    IsAtomic α where
```
YaelDillies, line 404: Do you have a concrete example where it is problematic? I'm not seeing it

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
 
+section StronglyAtomic
+
+variable {α : Type*} {a b : α} [Preorder α]
+
+/-- An order is strongly atomic if every nontrivial interval `[a,b]`
+contains an element covering `a`. -/
+@[mk_iff]
+class IsStronglyAtomic (α : Type*) [Preorder α] : Prop where
+  exists_covBy_le_of_lt' : ∀ (a b : α), a < b → ∃ x, a ⋖ x ∧ x ≤ b
```
YaelDillies, line 366: Why is this primed? It does not refer to any implementation detail
```suggestion
  exists_covBy_le_of_lt : ∀ (a b : α), a < b → ∃ x, a ⋖ x ∧ x ≤ b
```

Mathlib/Order/Atoms.lean
```diff
@@ -352,14 +355,114 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+    have hmem := (WellFounded.min_mem h (Set.Ioc a b) ⟨b, hab,rfl.le⟩)
+    exact ⟨⟨hmem.1,fun c hac hlt ↦ WellFounded.not_lt_min h
+      (Set.Ioc a b) ⟨b, hab,rfl.le⟩ ⟨hac, hlt.le.trans hmem.2⟩ hlt ⟩, hmem.2⟩
+
+theorem isStronglyCoatomic_of_wellFounded_gt (h : WellFounded ((· > ·) : α → α → Prop)) :
+    IsStronglyCoatomic α :=
+  isStronglyAtomic_dual_iff_is_stronglyCoatomic.1 <| isStronglyAtomic_of_wellFounded_lt (α := αᵒᵈ) h
+
+instance [WellFoundedLT α] : IsStronglyAtomic α :=
+    isStronglyAtomic_of_wellFounded_lt wellFounded_lt
```
YaelDillies, line 458:
```suggestion
  isStronglyAtomic_of_wellFounded_lt wellFounded_lt
```

Mathlib/Order/Atoms.lean
```diff
@@ -940,6 +1043,25 @@ protected theorem isCoatomic_iff [OrderTop α] [OrderTop β] (f : α ≃o β) :
 #align order_iso.is_coatomic_iff OrderIso.isCoatomic_iff
 
 end OrderIso
+section CompleteLattice
+
+variable [CompleteLattice α]
+
+instance [IsUpperModularLattice α] [IsAtomistic α] : IsStronglyAtomic α where
```
YaelDillies, line 1050: Ah now I get it. This to me is the one that shouldn't be an instance

YaelDillies: In general it would help a lot if you could explain why you resolved a conversation: whether you addressed it or ignored it. I am not reviewing from a computer right now so it is really hard for me to figure this information out on my own. It requires swapping tabs and uncollapsing discussions many times

Mathlib/Order/Atoms.lean
```diff
@@ -349,14 +352,110 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+
+instance IsStronglyAtomic.isAtomic (α : Type*) [PartialOrder α] [OrderBot α] [IsStronglyAtomic α] :
+    IsAtomic α where
+  eq_bot_or_exists_atom_le a := by
+    rw [or_iff_not_imp_left, ← Ne, ← bot_lt_iff_ne_bot]
+    refine fun hlt ↦ ?_
+    obtain ⟨x, hx, hxa⟩ := hlt.exists_covby_le
+    exact ⟨x, bot_covBy_iff.1 hx, hxa⟩
+
+instance IsStronglyCoatomic.isCoatomic (α : Type*) [PartialOrder α] [OrderTop α]
```
YaelDillies, line 404:
```suggestion
instance IsStronglyCoatomic.toIsCoatomic (α : Type*) [PartialOrder α] [OrderTop α]
```
same above

Mathlib/Order/Atoms.lean
```diff
@@ -349,14 +352,110 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+theorem isStronglyAtomic_dual_iff_is_stronglyCoatomic :
+    IsStronglyAtomic αᵒᵈ ↔ IsStronglyCoatomic α := by
+  simpa [isStronglyAtomic_iff, OrderDual.exists, OrderDual.forall,
+    OrderDual.toDual_le_toDual, and_comm, isStronglyCoatomic_iff] using forall_comm
+
+@[simp] theorem isStronglyCoatomic_dual_iff_is_stronglyAtomic :
+    IsStronglyCoatomic αᵒᵈ ↔ IsStronglyAtomic α := by
+  rw [← isStronglyAtomic_dual_iff_is_stronglyCoatomic]; rfl
+
+instance [IsStronglyAtomic α] : IsStronglyCoatomic αᵒᵈ := by
```
YaelDillies, line 390:
```suggestion
instance OrderDual.instIsStronglyCoatomic [IsStronglyAtomic α] : IsStronglyCoatomic αᵒᵈ := by
```
and below

Mathlib/Order/Atoms.lean
```diff
@@ -349,14 +352,110 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+    (h : Set.OrdConnected s) : IsStronglyCoatomic s :=
+  isStronglyAtomic_dual_iff_is_stronglyCoatomic.1 h.dual.isStronglyAtomic
+
+instance [IsStronglyAtomic α] {s : Set α} [Set.OrdConnected s] : IsStronglyAtomic s :=
+  Set.OrdConnected.isStronglyAtomic <| by assumption
+
+instance [IsStronglyCoatomic α] {s : Set α} [h : Set.OrdConnected s] : IsStronglyCoatomic s :=
+  Set.OrdConnected.isStronglyCoatomic <| by assumption
+
+instance [SuccOrder α] : IsStronglyAtomic α where
```
YaelDillies, line 426:
```suggestion
instance SuccOrder.toIsStronglyAtomic [SuccOrder α] : IsStronglyAtomic α where
```

Mathlib/Order/Atoms.lean
```diff
@@ -349,14 +352,110 @@ theorem isCoatomic_iff_forall_isCoatomic_Ici [OrderTop α] :
...
+      SuccOrder.succ_le_of_lt hab⟩
+
+instance [PredOrder α] : IsStronglyCoatomic α := by
+  rw [← isStronglyAtomic_dual_iff_is_stronglyCoatomic]; infer_instance
+
+end StronglyAtomic
+
 section WellFounded
 
+theorem isStronglyAtomic_of_wellFounded_lt (h : WellFounded ((· < ·) : α → α → Prop)) :
```
YaelDillies, line 438:
```suggestion
theorem IsStronglyAtomic.of_wellFounded_lt (h : WellFounded ((· < ·) : α → α → Prop)) :
```
for anonymous dot notation

Mathlib/Order/Atoms.lean
```diff
@@ -389,7 +488,7 @@ end BooleanAlgebra
 
 namespace CompleteBooleanAlgebra
 
--- See note [reducible non-instances]
+/-- See note [reducible non-instances] -/
```
YaelDillies, line 491: We keep references to notes inside regular comments usually
```suggestion
-- See note [reducible non-instances]
```

Mathlib/Order/Atoms.lean
```diff
@@ -937,6 +1036,31 @@ protected theorem isCoatomic_iff [OrderTop α] [OrderTop β] (f : α ≃o β) :
...
+    simp_rw [not_exists, and_comm (b := _ ≤ _), not_and] at hcon
+    specialize hcon (x ⊔ a) (sup_le (le_sSup _ _ hx) hab.le)
+    obtain (hbot | h_inf) := (h x hx).bot_covBy.eq_or_eq (c := x ⊓ a) (by simp) (by simp)
+    · exact False.elim <| hcon <|
+        (hbot ▸ IsUpperModularLattice.covBy_sup_of_inf_covBy) (h x hx).bot_covBy
+    rwa [inf_eq_left] at h_inf
+
+/-- A complete lower-modular lattice that is coatomistic is strongly coatomic.
+Not an instance to prevent loops. -/
+instance [IsLowerModularLattice α] [IsCoatomistic α] : IsStronglyCoatomic α := by
```
YaelDillies, line 1059:
```suggestion
theorem [IsLowerModularLattice α] [IsCoatomistic α] : IsStronglyCoatomic α := by
```

YaelDillies (commented):
Rest looks good.

Is “strongly atomic” a concept in the literature? I have never encountered it. Can you either provide a reference or clarify that this is made up for mathlib purposes

apnelson1:
> Rest looks good.
> 
> Is “strongly atomic” a concept in the literature? I have never encountered it. Can you either provide a reference or clarify that this is made up for mathlib purposes

A quick google search for 'strongly atomic lattice' confirms that the answer is yes, even if it's not so central as a concept. I don't think a citation is needed, any more than one is needed for the existing 'IsSimpleOrder', which is used in a way that differs from common usage of the term in the literature.

YaelDillies:
Well as a matter of fact I actually think we should document `IsSimpleOrder` as well, but okay for now

maintainer merge

YaelDillies:
Well as a matter of fact I actually think we should document `IsSimpleOrder` as well, but okay for now.

Btw your PR description is outdated wrt what's an instance and what's not, I think.

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #14006, opened 2024-06-20
[Merged by Bors] - fix: setOption linter also recognises a bare `profiler` option

grunweg:
Fix a missing edge case in the linter, when an option name
had only one component (as in the profiler option), and add a test.

At the same time, rewrite the linter to use the `name` API (much nicer!), and perform two small clean-ups to the test file.

---


Mathlib/Tactic/Linter/Style.lean
```diff
@@ -60,9 +60,9 @@ def setOptionLinter : Linter where run := withSetOptionIn fun stx => do
       return
     if let some (head) := stx.find? is_set_option then
       if let some (name) := parse_set_option head then
-        -- Drop a leading backtick.
-        let name := (toString name).drop 1
-        if name.startsWith "pp." || name.startsWith "profiler." || name.startsWith "trace." then
+        -- Drop a leading backtick and determine the first component of the option name.
+        let firstComponent := (((toString name).drop 1).splitOn ".").get! 0
+        if #["pp", "profiler", "trace"].contains firstComponent then
```
eric-wieser, line 65:
```suggestion
        if #[`pp, `profiler, `trace].contains name.getRoot then
```

test/LintStyle.lean
```diff
@@ -10,11 +10,11 @@ section setOption
 -- because the linter is run on `#guard_msgs` itself. This is a known issue, see e.g.
 -- https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/unreachableTactic.20linter.20not.20suppressed.20by.20.60.23guard_msgs.60
 -- We jump through an extra hoop here to silence the warning.
+set_option linter.setOption false
```
eric-wieser, line 13: Can you add a comment explaining this line?

eric-wieser: Ah nevermind, the comment was already there!

eric-wieser (approved):
bors d+

I think the string-processing that caused the bug can be skipped entirely?

Mathlib/Tactic/Linter/Style.lean
grunweg: Thanks for the pointer! Sadly, this doesn't work, as `parse_set_option` returns an `Ident`: I haven't been able to make it return a `name` instead.

eric-wieser: Isn't there an `Ident.getName`?

grunweg: Thanks for the pointer; I got it to work!

Mathlib/Tactic/Linter/Style.lean
```diff
@@ -31,11 +31,11 @@ namespace Style.SetOption
 
 /-- Whether a syntax element is a `set_option` command, tactic or term:
 Return the name of the option being set, if any. -/
-def parse_set_option : Syntax → Option (Ident)
+def parse_set_option : Syntax → Option (Name)
```
eric-wieser, line 34:
```suggestion
def parse_set_option : Syntax → Option Name
```

eric-wieser (approved):
bors d+

eric-wieser:
Ah whatever, it's already on the queue, no need to cancel it to remove those parens

Mathlib/Tactic/Linter/Style.lean
grunweg: Right; I've rolled this into #14294.

grunweg:
Thanks for the review!

PR #14009, opened 2024-06-20
chore: replace continuity -> fun_prop in remaining directories

grunweg:
After the current set of PRs, the only continuity remnants either need more attention or are merely for tagging continuity lemmas.

---

- [x] depends on: #11092


j-loreaux:
As I've mentioned elsewhere, we shouldn't be making `fun_prop` a default value until #11092 is merged. I'm adding this as a dependency.

Mathlib/LinearAlgebra/AffineSpace/ContinuousAffineEquiv.lean
```diff
@@ -126,6 +126,8 @@ variable (k P₁) in
 /-- Identity map as a `ContinuousAffineEquiv`. -/
 def refl : P₁ ≃ᵃL[k] P₁ where
   toEquiv := Equiv.refl P₁
+  continuous_toFun := by continuity
+  continuous_invFun := by continuity
```
j-loreaux, line 130: Are these because `fun_prop` doesn't work? If so, why not?

leanprover-community-bot-assistant:

grunweg:
Closing in favour of #28928 (which is almost identical).

PR #14012, opened 2024-06-21
[Merged by Bors] - feat(lint-style): fix `update-style-exceptions.py`; produce human-readable output by default

grunweg:
- Fix the output of `update-style-exceptions.py` by making `lake exe lint_style` optionally produce output in the right format: this regressed in #13620
- The current error messages are tailored for github annotations, which are not very readable for running the linter locally. Produce a human-readable and clickable error by default, but add a flag (which CI sets) for producing github-style output.

This entails adding a small CLI for the lint-style executable.


grunweg:
I'm happy to split out the "human-readable" part, but it wouldn't make this PR much shorter.

semorrison:
bors merge

PR #14013, opened 2024-06-21
[Merged by Bors] - feat(Algebra/Homology): definition of Functor.mapDerivedCategory

joelriou:
If `F` is an exact functor between abelian categories, we define the induced functor on the derived categories.

---


erdOne (commented):
Thanks!
maintainer merge

kbuzzard:
Thanks!

bors merge

PR #11570, opened 2024-03-21
feat: absolute differentials of a presheaf of rings

joelriou:
---

This is refactored in #14014.

- [x] depends on: #6845

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14014, opened 2024-06-21
[Merged by Bors] - feat: relative differentials as a presheaf of modules

joelriou:
In this PR, we define the type `M.Derivation φ` of derivations with values in a presheaf of `R`-modules `M` relative to
a morphism of `φ : S ⟶ F.op ⋙ R` of presheaves of commutative rings over categories `C` and `D` that are related by a functor `F : C ⥤ D`. In the particular case `F` is the identity functor, we construct the universal derivation.

---

- [x] depends on: #14030

This is a refactor of #11570


Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,188 @@
...
+
+open CategoryTheory
+
+variable {C : Type u₁} [Category.{v₁} C] {D : Type u₂} [Category.{v₂} D]
+
+namespace PresheafOfModules
+
+variable {S : Cᵒᵖ ⥤ CommRingCat.{u}} {F : C ⥤ D} {S' R : Dᵒᵖ ⥤ CommRingCat.{u}}
+   (M N : PresheafOfModules.{v} (R ⋙ forget₂ _ _))
+   (φ : S ⟶ F.op ⋙ R) (φ' : S' ⟶ R)
```
erdOne, line 48: I'm still not fully convinced that our main API should built around a specific type of sheaf homs. Can't we just pullback `M` along `F` when necessary?

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,188 @@
...
+/-- Given a morphism of presheaves of commutative rings `φ : S ⟶ F.op ⋙ R`,
+this is the type of relative `φ`-derivation of a presheaf of `R`-modules `M`. -/
+@[ext]
+structure Derivation where
+  /-- the underlying additive map `R.obj X →+ M.obj X` of a derivation -/
+  d {X : Dᵒᵖ} : R.obj X →+ M.obj X
+  d_mul {X : Dᵒᵖ} (a b : R.obj X) : d (a * b) = a • d b + b • d a := by aesop_cat
+  d_map {X Y : Dᵒᵖ} (f : X ⟶ Y) (x : R.obj X) :
+    d (R.map f x) = M.map f (d x) := by aesop_cat
+  d_app {X : Cᵒᵖ} (a : S.obj X) : d (φ.app X a) = 0 := by aesop_cat
```
erdOne, line 59: Can't we just use the newly introduced bundled derivation (in terms of `Module.Cat`) here?

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,188 @@
...
+
+/-- The postcomposition of a derivation by a morphism of presheaves of modules. -/
+@[simps! d_apply]
+def postcomp (f : M ⟶ N) : N.Derivation φ where
+  d := (f.app _).toAddMonoidHom.comp d.d
+  d_map _ _ := by simp [naturality_apply]
+  d_app {X} a := by
+    dsimp
+    erw [d_app, map_zero]
+
```
erdOne, line 85: We should also have `precomp`.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,188 @@
...
+/-- The universal property that a derivation `d : M.Derivation φ` must
+satisfy so that the presheaf of modules `M` can be considered as the presheaf of
+(relative) differentials of a presheaf of commutative rings `φ : S ⟶ F.op ⋙ R`. -/
+structure Universal where
```
erdOne, line 89: I think it would be useful to get an `Subsingleton` instance on `d.Universal`.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
joelriou: This is not a "specific type" of morphisms. This is *more general* than "any" `S ⟶ R`.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
joelriou: I would rather keep this for a separate PR.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
joelriou: With this degree of generality, we cannot use bundled derivations, but in the specific case of `Derivation'`, I will try to add a constructor which takes a family of bundles derivations, and a constructor for `IsUniversal` also.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
joelriou: Thanks for the suggestion. I shall add this.

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
erdOne: But if you are developing material (e.g. `relativeDifferentials'`) for abstract `S ⟶ R` below. I don't see how you can apply these back to morphisms of the form `S ⟶ F.op ⋙ R` (or $F^{-1}\mathcal{O}_S \to \mathcal{O}_R$) that allows use of this general API. In this sense, the material you are developing below is more restricted. 

I really believe that it would be cleaner if we can deal with pushforward-pullbacks separately later.

Another concern is, when you are viewing an abstract `(φ : S ⟶ R)` as a `(φ : S ⟶ (1 _).op ⋙ R)`, it is abusing the underlying defeq of `(1 _)` and `op` etc, and I'm not sure how happy lean is with this when the `R` and `S` and the category are more complex.

joelriou: Part of my point is that when we want to apply this to a morphism of schemes, then such a morphism is described as a morphism of topological spaces (which gives a certain functor `F` on categories of opens subsets) and a map `φ : Y.presheaf ⟶ base _* X.presheaf` which fits precisely in the API I am suggesting. We could define $Omega^1$ by using the adjoint map `base ^* Y.(pre)sheaf ⟶  X.(pre)sheaf`, but developing the API would be ackward because the pullback as a presheaf is not the same thing as the pullback as a sheaf.

The API you suggest is the *particular* case `Derivation'` in my PR. Using the more general API I suggest is a way to anticipate some issues when we want to sheafify.

erdOne: We wouldn't need to sheafify the sheaf of rings though, so "pullback as a presheaf is not the same thing as the pullback as a sheaf" is a non-issue as we will work with the pullback sheaf all along.

Maybe I should first ask how you plan to get `relativeDifferentials` from `relativeDifferentials'`.

joelriou: > Maybe I should first ask how you plan to get `relativeDifferentials` from `relativeDifferentials'`.

This is explained in the docstring of the file.

erdOne: Aah okay fair enough.

joelriou: Also, the equivalence of derivations for `S ⟶ F.op ⋙ R` and for `F^* S ⟶ R` is of independent interest, and it can be stated only if the more general definitions are available (or if we had used absolute derivations).

joelriou: Thanks for your understanding :-)

jcommelin:
Thanks :tada:

bors merge

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,231 @@
...
+import Mathlib.Algebra.Category.ModuleCat.Differentials.Basic
+
+/-!
+# The presheaf of differentials of a presheaf of modules
+
+In this file, we define the type `M.Derivation φ` of derivations
+with values in a presheaf of `R`-modules `M` relative to
+a morphism of `φ : S ⟶ F.op ⋙ R` of presheaves of commutative rings
+over categories `C` and `D` that are related by a functor `F : C ⥤ D`.
+We formalize the notion of universal derivation.
```
kbuzzard, line 16: Trying to read this first paragraph, I found myself not knowing what S was, and actually I'm still confused mathematically. If F : C => D then F.op >>> R is a presheaf of rings on D,and I'm guessing that S is a scheme because that's what it is in the next paragraph, and now I don't know what kind of object phi is supposed to be. Please feel free to add explanations so that I understand the first paragraph :-)

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,231 @@
...
+
+Geometrically, if `f : X ⟶ S` is a morphisms of schemes (or more generally
+a morphism of commutative ringed spaces), we would like to apply
+these definitions in the case where `F` is the pullback functors from
```
kbuzzard, line 20:
```suggestion
these definitions in the case where `F` is the pullback functor from
```

Mathlib/Algebra/Category/ModuleCat/Differentials/Presheaf.lean
```diff
@@ -0,0 +1,231 @@
...
+morphism $O_S ⟶ f_* O_X$.
+
+In order to prove that there exists a universal derivation, the target
+of which shall be called the presheaf of relative differentials of `φ`,
+we first study the case where `F` is the identity functor.
+In this case where we have a morphism of presheaves of commutative
+rings `φ' : S' ⟶ R`, we construct a derivation
+`DifferentialsConstruction.derivation'` which is universal.
+Then, the general case (TODO) shall be obtained by observing that
+derivations for `S ⟶ F.op ⋙ R` identify to derivations
```
kbuzzard, line 31:
```suggestion
derivations for `S ⟶ F.op ⋙ R` identify with derivations
```

kbuzzard (approved):
All the lean looks fine to me but I would like to be able to understand the module docstring better :-)

PR #14017, opened 2024-06-21
[Merged by Bors] - feat(CategoryTheory): if LR is abstractly isomorphic to the identity functor, then the unit is an isomorphism.

dagurtomas:
This PR proves that to show that the unit of an adjunction `L ⊣ R` is an isomorphism (i.e. that L is fully faithful), it is enough to give an arbitrary isomorphism `L ⋙ R ≅ 𝟭 C`, and the dual result. 

To do this, we give a general way to transport (co)monad structures on functors along isomorphisms of functors.

---


Mathlib/CategoryTheory/Monad/Basic.lean
```diff
@@ -396,4 +396,67 @@ instance : Inhabited (Comonad C) :=
...
+    slice_lhs 3 4 => rw [← F.map_comp]
+    simp only [hom_inv_id_app, Functor.map_id, id_comp, assoc]
+    rw [← i.hom.naturality_assoc, ← T.coassoc_assoc]
+    simp only [NatTrans.naturality_assoc]
+    congr 3
+    simp only [← Functor.map_comp, i.hom.naturality]
+
+end Comonad
+
+
```
dagurtomas, line 461:
```suggestion
```

Mathlib/CategoryTheory/Monad/Adjunction.lean
```diff
@@ -88,6 +88,42 @@ def adjToComonadIso (G : Comonad C) : G.adj.toComonad ≅ G :=
   ComonadIso.mk (NatIso.ofComponents fun X => Iso.refl _)
 #align category_theory.adjunction.adj_to_comonad_iso CategoryTheory.Adjunction.adjToComonadIso
 
+/--
+Given an adjunction `L ⊣ R`, if `L ⋙ R` is abstractly isomorphic to the identity functor, then the
+unit is an isomorphism.
+-/
+lemma isIso_unit_of_abstract_iso  (adj : L ⊣ R) (i : L ⋙ R ≅ 𝟭 C) : IsIso adj.unit := by
```
adamtopaz, line 95: Can you provide an actual iso as opposed to just an `IsIso` instance?

dagurtomas: Done.

Mathlib/CategoryTheory/Monad/Adjunction.lean
```diff
@@ -88,6 +88,62 @@ def adjToComonadIso (G : Comonad C) : G.adj.toComonad ≅ G :=
   ComonadIso.mk (NatIso.ofComponents fun X => Iso.refl _)
 #align category_theory.adjunction.adj_to_comonad_iso CategoryTheory.Adjunction.adjToComonadIso
 
+/--
+Given an adjunction `L ⊣ R`, if `L ⋙ R` is abstractly isomorphic to the identity functor, then the
+unit is an isomorphism.
+-/
+def unitAsIsoOfAbstractIso (adj : L ⊣ R) (i : L ⋙ R ≅ 𝟭 C) :  𝟭 C ≅ L ⋙ R where
```
semorrison, line 95: Let's remove `Abstract` from the name, here and below, so just `unitAsIsoOfIso`.

semorrison:
bors d+

dagurtomas:
Thanks for the reviews!

bors merge

PR #14023, opened 2024-06-21
perf(RingTheory.OreLocalization): make data irreducible 

mattrobball:
We generally do not want to unfold operations passing through `Quotient` if we can avoid it for performance reasons.

---


mattrobball:
!bench

grunweg:
Coming here from PR triage: can you comment on the status of this PR? Is this an experiment, basically not intended to land, or would you like this to get reviewed? In the latter case, please merge master. Thanks!

grunweg:
Revisiting this again from triage: can you elaborate on this PR's status, please? Thanks!

PR #14024, opened 2024-06-21
[Merged by Bors] - feat(Analysis/InnerProductSpace/Basic): add PreInnerProductSpace

yoh-tanimoto:
add the structure `PreInnerProductSpace`, which does not assume `definite` for the inner product, with a proof of the Cauchy-Schwarz inequality using the discriminant.

Motivation: Such structures arise very often in applications, and by quotienting the kernel one obtains `InnerProductSpace.Core`.

I duplicated most of the proofs from the namespace `InnerProductSpace.Core` and put them under the namespace `PreInnerProductSpace` . Is there a better way?


j-loreaux:
Well, you should remove the inner product versions for these that are strictly more general, as a first step.

@ADedecker do you have thoughts about this preinner product stuff before we get into the weeds of review?

ADedecker:
I'm sorry I have absolutely no time to help reviewing this, I'm currently buried under PHD-funding paperwork and the related stress... But I did think about this some time ago, so here are my thoughts.

I strongly think the correct approach for the general thzory is simply to assume only `SeminormedAddCommGroup E` in the definition of `InnerProductSpace` and in all the lemmas that work without definiteness : that way there is only one definition, and the choice to impose definiteness or not is obtained by choosing between `SeminormedAddCommGroup` and `NormedAddCommGroup`.

Of course this trick doesn't work for the "Core" version since it doesn't come with a norm. I am not so sure that it's the correct approach, but I would still say to just remove the definiteness from the definition, and add a `InnerProductSpace.normedAddCommGroupOfDefinite` constructor that also asks for a separate proof of definiteness. But please check it won't be too much of a hassle to use before going down that road.

Anyway, thanks for working on this!

yoh-tanimoto:
@j-loreaux @ADedecker I see, I thought this would be a small PR, but instead there seems to be much to do! I probably need some time to do anything reasonable. Meanwhile, if you have comments on how things should be organized, please do so.

j-loreaux:
I completely agree with Anatole here. I had forgotten we separated out the `NormedAddCommGroup` so that it is a parameter instead of having `InnerProductSpace` extend it. So, the sensible thing to do as a first pass on this PR is the following: Replace `NormedAddCommGroup` with `SeminormedAddCommGroup` in the definition of `InnerProductSpace`. See what breaks, then in those places that it breaks, check that definiteness is required, and in those situations add a `NormedAddCommGroup` assumption (without overlapping the `Seminormed` one). Does this make sense @yoh-tanimoto ?

yoh-tanimoto:
I tried, and the first issue is that, since I remove `definite` from `Core`, things break not with `NormedAddCommGroup` but when `definite` is used. What should I do with theorems using `definite`?  @j-loreaux

j-loreaux:
You should have two `.Core` things, one with definiteness and one with semi definiteness. On one you get a `NormedAddCommGroup` instance, and on the other you get a `SeminormedAddCommGroup` instance. It shouldn't be too big a deal because these should almost never be used anyway since you don't get good defeqs.for the norm.

yoh-tanimoto:
I made two `.Core`s. Now the problem is where I want to prove theorems for `cd : InnerProductSpace.Core 𝕜 F`, but the local notation `⟪x, y⟫` is given for `c : PreInnerProductSpace.Core 𝕜 F`. How can I update the notation?

j-loreaux:
You can just close the scope for the existing local notation (this might mean moving it into a `section`), and write a new local notation. However, what probably makes the most sense is to separate out the stuff involving `InnerProductSpace.Core` into its own `section` instead of having them intermingled.

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -113,8 +115,25 @@ class InnerProductSpace (𝕜 : Type*) (E : Type*) [RCLike 𝕜] [NormedAddCommG
...
   smul_left : ∀ x y r, inner (r • x) y = conj r * inner x y
+
+/-- An inner product space is a vector space with an additional operation called inner product.
+The norm could be derived from the inner product, instead we require the existence of a norm and
+the fact that `‖x‖^2 = re ⟪x, x⟫` to be able to put instances on `𝕂` or product
+spaces.
+
+To construct a norm from an inner product, see `InnerProductSpace.ofCore`.
+-/
+class InnerProductSpace (𝕜 : Type*) (E : Type*) [RCLike 𝕜] [NormedAddCommGroup E] extends
```
sgouezel, line 126: Can you explain why you want to have two different classes instead of just one named `InnerProductSpace` but only assuming `[RCLike 𝕜] [SeminormedAddCommGroup E]`? Then, depending on whether you want to have a norm or a seminorm, you would assume `[SeminormedAddCommGroup E]` or `[NormedAddCommGroup E]`, but use the same class `InnerProductSpace 𝕜 E`.

yoh-tanimoto: sorry, I misunderstood the discussion above. As it was suggested to add `InnerProductSpace.Core` and `PreInnerProductSpace.Core`, I thought it would apply to the class `Pre/InnerProductSpace` as well. I removerd `PreInnerProductSpace`.

The theorems are split in sections without and with `[NormedAddCommGroup E]`. I wonder whether this is the best way.

yoh-tanimoto: now for `InnerProductSpace.ofMatrix` in `Mathlib.LinearAlgebra.Matrix.PosDef`, I had to weaken the definition (it is `InnerProductSpace` only with `SeminormedAddCommGroup`, not `NormedAddCommGroup`. Is this fine?

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -360,6 +440,88 @@ theorem norm_inner_le_norm (x y : F) : ‖⟪x, y⟫‖ ≤ ‖x‖ * ‖y‖ :=
...
+theorem inner_self_ne_zero {x : F} : ⟪x, x⟫ ≠ 0 ↔ x ≠ 0 :=
+  inner_self_eq_zero.not
+#align inner_product_space.core.inner_self_ne_zero InnerProductSpace.Core.inner_self_ne_zero
+
+/-- Norm constructed from an `InnerProductSpace.Core` structure, defined to be the square root
+of the scalar product. -/
+def toNorm : Norm F where norm x := √(re ⟪x, x⟫)
+#align inner_product_space.core.to_has_norm InnerProductSpace.Core.toNorm
+
+attribute [local instance] toNorm
```
j-loreaux, line 523: Similarly, I think you don't need a new definition. You should just be able to make the `PreInnerProductSpace.Core.toNorm` a local instance.

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -708,6 +842,51 @@ theorem real_inner_mul_inner_self_le (x y : F) : ⟪x, y⟫_ℝ * ⟪x, y⟫_ℝ
...
+#align ext_inner_left ext_inner_left
+
+theorem ext_inner_right {x y : E} (h : ∀ v, ⟪x, v⟫ = ⟪y, v⟫) : x = y := by
+  rw [← sub_eq_zero, ← @inner_self_eq_zero 𝕜, inner_sub_left, sub_eq_zero, h (x - y)]
+#align ext_inner_right ext_inner_right
+
+@[simp]
+theorem inner_self_nonpos {x : E} : re ⟪x, x⟫ ≤ 0 ↔ x = 0 := by
+  rw [← norm_sq_eq_inner, (sq_nonneg _).le_iff_eq, sq_eq_zero_iff, norm_eq_zero]
+#align inner_self_nonpos inner_self_nonpos
```
j-loreaux, line 882: Unless I missed something `𝕜` is now explicit here. Can we change it back for consistency?

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -1946,8 +2195,19 @@ theorem RCLike.inner_apply (x y : 𝕜) : ⟪x, y⟫ = conj x * y :=
   rfl
 #align is_R_or_C.inner_apply RCLike.inner_apply
 
-/-! ### Inner product space structure on subspaces -/
+end RCLike
+
+section PreSubmodule
```
j-loreaux, line 2200: This `Pre` prefix in section names is just confusing. I know what you mean because of the context of this PR, but anyone else looking at this later is just going to ask: "what's a `PreSubmodule`?"

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -1977,10 +2237,19 @@ theorem orthonormal_span {ι : Type*} {v : ι → E} (hv : Orthonormal 𝕜 v) :
     Submodule.subset_span (Set.mem_range_self i)
 #align orthonormal_span orthonormal_span
 
+end PreSubmodule
+
 /-! ### Families of mutually-orthogonal subspaces of an inner product space -/
 
+section PreOrthogonalFamily
```
j-loreaux, line 2244: same comment

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -2160,6 +2429,22 @@ theorem OrthogonalFamily.summable_iff_norm_sq_summable [CompleteSpace E] (f : 
       linarith
 #align orthogonal_family.summable_iff_norm_sq_summable OrthogonalFamily.summable_iff_norm_sq_summable
 
+end PreOrthogonalFamily
```
j-loreaux, line 2432: ditto

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -2189,12 +2474,22 @@ theorem DirectSum.IsInternal.collectedBasis_orthonormal [DecidableEq ι] {V : ι
   simpa only [hV_sum.collectedBasis_coe] using hV.orthonormal_sigma_orthonormal hv_family
 #align direct_sum.is_internal.collected_basis_orthonormal DirectSum.IsInternal.collectedBasis_orthonormal
 
+
 end OrthogonalFamily
 
-section RCLikeToReal
+
+section PreRCLikeToReal
```
j-loreaux, line 2481: ditto

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -2310,6 +2623,18 @@ theorem ContinuousLinearMap.reApplyInnerSelf_apply (T : E →L[𝕜] E) (x : E)
   rfl
 #align continuous_linear_map.re_apply_inner_self_apply ContinuousLinearMap.reApplyInnerSelf_apply
 
+end ReApplyInnerSelf
+
+section PreReApplyInnerSelf
```
j-loreaux, line 2628: ditto

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -5,10 +5,11 @@ Authors: Zhouhang Zhou, Sébastien Gouëzel, Frédéric Dupuis
 -/
 import Mathlib.Algebra.DirectSum.Module
 import Mathlib.Algebra.Module.LinearMap.Basic
+import Mathlib.Algebra.QuadraticDiscriminant
```
j-loreaux, line 8: Why do we need this import now but we didn't before?

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -307,39 +331,95 @@ theorem inner_sub_sub_self (x y : F) : ⟪x - y, x - y⟫ = ⟪x, x⟫ - ⟪x, y
...
   ring
 #align inner_product_space.core.cauchy_schwarz_aux InnerProductSpace.Core.cauchy_schwarz_aux
 
 /-- **Cauchy–Schwarz inequality**.
-We need this for the `Core` structure to prove the triangle inequality below when
-showing the core is a normed group.
+We need this for the `PreInnerProductSpace.Core` structure to prove the triangle inequality below
+when showing the core is a normed group and to take the quotient.
 -/
 theorem inner_mul_inner_self_le (x y : F) : ‖⟪x, y⟫‖ * ‖⟪y, x⟫‖ ≤ re ⟪x, x⟫ * re ⟪y, y⟫ := by
```
j-loreaux, line 381: I have a feeling there's a much easier-to-formalize proof here, but I don't have time at the moment. I'll come back to it.

j-loreaux (commented):
Thanks again for taking this on. There's still room for improvement, but it's starting to look nice.

Mathlib/Analysis/InnerProductSpace/Basic.lean
j-loreaux: What you've done there is correct.

Mathlib/Analysis/InnerProductSpace/Basic.lean
yoh-tanimoto: OK, I leave this as is for now.

Mathlib/Analysis/InnerProductSpace/Basic.lean
yoh-tanimoto: I changed the proof of the Cauchy-Schwarz inequality from the previous one (using `definite`) to the current one, so I needed the traditional argument involving discriminant.

Mathlib/Analysis/InnerProductSpace/Basic.lean
yoh-tanimoto: I see, I just thought that the same section title for different assumptions isn't good. I change to `XXX_Seminormed`, is this better?

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -307,39 +331,95 @@ theorem inner_sub_sub_self (x y : F) : ⟪x - y, x - y⟫ = ⟪x, x⟫ - ⟪x, y
...
+    exact discrim_le_zero hdiscrim
+  rw [seminormSq, seminormSq, sq] at hnegdiscrim
+  nth_rw 1 [norm_inner_symm x y] at hnegdiscrim
+  linarith
+
+/-- Seminorm constructed from an `PreInnerProductSpace.Core` structure, defined to be the square root
 of the scalar product. -/
-def toNorm : Norm F where norm x := √(re ⟪x, x⟫)
-#align inner_product_space.core.to_has_norm InnerProductSpace.Core.toNorm
+def toPreNorm : Norm F where norm x := √(re ⟪x, x⟫)
```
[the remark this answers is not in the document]
yoh-tanimoto: done.

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -360,6 +440,88 @@ theorem norm_inner_le_norm (x y : F) : ‖⟪x, y⟫‖ ≤ ‖x‖ * ‖y‖ :=
...
+`InnerProductSpace.Core.toInner` because it takes `InnerProductSpace.Core` as an explicit
+argument. -/
+def toInner' : Inner 𝕜 F :=
+  cd.toInner
+
+attribute [local instance] toInner'
+
+/-- The norm squared function for `InnerProductSpace.Core` structure. -/
+def normSq (x : F) :=
+  reK ⟪x, x⟫
```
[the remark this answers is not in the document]
yoh-tanimoto: done (I restored `normSq` above and removed this).

Mathlib/Analysis/InnerProductSpace/Basic.lean
yoh-tanimoto: sorry that I missed this, yes.

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -322,20 +376,48 @@ theorem cauchy_schwarz_aux (x y : F) :
...
+          _ = ‖⟪x, y⟫‖ / ‖⟪x, y⟫‖ * ‖⟪x, y⟫‖ / ‖⟪x, y⟫‖ * re ⟪x, x⟫ * t * t +
+          ‖⟪x, y⟫‖ / ‖⟪x, y⟫‖ * 2* ‖⟪x, y⟫‖ * t + normSq y := by ring
+      rw [div_self hzero, one_mul, one_mul, div_self hzero, one_mul] at this
+      exact this
+  have hnegdiscrim : (2 * ‖⟪x, y⟫‖)^2 - 4 * normSqF x * normSqF y ≤ 0 := by
+    rw [← discrim]
+    exact discrim_le_zero hdiscrim
+  rw [normSq, normSq, sq] at hnegdiscrim
+  nth_rw 1 [norm_inner_symm x y] at hnegdiscrim
+  linarith
```
j-loreaux, line 417: I've tried to make this proof a bit more formalization friendly by using some handy automation. The key tools to make this simpler were: `field_simp`, `convert`, `ring` and `positivity`. `suffices` is also nice for restructuring proofs, and `obtain` is just handy all around.
```suggestion
  suffices discrim (normSqF x) (2 * ‖⟪x, y⟫_𝕜‖) (normSqF y) ≤ 0 by
    rw [norm_inner_symm y x]
    rw [discrim, normSq, normSq, sq] at this
    linarith
  refine discrim_le_zero fun t ↦ ?_
  by_cases hzero : ⟪x, y⟫ = 0
  · simp only [mul_assoc, ← sq, hzero, norm_zero, mul_zero, zero_mul, add_zero, ge_iff_le]
    obtain ⟨hx, hy⟩ : (0 ≤ normSqF x ∧ 0 ≤ normSqF y) := ⟨inner_self_nonneg, inner_self_nonneg⟩
    positivity
  · have hzero' : ‖⟪x, y⟫‖ ≠ 0 := norm_ne_zero_iff.2 hzero
    convert cauchy_schwarz_aux' (𝕜 := 𝕜) (⟪x, y⟫ • x) y (t / ‖⟪x, y⟫‖) using 3
    · field_simp
      rw [← sq, normSq, normSq, inner_smul_right, inner_smul_left, ← mul_assoc _ _ ⟪x, x⟫,
        mul_conj, ← ofReal_pow]
      simp [- ofReal_pow, mul_re]
      ring
    · field_simp
      rw [inner_smul_left, mul_comm _ ⟪x, y⟫_𝕜, mul_conj, ← ofReal_pow, ofReal_re]
      ring
```

j-loreaux (commented):
From my perspective, I think this is almost finished.

yoh-tanimoto:
I am sorry for my delayed commits.
@j-loreaux I just accepted your suggestions, thank you very much for the simplifications, I need to learn these tactics!

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -231,18 +259,19 @@ theorem inner_zero_left (x : F) : ⟪0, x⟫ = 0 := by
...
   rw [← inner_conj_symm, inner_zero_left]; simp only [RingHom.map_zero]
 
-theorem inner_self_eq_zero {x : F} : ⟪x, x⟫ = 0 ↔ x = 0 :=
-  ⟨c.definite _, by
-    rintro rfl
-    exact inner_zero_left _⟩
+theorem inner_self_of_eq_zero {x : F} : x = 0 → ⟪x, x⟫ = 0 := by
+    intro h
+    rw [h]
+    exact inner_zero_left _
```
j-loreaux, line 265: Almost the original proof works, and your indentation has issues:
```suggestion
  rintro rfl
  exact inner_zero_left _
```

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -231,18 +259,19 @@ theorem inner_zero_left (x : F) : ⟪0, x⟫ = 0 := by
...
 
-theorem normSq_eq_zero {x : F} : normSqF x = 0 ↔ x = 0 :=
-  Iff.trans
-    (by simp only [normSq, ext_iff, map_zero, inner_self_im, eq_self_iff_true, and_true_iff])
-    (@inner_self_eq_zero 𝕜 _ _ _ _ _ x)
+theorem normSq_eq_zero_of_eq_zero {x : F} : x = 0 → normSqF x = 0 := by
+  intro h
+  rw [normSq]
+  rw [inner_self_of_eq_zero h]
+  exact RCLike.zero_re'
```
j-loreaux, line 271: :golf:
```suggestion
  rintro rfl
  simp [normSq, inner_self_of_eq_zero]
```

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -231,18 +259,19 @@ theorem inner_zero_left (x : F) : ⟪0, x⟫ = 0 := by
...
 
-theorem inner_self_ne_zero {x : F} : ⟪x, x⟫ ≠ 0 ↔ x ≠ 0 :=
-  inner_self_eq_zero.not
+theorem ne_zero_of_inner_self_ne_zero {x : F} : ⟪x, x⟫ ≠ 0 → x ≠ 0 := by
+  exact fun P Q => P (inner_self_of_eq_zero Q)
```
j-loreaux, line 275: This is just the contrapositive of the existing result, which we can obtain with `mt` (modus tollens)
```suggestion
theorem ne_zero_of_inner_self_ne_zero {x : F} : ⟪x, x⟫ ≠ 0 → x ≠ 0 :=
  mt inner_self_of_eq_zero
```

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -274,12 +303,41 @@ theorem inner_add_add_self (x y : F) : ⟪x + y, x + y⟫ = ⟪x, x⟫ + ⟪x, y
...
+    re ⟪(ofReal t : 𝕜) • x, (ofReal t : 𝕜) • x⟫ = normSqF x * t * t := by
+  rw [inner_smul_left, inner_smul_right, conj_ofReal, RCLike.re_ofReal_mul, RCLike.re_ofReal_mul,
+  ← mul_assoc, mul_comm, ← mul_assoc, normSq]
+
+theorem inner_smul_left_re (x y : F) {t : ℝ} : ⟪(ofReal t : 𝕜) • x, y⟫ = ⟪x, y⟫  * t := by
+  rw [inner_smul_left, conj_ofReal, mul_comm]
+
+theorem inner_smul_right_re (x y : F) {t : ℝ} : ⟪x, (ofReal t : 𝕜) • y⟫ = ⟪x, y⟫  * t := by
+  rw [inner_smul_right, mul_comm]
+
```
j-loreaux, line 316: I think these lemmas are slightly misnamed, and if you prove them in a slightly different order then it's easier:
```suggestion
theorem inner_smul_ofReal_left (x y : F) {t : ℝ} : ⟪(t : 𝕜) • x, y⟫ = ⟪x, y⟫  * t := by
  rw [inner_smul_left, conj_ofReal, mul_comm]

theorem inner_smul_ofReal_right (x y : F) {t : ℝ} : ⟪x, (t : 𝕜) • y⟫ = ⟪x, y⟫  * t := by
  rw [inner_smul_right, mul_comm]

theorem re_inner_smul_ofReal_smul_self (x : F) {t : ℝ} :
    re ⟪(t : 𝕜) • x, (t : 𝕜) • x⟫ = normSqF x * t * t := by
  apply ofReal_injective (K := 𝕜)
  simp [inner_self_ofReal_re, inner_smul_ofReal_left, inner_smul_ofReal_right, normSq]

```

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -320,6 +392,75 @@ theorem norm_inner_le_norm (x y : F) : ‖⟪x, y⟫‖ ≤ ‖x‖ * ‖y‖ :=
...
+  cd.toInner
+
+attribute [local instance] toInner'
+
+local notation "normSqF" => @normSq 𝕜 F _ _ _ _
+
+theorem inner_self_eq_zero {x : F} : ⟪x, x⟫ = 0 ↔ x = 0 :=
+  ⟨cd.definite _, by
+    rintro rfl
+    exact inner_zero_left _⟩
```
j-loreaux, line 452:
```suggestion
  ⟨cd.definite _, inner_self_of_eq_zero⟩
```

j-loreaux (commented):
These are the last corrections. Ping me when you're finished and I'll merge.

yoh-tanimoto:
@j-loreaux done!

j-loreaux:
Can you please merge master, then wait for CI to finish? After that, assuming CI is successful, you're welcome to merge.

bors d+

PR #14028, opened 2024-06-22
[Merged by Bors] - feat(Bicategory/NaturalTransformation/Strong): define strong natural transformations

callesonne:
We define strong natural transformations of oplax functors, and provide API for going between strong and oplax natural transformations. We also move & rename the file `NaturalTransformation.lean`, which defines oplax natural transformations, to `NaturalTransformation/Oplax.lean`. This should make it easier to also add lax natural transformations in the future.

---


Mathlib/CategoryTheory/Bicategory/NaturalTransformation/Strong.lean
```diff
@@ -0,0 +1,202 @@
...
+that is "natural up to 2-isomorphisms".
+
+More precisely, it consists of the following:
+* a 1-morphism `η.app a : F.obj a ⟶ G.obj a` for each object `a : B`.
+* a 2-isomorphism `η.naturality f : F.map f ≫ app b ⟶ app a ≫ G.map f` for each 1-morphism
+`f : a ⟶ b`.
+* These 2-isomorphisms satisfy the naturality condition, and preserve the identities and the
+compositions modulo some adjustments of domains and codomains of 2-morphisms.
+-/
+structure StrongNatTrans (F G : OplaxFunctor B C) where
```
joelriou, line 61:
```suggestion
structure StrongOplaxNatTrans (F G : OplaxFunctor B C) where
```
The name has be changed, otherwise there will be a problem with lax/pseudo-functors.

callesonne: Very good point, thanks!

Mathlib/CategoryTheory/Bicategory/NaturalTransformation/Strong.lean
```diff
@@ -0,0 +1,202 @@
...
+def vcomp (η : StrongOplaxNatTrans F G) (θ : StrongOplaxNatTrans G H) : StrongOplaxNatTrans F H :=
+  mkOfOplax (OplaxNatTrans.vcomp η.toOplax θ.toOplax)
+    { naturality := λ {a b} f ↦
+        (α_ _ _ _).symm ≪≫ whiskerRightIso (η.naturality f) (θ.app b) ≪≫
+        (α_ _ _ _) ≪≫ whiskerLeftIso (η.app a) (θ.naturality f) ≪≫ (α_ _ _ _).symm }
+
+variable (B C)
+
+@[simps id comp]
+instance : CategoryStruct (Pseudofunctor B C) where
```
joelriou, line 193: As you are generating `simps` lemma, it may be appropriate to give an explicit name to this instance (and put it in the `Pseudofunctor` namespace), so that it will look better when we need to invoke these lemmas explicitly.

joelriou:
Thanks!

bors d+

Mathlib/CategoryTheory/Bicategory/NaturalTransformation/Strong.lean
callesonne: Thanks for the suggestion, done!

PR #12273, opened 2024-04-19
[Merged by Bors] - feat(Order/KonigLemma): Kőnig's infinity lemma

apnelson1:
This PR proves Kőnig's infinity lemma in terms of covers in a strongly atomic order, and it gives a common formulation for directed systems, which is useful for applications such as Ramsey theory. 

One of the things we needed is a `Fintype` instance for `RelEmbedding r s`. This would have caused a bad diamond with 
`SimpleGraph.Hom.instFintype`, so instead we remove the latter and provide their common generalization `RelHom.instFintype` in `Fintype.Pi`.

---

- [x] depends on: #14004

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14034, opened 2024-06-22
[Merged by Bors] - feat(Order/WellFounded): well-founded of strictly monotone

apnelson1:
This PR is separated off from #12273, which no longer depends on it. 

We show that an order with a strictly monotone function into an order with WellFoundedLT is itself WellFoundedLT. 
We also add three dual variants, and provide an instance showing that a Nat-graded order is thus well-founded. 

---


JADekker:
Mathematically this is not my expertise, but in terms of style I think this looks good to me!

Mathlib/Order/WellFounded.lean
```diff
@@ -286,3 +286,27 @@ theorem WellFounded.induction_bot {α} {r : α → α → Prop} (hwf : WellFound
 #align well_founded.induction_bot WellFounded.induction_bot
 
 end Induction
+
+section WellFoundedLT
+
+variable [Preorder α] [Preorder β] {f : α → β}
+
+theorem WellFoundedLT.of_strictMono [WellFoundedLT β] (hf : StrictMono f) : WellFoundedLT α where
```
j-loreaux, line 294: I would actually name this different to unlock dot notation. Same for the others in this file.
```suggestion
theorem StrictMono.wellFoundedLT [WellFoundedLT β] (hf : StrictMono f) : WellFoundedLT α where
```

j-loreaux:
Looks fine other than the comment above.

bors d+

PR #14042, opened 2024-06-23
chore: adaptations for nightly-2024-06-22

kim-em:


Mathlib/Data/Prod/Basic.lean
```diff
@@ -63,20 +61,26 @@ theorem map_mk (f : α → γ) (g : β → δ) (a : α) (b : β) : map f g (a, b
   rfl
 #align prod.map_mk Prod.map_mk
 
-theorem map_fst (f : α → γ) (g : β → δ) (p : α × β) : (map f g p).1 = f p.1 :=
+-- I'm skeptical about having this as a `simp` lemma, despite it having been in the past
```
fpvandoorn, line 64: Do we want this comment in the source code? If so, maybe add your name?

fpvandoorn (approved):
LGTM

bors d+

Mathlib/Data/Prod/Basic.lean
semorrison: I've rephrased without "I".

PR #14043, opened 2024-06-23
[Merged by Bors] - feat(Data/ENat): `sSup_eq_zero` and `sInf_eq_zero`

Rida-Hamadani:
To make working with `ENat` easier.

---

### DIFF
--- a/Mathlib/Data/ENat/Lattice.lean
+++ b/Mathlib/Data/ENat/Lattice.lean
@@ -51,8 +51,8 @@ lemma sSup_eq_zero : sSup s = 0 ↔ ∀ a ∈ s, a = 0 :=
   sSup_eq_bot
 
 lemma sInf_eq_zero : sInf s = 0 ↔ 0 ∈ s :=
-  ⟨fun h ↦ have ⟨_, h₁, h₂⟩  := (sInf_eq_bot.mp h) 1 (by decide)
-  lt_one_iff_eq_zero.mp h₂ ▸ h₁, inf_eq_bot_of_bot_mem⟩
+  rw [←lt_one_iff_eq_zero]
+  simp only [sInf_lt_iff, lt_one_iff_eq_zero, exists_eq_right]
 
 lemma sSup_eq_zero' : sSup s = 0 ↔ s = ∅ ∨ s = {0} :=
   sSup_eq_bot'
--- a/Mathlib/Data/ENat/Lattice.lean
+++ b/Mathlib/Data/ENat/Lattice.lean
@@ -50,8 +50,8 @@ variable {s : Set ℕ∞}
 lemma sSup_eq_zero : sSup s = 0 ↔ ∀ a ∈ s, a = 0 :=
   sSup_eq_bot
 
-lemma sInf_eq_zero : sInf s = 0 ↔ 0 ∈ s :=
-  rw [←lt_one_iff_eq_zero]
+lemma sInf_eq_zero : sInf s = 0 ↔ 0 ∈ s := by
+  rw [← lt_one_iff_eq_zero]
   simp only [sInf_lt_iff, lt_one_iff_eq_zero, exists_eq_right]
 
 lemma sSup_eq_zero' : sSup s = 0 ↔ s = ∅ ∨ s = {0} :=
--- a/Mathlib/Data/ENat/Lattice.lean
+++ b/Mathlib/Data/ENat/Lattice.lean
@@ -5,7 +5,6 @@ Authors: Yury Kudryashov
 -/
 import Mathlib.Data.Nat.Lattice
 import Mathlib.Data.ENat.Basic
-import Mathlib.Order.CompletePartialOrder
 
 #align_import data.enat.lattice from "leanprover-community/mathlib"@"f7fc89d5d5ff1db2d1242c7bb0e9062ce47ef47c"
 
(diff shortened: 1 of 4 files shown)


Mathlib/Data/ENat/Lattice.lean
```diff
@@ -45,4 +45,27 @@ lemma coe_iSup : BddAbove (range f) → ↑(⨆ i, f i) = ⨆ i, (f i : ℕ∞)
...
+  refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
+  · rw [← bot_eq_zero, sSup_eq_bot, bot_eq_zero] at h
+    by_cases h' : s = ∅
+    · apply Or.inr h'
+    · apply Or.inl
+      ext x
+      refine ⟨h x, fun hx ↦ ?_⟩
+      subst hx
+      obtain ⟨u, hu⟩ := nonempty_def.mp <| nonempty_iff_ne_empty.mpr h'
+      exact h u hu ▸ hu
```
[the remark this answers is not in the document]
Ruben-VandeVelde: But then I found some even better ones:
```lean
-- Put in Mathlib.Order.CompletePartialOrder
lemma sSup_eq_bot' [CompleteLattice α] {s : Set α} : sSup s = ⊥ ↔ s = ∅ ∨ s = {⊥} := by
  rw [sSup_eq_bot, ← subset_singleton_iff_eq, subset_singleton_iff]

lemma ENat.sSup_eq_zero : sSup s = 0 ↔ s = ∅ ∨ s = {0} := by
  rw [← bot_eq_zero, sSup_eq_bot']
```
Note that I swapped the order on the RHS to match `subset_singleton_iff_eq`

Ruben-VandeVelde (commented):
Thanks!

Mathlib/Data/ENat/Lattice.lean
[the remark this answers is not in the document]
Rida-Hamadani: Awesome, thank you for the great review!

Mathlib/Order/CompletePartialOrder.lean
```diff
@@ -82,3 +82,6 @@ end CompletePartialOrder
 instance CompleteLattice.toCompletePartialOrder [CompleteLattice α] : CompletePartialOrder α where
   sSup := sSup
   lubOfDirected _ _ := isLUB_sSup _
+
+lemma sSup_eq_bot' [CompleteLattice α] {s : Set α} : sSup s = ⊥ ↔ s = ∅ ∨ s = {⊥} := by
```
b-mehta, line 86: It seems like this lemma belongs in the same file as `sSup_eq_bot`. What do you think?

Mathlib/Data/ENat/Lattice.lean
```diff
@@ -5,6 +5,7 @@ Authors: Yury Kudryashov
 -/
 import Mathlib.Data.Nat.Lattice
 import Mathlib.Data.ENat.Basic
+import Mathlib.Order.CompletePartialOrder
```
b-mehta, line 8: With my other suggestion, I think you don't need to change the imports here...?

Mathlib/Data/ENat/Lattice.lean
```diff
@@ -45,4 +46,13 @@ lemma coe_iSup : BddAbove (range f) → ↑(⨆ i, f i) = ⨆ i, (f i : ℕ∞)
 @[norm_cast] lemma coe_iInf [Nonempty ι] : ↑(⨅ i, f i) = ⨅ i, (f i : ℕ∞) :=
   WithTop.coe_iInf (OrderBot.bddBelow _)
 
+variable {s : Set ℕ∞}
+
+lemma sSup_eq_zero : sSup s = 0 ↔ s = ∅ ∨ s = {0} :=
+  sSup_eq_bot'
```
b-mehta, line 52: How about having `sSup_eq_zero` mimicking `sSup_eq_bot`, and `sSup_eq_zero'` mimicking `sSup_eq_bot'`?

Mathlib/Order/CompletePartialOrder.lean
Rida-Hamadani: I agree, this seems more correct.

Mathlib/Data/ENat/Lattice.lean
```diff
@@ -45,4 +45,16 @@ lemma coe_iSup : BddAbove (range f) → ↑(⨆ i, f i) = ⨆ i, (f i : ℕ∞)
...
   WithTop.coe_iInf (OrderBot.bddBelow _)
 
+variable {s : Set ℕ∞}
+
+lemma sSup_eq_zero : sSup s = 0 ↔ ∀ a ∈ s, a = 0 :=
+  sSup_eq_bot
+
+lemma sInf_eq_zero : sInf s = 0 ↔ 0 ∈ s :=
+  ⟨fun h ↦ have ⟨_, h₁, h₂⟩  := (sInf_eq_bot.mp h) 1 (by decide)
+  lt_one_iff_eq_zero.mp h₂ ▸ h₁, inf_eq_bot_of_bot_mem⟩
```
b-mehta, line 55:
```suggestion
  rw [←lt_one_iff_eq_zero]
  simp only [sInf_lt_iff, lt_one_iff_eq_zero, exists_eq_right]
```
imo this is cleaner, but up to you

b-mehta (approved):
bors d+

PR #14044, opened 2024-06-23
[Merged by Bors] - feat(SetTheory/Surreal/Basic): surreal number multiplication

hwatheod:
prove multiplication of surreal numbers is well-defined, thus they form an ordered commutative ring

This is a port of a Lean 3 proof from [this mathlib branch](https://github.com/leanprover-community/mathlib/blob/surreal_mul_symm'/src/set_theory/surreal/basic.lean). It shows that multiplication of surreal numbers is well-defined, and therefore the surreal numbers form an ordered commutative ring.

Because `Surreal` is now a ring, the simp normal-form for scalar multiplication (by integers) is now `*`. This required changes throughout `SetTheory/Surreal/Dyadic`.

For the most part, I translated the Lean 3 code directly to Lean 4, although in a few cases, I couldn't get that to work and wrote a new proof instead.

There were some lemmas used that were in the mathlib branch that were not ported to mathlib4. I have ported those lemmas and put them in an appropriate place in mathlib4.

I copied over the author names from the Lean 3 code and added myself.

---


Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -1,9 +1,10 @@
 /-
 Copyright (c) 2019 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Mario Carneiro, Scott Morrison
+Authors: Mario Carneiro, Scott Morrison, Violeta Hernández Palacios, Junyan Xu, Theodore Hwa
 -/
 import Mathlib.Algebra.Order.Hom.Monoid
+import Mathlib.Logic.Hydra
```
semorrison, line 7: I'd suggest deferring any material that needs this import to a separate file.

semorrison: e.g. `Surreal.Multiplication`.

hwatheod: The `Hydra` material is needed at the early parts of the multiplication proof so I don't see a way to do that without moving the whole multiplication proof to a separate file. Maybe that is a good idea anyway since the multiplication proof is quite complicated on its own. I will do that.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -439,6 +427,643 @@ lemma bddBelow_of_small (s : Set Surreal.{u}) [Small.{u} s] : BddBelow s := by
...
+and we utilize symmetry (permutation and negation of arguments) to minimize calculation.
+
+  strategy: extract specialized versions of the induction hypothesis that easier to apply
+  (example: `ih1`, ...),
+  show they are invariant under certain symmetries (permutation and negation of variables),
+  and that the induction hypothesis indeed implies the specialized versions.
+  (Actually the induction hypotheses themselves already have those symmetries ..? No, not usually
+  the negation symmetries).
+
+  add the numeric hypothesis only at the last moment ..
```
alreadydone: This part is in draft form and that was the main reason why I didn't open a PR. Not sure how much I could recall about what I intended to say ...

alreadydone (commented):
Thanks for porting and opening the PR! Here are some initial comments, and I'll probably do some cleanups myself later. (There's a algebraic geometry formalization workshop this week, so my time will be limited ...)

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -439,6 +427,643 @@ lemma bddBelow_of_small (s : Set Surreal.{u}) [Small.{u} s] : BddBelow s := by
...
+  all_goals (
+    apply Numeric.sub
+    apply Numeric.add
+    apply ihnx ih
+    pick_goal 2
+    apply ihny ih
+    pick_goal 3
+    apply ihnxy ih
+  )
+  all_goals { solve_by_elim [IsOption.mk_left, IsOption.mk_right] }
```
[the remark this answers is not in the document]
hwatheod: Removing the parentheses works.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -439,6 +427,643 @@ lemma bddBelow_of_small (s : Set Surreal.{u}) [Small.{u} s] : BddBelow s := by
...
+
+
```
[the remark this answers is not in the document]
hwatheod: I ported manually.  I might have removed the `omit` and `include` statements without deleting the line.  Anyway I went through and removed all the extraneous blank lines.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -439,6 +427,643 @@ lemma bddBelow_of_small (s : Set Surreal.{u}) [Small.{u} s] : BddBelow s := by
...
+  · exact (cutExpand_add_left {x₁}).2 (cutExpand_pair_left h)
+  · exact (cutExpand_add_left {x₁}).2 (cutExpand_pair_right h)
+
+
+lemma ih₂₁ : ih24 x₂ x₁ y := ih₁₂ (by
+  simp_rw [ArgsRel, InvImage, Args.toMultiset, Multiset.pair_comm] at ih' ⊢
+  suffices {x₁, y, x₂} = {x₂, y, x₁} by rwa [← this]
+  dsimp [← Multiset.singleton_add] at ih' ⊢
+  abel
+)
```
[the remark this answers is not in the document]
hwatheod: I cleaned up some things in the next few lemmas following your other suggestions, but leaving this open in case you think there is more cleanup needed.

Mathlib/SetTheory/Surreal/Multiplication.lean
```diff
@@ -210,24 +210,24 @@ lemma ih1_neg_right : ih1 x y → ih1 x (-y) :=
...
+lemma ih1_of_ih : ih1 x y := by
   rintro x₁ x₂ y' h₁ h₂ (rfl|hy) <;> apply ih (Args.P24 _ _ _)
   swap
   refine TransGen.tail ?_ (cutExpand_pair_right hy)
   all_goals { exact TransGen.single (cutExpand_double_left h₁ h₂) }
 
-lemma ih1yx : ih1 y x :=
-  ih1xy $ by
+lemma ih1_rev_of_ih : ih1 y x :=
+  ih1_of_ih $ by
```
alreadydone: I renamed `ih1` to `IH1` according to naming convention, and then I can rename `ih1xy` to `ih1`, and `ih1yx` to `ih1_swap`. (I think `swap` is more common than `rev`, which is applied to lists/strings where there may be more than one element.)

Mathlib/SetTheory/Surreal/Multiplication.lean
```diff
@@ -210,24 +210,24 @@ lemma ih1_neg_right : ih1 x y → ih1 x (-y) :=
...
-lemma ihnx (h : IsOption x' x) : (x' * y).Numeric :=
+lemma left_numeric_of_ih (h : IsOption x' x) : (x' * y).Numeric :=
   ih (Args.P1 x' y) $ TransGen.single $ cutExpand_pair_left h
 
-lemma ihny (h : IsOption y' y) : (x * y').Numeric :=
+lemma right_numeric_of_ih (h : IsOption y' y) : (x * y').Numeric :=
   ih (Args.P1 x y') $ TransGen.single $ cutExpand_pair_right h
 
-lemma ihnxy (hx : IsOption x' x) (hy : IsOption y' y) : (x' * y').Numeric :=
+lemma left_right_numeric_of_ih (hx : IsOption x' x) (hy : IsOption y' y) : (x' * y').Numeric :=
```
alreadydone: I renamed these to `numeric_(option_)mul(_option)`.

alreadydone (commented):
Thanks for your work! I've pushed some more cleanups. I think the only thing remaining is to clean up the draft docs, which I should be able to do by the end of tomorrow.

alreadydone (approved):
Okay I think I've done cleaning up the PR. Maybe @MichaelStollBayreuth (coauthor of the original proof) would be interested in taking a look, and then I think we can merge this!

alreadydone:
Also, I haven't seen @vihdzp around for a while but I think this is worth a notification. also cc @apurvanakade

MichaelStollBayreuth:
I'll try to have a look soon-ish, but won't have a lot of time this week because of the AIM workshop (plus teaching etc. as usual).

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -30,27 +30,22 @@ Surreal numbers inherit the relations `≤` and `<` from games (`Surreal.instLE`
...
 
-One can also map all the ordinals into the surreals!
-
-### Multiplication of surreal numbers
-
-The proof that multiplication lifts to surreal numbers is surprisingly difficult and is currently
-missing in the library. A sample proof can be found in Theorem 3.8 in the second reference below.
-The difficulty lies in the length of the proof and the number of theorems that need to proven
-simultaneously. This will make for a fun and challenging project.
+In Mathlib.SetTheory.Surreal.Multiplication, we define multiplication and show that the
```
semorrison, line 35:
```suggestion
In `Mathlib.SetTheory.Surreal.Multiplication`, we define multiplication and show that the
```

Mathlib/SetTheory/Surreal/Multiplication.lean
```diff
@@ -0,0 +1,530 @@
+/-
+Copyright (c) 2024 Mario Carneiro. All rights reserved.
```
semorrison, line 2: This should be Theodore Hwa, now?

semorrison:
Two minor comments, but otherwise I'm happy.

bors d+

PR #14054, opened 2024-06-23
[Merged by Bors] - feat(RingTheory): localization and finiteness

chrisflav:
We show that `Module.Finite` is preserved under localizations and that if a module is finite after localizing at a spanning set of elements of the ring, it is finite.

---

The `LocalizedModuleIntegers` file is mostly a copy of [IsLocalization.IsInteger](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Localization/Integer.html#IsLocalization.IsInteger).

### DIFF
--- a/Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
+++ b/Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
@@ -31,14 +31,10 @@ open Function
 
 namespace IsLocalizedModule
 
-section
-
 /-- Given `x : M'`, `M'` a localization of `M` via `f`, `IsInteger f x` iff `x` is in the image of
 the localization map `f`. -/
 def IsInteger (x : M') : Prop :=
   x ∈ LinearMap.range f
-
-end
 
 lemma isInteger_zero : IsInteger f (0 : M') :=
   Submodule.zero_mem _
--- a/Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
+++ b/Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
@@ -1,0 +1,144 @@
+/-
+Copyright (c) 2024 Christian Merten. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Merten
+-/
+import Mathlib.Algebra.Module.LocalizedModule
+import Mathlib.Algebra.Module.Submodule.Pointwise
+
+/-!
+
+# Integer elements of a localized module
+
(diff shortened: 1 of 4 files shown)


Mathlib/RingTheory/Localization/Finiteness.lean
```diff
@@ -0,0 +1,107 @@
...
+variable {R : Type u} [CommRing R] (S : Submonoid R) {M : Type w} [AddCommMonoid M] [Module R M]
+
+/--
+If there exists a finite set `{ r }` of `R` such that `Mᵣ` is `Rᵣ`-finite for each `r`,
+then `M` is a finite `R`-module.
+
+General version for any modules `Mᵣ` and rings `Rᵣ` satisfying the correct universal properties.
+See `Module.Finite.of_localizationSpan_finite` for the specialized version.
+-/
+theorem of_localizationSpan_finite' (t : Finset R) (ht : Ideal.span (t : Set R) = ⊤)
```
erdOne, line 71: Can you give a `Set R` version? (using `Ideal.span_eq_top_iff_finite`)

chrisflav: Done.

erdOne:
Aside from the code duplication which makes me a bit sad, LGTM.
Thanks!
maintainer merge

Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
```diff
@@ -0,0 +1,144 @@
...
+
+
+variable {R : Type*} [CommSemiring R] {S : Submonoid R} {M : Type*} [AddCommMonoid M]
+  [Module R M] {M' : Type*} [AddCommMonoid M'] [Module R M'] (f : M →ₗ[R] M')
+
+open Function
+
+namespace IsLocalizedModule
+
+section
```
kbuzzard, line 34: Does this (and the `end` a few lines later) do anything?

kbuzzard (approved):
I'm not mad keen on `IsInteger` as a name, but I cannot think of a better name for the concept and so I guess it can be refactored later if I think of one. I don't really understand one of the `section`s so I've left a comment about it; feel free to delete the section if this is an oversight on your part.

bors d+

Mathlib/Algebra/Module/LocalizedModuleIntegers.lean
chrisflav: No, I removed it.

chrisflav:
Thanks for the reviews!

bors r+

PR #14056, opened 2024-06-23
[Merged by Bors] - chore: Fix unicode issues in references.bib

dupuisf:
This PR fixes two Unicode issues in references.bib, as per the discussion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Contribution.20to.20doc-gen4.3F/near/446313567), with a view towards running a [bibtex parser](https://github.com/dupuisf/BibtexQuery) written in Lean on it:

- I removed one instance of an `ö` character in a citation key, which is anyway problematic for some implementations of bibtex (see e.g. [this thread](https://tex.stackexchange.com/questions/359215/utf8-character-in-citation-keys)). According to grep, this was only referenced in one file in mathlib, which I changed accordingly.
- The one other problematic entry for my bibtex parser is one where there is an `\"o` inside a string delimited using quotation marks. I have an upcoming fix for this in my parser, but using UTF-8 here is anyway cleaner.

---

### DIFF
--- a/Mathlib/Analysis/Analytic/Inverse.lean
+++ b/Mathlib/Analysis/Analytic/Inverse.lean
@@ -353,7 +353,7 @@ analytic function. Another option would be to compute explicitly its terms (with
 coefficients) to obtain an explicit geometric bound, but this would be very painful.
 
 Instead, we will use the above intuition, but in a slightly different form, with finite sums and an
-induction. I learnt this trick in [pöschel2017siegelsternberg]. Let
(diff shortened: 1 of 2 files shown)


Ruben-VandeVelde:
I wonder if we can lint for these issues

Ruben-VandeVelde (approved):
But regardless, let's

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14057, opened 2024-06-23
[Merged by Bors] - chore: move `lint_style` executable to the `scripts` directory

grunweg:
Otherwise, importing `Mathlib.Tactic` makes defining own executables impossible, as the `main` function in `Linter.TextBased` would collide with any other main function. Move the code for the lint_style executable to scripts (which arguably might be a better place anyway); we leave the text-based linters in Tactic/Linter.

---

### DIFF
--- a/scripts/lint_style.lean
+++ b/scripts/lint_style.lean
@@ -1,0 +1,45 @@
+/-
+Copyright (c) 2024 Michael Rothgang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Michael Rothgang
+-/
+
+import Mathlib.Tactic.Linter.TextBased
+import Cli.Basic
+
+/-!
+# Text-based style linters
+
+This files defines the `lint_style` executable which runs all text-based style linters.
+The linters themselves are defined in `Mathlib.Tactic.Linter.TextBased`.
+-/
+
+open Cli
+
+/-- Implementation of the `lint_style` command line program. -/
+def lintStyleCli (args : Cli.Parsed) : IO UInt32 := do
+  let errorStyle := match (args.hasFlag "github", args.hasFlag "update") with
+    | (true, _) => ErrorFormat.github
+    | (false, true) => ErrorFormat.exceptionsFile
+    | (false, false) => ErrorFormat.humanReadable
+  let mut numberErrorFiles : UInt32 := 0
+  for s in ["Archive.lean", "Counterexamples.lean", "Mathlib.lean"] do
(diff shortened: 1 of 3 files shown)


eric-wieser:
maintainer merge

LGTM, but let's get a second pair of eyes.

scripts/lint_style.lean
```diff
@@ -0,0 +1,45 @@
...
+def lintStyleCli (args : Cli.Parsed) : IO UInt32 := do
+  let errorStyle := match (args.hasFlag "github", args.hasFlag "update") with
+    | (true, _) => ErrorFormat.github
+    | (false, true) => ErrorFormat.exceptionsFile
+    | (false, false) => ErrorFormat.humanReadable
+  let mut numberErrorFiles : UInt32 := 0
+  for s in ["Archive.lean", "Counterexamples.lean", "Mathlib.lean"] do
+    let n ← lintAllFiles (System.mkFilePath [s]) errorStyle
+    numberErrorFiles := numberErrorFiles + n
+  return numberErrorFiles
```
adomani, line 29: Does this end up being the exit code of the function?  Should the value make sure that it is not a non-zero multiple of 256, to avoid returning 0 when in reality there is an error?

semorrison:
bors d+

Please fix the exit code issue, but otherwise LGTM!

scripts/lint_style.lean
grunweg: Good catch, in principle!
In practice, I don't think this issue is likely: `UInt8` overflows at 256, but `UInt32` like this and the other functions return overflows at 2^32. I think at this point, we'll have other problems... so I'll approve this for now. I you think this requires fixing, let me know and I'll change it in a follow-up!

grunweg:
Thanks for the review! Per my comment above, let me
bors r+

scripts/lint_style.lean
adomani: I am quite outside of my comfort zone here, but isn't this number reported back to the shell?  I think that there, reduction mod 256 still happens:
```bash
$ ( exit 255 ); echo $?
255

$ ( exit 256 ); echo $?
0
```

grunweg: Oh, that is a good point. (I wasn't aware of this detail about shell exit codes... TIL!) A quick search confirms that limiting to an exit code of at most 125 is the safest option.
https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux

Let me send a follow-up PR soon, and adapt mk_all at the same time.

PR #14058, opened 2024-06-23
[Merged by Bors] -  feat(lint-style): rewrite the linter for plain string adaptation notes in Lean

grunweg:
In true fashion, the linter (correctly) flags itself; we add these to the nolints file.

---


semorrison:
bors merge

PR #14060, opened 2024-06-23
feat(SetTheory/Ordinal/Clubs): define club sets and prove basic properties

YnirPaz:
Create a file where club sets are defined and their basic properties are proven.

I also created a new recursion principle for ordinals, bounded recursion.

---

- [ ] depends on: #19189


Mathlib/SetTheory/Ordinal/Clubs.lean
YaelDillies, line 1: We don't (usually) use plural in file names

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
+/-
+Copyright (c) 2024 Nir Paz. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Nir Paz
+-/
+import Mathlib.SetTheory.Cardinal.Cofinality
+
```
YaelDillies, line 7:
```suggestion
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+
+/-!
+# Club and stationary sets
+This file sets up the basic theory of clubs (closed and unbounded sets) and stationary sets.
```
YaelDillies, line 11:
```suggestion

This file sets up the basic theory of clubs (closed and unbounded sets) and stationary sets.
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+
+## Main definitions
```
YaelDillies, line 13:
```suggestion
## Main definitions

```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+* `IsClosed`: A set of ordinals is closed in `o` if all of its elements are
+  less than `o` and it contains all of its accumulation points below `o`.
+* `IsClub`: A set of ordinals is a club in `o` if it is closed in `o` and unbounded in `o`.
+
+## Main results
```
YaelDillies, line 18:
```suggestion
## Main results

```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+
+
+/-!
+# Club and stationary sets
+This file sets up the basic theory of clubs (closed and unbounded sets) and stationary sets.
+
+## Main definitions
+* `IsClosed`: A set of ordinals is closed in `o` if all of its elements are
+  less than `o` and it contains all of its accumulation points below `o`.
+* `IsClub`: A set of ordinals is a club in `o` if it is closed in `o` and unbounded in `o`.
```
YaelDillies, line 16: You need to fully qualify these (for doc-gen and clarity)
```suggestion
* `Ordinal.IsClosed`: A set of ordinals is closed in `o` if all of its elements are
  less than `o` and it contains all of its accumulation points below `o`.
* `Ordinal.IsClub`: A set of ordinals is a club in `o` if it is closed in `o` and unbounded in `o`.
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+noncomputable section
+
+open Classical Cardinal Set Order
+
+universe u
+
+namespace Ordinal
+
+/-- A set of ordinals is unbounded in an ordinal if it
+  is a subset of it and it is an unbounded set in it. -/
```
YaelDillies, line 31: Please don't indent docstrings by default
```suggestion
is a subset of it and it is an unbounded set in it. -/
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+
+/-- A positive ordinal is an accumulation point of a set of ordinals if there
+  are elements in the set arbitrarily close to the ordinal from below. -/
+def IsAcc (o : Ordinal) (S : Set Ordinal) : Prop :=
+  0 < o ∧ (∀ p < o, ∃ s ∈ S, s < o ∧ p < s)
+
+/-- A set of ordinals is closed in an ordinals if it is a subset of it
+  and it contains all of its accumulation points below the ordinal. -/
+def IsClosed (S : Set Ordinal) (o : Ordinal) : Prop :=
+  S ⊆ (Iio o) ∧ (∀ p < o, IsAcc p S → p ∈ S)
```
YaelDillies, line 43: Unnecessary brackets
```suggestion
  S ⊆ Iio o ∧ ∀ p < o, IsAcc p S → p ∈ S
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+
+/-- A set of ordinals is unbounded in an ordinal if it
+  is a subset of it and it is an unbounded set in it. -/
+def IsUnbounded (S : Set Ordinal) (o : Ordinal) : Prop :=
+  S ⊆ (Iio o) ∧ (∀ p < o, ∃ s ∈ S, p < s)
+
+/-- A positive ordinal is an accumulation point of a set of ordinals if there
+  are elements in the set arbitrarily close to the ordinal from below. -/
+def IsAcc (o : Ordinal) (S : Set Ordinal) : Prop :=
+  0 < o ∧ (∀ p < o, ∃ s ∈ S, s < o ∧ p < s)
```
YaelDillies, line 38:
```suggestion
  0 < o ∧ ∀ p < o, ∃ s ∈ S, s < o ∧ p < s
```

YaelDillies, line 38: Similarly to `Nat`, prefer `≠ 0` to `0 <` on `Ordinal`
```suggestion
  o ≠ 0 ∧ (∀ p < o, ∃ s ∈ S, s < o ∧ p < s)
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+/-- A set of ordinals is closed in an ordinals if it is a subset of it
+  and it contains all of its accumulation points below the ordinal. -/
+def IsClosed (S : Set Ordinal) (o : Ordinal) : Prop :=
+  S ⊆ (Iio o) ∧ (∀ p < o, IsAcc p S → p ∈ S)
+
+/-- A set of ordinals is a club in an ordinal if it is closed and unbounded in it. -/
+def IsClub (S : Set Ordinal) (o : Ordinal) : Prop :=
+  IsClosed S o ∧ IsUnbounded S o
+
+theorem isAcc_of_subset {o : Ordinal} {S T : Set Ordinal} (h : S ⊆ T) (ho : o.IsAcc S) :
```
YaelDillies, line 49:
```suggestion
theorem IsAcc.subset {o : Ordinal} {S T : Set Ordinal} (h : S ⊆ T) (ho : o.IsAcc S) :
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+    o.IsAcc T := ⟨ho.1, fun p plto ↦ (ho.2 p plto).casesOn fun s hs ↦ ⟨s, h hs.1, hs.2⟩⟩
+
+theorem isLimit_of_isUnbounded {o : Ordinal} {U} (h : IsUnbounded U o)
```
YaelDillies, line 52:
```suggestion
theorem IsUnbounded.isLimit {o : Ordinal} {U} (h : IsUnbounded U o)
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+    (ho : o ≠ 0) : IsLimit o := by
+  refine' isLimit_of_not_succ_of_ne_zero (fun ⟨x, hx⟩ ↦ _) ho
+  rcases h.2 x (hx ▸ lt_succ x) with ⟨p, hp⟩
+  exact (hx ▸ succ_le_iff.mpr hp.2).not_lt (h.1 hp.1)
+
+theorem inter_suffix_of_isUnbounded {o : Ordinal} {U} (hU : IsUnbounded U o) {p : Ordinal}
```
YaelDillies, line 58: "suffix" is an obscure term here. Also dot notation is available
```suggestion
theorem IsUnbounded.inter_Ioo_nonempty {o : Ordinal} {U} (hU : IsUnbounded U o) {p : Ordinal}
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+    (hp : p < o) : (U ∩ (Ioo p o)).Nonempty :=
```
YaelDillies, line 59:
```suggestion
    (hp : p < o) : (U ∩ Ioo p o).Nonempty :=
```

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+  (hU.2 p hp).casesOn (fun x ⟨xmemU, pltx⟩ ↦ ⟨x, xmemU, ⟨pltx, hU.1 xmemU⟩⟩)
+
+
+section ClubIntersection
+
+variable {o : Ordinal.{u}} {S : Set (Set Ordinal)}
+
+theorem isClosed_sInter_of_isClosed (hS : S.Nonempty) (h : ∀ C ∈ S, IsClosed C o) :
```
YaelDillies, line 67:
```suggestion
theorem IsClosed.sInter (hS : S.Nonempty) (h : ∀ C ∈ S, IsClosed C o) :
```
for *anonymous dot notation*

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+    IsClosed (⋂₀ S) o :=
+  ⟨fun s hs ↦ by
+    rcases hS with ⟨C, CmemS⟩
+    exact (h C CmemS).1 ((sInter_subset_of_mem CmemS) hs),
+  fun p plto pAcc ↦ by
+    rw [mem_sInter]
+    intro C CmemS
+    exact (h C CmemS).2 p plto (isAcc_of_subset (sInter_subset_of_mem CmemS) pAcc)⟩
```
YaelDillies, line 75:
```suggestion
    IsClosed (⋂₀ S) o where
  mp s hs := by
    rcases hS with ⟨C, CmemS⟩
    exact (h C CmemS).1 ((sInter_subset_of_mem CmemS) hs)
  mpr p plto pAcc := by
    rw [mem_sInter]
    intro C CmemS
    exact (h C CmemS).2 p plto (isAcc_of_subset (sInter_subset_of_mem CmemS) pAcc)
```
(personally, I think that's much nicer when the `⟨_, _⟩` would otherwise span several lines)

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+  rw [lt_bsup] at pltsup
+  rcases pltsup with ⟨i, hi, plt⟩
+  rcases h i hi with ⟨q, qmemC, qmemIoo⟩
+  use q; use qmemC
+  exact ⟨lt_of_lt_of_le qmemIoo.2 (le_bsup _ _ _), plt.trans qmemIoo.1⟩
+
+/--
+The intersection of less than `o.cof` clubs in `o` is a club in `o`.
+-/
+theorem isClub_sInter (hCof : ℵ₀ < o.cof) (hS : ∀ C ∈ S, IsClub C o) (hSemp : S.Nonempty)
```
YaelDillies, line 220:
```suggestion
theorem IsClub.sInter (hCof : ℵ₀ < o.cof) (hS : ∀ C ∈ S, IsClub C o) (hSemp : S.Nonempty)
```

YaelDillies (commented):
You provide `sInter` lemmas. Can we also have the `iInter` versions?

zeramorphic:
A couple of thoughts about the structure of this PR:

`IsClosed s o` almost equivalent to `IsClosed s` for the topology on `Ordinal` defined in `Ordinal/Topology.lean`. If you define the topology `Preorder.topology` on `Set.Iio o`, the topological definition of closure agrees with yours. It might be better to define your lemmas in terms of the pre-existing definition of a closed set.

In fact, you can translate all of your new definitions into Mathlib's ones but ranging over `Iio o` rather than `Ordinal`: `IsUnbounded` -> `Set.Unbounded`, and `IsAcc` -> `Order.IsSuccLimit`. It might be worth looking at whether `IsAcc` agrees with `ClusterPt` or `AccPt` (or both?) with the preorder topology.

This isn't to say that your definitions are useless; I think they are important equivalent ways to say these ideas. But I'll defer to someone more experienced in how Mathlib works (pinging @YaelDillies) to determine what the correct spelling is.

YaelDillies:
I think the question really is "Do you need to talk about ordinals strictly less than `o`, or elements of `o`?". In the former case, we need the definitions in this PR. In the latter, we might as well reuse the existing definitions

YnirPaz:
@zeramorphic What's the practical benefit of using the topology definition? In theory it might give access to some API from mathlib, but in this case I don't see how anything beyond the very basic properties (like intersection of closed sets is closed) I already proved will ever be used when reasoning about these objects.

The more explicit definition is both (i) much closer to the one set theorists actually use (I've never seen topological reasoning used in this context), and (ii) very clearly corresponds to the "normal" mathematical definition.

I kinda think the topology version would just be a slightly obfuscated, slightly less practical definition for no real benefit, and would never be directly used in favor of the equivalent definitions I used here (I hope I'm not coming across too agressive, haha).

Your other comments look good, using `Set.Unbounded` might help with a few annoying proofs, so I'll try it (:

YaelDillies:
> The more explicit definition is both (i) much closer to the one set theorists actually use (I've never seen topological reasoning used in this context), and (ii) very clearly corresponds to the "normal" mathematical definition.

I should mention that you don't need `A` to be *defined* as `B` for people to use `A` as `B`. You are also free to prove `A ↔ B`. Here this would eg mean replacing `IsOrdinal.IsClosed` by `IsClosed` and showing `IsClosed` is equivalent to what was previously the definition of `IsOrdinal.IsClosed`

zeramorphic:
> I think the question really is "Do you need to talk about ordinals strictly less than `o`, or elements of `o`?". In the former case, we need the definitions in this PR. In the latter, we might as well reuse the existing definitions

I think this is a really important point. If we envision using clubs as plain old sets of ordinals, rather than inhabitants of `Set (Iio o)`, I think the definitions in the PR would be better.

YnirPaz:
> > I think the question really is "Do you need to talk about ordinals strictly less than `o`, or elements of `o`?". In the former case, we need the definitions in this PR. In the latter, we might as well reuse the existing definitions
> 
> I think this is a really important point. If we envision using clubs as plain old sets of ordinals, rather than inhabitants of `Set (Iio o)`, I think the definitions in the PR would be better.

Then I think the current definitions are better. For example I'd want to intersect clubs/stationary sets with `Iio o`'s and ask if they are still clubs/stationary sets, and having all of them be of different types would be a mess.

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+  · subst hi'
+    exact this.resolve_left (pos_of_gt rltf0).ne.symm
+  · have rltfi := rltf0.trans <| hf.2.1 0 i omega_pos hi (Ordinal.pos_iff_ne_zero.mpr hi')
+    exact this.resolve_left (pos_of_gt rltfi).ne.symm
+
+/-- If between every 2 consecutive elements of an increasing `δ`-sequence
+  there is an element of `C`, and `δ` is a limit ordinal,
+  then the supremum of the sequence is an accumulation point of `C`. -/
+theorem isAcc_bsup_of_between {δ : Ordinal} (C : Set Ordinal) (δLim : δ.IsLimit)
+    (s : Π o < δ, Ordinal) (sInc : ∀ o, (h : o < δ) → s o h < s (o + 1) (δLim.succ_lt h))
```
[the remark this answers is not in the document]
YnirPaz: I kind of defaulted to using `Π o < δ, Ordinal` because that's how fundamental sequences are defined, but `Iio δ` is probably better in many places (maybe including fundamental sequences, but that ship has sailed). I'll change it and work with `sup` instead of `bsup`.

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,243 @@
...
+Given a limit ordinal `o` and a property on pairs of ordinals `P`, such that
+for any `p < o` there is a `q < o` above `p` so that `P p q`, we can construct
+an increasing `ω`-sequence below `o` that satisfies `P` between every 2 consecutive elements.
+Additionaly, the sequence can begin arbitrarily high in `o`. That is, above any `r < o`.
+-/
+theorem exists_omega_seq_succ_prop (opos : 0 < o) {P : Ordinal → Ordinal → Prop}
+    (hP : ∀ p < o, ∃ q < o, (p < q ∧ P p q)) {r} (rlto : r < o) : ∃ f : Π p < ω, (Iio o),
+    (∀ i : Ordinal.{u}, (hi : i < ω) → P (f i hi) (f (i + 1) (omega_isLimit.2 i hi)))
+    ∧ (∀ i j, (hi : i < ω) → (hj : j < ω) → (i < j) → f i hi < f j hj)
+    ∧ r < f 0 omega_pos := by
```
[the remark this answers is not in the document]
YnirPaz: I used `ω` because this is supposed to be used with theorems like `isAcc_bsup_of_between` that work with arbitrary limit ordinals, doing it with `ℕ` first is mostly an implementation detail.

YnirPaz:
@alreadydone I changed things to work with `Iio` instead of `Π o <`, and now I actually think `Π o <` is better, because `Iio δ` is in a higher universe than `δ`. `Ordinal.sup` requires the range universe to be at least as high as the domain universe, but if the domain is `Iio ω` for example than we have to lift the output to universe `1`. With `Ordinal.bsup`, if we index with `Π o < ω`  then lean resolves with `ω : Ordinal.{0}`, and same thing when we index with any ordinal in the same universe as the range.

I pushed this not-done version, the problem starts at `isClub_sInter`. If we define `g` without lifting the output, we can't resolve the universe contraints of `sup g`, because `u` could be `0`. Then everything after that would have to deal with many lifts. Unless I'm missing something this makes `sup` much harder to work with than `bsup`, and I don't think its worth it. What's more, we should keep things working with the same definitions everywhere, so I would rather not use `Iio` for the domain of a sequence anywhere.
If that's okay with you I will revert these changes.

alreadydone:
Oh, I missed that there are existing APIs about "bfamilies" of the form `(a : Ordinal.{u}) → a < o → α`. I agree that those APIs make it convenient to work with such families rather than `Iio o → α`, so I won't insist you that change it.

On the other hand, you should be able to use `sSup (Set.range f)` (where `f` is a family) to avoid difficulties around universes. For `sSup` behaves as expected you just need it's range is bounded above, and there are lemmas like [Ordinal.bddAbove_iff_small](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Arithmetic.html#Ordinal.bddAbove_iff_small) and [Cardinal.small_iff_lift_mk_lt_univ](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Basic.html#Cardinal.small_iff_lift_mk_lt_univ) that could be useful.

Mathlib/SetTheory/Ordinal/Clubs.lean
```diff
@@ -0,0 +1,241 @@
...
+  fun p plto pAcc ↦ by
+    rw [mem_sInter]
+    intro C CmemS
+    exact (h C CmemS).2 p plto (isAcc_of_subset (sInter_subset_of_mem CmemS) pAcc)⟩
+
+/-- Given less than `o.cof` unbounded sets in `o` and some `q < o`, there is a `q < p < o`
+  such that `Ioo q p` contains an element of every unbounded set. -/
+theorem exists_above_of_lt_cof (p : Iio o) (hSemp : Nonempty S) --change!
+    (hSunb : ∀ U ∈ S, IsUnbounded U o) (hScard : #S < Cardinal.lift.{u + 1, u} o.cof) :
+    ∃ q : Iio o, p < q ∧ (∀ U ∈ S, (U ∩ Ioo p q).Nonempty) := by
```
alreadydone, line 81: This is the proof that I think can benefit from switching from `Ordinal.sup` to `sSup`.

Mathlib/SetTheory/Ordinal/Clubs.lean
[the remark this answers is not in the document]
alreadydone: [Ordinal.lift_type_eq](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Basic.html#Ordinal.lift_type_eq) or Ordinal.type_eq and (the definition) [Ordinal.type_nat_lt](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Basic.html#Ordinal.type_nat_lt) will be useful to construct a RelIso between `ℕ` and `Iio ω`. To compute the `type` of `Iio o` we may extract from [the proof of `small_Iio`](https://github.com/leanprover-community/mathlib4/blob/04e23488cdc6c40e3480f98100698e0566c1c192/Mathlib/SetTheory/Ordinal/Arithmetic.lean#L1169-L1178).

Mathlib/SetTheory/Ordinal/Clubs.lean
YnirPaz: I can change the name, but there is no input of type `IsClosed`, so how can dot notation be used? Maybe I don't understand exactly what anonymous means here.

YaelDillies: It's a new cousin of your usual beloved dot notation. It lets you write `.bar : Foo` to mean `Foo.bar : Foo`. The point being that, when the expected type is known to be `Foo`, `.bar` is interpreted as `Foo.bar`.

Mathlib/SetTheory/Ordinal/Clubs.lean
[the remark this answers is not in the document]
YnirPaz: I defined the RelIso, since this construction is already done I'll keep it as it is, but we have it for the future.

Mathlib/SetTheory/Ordinal/Arithmetic.lean
```diff
@@ -335,6 +338,60 @@ theorem limitRecOn_limit {C} (o H₁ H₂ H₃ h) :
...
+@[simp]
+theorem boundedLimitRec_limit {l : Ordinal} (lLim : l.IsLimit) {C} (o oltl H₁ H₂ H₃ oLim) :
+    @boundedLimitRec l lLim C H₁ H₂ H₃ o oltl = H₃ o oltl oLim (fun x hx ↦
+    @boundedLimitRec l lLim C H₁ H₂ H₃ x (hx.trans oltl)) := by
+  simp_all only [boundedLimitRec, eq_mpr_eq_cast, eq_mp_eq_cast, limitRecOn_limit,
+    dite_true, cast_cast, cast_eq]
+
+/-- Bounded recursion on the ordinals with a constant return type. Similar to `boundedLimitRec`,
+  but with a constant motive function. Useful for defining function to ordinals. -/
+def boundedLimitRec' {α : Sort*} {l : Ordinal} (lLim : l.IsLimit) (H₁ : α)
```
vihdzp, line 373: Is this really useful? 20 extra lines of API to save on a few characters doesn't seem like a good tradeoff to me.

Mathlib/SetTheory/Ordinal/Basic.lean
```diff
@@ -1213,6 +1216,42 @@ theorem enum_zero_eq_bot {o : Ordinal} (ho : 0 < o) :
...
+    · convert this.1.3
+      exact (iff_true_right this.1.3).mpr fun x ↦ congrArg ULift.up (this.1.3 x.down)
+  have aux : IsWellOrder (ULift.{u + 1, u} (Quotient.out o).α) (· < ·) := by
+    have : o.out.r ≃r fun x y : ULift.{u + 1, u} o.out.α ↦ x < y :=
+      ⟨⟨fun x ↦ ⟨x⟩, by aesop, congrFun rfl, congrFun rfl⟩, by aesop⟩
+    exact this.isWellOrder
+  convert this.symm.ordinal_type_eq
+  have : type fun x x_1 : ULift.{u + 1, u} (Quotient.out o).α ↦ x < x_1 =
+      lift.{u + 1, u} (type fun x x' : (Quotient.out o).α ↦ x < x') := rfl
+  rw [this, type_lt]
```
vihdzp, line 1248: This should be moved earlier in the file, near where `enumIso` is.
```lean 4
theorem type_Iio (o : Ordinal.{u}) : type (α := Iio o) (· < ·) = lift.{u + 1, u} o := by
  rw [←type_out o, ←lift_id.{u + 1} (type _), lift_type_eq.{u + 1, u, u + 1}]
  exact ⟨enumIso o.out.r⟩
```

Mathlib/SetTheory/Ordinal/Basic.lean
```diff
@@ -1213,6 +1216,42 @@ theorem enum_zero_eq_bot {o : Ordinal} (ho : 0 < o) :
...
+
+theorem relIso_nat_omega : Nonempty (((· < ·) : ℕ → ℕ → Prop)
```
vihdzp, line 1250: We could give an explicit isomorphism instead. This would be placed near the end of `SetTheory.Ordinal.Arithmetic`.
```lean 4
def relIso_nat_omega : ℕ ≃o Iio ω where
  toFun n := ⟨n, nat_lt_omega n⟩
  invFun n := Classical.choose (lt_omega.1 n.2)
  left_inv n := by
    have h : ∃ m : ℕ, n = (m : Ordinal) := ⟨n, rfl⟩
    exact (Ordinal.natCast_inj.1 (Classical.choose_spec h)).symm
  right_inv n := Subtype.eq (Classical.choose_spec (lt_omega.1 n.2)).symm
  map_rel_iff' := @natCast_le
  ```

**Review by vihdzp — COMMENTED** [review]
I'm really glad we're getting work done on this part of the library. That being said, I think there's some important questions to be asked about your current API.

**Comment by vihdzp** [comment]
> @zeramorphic What's the practical benefit of using the topology definition? In theory it might give access to some API from mathlib, but in this case I don't see how anything beyond the very basic properties (like intersection of closed sets is closed) I already proved will ever be used when reasoning about these objects.

That's not the point, really. The issue is that we have two definitions for the same concept, and that's always annoying. The first thing to do would be to show the topological definitions are equivalent to your simpler order-theoretic characterizations, at which point we just use those.

Topologists do make use of the topology of ordinals; that's why the file exist. Us set theorists admittedly don't care too much about that, but if we can benefit their work through ours, that's a good thing.

File: Mathlib/SetTheory/Ordinal/Clubs.lean (hunk 18 above)
Comment (vihdzp) [reply Mathlib/SetTheory/Ordinal/Clubs.lean]: For historical context. `Ordinal.sup` was originally `Ordinal.min`, with essentially the same definition. There's really no reason to have it anymore, it's just some technical debt no one's bothered to deal with.

File: Mathlib/SetTheory/Ordinal/Club.lean [hunk 22]
```diff
@@ -0,0 +1,263 @@
...
+noncomputable section
+
+open Classical Cardinal Set Order
+
+universe u v
+
+namespace Ordinal
+
+/-- A set of ordinals is unbounded in an ordinal if it is an unbounded subset of it. -/
+def IsUnbounded (S : Set Ordinal) (o : Ordinal) : Prop :=
```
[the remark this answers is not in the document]
YnirPaz: I tried that in a previous commit, but as I wrote [here](https://github.com/leanprover-community/mathlib4/pull/14060#issuecomment-2254207978), it was harder to work with and I don't think it's the right way to do things.

Mathlib/SetTheory/Ordinal/Club.lean
```diff
@@ -0,0 +1,263 @@
...
+
+/-!
+# Club and stationary sets
+
+This file sets up the basic theory of clubs (closed and unbounded sets) and stationary sets.
+
+## Main definitions
+
+* `Ordinal.IsClosed`: A set of ordinals `S` is closed in `o` if `S ⊆ Iio o`
+  and `S` contains every `x < o` such that `x.IsAcc S`.
```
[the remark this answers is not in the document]
YnirPaz: Alright I'll stop being stubborn and use the topological definition. To be clear, if we're sticking with `Set Ordinal` and not using `Set (Iio i)` (which I wrote about below), should the definition be that a set of ordinals is closed in `o` if its coercion to `Set (Iio o)` is closed in the order on `Iio o`?

Mathlib/SetTheory/Ordinal/Arithmetic.lean
```diff
@@ -2637,6 +2637,21 @@ theorem sup_mul_nat (o : Ordinal) : (sup fun n : ℕ => o * n) = o * ω := by
...
+def relIso_nat_omega : ℕ ≃o Iio ω where
+  toFun n := ⟨n, nat_lt_omega n⟩
+  invFun n := Classical.choose (lt_omega.1 n.2)
+  left_inv n := by
+    have h : ∃ m : ℕ, n = (m : Ordinal) := ⟨n, rfl⟩
+    exact (Ordinal.natCast_inj.1 (Classical.choose_spec h)).symm
+  right_inv n := Subtype.eq (Classical.choose_spec (lt_omega.1 n.2)).symm
+  map_rel_iff' := @natCast_le
+
+theorem relIso_nat_omega.symm_eq {o : Ordinal} (h : o < ω) :
```
vihdzp, line 2649: Isn't this just [`Equiv.symm_apply_apply`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Equiv/Defs.html#Equiv.symm_apply_apply)?

Mathlib/SetTheory/Ordinal/Club.lean
```diff
@@ -0,0 +1,263 @@
...
+  rw [← hq'] at fltq
+  rcases lift_down fltq.le with ⟨fUdown, fUlift⟩
+  use fUdown
+  constructor
+  · simp_all only [lift_inj, mem_inter_iff, f]
+  · constructor
+    exact lift_lt.mp <| fUlift ▸ (this.1)
+    exact lift_lt.mp <| hq' ▸ (fUlift ▸ this).2
+
+theorem strictMono_of_succ_lt_omega (f : Π p < ω, Iio o)
```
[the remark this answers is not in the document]
YnirPaz: I changed the proof to use `relIso_nat_omega`, but I do want to keep this one with `Π i < ω, α` because there are (and will be more) theorems about such sequences with length `δ` for a limit ordinal `δ`.

Mathlib/SetTheory/Ordinal/Basic.lean
```diff
@@ -413,6 +413,9 @@ theorem eq_zero_or_pos : ∀ a : Ordinal, a = 0 ∨ 0 < a :=
   eq_bot_or_bot_lt
 #align ordinal.eq_zero_or_pos Ordinal.eq_zero_or_pos
 
+theorem pos_of_gt {o p : Ordinal} (h : p < o) : 0 < o :=
```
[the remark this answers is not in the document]
YnirPaz: Ah... I did this but it's causing build errors in some files that use `pos_of_gt`, which don't show up for me locally. Sorry, I'm not sure how to deal with this, what should I do here?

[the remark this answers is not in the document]
vihdzp: Don't remove the existing theorem, instead make a new one `pos_of_gt'`. Since this is a change whose relevancy spans the entire library, it's better to put it in a separate PR.

Mathlib/SetTheory/Ordinal/Club.lean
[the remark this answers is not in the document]
vihdzp: I strongly disagree. If `Iio o` is harder to work with, then we're missing some important API. You should be able to take the supremum of a family indexed by it using `Ordinal.bsup` (which thinking about it, should probably have its type signature changed to reflect that).

Mathlib/SetTheory/Ordinal/Club.lean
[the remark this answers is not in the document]
vihdzp: We should be using `Set (Iio)` though, see my comment there.

Mathlib/SetTheory/Ordinal/Club.lean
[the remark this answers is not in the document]
vihdzp: Actually, we should probably phase both of those out in favor of the common notation for conditionally complete lattices. See here: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ordinal.20technical.20debt

In the meanwhile, I recommend you stick to `bsup`.

.vscode/settings.json
```diff
@@ -21,4 +21,4 @@
     "**/.elan/**": true,
     "**/.lake/**": true
   }
-}
+}
```
dupuisf, line 24: Can you make sure this file doesn't get changed?

dupuisf:
To proceed further with this PR, can you make sure that all the outstanding comments by reviewers have been resolved (one way or another)? If a reviewer comment has been dealt with or is no longer relevant, you can click on "Resolve conversation" to make this clear. If you are currently working on the PR and feel that the PR is not ready for review yet, you can label it as "WIP" (i.e. work in progress). Also, feel free to ask for help on Zulip at any time!

vihdzp:
This PR has clearly amassed a pretty large line count in both content and discussion, so I think it'd be good to split off the less controversial changes, such as `pos_of_gt'` and bounded recursion. You can even make an initial PR defining `IsClosed` and `IsAcc`, and then follow that up with this PR defining club sets.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
     exact le_bsup.{u, u} _ _ (ha.2 _ hba)
 #align ordinal.enum_ord_is_normal_iff_is_closed Ordinal.enumOrd_isNormal_iff_isClosed
 
+open Set Filter
+
+/-- A positive ordinal is an accumulation point of a set of ordinals if there
+are elements in the set arbitrarily close to the ordinal from below. -/
+def IsAcc (o : Ordinal) (S : Set Ordinal) : Prop :=
+  o ≠ 0 ∧ ∀ p < o, ∃ s ∈ S, p < s ∧ s < o
```
vihdzp, line 235: You can write this as
```suggestion
  o ≠ 0 ∧ ∀ p < o, (S ∩ Set.Ioo p o).Nonempty
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+
+/-- A set of ordinals is closed in an ordinal if it contains all of
```
vihdzp, line 237: I don't like the use of "in an ordinal". This phrasing makes sense in pen-and-paper mathematics where you treat ordinals as sets, but that's not quite what we're doing here. Maybe "below an ordinal" instead?

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
     exact le_bsup.{u, u} _ _ (ha.2 _ hba)
 #align ordinal.enum_ord_is_normal_iff_is_closed Ordinal.enumOrd_isNormal_iff_isClosed
 
+open Set Filter
+
+/-- A positive ordinal is an accumulation point of a set of ordinals if there
```
vihdzp, line 232:
```suggestion
/-- An ordinal is an accumulation point of a set of ordinals if it is positive and there
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+    {p : Ordinal} (hp : p < o) : (U ∩ Ioo p o).Nonempty := by
+  rcases hU.2 p hp with ⟨q, hq⟩
+  exact ⟨q, hq.1, hq.2.1, hq.2.2⟩
+
+theorem isClosed_zero (S : Set Ordinal) : IsClosed S 0 := fun _ h ↦
+  False.elim <| (Ordinal.zero_le _).not_lt h
+
+theorem isAcc_of_accPt {o p : Ordinal} (A : Set Ordinal) (hpos : p ≠ 0) (hp : p < o)
+    (h : AccPt ⟨p, hp⟩ (𝓟 {x : Iio o | x.1 ∈ A})) : IsAcc p A := by
+  use hpos
```
vihdzp, line 260: You can remove the `hpos` hypothesis by proving `not_accPt_zero` earlier.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+
+theorem IsAcc.inter_Ioo_nonempty {o : Ordinal} {U : Set Ordinal} (hU : o.IsAcc U)
+    {p : Ordinal} (hp : p < o) : (U ∩ Ioo p o).Nonempty := by
+  rcases hU.2 p hp with ⟨q, hq⟩
+  exact ⟨q, hq.1, hq.2.1, hq.2.2⟩
+
+theorem isClosed_zero (S : Set Ordinal) : IsClosed S 0 := fun _ h ↦
+  False.elim <| (Ordinal.zero_le _).not_lt h
+
+theorem isAcc_of_accPt {o p : Ordinal} (A : Set Ordinal) (hpos : p ≠ 0) (hp : p < o)
```
vihdzp, line 258:
```suggestion
theorem AccPt.IsAcc {o p : Ordinal} (A : Set Ordinal) (hpos : p ≠ 0) (hp : p < o)
```
properly namespaced if needed.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
     exact le_bsup.{u, u} _ _ (ha.2 _ hba)
 #align ordinal.enum_ord_is_normal_iff_is_closed Ordinal.enumOrd_isNormal_iff_isClosed
```
vihdzp: We got rid of `#align` ages ago, can you merge master?

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+    exact this.lt_or_eq.resolve_right <| Subtype.coe_ne_coe.mpr rnep
+  · let I : Set (Iio o) := Ioo ⟨q, qltp.trans hp⟩ ⟨p + 1,
+      (succ_le_of_lt hp).lt_or_eq.resolve_right h'⟩
+    have Inh : I ∈ nhds ⟨p, hp⟩ := IsOpen.mem_nhds isOpen_Ioo ⟨qltp, lt_add_one p⟩
+    rcases h I Inh with ⟨r, ⟨rmemI, rmemA⟩, rnep⟩
+    use r, rmemA, rmemI.1
+    have : r.1 ≤ p := lt_succ_iff.mp rmemI.2
+    exact this.lt_or_eq.resolve_right <| Subtype.coe_ne_coe.mpr rnep
+
+theorem accPt_of_isAcc {o p : Ordinal} (A : Set Ordinal) (hp : p < o)
```
vihdzp, line 278:
```suggestion
theorem IsAcc.AccPt {o p : Ordinal} (A : Set Ordinal) (hp : p < o)
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+    (h : IsAcc p A) : AccPt ⟨p, hp⟩ (𝓟 {x : Iio o | x.1 ∈ A}) := by
+  rw [accPt_iff_nhds]
+  intro u hu
+  have lb := exists_Ioc_subset_of_mem_nhds hu ⟨⟨0, pos_of_gt' hp⟩,
+    (Ordinal.zero_le p).lt_or_eq.resolve_right h.1.symm⟩
+  rcases lb with ⟨a, pmem, hi⟩
```
vihdzp, line 284:
```suggestion
  obtain ⟨a, pmem, hi⟩ := exists_Ioc_subset_of_mem_nhds hu ⟨⟨0, pos_of_gt' hp⟩,
    (Ordinal.zero_le p).lt_or_eq.resolve_right h.1.symm⟩
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+  rcases h.2 a pmem with ⟨x, hx⟩
+  use ⟨x, hx.2.2.trans hp⟩
+  exact ⟨⟨hi ⟨hx.2.1, hx.2.2.le⟩, hx.1⟩, fun h ↦ hx.2.2.ne <| Subtype.mk_eq_mk.mp h⟩
```
vihdzp, line 287: You can combine these two if you want.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+  use hpos
+  rw [accPt_iff_nhds] at h
+  intro q qltp
+  by_cases h' : Order.succ p = o
+  · let I : Set (Iio o) := Ioi ⟨q, qltp.trans hp⟩
+    have Inh : I ∈ nhds ⟨p, hp⟩ := IsOpen.mem_nhds isOpen_Ioi qltp
+    rcases h I Inh with ⟨r, ⟨rmemI, rmemA⟩, rnep⟩
+    use r, rmemA, rmemI
+    have : r ≤ p := Order.lt_succ_iff.mp (h' ▸ r.2)
+    exact this.lt_or_eq.resolve_right <| Subtype.coe_ne_coe.mpr rnep
```
vihdzp, line 269: If you move the `have` block one line earlier, you can combine `use` and `exact` into a single `exact`.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+  rcases h I this with ⟨x, ⟨hx, _⟩, xnz⟩
+  exact xnz (SetCoe.ext (lt_one_iff_zero.mp hx))
+  · have := (not_lt_iff_eq_or_lt.mp h').resolve_right fun h ↦
+      ho <| Ordinal.lt_one_iff_zero.mp h
+    rcases h Set.univ univ_mem with ⟨x, _, hx⟩
+    exact hx <| SetCoe.ext <| Ordinal.lt_one_iff_zero.mp (this ▸ x.2)
+
+theorem isClosed_iff {o : Ordinal} (A : Set Ordinal) :
+    IsClosed A o ↔ _root_.IsClosed {x : Iio o | x.1 ∈ A} := by
+  by_cases ho : o = 0
```
vihdzp, line 310:
```suggestion
  obtain rfl | ho : eq_or_ne o 0
```
This automatically rewrites `o = 0` in the first case.

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+  · rw [ho, eq_true (isClosed_zero A)]
+    have : {x : Iio o | x.1 ∈ A} = ∅ := by
+      apply eq_empty_iff_forall_not_mem.mpr
+      intro x
+      have := x.2
+      simp_rw [ho] at this
+      cases (Ordinal.zero_le _).not_lt this
```
vihdzp, line 317:
```suggestion
    have : {x : Iio 0 | x.1 ∈ A} = ∅ := Set.eq_empty_of_isEmpty _
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+    have : _root_.IsClosed {x : Iio o | x.1 ∈ A} := by
+      rw [this]
+      exact isClosed_empty
+    have := ho ▸ (eq_true this)
+    rw [this]
+  rw [isClosed_iff_clusterPt]
+  constructor
+  · intro h
+    intro r hr
```
vihdzp, line 326:
```suggestion
  · intro h r hr
```

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+    change r.1 ∈ A
+    match clusterPt_principal hr with
+    | .inl h' => exact h'
+    | .inr h' =>
+      have rnz : r.1 ≠ 0 := fun h ↦ by
+        have := not_accPt_zero ho {x : Iio o | x.1 ∈ A}
+        simp_rw [← h] at this
+        exact this h'
+      have := (isAcc_iff_accPt A r.2 rnz)
+      exact h r.1 r.2 (this.mpr h')
```
vihdzp, line 336: You can merge this into a single `exact`

Mathlib/SetTheory/Ordinal/Topology.lean
```diff
@@ -234,4 +234,126 @@ theorem enumOrd_isNormal_iff_isClosed (hs : s.Unbounded (· < ·)) :
...
+  · intro h p plto hp
+    apply h ⟨p, plto⟩
+    have := (isAcc_iff_accPt A plto hp.1).mp hp
+    apply ((acc_iff_cluster (⟨p, plto⟩ : Iio o) _).mp this).mono inf_le_right
```
vihdzp, line 340: Same here.

vihdzp:
Well, a lot has changed in this past month or so. Can you merge master and update the PR description?

YnirPaz:
> Well, a lot has changed in this past month or so. Can you merge master and update the PR description?

I kind of forgot about this PR and was planning on PRing all of this gradually. But seeing how long it takes for things to get merged, I'll update this one and maybe get it all done at once.

btw I moved and updated some related old code [here](https://github.com/YnirPaz/InfiniteCombinatorics) and am planning on adding to it soon.

Mathlib/Algebra/Order/Monoid/Canonical/Defs.lean
```diff
@@ -175,6 +175,9 @@ end CanonicallyOrderedCommMonoid
 theorem pos_of_gt {M : Type*} [CanonicallyOrderedAddCommMonoid M] {n m : M} (h : n < m) : 0 < m :=
   lt_of_le_of_lt (zero_le _) h
 
+theorem pos_of_gt' {M : Type*} [Preorder M] [OrderBot M] {n m : M} (h : n < m) : ⊥ < m :=
```
vihdzp, line 178: I think this was already discussed, but this theorem already exists as `bot_lt_of_lt`.

Mathlib/SetTheory/Ordinal/Arithmetic.lean
```diff
@@ -2434,6 +2434,30 @@ theorem sup_mul_nat (o : Ordinal) : (sup fun n : ℕ => o * n) = o * ω := by
...
+  right_inv n := Subtype.eq (Classical.choose_spec (lt_omega0.1 n.2)).symm
+  map_rel_iff' := fun {n m} ↦ by simp only [coe_fn_mk, Subtype.mk_le_mk, Nat.cast_le]
+
+theorem relIso_nat_omega.symm_eq {o : Ordinal} (h : o < ω) :
+    relIso_nat_omega.symm ⟨o, h⟩ = o := by
+  rcases lt_omega0.mp h with ⟨n, rfl⟩
+  exact congrArg Nat.cast <| relIso_nat_omega.symm_apply_apply n
+
+theorem strictMono_of_succ_lt_omega {o : Ordinal} (f : Iio ω → Iio o)
+    (hf : ∀ i, f i < f ⟨succ i, isLimit_omega0.succ_lt i.2⟩) (i j) (h : i < j) : f i < f j := by
```
vihdzp, line 2453:
```suggestion
    (hf : ∀ i, f i < f ⟨succ i, isLimit_omega0.succ_lt i.2⟩) {i j} (h : i < j) : f i < f j := by
```

Mathlib/SetTheory/Ordinal/Club.lean
```diff
@@ -0,0 +1,285 @@
...
+noncomputable section
+
+open Classical Cardinal Set Order Filter
+
+universe u v
+
+namespace Ordinal
+
+/-- A set of ordinals is a club below an ordinal if it is closed and unbounded in it. -/
+def IsClub (C : Set Ordinal) (o : Ordinal) : Prop :=
```
vihdzp, line 35: Question, does it make sense to have two notions of club sets? One for "club below an ordinal" and also a global one. An example for the latter might be the fixed points of a normal function.

On a similar note, does it make sense to define this for general linear/well orders with the order topology?

Mathlib/SetTheory/Ordinal/Arithmetic.lean
```diff
@@ -2434,6 +2434,30 @@ theorem sup_mul_nat (o : Ordinal) : (sup fun n : ℕ => o * n) = o * ω := by
     exact sup_eq_zero_iff.2 fun n => zero_mul (n : Ordinal)
   · exact (mul_isNormal ho).apply_omega0
 
+/-- The natural isomorphism between ℕ and the first ω ordinals. -/
+def relIso_nat_omega : ℕ ≃o Iio ω where
```
vihdzp, line 2438: This could be its own PR.

Mathlib/SetTheory/Ordinal/Club.lean
YnirPaz: I assumed we'll just use "the pointwise lift is a club below the universal ordinal" directly for a proper club, but the more I think about it the more annoying that sounds. Things like "the restriction of a universal club at a limit point creates a club below that limit point" would need some universe hopping. So I guess the question is can a definition like this have enough API that we wouldn't have to deal with universes when using it. And I don't know (of course this wouldn't be in this PR anyways). You have experience with universe shenanigans, do you think this would make things more tidy?

I've never heard about clubs over non-wellorders and a quick search didn't yield anything. I don't know if people outside this field ever care a lot if their closed sets in some order topology are also unbounded, to the point that they satisfy many things arbitrary closed sets don't.

Mathlib/SetTheory/Ordinal/Arithmetic.lean
YnirPaz: #19189

YaelDillies:
You got the `- [ ] depends on:` syntax wrong, so the bot didn't pick up the dependency. I should have fixed it now

PR #14065, opened 2024-06-23
[Merged by Bors] - feat(ContinuousFunctionalCalculus): add several lemmas involving the CFC and `algebraMap`

dupuisf:
This PR adds several lemmas about the interaction between the (non-unital) CFC and `algebraMap`. Several lemmas require some sort of nontriviality statement for the CFC (i.e. the predicate can't be false everywhere), I just stated it as an `hp : p 0` hypothesis in the lemma statements; it seems like the most convenient way in practice and probably the easiest to automate.

---

- [x] depends on: #14100


Mathlib/Topology/ContinuousFunction/FunctionalCalculus.lean
```diff
@@ -582,6 +582,45 @@ lemma eq_one_of_spectrum_subset_one (h_spec : spectrum R a ⊆ {1}) (ha : p a :=
     a = 1 := by
   simpa using eq_algebraMap_of_spectrum_subset_singleton a 1 h_spec
 
+lemma spectrum_algebraMap_subset (r : R) : spectrum R (algebraMap R A r) ⊆ {r} := by
+  rw [← cfc_const r 0 (cfc_predicate_zero R),
+    cfc_map_spectrum (fun _ ↦ r) 0 (cfc_predicate_zero R)]
+  rintro - ⟨x, -, rfl⟩
+  simp
+
+lemma spectrum_algebraMap_eq [Nontrivial A] (r : R) :
```
j-loreaux, line 591: Can you namespace the ones that don't have `cfc` to the `CFC` namespace? (including the few lemmas above that you didn't write)

It's strange to have `spectrum_zero_eq` for example, in the root namespace, with a `ContinuousFunctionalCalculus` hypothesis, especially when we have `spectrum.zero_eq` already.

j-loreaux:
bors d+

Mathlib/Topology/ContinuousFunction/FunctionalCalculus.lean
dupuisf: Yes, fair enough.

PR #14066, opened 2024-06-24
[Merged by Bors] - feat(Topology/UniformSpace): add `UniformContinuous.iterate `

pitmonticone:
This PR adds the theorem `UniformContinuous.iterate` claiming that if a function `T` is uniformly continuous in a uniform space `β`, then its `n`-th iterate `T^[n]` is also uniformly continuous.

Co-authored-by: @D-Thomine 

---

### DIFF
--- a/Mathlib/Topology/UniformSpace/Basic.lean
+++ b/Mathlib/Topology/UniformSpace/Basic.lean
@@ -1122,12 +1122,11 @@ nonrec theorem UniformContinuous.comp [UniformSpace β] [UniformSpace γ] {g : �
 
 /--If a function `T` is uniformly continuous in a uniform space `β`,
 then its `n`-th iterate `T^[n]` is also uniformly continuous.-/
-theorem uniformContinuous_ite [UniformSpace β] (T : β → β) (n : ℕ) (h : UniformContinuous T) :
+theorem UniformContinuous.iterate [UniformSpace β] (T : β → β) (n : ℕ) (h : UniformContinuous T) :
     UniformContinuous T^[n] := by
   induction' n with n hn
   · exact uniformContinuous_id
   · exact Function.iterate_succ _ _ ▸ UniformContinuous.comp hn h
-#align uniform_continuous_ite uniformContinuous_ite
 
 theorem Filter.HasBasis.uniformContinuous_iff {ι'} [UniformSpace β] {p : ι → Prop}
     {s : ι → Set (α × α)} (ha : (𝓤 α).HasBasis p s) {q : ι' → Prop} {t : ι' → Set (β × β)}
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/UniformSpace/Basic.lean
```diff
@@ -1120,6 +1120,15 @@ nonrec theorem UniformContinuous.comp [UniformSpace β] [UniformSpace γ] {g : 
   hg.comp hf
 #align uniform_continuous.comp UniformContinuous.comp
 
+/--If a function `T` is uniformly continuous in a uniform space `β`,
+then its `n`-th iterate `T^[n]` is also uniformly continuous.-/
+theorem uniformContinuous_ite [UniformSpace β] (T : β → β) (n : ℕ) (h : UniformContinuous T) :
```
sgouezel, line 1125: The name is bad, because `ite` means `if then else`. What about `UniformContinuous.iterate`?

Mathlib/Topology/UniformSpace/Basic.lean
```diff
@@ -1120,6 +1120,15 @@ nonrec theorem UniformContinuous.comp [UniformSpace β] [UniformSpace γ] {g : 
...
+    UniformContinuous T^[n] := by
+  induction' n with n hn
```
sgouezel, line 1127: Does `induction` work instead of `induction'`? (It is more canonical).

Mathlib/Topology/UniformSpace/Basic.lean
```diff
@@ -1120,6 +1120,15 @@ nonrec theorem UniformContinuous.comp [UniformSpace β] [UniformSpace γ] {g : 
...
+  · exact uniformContinuous_id
+  · exact Function.iterate_succ _ _ ▸ UniformContinuous.comp hn h
+#align uniform_continuous_ite uniformContinuous_ite
```
sgouezel, line 1130: No #align for new statements, just remove this line.

Mathlib/Topology/UniformSpace/Basic.lean
pitmonticone: Ok.

Mathlib/Topology/UniformSpace/Basic.lean
pitmonticone: No, it doesn't work if I naively substitute it for `induction'`.

Mathlib/Topology/UniformSpace/Basic.lean
pitmonticone: Done.

pitmonticone:
Ok, I should have addressed all the review requests (apart from the `induction` vs. `induction'` that seems to be minor). 

Thank you very much @sgouezel for your review!

Mathlib/Topology/UniformSpace/Basic.lean
sgouezel: Sure, naive substitution won't work, but `induction'` is suboptimal and will probably be removed some day, so better learn the new syntax. Can you figure it out (for instance looking for `induction n` around)? If you can't, I can show you how it should be done, but I think it's more instructive to let you find out.

pitmonticone: Sure, I'll do it as soon as I can.

pitmonticone: Done.

pitmonticone:
Ok, I should have addressed all the review requests.

Thank you very much @sgouezel for your review!

sgouezel:
bors r+
Thanks!

PR #14069, opened 2024-06-24
[Merged by Bors] - feat: if the body constrains universes, make it explicit in the signature.

kim-em:
In many places in Mathlib universes in the type signature are invisibly constrained by the RHS of the `def`.

This PR makes all these explicit in the type signature.

Likely we will change the Lean behaviour to disallow this in https://github.com/leanprover/lean4/pull/4482 (i.e. this is the backport to `master` of the fixes required for that).

### DIFF
--- a/Mathlib/Data/Fintype/Prod.lean
+++ b/Mathlib/Data/Fintype/Prod.lean
@@ -85,14 +85,14 @@ instance Pi.infinite_of_left {ι : Sort*} {π : ι → Sort _} [∀ i, Nontrivia
 #align pi.infinite_of_left Pi.infinite_of_left
 
 /-- If at least one `π i` is infinite and the rest nonempty, the pi type of all `π` is infinite. -/
-theorem Pi.infinite_of_exists_right {ι : Type*} {π : ι → Type*} (i : ι) [Infinite <| π i]
+theorem Pi.infinite_of_exists_right {ι : Sort _} {π : ι → Sort _} (i : ι) [Infinite <| π i]
     [∀ i, Nonempty <| π i] : Infinite (∀ i : ι, π i) :=
   let ⟨m⟩ := @Pi.instNonempty ι π _
   Infinite.of_injective _ (update_injective m i)
 #align pi.infinite_of_exists_right Pi.infinite_of_exists_right
 
 /-- See `Pi.infinite_of_exists_right` for the case that only one `π i` is infinite. -/
-instance Pi.infinite_of_right {ι : Sort _} {π : ι → Sort _} [∀ i, Infinite <| π i] [Nonempty ι] :
+instance Pi.infinite_of_right {ι : Sort _} {π : ι → Type _} [∀ i, Infinite <| π i] [Nonempty ι] :
     Infinite (∀ i : ι, π i) :=
   Pi.infinite_of_exists_right (Classical.arbitrary ι)
 #align pi.infinite_of_right Pi.infinite_of_right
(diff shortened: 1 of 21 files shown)


Mathlib/Data/Fintype/Prod.lean
```diff
@@ -85,14 +85,14 @@ instance Pi.infinite_of_left {ι : Sort*} {π : ι → Sort _} [∀ i, Nontrivia
 #align pi.infinite_of_left Pi.infinite_of_left
 
 /-- If at least one `π i` is infinite and the rest nonempty, the pi type of all `π` is infinite. -/
-theorem Pi.infinite_of_exists_right {ι : Type*} {π : ι → Type*} (i : ι) [Infinite <| π i]
+theorem Pi.infinite_of_exists_right {ι : Sort _} {π : ι → Sort _} (i : ι) [Infinite <| π i]
```
sgouezel, line 88:
```suggestion
theorem Pi.infinite_of_exists_right {ι : Sort*} {π : ι → Sort*} (i : ι) [Infinite <| π i]
```
`Sort _` is a code smell to me, as it could indicate that there is something nontrivial to be inferred on universes, that you're leaving Lean to resolve by itself. Better use `Sort*` when it works, or `Sort (max u v)` say when needed.

On Line 79 of the same file, there is the same kind of issue (a `Sort _` which only makes sense in Type, and so should be a `Type*`). I am surprised that this one did not create issues for you. Could you nevertheless still fix it?

Mathlib/Data/Fintype/Prod.lean
```diff
@@ -85,14 +85,14 @@ instance Pi.infinite_of_left {ι : Sort*} {π : ι → Sort _} [∀ i, Nontrivia
...
     [∀ i, Nonempty <| π i] : Infinite (∀ i : ι, π i) :=
   let ⟨m⟩ := @Pi.instNonempty ι π _
   Infinite.of_injective _ (update_injective m i)
 #align pi.infinite_of_exists_right Pi.infinite_of_exists_right
 
 /-- See `Pi.infinite_of_exists_right` for the case that only one `π i` is infinite. -/
-instance Pi.infinite_of_right {ι : Sort _} {π : ι → Sort _} [∀ i, Infinite <| π i] [Nonempty ι] :
+instance Pi.infinite_of_right {ι : Sort _} {π : ι → Type _} [∀ i, Infinite <| π i] [Nonempty ι] :
```
sgouezel, line 95:
```suggestion
instance Pi.infinite_of_right {ι : Sort*} {π : ι → Type*} [∀ i, Infinite <| π i] [Nonempty ι] :
```

Mathlib/Data/Fintype/Prod.lean
```diff
@@ -103,7 +103,8 @@ instance Function.infinite_of_left {ι π : Sort _} [Nontrivial π] [Infinite ι
 #align function.infinite_of_left Function.infinite_of_left
 
 /-- Non-dependent version of `Pi.infinite_of_exists_right` and `Pi.infinite_of_right`. -/
-instance Function.infinite_of_right {ι π : Sort _} [Infinite π] [Nonempty ι] : Infinite (ι → π) :=
+instance Function.infinite_of_right {ι : Sort _} {π : Type _} [Infinite π] [Nonempty ι] :
```
sgouezel, line 106:
```suggestion
instance Function.infinite_of_right {ι : Sort*} {π : Type*} [Infinite π] [Nonempty ι] :
```
There is the same issue on Line 101, right?

Mathlib/Logic/Equiv/Basic.lean
```diff
@@ -1976,7 +1976,7 @@ end
 
 section BinaryOp
 
-variable (e : α₁ ≃ β₁) (f : α₁ → α₁ → α₁)
+variable {α₁ β₁ : Type _} (e : α₁ ≃ β₁) (f : α₁ → α₁ → α₁)
```
sgouezel, line 1979:
```suggestion
variable {α₁ β₁ : Type*} (e : α₁ ≃ β₁) (f : α₁ → α₁ → α₁)
```

YaelDillies (commented):
The other changes look good.

Just to make sure, by RHS you mean `example : LHS := RHS`, not `example : LHS = RHS := ...`? I would rather call that the "body" of the declaration.

sgouezel:
bors r+
Thanks!

PR #14070, opened 2024-06-24
experiment: globally increase maxSynthPendingDepth

kim-em:


semorrison:
!bench

semorrison:
I think the +1.4% walllclock and +0.3% instructions make this slightly too expensive to consider now, but it's not nearly as bad as I feared.

PR #13082, opened 2024-05-21
[Merged by Bors] - chore: move LinearOrder Nat instance to Mathlib.Data.Nat.Defs

Ruben-VandeVelde:
---

Preliminary to removing the order dependency from `Mathlib/Data/Nat/Defs.lean`, but that needs a little more work

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14081, opened 2024-06-24
chore: Move Nat.instLinearOrder to Init.Order.Defs

Ruben-VandeVelde:
It turns out that pushing the dependency out is quite annoying, so as an alternative, we propose considering Preorder, PartialOrder and LinearOrder to be foundational type classes that all of Mathlib can depend on.

---

### DIFF
--- a/Mathlib/Data/Char.lean
+++ b/Mathlib/Data/Char.lean
@@ -3,9 +3,9 @@ Copyright (c) 2018 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
-import Mathlib.Init.Data.Nat.Lemmas
+import Mathlib.Init.Data.Nat.Notation
 import Mathlib.Init.Order.Defs
-import Mathlib.Data.Nat.Defs
+import Mathlib.Tactic.SplitIfs
 
 #align_import data.char from "leanprover-community/mathlib"@"c4658a649d216f57e99621708b09dcb3dcccbd23"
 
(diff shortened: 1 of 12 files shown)


YaelDillies:
Ahah! That's basically the criticism I had against #13082. I am perfectly happy with the revert.

maintainer merge

semorrison:
I'm concerned about moving material into `Init`, when we are trying to get rid of all these files. Can we achieve your end here without backsliding on this?

semorrison:
Further (this is less urgent, but flagging a medium-term concern) I am cautious about considering `LinearOrder` as fundamental --- it is clear that for programming purposes we are going to need a substantial revision of the treatment of decidable inequalities and typeclasses guaranteeing properties of these functions. No one as yet has a concrete design, but it's clear that radical changes are needed at some point. I would suspect that `LinearOrder` itself will need to be revised as part of this (and hopefully will become less necessary in some applications).

I don't think reorganizations like this PR really add to the difficulty of refactors like this, but I think it's worth keeping in mind.

Ruben-VandeVelde:
The idea (as I understand it right now) is that we'd move this entire file into a new library (within this repository) for mathlib to depend on, so I didn't see much point in moving it out of Init at this point. I have no opinion on the programming treatment of ordering.

semorrison:
Oh, I think the short term plan is actually the opposite, to move all of the content of the `Init` files out into Mathlib proper, and then delete them.

semorrison:
I would *love* to start splitting out the bottom of Mathlib into separate libraries within the Mathlib repo, but as far as I understand this proposal is still foundering on the shoals of decision-making. :-)

riccardobrasca:
What is the status of this?

YaelDillies:
What about you move the instance to `Data.Nat.Defs` for now?

PR #14087, opened 2024-06-24
[Merged by Bors] - feat: commutation with shifts of functors from quotient categories

joelriou:
When a relation on morphisms `r : HomRel C` is compatible with the shift by an additive monoid, and `F : C ⥤ D` is a functor which commutes with the shift and is compatible with `r`, then the induced functor `Quotient.lift r F _ : Quotient r ⥤ D` also commutes with the shift.

---


erdOne (commented):
Thanks!
maintainer merge

semorrison:
bors merge

PR #14091, opened 2024-06-24
[Merged by Bors] - feat: add basic `Digraph` file

jt18chev:
Added the file `Basic.lean` which includes the basic definition of a digraph along with some basic theorems.

---


vasnesterov:
Hi there! I spotted this PR in the queue and decided to take a look. This is a great contribution! I left a few minor comments for you.

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,252 @@
...
+import Mathlib.Tactic.ApplyFun
+
+/-!
+# Digraphs
+
+This module defines directed graphs on a vertex type `V`.
+
+## Main definitions
+
+* `Digraph` is a structure for the relation
```
semorrison, line 17: This bullet point doesn't convey much to me. You can't say "the relation" without the prior text giving some indication which relation!

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,252 @@
...
+
+* `CompleteAtomicBooleanAlgebra` instance: Under the subgraph relation, `Digraph` forms a
+  `CompleteAtomicBooleanAlgebra`. In other words, this is the complete lattice of spanning subgraphs
```
semorrison, line 20: Really "spanning" here? The empty digra

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,252 @@
...
+  of the complete graph.
+
+## Todo
+
+* The implementation of digraphs is currently incomplete. It was originally created by Kyle Miller
```
semorrison, line 25: I'd just remove this TODO. It's not actionable.

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,252 @@
...
+
+open Finset Function
+
+universe u v w
+
+/-- A directed graph is a relation `Adj` on a vertex type `V`.
+The relation describes which pairs of vertices are adjacent.
+-/
+@[ext]
+structure Digraph (V : Type u) where
```
semorrison, line 40: Is this the same as Quiver?

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,252 @@
...
+
+/-- The complete graph on a type `V` is the digraph with all pairs of vertices
+adjacent. In `Mathlib`, this is usually referred to as `⊤`. Note that all vertices are adjacent to
+themselves. -/
+def Digraph.completeGraph (V : Type u) : Digraph V where Adj := ⊤
+
+/-- The graph with no edges on a given vertex type `V`. `Mathlib` prefers the notation `⊥`. -/
+def Digraph.emptyGraph (V : Type u) : Digraph V where Adj _ _ := False
+
+/-- Two vertices are adjacent in the complete bipartite graph on two vertex types
```
semorrison, line 67: Say digraph rather than graph in this doc string? And perhaps below.

semorrison:
Otherwise looks good!

Mathlib/Combinatorics/Digraph/Basic.lean
kmill: Yes, and it's not a class and not meant to be used as a class.

The main function of `Digraph` is to use graph theory terminology for a basic concept.

Mathlib/Combinatorics/Digraph/Basic.lean
kmill: It does because it's only empty in the sense of having no edges — it still has all the vertices `V`.

semorrison: Oops, misled by the fact that spanning trees are connected by virtue of being trees, I was thinking that a spanning subgraph should be connected. 

Sorry for the noise!

Mathlib/Combinatorics/Digraph/Basic.lean
semorrison: Possibly we could have a comment somewhere to this effect, but it's not essential.

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,236 @@
...
+Authors: Kyle Miller, Jack Cheverton, Jeremy Tan
+-/
+import Mathlib.Order.CompleteBooleanAlgebra
+import Mathlib.Data.Fintype.Pi
+
+/-!
+# Digraphs
+
+This module defines directed graphs on a vertex type `V`. Although the structure is definitionally
+equivalent to a `Quiver` in a specialized universe, the intention is to consider graph theory
```
jcommelin, line 13: It's not defeq to `Quiver` in Lean, right? Because it's a brand new structure.
I agree that it is equivalent in the `Equiv` sense.

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,236 @@
...
+
+/-- A digraph is a relation `Adj` on a vertex type `V`.
+The relation describes which pairs of vertices are adjacent. -/
+@[ext]
+structure Digraph (V : Type*) where
+  /-- The adjacency relation of a digraph. -/
+  Adj : V → V → Prop
+
+noncomputable instance {V : Type*} [Fintype V] : Fintype (Digraph V) := by
+  classical exact Fintype.ofInjective Digraph.Adj (fun _ _ ↦ Digraph.ext)
```
Command-Master, line 36: Can't you use `Fintype.ofSurjective` here to get this computabily?

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,236 @@
...
+theorem adj_injective : Injective (Adj : Digraph V → V → V → Prop) :=
+  fun _ _ ↦ Digraph.ext
+
+@[simp]
+theorem adj_inj {G H : Digraph V} : G.Adj = H.Adj ↔ G = H :=
+  adj_injective.eq_iff
+
+section Order
+
+/-- The relation that one `Digraph` is a subgraph of another.
```
Command-Master, line 76: Perhaps the docstring should mention that this is a spanning subgraph, not any subgraph

Mathlib/Combinatorics/Digraph/Basic.lean
kmill: How about "structurally equivalent"?

The comment about "not intended to be used as a class" is true, but it might give the wrong impression. The purpose of it not being a class is to be able to work with the whole lattice of digraphs on a given vertex type. Classes work best when there's the assumption that there's *the* digraph structure on the vertex type.

The more apt comparison by the way is to a plain relation. I'd start it out with something like "This module defines directed graphs on a vertex type `V`, which is simply the notion of a relation `V -> V -> Prop`. The intention is to develop the notion using the language of graph theory." Then in the next paragraph it could mention the relation to quivers.

This will also help graph theorists understand this module, since many have never heard of quivers before.

Mathlib/Combinatorics/Digraph/Basic.lean
kmill: Move this after `mk'` and then you can use it with `Fintype.ofBijective` (it saves needing that `Digraph V` have `DecidableEq`):
```lean
instance {V : Type*} [DecidableEq V] [Fintype V] : Fintype (Digraph V) :=
  Fintype.ofBijective Digraph.mk' <| by
    refine ⟨Embedding.injective _, ?_⟩
    classical
    intro G
    use fun v w ↦ G.Adj v w
    ext v w
    simp
```

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,235 @@
...
+import Mathlib.Order.CompleteBooleanAlgebra
+import Mathlib.Data.Fintype.Pi
+
+/-!
+# Digraphs
+
+This module defines directed graphs on a vertex type `V`,
+which is the same notion as a relation `V → V → Prop`.
+While this might be too simple of a notion to deserve the grandeur of a new definition,
+the intention here is to develop relations using the lanugage of graph theory.
```
Rida-Hamadani, line 15:
```suggestion
the intention here is to develop relations using the language of graph theory.
```

Mathlib/Combinatorics/Digraph/Basic.lean
```diff
@@ -0,0 +1,235 @@
...
+theorem sdiff_adj (x y : Digraph V) (v w : V) : (x \ y).Adj v w ↔ x.Adj v w ∧ ¬y.Adj v w := Iff.rfl
+
+instance supSet : SupSet (Digraph V) where
+  sSup s := { Adj := fun a b ↦ ∃ G ∈ s, Adj G a b }
+
+instance infSet : InfSet (Digraph V) where
+  sInf s := { Adj := fun a b ↦ (∀ ⦃G⦄, G ∈ s → Adj G a b) }
+
+@[simp]
+theorem sSup_adj {s : Set (Digraph V)} {a b : V} : (sSup s).Adj a b ↔ ∃ G ∈ s, Adj G a b := Iff.rfl
```
Rida-Hamadani, line 150: Seems weird to me that `sSup_adj` lemma (explicitly) takes `{a b : V}` while the others don't
```suggestion
theorem sSup_adj {s : Set (Digraph V)} : (sSup s).Adj a b ↔ ∃ G ∈ s, Adj G a b := Iff.rfl
```
Or adding the parameter to the other lemmas

jcommelin (approved):
lgtm

b-mehta (approved):
bors merge

I'd love to see more theory developed for these!

PR #14100, opened 2024-06-24
[Merged by Bors] - refactor: enforce the predicate on the `ContinuousFunctionalCalculus` is nontrivial

j-loreaux:
add a condition to the unital and non-unital continuous functional calculus class which ensures that the predicate is nontrivial by requiring `p 0` to hold. This guarantees in unital algebras that `p (algebraMap R A r)` holds, which is about all one can expect given that sometimes `p := IsStarNormal`.

---

see #14065 for motivation


j-loreaux (commented):
@adedecker I know you're busy with other things, but I would appreciate it if you could take a few minutes to review this. Frederic and I talked about it and think it makes sense, but I wanted to make sure you agree.

Mathlib/Topology/ContinuousFunction/FunctionalCalculus.lean
```diff
@@ -343,9 +349,21 @@ lemma cfc_id' : cfc (fun x : R ↦ x) a = a := cfc_id R a
 lemma cfc_map_spectrum : spectrum R (cfc f a) = f '' spectrum R a := by
   simp [cfc_apply f a, cfcHom_map_spectrum (p := p)]
 
+lemma cfc_const (r : R) (a : A) (ha : p a := by cfc_tac) :
+    cfc (fun _ ↦ r) a = algebraMap R A r := by
+  rw [cfc_apply (fun _ : R ↦ r) a, ← AlgHomClass.commutes (cfcHom ha (p := p)) r]
+  congr
+
 lemma cfc_predicate : p (cfc f a) :=
   cfc_apply f a ▸ cfcHom_predicate (A := A) ha _
```
dupuisf, line 359: Can you remove the `p a` and continuity conditions on this one, now that we know that `p` is true on the junk value? (And the same for the non-unital case)

dupuisf (approved):

dupuisf:
@ADedecker I understand you're a bit busy at the moment, so I'll just go ahead and merge this.

bors r+

PR #14107, opened 2024-06-24
[Merged by Bors] - feat(SpecialFunctions/Log): Add extended exp

LorenzoLuccioli:
We define an extended version of the exponential from `EReal` to `ENNReal` and prove some properties in relation to the extended log.

- [x] Rename `ENNReal.lean` to `ENNRealLog.lean`
- [x] Add `ENNRealExp.lean`
- [x] Add `ENNRealLogExp.lean`

See [this](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Porting.20TopologicalEntropy.20into.20Mathlib/near/446665481) Zulip discussion.

Co-authored-by: @RemyDegenne @pitmonticone @D-Thomine 

---


Mathlib/Analysis/SpecialFunctions/Log/ENNRealExp.lean
```diff
@@ -0,0 +1,104 @@
...
+  | ⊤ => ∞
+  | (x : ℝ) => ENNReal.ofReal (Real.exp x)
+
+@[simp] lemma exp_bot : exp ⊥ = 0 := rfl
+@[simp] lemma exp_zero : exp 0 = 1 := by simp [exp]
+@[simp] lemma exp_top : exp ⊤ = ∞ := rfl
+@[simp] lemma exp_coe (x : ℝ) : exp x = ENNReal.ofReal (Real.exp x) := rfl
+
+@[simp] lemma exp_eq_zero_iff {x : EReal} : exp x = 0 ↔ x = ⊥ := by
+  induction x using EReal.rec <;> simp [Real.exp_pos]
```
sgouezel, line 48:
```suggestion
  induction x <;> simp [Real.exp_pos]
```
Same thing below.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,170 @@
...
+/-!
+# Properties of the extended logarithm and exponential
+
+We prove that `log` and `exp` define an order isomorphism between `ℝ≥0∞` and `EReal`.
+
+## Main Definitions
+- `ENNReal.logOrderIso`: The order isomorphism between `ℝ≥0∞` and `EReal` defined by `log`
+and `exp`.
+- `ENNReal.log_homeomorph`: `log` as a homeomorphism.
+- `ENNReal.exp_homeomorph`: `exp` as a homeomorphism.
```
sgouezel, line 19:
```suggestion
- `EReal.exp_homeomorph`: `exp` as a homeomorphism.
```
Also, it feels weird to have the order_iso version of the log but not of the exponential.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,170 @@
...
+Authors: Damien Thomine, Pietro Monticone, Rémy Degenne, Lorenzo Luccioli
+-/
+import Mathlib.Analysis.SpecialFunctions.Log.ENNRealExp
+import Mathlib.Analysis.SpecialFunctions.Log.ENNRealLog
+import Mathlib.MeasureTheory.Constructions.BorelSpace.Basic
+
+/-!
+# Properties of the extended logarithm and exponential
+
+We prove that `log` and `exp` define an order isomorphism between `ℝ≥0∞` and `EReal`.
```
sgouezel, line 13:
```suggestion
We prove that `log` and `exp` define order isomorphisms between `ℝ≥0∞` and `EReal`.
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,170 @@
...
+
+@[simp] lemma logOrderIso_apply (x : ℝ≥0∞) : logOrderIso x = log x := rfl
+@[simp] lemma logOrderIso_symm_apply (x : EReal) : logOrderIso.symm x = exp x := rfl
+
+end OrderIso
+
+section Continuity
+
+/-- `log` as a homeomorphism. -/
+noncomputable def log_homeomorph : ℝ≥0∞ ≃ₜ EReal := logOrderIso.toHomeomorph
```
sgouezel, line 130:
```suggestion
noncomputable def logHomeomorph : ℝ≥0∞ ≃ₜ EReal := logOrderIso.toHomeomorph
```
This is probably a better name, as this is a single unit, so no underscore in it. No need to deprecate the previous version as it was introduced very recently. Same thing for the exponential.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,170 @@
...
+@[measurability, fun_prop]
+lemma _root_.Measurable.ereal_exp {α : Type*} {_ : MeasurableSpace α}
+    {f : α → EReal} (hf : Measurable f) :
+    Measurable fun x ↦ exp (f x) := measurable_exp.comp hf
+
+end Measurability
+
+end ENNReal
+
+instance : TopologicalSpace.MetrizableSpace EReal :=
```
sgouezel, line 169: Could you even register a polish space instance?

pitmonticone: We are not sure how to do it. @LorenzoLuccioli

LorenzoLuccioli: I just committed an attempt at adding the `PolishSpace` instance: 
https://github.com/leanprover-community/mathlib4/blob/84f997b1aafbf27bda3206e2ff634b5b633b1ff3/Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean#L242-L249

I was missing the instance `(uniformity EReal).IsCountablyGenerated` and I tried to prove it adding the `PseudoEMetricSpace` instance to `EReal`. The problem now is that Lean does not recognize that the `UniformSpace` structure induced by the `EDist` and the one induced by the topology are the same, do you know how to prove this?

Moreover I'm not sure whether my implementation of `EDist` satisfies the best practices, and I think it should be moved to some other file, I just left it here temporarily.

LorenzoLuccioli: I solved the problem with the help of @RemyDegenne, proving that `ENNReal` is a `PolishSpace` since it is isomorphic to a real interval, then we recover the property for `EReal` through the new isomorphism with `ENNReal`. Since it was not needed anymore I erased the definition of `EDist`.

Now I just have to figure out where to position the `PolishSpace` instance for `ENNReal`. 
We cannot put it in the ENNReal [file](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Instances/EReal.html), since it is imported by the file where `PolishSpace` is defined.
We could leave it here, but it seems weird to do that, since it does not really use anything about the log or exp.
`find_home` suggests to put it in the Polish spaces [file](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/MetricSpace/Polish.html), this would avoid any additional import, but in that file there are no instances of specific types to be Polish spaces.
Do you have any suggestions?

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
LorenzoLuccioli: `logOrderIso` actually contains the order isomorphism version of the exponential as the inverse function:
https://github.com/leanprover-community/mathlib4/blob/7ba4a1a7efa52aa3873d0ba28eb391c73280e7ea/Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean#L115-L120

Should we rename it to something like `logExpOrderIso` so it is more evident? Or do you mean that we should actually define an `expOrderIso` as a separate structure (or maybe as `logOrderIso.symm`)?

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
sgouezel: I think it would make sense to put it in the file defining PolishSpace, since it can not be put earlier.

Note that if you register the PolishSpace instance, you can remove the metrizable one as it follows from it.

Mathlib/Topology/MetricSpace/Polish.lean
```diff
@@ -229,6 +227,14 @@ theorem exists_polishSpace_forall_le {ι : Type*} [Countable ι] [t : Topologica
     .iInf ⟨none, Option.forall.2 ⟨le_rfl, hm⟩⟩ <| Option.forall.2 ⟨p, h'm⟩⟩
 #align polish_space.exists_polish_space_forall_le PolishSpace.exists_polishSpace_forall_le
 
+instance : PolishSpace ENNReal :=
+  ClosedEmbedding.polishSpace ⟨ENNReal.orderIsoUnitIntervalBirational.toHomeomorph.embedding,
+    ENNReal.orderIsoUnitIntervalBirational.range_eq ▸ isClosed_univ⟩
+
+instance : PolishSpace EReal :=
```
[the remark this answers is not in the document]
pitmonticone: @sgouezel Which EReal file? Don't we get build cycle error?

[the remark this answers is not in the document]
pitmonticone: Ah, ok. That's perfect.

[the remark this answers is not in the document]
pitmonticone: Done. Thanks for your help.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,182 @@
...
+
+## Tags
+ENNReal, EReal, logarithm, exponential
+-/
+
+open EReal ENNReal
+section LogExp
+namespace EReal
+
+@[simp] lemma log_exp (x : EReal) : log (exp x) = x := by
```
sgouezel, line 35:
```suggestion
@[simp] lemma EReal.log_exp (x : EReal) : log (exp x) = x := by
```
No need to open a namespace for a single lemma. Same thing for the next lemma.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,182 @@
...
+lemma exp_mul (x : EReal) (y : ℝ) : exp (x * y) = (exp x) ^ y := by
+  rw [← log_eq_iff, log_rpow, log_exp, log_exp, mul_comm]
+
+end EReal
+end Exp
+
+namespace ENNReal
+section OrderIso
+
+/-- `ENNReal.log` and its inverse `ENNReal.exp` are an order isomorphism between `ℝ≥0∞` and
```
sgouezel, line 119:
```suggestion
/-- `ENNReal.log` and its inverse `EReal.exp` are an order isomorphism between `ℝ≥0∞` and
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,182 @@
...
+
+end Continuity
+
+section Measurability
+
+@[measurability, fun_prop]
+lemma measurable_log : Measurable log := continuous_log.measurable
+
+@[measurability, fun_prop]
+lemma measurable_exp : Measurable exp := continuous_exp.measurable
```
sgouezel, line 164:
```suggestion
lemma _root_.EReal.measurable_exp : Measurable exp := continuous_exp.measurable
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,182 @@
...
+def logOrderIso : ℝ≥0∞ ≃o EReal where
+  toFun := log
+  invFun := exp
+  left_inv x := exp_log x
+  right_inv x := log_exp x
+  map_rel_iff' := by simp only [Equiv.coe_fn_mk, log_le_log_iff, forall_const]
+
+@[simp] lemma logOrderIso_apply (x : ℝ≥0∞) : logOrderIso x = log x := rfl
+@[simp] lemma logOrderIso_symm_apply (x : EReal) : logOrderIso.symm x = exp x := rfl
+
```
sgouezel, line 131: could you also add `expOrderIso` in the right namespace? (Replacing the lemma `logOrderIso_symm_apply` with a lemma `logOrderIso_symm`, and adding suitable lemmas).

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+/-!
+# Properties of the extended logarithm and exponential
+
+We prove that `log` and `exp` define order isomorphisms between `ℝ≥0∞` and `EReal`.
+## Main Definitions
+- `ENNReal.logOrderIso`: The order isomorphism between `ℝ≥0∞` and `EReal` defined by `log`
+and `exp`.
+- `ENNReal.logHomeomorph`: `log` as a homeomorphism.
+- `EReal.expHomeomorph`: `exp` as a homeomorphism.
```
sgouezel, line 19: Mention `EReal.expOrderIso` around here?

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+
+## Main Results
+- `ENNReal.log_exp`, `ENNReal.exp_log`: `log` and `exp` are inverses of each other.
```
sgouezel, line 22:
```suggestion
- `EReal.log_exp`, `ENNReal.exp_log`: `log` and `exp` are inverses of each other.
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+- `ENNReal.exp_nmul`, `ENNReal.exp_mul`: `exp` satisfies the identities `exp (n * x) = (exp x) ^ n`
```
sgouezel, line 23:
```suggestion
- `EReal.exp_nmul`, `EReal.exp_mul`: `exp` satisfies the identities `exp (n * x) = (exp x) ^ n`
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+and `exp (x * y) = (exp x) ^ y`.
+- `instMetrizableSpaceEReal`: `EReal` is a metrizable space.
```
sgouezel, line 25:
```suggestion
- `EReal` is a Polish space.
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+end LogExp
+
+section Log
+namespace ENNReal
+
+@[simp]
+lemma log_lt_log_iff {x y : ℝ≥0∞} : log x < log y ↔ x < y := by
+  refine ⟨fun h ↦ ?_, fun h ↦ log_strictMono h⟩
+  rw [← exp_log x, ← exp_log y]
+  exact exp_strictMono h
```
sgouezel, line 57: This is true for any strict mono function, so it has to be a one-liner. Indeed:
```suggestion
lemma log_lt_log_iff {x y : ℝ≥0∞} : log x < log y ↔ x < y := log_strictMono.lt_iff_lt
```

But then this lemma and the lemmas that follow don't use the exponential, and should probably go in the file about log.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+
+@[simp] lemma log_lt_zero_iff {x : ℝ≥0∞} : log x < 0 ↔ x < 1 := log_one ▸ @log_lt_log_iff x 1
+
+@[simp] lemma zero_lt_log_iff {x : ℝ≥0∞} : 0 < log x ↔ 1 < x := log_one ▸ @log_lt_log_iff 1 x
+
+@[simp]
+lemma log_le_log_iff {x y : ℝ≥0∞} : log x ≤ log y ↔ x ≤ y := by
+  refine ⟨fun h ↦ ?_, fun h ↦ log_monotone h⟩
+  rw [← exp_log x, ← exp_log y]
+  exact exp_monotone h
```
sgouezel, line 71:
```suggestion
lemma log_le_log_iff {x y : ℝ≥0∞} : log x ≤ log y ↔ x ≤ y := log_strictMono.le_iff_le
```

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,186 @@
...
+@[simp] lemma zero_le_log_iff {x : ℝ≥0∞} : 0 ≤ log x ↔ 1 ≤ x := log_one ▸ @log_le_log_iff 1 x
+
+end ENNReal
+end Log
+
+section Exp
+namespace EReal
+@[simp]
+lemma exp_lt_exp_iff {a b : EReal} : exp a < exp b ↔ a < b := by
+  conv_rhs => rw [← log_exp a, ← log_exp b, log_lt_log_iff]
```
sgouezel, line 84: instead of this, you can use `exp_strictMono`, like I did above. And then move this and the following lemmas to the file solely on exp.

Mathlib/Analysis/SpecialFunctions/Log/ENNRealLogExp.lean
```diff
@@ -0,0 +1,140 @@
...
+section Measurability
+
+@[measurability, fun_prop]
+lemma measurable_log : Measurable log := continuous_log.measurable
+
+@[measurability, fun_prop]
+lemma _root_.EReal.measurable_exp : Measurable exp := continuous_exp.measurable
+
+@[measurability, fun_prop]
+lemma _root_.Measurable.ereal_log {α : Type*} {_ : MeasurableSpace α}
```
sgouezel, line 125:
```suggestion
lemma _root_.Measurable.ennreal_log {α : Type*} {_ : MeasurableSpace α}
```

sgouezel (commented):
bors d+
Thanks!

PR #2603, opened 2023-03-03
[Merged by Bors] - feat: port Algebra.Star.StarAlgHom

ADedecker:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14110, opened 2024-06-25
[Merged by Bors] - refactor(Algebra/Star/StarAlgHom): Reverting {_ : ...} to [...]

mans0954:
During porting (#2603) parameters to some instances were made implicit to avoid [dangerous instances](https://github.com/leanprover-community/mathlib4/pull/2603/files#diff-8017bfdc9f21755fb88462d9ba5909d1590cfe91939577ce589ef704a831df57R700). Is this still necessary?

---


mans0954:
!bench

grunweg:
Just for the record: the dangerous instance linter has not been ported from mathlib to mathlib4 yet. (So, someone who understands this issue should look at it; linters alone would not catch such issues right now. I don't understand this question well enough to comment.)

eric-wieser (approved):
LGTM, this doesn't look related to dangerous instances to me.

j-loreaux:
Right, it wasn't about dangerous instances, but rather the order in which arguments were synthesized. I think Gabriel came up with a solution during porting. Given that the bench doesn't show any significant change, I think it's fine to change them back.

bors merge

PR #14117, opened 2024-06-25
[Merged by Bors] - chore: downgrade priority of `SetLike.smul` instances 

mattrobball:
Searching through the `SetLike` hierarchy when we want specific data like `SMul` seems like something to avoided unless necessary. The current priorities of `900` seem to too high (recall default is `1000` and our low value is `100`). We change the values to `50`.

---


mattrobball:
!bench

j-loreaux:
@mattrobball do you still want this change, or do you not care anymore since there was no benefit?

mattrobball:
I still think the new value is more reasoned.

j-loreaux:
bors merge

PR #14120, opened 2024-06-25
[Merged by Bors] - chore (LinearAlgebra.Dimension): make `instNontrivial` a local instance 

mattrobball:
The universes are a bit strange in this since the only occurrence of one is in the `HasRankNullity` class. Since this isn't needed outside this file, we should probably make it `local`.

---


semorrison:
bors merge

PR #14121, opened 2024-06-25
[Merged by Bors] - chore (Subgroup.Order): move ordered facts out of `Basic` and into `Order`

mattrobball:
We don't want to import the order hierarchy without really needing it.

---


mattrobball:
!bench

kbuzzard:
Very nice!

bors merge

PR #14122, opened 2024-06-25
[Merged by Bors] - chore (Ring.Prod): move order instances on product rings

mattrobball:
We shouldn't import the whole order hierarchy when we want ring instances on products.

---


mattrobball:
!bench

kbuzzard:
I feel like you're slowly unpicking a gigantic knot here.

bors merge

PR #14127, opened 2024-06-25
feat(CategoryTheory): monoidal structure on the category of presheaves of modules

dagurtomas:

---


joelriou:
After the refactor #16667, I have rewritten this PR as #17300. Performance issues have disappeared in the process.

PR #14129, opened 2024-06-25
[Merged by Bors] - chore (Set.Pointwise.Basic): avoid importing ordered rings 

mattrobball:
We shouldn't need to import ordered algebraic classes for basic properties of pointwise multiplication of sets.

---


semorrison:
bors merge

PR #11147, opened 2024-03-04
[Merged by Bors] - feat(Order/KrullDimension): add some basic lemmas about krull dimension of a preorder

jjaassoonn:
- krull dimension, height, coheight interacting with functions
- order dual

These will be specialized to ring theoretic krull dimension later

---

- [x] depends on:#11730

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14134, opened 2024-06-25
[Merged by Bors] - feat(RingTheory/Ideal/KrullDimension): define the Krull dimension of a ring

FMLJohn:
The ring theoretic krull dimension is the order theoretic krull dimension applied to its prime spectrum. Unfolding this definition, it is the length of longest series of prime ideals ordered by inclusion.

---

@mariainesdff

- [x] depends on: #11147


Mathlib/Order/KrullDimension.lean
```diff
@@ -27,6 +32,17 @@ sum of `-∞` and the Krull dimension of any other varieties.
...
 
 variable (α : Type*) [Preorder α]
+variable {β : Type _} (r : Rel β β)
+
+/--
+Krull dimension of a set `α` with a binary relation `r` is the supremum of the rightmost index of
+all relation series of `r`. If there is no relation series `a₀, a₁, ..., aₙ` in `α`, then its Krull
+dimension is defined to be negative infinity; if the length of all relation series `a₀, a₁, ..., aₙ`
+is unbounded, its Krull dimension is defined to be positive infinity. (Not sure if this definition
+is useful.)
```
mariainesdff, line 42: Based on the above comment, can I ask why you include this definition?

Mathlib/Order/KrullDimension.lean
```diff
@@ -46,3 +64,108 @@ noncomputable def height (a : α) : WithBot (WithTop ℕ) := krullDim (Set.Iic a
...
+
+lemma eq_zero_of_unique [Unique α] : krullDim α = 0 := by
+  rw [eq_len_of_finiteDimensionalType (α := α), Nat.cast_eq_zero]
+  refine (LTSeries.longestOf_len_unique (default : LTSeries α) fun q ↦ show _ ≤ 0 from ?_).symm
+  by_contra r
+  rw [not_le] at r
+  exact ne_of_lt (q.step ⟨0, r⟩) <| Subsingleton.elim _ _
+
+lemma le_of_strictMono (f : α → β) (hf : StrictMono f) : krullDim α ≤ krullDim β := by
+  refine krullDimOfRel.le_of_map (· < ·) f hf
```
mariainesdff, line 141: Do you really need `refine` here?

Mathlib/Order/KrullDimension.lean
```diff
@@ -46,3 +64,108 @@ noncomputable def height (a : α) : WithBot (WithTop ℕ) := krullDim (Set.Iic a
...
+lemma le_of_strictComono_and_surj
+    (f : α → β) (hf : StrictComono f) (hf' : Function.Surjective f) :
+    krullDim β ≤ krullDim α :=
+iSup_le $ λ p ↦ le_sSup ⟨p.comap _ hf hf', rfl⟩
+
+lemma eq_of_orderIso (f : α ≃o β) : krullDim α = krullDim β :=
+  krullDimOfRel.eq_of_relIso (· < ·) ⟨f, fun {_ _} => f.lt_iff_lt⟩
+
+lemma eq_iSup_height : krullDim α = ⨆ (a : α), height α a := by
+{ refine' le_antisymm (iSup_le $ λ i ↦ le_iSup_of_le (i ⟨i.length, lt_add_one _⟩)
```
mariainesdff, line 155: I have been warned that `refine'` should be replaced by `refine`.

Mathlib/Order/KrullDimension.lean
jjaassoonn: `krullDimOfRel` seems not to be a very useful definition. The only other relation I can think of that we want to consider Krull dimension of is `covby` for submodules. The length of a module can be defined as krull dimension of the poset submodules; the length of a composition series; the krull dimension of `covby` relation. But the last is conceptually the same as length of the composition series. So probably `krullDimOfRel` is not too useful a definition so I think we should remove it.

jjaassoonn:
Note that the basic part of this PR is actually #11147

Mathlib/Order/KrullDimension.lean
mariainesdff: Thanks; then I agree that this could be deleted.

mariainesdff:
> Note that the basic part of this PR is actually #11147

Ah thanks, I didn't know about that one. I will list it as a dependency.

alreadydone (commented):
Repeating my comment on Zoom: I think it would be nice to add some examples, including Subsingleton (= ⊥), Nontrivial (> ⊥), Field (= 0), IsArtinianRing (≤ 0), Ring.DimensionLEOne (≤ 1), IsDomain + ¬ IsField (≥ 1), IsNoetherianRing (< ⊤).

And you might state [these](https://math.stackexchange.com/questions/4368191/bound-for-the-krull-dimension-of-a-polynomial-ring) [two](https://math.stackexchange.com/questions/1266827/krull-dimension-of-polynomial-rings-over-noetherian-rings) results about Polynomial as proof_wanted.

FMLJohn:
> Repeating my comment on Zoom: I think it would be nice to add some examples, including Subsingleton (= ⊥), Nontrivial (> ⊥), Field (= 0), IsArtinianRing (≤ 0), Ring.DimensionLEOne (≤ 1), IsDomain + ¬ IsField (≥ 1), IsNoetherianRing (< ⊤).
> 
> And you might state [these](https://math.stackexchange.com/questions/4368191/bound-for-the-krull-dimension-of-a-polynomial-ring) [two](https://math.stackexchange.com/questions/1266827/krull-dimension-of-polynomial-rings-over-noetherian-rings) results about Polynomial as proof_wanted.

I have added the result that the ring theoretic Krull dimension of a commutative ring is equal to the topological Krull dimension of its prime spectrum. I have also included the lemma that the Krull dimension of a field is zero. For other results, including those on Artinian rings and Noetherian rings, I have noticed that they are a bit long. Will it be appropriate to push them in some other pull request?

Also, how shall I mark results as proof_wanted? Shall I add some comments into the lean file? For example, "we still need the proofs the following results ..."?

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+-/
+noncomputable abbrev ringKrullDim (R : Type _) [CommRing R] : WithBot (WithTop ℕ) :=
+  krullDim (PrimeSpectrum R)
+
+namespace ringKrullDim
+
+open PrimeSpectrum OrderDual in
+theorem eq_topologicalKrullDim (R : Type _) [CommRing R] :
+    ringKrullDim R = topologicalKrullDim (PrimeSpectrum R) :=
+  Eq.symm $ krullDim_orderDual.symm.trans $ krullDim_eq_of_orderIso $ OrderIso.symm {
```
erdOne, line 38: Is `PrimeSpectrum.pointsEquivIrreducibleCloseds` relevant here?

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+      simp [zeroLocus_subset_zeroLocus_iff, q.isPrime.radical] }
+
+/-- If `f : R →+* S` is surjective, then `ringKrullDim S ≤ ringKrullDim R`. -/
+theorem le_of_surj (R S : Type _) [CommRing R] [CommRing S] (f : R →+* S)
+    (hf : Function.Surjective f) : ringKrullDim S ≤ ringKrullDim R :=
+  krullDim_le_of_strictMono (PrimeSpectrum.comap f) (Monotone.strictMono_of_injective
+    (fun _ _ hab ↦ Ideal.comap_mono hab) (PrimeSpectrum.comap_injective_of_surjective f hf))
+
+/-- If `I` is an ideal of `R`, then `ringKrullDim (R ⧸ I) ≤ ringKrullDim R`. -/
+theorem le_of_quot (R : Type _) [CommRing R] (I : Ideal R) :
```
erdOne, line 58:
```suggestion
theorem quot_le (R : Type _) [CommRing R] (I : Ideal R) :
```

Mathlib/Order/Monotone/Basic.lean
```diff
@@ -98,6 +98,12 @@ def StrictMono (f : α → β) : Prop :=
   ∀ ⦃a b⦄, a < b → f a < f b
 #align strict_mono StrictMono
 
+/--
+A function `f : α → β` is said to be strictly comonotonic (dual to strictly monotonic)
+if and only if `a < b` is implied by `f a < f b` for all `a, b : β`.
+-/
+def StrictComono (f : α → β) : Prop := ∀ ⦃a b⦄, f a < f b → a < b
+
```
jjaassoonn, line 106: Is this still relevant?

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+3. the Krull dimension of any quotient ring of a commutative ring must be less than or equal to
+   the Krull dimension of that commutative ring;
+4. two isomorphic commutative rings have the same Krull dimension;
+5. the Krull dimension of a field is zero.
+-/
+
+/--
+The ring theoretic Krull dimension is the Krull dimension of its spectrum ordered by inclusion.
+-/
+noncomputable abbrev ringKrullDim (R : Type _) [CommRing R] : WithBot (WithTop ℕ) :=
```
jjaassoonn, line 30:
```suggestion
noncomputable abbrev ringKrullDim (R : Type*) [CommRing R] : WithBot (WithTop ℕ) :=
```

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+      simp_rw [vanishingIdeal_zeroLocus_eq_radical, p.isPrime.radical]
+    right_inv := fun s ↦ by
+      simp only [zeroLocus_vanishingIdeal_eq_closure, s.3.closure_eq]; rfl
+    map_rel_iff' := by
+      intro p q
+      change zeroLocus _ ≤ zeroLocus _ ↔ _
+      simp [zeroLocus_subset_zeroLocus_iff, q.isPrime.radical] }
+
+/-- If `f : R →+* S` is surjective, then `ringKrullDim S ≤ ringKrullDim R`. -/
+theorem le_of_surj (R S : Type _) [CommRing R] [CommRing S] (f : R →+* S)
```
jjaassoonn, line 52:
```suggestion
theorem le_of_surj {R S : Type*} [CommRing R] [CommRing S] (f : R →+* S)
```

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+  krullDim_le_of_strictMono (PrimeSpectrum.comap f) (Monotone.strictMono_of_injective
+    (fun _ _ hab ↦ Ideal.comap_mono hab) (PrimeSpectrum.comap_injective_of_surjective f hf))
+
+/-- If `I` is an ideal of `R`, then `ringKrullDim (R ⧸ I) ≤ ringKrullDim R`. -/
+theorem le_of_quot (R : Type _) [CommRing R] (I : Ideal R) :
+    ringKrullDim (R ⧸ I) ≤ ringKrullDim R :=
+  le_of_surj _ _ (Ideal.Quotient.mk I) Ideal.Quotient.mk_surjective
+
+/-- If `R` and `S` are isomorphic, then `ringKrullDim R = ringKrullDim S`. -/
+theorem eq_of_ringEquiv (R S : Type _) [CommRing R] [CommRing S] (e : R ≃+* S) :
```
jjaassoonn, line 63:
```suggestion
theorem eq_of_ringEquiv {R S : Type*} [CommRing R] [CommRing S] (e : R ≃+* S) :
```

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+    ringKrullDim R = ringKrullDim S :=
+  le_antisymm (le_of_surj S R (RingEquiv.symm e) (EquivLike.surjective (RingEquiv.symm e)))
+    (le_of_surj R S e (EquivLike.surjective e))
+
+section Field
+
+theorem eq_zero_of_field (F : Type _) [Field F] : ringKrullDim F = 0 :=
```
jjaassoonn, line 70:
```suggestion
theorem eq_zero_of_field (F : Type*) [Field F] : ringKrullDim F = 0 :=
```

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,78 @@
...
+  krullDim_eq_zero_of_unique
+
+theorem eq_zero_of_isField (F : Type _) [CommRing F] (hF : IsField F) : ringKrullDim F = 0 :=
```
jjaassoonn, line 73:
```suggestion
theorem eq_zero_of_isField (F : Type*) [CommRing F] (hF : IsField F) : ringKrullDim F = 0 :=
```

Mathlib/RingTheory/Ideal/KrullDimension.lean
FMLJohn: How about changing it to `quotient_le`? I feel that this is more suitable in terms of the Mathlib4 naming convention.

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -21,7 +21,9 @@ Key results in this file include:
 3. the Krull dimension of any quotient ring of a commutative ring must be less than or equal to
    the Krull dimension of that commutative ring;
 4. two isomorphic commutative rings have the same Krull dimension;
-5. the Krull dimension of a field is zero.
+5. for a commutative ring whose Krull dimension is zero, any prime ideal is maximal;
+6. for a commutative ring whose Krull dimension is zero, any prime ideal is minimal;
+7. the Krull dimension of a field is zero.
```
kbuzzard, line 26: This part of the module docstring should probably follow the conventions here https://leanprover-community.github.io/contribute/doc.html#header-comment , now there are so many key results. You should make a new `## Main statements` subheading and list the statements together with what the theorems are called in Lean. See e.g. `RingTheory/Ideal/IsPrincipal.lean` or many other mathlib files for the formatting you should use, ``* `theorem_name`: theorem_explanation``. Then it will display nicely in the online API docs.

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,100 @@
+/-
```
kbuzzard, line 1: I don't really see why this file is in the `Ideal` subdirectory, probably it should just be `RingTheory/KrullDimension.lean`.

Mathlib/RingTheory/Ideal/KrullDimension.lean
```diff
@@ -0,0 +1,100 @@
...
+7. the Krull dimension of a field is zero.
+-/
+
+/--
+The ring theoretic Krull dimension is the Krull dimension of its spectrum ordered by inclusion.
+-/
+noncomputable abbrev ringKrullDim (R : Type*) [CommRing R] : WithBot (WithTop ℕ) :=
+  krullDim (PrimeSpectrum R)
+
+namespace ringKrullDim
```
kbuzzard, line 35: I am a bit unclear about whether this is the right approach for naming. For example with `finprod` we just have loads of theorems whose names start with `finprod_` rather than a namespace `finprod`.

kbuzzard (changes requested):
Thanks! I left some comments. Re "proof wanted": search mathlib for `proof_wanted` to see what the suggestion means. It's `proof_wanted <statement>`. It would be very easy to add the results for `Subsingleton`  (= ⊥) and 
`Nontrivial` (> ⊥) in this PR.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,115 @@
...
+  @krullDim_eq_zero_of_unique _ _ <| @PrimeSpectrum.instUnique _ hF.toField
+
+end Field
+
+end ringKrullDim
+
+proof_wanted Polynomial.ringKrullDim_le (R : Type*) [CommRing R] :
+  ringKrullDim (Polynomial R) ≤ 2 * (ringKrullDim R) + 1
+
+proof_wanted MvPolynomial.fin_ringKrullDim_eq_plus_of_isNoetherianRing
```
jcommelin, line 113:
```suggestion
proof_wanted MvPolynomial.fin_ringKrullDim_eq_add_of_isNoetherianRing
```

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,115 @@
...
+The ring theoretic Krull dimension is the Krull dimension of its spectrum ordered by inclusion.
+-/
+noncomputable abbrev ringKrullDim (R : Type*) [CommRing R] : WithBot (WithTop ℕ) :=
+  krullDim (PrimeSpectrum R)
+
+lemma ringKrullDim_eq_bot_of_subsingleton (R : Type*) [CommRing R] [Subsingleton R] :
+    ringKrullDim R = ⊥ :=
+  krullDim_eq_bot_of_isEmpty
+
+lemma ringKrullDim_gt_bot_of_nontrivial (R : Type*) [CommRing R] [Nontrivial R] :
```
jcommelin, line 37: Why do you phrase this as `gt_bot` instead of `nonneg`?

jcommelin:
What happened with the idea of having a predicate `krullDimLT` that asserted that the krullDim of a given ring/space/... was LT a given natural number? Is there a discussion recorded somewhere with the pros and cons and the reason for going with a function valued in `WithTop (WithBot _)`?

FMLJohn:
> What happened with the idea of having a predicate `krullDimLT` that asserted that the krullDim of a given ring/space/... was LT a given natural number? Is there a discussion recorded somewhere with the pros and cons and the reason for going with a function valued in `WithTop (WithBot _)`?

No I have checked the streams on Zulip and I don't think there is any recorded discussion on that. The reason why we use `WithTop (WithBot _)` is that we eventually want `⊥ + ⊺ = ⊥`. It is true that for topological spaces `X` and `Y`, `dim(X × Y) = dim(X) + dim(Y)`. Now if `X` is empty and `Y` is infinite-dimensional, then `X × Y` is also empty. So we have `⊥ + ⊺ = dim(X) + dim(Y) = dim(X × Y) = ⊥`.

Mathlib/RingTheory/Ideal/KrullDimension.lean
FMLJohn: I have noticed that [RingTheory/Multiplicity](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Multiplicity.html#multiplicity) uses the same naming method as mine. So I don't think I need to change the way I name theorems here.

jjaassoonn:
I think Johan is referring to @kbuzzard's idea that instead of defining `krullDim`, we can introduce a family of predicate `krullDimLT`, then the negative infinity case is `krullDimLT 0`, positive infinity case is `\forall n, \not krullDimLT n` and krull dimension equal to `n` is `krullDimLT (n +1)` and `\not krullDimLT n`

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,113 @@
...
+* `ringKrullDim.eq_topologicalKrullDim`: the ring theoretic Krull dimension of a commutative ring
+is equal to the topological Krull dimension of its prime spectrum.
+
+* `ringKrullDim.eq_zero_of_field`: the Krull dimension of a field is zero.
+-/
+
+/--
+The ring theoretic Krull dimension is the Krull dimension of its spectrum ordered by inclusion.
+-/
+noncomputable abbrev ringKrullDim (R : Type*) [CommRing R] : WithBot (WithTop ℕ) :=
```
erdOne, line 30: Why is it an `abbrev`?

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,113 @@
...
+  krullDim (PrimeSpectrum R)
+
+namespace ringKrullDim
+
+lemma eq_bot_of_subsingleton (R : Type*) [CommRing R] [Subsingleton R] :
```
erdOne, line 35:
```suggestion
@[nontriviality]
lemma eq_bot_of_subsingleton (R : Type*) [CommRing R] [Subsingleton R] :
```

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,113 @@
...
+  · subst hIJ; exact hJ1
+  · let x : LTSeries (PrimeSpectrum R) := ⟨1, ![⟨I, hI⟩, ⟨J, hJ1.isPrime⟩],
+      fun i ↦ by fin_cases i; exact ⟨hJ2, fun r ↦ hIJ <| Eq.symm <| hJ1.eq_of_le hI.1 r⟩⟩
+    have : 1 ≤ ringKrullDim R :=
+      le_iSup (α := WithBot (WithTop ℕ)) (f := fun x : LTSeries _ ↦ x.length) x
+    rw [hdim] at this
+    norm_num at this
+
+theorem _root_.Ideal.IsPrime.mem_minimalPrimes_of_ringKrullDim_eq_zero
+    {I : Ideal R} (hI : I.IsPrime) (hdim : ringKrullDim R = 0) :
```
erdOne, line 82: There should be a lemma saying `krullDim a = 0 <-> forall x in a, x is maximal (minimal)`, then specialize.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,113 @@
...
+    have : 1 ≤ ringKrullDim R :=
+      le_iSup (α := WithBot (WithTop ℕ)) (f := fun x : LTSeries _ ↦ x.length) x
+    rw [hdim] at this
+    norm_num at this
+
+end DimensionZero
+
+section Field
+
+theorem eq_zero_of_field (F : Type*) [Field F] : ringKrullDim F = 0 :=
```
erdOne, line 98:
```suggestion
@[simp]
theorem eq_zero_of_field (F : Type*) [Field F] : ringKrullDim F = 0 :=
```

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,113 @@
...
+  krullDim_le_of_strictMono (PrimeSpectrum.comap f) (Monotone.strictMono_of_injective
+    (fun _ _ hab ↦ Ideal.comap_mono hab) (PrimeSpectrum.comap_injective_of_surjective f hf))
+
+/-- If `I` is an ideal of `R`, then `ringKrullDim (R ⧸ I) ≤ ringKrullDim R`. -/
+theorem quotient_le {R : Type*} [CommRing R] (I : Ideal R) :
+    ringKrullDim (R ⧸ I) ≤ ringKrullDim R :=
+  le_of_surjective _ Ideal.Quotient.mk_surjective
+
+/-- If `R` and `S` are isomorphic, then `ringKrullDim R = ringKrullDim S`. -/
+theorem eq_of_ringEquiv {R S : Type*} [CommRing R] [CommRing S] (e : R ≃+* S) :
```
erdOne, line 60: I would give this an alias `RingEquiv.ringKrullDim` to enable dot notation.

erdOne (commented):
Also please use `variable {R S : Type*} [CommRing R] [CommRing S]` at the top of the file.

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: I have changed it to `def`.

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: I have formalized the following lemmas:
```
lemma mem_maximals_of_krullDim_eq_zero {α : Type*} [Preorder α] :
    krullDim α = 0 → ∀ (a : α), a ∈ maximals (fun x x1 ↦ x ≤ x1) ⊤ := ...

lemma mem_minimals_of_krullDim_eq_zero {α : Type*} [Preorder α] :
    krullDim α = 0 → ∀ (a : α), a ∈ minimals (fun x x1 ↦ x ≤ x1) ⊤ := ...

lemma krullDim_eq_zero_iff_mem_maximals_of_nonempty {α : Type*} [Preorder α] [Nonempty α] :
    krullDim α = 0 ↔ ∀ a : α, a ∈ maximals (· ≤ ·) ⊤ := ...

lemma krullDim_eq_zero_iff_mem_minimals_of_nonempty {α : Type*} [Preorder α] [Nonempty α] :
    krullDim α = 0 ↔ ∀ a : α, a ∈ minimals (· ≤ ·) ⊤ := ...
```
Then I made consequent changes to the results
```
Ideal.IsPrime.isMaximal_of_ringKrullDim_eq_zero

Ideal.IsPrime.mem_minimalPrimes_of_ringKrullDim_eq_zero
```

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: I have written a new theorem `RingEquiv.ringKrullDim`, whose statement is exactly the same as that of `ringKrullDim_eq_of_ringEquiv`. Is it the correct way of creating an alias?

FMLJohn:
> Also please use `variable {R S : Type*} [CommRing R] [CommRing S]` at the top of the file.

Done!

FMLJohn:
> I think Johan is referring to @kbuzzard's idea that instead of defining `krullDim`, we can introduce a family of predicate `krullDimLT`, then the negative infinity case is `krullDimLT 0`, positive infinity case is `\forall n, \not krullDimLT n` and krull dimension equal to `n` is `krullDimLT (n +1)` and `\not krullDimLT n`

I have asked @kbuzzard recently and he said that both `WithBot (WithTop Nat)` and `krullDimLT` have advantages. We can define `krullDimLT` in another PR.

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: Could you please have a look and see if I have created the alias properly (i.e. have I put the alias in the correct place, etc.)? Thanks a lot!

Mathlib/Order/KrullDimension.lean
```diff
@@ -97,6 +98,82 @@ lemma krullDim_eq_zero_of_unique [Unique α] : krullDim α = 0 := by
   by_contra r
   exact ne_of_lt (q.step ⟨0, not_le.mp r⟩) <| Subsingleton.elim _ _
 
+lemma mem_maximals_of_krullDim_eq_zero :
```
erdOne, line 101: Golf of these
```

lemma LTSeries.length_le_krullDim (p : LTSeries α) : p.length ≤ krullDim α := le_sSup ⟨_, rfl⟩ 

lemma mem_maximals_of_krullDim_eq_zero (h : krullDim α = 0) (a : α) :
    a ∈ maximals (· ≤ ·) Set.univ := by
  simp only [maximals, Set.mem_univ, true_implies, true_and]
  intro b hab
  by_contra hba
  let x : LTSeries α := ⟨1, ![a, b], (by simpa [Unique.forall_iff] using lt_of_le_not_le hab hba)⟩
  simpa using WithBot.one_le_coe.mp (x.length_le_krullDim.trans_eq h)

lemma krullDim_eq_zero_iff_mem_maximals_of_nonempty [Nonempty α] :
    krullDim α = 0 ↔ ∀ a : α, a ∈ maximals (· ≤ ·) Set.univ := by
  refine ⟨mem_maximals_of_krullDim_eq_zero, fun h ↦ ?_⟩
  simp only [maximals, Set.mem_univ, true_implies, true_and] at h
  have (p : LTSeries α) : p.length = 0 := by
    by_contra hp
    have : NeZero p.length := ⟨hp⟩
    exact lt_irrefl _ ((h (p 0) (p.step 0).le).trans_lt (p.step 0))
  simp only [krullDim, this, Nat.cast_zero, ciSup_const]

lemma mem_minimals_of_krullDim_eq_zero (h : krullDim α = 0) (a : α) :
    a ∈ minimals (· ≤ ·) ⊤ :=
  mem_maximals_of_krullDim_eq_zero (krullDim_orderDual.trans h) a
  
lemma krullDim_eq_zero_iff_mem_minimals_of_nonempty [Nonempty α] :
    krullDim α = 0 ↔ ∀ a : α, a ∈ minimals (· ≤ ·) ⊤ := by
  rw [← krullDim_orderDual, krullDim_eq_zero_iff_mem_maximals_of_nonempty]
  rfl
```

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+    ringKrullDim R = ⊥ :=
+  krullDim_eq_bot_of_isEmpty
+
+lemma ringKrullDim_nonneg_of_nontrivial [Nontrivial R] :
+    0 ≤ ringKrullDim R :=
+  krullDim_nonneg_of_nonempty
+
+theorem ringKrullDim_eq_topologicalKrullDim :
+    ringKrullDim R = topologicalKrullDim (PrimeSpectrum R) :=
+  Eq.symm $ krullDim_orderDual.symm.trans $ krullDim_eq_of_orderIso
```
erdOne, line 46: We use `<|` and not `$` now.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+theorem ringKrullDim_eq_of_ringEquiv (e : R ≃+* S) :
+    ringKrullDim R = ringKrullDim S :=
+  le_antisymm (ringKrullDim_le_of_surjective e.symm e.symm.surjective)
+    (ringKrullDim_le_of_surjective e e.surjective)
+
+/-- An alias for `ringKrullDim_eq_of_ringEquiv`. -/
+theorem _root_.RingEquiv.ringKrullDim (e : R ≃+* S) :
+    ringKrullDim R = ringKrullDim S :=
+  le_antisymm (ringKrullDim_le_of_surjective e.symm e.symm.surjective)
+    (ringKrullDim_le_of_surjective e e.surjective)
```
erdOne, line 70:
```suggestion
alias RingEquiv.ringKrullDim := ringKrullDim_eq_of_ringEquiv
```
There is also this nice syntax.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+    (I : PrimeSpectrum R) (hdim : ringKrullDim R = 0) :
+    I.asIdeal.IsMaximal := by
+  have h := mem_maximals_of_krullDim_eq_zero hdim I
+  simp only [maximals, Set.top_eq_univ, Set.mem_univ, true_implies, true_and] at h
+  constructor; constructor
+  · exact Ideal.IsPrime.ne_top'
+  · intro J hIJ; by_contra hJ
+    obtain ⟨M, hM1, hM2⟩ := Ideal.exists_le_maximal J hJ
+    have := h $ show _ ≤ ⟨M, hM1.isPrime⟩ by exact (le_of_lt $ lt_of_lt_of_le hIJ hM2 : _ ≤ _)
+    exact (lt_self_iff_false _).1 $ lt_of_lt_of_le (lt_of_le_of_lt (by exact this) hIJ) hM2
```
erdOne, line 84: I think the next lemma is good enough.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+
+theorem _root_.Ideal.IsPrime.isMaximal_of_ringKrullDim_eq_zero
```
erdOne, line 86:
```suggestion
theorem IsPrime.isMaximal_of_ringKrullDim_eq_zero
```
We are already in the root namespace right?

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+
+theorem _root_.PrimeSpectrum.asIdeal.mem_minimalPrimes_of_ringKrullDim_eq_zero
+    (I : PrimeSpectrum R) (hdim : ringKrullDim R = 0) :
+    I.asIdeal ∈ minimalPrimes R := by
+  have h := mem_minimals_of_krullDim_eq_zero hdim I
+  simp only [minimals, Set.top_eq_univ, Set.mem_univ, true_implies, true_and] at h
+  exact ⟨⟨I.isPrime, OrderBot.bot_le I.asIdeal⟩, by
+    rintro J ⟨hJ, _⟩ hJI; exact h $ show ⟨J, hJ⟩ ≤ I by exact hJI⟩
+
+theorem _root_.Ideal.IsPrime.mem_minimalPrimes_of_ringKrullDim_eq_zero
```
erdOne, line 99:
```suggestion
theorem Ideal.IsPrime.mem_minimalPrimes_of_ringKrullDim_eq_zero
```

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+    obtain ⟨M, hM1, hM2⟩ := Ideal.exists_le_maximal J hJ
+    have := h $ show _ ≤ ⟨M, hM1.isPrime⟩ by exact (le_of_lt $ lt_of_lt_of_le hIJ hM2 : _ ≤ _)
+    exact (lt_self_iff_false _).1 $ lt_of_lt_of_le (lt_of_le_of_lt (by exact this) hIJ) hM2
+
+theorem _root_.Ideal.IsPrime.isMaximal_of_ringKrullDim_eq_zero
+    {I : Ideal R} (hI : I.IsPrime) (hdim : ringKrullDim R = 0) :
+    I.IsMaximal :=
+  PrimeSpectrum.asIdeal.isMaximal_of_ringKrullDim_eq_zero ⟨I, hI⟩ hdim
+
+theorem _root_.PrimeSpectrum.asIdeal.mem_minimalPrimes_of_ringKrullDim_eq_zero
```
erdOne, line 91: I think the next lemma is good enough.

erdOne: `T1Space (PrimeSpectrum R) ↔ ringKrullDim R = 0` is a fun lemma to have instead if you like something in terms of `PrimeSpectrum`. Maybe in a separate PR though.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+
+end DimensionZero
+
+section Field
+
+@[simp]
+theorem ringKrullDim_eq_zero_of_field (F : Type*) [Field F] : ringKrullDim F = 0 :=
+  krullDim_eq_zero_of_unique
+
+theorem ringKrullDim_eq_zero_of_isField {F : Type*} [CommRing F] (hF : IsField F) :
```
erdOne, line 112: Can we make this an iff for domains?

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,123 @@
...
+variable {R S : Type*} [CommRing R] [CommRing S]
+
+@[nontriviality]
+lemma ringKrullDim_eq_bot_of_subsingleton [Subsingleton R] :
+    ringKrullDim R = ⊥ :=
+  krullDim_eq_bot_of_isEmpty
+
+lemma ringKrullDim_nonneg_of_nontrivial [Nontrivial R] :
+    0 ≤ ringKrullDim R :=
+  krullDim_nonneg_of_nonempty
```
erdOne, line 42: We would probably want iff versions of these or else there is little reason to let `ringKrullDim` be valued in `WithBot`. We are missing the analogous lemmas for `krullDim` though so maybe we can add this in a separate PR.

Mathlib/Order/KrullDimension.lean
FMLJohn: Thank you! Changed!

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: So shall I delete `PrimeSpectrum.asIdeal.isMaximal_of_ringKrullDim_eq_zero`?

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: OK! I'll prove this in a new PR!

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: > I think the next lemma is good enough.

So shall I delete `PrimeSpectrum.asIdeal.mem_minimalPrimes_of_ringKrullDim_eq_zero`?

Mathlib/RingTheory/KrullDimension.lean
erdOne: Yes that's my suggestion. Same below.

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: I have formalised the theorem `ringKrullDim_eq_zero_iff_isField_of_isDomain`, which says that the Krull dimension of an integral domain is zero if and only if it is a field. Could you please help me check if any golfing can be done? Also, shall I preserve the theorems `ringKrullDim_eq_zero_of_field` and `ringKrullDim_eq_zero_of_isField` or delete them?

FMLJohn: > ringKrullDim_eq_zero_of_isField

But the statement of `ringKrullDim_eq_zero_of_isField` is different from that of `ringKrullDim_eq_zero_of_field`. One says that `isField` implies that the Krull dimension is 0; the other says that `Field` implies that the Krull dimension is 0. How can I make it an alias?

FMLJohn: > `ringKrullDim_eq_zero_of_field` should be kept. `ringKrullDim_eq_zero_of_isField` can be an `alias`. I was thinking along the lines of
> 
> ```
> theorem PrimeSpectrum.mem_minimalPrimes_iff {p : PrimeSpectrum R} :
>     p.asIdeal ∈ minimalPrimes R ↔ p ∈ minimals (· ≤ ·) Set.univ :=
>   have : asIdeal '' Set.univ = {p : Ideal R | p.IsPrime ∧ ⊥ ≤ p} := by
>     ext x; simpa using ⟨fun H ↦ H.choose_spec ▸ H.choose.2, fun H ↦ ⟨⟨_, H⟩, rfl⟩⟩
>   Iff.trans (this ▸ Iff.rfl) (map_mem_minimals_iff (s := LE.le) (f := asIdeal) (by simp) (by simp))
> 
> theorem ringKrullDim_eq_zero_iff_forall_mem_minimalPrimes [Nontrivial R] :
>   ringKrullDim R = 0 ↔ ∀ (I : Ideal R), I.IsPrime → I ∈ minimalPrimes R := by
>   refine krullDim_eq_zero_iff_mem_minimals_of_nonempty.trans ?_
>   simp_rw [← PrimeSpectrum.mem_minimalPrimes_iff]
>   exact ⟨(· ⟨·, ·⟩), (· _ <| PrimeSpectrum.isPrime ·)⟩
>   
> attribute [local instance] Ideal.bot_prime
> 
> theorem ringKrullDim_eq_zero_iff_isField_of_isDomain [IsDomain R] :
>     ringKrullDim R = 0 ↔ IsField R := by
>   simp only [ringKrullDim_eq_zero_iff_forall_mem_minimalPrimes, minimalPrimes,
>     Ideal.minimalPrimes_eq_subsingleton_self, Set.mem_singleton_iff,
>     ← not_iff_comm.mp Ring.not_isField_iff_exists_prime, ne_eq, not_exists, not_and, not_imp_not]
> ```
> 
> The first lemma is also useful for `Ideal.IsPrime.mem_minimalPrimes_of_ringKrullDim_eq_zero` and the second should be added anyways (also the maximal version).

But the statement of `ringKrullDim_eq_zero_of_isField` is different from that of `ringKrullDim_eq_zero_of_field`. One says that `isField` implies that the Krull dimension is 0; the other says that `Field` implies that the Krull dimension is 0. How can I make it an alias?

erdOne: I was originally thinking of `alias ⟨_, ringKrullDim_eq_zero_of_isField⟩ := ringKrullDim_eq_zero_iff_isField_of_isDomain` but you're right let's keep both.

erdOne:
Thanks!
maintainer merge

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,120 @@
+/-
+Copyright (c) 2024 Jujian Zhang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fangming Li, Jujian Zhang
+-/
+import Mathlib.AlgebraicGeometry.PrimeSpectrum.Basic
```
mattrobball, line 6: This should import `Mathlib.RingTheory.PrimeSpectrum` and results that rely on Zariski topology should be moved to `AlgebraicGeometry`.

Mathlib/RingTheory/KrullDimension.lean
```diff
@@ -0,0 +1,120 @@
...
+
+theorem ringKrullDim_eq_zero_iff_forall_mem_minimalPrimes [Nontrivial R] :
+    ringKrullDim R = 0 ↔ ∀ (I : Ideal R), I.IsPrime → I ∈ minimalPrimes R := by
+  refine krullDim_eq_zero_iff_forall_mem_minimals_of_nonempty.trans ?_
+  simp_rw [← PrimeSpectrum.mem_minimalPrimes_iff]
+  exact ⟨(· ⟨·, ·⟩), (· _ <| PrimeSpectrum.isPrime ·)⟩
+
+end DimensionZero
+
+section Field
```
mattrobball, line 95: Let's split this section into a separate file.

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: But in the current version of Mathlib of this pull request, there is no `Mathlib.RingTheory.PrimeSpectrum`. Do I need to run `git merge origin/master`?

Mathlib/RingTheory/KrullDimension.lean
FMLJohn: Shall I create a folder `Mathlib.RingTheory.KrullDimension` and create the files `Mathlib.RingTheory.KrullDimension.Basic` and `Mathlib.RingTheory.KrullDimension.Field` and then move the content of this section into `Mathlib.RingTheory.KrullDimension.Field`?

Mathlib/RingTheory/KrullDimension.lean
mattrobball: It's good practice to merge master if it has been a week or more. Eg, GitHub can't tell if there is a linter failure in your PR because of new changes to master.

FMLJohn: But after I ran `git merge origin/master`, there is still no `Mathlib.RingTheory.PrimeSpectrum`. How can I solve the problem?

FMLJohn: After I changed the import into `Mathlib.RingTheory.PrimeSpectrum`, the core definition `ringKrullDim` stopped compiling because the instance `Preorder (PrimeSpectrum R)` could not be synthesized. And all the contents of the file rely on the definition `ringKrullDim`. I'm not sure how to resolve the problem now?

erdOne: The current preorder on Spec R is given by the specialization order and depends on the topology. If it is easy to do you can redefine it using the induced order from ideals and move it to `RingTheory`. If not this could be done in a separate PR. But please still split the lemmas really depending on the topology into a new PR.

mattrobball: :/ Let's move this file to `AlgebraicGeometry.KrullDimension` and follow up with another PR moving back into `RingTheory` with the standard preorder and splitting off `Field` results.

FMLJohn: So for this PR, shall I create a file `AlgebraicGeometry/KrullDimension.lean`, or shall I create a folder `AlgebraicGeometry/KrullDimension` and then a file `AlgebraicGeometry/KrullDimension/Basic.lean`?

mattrobball: Just a single file is ok for now since it will be touched again in another PR. We should get the correct preorder instance before splitting

erdOne: Or we can wait for #15108.

FMLJohn: > Or we can wait for #15108.

Yeah I feel that this is a good idea. Also I am not sure whether redefining `ringKrullDim` is a good idea because @jjaassoonn and I have already formalized a bunch of results (e.g. the Krull dimensions of PIDs, Artinian rings, etc.) based on the definition
```
noncomputable def ringKrullDim (R : Type*) [CommRing R] : WithBot (WithTop ℕ) :=
  krullDim (PrimeSpectrum R)
```
and I'm not sure if our proofs will remain as concise as before after redefining `ringKrullDim`.

mattrobball: I don't think `ringKrullDim` needs to change. We just need the preorder instance in scope for it to work. I didn't realize it was already the usual partial order (ideal containment) and can just be moved easily.

mattrobball: > Or we can wait for #15108.

This will be on master shortly

FMLJohn: I have merged with master and changed the import into `Mathlib.RingTheory.PrimeSpectrum`, but `Preorder (PrimeSpectrum R)` still cannot be synthesized?

erdOne: I just merged master for you. Please try again.

erdOne: (and yes next time please run `git fetch` before merging and merge with `origin master` and not the `master` branch on your local machine. Or `git pull origin master` works too I think.)

FMLJohn: The theorem “ringKrullDim_le_of_surjective” stopped compiling after I changed the import because the proof of the theorem uses “PrimeSpectrum.comap”. But I don’t think it is reasonable to move “ringKrullDim_le_of_surjective” to “AlgebraicGeometry”, because the theorem is one of the most basic theorems about the Krull dimension of a ring. Is there any good way to solve this problem?

erdOne: The easy way out is to replace `comap` with its definition and the lemma with `Ideal.comap_injective_of_surjective` in the proof. The long term solution is to unbundle `comap` from continuity and move it to `RingTheory`

mattrobball: There should be a `PrimeSpectrum.comapFun`.

mattrobball: #15114 breaks out the function. It was done a bit hastily so please look it over.

jcommelin: @mattrobball Should #15114 be a dependency of this PR?

jcommelin:
@FMLJohn @erdOne @mattrobball what is the status here? I think this one fell through some cracks in the queue.

Mathlib/RingTheory/KrullDimension.lean
mattrobball: Assuming it looks fine yes. Otherwise, I would like to find a reasonable way to keep the line between commutative algebra and algebraic geometry at importing the Zariski topology.

mattrobball:
It looks like #14721 did not deprecate anything or provide an adaptation path here. @apnelson1 can you find this PR? Thanks.

mattrobball:
bors merge

PR #14084, opened 2024-06-24
[Merged by Bors] - feat: Functor.mapHomotopyCategory is a triangulated functor

joelriou:
In this PR, we show that if `F` is an additive functor, then the induced functor on the homotopy categories is a triangulated functor.

---

- [x] depends on: #14087

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14153, opened 2024-06-26
[Merged by Bors] - feat: Functor.mapDerivedCategory is a triangulated functor

joelriou:
In this PR, it is shown that an exact functor between abelian categories induces a triangulated functor between the corresponding derived categories.

---

- [x] depends on: #14084


semorrison:
bors merge

PR #14159, opened 2024-06-26
[Merged by Bors] - feat: lemmas needed for Sobolev inequality

fpvandoorn:
* generalize and rename `MeasureTheory.snorm_indicator_ge_of_bdd_below`

---

### DIFF
--- a/Mathlib/MeasureTheory/Function/LpSpace.lean
+++ b/Mathlib/MeasureTheory/Function/LpSpace.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Rémy Degenne, Sébastien Gouëzel
 -/
 import Mathlib.Analysis.Normed.Group.Hom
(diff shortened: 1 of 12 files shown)


Mathlib/MeasureTheory/Function/LpSpace.lean
```diff
@@ -752,6 +753,26 @@ protected lemma Memℒp.piecewise [DecidablePred (· ∈ s)] {g}
 
 end Indicator
 
+section Topology
+variable {X : Type*} [TopologicalSpace X] [MeasurableSpace X] [OpensMeasurableSpace X]
```
sgouezel, line 757: Could you move the assumption `[OpensMeasurableSpace X]` to the theorem that needs it, i.e., the second one.

sgouezel (commented):
bors d+
Thanks!

fpvandoorn:
bors merge

PR #14160, opened 2024-06-26
feat(Algebra/Group/Action/Defs): Redefine `Function.End`

wrenna-robson:
Function.End is currently defined, semireducibly, as just a function type. This occasionally leads to irritating behaviour where theorems about Function.End don't apply to a statement which is about a function. I am redefining Function.End as a one-field structure and seeing what that breaks, and if it doesn't break too much, I will also add support for coercing Equiv.Perm to it.

---


riccardobrasca:
What is the status of this PR? Can you please mark  `WIP` or `awaiting-review`? Thanks!

PR #13248, opened 2024-05-26
feat: basic concepts of auction theory

hcWang942:
## Description
Formalise some core concepts and results in auction theory: this includes definitions for first-price and second-price auctions, as well as several fundamental results and helping lemmas.

This is the very first PR of the project formalizing core concepts and results in auction theory.
Our group is working on more contributions on the formalization of game theory prefix. 

Co-authored-by: Ma Jiajun <hoxide@gmail.com>

## Reference
Roughgarden, Tim. ***Twenty Lectures on Algorithmic Game Theory***. Cambridge University Press, 2020. [Link](https://www.cambridge.org/core/books/twenty-lectures-on-algorithmic-game-theory/A9D9427C8F43E7DAEF8C702755B6D72B)
---
- [x] Will depend on #14163 once that PR is merged. The Fintype lemmas introduced by this PR have been added in that PR and will be removed from here once that PR gets merged

## Current plan for formalization of Game Theory
The current plan for the formalizing of Game Theory include:
#### 1. Auction Theory. 🎉   _(200+ lines, this PR)_ 
- Essential definitions of Sealed-bid auction, First-price auction and Second-price auction.
- First-price auction has no dominant strategy.
- Second-price auction has dominant strategy. (Second-price auction is DSIC)
#### 2. Mechanism design & Myerson's Lemma.    🎉 (400+ lines, pending for modification to Mathlib Standard)
-  Mechanism design
An allocation rule is implementable if there exists 
        - Dominant Strategy Incentive Compatible (DSIC) payment rule
        - An allocation rule is monotone if for every bidder’s gain is nondecreasing w.r.t. her/his bid
- Myerson's Lemma
                Implementable ⇔ Monotone
                In the above case, the DSIC payment rule is unique.                                     
#### 3. von Neumann‘s Minimax Theorem. 🎉 (800+ lines, pending for modification to Mathlib Standard)
- Equilibrium in zero sum game
- Formalization strategy: via Loomis’s theorem.

PR #14163, opened 2024-06-26
[Merged by Bors] - feat(Fintype.Basic): Add two theorems supporting Auction Theory formalizations

hcWang942:
This PR adds two missing theorems in `Mathlib.Data.Fintype.Basic`

---

This is used in another PR of mine formalizing some basic theorems in `GameTheory.AuctionTheory.Basic` .

One may check the previous PR for Auction Theory here: #13248 

I would sincerely appreciate a timely review of this PR, as I am eager to complete my first project on Auction Theory before my upcoming birthday. Thank you very much for your consideration and support!


Shreyas4991:
It seems you have rebased your PR on the other PR branch so now we have all these other commits. Can you remove those commits, branch off master (or reset HEAD to the latest master commit) and then commit only the two new theorems?

hcWang942:
> It seems you have rebased your PR on the other PR branch so now we have all these other commits which are not needed in this PR. In your branch `finset_nontrivial_two_theorems_add` can you remove those commits by resetting `HEAD` to the latest master commit and then commit only the two new theorems? Once you only have the one new commit in this branch for adding the two theorems, you will have to force push to this branch here.

May I know if it is correct to use the command
`
git fetch origin master
git reset --hard origin/master
`
Then commit and push the changes again.

Mathlib/Data/Fintype/Basic.lean
```diff
@@ -114,6 +114,14 @@ theorem univ_eq_empty_iff : (univ : Finset α) = ∅ ↔ IsEmpty α := by
   rw [← not_nonempty_iff, ← univ_nonempty_iff, not_nonempty_iff_eq_empty]
 #align finset.univ_eq_empty_iff Finset.univ_eq_empty_iff
 
+theorem univ_nontrivial_iff [Fintype α] :
```
tb65536, line 117: I think the `[Fintype α]` assumption is already present at this point of the file.

Shreyas4991:
> You don't need to `git fetch origin master`. I have a beginner friendly but slightly tedious idea. I am assuming you are currently on this branch at the last I commit I see on this page [9659d19](https://github.com/leanprover-community/mathlib4/pull/14163/commits/9659d19481b1b158d7fde7bc9d79bae15e624b07). if you open git log by running `git log` and scroll down, you will find the most recent commit that is not your commit in this PR. copy that commit ID (which I denote by `<commit_ID>` hereon). then `git reset --hard <commit_ID>`. Then add your two theorems (which you can in the respective file). add and commit that. Then `git push --force origin finset_nontrivial_two_theorems_add`

I am so so sorry. Instead of hitting reply, I seem to have hit edit on your comment. I hope the question you have was answered though.

Shreyas4991:
Could you please delete this file `https://github.com/leanprover-community/mathlib4/pull/14163/files#diff-ffd24c8f58ab0af7a8d4610ab7fc208cf0fef60e76059dcd605f7113179a96ca`

hcWang942:
> Could you please delete this file `https://github.com/leanprover-community/mathlib4/pull/14163/files#diff-ffd24c8f58ab0af7a8d4610ab7fc208cf0fef60e76059dcd605f7113179a96ca`
> 
> EDIT: You can do this from this webpage by going to the `files changed` tab at the top your PR description and deleting that large file with 66000 changes.
> 
> EDIT 2 : If I have to guess the cause of this file, git probably opened your text editor for the commit message after the reset and for reasons beyond me saved the file into the toplevel directory of your local clone of mathlib. This file can be safely discarded.

Thanks for the suggestion, relevant modification has been done!

Shreyas4991 (commented):
I am not convinced that `univ_nontrivial` belongs in mathlib. I don't think it is mathlib practice to have `mp` and  `mpr` versions of lemmas as separate globally named lemmas. The `iff` version suffices. `univ_nontrivial_iff` can be used with `rw` and `univ_nontrivial_iff.mpr` with `apply`, depending on which tactic you are using.

tb65536:
The one advantage of `univ_nontrivial` is that it can take `Nontrivial α` automatically. For instance, see `univ_nonempty_iff` and `univ_nonempty` a few lines above.

Shreyas4991:
> The one advantage of `univ_nontrivial` is that it can take `Nontrivial α` automatically. For instance, see `univ_nonempty_iff` and `univ_nonempty` a few lines above.

If so, is there a reason this theorem is not an instance declaration instead?

tb65536:
> > The one advantage of `univ_nontrivial` is that it can take `Nontrivial α` automatically. For instance, see `univ_nonempty_iff` and `univ_nonempty` a few lines above.
> 
> If so, is there a reason this theorem is not an instance declaration instead?

I think `Finset.Nonempty` and `Finset.Nontrivial` are just defs, not typeclasses.

Shreyas4991:
> > > The one advantage of `univ_nontrivial` is that it can take `Nontrivial α` automatically. For instance, see `univ_nonempty_iff` and `univ_nonempty` a few lines above.
> > 
> > If so, is there a reason this theorem is not an instance declaration instead?
> 
> I think `Finset.Nonempty` and `Finset.Nontrivial` are just defs, not typeclasses.

Okay then I take back that comment, but I still don't see why `univ_nonempty_iff.mpr ` is not giving you the same ease of use as `univ_nonempty`. That being said, I am not an official reviewer or maintainer, so please take this opinion with a pinch of salt.

tb65536 (approved):
I think the above discussion has resolved and this is still waiting for review? Personally, this PR looks good to me.

hcWang942:
> I think the above discussion has resolved and this is still waiting for review? Personally, this PR looks good to me.

Thanks for the kind reply, may I know what more can I do now? Or I just need to wait patiently.

Shreyas4991 (approved):

j-loreaux:
Arguably, `Finset.univ_nontrivial` could / should be marked `@[simp]`, but I think I prefer it as is so that we don't insert this generic typeclass search into the `simp` calls.

bors merge

PR #14124, opened 2024-06-25
[Merged by Bors] - feat: the norm raised to a power is C^1

fpvandoorn:
* From the Sobolev inequality project
* Generalize `continuousAt_rpow_const`

Co-authored-by: Heather Macbeth 25316162+hrmacbeth@users.noreply.github.com

---
- [x] depends on: #14123 [one proof will fail until then]

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14165, opened 2024-06-26
[Merged by Bors] - feat: prove the Gagliardo-Nirenberg-Sobolev inequality

fpvandoorn:

---

- [x] depends on: #14124
- [x] depends on: #14159


Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,619 @@
...
+/-- Special case of the grid-lines lemma `lintegral_mul_prod_lintegral_pow_le`, taking the extremal
+exponent `p = (#ι - 1)⁻¹`. -/
+theorem lintegral_prod_lintegral_pow_le [Nontrivial ι]
+    {p : ℝ} (hp : Real.IsConjExponent #ι p)
+    {f} (hf : Measurable f) :
+    ∫⁻ x, ∏ i, (∫⁻ xᵢ, f (update x i xᵢ) ∂μ i) ^ ((1 : ℝ) / (#ι - 1 : ℝ)) ∂.pi μ
+    ≤ (∫⁻ x, f x ∂.pi μ) ^ p := by
+  have h0 : (1:ℝ) < #ι := by norm_cast; exact Fintype.one_lt_card
+  have h1 : (0:ℝ) < #ι - 1 := by linarith
+  have h3 : 0 ≤ ((1 : ℝ) / (#ι - 1 : ℝ)) := by positivity
```
Ruben-VandeVelde, line 241: I guess there used to be a `h2` here; you might renumber

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,619 @@
...
+    _ ≤ ‖fderiv ℝ u (update x i y)‖₊ * ‖deriv (update x i) y‖₊ :=
+        ContinuousLinearMap.le_opNNNorm ..
+    _ ≤ ‖fderiv ℝ u (update x i y)‖₊ := by simp [deriv_update, Pi.nnnorm_single]
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [MeasurableSpace E] [BorelSpace E]
+  [FiniteDimensional ℝ E] (μ : Measure E) [IsAddHaarMeasure μ]
+
+open FiniteDimensional
+
+set_option linter.unusedVariables false in
```
Ruben-VandeVelde, line 307: Why?

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,619 @@
...
+    convert hC hu h2u
+    ext
+    rw [← inv_inj, hp']
+    field_simp [n', NNReal.conjExponent]
+  -- the case `p > 1`
+  let q := Real.conjExponent p
+  have hq : Real.IsConjExponent p q := .conjExponent hp
+  have h0p : p ≠ 0 := zero_lt_one.trans hp |>.ne'
+  have h1p : (p : ℝ) ≠ 1 := hq.one_lt.ne'
+  -- have h3p : (p : ℝ) ≠ 0 := hq.pos.ne'
```
Ruben-VandeVelde, line 433:
```suggestion
```

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,619 @@
...
+theorem snorm_le_snorm_fderiv_of_eq [FiniteDimensional ℝ F] {p p' : ℝ≥0} (hp : 1 ≤ p)
+    (h2p : p < finrank ℝ E) (hp' : (p' : ℝ)⁻¹ = p⁻¹ - (finrank ℝ E : ℝ)⁻¹) :
+    ∃ C : ℝ≥0, ∀ {u : E → F} (hu : ContDiff ℝ 1 u) (h2u : HasCompactSupport u),
+    snorm u p' μ ≤ C * snorm (fderiv ℝ u) p μ := by
+  let F' := EuclideanSpace ℝ <| Fin <| finrank ℝ F
+  let e : F ≃L[ℝ] F' := toEuclidean
+  let C₁ : ℝ≥0 := ‖(e.symm : F' →L[ℝ] F)‖₊
+  let C₂ : ℝ≥0 := ‖(e : F →L[ℝ] F')‖₊
+  obtain ⟨C, hC⟩ := snorm_le_snorm_fderiv_of_eq_inner μ F' hp h2p hp'
+  refine ⟨C₁ * C * C₂, @fun u hu h2u ↦ ?_⟩ -- nope?
```
Ruben-VandeVelde, line 523: "nope?"?

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: It flags the hypotheses `hu` and `h2u` in the conclusion. But disabling the linter is not nice, I enabled it again.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: Oops, artifact from when the constant wasn't correct yet :-)

Mathlib/MeasureTheory/Function/LpSpace.lean
```diff
@@ -672,6 +672,13 @@ theorem memℒp_indicator_iff_restrict (hs : MeasurableSet s) :
   simp [Memℒp, aestronglyMeasurable_indicator_iff hs, snorm_indicator_eq_snorm_restrict hs]
 #align measure_theory.mem_ℒp_indicator_iff_restrict MeasureTheory.memℒp_indicator_iff_restrict
 
+/-- For a function `f` with support in `s`, the Lᵖ norms of `f` with respect to `μ` and
+`μ.restrict s` are the same. -/
+theorem snorm_restrict_eq {s : Set α} (hs : MeasurableSet s) (hsf : f.support ⊆ s) :
```
sgouezel, line 677: hs shouldn't be needed: otherwise, use the measurable hull of s instead. Since it contains s, it will also satisfy hsf

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,610 @@
...
+Let `n` be the dimension of the domain.
+
+## Main results
+
+* `MeasureTheory.snorm_le_snorm_fderiv_of_eq`:
+  The bound holds for `1 ≤ p < n` and `q⁻¹ = p⁻¹ - n⁻¹`
+* `MeasureTheory.snorm_le_snorm_fderiv_of_le`:
+  The bound holds when `1 ≤ p < n`, `0 ≤ q` and `p⁻¹ - n⁻¹ ≤ q⁻¹`.
+  Note that in this case the constant depends on the support of `u`.
+-/
```
sgouezel, line 28: Could you sketch in the file-level docstring the proof you are following, first in informal terms and then describing the formalization concepts you use to avoid DTT?

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,610 @@
...
+/-- The main inductive step in the grid-lines lemma for the Gagliardo-Nirenberg-Sobolev inequality.
+
+The grid-lines operation `GridLines.T` on a nonnegative function on a finitary product type is
+less than or equal to the grid-lines operation of its partial integral in one co-ordinate
+(the latter intuitively considered as a function on a space "one dimension down"). -/
+theorem T_insert_le_T_lmarginal_singleton (hp₀ : 0 ≤ p) (s : Finset ι)
+    (hp : (s.card : ℝ) * p ≤ 1)
+    (i : ι) (hi : i ∉ s) {f : (∀ i, A i) → ℝ≥0∞} (hf : Measurable f) :
+    T μ p f (insert i s) ≤ T μ p (∫⋯∫⁻_{i}, f ∂μ) s := by
+  calc T μ p f (insert i s)
```
sgouezel, line 87: Whenever a proof is more than 15 lines long, I think it deserves some comments explaining what you are doing. This one qualifies.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,610 @@
...
+expression `|u x| ^ (n / (n - 1))` is bounded above by the `n / (n - 1)`-th power of the Lebesgue
+integral of the Fréchet derivative of `u`.
+
+For a basis-free version, see `lintegral_pow_le_pow_lintegral_fderiv`. -/
+theorem lintegral_pow_le_pow_lintegral_fderiv_aux
+    {p : ℝ} (hp : Real.IsConjExponent #ι p)
+    {u : (ι → ℝ) → F} (hu : ContDiff ℝ 1 u)
+    (h2u : HasCompactSupport u) :
+    ∫⁻ x, (‖u x‖₊ : ℝ≥0∞) ^ p ≤ (∫⁻ x, ‖fderiv ℝ u x‖₊) ^ p := by
+  have : (1:ℝ) ≤ ↑#ι - 1 := by
```
sgouezel, line 262: Add some comments in the proof?

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,610 @@
...
+  have h2 : 0 ≤ ((1 : ℝ) / (#ι - 1 : ℝ)) := by positivity
+  have h3 : (#ι - 1 : ℝ) * ((1 : ℝ) / (#ι - 1 : ℝ)) ≤ 1 := by field_simp
+  have h4 : p = 1 + 1 / (↑#ι - 1) := by field_simp; rw [mul_comm, hp.sub_one_mul_conj]
+  rw [h4]
+  convert lintegral_mul_prod_lintegral_pow_le μ h2 h3 hf using 2
+  field_simp
+
+/-! ## The Gagliardo-Nirenberg-Sobolev inequality -/
+
+variable [Nontrivial ι] {F : Type*} [NormedAddCommGroup F] [NormedSpace ℝ F] [CompleteSpace F]
```
sgouezel, line 248: Do you really need the assumption that iota is nontrivial? Below, you assume `Real.IsConjExponent #ι p`, which probably implies it? Same thing in the other statements below.

Mathlib/MeasureTheory/Function/LpSpace.lean
fpvandoorn: I also thought that, but I wasn't sure how to get `snorm f p (μ.restrict s) = snorm f p (μ.restrict (toMeasurable μ s)`

sgouezel: Loogle tells me about [MeasureTheory.Measure.restrict_toMeasurable_of_sFinite](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Measure/Typeclasses.html#MeasureTheory.Measure.restrict_toMeasurable_of_sFinite)

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: Well spotted! This removes multiple other arguments.
In `snorm_le_snorm_fderiv_of_eq_inner` I added a new condition that `0 < n`, and I *think* that is necessary, but my brain hurts when I have to think about 0-dimensional vector spaces.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,610 @@
...
+compactly-supported function `u` on a normed space `E` of finite dimension `n`, equipped
+with Haar measure, let `1 < p < n` and let `p'⁻¹ := p⁻¹ - n⁻¹`.
+There exists a constant `C` depending only on `E` and `p`, such that the `Lᵖ'` norm of `u`
+is bounded above by `C` times the `Lᵖ` norm of the Fréchet derivative of `u`.
+
+Note: The codomain of `u` needs to be a Hilbert space.
+-/
+theorem snorm_le_snorm_fderiv_of_eq_inner {p p' : ℝ≥0} (hp : 1 ≤ p)
+    (h2p : p < finrank ℝ E) (hp' : (p' : ℝ)⁻¹ = p⁻¹ - (finrank ℝ E : ℝ)⁻¹) :
+    ∃ C : ℝ≥0, ∀ {u : E → F'} (_hu : ContDiff ℝ 1 u) (_h2u : HasCompactSupport u),
```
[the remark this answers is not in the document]
fpvandoorn: Done. Note that in the versions with a finite dimensional normed space as codomain it does depend on (the norm of) `F`.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: Done, and I also further gave a proof sketch in 5 long proofs.

Mathlib/MeasureTheory/Function/LpSpace.lean
fpvandoorn: I replaced the lemma, and this is actually nicer. I didn't realize we had to put some effort in making the set measurable when we applied it, so this is a nice simplification.

This does add a small amount of imports to the file `LpSpace`.

fpvandoorn:
I (hopefully) addressed all review comments.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,663 @@
...
+    _ ≤ ‖fderiv ℝ u (update x i y)‖₊ * ‖deriv (update x i) y‖₊ :=
+        ContinuousLinearMap.le_opNNNorm ..
+    _ ≤ ‖fderiv ℝ u (update x i y)‖₊ := by simp [deriv_update, Pi.nnnorm_single]
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [MeasurableSpace E] [BorelSpace E]
+  [FiniteDimensional ℝ E] (μ : Measure E) [IsAddHaarMeasure μ]
+
+open FiniteDimensional
+
+variable (F) in
```
sgouezel, line 343: still needed?

Mathlib/MeasureTheory/Function/LpSpace.lean
```diff
@@ -672,6 +673,15 @@ theorem memℒp_indicator_iff_restrict (hs : MeasurableSet s) :
   simp [Memℒp, aestronglyMeasurable_indicator_iff hs, snorm_indicator_eq_snorm_restrict hs]
 #align measure_theory.mem_ℒp_indicator_iff_restrict MeasureTheory.memℒp_indicator_iff_restrict
 
+/-- For a function `f` with support in `s`, the Lᵖ norms of `f` with respect to `μ` and
+`μ.restrict s` are the same. -/
+theorem snorm_restrict_eq [SFinite μ] {s : Set α} (hsf : f.support ⊆ s) :
```
sgouezel, line 678: In fact I think it's possible to get rid also of the SFinite condition, but I'll do that after this PR is merged.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,661 @@
...
+  have h2 : 0 ≤ ((1 : ℝ) / (#ι - 1 : ℝ)) := by positivity
+  have h3 : (#ι - 1 : ℝ) * ((1 : ℝ) / (#ι - 1 : ℝ)) ≤ 1 := by field_simp
+  have h4 : p = 1 + 1 / (↑#ι - 1) := by field_simp; rw [mul_comm, hp.sub_one_mul_conj]
+  rw [h4]
+  convert lintegral_mul_prod_lintegral_pow_le μ h2 h3 hf using 2
+  field_simp
+
+/-! ## The Gagliardo-Nirenberg-Sobolev inequality -/
+
+variable {F : Type*} [NormedAddCommGroup F] [NormedSpace ℝ F] [CompleteSpace F]
```
sgouezel, line 277: The completeness assumption here is probably not necessary, but removing this can definitely wait for another PR.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,661 @@
...
+    _ ≤ ‖fderiv ℝ u (update x i y)‖₊ := by simp [deriv_update, Pi.nnnorm_single]
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [MeasurableSpace E] [BorelSpace E]
+  [FiniteDimensional ℝ E] (μ : Measure E) [IsAddHaarMeasure μ]
+
+open FiniteDimensional
+
+/-- The **Gagliardo-Nirenberg-Sobolev inequality**.  Let `u` be a continuously differentiable
+compactly-supported function `u` on a normed space `E` of finite dimension `n ≥ 2`, equipped
+with Haar measure. There exists a constant `C` depending only on `E`, such that the Lebesgue
```
sgouezel, line 345:
```suggestion
with Haar measure. There exists a constant `C` depending only on `E` and `μ`, such that the Lebesgue
```

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,661 @@
...
+integral of the pointwise expression `|u x| ^ (n / (n - 1))` is bounded above by `C` times the
+`n / (n - 1)`-th power of the Lebesgue integral of the Fréchet derivative of `u`. -/
+theorem lintegral_pow_le_pow_lintegral_fderiv {p : ℝ} (hp : Real.IsConjExponent (finrank ℝ E) p) :
+    ∃ C : ℝ≥0, ∀ {F : Type*} [NormedAddCommGroup F] [NormedSpace ℝ F] [CompleteSpace F]
```
sgouezel, line 349: I have a small problem with existential statements like that, that the constant `C` might a priori depend on the universe of `F`. So you can not really write that "it only depend on E and mu" (well, it's true, but not what the statement says). I wonder if it's not clearer to define a constant `gagliardoNirenbergSobolevConstant E \mu := ...` and use that in the statements, to lift this ambiguity (with a shorter notation to be used in the current file, to make things readable). It would also have the advantage to avoid the ugly forall quantifier in the statement.

Same thing for the other statements below.

What do you think?

sgouezel (commented):
It feels weird to have this in the MeasureTheory folder; this is firmly a statement in analysis, just like all Sobolev spaces theory. Maybe create a folder `Analysis/FunctionalSpaces` and put it there?

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: I am in favor, since I think that will be a lot nicer to work with in practice. I'm also doing that in the formalization of Carleson's theorem. I wasn't sure if that style was appropriate for Mathlib.

fpvandoorn: Since multiple statements in this file have slightly different constants, I'll probably use something like `lintegralPowLePowLintegralFderivConst`.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
fpvandoorn: I reformulated these sentences now that we've separated out the constants.

fpvandoorn:
Thanks for your quick reviews! 
I don't see how to quickly do the 2 suggested generalizations, so let's leave them for a later PR.
I incorporated your last suggestion, reformulated docstrings with this change, and added a paragraph in the module doc explaining this.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,711 @@
...
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [MeasurableSpace E] [BorelSpace E]
+  [FiniteDimensional ℝ E] (μ : Measure E) [IsAddHaarMeasure μ]
+
+open FiniteDimensional
+
+/-- The constant factor occurring in the conclusion of `lintegral_pow_le_pow_lintegral_fderiv`.
+It only depends on `E`, `μ` and `p`.
+It is determined by the ratio of the measures on `E` and `ℝⁿ` and
+the operator norm of `e` (raised to suitable powers involving `p`).-/
+def lintegralPowLePowLIntegralFDerivConst (p : ℝ) : ℝ≥0 := by
```
sgouezel, line 352: `irreducible_def`, maybe, to make sure that Lean will not try to unfold it given how nasty it is. Same thing for all the other constants definition below.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
```diff
@@ -0,0 +1,711 @@
...
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [MeasurableSpace E] [BorelSpace E]
+  [FiniteDimensional ℝ E] (μ : Measure E) [IsAddHaarMeasure μ]
+
+open FiniteDimensional
+
+/-- The constant factor occurring in the conclusion of `lintegral_pow_le_pow_lintegral_fderiv`.
+It only depends on `E`, `μ` and `p`.
+It is determined by the ratio of the measures on `E` and `ℝⁿ` and
+the operator norm of `e` (raised to suitable powers involving `p`).-/
```
sgouezel, line 351: Can you say in the docstring what `e` is?

sgouezel (commented):
bors d+
Thanks!

fpvandoorn:
bors merge

Mathlib/MeasureTheory/Function/LpSpace.lean
sgouezel: I have a proof removing the `SFinite` assumption, but I will first wait for #14260 to be merged, to avoid introducing new lemmas with a wrong name.

Mathlib/MeasureTheory/Integral/SobolevInequality.lean
sgouezel: Done in #14262

Mathlib/MeasureTheory/Function/LpSpace.lean
sgouezel: See #14289

PR #14049, opened 2024-06-23
Archive: `a^n+1` is prime only if `n` is a power of two

rwst:
The fact that primes of the form `a^n+1` must have `n` a power of two was stated and
the proof outlined in three sentences by Euler in the year 1738 [E026]. A proof can be
found in Hardy & Wright, An introduction to the theory of numbers (5th ed., 1979),
p.15, Theorem 17. In the proof below we followed Euler's thoughts.

---
- [ ] depends on: #14357
-->

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14166, opened 2024-06-26
[Merged by Bors] - feat(Nat/Factorization): add `exists_eq_two_pow_mul_odd`

rwst:
Any nonzero natural number is the product of an odd part `m` and a power of two `2 ^ k`.

This is a trivial consequence of the existing `exists_eq_pow_mul_and_not_dvd`, but performs a convenient restatement in terms of `Odd`.

Needed by #14049.

Co-authored-by: Thomas Browning <tb65536@uw.edu>

---


tb65536 (approved):
LGTM, but someone else should take a look since I helped write it.

eric-wieser:
bors merge

Thanks! I updated the PR description to explain why I think this lemma is just over the line of "too trivial a consequence of an existing result"

PR #14168, opened 2024-06-26
[Merged by Bors] - chore: deprecate AlgHom.map_* lemmas

Ruben-VandeVelde:
These are redundant. Unfortunately there's two cases involving `GradedAlgebra` where a direct replacement times out, so I ignored those for now.

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #14172, opened 2024-06-26
[Merged by Bors] - doc: clarify finite adele ring docstring

kbuzzard:
---

OK so what's going on here is that I made a PR #13705 adding this declaration and some API, it was delegated, then Oliver suggested some changes to docstrings which I thought I'd clicked "apply suggestion" on but apparently I missed one; I then merged the PR without one of his suggestions, and it's a good suggestion. He then kindly DMed me asking me if his suggestion was wrong, which led me to discover my error :-) This PR adds his suggestion back in.


ocfnash:
Thanks!

bors merge

PR #14177, opened 2024-06-27
[Merged by Bors] - feat: define the weak operator topology

dupuisf:
This PR defines the [weak operator topology](https://en.wikipedia.org/wiki/Weak_operator_topology) on maps between normed spaces and proves some of its basic properties. It creates a type copy of `E →L[𝕜] F` endowed with the WOT rather than the norm topology.

Along the way, we define several lemmas that are useful for proving properties of a topology that is induced by a bundled hom.

Note: I've defined the global notation `E →WOT[𝕜] F` for the type copy; I don't think this is likely to collide with anything else, but I can make it scoped if this is deemed preferable.

---


Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,276 @@
...
+instance instAddCommGroup : AddCommGroup (E →WOT[𝕜] F) :=
+  inferInstanceAs <| AddCommGroup (E →L[𝕜] F)
+
+unseal ContinuousLinearMapWOT in
+instance instModule : Module 𝕜 (E →WOT[𝕜] F) := inferInstanceAs <| Module 𝕜 (E →L[𝕜] F)
+
+variable (𝕜) (E) (F)
+
+unseal ContinuousLinearMapWOT in
+/-- The equivalence that sends a continuous linear map to the type copy endowed with the
```
j-loreaux, line 83:
```suggestion
/-- The linear equivalence that sends a continuous linear map to the type copy endowed with the
```

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,276 @@
...
+  coe f :=  ((ContinuousLinearMap.toWOT 𝕜 E F).symm f : E → F)
+  coe_injective' := by intro; simp
+
+instance instContinuousLinearMapClass : ContinuousLinearMapClass (E →WOT[𝕜] F) 𝕜 E F where
+  map_add f x y := by simp only [DFunLike.coe]; simp
+  map_smulₛₗ f r x := by simp only [DFunLike.coe]; simp
+  map_continuous f := ContinuousLinearMap.continuous ((ContinuousLinearMap.toWOT 𝕜 E F).symm f)
+
+lemma _root_.ContinuousLinearMap.toWOT_apply {A : E →L[𝕜] F} {x : E} :
+    ((ContinuousLinearMap.toWOT 𝕜 E F) A) x = A x := rfl
```
j-loreaux, line 99: Are you thinking we don't want this to be `@[simp]` because it switches from WOT to CLM and we don't want to break the barrier? If so, you can just resolve this.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,276 @@
...
+
+unseal ContinuousLinearMapWOT in
+lemma ext_iff {A B : E →WOT[𝕜] F} : A = B ↔ ∀ x, A x = B x := ContinuousLinearMap.ext_iff
+
+lemma ext_dual_iff {A B : E →WOT[𝕜] F} :
+    A = B ↔ ∀ x (y : F⋆), y (A x) = y (B x) := by
```
j-loreaux, line 105: Don't we want an actual `@[ext]` lemma too? If so, which one do we want? The `FunLike` one should be copied over in either case. (I think we maybe want `ext_dual_iff.mpr` to be the `@[ext]` lemma for this type).

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,276 @@
...
+  toFun := fun A ⟨x, y⟩ => y (A x)
+  map_add' := fun x y => by ext; simp
+  map_smul' := fun x y => by ext; simp
+
+/-- The weak operator topology is the coarsest topology such that `fun A => y (A x)` is
+continuous for all `x, y`. -/
+instance instTopologicalSpace : TopologicalSpace (E →WOT[𝕜] F) :=
+  .induced (inducingFn _ _ _) Pi.topologicalSpace
+
+lemma continuous_inducingFn : Continuous (inducingFn 𝕜 E F) :=
```
j-loreaux, line 149:
```suggestion
@[fun_prop]
lemma continuous_inducingFn : Continuous (inducingFn 𝕜 E F) :=
```

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,276 @@
...
+  continuous_induced_dom
+
+lemma continuous_dual_apply (x : E) (y : F⋆) : Continuous fun (A : E →WOT[𝕜] F) => y (A x) := by
+  refine (continuous_pi_iff.mp continuous_inducingFn) ⟨x, y⟩
+
+lemma continuous_of_dual_apply_continuous {α : Type*} [TopologicalSpace α] {g : α → E →WOT[𝕜] F}
```
j-loreaux, line 155:
```suggestion
@[fun_prop]
lemma continuous_of_dual_apply_continuous {α : Type*} [TopologicalSpace α] {g : α → E →WOT[𝕜] F}
```

Mathlib/Analysis/InnerProductSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,49 @@
...
+-/
+
+open scoped Topology
+
+namespace ContinuousLinearMapWOT
+
+variable {𝕜 : Type*} {E : Type*} {F : Type*} [RCLike 𝕜] [AddCommGroup E] [TopologicalSpace E]
+  [Module 𝕜 E] [NormedAddCommGroup F] [InnerProductSpace 𝕜 F]
+
+lemma ext_inner_iff {A B : E →WOT[𝕜] F} : A = B ↔ ∀ x y, ⟪y, A x⟫_𝕜 = ⟪y, B x⟫_𝕜 := by
```
j-loreaux, line 25: I think we may also want `ext_inner_iff.mpr` to be an `@[ext]` lemma.

j-loreaux (commented):
This is a first pass. I may have more thoughts about simplifying the proof of the `hasBasis` theorem later. This mostly looks really nice though, thanks!

Mathlib/Analysis/InnerProductSpace/WeakOperatorTopology.lean
dupuisf: Does `ext` behave well where there are competing `ext` lemmas? If we have both this and `ext_dual_iff.mpr` tagged as `ext`, they would both apply in the Hilbert space case.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
dupuisf: Yeah I'm not too sure either whether we want this to be a simp lemma or not. Let's just leave it like this, we can always add the tag later if not having it turns out to be annoying.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
dupuisf: I agree `ext_dual_iff.mpr` seems like the most natural choice. I've added it as `ext_dual`.

dupuisf (commented):
Thanks @j-loreaux ! Regarding the `HasBasis` proof, note that I mostly cribbed it from `LinearMap.hasBasis_weakBilin`; if you have ideas to simplify it, they probably apply there as well!

j-loreaux:
> Does `ext` behave well where there are competing `ext` lemmas? If we have both this and `ext_dual_iff.mpr` tagged as `ext`, they would both apply in the Hilbert space case.

I think this was a problem at one point, but I don't think it is any longer. I believe we have resurrected the Lean 3 behavior where the one declared later has priority, so in this case, the inner product version. You could test it though.

Mathlib/Analysis/InnerProductSpace/WeakOperatorTopology.lean
j-loreaux: I think this was a problem at one point, but I don't think it is any longer. I believe we have resurrected the Lean 3 behavior where the one declared later has priority, so in this case, the inner product version. You could test it though.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,292 @@
...
+      simp only [Set.empty_pi, Set.preimage_univ, Set.subset_univ, and_true_iff]
+      exact ⟨(p 0).ball 0 1, p.basisSets_singleton_mem 0 one_pos⟩
+  · intro U (hU : U ∈ p.basisSets)
+    rw [SeminormFamily.basisSets_iff] at hU
+    obtain ⟨s, r, hr, hU⟩ := hU
+    rw [hU]
+    refine ⟨(s, fun _ => r), ⟨by simp only [s.finite_toSet], fun y _ => hr⟩, fun x hx => ?_⟩
+    simp only [Set.mem_preimage, Set.mem_pi, Finset.mem_coe, mem_ball_zero_iff] at hx
+    simp only [_root_.id, Seminorm.mem_ball, sub_zero]
+    exact Seminorm.finset_sup_apply_lt hr fun y hy => hx y hy
```
j-loreaux, line 260: My conclusion is that this lemma is missing:
```lean
/-- If a family of seminorms is continuous, then their basis sets are neighborhoods of zero. -/
lemma SeminormFamily.basisSets_mem_nhds {𝕜 E ι : Type*} [NormedField 𝕜]
    [AddCommGroup E] [Module 𝕜 E] [TopologicalSpace E] (p : SeminormFamily 𝕜 E ι)
    (hp : ∀ i, Continuous (p i)) (U : Set E) (hU : U ∈ p.basisSets) : U ∈ 𝓝 (0 : E) := by
  obtain ⟨s, r, hr, rfl⟩ := p.basisSets_iff.mp hU
  clear hU
  refine Seminorm.ball_mem_nhds ?_ hr
  classical
  induction s using Finset.induction_on with
  | empty => simpa using continuous_zero
  | insert _ hs => simpa using hp _ |>.sup <| hs
```
Note that the induction above is actually a bit necessary because the `empty` case won't hold in general (see `Continuous.finset_sup` for the condition that the codomain of the functions satisfies `OrderBot`).

With this, and the `lift` tactic, and various other small cleanups, we get:
```suggestion
lemma hasBasis_seminorms : (𝓝 (0 : E →WOT[𝕜] F)).HasBasis (seminormFamily 𝕜 E F).basisSets id := by
  let p := seminormFamily 𝕜 E F
  rw [nhds_induced, nhds_pi]
  simp [map_zero, zero_apply]
  have h := Filter.hasBasis_pi (fun _ : (E × F⋆) ↦ Metric.nhds_basis_ball (x := 0)) |>.comap
    (inducingFn 𝕜 E F)
  refine h.to_hasBasis' ?_ ?_
  · rintro ⟨s, U₂⟩ ⟨hs, hU₂⟩
    lift s to Finset (E × F⋆) using hs
    by_cases hU₃ : s.Nonempty
    · refine ⟨(s.sup p).ball 0 <| s.inf' hU₃ U₂, p.basisSets_mem _ <| (Finset.lt_inf'_iff _).2 hU₂,
        fun x hx y hy => ?_⟩
      simp only [Set.mem_preimage, Set.mem_pi, mem_ball_zero_iff]
      rw [id, Seminorm.mem_ball_zero] at hx
      have hp : p y ≤ s.sup p := Finset.le_sup hy
      refine lt_of_le_of_lt (hp x) (lt_of_lt_of_le hx ?_)
      exact Finset.inf'_le _ hy
    · rw [Finset.not_nonempty_iff_eq_empty.mp hU₃]
      exact ⟨(p 0).ball 0 1, p.basisSets_singleton_mem 0 one_pos, by simp⟩
  · suffices ∀ U ∈ p.basisSets, U ∈ 𝓝 (0 : E →WOT[𝕜] F) by
      rw [nhds_induced, nhds_pi] at this
      simpa [map_zero, zero_apply]
    exact p.basisSets_mem_nhds fun ⟨x, y⟩ ↦ continuous_dual_apply x y |>.norm
```

Mathlib/Analysis/InnerProductSpace/WeakOperatorTopology.lean
dupuisf: I had a quick look at the `ext` code, and it uses a discrimination tree (as I assumed), so the order of declaration probably doesn't matter. However, we can set priorities on ext lemmas so I'll just do that.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
dupuisf: Any idea where that new lemma could go? It can't go in `WithSeminorms.lean` because it uses `Continuous.sup`.

dupuisf: Alright, I modified the new lemma to avoid additional imports.

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
```diff
@@ -0,0 +1,283 @@
...
+
+instance instT3Space : T3Space (E →WOT[𝕜] F) := embedding_inducingFn.t3Space
+
+instance instContinuousAdd : ContinuousAdd (E →WOT[𝕜] F) := .induced (inducingFn 𝕜 E F)
+instance instContinuousNeg : ContinuousNeg (E →WOT[𝕜] F) := .induced (inducingFn 𝕜 E F)
+instance instContinuousSMul : ContinuousSMul 𝕜 (E →WOT[𝕜] F) := .induced (inducingFn 𝕜 E F)
+
+instance instTopologicalAddGroup : TopologicalAddGroup (E →WOT[𝕜] F) where
+
+instance instUniformSpace : UniformSpace (E →WOT[𝕜] F) := .comap (inducingFn 𝕜 E F) inferInstance
```
j-loreaux, line 207: What about completeness? This should hold at least when `E` and `F` are Hilbert spaces.

j-loreaux (commented):
If you want, you can save completeness for a future PR.

bors d+

Mathlib/Analysis/NormedSpace/WeakOperatorTopology.lean
dupuisf: Yeah, I think I'll leave that for a future PR!

PR #14185, opened 2024-06-27
[Merged by Bors] - feat(MeasureTheory/Constructions/Pi): add `measurePreserving_pi`

xroblot:
See the Zulip [thread](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/MeasureTheory.2EmeasurePreserving_pi/near/447399178)

---


Mathlib/MeasureTheory/Constructions/Pi.lean
```diff
@@ -970,6 +970,23 @@ theorem volume_preserving_piFinsetUnion (α : ι → Type*) [DecidableEq ι] {s
     MeasurePreserving (MeasurableEquiv.piFinsetUnion α h) volume volume :=
   measurePreserving_piFinsetUnion h (fun _ ↦ volume)
 
+theorem measurePreserving_pi {β : ι → Type*} [∀ i, MeasurableSpace (β i)]
+    (ν : (i : ι) → Measure (β i)) [∀ i, SigmaFinite (ν i)] {f : (i : ι) → (α i) ≃ᵐ (β i)}
```
sgouezel, line 974: You are assuming that the `f i` are measurable equivalences, but this should be true for any map, right?

xroblot: Right. I got a bit lazy here. It's fixed in the new version (and the proof is actually shorter). Thanks!

 I also added a `volume` version of the result since it seems that many results in this file are stated both ways.

Mathlib/MeasureTheory/Constructions/Pi.lean
```diff
@@ -970,6 +970,26 @@ theorem volume_preserving_piFinsetUnion (α : ι → Type*) [DecidableEq ι] {s
     MeasurePreserving (MeasurableEquiv.piFinsetUnion α h) volume volume :=
   measurePreserving_piFinsetUnion h (fun _ ↦ volume)
 
+theorem measurePreserving_pi {β : ι → Type*} [∀ i, MeasurableSpace (β i)]
```
sgouezel, line 973: Can't you deduce the sigma-finiteness of `mu i` from the one of `nu i` instead of assuming it, thanks to [MeasureTheory.MeasurePreserving.sigmaFinite](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Dynamics/Ergodic/MeasurePreserving.html#MeasureTheory.MeasurePreserving.sigmaFinite)? Same thing for the next lemma.

xroblot: Indeed. Thanks!

urkud:
Thanks! :tada:
bors merge

PR #14187, opened 2024-06-27
[Merged by Bors] - feat(CategoryTheory/Shift): more compatibilities for NatTrans.CommShift

joelriou:
In this file, we show various compatibilities for the commutation of natural transformations with shifts. In particular, if `F`, `G`, `H` are composable functors which commute with shifts on the categories, then the associator isomorphism
`(F ⋙ G) ⋙ H ≅ F ⋙ (G ⋙ H)` commutes with the shifts.

---

This shall be used in order to obtain more compatibilities for the triangulated functor `Functor.mapDerivedCategory` defined in #14153.


semorrison:
bors merge

PR #14193, opened 2024-06-27
[Merged by Bors] - doc(RingTheory/Flat/Algebra) fix docstring

jaclynlang:
Switch the roles of R and S in the docstring on line 62

---

### DIFF
--- a/Mathlib/RingTheory/Flat/Algebra.lean
+++ b/Mathlib/RingTheory/Flat/Algebra.lean
@@ -59,7 +59,7 @@ theorem isBaseChange [Algebra R S] (R' : Type w) (S' : Type t) [CommRing R'] [Co
 
 end Algebra.Flat
 
-/-- A `RingHom` is flat if `S` is flat as an `R` algebra. -/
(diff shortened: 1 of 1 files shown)


Mathlib/RingTheory/Flat/Algebra.lean
```diff
@@ -59,7 +59,7 @@ theorem isBaseChange [Algebra R S] (R' : Type w) (S' : Type t) [CommRing R'] [Co
 
 end Algebra.Flat
 
-/-- A `RingHom` is flat if `R` is flat as an `S` algebra. -/
+/-- A `RingHom` is flat if `S` is flat as an `R` algebra. -/
```
kbuzzard, line 62:
```suggestion
/-- A ring homomorphism `f : R →+* S` is flat if `S` is flat as an `R` algebra. -/
```

It's best to include the roles of the variables, because when someone hovers over RingHom.Flat in VS Code for their ring homomorphism from A to B, or maybe from S to R, they'll see you talking about R and S without any context.

kbuzzard (commented):
Thanks! Please specify what R and S are to make the docstring even better! Then feel free to merge.

bors d+

PR #14194, opened 2024-06-27
feat(Order/Monotone/Basic): add `prod_map_iterate`

pitmonticone:
This PR adds `prod_map_iterate`.

I don't know what is supposed to be the optimal location for this theorem.

`#find_home! prod_map_iterate` suggested this file, but I am not sure where exactly it should be put within that file.

Co-authored-by: @D-Thomine 

---

### DIFF
--- a/Mathlib/Order/Monotone/Basic.lean
+++ b/Mathlib/Order/Monotone/Basic.lean
@@ -1256,3 +1256,10 @@ theorem const_strictMono [Nonempty β] : StrictMono (const β : α → β → α
 #align function.const_strict_mono Function.const_strictMono
 
 end Function
+
+theorem prod_map_iterate {X Y : Type _} (S : X → X) (T : Y → Y) (n : ℕ) :
+    (Prod.map S T)^[n] = Prod.map S^[n] T^[n] := by
+  induction' n with n hn
+  · rw [Function.iterate_zero, Function.iterate_zero, Function.iterate_zero, Prod.map_id]
+  · rw [Function.iterate_succ, hn, Prod.map_comp_map, ← Function.iterate_succ,
+    ← Function.iterate_succ]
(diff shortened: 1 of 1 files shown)


Mathlib/Order/Monotone/Basic.lean
```diff
@@ -1256,3 +1256,10 @@ theorem const_strictMono [Nonempty β] : StrictMono (const β : α → β → α
 #align function.const_strict_mono Function.const_strictMono
 
 end Function
+
+theorem prod_map_iterate {X Y : Type _} (S : X → X) (T : Y → Y) (n : ℕ) :
+    (Prod.map S T)^[n] = Prod.map S^[n] T^[n] := by
+  induction' n with n hn
```
sgouezel, line 1262: What about switching from `induction'` to `induction`?

pitmonticone:
@sgouezel I don't know what is supposed to be the optimal location for this theorem.

`#find_home! prod_map_iterate` suggested this file, but I am not sure where exactly it should be put within that file.

Mathlib/Order/Monotone/Basic.lean
pitmonticone: Done. Thanks.

sgouezel:
Have you tried to ask loogle if it knows about something involving Prod.Map and Nat.iterate? Per https://loogle.lean-lang.org/?q=Prod.map%2C%20Nat.iterate, it seems that the result in this PR is already available in mathlib...

pitmonticone:
> Have you tried to ask loogle if it knows about something involving Prod.Map and Nat.iterate? Per https://loogle.lean-lang.org/?q=Prod.map%2C%20Nat.iterate, it seems that the result in this PR is already available in mathlib...

Oh, I see what happened: `Function.iterate_prod_map` had been added in `Mathlib.Dynamics.PeriodicPts` about 11 months ago, while the `prod_map_iterate` I am adding in this PR had been written more than a year ago... 

I am sorry @sgouezel, I am going to close this PR if it's ok.

sgouezel:
Yes, you can definitely close the PR.

PR #14195, opened 2024-06-27
[Merged by Bors] - feat(Logic/Embedding/Set): Injections from subtypes of disjoint sets

apnelson1:
For disjoint `s t : Set α`, there is a natural injection from `↑s ⊕ ↑t` to `α`. We provide this as an embedding, and do the same for an indexed family of pairwise disjoint sets. 

---


Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,55 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
...
+
+/-- For disjoint `s t : Set α`, the natural injection from `↑s ⊕ ↑t` to `α`. -/
+@[simps] def Disjoint.sumSubtypeEmbedding (h : Disjoint s t) : s ⊕ t ↪ α where
+  toFun := Sum.elim (↑) (↑)
+  inj' := by
+    rintro (⟨a, ha⟩ | ⟨a, ha⟩) (⟨b, hb⟩ | ⟨b, hb⟩)
+    · simp [Subtype.val_inj]
+    · simpa using h.ne_of_mem ha hb
+    · simpa using h.symm.ne_of_mem ha hb
+    simp [Subtype.val_inj]
```
YaelDillies, line 168: This could equivalently be written as
```suggestion
@[simps] def Disjoint.sumSubtypeEmbedding (h : Disjoint s t) : s ⊕ t ↪ α :=
  (Equiv.Set.union h).toEmbedding.trans (.subtype _)
```
(except for the fact that `Equiv.Set.union` takes in `s ∩ t ⊆ ∅` rather than `Disjoint s t`, which should be fixed), so maybe not worth a separate definition?

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,55 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
...
+
+@[simp] theorem Disjoint.sumSubtypeEmbedding_range {s t : Set α} (h : Disjoint s t) :
+    range h.sumSubtypeEmbedding = s ∪ t := by
+  ext
+  simp
+
+/-- For an indexed family `s : ι → Set α` of disjoint sets,
+the natural injection from the sigma-type `(i : ι) × ↑(s i)` to `α`. -/
+@[simps] def Pairwise.disjointSigmaSubtypeEmbedding {s : ι → Set α} (h : Pairwise (Disjoint on s)) :
+    (i : ι) × s i ↪ α where
```
YaelDillies, line 188: Same here with [Set.unionEqSigmaOfDisjoint](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Lattice.html#Set.unionEqSigmaOfDisjoint)

semorrison:
(@YaelDillies, it's helpful when you leave review comments requesting changes if you can add the `awaiting-author` label, so the PR comes off the queue in the meantime.)

Mathlib/Logic/Embedding/Set.lean
apnelson1: This has the advantage of not needing decidability. Also, the simp lemmas below would look quite clunky without a definition.

Mathlib/Logic/Embedding/Set.lean
apnelson1: Same response as above, except this time noncomputability.

YaelDillies:
Sorry, I do that except that here it was a quick back and forth and maybe someone else would have wanted to comment. If you see such a PR and have nothing else to comment then please add `awaiting-author`.

Mathlib/Logic/Embedding/Set.lean
YaelDillies: > the simp lemmas below would look quite clunky without a definition

That's usually not a concern since there are simp lemmas dealing with the intermediate expression




> This has the advantage of not needing decidability

That is a fair point

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,55 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
...
+  toFun x := x.2.1
+  inj' := by
+    rintro ⟨i,⟨x,hx⟩⟩ ⟨j,⟨-,hx'⟩⟩ rfl
```
YaelDillies, line 191:
```suggestion
    rintro ⟨i, x, hx⟩ ⟨j, -, hx'⟩ rfl
```

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,55 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
 #align subtype_or_equiv_symm_inr subtypeOrEquiv_symm_inr
 
 end Subtype
+
+section Disjoint
+
+variable {α ι : Type*} {s t r : Set α}
+
+/-- For disjoint `s t : Set α`, the natural injection from `↑s ⊕ ↑t` to `α`. -/
+@[simps] def Disjoint.sumSubtypeEmbedding (h : Disjoint s t) : s ⊕ t ↪ α where
```
YaelDillies, line 161: What does `simps` generate? I think you're missing `@[norm_cast] lemma Disjoint.coe_sumSubtypeEmbedding`

YaelDillies, line 161:
```suggestion
@[simps] def Function.Embedding.sumSet (h : Disjoint s t) : s ⊕ t ↪ α where
```
seems more in line with the existing `Equiv.Set.union`

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,55 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
...
+  simp
+
+@[simp] theorem Disjoint.sumSubtypeEmbedding_range {s t : Set α} (h : Disjoint s t) :
+    range h.sumSubtypeEmbedding = s ∪ t := by
+  ext
+  simp
+
+/-- For an indexed family `s : ι → Set α` of disjoint sets,
+the natural injection from the sigma-type `(i : ι) × ↑(s i)` to `α`. -/
+@[simps] def Pairwise.disjointSigmaSubtypeEmbedding {s : ι → Set α} (h : Pairwise (Disjoint on s)) :
```
YaelDillies, line 187: And this could be
```suggestion
@[simps] def Function.Embedding.sigmaSet {s : ι → Set α} (h : Pairwise (Disjoint on s)) :
```

Mathlib/Logic/Embedding/Set.lean
YaelDillies: Please show the equality with my spelling, though

Mathlib/Logic/Embedding/Set.lean
YaelDillies: Same here

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -3,19 +3,24 @@ Copyright (c) 2017 Johannes Hölzl. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Johannes Hölzl, Mario Carneiro
 -/
+import Mathlib.Data.Set.Notation
+import Mathlib.Order.SetNotation
 import Mathlib.Logic.Embedding.Basic
+import Mathlib.Logic.Pairwise
 import Mathlib.Data.Set.Image
 
+
```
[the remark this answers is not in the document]
YaelDillies: 🏓

Mathlib/Logic/Embedding/Set.lean
```diff
@@ -147,3 +152,61 @@ theorem subtypeOrEquiv_symm_inr (p q : α → Prop) [DecidablePred p] (h : Disjo
...
+@[norm_cast] lemma Function.Embedding.coe_sumSet (h : Disjoint s t) :
+    (Function.Embedding.sumSet h : s ⊕ t → α) = Sum.elim (↑) (↑) := rfl
+
+@[simp] theorem Disjoint.sumSubtypeEmbedding_preimage_inl (h : Disjoint s t) :
+    .inl ⁻¹' (Function.Embedding.sumSet h ⁻¹' r) = r ∩ s := by
+  ext
+  simp
+
+@[simp] theorem Disjoint.sumSubtypeEmbedding_preimage_inr {s t r : Set α} (h : Disjoint s t) :
+    .inr ⁻¹' (Function.Embedding.sumSet h ⁻¹' r) = r ∩ t := by
```
YaelDillies, line 179: Those brackets everywhere are unnecessary
```suggestion
    .inr ⁻¹' Function.Embedding.sumSet h ⁻¹' r = r ∩ t := by
```

YaelDillies (commented):
Looks otherwise okay

maintainer merge

riccardobrasca:
Thanks!

bors d+

Mathlib/Logic/Embedding/Set.lean
apnelson1: Doesn't work.

YaelDillies: Uh, we should fix that. Pretty sure it worked in Lean 4. But okay for now

PR #14196, opened 2024-06-27
[Merged by Bors] - chore: add exception for `Data.Set.Subset` to `noshake.json`.

apnelson1:
We add `Data.Set.Subset` to the set of files ignored by the shake linter, which gives a false positive when this file is just imported for the notation.

---

### DIFF
--- a/Mathlib.lean
+++ b/Mathlib.lean
@@ -2312,6 +2312,7 @@ import Mathlib.Data.Set.List
 import Mathlib.Data.Set.MemPartition
 import Mathlib.Data.Set.MulAntidiagonal
 import Mathlib.Data.Set.NAry
+import Mathlib.Data.Set.Notation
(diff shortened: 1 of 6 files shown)


Ruben-VandeVelde (approved):
maintainer merge

j-loreaux:
It seems a bit odd to me that we would be `noshake`ing an entire file just because some notation is defined there. Perhaps a better solution could be to separate out the notation into its own file and `noshake` that?

apnelson1:
> It seems a bit odd to me that we would be `noshake`ing an entire file just because some notation is defined there. Perhaps a better solution could be to separate out the notation into its own file and `noshake` that?

I've just done this - a new file `Data.Set.Notation` now contains the `↓∩` notation as well as the `↑s` notation for `(↑) '' s`.

semorrison:
bors merge

PR #14197, opened 2024-06-27
[Merged by Bors] - feat(Data/Matroid/Closure): matroid closure

apnelson1:
This PR defines the closure of a set in a matroid, and provides some API. 

---


Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+lemma cl_def' (M : Matroid α) (X : Set α) (hX : X ⊆ M.E := by aesop_mat) :
+    M.cl X = ⋂₀ {F | M.Flat F ∧ X ⊆ F} := by
+  nth_rw 2 [← inter_eq_self_of_subset_left hX]
+  rfl
+
+@[aesop unsafe 10% (rule_sets := [Matroid])]
+lemma cl_subset_ground (M : Matroid α) (X : Set α) : M.cl X ⊆ M.E :=
+  sInter_subset_of_mem ⟨M.ground_flat, inter_subset_right⟩
+
+lemma ground_subset_cl_iff : M.E ⊆ M.cl X ↔ M.cl X = M.E := by
```
madvorak, line 72: I'd rather write `M.E ⊆ M.cl X ↔ M.E = M.cl X`.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+  simp [M.cl_subset_ground X, subset_antisymm_iff]
+
+lemma cl_eq_sInter_of_subset (X : Set α) (hX : X ⊆ M.E := by aesop_mat) :
+    M.cl X = ⋂₀ {F : Set α | M.Flat F ∧ X ⊆ F} := by
```
madvorak, line 76: Other definitions don't specify ` : Set α` in the set-builder notation. If I were writing it, I'd add the redundant type information everywhere as it is done here.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+    (iUnion_subset (fun i ↦ M.cl_subset_cl (subset_iUnion _ _)))).antisymm ?_
+  rw [← cl_inter_ground, iUnion_inter]
+  refine' M.cl_subset_cl (iUnion_subset (fun i ↦ (M.subset_cl _).trans _))
+  rw [cl_inter_ground]
+  exact subset_iUnion (fun i ↦ M.cl (Xs i)) i
+
+lemma cl_iUnion_congr {ι : Type*} (Xs Ys : ι → Set α) (h : ∀ i, M.cl (Xs i) = M.cl (Ys i)) :
+    M.cl (⋃ i, Xs i) = M.cl (⋃ i, Ys i) := by
+  rw [← M.cl_iUnion_cl_eq_cl_iUnion]
+  simp [h, M.cl_iUnion_cl_eq_cl_iUnion]
```
madvorak, line 163: Both lines can be replaced by `simp [h, ← M.cl_iUnion_cl_eq_cl_iUnion]`.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+
+@[simp] lemma cl_empty_union_cl_eq (M : Matroid α) (X : Set α) :
+    M.cl ∅ ∪ M.cl X = M.cl X :=
+  union_eq_self_of_subset_left (M.cl_subset_cl (empty_subset _))
+
+lemma cl_insert_eq_of_mem_cl (he : e ∈ M.cl X) : M.cl (insert e X) = M.cl X := by
+  rw [← cl_insert_cl_eq_cl_insert, insert_eq_of_mem he, cl_cl]
+
+lemma mem_cl_self (M : Matroid α) (e : α) (he : e ∈ M.E := by aesop_mat) : e ∈ M.cl {e} :=
+  mem_cl_of_mem' _ rfl
```
madvorak, line 207: This use of underscore feels pointless when it can be `mem_cl_of_mem' M rfl` literally.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+All such choices come with tradeoffs; the definition we opt for satisfies `M.cl X = M.cl (X ∩ M.E)`.
+The drawback here is that the statement `X ⊆ M.cl X` requires the assumption `X ⊆ M.E`.
+But all the other properties of closure work nicely without extra assumptions;
+closure is monotone, idempotent, and `M.cl X ⊆ M.E` for all `X`.
+-/
+
+open Set
+namespace Matroid
+
+variable {α ι : Type*} {M : Matroid α} {F I J X Y B C R : Set α} {e f x y : α}
```
madvorak, line 44: `B` and `C` are not used. Neither are `I` and `J`.

From the last part, only `{e : α}` is used.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+lemma cl_iUnion_cl_eq_cl_iUnion (M : Matroid α) (Xs : ι → Set α) :
+    M.cl (⋃ i, M.cl (Xs i)) = M.cl (⋃ i, Xs i) := by
+  refine (M.cl_subset_cl_of_subset_cl
+    (iUnion_subset (fun i ↦ M.cl_subset_cl (subset_iUnion _ _)))).antisymm ?_
+  rw [← cl_inter_ground, iUnion_inter]
+  refine' M.cl_subset_cl (iUnion_subset (fun i ↦ (M.subset_cl _).trans _))
+  rw [cl_inter_ground]
+  exact subset_iUnion (fun i ↦ M.cl (Xs i)) i
+
+lemma cl_iUnion_congr {ι : Type*} (Xs Ys : ι → Set α) (h : ∀ i, M.cl (Xs i) = M.cl (Ys i)) :
```
madvorak, line 160: Local argument `{ι : Type*}` is redundant as it is a variable for the whole file.

madvorak, line 160: Arguments `Xs` and `Ys` could be implicit.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+
+@[aesop unsafe 10% (rule_sets := [Matroid])]
+lemma mem_ground_of_mem_cl (he : e ∈ M.cl X) : e ∈ M.E := (M.cl_subset_ground _) he
+
+lemma cl_iUnion_cl_eq_cl_iUnion (M : Matroid α) (Xs : ι → Set α) :
+    M.cl (⋃ i, M.cl (Xs i)) = M.cl (⋃ i, Xs i) := by
+  refine (M.cl_subset_cl_of_subset_cl
+    (iUnion_subset (fun i ↦ M.cl_subset_cl (subset_iUnion _ _)))).antisymm ?_
+  rw [← cl_inter_ground, iUnion_inter]
+  refine' M.cl_subset_cl (iUnion_subset (fun i ↦ (M.subset_cl _).trans _))
```
madvorak, line 156: Why not `refine M.cl_subset_cl (iUnion_subset (fun i ↦ (M.subset_cl _).trans ?_))` ?

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,207 @@
...
+@[aesop unsafe 20% (rule_sets := [Matroid])]
+lemma Flat.subset_ground (hF : M.Flat F) : F ⊆ M.E :=
+  hF.2
+
+@[simp] lemma ground_flat (M : Matroid α) : M.Flat M.E :=
+  ⟨fun _ _ _ ↦ Basis.subset_ground, Subset.rfl⟩
+
+/-- The closure of a subset of the ground set is the intersection of the flats containing it.
+  A set `X` that doesn't satisfy `X ⊆ M.E` has the junk value `M.cl X := M.cl (X ∩ M.E)`. -/
+def cl (M : Matroid α) (X : Set α) : Set α := ⋂₀ {F | M.Flat F ∧ X ∩ M.E ⊆ F}
```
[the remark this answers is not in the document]
madvorak: I didn't take into account renaming constants containing `cl` in their name.

Mathlib/Data/Matroid/Closure.lean
apnelson1: I think that the current direction is better for rewriting using the RHS - it is far more like to `rw [h : M.cl X = M.E]` than the other way round.

Mathlib/Data/Matroid/Closure.lean
apnelson1: I tend to prefer things without the type annotation here - informally we are taking the intersection of the 'set of flats of M such that ...' so the type annotation is implicit in normal speech because of `M.Flat F`.

Mathlib/Data/Matroid/Closure.lean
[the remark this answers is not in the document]
apnelson1: The notation is to line up with the normal pen-and-paper notation for closure.

Mathlib/Data/Matroid/Closure.lean
madvorak: OK, ignore my comment then. I was thinking about readability of the statement rather than its uses. Your point is more relevant.

Mathlib/Data/Matroid/Closure.lean
madvorak: Alright! I was just checking it is intentional here.

Mathlib/Data/Matroid/Closure.lean
apnelson1: Thanks! This is because the PR is just the beginning of the repo file, which requires more in its full version.

madvorak:
LGTM.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,199 @@
...
+-/
+
+open Set
+namespace Matroid
+
+variable {ι α : Type*} {M : Matroid α} {F X Y R : Set α} {e : α}
+
+/-- A flat is a maximal set having a given basis  -/
+def Flat (M : Matroid α) (F : Set α) : Prop :=
+  (∀ ⦃I X⦄, M.Basis I F → M.Basis I X → X ⊆ F) ∧ F ⊆ M.E
```
jcommelin, line 48:
```suggestion
@[mk_iff]
structure Flat (M : Matroid α) (F : Set α) : Prop where
  subset_of_basis_of_basis : ∀ ⦃I X⦄, M.Basis I F → M.Basis I X → X ⊆ F
  subset_ground : F ⊆ M.E
```

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,199 @@
...
+
+@[aesop unsafe 20% (rule_sets := [Matroid])]
+lemma Flat.subset_ground (hF : M.Flat F) : F ⊆ M.E :=
+  hF.2
```
jcommelin, line 52:
```suggestion
attribute [aesop unsafe 20% (rule_sets := [Matroid])] Flat.subset_ground
```

Mathlib/Data/Matroid/Closure.lean
jcommelin: Should this be a simp lemma?

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,199 @@
...
+
+If `X : Set α` satisfies `X ⊆ M.E`, then it is clear how `M.cl X` should be defined.
+But `M.cl X` needs to be defined for all `X : Set α`, so a convention is needed for how `M.cl`
+handles sets containing junk elements outside `M.E`. Ideally, this convention should minimize
+the need to keep track of which sets are actually contained in the ground set.
+
+All such choices come with tradeoffs; the definition we opt for satisfies `M.cl X = M.cl (X ∩ M.E)`.
+The drawback here is that the statement `X ⊆ M.cl X` requires the assumption `X ⊆ M.E`.
+But all the other properties of closure work nicely without extra assumptions;
+closure is monotone, idempotent, and `M.cl X ⊆ M.E` for all `X`.
```
jcommelin, line 38: Thinking out loud for a bit...

If you choose to add `X \ M.E` to the closure, then it is still monotone and idempotent. You get `X ⊆ M.cl X` without assumptions, but now `M.cl X ⊆ M.E` will need an assumption.

On the other hand, it makes `M.cl` an example of a [ClosureOperator](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Closure.html#ClosureOperator). That would give you some API for free.

Naive question: is `M.cl X` an example of a flat? If so, I suspect that `M.cl` is part of a Galois correspondence (even Galois insertion) between flats and subsets. That would also give you a bunch of API for free.

Happy to hear your thoughts.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,199 @@
...
+
+@[simp] lemma cl_union_cl_left_eq (M : Matroid α) (X Y : Set α) :
+    M.cl (M.cl X ∪ Y) = M.cl (X ∪ Y) := by
+  rw [← cl_cl_union_cl_eq_cl_union, cl_cl, cl_cl_union_cl_eq_cl_union]
+
+@[simp] lemma cl_insert_cl_eq_cl_insert (M : Matroid α) (e : α) (X : Set α) :
+    M.cl (insert e (M.cl X)) = M.cl (insert e X) := by
+  simp_rw [← singleton_union, cl_union_cl_right_eq]
+
+@[simp] lemma cl_union_cl_empty_eq (M : Matroid α) (X : Set α) :
```
jcommelin, line 187: Is `M.cl ∅ = ∅` true? If so, that seems like a good simp lemma?

jcommelin (commented):
Some naive comments by someone who knows a bit of Lean/mathlib but nothing about matroids. My apologies for any stupid questions.

One more consideration: in other parts of mathlib we write out the full name `closure`. Here you have abbreviated to `cl`. I think that it would be good for uniformity to use `closure` everywhere.

Mathlib/Data/Matroid/Closure.lean
apnelson1: No - for instance, if `M` is the matroid arising from a function `v : α → V` for a vector space `V`, then `M.cl ∅ = v ⁻¹' {0}`.

jcommelin: Thanks! That is a helpful example.

Mathlib/Data/Matroid/Closure.lean
apnelson1: Thanks for this - I wasn't familiar with the `GaloisConnection` and `ClosureOperator` parts of mathlib, and gave this quite a lot of thought. 

Mathematically, you're absolutely on the money - if we have `M.E = univ`, then `M.closure` would give a `ClosureOperator` on `Set α` for which `Flat ↔ IsClosed`, and we thus have a Galois insertion with the types `Set α` and `{F // M.Flat F}`. 

But because we may have `M.E ⊂ univ` and `IsClosed univ` is always true, we can't have all three of the following at once : 
(1) `M.E ⊂ univ`,
(2) every `Flat` is a subset of the ground set,
(3) `M.closure` is a `ClosureOperator` on `Set α` for which `Flat ↔ IsClosed`

For instance, if we included `X \ M.E` in `M.cl X`, then we would get a `ClosureOperator`, but `IsClosed` would not correspond to `Matroid.Flat`, since (for example) the set `univ` would be closed but not a flat. Perhaps having this `ClosureOperator` would still be somewhat useful, but losing the property that `M.cl X ⊆ M.E` is not great for `aesop_mat`, and I think in practice would add a lot of extra assumptions to lemmas.  

I believe that (1) and (2) are important enough for the design of matroids that we have to give up (3). 

However, when it comes to piggybacking off the `ClosureOperator` and `GaloisInsertion` API, we can do something using subtypes. 
See `Matroid.subtypeClosure` as defined below. It allowed me to golf the proof of `closure_iUnion_closure_eq_closure_iUnion`,
and I think it'll be nice for making the `Flat` API simpler when we get there. 

That all said, after pushing, I will play around with including `X \ M.E` in `M.cl X` to see what it entails in terms of extra assumptions.

Mathlib/Data/Matroid/Closure.lean
```diff
@@ -0,0 +1,246 @@
...
+    (iInter_subset _ (Classical.arbitrary _)).trans (hFs _).subset_ground⟩
+  obtain ⟨J, hIJ, hJ⟩ := hI.indep.subset_basis_of_subset (hI.subset.trans (iInter_subset _ i))
+  refine' subset_union_right.trans ((hFs i).1 (X := Fs i ∪ X) hIJ _)
+  convert hIJ.basis_union (hIX.basis_union_of_subset hIJ.indep hJ) using 1
+  rw [← union_assoc, union_eq_self_of_subset_right hIJ.subset]
+
+/-- The property of being a flat gives rise to a `ClosureOperator` on the subsets of `M.E`,
+in which the `IsClosed` sets correspond to `Flat`s.
+(We can't define such an operator on all of `Set α`,
+since this would incorrectly force `univ` to always be a flat.)-/
```
jcommelin, line 72:
```suggestion
since this would incorrectly force `univ` to always be a flat.) -/
```

Mathlib/Data/Matroid/Closure.lean
jcommelin: Thanks for writing down these reflections! Would you mind incorporating (parts of) this comment into the "Implementation notes" in the module docstring?

After that (and once CI passes), I think this PR is ready to be merged.

apnelson1: Great, thanks. At the moment I'm experimenting with the `X ⊆ M.cl X` design to see how much havoc it wreaks in my larger repo. If that turns out to be 'a lot', I'll go ahead with what you say.

jcommelin: One other option is to have both, where the current version would be consider "canonical" and the other is only there as implementation detail for setting up API. Because I do think that the other would give you a closure operator and a galois correspondence.
It could even be there only for illustrative purposes.

jcommelin (commented):
Thanks :tada:

bors merge

mattrobball:
bors r- 

Please merge master and fix the issues. Thanks.

apnelson1:
> bors r-
> 
> Please merge master and fix the issues. Thanks.

Sorry about that! I think we're now good to go.

mattrobball:
Not your fault when the world shifts under you. Thanks.

bors merge

PR #14201, opened 2024-06-27
[Merged by Bors] - chore(AddConstMap): don't extend `FunLike`

urkud:
Instead, take it as an argument as other bundled homs do.
Also change the `npow` field in the `Monoid` instance so that `coe_pow` is a `rfl` lemma now.

---


dupuisf:
Thanks!

bors r+

PR #14217, opened 2024-06-28
[Merged by Bors] - chore(Data/PNat): drop `bit*` lemmas

urkud:
Add `OfNat` lemmas instead.

---


semorrison:
bors merge

PR #14229, opened 2024-06-28
[Merged by Bors] - chore: remove conditional dependency on doc-gen

kim-em:
Mathlib's documentation generation CI uses a downstream repository, which just has a conventional `require` on both Mathlib and docgen.

As far as I'm aware (please correct me!) this isn't actually needed by anyone.

We are very keen to get rid of `meta if`, move to `lakefile.toml`s, and avoid instructing users to ever use `lake -R`.

If this change works, I hope that the downstream mathematical repositories that are also using docgen can all follow suit.

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -18,9 +18,6 @@ package mathlib where
 ## Mathlib dependencies on upstream projects.
 -/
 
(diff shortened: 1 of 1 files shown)


fpvandoorn:
LGTM

bors merge

PR #14232, opened 2024-06-28
[Merged by Bors] - feat(FieldTheory/Separable): add elementwise `IsSeparable` and rename the old to `Algebra.IsSeparable`

jjdishere:
Make `IsSeparable` more aligned with `IsIntegral`. Rename previous `IsSeparable` into `Algebra.IsSeparable` (like `Algebra.IsIntegral`), add notation `IsSeparable` for separablity of elements (like `IsIntegral`).

Shorten statements using the new defined `IsSeparable`. Rename some shortened theorems.

---


riccardobrasca:
!bench

Mathlib/FieldTheory/Separable.lean
```diff
@@ -555,6 +561,14 @@ variable (F K : Type*) [CommRing F] [Ring K] [Algebra F K]
 -- Note that right now a Galois extension (class `IsGalois`) is defined to be an extension which
 -- is separable and normal, so if the definition of separable changes here at some point
 -- to allow non-algebraic extensions, then the definition of `IsGalois` must also be changed.
+
+/--
+An element `x` of an algebra `A` over a commutative ring `R` is said to be *separable*, if its
+minimal polynamial over `R` is separable. Note that the minimal polynomial of any element not
+integral over `R` is defined to be `0`, which is not a separable polynomial.
+-/
+notation " IsSeparable " F:max x:max => (Polynomial.Separable (minpoly F x))
```
erdOne, line 570: I personally think this deserves to be a def. Among other things, it unlocks dot notation.

riccardobrasca: Good point, let's make it a `def`.

jjdishere: Good! I am working on this. I am also trying to decompose many theorems in `Algebra.IsSeparable` into a theorem of `IsSeparable` and a theorem using the previous one directly. This takes more time that I thought.

riccardobrasca: What you can do is to open another PRs. Here you are just renaming stuff, right? I think it is better to not do anything else in this PR.

jjdishere: I see. Thank you!

jjdishere:
This PR only contains renaming now. I'll open a new PR adding lemmas of `IsSeparable`.

PR #14233, opened 2024-06-28
[Merged by Bors] - perf (Algebra.AddConstMap): scope simp lemmas

mattrobball:
The keys for these `simp` lemmas will match quite a bit. To avoid this, we scope them to the `AddConstMapClass` namespace so that you only have them when you need them and not when you don't.

---


mattrobball:
!bench

mattrobball:
Note this is over 1 minute is savings on the very fast benchmark server

dupuisf:
Thanks a lot for figuring this out!

bors r+

PR #14234, opened 2024-06-28
[Merged by Bors] - feat(RingTheory/Flat/Basic): a module is flat iff tensoring is left exact (non-categorical formulation)

jjaassoonn:
This is stated in terms of `Function.Exact`

---


Mathlib/RingTheory/Flat/Basic.lean
```diff
@@ -269,6 +271,88 @@ lemma iff_lTensor_preserves_injective_linearMap [Small.{v} R] :
...
+        (LinearMap.exact_iff.mp exact).symm
+  have inj0 : Function.Injective ι := by simpa [ι] using Subtype.val_injective
+  have eq0 : g = ι.comp π := by aesop
+
+  suffices exact1 : Function.Exact (f.rTensor M) (π.rTensor M) by
+    rw [eq0, rTensor_comp]
+    apply Function.Exact.comp_injective (exact := exact1)
+      (inj := iff_rTensor_preserves_injective_linearMap R M |>.mp flat _ inj0)
+      (h0 := map_zero _)
+
```
jjaassoonn, line 319: We could reuse the proof in `lTensor_exact`, but I think it is faster to just copy and paste

Mathlib/Algebra/Exact.lean
```diff
@@ -51,6 +51,14 @@ lemma Exact.of_comp_of_mem_range [Zero P] (h1 : g ∘ f = 0)
   fun y => Iff.intro (h2 y) <|
     Exists.rec ((forall_apply_eq_imp_iff (p := (g · = 0))).mpr (congrFun h1) y)
 
+lemma Exact.comp_injective [Zero P] [Zero P'] (exact : Exact f g)
```
erdOne, line 54: We have `Injective.comp_exact_iff_exact`. Do you need the non-linearMap version?

Mathlib/RingTheory/Flat/Basic.lean
```diff
@@ -269,6 +271,86 @@ lemma iff_lTensor_preserves_injective_linearMap [Small.{v} R] :
       (L : N →ₗ[R] N'), Function.Injective L → Function.Injective (L.lTensor M) := by
   simp_rw [iff_rTensor_preserves_injective_linearMap, LinearMap.lTensor_inj_iff_rTensor_inj]
 
+variable (R M) in
+lemma lTensor_exact [Small.{v} R] [flat : Flat R M] ⦃N N' N'' : Type v⦄
```
erdOne, line 275: I find it strange that these universe constraints exists. I thought the main purpose of the nonstandard definitions is to avoid these.

Mathlib/Algebra/Exact.lean
jjaassoonn: Not really, but somebody might do. I am not sure if I should delete it

Mathlib/RingTheory/Flat/Basic.lean
jjaassoonn: This universe constraint is harder to avoid, it comes from Baer's criterion where we need to lift every ideal to a module. So if module lives in a universe that is smaller than the ring, Baer's criterion might not be true. @alreadydone has thought a lot about this.

erdOne: I think the only explicit variables should be `exact` and `M`. Same in `rTensor_exact`.

Mathlib/Algebra/Exact.lean
riccardobrasca: I think we can keep it.

riccardobrasca (approved):
LGTM.

riccardobrasca:
Thanks!

bors merge

PR #14170, opened 2024-06-26
[Merged by Bors] - feat(CategoryTheory): Split equalizers

mckoen:
Defines what it means for a triple of morphisms `f g : X ⟶ Y`, `ι : W ⟶ X` to be a split equalizer. In addition, shows that every split equalizer is an equalizer and absolute.

This dualises everything in [Mathlib.CategoryTheory.Limits.Shapes.SplitCoequalizer](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Limits/Shapes/SplitCoequalizer.html).

This contribution was created as part of the AIM workshop "Formalizing Algebraic Geometry" in June 2024.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14236, opened 2024-06-28
[Merged by Bors] - feat(CategoryTheory): (co)limits in the category of coalgebras

mckoen:
Shows that the forgetful functor `forget T : Coalgebra T ⥤ C` for a comonad `T : C ⥤ C` creates colimits and creates any limits which `T` preserves. This is used to show that `Coalgebra T` has any colimits which `C` has, and any limits which `C` has and `T` preserves. This is generalised to the case of a comonadic functor `D ⥤ C`.

Dualises everything currently in [Mathlib.CategoryTheory.Monad.Limits](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Monad/Limits.html).

This contribution was created as part of the AIM workshop "Formalizing Algebraic Geometry" in June 2024.

---

- [x] depends on: #14170 


Mathlib/CategoryTheory/Monad/Adjunction.lean
```diff
@@ -361,4 +396,10 @@ instance (priority := 100) monadicOfReflective [Reflective R] :
   eqv := { full := Reflective.comparison_full _ }
 #align category_theory.monadic_of_reflective CategoryTheory.monadicOfReflective
 
+/-- Any coreflective inclusion has a comonadic left adjoint. -/
```
dagurtomas, line 399: Is this the same theorem in [Riehl]? In any case, can you add a reference to [Riehl] like above?

Mathlib/CategoryTheory/Monad/Limits.lean
```diff
@@ -406,4 +408,330 @@ noncomputable def leftAdjointPreservesTerminalOfReflective (R : D ⥤ C) [Reflec
...
+@[simps]
+def γ : D ⋙ T.forget ⟶ D ⋙ T.forget ⋙ ↑T  where app j := (D.obj j).a
+
+/-- (Impl) This new cocone is used to construct the coalgebra structure -/
+@[simps! ι_app]
+def newCocone : Cocone (D ⋙ forget T) where
+  pt := T.obj c.pt
+  ι := γ D ≫ whiskerRight c.ι (T : C ⥤ C) ≫ (Functor.constComp J _ (T : C ⥤ C)).hom
+
+/-- The coalgebra structure which will be the apex of the new colimit cone for `D`. -/
```
dagurtomas, line 430: Isn't it called something other than "apex" for cocones? Alternatively, you can use the mathlib terminology "cocone point".

Mathlib/CategoryTheory/Monad/Limits.lean
```diff
@@ -406,4 +408,330 @@ noncomputable def leftAdjointPreservesTerminalOfReflective (R : D ⥤ C) [Reflec
...
+
+/-- If `C` has limits then any coreflective subcategory has limits. -/
+theorem hasLimits_of_coreflective (R : D ⥤ C) [Coreflective R] [HasLimitsOfSize.{v, u} C] :
+    HasLimitsOfSize.{v, u} D :=
+  ⟨fun _ => hasLimitsOfShape_of_coreflective R⟩
+
+/-- The coreflector always preserves terminal objects. Note this in general doesn't apply to any
+other colimit.
+-/
+noncomputable def rightAdjointPreservesTerminalOfCoreflective (R : D ⥤ C) [Coreflective R] :
```
dagurtomas, line 718:
```suggestion
/-- The coreflector always preserves initial objects. Note this in general doesn't apply to any
other colimit.
-/
noncomputable def rightAdjointPreservesInitialOfCoreflective (R : D ⥤ C) [Coreflective R] :
```

dagurtomas (approved):

Mathlib/CategoryTheory/Monad/Limits.lean
mckoen: I think "nadir" is an option, but since we use "apex" and not "zenith" for cones, I prefer "(co)cone point." I'll change both.

Mathlib/CategoryTheory/Monad/Adjunction.lean
mckoen: This is just the dual statement, so I think it's appropriate to also refer to 5.3.3. I will add a comment.

jcommelin (commented):
Thanks :tada:

bors merge

PR #9317, opened 2023-12-28
[Merged by Bors] - feat: Turán's theorem

Parcly-Taxel:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14239, opened 2024-06-28
[Merged by Bors] - doc:add some comments and stacks tags

kbuzzard:
Add some stacks tags to etale, flat, unramified etc morphisms.

---

### DIFF
--- a/Mathlib/RingTheory/Etale/Basic.lean
+++ b/Mathlib/RingTheory/Etale/Basic.lean
@@ -138,6 +138,8 @@ where `Rₘ` and `Sₘ` are the localisations of `R` and `S` at a multiplicative
 subset `M` of `R`.
 -/
 
+def FiniteAdele := 37
+
 /-! Let R, S, Rₘ, Sₘ be commutative rings -/
 variable {R S Rₘ Sₘ : Type u} [CommRing R] [CommRing S] [CommRing Rₘ] [CommRing Sₘ]
 /-! Let M be a multiplicatively closed subset of `R` -/
@@ -161,7 +163,7 @@ theorem localization_base [FormallyEtale R Sₘ] : FormallyEtale Rₘ Sₘ :=
     ⟨FormallyUnramified.localization_base M, FormallySmooth.localization_base M⟩
 #align algebra.formally_etale.localization_base Algebra.FormallyEtale.localization_base
 
-/-- The localization of a formally etale map is formally etale. -/
+/-- The localization of a formally étale map is formally étale. -/
 theorem localization_map [FormallyEtale R S] : FormallyEtale Rₘ Sₘ := by
   haveI : FormallyEtale S Sₘ := FormallyEtale.of_isLocalization (M.map (algebraMap R S))
   haveI : FormallyEtale R Sₘ := FormallyEtale.comp R S Sₘ
--- a/Mathlib/RingTheory/Etale/Basic.lean
+++ b/Mathlib/RingTheory/Etale/Basic.lean
@@ -131,21 +131,21 @@ We now consider a commutative square of commutative rings
 R -----> S
 |        |
 |        |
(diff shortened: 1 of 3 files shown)


Mathlib/RingTheory/Etale/Basic.lean
```diff
@@ -122,10 +124,28 @@ end BaseChange
...
 section Localization
 
+/-!
+
+We now consider a commutative square of commutative rings
+
+R -----> S
+|        |
+|        |
+\/       \/
```
semorrison, line 134: Often just `v` suffices for arrow heads in ASCII. :-)

Mathlib/RingTheory/Etale/Basic.lean
```diff
@@ -122,10 +124,28 @@ end BaseChange
...
+
+where `Rₘ` and `Sₘ` are the localisations of `R` and `S` at a multiplicatively closed
+subset `M` of `R`.
+-/
+
+-- Let R, S, Rₘ, Sₘ be commutative rings
 variable {R S Rₘ Sₘ : Type u} [CommRing R] [CommRing S] [CommRing Rₘ] [CommRing Sₘ]
+-- Let M be a multiplicatively closed subset of `R`
 variable (M : Submonoid R)
+-- Assume that the rings are in a commutative as above
```
semorrison, line 145:
```suggestion
-- Assume that the rings are in a commutative diagram as above
```

Mathlib/RingTheory/Etale/Basic.lean
```diff
@@ -122,10 +124,28 @@ end BaseChange
...
+|        |
+|        |
+\/       \/
+Rₘ ----> Sₘ
+
+where `Rₘ` and `Sₘ` are the localisations of `R` and `S` at a multiplicatively closed
+subset `M` of `R`.
+-/
+
+-- Let R, S, Rₘ, Sₘ be commutative rings
```
semorrison, line 141: These could even be single `/-! ... -/` comments.

Mathlib/RingTheory/Etale/Basic.lean
```diff
@@ -141,6 +161,7 @@ theorem localization_base [FormallyEtale R Sₘ] : FormallyEtale Rₘ Sₘ :=
     ⟨FormallyUnramified.localization_base M, FormallySmooth.localization_base M⟩
 #align algebra.formally_etale.localization_base Algebra.FormallyEtale.localization_base
 
+/-- The localization of a formally etale map is formally etale. -/
```
semorrison, line 164:
```suggestion
/-- The localization of a formally étale map is formally etale. -/
```

semorrison:
bors d+

Mathlib/RingTheory/Etale/Basic.lean
kbuzzard: We had a discussion about this during the session, and I pointed out that searching for adèle in Lean was a pain for me because I didn't have an è on my keyboard. People on the call whose first language contained accents immediately told me how to do it on Ubuntu (Alt-Gr # e) and now look at me I'm writing accents in github comments, but do you think there's an argument @semorrison for de-accenting words like this? Alternatively, should the Lean declaration be `FiniteAdèleRing`?

Parcly-Taxel: I never put the accent on the "a" in `TuranGraph`/`IsTuranMaximal` in #9317, but (knowing that you can type any character in Ubuntu using Ctrl-Shift-U then the Unicode hex codepoint) put the accent everywhere _in the commentary_. The Lean declaration should be `FiniteAdeleRing`.

kbuzzard: OK Michael Stoll on the Zulip has convinced me you're right.

PR #14240, opened 2024-06-28
[Merged by Bors] - feat: make `discrete_of_t1_of_finite` an instance and rename

j-loreaux:
Also adds the lemma `Set.Finite.continuousOn` which seems to be missing.

---


j-loreaux:
we should benchmark this when CI is done to make sure any instance loops created don't cause problems.

grunweg:
As CI is done, let me benchmark this as mentioned above. (Let me know if this is unwanted and I'll stop!!)

grunweg:
!bench

dupuisf:
Thanks!

bors r+

PR #14243, opened 2024-06-28
[Merged by Bors] - feat: add polynomial Wronskian

seewoo5:
Define wronskian for polynomials over commutative rings, and prove some basic properties. This is mainly for the porting of our (me and @jcpaik) formalization of [Mason-Stothers theorem](https://github.com/seewoo5/lean-poly-abc). We already discussed about this in [Zulip](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Ring.20with.20derivation/near/447378869). Especially, it may possible to improve as
* Replace `CommRing` with `Ring`
* or more generally, Wronskian can be defined over any ring with `Derivation`, and most of the theorems we proved should also hold in this context

---

- [x] depends on: #14281


Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+  the Wronskian is nonzero. (Otherwise, `a = b = 1` gives a counterexample.)
+-/
+
+noncomputable section
+
+open scoped Polynomial Classical
+
+open Polynomial
+
+variable {R : Type _} [CommRing R]
```
eric-wieser, line 29:
```suggestion
variable {R : Type*} [CommRing R]
```

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+
+/-- Wronskian: W(a, b) = ab' - a'b. -/
+def wronskian (a b : R[X]) : R[X] :=
+  a * (derivative b) - (derivative a) * b
+
+@[simp]
+theorem wronskian_zero_left (a : R[X]) : wronskian 0 a = 0 := by
+  simp_rw [wronskian]; simp only [MulZeroClass.zero_mul, derivative_zero, sub_self]
```
eric-wieser, line 37:
```suggestion
  simp_rw [wronskian]; simp only [zero_mul, derivative_zero, sub_self]
```
and similarly below.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+
+theorem wronskian_eq_of_sum_zero {a b c : R[X]} (h : a + b + c = 0) :
+    wronskian a b = wronskian b c := by
+  rw [← neg_eq_iff_add_eq_zero] at h
+  rw [← h]
+  rw [wronskian_neg_right]
+  rw [wronskian_add_right]
+  rw [wronskian_self]
+  rw [add_zero]
+  rw [← wronskian_anticomm]
```
eric-wieser, line 65: Can you combine these into a single rewrite chain?

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+    _ < a.degree + b.degree := by
+      rw [max_lt_iff]
+      constructor
+      case left => .
+      {
+        apply lt_of_le_of_lt
+        exact degree_mul_le a (derivative b)
+        rw [WithBot.add_lt_add_iff_left (degree_ne_bot ha)]
+        exact Polynomial.degree_derivative_lt hb
+      }
```
eric-wieser, line 84:
```suggestion
      case left =>
        apply lt_of_le_of_lt
        exact degree_mul_le a (derivative b)
        rw [WithBot.add_lt_add_iff_left (degree_ne_bot ha)]
        exact Polynomial.degree_derivative_lt hb
```
and elsewhere

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+
+theorem wronskian.natDegree_lt_add {a b : R[X]} (hw : wronskian a b ≠ 0) :
+    (wronskian a b).natDegree < a.natDegree + b.natDegree := by
+  have ha : a ≠ 0 := by intro h; subst h; rw [wronskian_zero_left] at hw; exact hw rfl
+  have hb : b ≠ 0 := by intro h; subst h; rw [wronskian_zero_right] at hw; exact hw rfl
+  rw [← WithBot.coe_lt_coe, WithBot.coe_add]
+  convert ← wronskian.degree_lt_add ha hb
+  exact Polynomial.degree_eq_natDegree hw
+  exact Polynomial.degree_eq_natDegree ha
+  exact Polynomial.degree_eq_natDegree hb
```
eric-wieser, line 101: Please use `.` to create a subgoal for each of these `exact`s.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,110 @@
...
+  the sum of degrees of `a` and `b`
+- `wronskian.natDegree_lt_add`: `natDegree` version of the above theorem. We need to assume that
+  the Wronskian is nonzero. (Otherwise, `a = b = 1` gives a counterexample.)
+-/
+
+noncomputable section
+
+open scoped Polynomial Classical
+
+open Polynomial
```
eric-wieser, line 27:
```suggestion
namespace Polynomial
```
since you probably don't want to make `wronskian` in the global namespace.

eric-wieser (commented):
This generally looks good; I've left a first round of mostly stylistic comments.

Mathlib/RingTheory/Polynomial/Wronskian.lean
seewoo5: I combined all the `rw`s *except for the first one*, is it okay or could you help me to combine two further?
https://github.com/leanprover-community/mathlib4/blob/9861215de2277c9d3341cd47e5aa9c3ac0dbc880/Mathlib/RingTheory/Polynomial/Wronskian.lean#L59-L60

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,101 @@
...
+theorem wronskian_self (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem wronskian_anticomm (a b : R[X]) : wronskian a b = -wronskian b a := by
+  rw [wronskian, wronskian]; ring
+
+theorem wronskian_eq_of_sum_zero {a b c : R[X]} (h : a + b + c = 0) :
+    wronskian a b = wronskian b c := by
+  rw [← neg_eq_iff_add_eq_zero] at h
+  rw [← h, wronskian_neg_right, wronskian_add_right, wronskian_self, add_zero, ← wronskian_anticomm]
+
```
eric-wieser, line 61: As discussed on Zulip; can you prove this for `LinearMap.IsAlt`, and show `wronskianBilin` and `IsAlt wronskianBIlin`?

seewoo5: Added as #14281 ! I may update the proof with this after the PR is merged.

eric-wieser: Thanks!

seewoo5:
@eric-wieser Could you help me for setup? I'm trying to understand how the `BilinForm` works and how to adapt it to Wronskian. In fact, I cannot figure out how to "state" the theorem `wronskianBilin`. I found that the core definition of `BilinForm` is defined as `mk₂` in `LinearAlgebra/BilinearMap.lean`, but not sure where to start. For example, do I need to re-write the definition of current Wronskian? Once I figure out the statements, I think I can prove it myself.

eric-wieser:
With
```lean
import Mathlib.Algebra.Polynomial.AlgebraMap
import Mathlib.LinearAlgebra.SesquilinearForm
import Mathlib.Algebra.Algebra.Bilinear
```
you can write
```lean
/-- `Polynomial.wronskian` as a bilinear map. -/
def wronskianBilin : R[X] →ₗ[R] R[X] →ₗ[R] R[X] :=
  (LinearMap.mul R R[X]).compl₂ derivative - (LinearMap.mul R R[X]).comp derivative

@[simp]
theorem wronskianBilin_apply (a b : R[X]) : wronskianBilin a b = wronskian a b := rfl
```

seewoo5:
How can I state `IsAltwronskianBilin`? I tried the followings, which all didn't work (although I can prove it - whatever it is - with existing `wronskian_anticomm` that I may rename as `wronskian_is_alt`)
```lean
theorem IsAltwronskianBilin : wronskianBilin.IsAlt := by sorry
theorem IsAltwronskianBilin : LinearMap.IsAlt wronskianBilin := by sorry 
```

eric-wieser:
That doesn't work because Lean can't work out which `R` you are talking about. See my edit above for the fix, then you can write `(wronskianBilin R).IsAlt`.

seewoo5:
I think I'm almost done assuming `IsAltwronskianBilin` (I just `sorry`ed it and proved everything else). I have

```lean
theorem IsAltwronskianBilin : (wronskianBilin R).IsAlt := by
  intro a
  rw [wronskianBilin]
  -- sorry
```
and I though `ring` or `ring_nf` would just finish from here, but it seems not.

eric-wieser:
You should use the `wronskianBilin_apply` lemma I wrote above, rather than unfolding `wronskianBilin`

seewoo5:
That's actually the confusing point for me now. The current version proves everything with `wronskian`, and now we are trying to prove them using the lemma `eq_of_add_add_eq_zero` for alternating forms. So I tried to prove everything on the level of `wronskianBilin` instead of `wronskian`, and then use `rw` to rewrite every current lemma with `wronskianBilin`. For example, my unpushed proof includes something like
```lean
theorem wronskian_neg_right (a b : R[X]) : wronskian a (-b) = -wronskian a b := by
  rw [←wronskianBilin_apply, ←wronskianBilin_apply]; exact (wronskianBilin R a).map_neg b

theorem wronskian_eq_of_sum_zero {a b c : R[X]} (hAdd : a + b + c = 0) :
    wronskian a b = wronskian b c := by
  have h : wronskianBilin R a b = wronskianBilin R b c :=
    by exact LinearMap.IsAlt.eq_of_add_add_eq_zero IsAltwronskianBilin hAdd
  rw [←wronskianBilin_apply, ←wronskianBilin_apply]; exact h
```
So we utilizes the fact that `wronskianBilin` is indeed a bilinear map and reuse theorems there. But if we prove `IsAltwronskianBilin` with `wronskian` + ring tactics through `wronskianBilin_apply`, I felt like I'm doing in a reverse way.

eric-wieser:
A useful trick here is that you can simplify that to
```lean
theorem wronskian_neg_right (a b : R[X]) : wronskian a (-b) = -wronskian a b :=
  (wronskianBilin R a).map_neg b

theorem wronskian_eq_of_sum_zero {a b c : R[X]} (hAdd : a + b + c = 0) :
    wronskian a b = wronskian b c :=
  IsAltwronskianBilin.eq_of_add_add_eq_zero hAdd
```
I think it's ok to prove one result in the "reverse" way if it makes all the other results simpler.

seewoo5:
> A useful trick here is that you can simplify that to
> 
> ```lean
> theorem wronskian_neg_right (a b : R[X]) : wronskian a (-b) = -wronskian a b :=
>   (wronskianBilin R a).map_neg b
> 
> theorem wronskian_eq_of_sum_zero {a b c : R[X]} (hAdd : a + b + c = 0) :
>     wronskian a b = wronskian b c :=
>   IsAltwronskianBilin.eq_of_add_add_eq_zero hAdd
> ```

Could you explain how these simplifications works (without explicitly mentioning `rw`s)? This is somewhat unexpected for me.

eric-wieser:
The trick is that `wronskianBilin R a b = wronskian a b` is not only true, but true *by definition* - we say they are definitionally equal, or "defeq" for short. This means we can proof it by `rfl`, and that any proof of `P (wronskianBilin R a b)` can use used to populate a proof of `P (wronskian a b)`.

In some cases, this is called defeq abuse; it's not always reasonable for the caller to know when `foo` is defeq to `bar`. A clear exception to this rule is that the caller should always expect `fooBilin` to be defeq to `foo`, `fooLinearMap` to be defeq to `foo`, etc.
Usually we would write "`foo`, as a bilinear map" in the docstring of `fooBilin` to make that expectation clear.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,116 @@
...
+  (wronskianBilin R).map_add₂ a b c
+
+theorem wronskian_is_alt (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem IsAltwronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
+
+theorem wronskian_self_eq_zero (a : R[X]) : wronskian a a = 0 := wronskian_is_alt a
+
+theorem wronskianBilin_neq_eq (a b : R[X]) : -(wronskianBilin R) a b = (wronskianBilin R) b a :=
+  LinearMap.IsAlt.neg IsAltwronskianBilin a b
```
eric-wieser, line 70: I would recommend dropping this lemma, since it's not the simp-normal form; the one below is all you need, and you can replace `:= wronskianBilin_neq_eq a b` below with this proof.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,116 @@
...
+
+theorem wronskian_add_right (a b c : R[X]) : wronskian a (b + c) = wronskian a b + wronskian a c :=
+  (wronskianBilin R a).map_add b c
+
+theorem wronskian_add_left (a b c : R[X]) : wronskian (a + b) c = wronskian a c + wronskian b c :=
+  (wronskianBilin R).map_add₂ a b c
+
+theorem wronskian_is_alt (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem IsAltwronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
```
eric-wieser, line 65: The naming convention here is:
```suggestion
theorem isAlt_wronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
```
that is: separate declarations with underscores, and lowercase the first letter.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,115 @@
...
+
+theorem wronskian_add_right (a b c : R[X]) : wronskian a (b + c) = wronskian a b + wronskian a c :=
+  (wronskianBilin R a).map_add b c
+
+theorem wronskian_add_left (a b c : R[X]) : wronskian (a + b) c = wronskian a c + wronskian b c :=
+  (wronskianBilin R).map_add₂ a b c
+
+theorem wronskian_is_alt (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem IsAlt_wronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
```
eric-wieser, line 66:
```suggestion
theorem isAlt_wronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
```
The naming convention is to lower-case the first letter of each token in the theorem name, rather tahn using the name unchanged. I think all the other names in this PR are already correct.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,115 @@
...
+theorem wronskian_neg_eq (a b : R[X]) : -wronskian a b = wronskian b a :=
+  LinearMap.IsAlt.neg IsAlt_wronskianBilin a b
+
+theorem wronskian_eq_of_sum_zero {a b c : R[X]} (hAdd : a + b + c = 0) :
+    wronskian a b = wronskian b c := IsAlt_wronskianBilin.eq_of_add_add_eq_zero hAdd
+
+private theorem degree_ne_bot {a : R[X]} (ha : a ≠ 0) : a.degree ≠ ⊥ := by
+  intro h; rw [Polynomial.degree_eq_bot] at h; exact ha h
+
+theorem wronskian.degree_lt_add {a b : R[X]} (ha : a ≠ 0) (hb : b ≠ 0) :
```
eric-wieser, line 79:
```suggestion
theorem degree_wronskian_lt_add {a b : R[X]} (ha : a ≠ 0) (hb : b ≠ 0) :
```
dots in theorem names usually are reserved for when the left side is a type like `Polynomial`. One more below too.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,115 @@
...
+  (wronskianBilin R).map_add₂ a b c
+
+theorem wronskian_is_alt (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem IsAlt_wronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
+
+theorem wronskian_self_eq_zero (a : R[X]) : wronskian a a = 0 := wronskian_is_alt a
+
+theorem wronskian_neg_eq (a b : R[X]) : -wronskian a b = wronskian b a :=
+  LinearMap.IsAlt.neg IsAlt_wronskianBilin a b
```
eric-wieser, line 71:
```suggestion
  isAlt_wronskianBilin.IsAlt.neg a b
```
(untested, assumes the name change above)

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,115 @@
...
+This file defines Wronskian of a pair of polynomials, which is `W(a, b) = ab' - a'b`.
+We also prove basic properties of it.
+
+## Main declarations
+
+- `wronskian_eq_sum_zero`: We have `W(a, b) = W(b, c)` when `a + b + c = 0`.
+- `wronskian.degree_lt_add`: Degree of Wronskian `W(a, b)` is strictly smaller than
+  the sum of degrees of `a` and `b`
+- `wronskian.natDegree_lt_add`: `natDegree` version of the above theorem. We need to assume that
+  the Wronskian is nonzero. (Otherwise, `a = b = 1` gives a counterexample.)
```
eric-wieser, line 22: Don't forget to update these to reflect the name changes from review.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,115 @@
...
+  (wronskianBilin R a).map_add b c
+
+theorem wronskian_add_left (a b c : R[X]) : wronskian (a + b) c = wronskian a c + wronskian b c :=
+  (wronskianBilin R).map_add₂ a b c
+
+theorem wronskian_is_alt (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]
+
+theorem isAlt_wronskianBilin : (wronskianBilin R).IsAlt := by intro a; exact wronskian_is_alt a
+
+theorem wronskian_self_eq_zero (a : R[X]) : wronskian a a = 0 := wronskian_is_alt a
```
eric-wieser, line 68: You have the same lemma twice here!
```suggestion
theorem wronskian_self_eq_zero (a : R[X]) : wronskian a a = 0 := by rw [wronskian, mul_comm, sub_self]

theorem isAlt_wronskianBilin : (wronskianBilin R).IsAlt := wronskian_self_eq_zero
```

seewoo5:
@eric-wieser I believe it is now ready to be merged - let me know if it needs further updates.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,114 @@
...
+
+This file defines Wronskian of a pair of polynomials, which is `W(a, b) = ab' - a'b`.
+We also prove basic properties of it.
+
+## Main declarations
+
+- `wronskian_eq_of_sum_zero`: We have `W(a, b) = W(b, c)` when `a + b + c = 0`.
+- `degree_wronskian_lt_add`: Degree of Wronskian `W(a, b)` is strictly smaller than
+  the sum of degrees of `a` and `b`
+- `natDegree_wronskian_lt_add`: `natDegree` version of the above theorem. We need to assume that
```
eric-wieser, line 21: to make these clickable in the docs (may need line-rewrapping)
```suggestion

- `Polynomial.wronskian_eq_of_sum_zero`: We have `W(a, b) = W(b, c)` when `a + b + c = 0`.
- `Polynomial.degree_wronskian_lt_add`: Degree of Wronskian `W(a, b)` is strictly smaller than
  the sum of degrees of `a` and `b`
- `Polynomial.natDegree_wronskian_lt_add`: `natDegree` version of the above theorem. We need to assume that
```

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,114 @@
...
+  · exact Polynomial.degree_eq_natDegree hb
+
+-- Note: the following is false!
+-- Counterexample: b = a = 1 →
+-- (wronskian a b).natDegree = a.natDegree = b.natDegree = 0
+/-
+lemma wronskian.natDegree_lt_add {a b : R[X]}
+  (ha : a ≠ 0) (hb : b ≠ 0) :
+  (wronskian a b).natDegree < a.natDegree + b.natDegree := sorry
+-/
```
eric-wieser, line 114: I think this would be better placed in the docstring of `natDegree_wronskian_lt_add`, something like

> Note this would be false with just `(ha : a ≠ 0) (hb : b ≠ 0)`, as when `a = b = 1` we have `(wronskian a b).natDegree = a.natDegree = b.natDegree = 0`.

eric-wieser (approved):
maintainer merge

My one lingering question is "are you sure you don't want the n-ary wronskian instead of the binary one?". That could be written as an `AlternatingMap`.

If you don't want it, let's add that as a `TODO` in the module docstring.

I'll let someone else give this the final review, but it looks great to me now.

Mathlib/RingTheory/Polynomial/Wronskian.lean
```diff
@@ -0,0 +1,111 @@
...
+theorem isAlt_wronskianBilin : (wronskianBilin R).IsAlt := wronskian_self_eq_zero
+
+theorem wronskian_neg_eq (a b : R[X]) : -wronskian a b = wronskian b a :=
+  LinearMap.IsAlt.neg isAlt_wronskianBilin a b
+
+theorem wronskian_eq_of_sum_zero {a b c : R[X]} (hAdd : a + b + c = 0) :
+    wronskian a b = wronskian b c := isAlt_wronskianBilin.eq_of_add_add_eq_zero hAdd
+
+private theorem degree_ne_bot {a : R[X]} (ha : a ≠ 0) : a.degree ≠ ⊥ := by
+  intro h; rw [Polynomial.degree_eq_bot] at h; exact ha h
```
eric-wieser, line 76: Can you promote this to a non-private theorem next to `Polynomial.degree_eq_bot`? The proof should just be `Polynomial.degree_eq_bot.not`, and the statement should be an `Iff`.

seewoo5:
> maintainer merge
> 
> My one lingering question is "are you sure you don't want the n-ary wronskian instead of the binary one?". That could be written as an `AlternatingMap`.
> 
> If you don't want it, let's add that as a `TODO` in the module docstring.
> 
> I'll let someone else give this the final review, but it looks great to me now.

At least for me, I don't need it right now (for polynomial FLT). But I have an (unordinary) application in my mind that uses n-ary Wronskian. The polynomial FLT can be extended to arbitrary number of polynomials as follows: for any algebraically closed characteristic zero field $k$, the Fermat-Catalan equation

$$
a_{1}^{m_{1}} + a_{2}^{m_{2}} + \cdots + a_{n-1}^{m_{n-1}} = a_{n}^{m_{n}}
$$

does not admit a non-constant solution in $k[t]$ whenever $n(n-2) \leq m_{i}$ for all $1\leq i \leq n$ (see [this](https://staff.math.su.se/shapiro/ProblemSolving/BayatTeimoori.pdf) paper). The proof is almost same as the original Mason-Stothers theorem, and it certainly uses n-ary Wronskian. I haven't tried to formalize this proof yet, but if someone wants to formalize this (or possibly more general version that is much close to the Mason-Stothers theorem, not only for the Fermat-Catalan equations), they may need the n-ary version.

But Wronskian itself is more useful for testing linear independence of sufficiently nice functions, not only polynomials. In that direction, ultimately it would be better to define Wronskian for more general class of functions - or even for abstract rings with derivations - not only for polynomials. So I think it is fine to leave it as-is at this point.

riccardobrasca:
Can you please make the linter happy? Thanks!

seewoo5:
```
/- The `simpNF` linter reports:
SOME SIMP LEMMAS ARE NOT IN SIMP-NORMAL FORM.
see note [simp-normal form] for tips how to debug this.
https://leanprover-community.github.io/mathlib_docs/notes.html#simp-normal%20form -/
-- Mathlib.Algebra.Polynomial.Degree.Definitions
Error: ././././Mathlib/Algebra/Polynomial/Degree/Definitions.lean:1[23](https://github.com/leanprover-community/mathlib4/actions/runs/9868167114/job/27249725253?pr=14243#step:22:24):1: error: @Polynomial.degree_ne_bot simp can prove this:
  by simp only [@ne_eq, @Polynomial.degree_eq_bot]
One of the lemmas above could be a duplicate.
If that's not the case try reordering lemmas or adding @[priority].
```
If I understood correctly, this says that the newly added theorem `degree_ne_bot` is somewhat duplicates with `degree_eq_bot`. Should I remove this from the `Definitions.lean` and prove the original statement in Wronskian via `Polynomial.degree_eq_bot.ne`?

eric-wieser:
It's just telling you to remove `@[simp]`

riccardobrasca:
Thanks!

bors merge

PR #14250, opened 2024-06-29
feat : create FactorisationProperties.lean (new file)

Colin166:
---


Colin166:
Got rid of the build issue but now have a new "label new contributors" error

Colin166:
I need help. I don't know what's going on. I have the correct labels up and I've rerun all jobs.

loefflerd:
You've created your own personal fork of Mathlib and you're trying to PR from that. The CI scripts are set up on the assumption that you make a Git branch within the `leanprover_community` repo, rather than a separate fork.

loefflerd:
(Incidentally, it is very unfortunate that the error message is so confusing here – I think you would have every right to post on mathlib zulip pointing out the experience you've had and suggesting hacking the CI scripts to produce a more comprehensible error message.)

PR #14251, opened 2024-06-29
[Merged by Bors] - chore: add a `Repr` for `Sym2`

eric-wieser:
This matches the approach used by `Multiset`

---

### DIFF
--- a/Mathlib/Data/Sym/Sym2.lean
+++ b/Mathlib/Data/Sym/Sym2.lean
@@ -764,7 +764,7 @@ instance [IsEmpty α] : IsEmpty (Sym2 α) :=
 instance [Nontrivial α] : Nontrivial (Sym2 α) :=
   diag_injective.nontrivial
 
--- TODO: use a sort order if available, gh-18166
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Sym/Sym2.lean
```diff
@@ -764,4 +764,8 @@ instance [IsEmpty α] : IsEmpty (Sym2 α) :=
 instance [Nontrivial α] : Nontrivial (Sym2 α) :=
   diag_injective.nontrivial
 
+-- TODO: use a sort order if available, gh-18166
```
semorrison, line 767: Let's use full URLs when referencing issues or prs.

Actually, what is this, the number is too high to be an issue??

eric-wieser:
```suggestion
-- TODO: use a sort order if available, https://github.com/leanprover-community/mathlib/issues/18166
```

YaelDillies (approved):
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14254, opened 2024-06-29
[Merged by Bors] - feat(Algebra/Category/ModuleCat/{Presheaf, Sheaf}): forgetful functor from sheaves of modules over sheaf of ring R to sheaves of R(X)-module when X is initial

jjaassoonn:
Let $R : C^{op} \to Ring$ be a sheaf of rings and $M$ a sheaf of $R$-modules. Then $M$ is naturally a sheaf of $R(\top)$-modules where $\top$ is an initial object.

In the case of a sheafed space $X$, for any $x \in X$ and $M$ a sheaf of $\mathcal{O} _X$-modules, then $M$ is also a sheaf of $\Gamma(\mathcal{O} _X, X)$-modules; thus we can take stalks to naturally have that $M_x$ is a $\Gamma(\mathcal{O} _X, X)$-module. Thus we can use this construction to define $M_x$ as a module over $\mathcal{O} _{X, x}$.

Co-authored-by: Sophie Morel <sophie.morel@ens-lyon.fr>

---

- [x] depends on: #14283


Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,81 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+morphism level `(f : M ⟶ N) ↦ (c ↦ f(c))`.
+-/
+@[simps]
+noncomputable def forgetToPresheafModuleCatObj
+    (X : Cᵒᵖ) (hX : Limits.IsInitial X) (M : PresheafOfModules.{v} R) :
+    Cᵒᵖ ⥤ ModuleCat (R.1.obj X) :=
+{ obj := fun c =>
+      letI : Module (R.1.obj X) (M.presheaf.obj c) :=
+        Module.compHom (M.presheaf.obj c) (R.1.map (hX.to c))
+      ModuleCat.of _ (M.presheaf.obj c)
```
joelriou, line 524: Here, you should use `ModuleCat.restrictScalars` applied to the obvious map of rings and the object `M.obj c`.

Mathlib/Algebra/Category/ModuleCat/Sheaf.lean
```diff
@@ -92,6 +92,26 @@ def toSheaf : SheafOfModules.{v} R ⥤ Sheaf J AddCommGrp.{v} where
...
 
+/--
+The forgetful functor from sheaves of modules over sheaf of ring `R` to sheaves of `R(X)`-module
+when `X` is initial.
+-/
+@[simps]
+noncomputable def forgetToSheafModuleCat (X : Cᵒᵖ) (hX : Limits.IsInitial X)  :
+    SheafOfModules R ⥤ Sheaf J (ModuleCat (R.1.obj X)) where
+  obj M := ⟨(PresheafOfModules.forgetToPresheafModuleCat X hX).obj M.1, by
+    rw [CategoryTheory.Presheaf.isSheaf_iff_isSheaf_forget J
```
joelriou, line 103: Instead of `isSheaf_iff_isSheaf_forget`, I feel it should be possible to apply `isSheaf_iff_isSheaf_comp`(?) to the `forget₂` functor from modules to abelian groups. Then, the goal should be proved `by exact M.isSheaf`.

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
jjaassoonn: Thanks!

Mathlib/Algebra/Category/ModuleCat/Sheaf.lean
jjaassoonn: Thanks `isSheaf_iff_isSheaf_comp` indeed works, but I needed to introduce some extra universe constraint. Can you check if this is acceptable please?

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+
+The functor is implemented as, on object level `M ↦ (c ↦ M(c))` where the `R(X)`-module structure
+on `M(c)` is given by restriction of scalars along the unique morphism `R(c) ⟶ R(X)`; and on
+morphism level `(f : M ⟶ N) ↦ (c ↦ f(c))`.
+-/
+@[simps]
+noncomputable def forgetToPresheafModuleCatObj
+    (X : Cᵒᵖ) (hX : Limits.IsInitial X) (M : PresheafOfModules.{v} R) :
+    Cᵒᵖ ⥤ ModuleCat (R.1.obj X) :=
+{ obj := fun c =>
```
joelriou, line 521: This syntax is more convenient:
```suggestion
    Cᵒᵖ ⥤ ModuleCat (R.1.obj X) where
  obj c :=
```

Mathlib/Algebra/Category/ModuleCat/Sheaf.lean
joelriou: Could you make explicit the universe `w` in the statement: `SheafOfModules.{w} R ⥤ ...`?

I do not think that your previous version was not more general, because by writing `CategoryTheory.forget.{max v₁ u₁}` in the proof, it was forcing `w := max v₁ u₁`. For the applications to topological spaces (e.g. schemes), this is general enough. (Since I have shown that a limit in `AddCommGrp` exists iff a certain type is small, I think that `forget₂ (ModuleCat _) (AddCommGrp)` reflects all limits, regardless of universes for the index categories. Then, `forgetToSheafModuleCat` should be defined regardless of the universes, but I do not feel it is necessary to do it now!)

joelriou: In #14283, I show that the forget functor from `ModuleCat` to `AddCommGrp` reflects all limits (regardless of universes) [for the preservation we may need that the ring is `v`-small]. Then, I think that you may now use `isSheaf_of_isSheaf_comp` applied to this `forget₂` in order to remove universe restrictions.

erdOne:
(I have removed Kevin's comment in the PR message and fixed the LaTeX. You need spaces before underscores so that Github markdown does not register it as italics)

Mathlib/Algebra/Category/ModuleCat/Sheaf.lean
jjaassoonn: Thanks! This works perfectly!

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+
+/--
+Implementation of the functor `PresheafOfModules R ⥤ Cᵒᵖ ⥤ ModuleCat (R.obj X)`
+when `X` is initial.
+
+The functor is implemented as, on object level `M ↦ (c ↦ M(c))` where the `R(X)`-module structure
+on `M(c)` is given by restriction of scalars along the unique morphism `R(c) ⟶ R(X)`; and on
+morphism level `(f : M ⟶ N) ↦ (c ↦ f(c))`.
+-/
+noncomputable def forgetToPresheafModuleCatMap
```
jcommelin, line 549: Should there be `simps` on this?
```suggestion
@[simps]
noncomputable def forgetToPresheafModuleCatMap
```

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+
+/--
+The forgetful functor from presheaves of modules over a presheaf of rings `R` to presheaves of
+`R(X)`-modules where `X` is an initial object.
+
+The functor is implemented as, on object level `M ↦ (c ↦ M(c))` where the `R(X)`-module structure
+on `M(c)` is given by restriction of scalars along the unique morphism `R(c) ⟶ R(X)`; and on
+morphism level `(f : M ⟶ N) ↦ (c ↦ f(c))`.
+-/
+noncomputable def forgetToPresheafModuleCat (X : Cᵒᵖ) (hX : Limits.IsInitial X) :
```
jcommelin, line 568: And here?
```suggestion
@[simps]
noncomputable def forgetToPresheafModuleCat (X : Cᵒᵖ) (hX : Limits.IsInitial X) :
```

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+noncomputable def forgetToPresheafModuleCatObj
+    (X : Cᵒᵖ) (hX : Limits.IsInitial X) (M : PresheafOfModules.{v} R) :
+    Cᵒᵖ ⥤ ModuleCat (R.1.obj X) where
+  obj c :=
+    ModuleCat.restrictScalars (R.1.map (hX.to c)) |>.obj <| M.obj c
+  map := fun {c₁ c₂} f =>
+    { toFun := fun x => M.presheaf.map f x
+      map_add' := M.presheaf.map f |>.map_add
+      map_smul' := fun r (m : M.presheaf.obj c₁) => by
+        change M.presheaf.map f (R.1.map (hX.to c₁) _ • m) = R.1.map (hX.to c₂) _ • _
```
jcommelin, line 527: Are we missing (d)simp lemmas to get the goal in the right shape? I would prefer if we could avoid `change`.

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+        rw [M.map_smul, ← CategoryTheory.comp_apply, ← R.map_comp]
+        congr
+        apply hX.hom_ext }
+  map_id := fun c => by
+      ext x
+      change M.presheaf.map _ x = x
```
jcommelin, line 533: Idem

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,79 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+      rw [M.presheaf.map_id]
+      rfl
+  map_comp := fun {c₁ c₂ c₃} f g => by
+      ext x
+      change M.presheaf.map _ x = (M.presheaf.map _ ≫ M.presheaf.map _) x
```
jcommelin, line 538: Idem

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
jjaassoonn: This is resolved

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
jjaassoonn: This is resolved

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
jjaassoonn: This is resolved by explicitly tell Lean `m : ModuleCat.restrictScalars ....`

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
jjaassoonn: The generated `simp`-lemma gives a `simpNF`-linter error

Mathlib/Algebra/Category/ModuleCat/Presheaf.lean
```diff
@@ -497,4 +497,74 @@ def unitHomEquiv (M : PresheafOfModules R) :
...
+
+/--
+Implementation of the functor `PresheafOfModules R ⥤ Cᵒᵖ ⥤ ModuleCat (R.obj X)`
+when `X` is initial.
+
+The functor is implemented as, on object level `M ↦ (c ↦ M(c))` where the `R(X)`-module structure
+on `M(c)` is given by restriction of scalars along the unique morphism `R(c) ⟶ R(X)`; and on
+morphism level `(f : M ⟶ N) ↦ (c ↦ f(c))`.
+-/
+noncomputable def forgetToPresheafModuleCatMap
```
semorrison, line 543:
```suggestion
-- `@[simps]` here is not needed,
-- and in any case produces lemmas that the `simpNF` linter complains about.
noncomputable def forgetToPresheafModuleCatMap
```
Just so future readers don't miss it!

semorrison:
bors d+

PR #13943, opened 2024-06-19
[Merged by Bors] - chore(GroupTheory/MonoidLocalization): Change `Localization` to an abbrev of `OreLocalization`.

erdOne:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14257, opened 2024-06-29
Tinker with 13943 instance prios

kbuzzard:
---

This is #13943 with
```lean
instance (priority := 900) : DivisionRing R[R⁰⁻¹] where
```
in RingTheory.OreLocalization.Ring


kbuzzard:
!bench

kbuzzard:
OK so that might be a step in the right direction (e.g. we've gone from
```
+ ~Mathlib.AlgebraicGeometry.AffineScheme                  instructions   -34.5%
```
to
```
+ ~Mathlib.AlgebraicGeometry.AffineScheme                  instructions   -15.1%
```
but still more work to do

PR #14270, opened 2024-06-29
feat(FermatLastTheorem): Generalize FLT statement

jcpaik:
Generalize the statement of `FermatLastTheorem` to accommodate different kinds of rings (e.g. $R = k[X]$) and trivial solutions. A relevant Zulip discussion [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Mason-Stothers.20theorem.20and.20Polynomial.20FLT).

Fermat's Last Theorem states the following for ring $R = \mathbb{Z}$ (or semiring $R = \mathbb{N}$).

**Statement 1** ('zero' version): For any solutions to $a^n + b^n = c^n$ we have $abc=0$.

The Mason--Stothers theorem gives a polynomial variant of FLT for ring $R = k[X]$ with $k$ characteristic zero. But we have to change the notion of trivial solutions.

**Statement 2** ('unit' version):  For any solutions to $a^n + b^n = c^n$, the triple $(a, b, c)$ is a common multiple of $(a', b', c')$ where each $a', b', c'$ is either zero or unit.

We note the followings.

1. For $R =  \mathbb{N}, \mathbb{Z}$ the statements 1 and 2 are equivalent.
2. However, for $R = \mathbb{Q}$ statement 1 is equivalent to full FLT but statement 2 is trivial.
3. The equivalence of statements 1 and 2 for $R =  \mathbb{N}, \mathbb{Z}$ also requires a bit of juggling with definitions albeit trivial.
4. For $R =  k[X]$, statement 1 is likely false but statement 2 is true.
5. In "[Fermat's Last Theorem over Number Fields](https://revistas.rcaap.pt/boletimspm/article/view/21033/15548)" by N. Freitas, the author mostly mentions statement 1 as a generalization, but also mentions many 'trivial' solutions outside statement 1.

The relationship between statements 1 and 2 in general:

1. If $R$ is a Dedekind domain (e.g. integral ring) then it seems that zero version implies the unit version.
2. If the sum of two units in $R$ is never a unit in $R$, then the unit version implies the zero version. [Here](https://math.stackexchange.com/questions/2107233/when-are-u-and-1-u-never-both-units) it is mentioned that the integral ring of $\mathbb{Q}(\sqrt{d})$ for $d < 0$ satisfy the condition unless $d = 2, 3, 6$.

This PR attempts to capture such minor differences in the notion of 'trivial' solutions.

1. We provide a general statement for stating FLT with arbitrary notion of 'trivial' solutions. Then we give Statements 1 and 2 as specific instances.
3. Since we are much more familiar with 'zero' version (Statement 1), we define the main FLT for integers with Statement 1.
4. We show the equivalence between Statements 1 and 2 for $R =  \mathbb{N}, \mathbb{Z}$
5. We state the polynomial FLT with statement 2. This will be proved when we port the project I did with @seewoo5 [here](https://github.com/seewoo5/lean-poly-abc).

---

### DIFF
--- a/Mathlib/NumberTheory/FLT/Basic.lean
+++ b/Mathlib/NumberTheory/FLT/Basic.lean
@@ -1,12 +1,14 @@
 /-
 Copyright (c) 2023 Kevin Buzzard. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Kevin Buzzard, Yaël Dillies
+Authors: Kevin Buzzard, Yaël Dillies, Jineon Baek
 -/
+import Mathlib.Algebra.CharP.Defs
 import Mathlib.Algebra.GCDMonoid.Finset
 import Mathlib.Algebra.GCDMonoid.Nat
 import Mathlib.Algebra.GroupWithZero.Divisibility
 import Mathlib.Algebra.Order.Ring.Abs
+import Mathlib.Algebra.Polynomial.Basic
 import Mathlib.Data.Rat.Defs
 import Mathlib.RingTheory.PrincipalIdealDomain
 import Mathlib.Tactic.NormNum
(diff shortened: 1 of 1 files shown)


jcpaik:
The current commits are full of sorries. I will proceed once we reach a general consensus on the right generalized statement.

kbuzzard:
I really don't mind what we do here. My original motivation with the definition was that I felt that mathlib should have some famous mathematical statements, such as FLT and the Riemann Hypothesis etc, even if we don't have proofs. If you want to manipulate things so that your proof of FLT for function fields fits into this framework this is fine by me, it doesn't affect my work.

PR #14271, opened 2024-06-29
[Merged by Bors] - feat(RingTheory): presentations of algebras

chrisflav:
Adds basic API for presentations of algebras and constructors for localization and base change.

Co-authored-by: Andrew Yang @erdOne
Co-authored-by: Jung Tao Cheng @rontaucheng 

---

### DIFF
--- a/Mathlib/RingTheory/Generators.lean
+++ b/Mathlib/RingTheory/Generators.lean
@@ -179,6 +179,8 @@ def extendScalars (P : Generators R T) : Generators S T where
   σ' x := map (algebraMap R S) (P.σ x)
   aeval_val_σ' s := by simp [@aeval_def S, ← IsScalarTower.algebraMap_eq, ← @aeval_def R]
 
+/-- If `P` is a family of generators of `S` over `R` and `T` is an `R`-algebra, we
+obtain a natural family of generators of `T ⊗[R] S` over `T`. -/
 noncomputable
 def baseChange {T} [CommRing T] [Algebra R T] (P : Generators R S) : Generators T (T ⊗[R] S) := by
   apply Generators.ofSurjective (fun x ↦ 1 ⊗ₜ[R] P.val x)
--- a/Mathlib/RingTheory/Generators.lean
+++ b/Mathlib/RingTheory/Generators.lean
@@ -6,7 +6,7 @@ Authors: Andrew Yang
 import Mathlib.RingTheory.Ideal.Cotangent
 import Mathlib.RingTheory.Localization.Away.Basic
 import Mathlib.RingTheory.MvPolynomial.Tower
-import Mathlib.RingTheory.TensorProduct.MvPolynomial
+import Mathlib.RingTheory.TensorProduct.Basic
 
 /-!
 
--- a/Mathlib/RingTheory/Generators.lean
+++ b/Mathlib/RingTheory/Generators.lean
@@ -93,7 +93,7 @@ lemma algebraMap_surjective : Function.Surjective (algebraMap P.Ring S) := (⟨_
 section Construction
 
 /-- Construct `Generators` from an assignment `I → S` such that `R[X] → S` is surjective. -/
-@[simps vars val]
+@[simps val, simps (config := .lemmasOnly) vars]
 noncomputable
 def ofSurjective {vars} (val : vars → S) (h : Function.Surjective (aeval (R := R) val)) :
     Generators R S where
@@ -131,7 +131,7 @@ variable (r : R) [IsLocalization.Away r S]
 
 /-- If `S` is the localization of `R` away from `r`, we obtain a canonical generator mapping
 to the inverse of `r`. -/
-@[simps vars val, simps (config := .lemmasOnly) σ]
+@[simps val, simps (config := .lemmasOnly) vars σ]
 noncomputable
 def localizationAway : Generators R S where
   vars := Unit
@@ -154,7 +154,7 @@ variable {T} [CommRing T] [Algebra R T] [Algebra S T] [IsScalarTower R S T]
 
 /-- Given two families of generators `S[X] → T` and `R[Y] → S`,
 we may constuct the family of generators `R[X, Y] → T`. -/
-@[simps vars val, simps (config := .lemmasOnly) σ]
+@[simps val, simps (config := .lemmasOnly) vars σ]
 noncomputable
 def comp (Q : Generators S T) (P : Generators R S) : Generators R T where
   vars := Q.vars ⊕ P.vars
@@ -171,7 +171,7 @@ def comp (Q : Generators S T) (P : Generators R S) : Generators R T where
 variable (S) in
 /-- If `R → S → T` is a tower of algebras, a family of generators `R[X] → T`
 gives a family of generators `S[X] → T`. -/
-@[simps vars val]
+@[simps val, simps (config := .lemmasOnly) vars]
 noncomputable
 def extendScalars (P : Generators R T) : Generators S T where
   vars := P.vars
@@ -181,6 +181,7 @@ def extendScalars (P : Generators R T) : Generators S T where
 
 /-- If `P` is a family of generators of `S` over `R` and `T` is an `R`-algebra, we
 obtain a natural family of generators of `T ⊗[R] S` over `T`. -/
+@[simps! val, simps! (config := .lemmasOnly) vars]
 noncomputable
 def baseChange {T} [CommRing T] [Algebra R T] (P : Generators R S) : Generators T (T ⊗[R] S) := by
   apply Generators.ofSurjective (fun x ↦ 1 ⊗ₜ[R] P.val x)
@@ -336,6 +337,9 @@ section Cotangent
 
 /-- The kernel of a presentation. -/
 abbrev ker : Ideal P.Ring := RingHom.ker (algebraMap P.Ring S)
+
+lemma ker_eq_ker_aeval_val : P.ker = RingHom.ker (aeval P.val) :=
+  rfl
 
 /-- The cotangent space of a presentation.
 This is a type synonym so that `P = R[X]` can act on it through the action of `S` without creating
--- a/Mathlib/RingTheory/Generators.lean
+++ b/Mathlib/RingTheory/Generators.lean
@@ -4,6 +4,9 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Andrew Yang
 -/
 import Mathlib.RingTheory.Ideal.Cotangent
+import Mathlib.RingTheory.Localization.Away.Basic
+import Mathlib.RingTheory.MvPolynomial.Tower
+import Mathlib.RingTheory.TensorProduct.MvPolynomial
 
 /-!
 
@@ -24,10 +27,6 @@ import Mathlib.RingTheory.Ideal.Cotangent
   A hom between `P` and `P'` is an assignment `X → P'` such that the arrows commute.
 - `Algebra.Generators.Cotangent`: The cotangent space wrt `P = R[X] → S`, i.e. the
   space `I/I²` with `I` being the kernel of the presentation.
-
-## TODO
-* Define `Algebra.Presentation` that extends `Generators` and contains the data of a family of
-  relations that spans the kernel of the presentation.
 
 -/
 
@@ -126,6 +125,31 @@ def self : Generators R S where
   σ' := X
   aeval_val_σ' := aeval_X _
 
+section Localization
+
+variable (r : R) [IsLocalization.Away r S]
+
+/-- If `S` is the localization of `R` away from `r`, we obtain a canonical generator mapping
+to the inverse of `r`. -/
+@[simps vars val, simps (config := .lemmasOnly) σ]
+noncomputable
+def localizationAway : Generators R S where
+  vars := Unit
+  val _ := IsLocalization.Away.invSelf r
+  σ' s :=
+    letI a : R := (IsLocalization.Away.sec r s).1
+    letI n : ℕ := (IsLocalization.Away.sec r s).2
+    C a * X () ^ n
+  aeval_val_σ' s := by
+    rw [_root_.map_mul, algHom_C, map_pow, aeval_X]
+    simp only [← IsLocalization.Away.sec_spec, map_pow, IsLocalization.Away.invSelf]
+    rw [← IsLocalization.mk'_pow, one_pow, ← IsLocalization.mk'_one (M := Submonoid.powers r) S r]
+    rw [← IsLocalization.mk'_pow, one_pow, mul_assoc, ← IsLocalization.mk'_mul]
+    rw [mul_one, one_mul, IsLocalization.mk'_pow]
+    simp
+
+end Localization
+
 variable {T} [CommRing T] [Algebra R T] [Algebra S T] [IsScalarTower R S T]
 
 /-- Given two families of generators `S[X] → T` and `R[Y] → S`,
@@ -154,6 +178,32 @@ def extendScalars (P : Generators R T) : Generators S T where
   val := P.val
   σ' x := map (algebraMap R S) (P.σ x)
   aeval_val_σ' s := by simp [@aeval_def S, ← IsScalarTower.algebraMap_eq, ← @aeval_def R]
+
+noncomputable
+def baseChange {T} [CommRing T] [Algebra R T] (P : Generators R S) : Generators T (T ⊗[R] S) := by
+  apply Generators.ofSurjective (fun x ↦ 1 ⊗ₜ[R] P.val x)
+  intro x
+  induction x using TensorProduct.induction_on with
+  | zero => exact ⟨0, map_zero _⟩
+  | tmul a b =>
+    let X := P.σ b
+    use a • MvPolynomial.map (algebraMap R T) X
+    simp only [LinearMapClass.map_smul, X, aeval_map_algebraMap]
+    have : ∀ y : P.Ring,
+      aeval (fun x ↦ (1 ⊗ₜ[R] P.val x : T ⊗[R] S)) y = 1 ⊗ₜ aeval (fun x ↦ P.val x) y := by
+      intro y
+      induction y using MvPolynomial.induction_on with
+      | h_C a =>
+        rw [aeval_C, aeval_C, TensorProduct.algebraMap_apply, algebraMap_eq_smul_one, smul_tmul,
+          algebraMap_eq_smul_one]
+      | h_add p q hp hq => simp [map_add, tmul_add, hp, hq]
+      | h_X p i hp => simp [hp]
+    rw [this, P.aeval_val_σ, smul_tmul', smul_eq_mul, mul_one]
+  | add x y ex ey =>
+    obtain ⟨a, ha⟩ := ex
+    obtain ⟨b, hb⟩ := ey
+    use (a + b)
+    rw [map_add, ha, hb]
 
 end Construction
 
--- a/Mathlib/RingTheory/Localization/Away/Basic.lean
+++ b/Mathlib/RingTheory/Localization/Away/Basic.lean
@@ -61,8 +61,8 @@ theorem mul_invSelf : algebraMap R S x * invSelf x = 1 := by
   apply IsLocalization.mk'_one
 #align is_localization.away.mul_inv_self IsLocalization.Away.mul_invSelf
 
-/-- For `s : S`, this is a choice of `(x, n) : R × ℕ` such that
-`s * algebraMap R S (s ^ n) = algebraMap R S r`. -/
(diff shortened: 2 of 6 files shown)


Mathlib/RingTheory/Localization/Away/Basic.lean
```diff
@@ -61,6 +61,18 @@ theorem mul_invSelf : algebraMap R S x * invSelf x = 1 := by
   apply IsLocalization.mk'_one
 #align is_localization.away.mul_inv_self IsLocalization.Away.mul_invSelf
 
+/-- For `s : S`, this is a choice of `(x, n) : R × ℕ` such that
+`s * algebraMap R S (s ^ n) = algebraMap R S r`. -/
```
erdOne, line 65:
```suggestion
/-- For `s : S` with `S` being the localization away from `f`,
this is a choice of `(r, n) : R × ℕ` such that `s * algebraMap R S (f ^ n) = algebraMap R S r`. -/
```

MichaelStollBayreuth, line 65: Where is `x` in the statement? What is `r`?
Also, `x` is a `variable` here, so the name `x` should not be used to describe something else.
Did you mean to say
"For `s : S`, this is a choice of `(r, n) : R × ℕ` such that `s * algebraMap R S (x ^ n) = algebraMap R S r`?
(I.e., expressing `s` as `r/x^n` informally.)

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+1. `relations`: The type of relations.
+2. `relation : relations → MvPolynomial vars R`: The assignment of
+each relation to a polynomial in the generators.
+-/
+@[nolint checkUnivs]
+structure Algebra.Presentation extends Algebra.Generators.{w} R S where
+  /-- The type of relations.  -/
+  relations : Type t
+  /-- The assignment of each relation to a polynomial in the generators. -/
+  relation : relations → MvPolynomial vars R
```
erdOne, line 57: I think the names of these two are too similar and might be confusing. Maybe make `relations` -> `rels` to mimic `vars`.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+  /-- The relations span the kernel of the canonical map. -/
+  ker_algebraMap_eq_span_range_relation :
+    RingHom.ker (aeval val) = Ideal.span (Set.range <| relation)
+
+namespace Algebra.Presentation
+
+variable {R S}
+variable (P : Presentation.{t, w} R S)
+
+lemma ideal_eq_span_range_relation : P.ker = Ideal.span (Set.range <| P.relation) :=
```
erdOne, line 67: I'd reverse it and call it `span_range_relation_eq_ker`.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+  P.ker_algebraMap_eq_span_range_relation
+
+lemma algebraMap_relation (i) : aeval P.val (P.relation i) = 0 := by
```
erdOne, line 70:
```suggestion
@[simp] lemma algebraMap_relation (i) : aeval P.val (P.relation i) = 0 := by
```

Mathlib/RingTheory/MvPolynomial/Localization.lean
```diff
@@ -69,3 +79,78 @@ lemma isLocalization_C_mk' (a : R) (m : M) :
...
+    · rintro p ⟨q, rfl⟩
+      simp
+    · simp
+    · intro p q hp hq
+      simp [hp, hq]
+    · intro a x hx
+      simp [hx]
+
+@[simp]
+lemma auxHom_mk (p : MvPolynomial Unit R) :
```
MichaelStollBayreuth, line 105: Should this also be `private`? It uses a private definition in the statement, which will then not be understandable outside this file.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+## Main definition
+
+- `Algebra.Presentation`: A presentation of a `R`-algebra `S` is a family of
+  generators with
+  1. `relations`: The type of relations.
+  2. `relation : relations → MvPolynomial vars R`: The assignment of
+     each relation to a polynomial in the generators.
+- `Algebra.Presentation.IsFinite`: A presentation is called finite, if both variables and relations
+  are finite.
+- `Algebra.dimension`: The dimension of a presentation is the number of generators minus the number
```
MichaelStollBayreuth, line 27: Surely, this should be `Algebra.Presentation.dimension` ?

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+-/
+@[nolint checkUnivs]
+structure Algebra.Presentation extends Algebra.Generators.{w} R S where
+  /-- The type of relations.  -/
+  relations : Type t
+  /-- The assignment of each relation to a polynomial in the generators. -/
+  relation : relations → MvPolynomial vars R
+  /-- The relations span the kernel of the canonical map. -/
+  ker_algebraMap_eq_span_range_relation :
+    RingHom.ker (aeval val) = Ideal.span (Set.range <| relation)
```
MichaelStollBayreuth, line 60:
```suggestion
    RingHom.ker (aeval val) = Ideal.span (Set.range relation)
```
Left-over from a previous version?

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+lemma quotientEquiv_symm (x : S) : P.quotientEquiv.symm x = P.σ x :=
+  rfl
+
+/--
+Dimension of a presentation defined as the cardinality of the generators
+minus the cardinality of the relations.
+
+Note: this definition is completely non-sensical for non-finite presentations and
+even then for this to make sense, you should assume that the presentation
+is standard smooth.
```
erdOne, line 95:
```suggestion
Note: this definition is completely non-sensical for non-finite presentations and
even then for this to make sense, you should assume that the presentation
is a complete intersection.
```

Mathlib/RingTheory/Presentation.lean
MichaelStollBayreuth, line 67:
```suggestion
lemma ideal_eq_span_range_relation : P.ker = Ideal.span (Set.range P.relation) :=
```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+-/
+noncomputable def dimension : ℕ :=
+  Nat.card P.vars - Nat.card P.relations
+
+/-- A presentation is called finite if there are only finitely-many
```
erdOne, line 100:
```suggestion
/-- A presentation is finite if there are only finitely-many
```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+variable (P : Presentation.{t, w} R S)
+
+lemma ideal_eq_span_range_relation : P.ker = Ideal.span (Set.range <| P.relation) :=
+  P.ker_algebraMap_eq_span_range_relation
+
+lemma algebraMap_relation (i) : aeval P.val (P.relation i) = 0 := by
+  rw [← RingHom.mem_ker, ker_algebraMap_eq_span_range_relation]
+  exact Ideal.subset_span ⟨i, rfl⟩
+
+/-- The polynomial algebra wrt a family of generators module a family of relations. -/
```
MichaelStollBayreuth, line 74:
```suggestion
/-- The polynomial algebra wrt a family of generators modulo a family of relations. -/
```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+lemma ideal_fg_of_isFinite [P.IsFinite] : P.ker.FG := by
+  use (Set.finite_range P.relation).toFinset
+  simp [ideal_eq_span_range_relation]
+
+/-- If a presentation is finite, the corresponding quotient is
+of finite presentation. -/
+instance [P.IsFinite] : FinitePresentation R P.Quotient :=
+  FinitePresentation.quotient P.ideal_fg_of_isFinite
+
+lemma finitePresentation_of_presentation_of_isFinite [P.IsFinite] :
```
erdOne, line 117:
```suggestion
lemma finitePresentation_of_isFinite [P.IsFinite] :
```
This is already in the `Presentation` namespace right?

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+      induction x using MvPolynomial.induction_on with
+      | h_C a =>
+        simp only [Generators.algebraMap_apply, algHom_C, TensorProduct.algebraMap_apply,
+          id.map_eq_id, RingHom.id_apply, e]
+        erw [← MvPolynomial.algebraMap_eq, AlgEquiv.commutes]
+        simp
+      | h_add p q hp hq => simp only [map_add, hp, hq]
+      | h_X p i hp =>
+        simp [_root_.map_mul, e, hp]
+        rfl
```
erdOne, line 193: Not sure how robust is `simp; rfl`. Maybe squeeze it.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+    erw [H, ker_algebraMap_eq_span_range_relation] at H'
+    erw [← Ideal.mem_comap, Ideal.comap_symm, Ideal.map_map, Ideal.map_span, ← Set.range_comp] at H'
+    convert H'
+    simp
+    rfl
```
erdOne, line 198: Similarly.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+lemma localizationAway_dimension_zero : (localizationAway r (S := S)).dimension = 0 := by
+  simp [Presentation.dimension, localizationAway]
+
+end Localization
+
+variable {T} [CommRing T] [Algebra R T] (P : Presentation R S)
+
+private lemma ker_algebraMap_eq_span_range_relation_baseChange :
+    RingHom.ker (aeval (R := T) (S₁ := T ⊗[R] S) P.baseChange.val) =
+      Ideal.span (Set.range fun i ↦ (MvPolynomial.map (algebraMap R T)) (P.relation i)) := by
```
MichaelStollBayreuth, line 173: One could write the `fun i ↦ ...` part more concisely as `(MvPolynomial.map (algebraMap R T)) ∘ P.relation`, but I haven't check if this makes things better or worse later.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,220 @@
...
+        simp only [Generators.algebraMap_apply, algHom_C, TensorProduct.algebraMap_apply,
+          id.map_eq_id, RingHom.id_apply, e]
+        erw [← MvPolynomial.algebraMap_eq, AlgEquiv.commutes]
+        simp
+      | h_add p q hp hq => simp only [map_add, hp, hq]
+      | h_X p i hp =>
+        simp [_root_.map_mul, e, hp]
+        rfl
+    erw [H, ker_algebraMap_eq_span_range_relation] at H'
+    erw [← Ideal.mem_comap, Ideal.comap_symm, Ideal.map_map, Ideal.map_span, ← Set.range_comp] at H'
```
MichaelStollBayreuth, line 195: Can these be combined into a single `erw`?

Mathlib/RingTheory/Presentation.lean
chrisflav: It seems to make things worse, so I left it as is.

chrisflav:
Thanks for the quick reviews!

Mathlib/RingTheory/Presentation.lean
MichaelStollBayreuth: An alternative (which is just alternative notation, AFAIK) is
`((MvPolynomial.map (algebraMap R T)) (P.relation ·))` .

MichaelStollBayreuth: (But this is not really important...)

MichaelStollBayreuth:
A question for native English speakers: Which is correct?
* "A `R`-algebra" 
* "An `R`-algebra"

My gut feeling says it's the second one (since it corresponds to what I would say when reading it), but I wouldn't want to claim I know all the rules.

Mathlib/RingTheory/Presentation.lean
chrisflav: The scope of `·` seems to be interpreted as restricted to `P.relation`, so this causes a type error.

kbuzzard:
> A question for native English speakers: Which is correct?
> 
> * "A `R`-algebra"
> * "An `R`-algebra"
> 
> My gut feeling says it's the second one (since it corresponds to what I would say when reading it), but I wouldn't want to claim I know all the rules.

Definitely "an `R`-algebra". Indeed the rule is "write what you would say".

chrisflav:
> > A question for native English speakers: Which is correct?
> > 
> > * "A `R`-algebra"
> > * "An `R`-algebra"
> > 
> > My gut feeling says it's the second one (since it corresponds to what I would say when reading it), but I wouldn't want to claim I know all the rules.
> 
> Definitely "an `R`-algebra". Indeed the rule is "write what you would say", and what you say depends on the sound not the actual letter (so e.g. "a eucalyptus tree" as the word is pronounced "Yucaliptus")

Fixed.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,224 @@
...
+1. `rels`: The type of relations.
+2. `relation : relations → MvPolynomial vars R`: The assignment of
+each relation to a polynomial in the generators.
+-/
+@[nolint checkUnivs]
+structure Algebra.Presentation extends Algebra.Generators.{w} R S where
+  /-- The type of relations.  -/
+  rels : Type t
+  /-- The assignment of each relation to a polynomial in the generators. -/
+  relation : rels → MvPolynomial vars R
```
erdOne, line 57:
```suggestion
  relation : rels → MvPolynomial vars R
```

erdOne, line 57:
```suggestion
  relation : rels → toGenerators.Ring
```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,224 @@
...
+  /-- The relations span the kernel of the canonical map. -/
+  ker_algebraMap_eq_span_range_relation :
+    RingHom.ker (aeval val) = Ideal.span (Set.range relation)
```
erdOne, line 60:
```suggestion
    toGenerators.ker = Ideal.span (Set.range relation)
```
I wonder if we should just call it `span_range_relation_eq_ker` and replace the lemma below.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,224 @@
...
+namespace Algebra.Presentation
+
+variable {R S}
+variable (P : Presentation.{t, w} R S)
+
+lemma span_range_relation_eq_ker : Ideal.span (Set.range P.relation) = P.ker :=
+  P.ker_algebraMap_eq_span_range_relation.symm
+
+@[simp]
+lemma algebraMap_relation (i) : aeval P.val (P.relation i) = 0 := by
```
erdOne, line 71:
```suggestion
lemma aeval_val_relation (i) : aeval P.val (P.relation i) = 0 := by
```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,224 @@
...
+  erw [← RingHom.comap_ker]
+  simp only [Generators.localizationAway_vars, AlgEquiv.toAlgHom_eq_coe, AlgHom.toRingHom_eq_coe,
+    AlgEquiv.toAlgHom_toRingHom]
+  show Ideal.comap _ (RingHom.ker (mvPolynomialQuotientEquiv S r)) =
+    Ideal.span {C r * X () - 1}
+  simp [RingHom.ker_equiv, ← RingHom.ker_eq_comap_bot]
+
+/-- If `S` is the localization of `R` away from `r`, we can construct a natural
+presentation of `S` as `R`-algebra with a single generator `X` and the relation `r * X - 1 = 0`. -/
+@[simps rels relation]
```
erdOne, line 151: Working with the `Generators` api a bit more, I've come to realize having simp lemmas on the indexing types are not a good thing. Maybe make `rels` `lemmasOnly`.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,224 @@
...
+      TensorProduct.includeRight_apply]
+    erw [aeval_map_algebraMap]
+    show _ = TensorProduct.includeRight _
+    erw [map_aeval, TensorProduct.includeRight.comp_algebraMap]
+    rfl
+
+/-- If `P` is a presentation of `S` over `R` and `T` is an `R`-algebra, we
+obtain a natural presentation of `T ⊗[R] S` over `T`. -/
+noncomputable
+def baseChange : Presentation T (T ⊗[R] S) where
```
erdOne, line 218:
```suggestion
@[simps relation]
def baseChange : Presentation T (T ⊗[R] S) where
```

Mathlib/RingTheory/Generators.lean
```diff
@@ -126,6 +125,31 @@ def self : Generators R S where
   σ' := X
   aeval_val_σ' := aeval_X _
 
+section Localization
+
+variable (r : R) [IsLocalization.Away r S]
+
+/-- If `S` is the localization of `R` away from `r`, we obtain a canonical generator mapping
+to the inverse of `r`. -/
+@[simps vars val, simps (config := .lemmasOnly) σ]
```
erdOne, line 134: Similar here. `vars` should not be simp IMO.

Mathlib/RingTheory/Generators.lean
```diff
@@ -155,6 +179,34 @@ def extendScalars (P : Generators R T) : Generators S T where
   σ' x := map (algebraMap R S) (P.σ x)
   aeval_val_σ' s := by simp [@aeval_def S, ← IsScalarTower.algebraMap_eq, ← @aeval_def R]
 
+/-- If `P` is a family of generators of `S` over `R` and `T` is an `R`-algebra, we
+obtain a natural family of generators of `T ⊗[R] S` over `T`. -/
+noncomputable
+def baseChange {T} [CommRing T] [Algebra R T] (P : Generators R S) : Generators T (T ⊗[R] S) := by
```
erdOne, line 185:
```suggestion
@[simps! val]
def baseChange {T} [CommRing T] [Algebra R T] (P : Generators R S) : Generators T (T ⊗[R] S) := by
```

Mathlib/RingTheory/Presentation.lean
chrisflav: You mean, state this the other way round?

erdOne: Yes.

chrisflav: Done.

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,231 @@
...
+- `Algebra.Presentation.dimension`: The dimension of a presentation is the number of generators
+  minus the number of relations.
+
+We also give constructors for localization and base change.
+
+## TODO
+
+- Define composition of presentations.
+- Define `Hom`s of presentations.
+
```
erdOne, line 36:
```suggestion
## Notes
This contribution was created as part of the AIM workshop
"Formalizing algebraic geometry" in June 2024.

```

Mathlib/RingTheory/Presentation.lean
```diff
@@ -0,0 +1,231 @@
...
+  /-- The relations span the kernel of the canonical map. -/
+  span_range_relation_eq_ker :
+    Ideal.span (Set.range relation) = toGenerators.ker
+
+namespace Algebra.Presentation
+
+variable {R S}
+variable (P : Presentation.{t, w} R S)
+
+lemma ker_eq_span_range_relation : P.ker = Ideal.span (Set.range P.relation) :=
```
erdOne, line 70: I think we can remove this.

erdOne:
LGTM. What does @MichaelStollBayreuth think?

MichaelStollBayreuth (approved):
LGTM!

MichaelStollBayreuth:
maintainer merge

riccardobrasca:
Thanks!

bors merge

chrisflav:
Thanks for all the reviews!

PR #14272, opened 2024-06-30
[Merged by Bors] - fix(scripts/{lint_style,mk_all}): limit return value to 125

grunweg:
so the executable's exit code can be used further in shell scripts, such as mathlib's CI.
See https://unix.stackexchange.com/a/418802 for background why a maximum of 125 is safest (on Linux).

---

### DIFF
--- a/scripts/lint_style.lean
+++ b/scripts/lint_style.lean
@@ -26,7 +26,9 @@ def lintStyleCli (args : Cli.Parsed) : IO UInt32 := do
   for s in ["Archive.lean", "Counterexamples.lean", "Mathlib.lean"] do
     let n ← lintAllFiles (System.mkFilePath [s]) errorStyle
     numberErrorFiles := numberErrorFiles + n
-  return numberErrorFiles
+  -- Make sure to return an exit code of at most 125, so this return value can be used further
+  -- in shell scripts.
+  return min numberErrorFiles 125
 
 /-- Setting up command line options and help text for `lake exe lint_style`. -/
 -- so far, no help options or so: perhaps that is fine?
--- a/scripts/lint_style.lean
+++ b/scripts/lint_style.lean
@@ -26,10 +26,9 @@ def lintStyleCli (args : Cli.Parsed) : IO UInt32 := do
   for s in ["Archive.lean", "Counterexamples.lean", "Mathlib.lean"] do
     let n ← lintAllFiles (System.mkFilePath [s]) errorStyle
     numberErrorFiles := numberErrorFiles + n
-  -- Make sure this return value can be used further in shell scripts:
-  -- exit codes 125 and 126 are disallowed, as is any number higher than 128.
-  if numberErrorFiles == 128 then return 128
-  else return min numberErrorFiles 125
+  -- Make sure to return an exit code of at most 125, so this return value can be used further
+  -- in shell scripts.
+  return min numberErrorFiles 125
 
 /-- Setting up command line options and help text for `lake exe lint_style`. -/
 -- so far, no help options or so: perhaps that is fine?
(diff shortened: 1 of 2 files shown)


grunweg:
@adomani Since you raised this issue (thanks again for pointing me in this direction!), would you like to take a look?

scripts/lint_style.lean
```diff
@@ -26,7 +26,9 @@ def lintStyleCli (args : Cli.Parsed) : IO UInt32 := do
   for s in ["Archive.lean", "Counterexamples.lean", "Mathlib.lean"] do
     let n ← lintAllFiles (System.mkFilePath [s]) errorStyle
     numberErrorFiles := numberErrorFiles + n
-  return numberErrorFiles
+  -- Make sure to return an exit code of at most 125, so this return value can be used further
+  -- in shell scripts.
+  return min numberErrorFiles 125
```
adomani, line 31:
```suggestion
  return min (numberErrorFiles % 128) 125
```
I wonder whether this can be (ever so slightly) more meaningful:
* reduce module 128=2^7 (since 126 and 127 are excluded for other reasons, but 128 appears to be the meaningful number);
* truncate at 125 to avoid 126 and 127.

scripts/mk_all.lean
```diff
@@ -68,7 +68,9 @@ def mkAllCLI (args : Parsed) : IO UInt32 := do
       updates := updates + 1
   if updates == 0 then
     IO.println "No update necessary"
-  return updates
+  -- Make sure to return an exit code of at most 125, so this return value can be used further
+  -- in shell scripts.
+  return min updates 125
```
adomani, line 73:
```suggestion
  return min (numberErrorFiles % 128) 125
```
See above for a justification.

adomani (approved):
This looks great, thanks!

~~Modulo the modulo,~~

maintainer merge

scripts/lint_style.lean
grunweg: But this would mishandle the case of exactly 128 errors, right? I just pushed a commit allowing 128 explicitly - it is somewhat longer. Do you think returning an exit code of 128 is an important feature? I am not convinced the longer code is worth it, but could be swayed otherwise.

scripts/mk_all.lean
grunweg: Changed this similarly; let's keep the discussion in the other comment.

scripts/lint_style.lean
adomani: Sorry, I got this wrong.  We should special-case *actual value* 0 and everything else.

How about something like
```lean
if numberErrorFiles == 0 then return 0 else
  return max 1 (numberErrorFiles % 125)
```

grunweg: This makes more sense! Why should we do *modulo* though, instead of saturating? I'd prefer 130 files to return a code of 125 over 2 (or 5), WDYT?

adomani: Sure, I do not have a preference.  There might even be some value into specifying that exit code 1 means there were some errors and leave other error codes for potentially other meanings.  Feel free to pick what you prefer: let's simply make sure that exit code 0 is if and only if `numberErrorFiles = 0`!  :smile:

grunweg: I'll let a maintainer weigh in on their preference: I have a weak preference for simply reverting the most recent commit; I'm happy to do either of the other options.

semorrison: Saturation sounds better to me, but we're in the weeds here. :-)

semorrison:
bors d+

grunweg:
It seems that the current solution satisfies everybody - let's land this. Thanks for the reviews and the keen comments!
bors r+

PR #14273, opened 2024-06-30
feat: allow ignoring individual style linter errors; rewrite `update-style-exceptions` in Lean

grunweg:
- add an --update flag to just add exceptions for all current errors
(People have asked for this since the "file length linter" was created; now it's easily possible!)
- reimplement the `update-style-exceptions` script in Lean, as `lean exe lint_style --regenerate`.
As long as there are Python style linters, this calls into the Python linters to update the output.
(In effect, this PR flips the call responsibility: now the Lean code calls the Python code, not vice versa.)

This simplifies the style linter infrastructure: two shell and one python script are replaced by one Lean executable and one Python script (which ought to be transient, and disappear when all Python linters are rewritten).

---

- [ ] depends on: #14697
Rebased and cleaned up version of #13623. Commits can be reviewed individually.


grunweg:
@adomani @semorrison @YaelDillies Would one of you like to review another "rewrite Python style linters" PR?

.github/workflows/nolints.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         shell: bash
         run: |
           env LEAN_ABORT_ON_PANIC=1 lake exe runLinter --update Mathlib
-          ./scripts/update-style-exceptions.sh
+          lake exe lint_style --regenerate
```
YaelDillies, line 60: Why is the flag called `--regenerate` and not `--update` as in `lake exe shake --update`?

grunweg: I considered that --- note that I defined *two* flags, --update and --regenerate.
The former only ignores new errors: this is suitable when you e.g. don't want to split a file which your change pushed over the limit. The latter regenerates all exceptions; in practice, this will incur more churn since the current number of lines is encoded in the file... so, to me only making the latter change occasionally (but not as a side effect) to intentional.

With shake, --update corresponds to my --regenerate, so is a super-set of this.
In terms of consistency, I think this is fine: the default action should be to call --update; *not* --regenerate. (If I get around to fixing the nolints bot, the file will be regenerated once a week.)

I could rename `--update` to `--ignore` if this is clearer; perhaps even with an error on using `--update` to clarify the distinction.

YaelDillies: So `--regenerate` is the same as "delete `nostyle`" + `--update`, right? Maybe it doesn't need to exist as a separate flag

grunweg: Yup. But then, why did we have `update-style-exceptions.py`? Or, put differently, how are we going to *remove* entries from the list? Automatically updating seems useful enough (to expert users) to make it low-friction; it just doesn't need to be the *first* operation users are exposed to.

YaelDillies: Well you are worried about unnecessarily updating existing entries, not about correctly removing redundant entries, right? So `--update` could
* Recompute the file
* Add the lines to be added
* Remove the lines to be removed
* Not touch the lines that are merely modified (so long as not touching them is not causing a linting error)

grunweg: That will be more implementation work: but indeed, that is possible. Let me ask on zulip what a good user experience would be.

grunweg:
The branch MR-regenerate-style-exceptions2 has a rewrite of these changes, blocked on preliminary refactoring in #14676.
(The branch is functionally complete, with one small bug to fix.)

grunweg:
Closing, as #14697 has landed now.

PR #14274, opened 2024-06-30
[Merged by Bors] - feat: the Dialectica category

digama0:
Co-authored with Valeria de Paiva. This defines the category `Dial(C)` which is a categorical interpretation of Gödel's "Dialectica interpretation" of higher order arithmetic.

---


Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,98 @@
...
+  ([pdf](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-213.pdf)).
+-/
+
+noncomputable section
+
+namespace CategoryTheory
+
+open Limits
+
+theorem Functor.map_le_map {C D} [Preorder C] [Preorder D]
```
joelriou, line 25: I believe this is `Functor.monotone` in `CategoryTheory.Category.Preorder`.

Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,98 @@
...
+  rel : Subobject (src ⨯ tgt)
+
+namespace Dial
+
+local notation "π₁" => prod.fst
+local notation "π₂" => prod.snd
+local notation "π(" a ", " b ")" => prod.lift a b
+
+/-- A morphism in the `Dial C` category from `⟨U, X, α⟩` to `⟨V, Y, β⟩` is a pair
+`(f : U ⟶ V, F : U ⨯ Y ⟶ X)` such that `α(u, F(u, y)) ≤ β(f(u), y)`. -/
```
joelriou, line 50: The expression `α(u, F(u, y)) ≤ β(f(u), y)` does not fully makes sense to me. At the very least, I find it more confusing. It would be nice to at least say explicitly that this is an inequality in subobjects of `V ⨯ Y`. Also, writing `{ α(u, F(u, y)) } ≤ { β(f(u), y) }` would makes things more clear.

digama0: Would it be better to use subset notation like `{ (u,y) | α(u, F(u, y)) } ⊆{ (u,y) | β(f(u), y) }`? For me the definition with point-free descriptions of pullbacks is completely impenetrable, and I need to frequently consult this "internal language" version in order to make sense of things. (It's also not an inequality in `V ⨯ Y`, it's in `U ⨯ Y`. I was hoping the choice of letters would make the typing clear.)

semorrison:
While I'm happy to see monoidal closed categories being used for something I know nothing about, I also know nothing about the dialectica category!

Might you, or perhaps Valeria, be able to say something here or in the module-doc about what this will be used for? (Or if this will be a leaf node in Mathlib?)

Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,96 @@
...
+local notation "π(" a ", " b ")" => prod.lift a b
+
+/-- A morphism in the `Dial C` category from `⟨U, X, α⟩` to `⟨V, Y, β⟩` is a pair
+`(f : U ⟶ V, F : U ⨯ Y ⟶ X)` such that `{(u,y) | α(u, F(u, y))} ≤ {(u,y) | β(f(u), y)}`. -/
+@[ext] structure Hom (X Y : Dial C) where
+  /-- Maps the sources -/
+  f : X.src ⟶ Y.src
+  /-- Maps the targets (contravariantly) -/
+  F : X.src ⨯ Y.tgt ⟶ X.tgt
+  /-- This says `α(u, F(u, y)) ≤ β(f(u), y)` using subobject pullbacks -/
```
joelriou, line 54: This docstring should also be updated.

Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,96 @@
...
+  }
+  id_comp f := by simp; rfl
+  comp_id f := by simp; rfl
+  assoc f g h := by
+    simp only [Category.assoc, Hom.mk.injEq, true_and]
+    rw [← Category.assoc, ← Category.assoc]; congr 1
+    ext <;> simp
+
+@[ext] theorem Hom.ext' {X Y : Dial C} :
+    ∀ (x y : X ⟶ Y), x.f = y.f → x.F = y.F → x = y := Hom.ext
```
joelriou, line 96: This is the standard name we use in multiple other places in category theory:
```suggestion
@[ext] theorem hom_ext {X Y : Dial C} {x y : X ⟶ Y} (h : x.f = y.f) : x = y := by
   Hom.ext x y h
```

Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,96 @@
...
+import Mathlib.CategoryTheory.Limits.Shapes.FiniteProducts
+
+/-!
+# Dialectica category
+
+We define the category `Dial` of the Dialectica interpretation, after:
+
+* Valeria de Paiva, The Dialectica Categories.
+  University of Cambridge, Computer Laboratory, PhD Thesis, Technical Report 213, 1991
+  ([pdf](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-213.pdf)).
```
joelriou, line 16: Could you add this to the `docs/references.bib` file?

Mathlib/CategoryTheory/Dialectica/Basic.lean
digama0:
```suggestion
  /-- This says `{(u, y) | α(u, F(u, y))} ⊆  {(u, y) | β(f(u), y)}` using subobject pullbacks -/
```

joelriou:
I think it would be nice to include a constructor for isomorphisms in `Basic.lean`, something like
```lean
@[simps]
def isoMk {X Y : Dial C} (e₁ : X.src ≅ Y.src) (e₂ : X.tgt ≅ Y.tgt)
    (eq : X.rel = (Subobject.pullback (prod.map e₁.hom e₂.hom)).obj Y.rel) : X ≅ Y where
  hom := { f := e₁.hom, F := π₂ ≫ e₂.inv, le := sorry }
  inv := { f := e₁.inv, F := π₂ ≫ e₂.hom, le := sorry }
```
(I am not completely sure what is the most convenient phrasing of `eq` here.)

Then in `Monoidal.lean`, definitions would be easier because only one compatibility would be necessary:
```lean
/-- Left unit cancellation `1 ⊗ X ≅ X` in `Dial C`. -/
@[simps!]
def leftUnitor (X : Dial C) : tensorObj tensorUnit X ≅ X :=
  isoMk (Limits.prod.leftUnitor _) (Limits.prod.leftUnitor _) sorry

/-- Right unit cancellation `X ⊗ 1 ≅ X` in `Dial C`. -/
@[simps!] def rightUnitor (X : Dial C) : tensorObj X tensorUnit ≅ X :=
  isoMk (Limits.prod.rightUnitor _) (Limits.prod.rightUnitor _) sorry

/-- The associator for tensor, `(X ⊗ Y) ⊗ Z ≅ X ⊗ (Y ⊗ Z)` in `Dial C`. -/
@[simps!]
def associator (X Y Z : Dial C) : tensorObj (tensorObj X Y) Z ≅ tensorObj X (tensorObj Y Z) :=
  isoMk (Limits.prod.associator _ _ _) (Limits.prod.associator _ _ _) sorry
```

Could you split this PR into two? One for `Basic` and `Monoidal`, and a second one for `Closed`.

Mathlib/CategoryTheory/Dialectica/Basic.lean
```diff
@@ -0,0 +1,118 @@
...
+  let F2 := prod.map F.f (𝟙 Z.tgt)
+  have h1 := (Subobject.pullback F1).monotone F.le
+  have h2 := (Subobject.pullback F2).monotone G.le
+  rw [← Subobject.pullback_comp, ← Subobject.pullback_comp] at h1 h2
+  rw [(_ : F1 ≫ _ = _)] at h1
+  rw [(_ : F2 ≫ _ = _), (_ : F2 ≫ _ = _)] at h2
+  · exact le_trans h1 h2
+  · simp [F2]
+  · simp [F1, F2]
+  · simp [F1]
```
joelriou, line 74:
```suggestion
theorem comp_le_lemma {X Y Z : Dial C} (F : Dial.Hom X Y) (G : Dial.Hom Y Z) :
    (Subobject.pullback π(π₁, π(π₁, prod.map F.f (𝟙 _) ≫ G.F) ≫ F.F)).obj X.rel ≤
    (Subobject.pullback (prod.map (F.f ≫ G.f) (𝟙 Z.tgt))).obj Z.rel := by
  refine' le_trans _
    (((Subobject.pullback (π(π₁, prod.map F.f (𝟙 _) ≫ G.F))).monotone F.le).trans
    (le_trans _ (((Subobject.pullback (prod.map F.f (𝟙 Z.tgt))).monotone G.le).trans _)))
  all_goals simp [← Subobject.pullback_comp]
```

Mathlib/CategoryTheory/Dialectica/Monoidal.lean
```diff
@@ -0,0 +1,154 @@
...
+      tensorHom (𝟙 X) (braiding Y Z).hom ≫ (associator X Z Y).inv ≫
+      tensorHom (braiding X Z).hom (𝟙 Y) := by aesop_cat
+
+instance : SymmetricCategory (Dial C) where
+  braiding := braiding
+  braiding_naturality_right := braiding_naturality_right
+  braiding_naturality_left := braiding_naturality_left
+  hexagon_forward := hexagon_forward
+  hexagon_reverse := hexagon_reverse
+  symmetry := symmetry
```
joelriou, line 154:
```suggestion
  symmetry := symmetry

end Dial

end CategoryTheory
```

joelriou:
After the suggested changes are done and additional documentation at the beginning of the `Basic` file is added, this can be merged.

vcvpaiva:
hi, a proposed additional documentation to be added at the beginning of the Basic file. please feel free to cut it to size, if too verbose.

Dialectica categories are important models of linear type theory. They satisfy most of the distinctions that linear logic was meant to introduce and many models do not satisfy, like the independence of constants. Many linear type theories are being used at the moment--https://ncatlab.org/nlab/show/linear+type+theory describes some of them: for quantum systems, for effects in programming, for linear dependent types. In particular, dialectica categories are connected to polynomial functors, being a slightly more sophisticated version of polynomial types, as discussed, for instance, in Moss and von Glehn's `Dialectica models of type theory' https://arxiv.org/abs/2105.00283. As such they are related to the polynomial constructions being developed by Awodey, Riehl, and Hazratpour. For the non-dependent version developed here several applications are known to Petri Nets, small cardinals in Set Theory, state in imperative programming, and others, see https://github.com/vcvpaiva/DialecticaCategories.

digama0:
I edited Valeria's paragraph above into the file. Let me know if I should make any additional edits to make it better fit into the standard module doc style.

PR #14275, opened 2024-06-30
[Merged by Bors] - feat(LocallyConvex): add `isVonNBounded_union` etc

urkud:
---


Mathlib/Analysis/LocallyConvex/Bounded.lean
```diff
@@ -92,11 +92,19 @@ theorem IsVonNBounded.subset {s₁ s₂ : Set E} (h : s₁ ⊆ s₂) (hs₂ : Is
     IsVonNBounded 𝕜 s₁ := fun _ hV => (hs₂ hV).mono_right h
 #align bornology.is_vonN_bounded.subset Bornology.IsVonNBounded.subset
 
+@[simp]
+theorem isVonNBounded_union {s t : Set E} :
```
eric-wieser, line 96: Can you add the `iUnion` version too (we already have `absorbs_iUnion`)? Don't forget to ensure the index is in `Sort` not `Type`!

eric-wieser (approved):
bors d+

PR #14278, opened 2024-06-30
feat(SetTheory/Surreal/Basic): add some simp lemmas for Surreal.mk

hwatheod:
add some simp lemmas for Surreal.mk

---

### DIFF
--- a/Mathlib/SetTheory/Surreal/Basic.lean
+++ b/Mathlib/SetTheory/Surreal/Basic.lean
@@ -301,6 +301,15 @@ instance : One Surreal :=
 instance : Inhabited Surreal :=
   ⟨0⟩
 
+@[simp]
+lemma mk_eq_mk {x y : PGame.{u}} {hx hy} : mk x hx = mk y hy ↔ x ≈ y := by
+  rw [mk, mk, Quotient.eq]; rfl
+
+@[simp]
+lemma mk_eq_zero {x : PGame.{u}} {hx} : mk x hx = 0 ↔ x ≈ 0 := by
+  change mk x hx = mk 0 numeric_zero ↔ x ≈ 0
+  rw [mk_eq_mk]
+
 /-- Lift an equivalence-respecting function on pre-games to surreals. -/
 def lift {α} (f : ∀ x, Numeric x → α)
     (H : ∀ {x y} (hx : Numeric x) (hy : Numeric y), x.Equiv y → f x hx = f y hy) : Surreal → α :=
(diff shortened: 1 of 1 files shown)


Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -301,6 +301,15 @@ instance : One Surreal :=
...
   ⟨0⟩
 
+@[simp]
+lemma mk_eq_mk {x y : PGame.{u}} {hx hy} : mk x hx = mk y hy ↔ x ≈ y := by
+  rw [mk, mk, Quotient.eq]; rfl
+
+@[simp]
+lemma mk_eq_zero {x : PGame.{u}} {hx} : mk x hx = 0 ↔ x ≈ 0 := by
+  change mk x hx = mk 0 numeric_zero ↔ x ≈ 0
+  rw [mk_eq_mk]
```
eric-wieser, line 311:
```suggestion
@[simp]
lemma mk_eq_mk {x y : PGame.{u}} {hx hy} : mk x hx = mk y hy ↔ x ≈ y := Quotient.eq

@[simp]
lemma mk_eq_zero {x : PGame.{u}} {hx} : mk x hx = 0 ↔ x ≈ 0 := Quotient.eq
```

eric-wieser (commented):
I'm not sure these should be `@[simp]`; [`QuotientAddGroup.mk'_eq_mk'`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/GroupTheory/QuotientGroup.html#QuotientAddGroup.mk'_eq_mk') is morally very similar, and is also not `simp`.

Mathlib/SetTheory/Surreal/Basic.lean
hwatheod: Are there guidelines for when something should be `@[simp]` ? If I don't tag these as `@[simp]`, then I need to list them explicitly within `simp [...]` when I use them. In my current work, there is just one use of each of these two lemmas, and there may a few more later, but I don't anticipate a huge number.

hwatheod: The `mk_le_mk` a few lines below these changes is currently `@[simp]`.

hwatheod: Keeping the `@[simp]` for now.

hwatheod:
Closing this PR for now as further work as made it less and less clear that we want these to be `@[simp]` lemmas. May re-open in the future.

eric-wieser:
Even if the value of `simo` is dubious, having these lemmas without simp seems like a good thing to me, and I'd be happy to merge that version

hwatheod:
Re-opening with `simp` removed and some additional lemmas added, which are needed for the division proof

hwatheod:
Reopning

PR #14281, opened 2024-06-30
[Merged by Bors] - feat: add `LinearMap.IsAlt.eq_of_add_add_eq_zero`

seewoo5:
For an alternating form $B : M \times M \to R$ and $a, b, c \in M$ with $a + b+ c = 0$, we have $B(a, b) = B(b, c)$.
This can be used to prove the same property of Wronskian: see #14243.

---

### DIFF
--- a/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
+++ b/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
@@ -158,11 +158,7 @@ theorem isRefl (H : B₁.IsAlt) : B₁.IsRefl := LinearMap.IsAlt.isRefl H
 #align bilin_form.is_alt.is_refl LinearMap.BilinForm.IsAlt.isRefl
 
 theorem eq_of_add_add_eq_zero [IsCancelAdd R] {a b c : M} (H : B.IsAlt) (hAdd : a + b + c = 0) :
-    B a b = B b c := by
-  have : B a a + B a b + B a c = B a c + B b c + B c c := by
-    simp_rw [← map_add, ← map_add₂, hAdd, map_zero, LinearMap.zero_apply]
-  rw [H, H, zero_add, add_zero, add_comm] at this
-  exact add_left_cancel this
+    B a b = B b c := LinearMap.eq_of_add_add_eq_zero H hAdd
 
 protected theorem add {B₁ B₂ : BilinForm R M} (hB₁ : B₁.IsAlt) (hB₂ : B₂.IsAlt) : (B₁ + B₂).IsAlt :=
   fun x => (congr_arg₂ (· + ·) (hB₁ x) (hB₂ x) : _).trans <| add_zero _
--- a/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
+++ b/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
@@ -186,7 +186,7 @@ theorem isAlt_neg {B : BilinForm R₁ M₁} : (-B).IsAlt ↔ B.IsAlt :=
 #align bilin_form.is_alt_neg LinearMap.BilinForm.isAlt_neg
 
 theorem Alt_eq_of_sum_zero {B : BilinForm R₁ M₁} {a b c : M₁} (hB : B.IsAlt) (hSum : a + b + c = 0) :
-  B a b = B b c := by
+    B a b = B b c := by
   rw [← neg_eq_iff_add_eq_zero] at hSum
   rw [← hSum, B.neg_right, B.add_right, hB.self_eq_zero, ← hB.neg_eq, add_zero]
 
--- a/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
+++ b/Mathlib/LinearAlgebra/BilinearForm/Properties.lean
@@ -185,6 +185,11 @@ theorem isAlt_neg {B : BilinForm R₁ M₁} : (-B).IsAlt ↔ B.IsAlt :=
   ⟨fun h => neg_neg B ▸ h.neg, IsAlt.neg⟩
 #align bilin_form.is_alt_neg LinearMap.BilinForm.isAlt_neg
 
+theorem Alt_eq_of_sum_zero {B : BilinForm R₁ M₁} {a b c : M₁} (hB : B.IsAlt) (hSum : a + b + c = 0) :
+  B a b = B b c := by
+  rw [← neg_eq_iff_add_eq_zero] at hSum
+  rw [← hSum, B.neg_right, B.add_right, hB.self_eq_zero, ← hB.neg_eq, add_zero]
+
 /-! ### Linear adjoints -/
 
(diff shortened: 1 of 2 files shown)


Mathlib/LinearAlgebra/BilinearForm/Properties.lean
```diff
@@ -185,6 +185,11 @@ theorem isAlt_neg {B : BilinForm R₁ M₁} : (-B).IsAlt ↔ B.IsAlt :=
   ⟨fun h => neg_neg B ▸ h.neg, IsAlt.neg⟩
 #align bilin_form.is_alt_neg LinearMap.BilinForm.isAlt_neg
 
+theorem Alt_eq_of_sum_zero {B : BilinForm R₁ M₁} {a b c : M₁} (hB : B.IsAlt) (hSum : a + b + c = 0) :
```
eric-wieser, line 188:
```suggestion
theorem isAlt.eq_of_add_add_eq_zero {B : BilinForm R₁ M₁} {a b c : M₁} (hB : B.IsAlt) (hSum : a + b + c = 0) :
```
which means this can be use as `hB.eq_of_add_add_eq_zero` at the caller.

Note that `sum` is reserved for things like `List.sum` or `Finset.sum`, and we use `add` for expressions with `+`.

eric-wieser: In fact, if you move this above the `end IsAlt`, then you can drop the `IsAlt` prefix as it is added automatically

eric-wieser (commented):
Could you prove this for `LinearMap.IsAlt` too?

eric-wieser:
Here's a proof that works for semirings and not just rings:
```lean
theorem isAlt.eq_of_add_add_eq_zero [IsCancelAdd R] {B : LinearMap.BilinForm R M} {a b c : M} (hB : B.IsAlt) (hSum : a + b + c = 0) :
    B a b = B b c := by
  have : B a a + B a b + B a c = B a c + B b c + B c c := by
    simp_rw [← map_add, ← map_add₂, hSum, map_zero, zero_apply]
  rw [hB, hB, zero_add, add_zero, add_comm] at this
  exact add_left_cancel this
```
I'll leave adapting this to `LinearMap.IsAlt` to you :)

seewoo5:
> Here's a proof that works for semirings and not just rings:
> 
> ```lean
> theorem isAlt.eq_of_add_add_eq_zero [IsCancelAdd R] {B : LinearMap.BilinForm R M} {a b c : M} (hB : B.IsAlt) (hSum : a + b + c = 0) :
>     B a b = B b c := by
>   have : B a a + B a b + B a c = B a c + B b c + B c c := by
>     simp_rw [← map_add, ← map_add₂, hSum, map_zero, zero_apply]
>   rw [hB, hB, zero_add, add_zero, add_comm] at this
>   exact add_left_cancel this
> ```
> 
> I'll leave adapting this to `LinearMap.IsAlt` to you :)

I naively put this above `end IsAlt` as `theorem eq_of_add_add_eq_zero ...`, and it says that `zero_apply` in `simp_rw` made no progress. Do you have any idea?

eric-wieser:
It might need to be `LinearMap.zero_apply`. If you push what you tried, I can help more.

Another trick would be to just remove that lemma from the `simp_rw`, then use `simp?` to find what the lemma is really called.

seewoo5:
Updated (almost no change from your suggestion, which also worked well for `LinearMap.IsAlt`. I changed some names of variables and remove unnecessary terms, too.

Actually, I don't quite understand why we don't need `IsCancelAdd R₁` hypothesis for the same theorem in `SesquilinearForm.lean`. I was not able to find the relevant assumption in the file, but it works well without explicitly assuming it in the theorem, so I guess I'm missing the hypothesis which is somewhere in the file.

Mathlib/LinearAlgebra/SesquilinearForm.lean
```diff
@@ -304,6 +304,12 @@ theorem ortho_comm {x y} : IsOrtho B x y ↔ IsOrtho B y x :=
   H.isRefl.ortho_comm
 #align linear_map.is_alt.ortho_comm LinearMap.IsAlt.ortho_comm
 
+theorem eq_of_add_add_eq_zero {a b c : M₁} (hAdd : a + b + c = 0) : B a b = B b c := by
```
eric-wieser, line 307: You should move this up to the `AddCommMonoid` section above, where the `IsCancelAdd` argument will need to return.

You don't need it here because you have `AddCommGroup`.

Mathlib/LinearAlgebra/BilinearForm/Properties.lean
```diff
@@ -157,6 +157,13 @@ theorem neg_eq (H : B₁.IsAlt) (x y : M₁) : -B₁ x y = B₁ y x := LinearMap
 theorem isRefl (H : B₁.IsAlt) : B₁.IsRefl := LinearMap.IsAlt.isRefl H
 #align bilin_form.is_alt.is_refl LinearMap.BilinForm.IsAlt.isRefl
 
+theorem eq_of_add_add_eq_zero [IsCancelAdd R] {a b c : M} (H : B.IsAlt) (hAdd : a + b + c = 0) :
+    B a b = B b c := by
+  have : B a a + B a b + B a c = B a c + B b c + B c c := by
+    simp_rw [← map_add, ← map_add₂, hAdd, map_zero, LinearMap.zero_apply]
+  rw [H, H, zero_add, add_zero, add_comm] at this
+  exact add_left_cancel this
```
eric-wieser, line 165:
```suggestion
    B a b = B b c := LinearMap.IsAlt.eq_of_add_add_eq_zero H hAdd
```

Mathlib/LinearAlgebra/SesquilinearForm.lean
seewoo5: So if I understood correctly, since the theorem itself is true for `AddCommMonoid`, not just for `AddCommGroup`, it is better to be in `AddCommMonoid` section with `IsCancelAdd` argument, right?

Mathlib/LinearAlgebra/SesquilinearForm.lean
```diff
@@ -276,6 +276,13 @@ theorem IsAlt.self_eq_zero (x : M₁) : B x x = 0 :=
   H x
 #align linear_map.is_alt.self_eq_zero LinearMap.IsAlt.self_eq_zero
 
+theorem eq_of_add_add_eq_zero [IsCancelAdd M] {a b c : M₁} (hAdd : a + b + c = 0) :
```
eric-wieser, line 279:
```suggestion
theorem IsAlt.eq_of_add_add_eq_zero [IsCancelAdd M] {a b c : M₁} (hAdd : a + b + c = 0) :
```
Since you're no longer in a namespace here

eric-wieser:
bors merge

Thanks, this now looks great!

PR #14282, opened 2024-06-30
[Merged by Bors] - feat: more lemmas about `List.dedup` and other lattice operations

eric-wieser:
Also the trivial `Multiset` wrappers.

---

### DIFF
--- a/Mathlib/Data/List/Dedup.lean
+++ b/Mathlib/Data/List/Dedup.lean
@@ -145,7 +145,8 @@ theorem dedup_map_of_injective [DecidableEq β] {f : α → β} (hf : Function.I
     · rw [dedup_cons_of_not_mem h, dedup_cons_of_not_mem <| (mem_map_of_injective hf).not.mpr h, ih,
         map_cons]
 
-theorem Subset.dedup_append {xs ys : List α} (h : xs ⊆ ys) :
+/-- Note that the weaker `List.Subset.dedup_append_left` is proved later. -/
+theorem Subset.dedup_append_right {xs ys : List α} (h : xs ⊆ ys) :
     dedup (xs ++ ys) = dedup ys := by
   rw [List.dedup_append, Subset.union_eq_right (h.trans <| subset_dedup _)]
 
--- a/Mathlib/Data/List/Dedup.lean
+++ b/Mathlib/Data/List/Dedup.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
 import Mathlib.Data.List.Nodup
+import Mathlib.Data.List.Lattice
 
 #align_import data.list.dedup from "leanprover-community/mathlib"@"d9e96a3e3e0894e93e10aff5244f4c96655bac1c"
 
@@ -24,7 +25,7 @@ universe u
 
 namespace List
 
-variable {α : Type u} [DecidableEq α]
+variable {α β : Type*} [DecidableEq α]
 
 @[simp]
 theorem dedup_nil : dedup [] = ([] : List α) :=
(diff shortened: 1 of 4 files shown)


Mathlib/Data/List/Dedup.lean
```diff
@@ -132,6 +133,41 @@ theorem dedup_append (l₁ l₂ : List α) : dedup (l₁ ++ l₂) = l₁ ∪ ded
   · rw [dedup_cons_of_not_mem' h, insert_of_not_mem h]
 #align list.dedup_append List.dedup_append
 
+theorem dedup_map_of_injective [DecidableEq β] {f : α → β} (hf : Function.Injective f)
+    (xs : List α) :
+    (xs.map f).dedup = xs.dedup.map f := by
```
YaelDillies, line 138: @Timeroot had this lemma somewhere. Did it ever land? What is the PR?

Mathlib/Data/List/Dedup.lean
```diff
@@ -132,6 +133,41 @@ theorem dedup_append (l₁ l₂ : List α) : dedup (l₁ ++ l₂) = l₁ ∪ ded
...
+  induction xs with
+  | nil => simp
+  | cons x xs ih =>
+    rw [map_cons]
+    by_cases h : x ∈ xs
+    · rw [dedup_cons_of_mem h, dedup_cons_of_mem (mem_map_of_mem f h), ih]
+    · rw [dedup_cons_of_not_mem h, dedup_cons_of_not_mem <| (mem_map_of_injective hf).not.mpr h, ih,
+        map_cons]
+
+theorem Subset.dedup_append {xs ys : List α} (h : xs ⊆ ys) :
```
YaelDillies, line 148:
```suggestion
theorem Subset.dedup_append_left {xs ys : List α} (h : xs ⊆ ys) :
```
and can we have the right version stated using `List.Perm`?

Mathlib/Data/Multiset/FinsetOps.lean
```diff
@@ -280,6 +289,12 @@ theorem ndinter_eq_zero_iff_disjoint {s t : Multiset α} : ndinter s t = 0 ↔ D
   rw [← subset_zero]; simp [subset_iff, Disjoint]
 #align multiset.ndinter_eq_zero_iff_disjoint Multiset.ndinter_eq_zero_iff_disjoint
 
+alias ⟨_, Disjoint.ndinter_eq_zero⟩ := ndinter_eq_zero_iff_disjoint
+
+theorem Subset.ndinter_eq_left {s t : Multiset α} (h : s ⊆ t) : s.ndinter t = s := by
```
YaelDillies, line 294: Same, can we have the right verSion?

Mathlib/Data/List/Dedup.lean
YaelDillies: Hmm maybe, but the fact it's missing compared to the other direction is quite upsetting and might cause confusion in the future. At least add a comment explaining why the other lemma doesn't exist

eric-wieser: I added the lemma, but in the multiset file where the proof was trivial.

Mathlib/Data/Multiset/FinsetOps.lean
eric-wieser: I don't think the right version is meaningfully true? What statement do you want?

YaelDillies: Ah indeed. But then this lemma is something like `Multiset.filter_true_of_mem h` except that `filter_true_of_mem` doesn't exist

Mathlib/Data/Multiset/Dedup.lean
```diff
@@ -132,6 +137,25 @@ theorem Nodup.le_dedup_iff_le {s t : Multiset α} (hno : s.Nodup) : s ≤ t.dedu
...
+theorem Subset.dedup_add_left {s t : Multiset α} (h : t ⊆ s) :
+    dedup (s + t) = dedup s := by
+  rw [add_comm, Subset.dedup_add_right h]
+
+theorem Disjoint.dedup_add {s t : Multiset α} (h : Disjoint s t) :
+    dedup (s + t) = dedup s + dedup t := by
+  induction s, t using Quot.induction_on₂
+  exact congr_arg ((↑) : List α → Multiset α) <| List.Disjoint.dedup_append h
+
+/-- Note that the tronger `List.Subset.dedup_append_right` is proved earlier. -/
```
YaelDillies, line 154:
```suggestion
/-- Note that the stronger `List.Subset.dedup_append_right` is proved earlier. -/
```

Mathlib/Data/Multiset/FinsetOps.lean
eric-wieser: It exists for `List` though; I golfed the proof there

Mathlib/Data/List/Dedup.lean
eric-wieser: I assume we don't need to wait for an answer here? Of course, if @Timeroot comes along with a shorter proof after this is merged, we can always make a golf PR.

YaelDillies (approved):
Looks good now

maintainer merge

eric-wieser:
maintainer merge

The conspiracy for bots to ignore Yael continues

mattrobball:
!bench

mattrobball:
I would guess the import change and the single simp are good but benched so I don't have to speculate. If it comes back reasonable, please merge. Thanks!

bors delegate+

PR #14284, opened 2024-06-30
refactor: drop `MulEquivClass`

urkud:
---
If we decide to go with it, then I'll restore old theorem names as deprecated aliases.


urkud:
!bench

test/TCSynth.lean
```diff
@@ -87,7 +87,7 @@ section
 -- Initial issue: https://github.com/leanprover-community/mathlib4/issues/12232
 
 open Equiv in
-set_option synthInstance.maxHeartbeats 9000 in
+set_option synthInstance.maxHeartbeats 11000 in
```
urkud, line 90: I had to increase `synthInstance.maxHeartbeats` here.

urkud:
It looks like the consensus on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Drop.20MulEquivClass.3F) is to drop TODO instead.

PR #14204, opened 2024-06-27
[Merged by Bors] - chore: register `TensorProduct` with `induction_eliminator`

eric-wieser:
Split from #12605

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14285, opened 2024-06-30
[Merged by Bors] - feat: `QuadraticForm.baseChange` is unique when it exists

eric-wieser:
Proof adapted from @ScottCarnahan's [Zulip message](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Base.20change.20for.20bilinear.20maps.20and.20quadratic.20forms/near/375115826)

---

- [x] depends on: #14204


kbuzzard:
Very nice!

bors merge

PR #14288, opened 2024-06-30
[Merged by Bors] - chore: register `Sym2` with `induction`

eric-wieser:
Split from #12605

Also cleans up some proofs that used `revert` instead of `induction`.

---


TwoFX (approved):
bors r+

PR #14289, opened 2024-06-30
[Merged by Bors] - feat: drop sfiniteness in a lemma on L^p norm for a restricted measure

sgouezel:
---


fpvandoorn (commented):
Thanks!

bors merge

PR #13827, opened 2024-06-14
[Merged by Bors] - feat: If `s ≤ t` in colex, then `s \ {a} ≤ t \ {min t}` in colex

YaelDillies:
This proof is a horrible case bash. I have already spent a significant amount of time optimising it.

From LeanCamCombi

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14292, opened 2024-06-30
[Merged by Bors] - feat: `BilinForm.toQuadraticForm` is surjective in free modules

eric-wieser:
That is: if your module is free, you can always write `Q = B.toQuadraticMap`, which means that the base change exists.

This doesn't go as far as changing `QuadraticForm.baseChange`, as this would trade one generality (`Invertible (2 : R)`) for another (`Module.Free R M`).

---

- [x] depends on: #14215


Mathlib/Data/Sym/Sym2/Order.lean
eric-wieser, line 1: I can split this to a separate PR if that's desirable, but it is only 40 lines

Mathlib/Data/Finset/Prod.lean
```diff
@@ -428,6 +428,16 @@ theorem offDiag_insert (has : a ∉ s) : (insert a s).offDiag = s.offDiag ∪ {a
...
 
+theorem offDiag_filter_lt_eq_filter_le {ι}
+    [PartialOrder ι] [DecidableEq ι]
+    [DecidableRel (LE.le (α := ι))] [DecidableRel (LT.lt (α := ι))]
+    (s : Finset ι) :
+    s.offDiag.filter (fun i => i.1 < i.2) = s.offDiag.filter (fun i => i.1 ≤ i.2) := by
+  ext ⟨i, j⟩
+  simp only [mem_filter, mem_offDiag, ne_eq, and_congr_right_iff, and_imp]
+  rintro _ _ h
+  rw [Ne.le_iff_lt h]
```
YaelDillies, line 439: Can you use the newly introduced `Finset.filter_inj` from #13827 ?

Mathlib/Data/Sym/Sym2/Order.lean
```diff
@@ -0,0 +1,40 @@
...
+
+def sup [SemilatticeSup α] (x : Sym2 α) : α := Sym2.lift ⟨(· ⊔ ·), sup_comm⟩ x
+
+@[simp] theorem sup_mk [SemilatticeSup α] (a b : α) : s(a, b).sup = a ⊔ b := rfl
+
+def inf [SemilatticeInf α] (x : Sym2 α) : α := Sym2.lift ⟨(· ⊓ ·), inf_comm⟩ x
+
+@[simp] theorem inf_mk [SemilatticeInf α] (a b : α) : s(a, b).inf = a ⊓ b := rfl
+
+theorem inf_le_sup [Lattice α] (s : Sym2 α) : s.inf ≤ s.sup := by
```
YaelDillies, line 25:
```suggestion
protected theorem inf_le_sup [Lattice α] (s : Sym2 α) : s.inf ≤ s.sup := by
```

Mathlib/LinearAlgebra/QuadraticForm/Basis.lean
```diff
@@ -0,0 +1,77 @@
...
+    (s : Finset ι) (p : Sym2 ι → α) :
+    ∑ i in s.sym2.filter (¬ ·.IsDiag), p i =
+      ∑ i in s.offDiag.filter (fun i => i.1 < i.2), p s(i.1, i.2) := by
+  rw [Finset.offDiag_filter_lt_eq_filter_le]
+  conv_rhs => rw [← Finset.sum_subtype_eq_sum_filter]
+  refine (Finset.sum_equiv Sym2.sortEquiv.symm ?_ ?_).symm
+  · rintro ⟨⟨i₁, j₁⟩, hij₁⟩
+    simp [and_assoc]
+  · rintro ⟨⟨i₁, j₁⟩, hij₁⟩
+    simp
```
YaelDillies, line 29: `Algebra.Order.BigOperators.Group.Finset` should do

Mathlib/Data/Sym/Sym2/Order.lean
YaelDillies: Looks straightforward enough

Mathlib/LinearAlgebra/QuadraticForm/Basis.lean
eric-wieser: Let's see if this blows up the number of imports...

Mathlib/Data/Finset/Prod.lean
eric-wieser: That's about the sets being equal, not the conditions? I can add a primed lemma though.

Mathlib/Algebra/BigOperators/Group/Finset.lean
```diff
@@ -2634,3 +2636,16 @@ end AddCommMonoid
 
 @[deprecated (since := "2023-12-23")] alias Equiv.prod_comp' := Fintype.prod_equiv
 @[deprecated (since := "2023-12-23")] alias Equiv.sum_comp' := Fintype.sum_equiv
+
+-- TODO: move
```
YaelDillies, line 2640: I think it's okay here?

YaelDillies (commented):
Lean-wise this looks good. I'll let @mans0954 evaluate the math.

Mathlib/Algebra/BigOperators/Group/Finset.lean
eric-wieser: I was hoping the import-countinf bot would run

Mathlib/LinearAlgebra/QuadraticForm/Basis.lean
```diff
@@ -0,0 +1,63 @@
...
+namespace QuadraticMap
+
+variable {ι R M N} [LinearOrder ι]
+variable [CommRing R] [AddCommGroup M] [AddCommGroup N] [Module R M] [Module R N]
+
+/-- Given an ordered basis, produce a bilinear form associated with the quadratic form.
+
+Unlike `QuadraticMap.associated`, this is not symmetric; however, as a result it can be used even
+in characteristic two. When considered as a matrix, the form is triangular. -/
+noncomputable def toBilin (Q : QuadraticMap R M N) (bm : Basis ι R M) : LinearMap.BilinMap R M N :=
```
mans0954, line 25: (I think this is what I was trying to do here: https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Is.20Finset.20totally.20ordered.3F )

eric-wieser: I'm surprised I didn't have anything more interesting to say at the time

Mathlib/Algebra/BigOperators/Group/Finset.lean
eric-wieser: @adomani, do you know why the bot isn't running?

Ruben-VandeVelde:
>| File | Base Count | Head Count | Change |
>| --- | --- | --- | --- |
>| Mathlib.Algebra.BigOperators.Group.Finset | 498 | 504 | +6 (+1.20%) |

It is running, just keeps updating the hidden comment above

Mathlib/Algebra/BigOperators/Group/Finset.lean
eric-wieser: Nevermind, resolved!

adomani:
Yes, I find the fact that it keeps saying when it first posted, rather than when it was last edited, confusing.  Maybe there is a way of changing that?

YaelDillies:
I think it's a feature, so long as the comment is early enough in the list of comments. This fails to be the case only if:
* The PR predates the introduction of the PR summary bot
* The author has made a lot of commits before opening the PR

both of which should be pretty rare. So I think it's fine as it is!

mans0954 (approved):
The maths looks right to me.

YaelDillies:
maintainer merge

then!

Mathlib/Algebra/BigOperators/Group/Finset.lean
riccardobrasca: Can we at least get rid of the `TODO` line?

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Algebra/BigOperators/Group/Finset.lean
eric-wieser:
```suggestion
```

eric-wieser:
bors merge

PR #14294, opened 2024-06-30
[Merged by Bors] - feat: also lint when debug options are set

grunweg:
For the same reason as the other existing options: these are not meant to stay in production code.

---

- [ ] depends on: #14006


eric-wieser (approved):
bors merge

Thanks!

PR #14295, opened 2024-06-30
[Merged by Bors] - feat(CategoryTheory): two more versions of the Yoneda lemma

TwoFX:
---


joelriou:
Thanks!

bors merge

PR #14301, opened 2024-06-30
[Merged by Bors] - chore(Logic): reduce use of autoImplicit

grunweg:
The remaining uses are
- three declarations in Logic/Basic (which need it, somehow)
- in UnivLE, where removing this yields funky errors related to explicit universe levels
- in Logic/Equiv/Basic, which is a huge file; fixing it is less trivial

---


eric-wieser (approved):
bors merge

Thanks!

PR #14302, opened 2024-06-30
[Merged by Bors] - feat(AlgebraicGeometry): residue field of a point and evaluation

chrisflav:
Defines the residue field of a point of a locally ringed space and the corresponding evaluation map.

---

### DIFF
--- a/Mathlib/AlgebraicGeometry/GammaSpecAdjunction.lean
+++ b/Mathlib/AlgebraicGeometry/GammaSpecAdjunction.lean
@@ -69,6 +69,18 @@ def ΓToStalk (x : X) : Γ.obj (op X) ⟶ X.presheaf.stalk x :=
   X.presheaf.germ (⟨x, trivial⟩ : (⊤ : Opens X))
 #align algebraic_geometry.LocallyRingedSpace.Γ_to_stalk AlgebraicGeometry.LocallyRingedSpace.ΓToStalk
 
+lemma ΓToStalk_stalkMap {X Y : LocallyRingedSpace} (f : X ⟶ Y) (x : X) :
+    Y.ΓToStalk (f.val.base x) ≫ PresheafedSpace.stalkMap f.val x =
+      f.val.c.app (op ⊤) ≫ X.ΓToStalk x := by
+  dsimp only [LocallyRingedSpace.ΓToStalk]
+  rw [PresheafedSpace.stalkMap_germ']
+
+lemma ΓToStalk_stalkMap_apply {X Y : LocallyRingedSpace} (f : X ⟶ Y) (x : X)
+    (a : Y.presheaf.obj (op ⊤)) :
+    PresheafedSpace.stalkMap f.val x (Y.ΓToStalk (f.val.base x) a) =
+      X.ΓToStalk x (f.val.c.app (op ⊤) a) := by
+  simpa using congrFun (congrArg DFunLike.coe <| ΓToStalk_stalkMap f x) a
+
 /-- The canonical map from the underlying set to the prime spectrum of `Γ(X)`. -/
 def toΓSpecFun : X → PrimeSpectrum (Γ.obj (op X)) := fun x =>
   comap (X.ΓToStalk x) (LocalRing.closedPoint (X.presheaf.stalk x))
--- a/Mathlib/AlgebraicGeometry/ResidueField.lean
+++ b/Mathlib/AlgebraicGeometry/ResidueField.lean
@@ -1,0 +1,116 @@
+/-
+Copyright (c) 2024 Christian Merten. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Merten
+-/
+import Mathlib.AlgebraicGeometry.OpenImmersion
+import Mathlib.AlgebraicGeometry.GammaSpecAdjunction
+
+/-!
+
+# Residue fields of points
+
+Any point `x` of a locally ringed space `X` comes with a natural residue field, namely the residue
+field of the stalk at `x`. Moreover, for every open subset of `X` containing `x`, we have a canonical
+evaluation map from `Γ(X, U)` to the residue field of `X` at `x`.
+
+## Main definitions
+
+The following are in the `AlgebraicGeometry.LocallyRingedSpace` namespace:
+
+- `residueField`: the residue field of the stalk at `x`.
+- `evaluation`: for open subsets `U` of `X` containing `x`, the evaluation map from sections over `U` to
+  the residue field at `x`.
+- `evaluationMap`: a morphism of locally ringed spaces induces a morphism, i.e. extension, of
+  residue fields.
+
+-/
+
+universe u
+
+open CategoryTheory TopologicalSpace Opposite
+
+noncomputable section
+
+namespace AlgebraicGeometry.LocallyRingedSpace
+
+variable (X : LocallyRingedSpace.{u}) {U : Opens X}
+
+/-- The residue field of `X` at a point `x` is the residue field of the stalk of `X`
+at `x`. -/
+def residueField (x : X) : CommRingCat :=
+  CommRingCat.of <| LocalRing.ResidueField (X.stalk x)
+
+/--
+If `U` is an open of `X` containing `x`, we have a canonical ring map from the sections
+over `U` to the residue field of `x`.
+
+If we interpret sections over `U` as functions of `X` defined on `U`, then this ring map
+corresponds to evaluation at `x`.
+-/
+def evaluation (x : U) :
+    X.presheaf.obj (op U) ⟶ X.residueField x :=
+  X.presheaf.germ x ≫ LocalRing.residue _
+
+lemma evaluation_eq_zero_iff_mem_maximalIdeal (x : U) (f : X.presheaf.obj (op U)) :
+    X.evaluation x f = 0 ↔ (X.presheaf.germ x) f ∈ LocalRing.maximalIdeal (X.stalk x) :=
+  LocalRing.residue_eq_zero_iff _
+
+lemma evaluation_ne_zero_iff_isUnit {U : Opens X} (x : U) (f : X.presheaf.obj (op U)) :
+    X.evaluation x f ≠ 0 ↔ IsUnit ((X.presheaf.germ x) f) :=
+  LocalRing.residue_ne_zero_iff_isUnit _
+
+/-- The global evaluation map from `Γ(X, ⊤)` to the residue field at `x`. -/
+def Γevaluation (x : X) : LocallyRingedSpace.Γ.obj (op X) ⟶ X.residueField x :=
+  X.evaluation ⟨x, show x ∈ ⊤ from trivial⟩
+
+lemma Γevaluation_eq_zero_iff_mem_maximalIdeal (x : X) (f : LocallyRingedSpace.Γ.obj (op X)) :
+    X.Γevaluation x f = 0 ↔
+      (X.presheaf.germ ⟨x, show x ∈ ⊤ by trivial⟩) f ∈ LocalRing.maximalIdeal (X.stalk x) :=
+  LocalRing.residue_eq_zero_iff _
+
+lemma Γevaluation_ne_zero_iff_isUnit (x : U) (f : X.presheaf.obj (op U)) :
+    X.evaluation x f ≠ 0 ↔ IsUnit ((X.presheaf.germ x) f) :=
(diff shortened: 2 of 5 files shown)


Mathlib/AlgebraicGeometry/ResidueField.lean
```diff
@@ -0,0 +1,119 @@
...
+  mem_basicOpen_iff_evaluation_ne_zero X f ⟨x, trivial⟩
+
+variable {X Y : LocallyRingedSpace.{u}} (f : X ⟶ Y)
+
+/-- If `X ⟶ Y` is a morphism of locally ringed spaces and `x` a point of `X`, we obtain
+a morphism of residue fields in the other direction. -/
+def evaluationMap (x : X) : Y.residueField (f.val.base x) ⟶ X.residueField x :=
+  LocalRing.ResidueField.map (LocallyRingedSpace.stalkMap f x)
+
+lemma evaluation_naturality {V : Opens Y} (x : f ⁻¹ᵁ V) :
```
erdOne, line 95:
```suggestion
@[reassoc]
lemma evaluation_naturality {V : Opens Y} (x : f ⁻¹ᵁ V) :
```

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -369,6 +369,19 @@ def residue : R →+* ResidueField R :=
   Ideal.Quotient.mk _
 #align local_ring.residue LocalRing.residue
 
+variable {R}
+
+lemma residue_ker : RingHom.ker (residue R) = maximalIdeal R :=
```
erdOne, line 374:
```suggestion
lemma ker_residue : RingHom.ker (residue R) = maximalIdeal R :=
```

Mathlib/RingTheory/Ideal/LocalRing.lean
```diff
@@ -369,6 +369,19 @@ def residue : R →+* ResidueField R :=
...
+  Ideal.mk_ker
+
+lemma residue_eq_zero_iff (x : R) : residue R x = 0 ↔ x ∈ maximalIdeal R := by
+  rw [← RingHom.mem_ker, residue_ker]
+
+lemma residue_ne_zero_iff_isUnit (x : R) : residue R x ≠ 0 ↔ IsUnit x := by
+  rw [← not_iff_not, ne_eq, not_not, residue_eq_zero_iff, mem_maximalIdeal, mem_nonunits_iff]
```
erdOne, line 381: Do you think these should be simp lemmas?

Mathlib/AlgebraicGeometry/ResidueField.lean
```diff
@@ -0,0 +1,119 @@
+/-
+Copyright (c) 2024 Christian Merten. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Merten
+-/
+import Mathlib.AlgebraicGeometry.OpenImmersion
```
erdOne, line 6: Why is this necessary? I think everything `LocallyRingedSpace`-related should be in `Geometry/RingedSpace/OpenImmersion`.

Mathlib/AlgebraicGeometry/ResidueField.lean
```diff
@@ -0,0 +1,119 @@
...
+import Mathlib.AlgebraicGeometry.GammaSpecAdjunction
```
erdOne, line 7: We should split this file and move the `LocallyRingedSpace` adjunction away... (not in this PR)

Mathlib/AlgebraicGeometry/GammaSpecAdjunction.lean
```diff
@@ -69,6 +69,18 @@ def ΓToStalk (x : X) : Γ.obj (op X) ⟶ X.presheaf.stalk x :=
   X.presheaf.germ (⟨x, trivial⟩ : (⊤ : Opens X))
 #align algebraic_geometry.LocallyRingedSpace.Γ_to_stalk AlgebraicGeometry.LocallyRingedSpace.ΓToStalk
```
erdOne, line 70: Or the real problem is that this def doesn't belong here.

erdOne (commented):
Please move the file to `Geometry/RingedSpace/LocallyRingedSpace`.

Mathlib/RingTheory/Ideal/LocalRing.lean
chrisflav: Probably makes sense.

chrisflav:
Thanks for the quick review @erdOne! An unfortunate side effect of reducing the imports, is that the nice notations you introduced are no longer available. I think we should move these down the import hierarchy.

Mathlib/Geometry/RingedSpace/LocallyRingedSpace/ResidueField.lean
```diff
@@ -0,0 +1,122 @@
...
+    X.Γevaluation x f = 0 ↔
+      (X.presheaf.germ ⟨x, show x ∈ ⊤ by trivial⟩) f ∈ LocalRing.maximalIdeal (X.stalk x) :=
+  LocalRing.residue_eq_zero_iff _
+
+lemma Γevaluation_ne_zero_iff_isUnit (x : U) (f : X.presheaf.obj (op U)) :
+    X.evaluation x f ≠ 0 ↔ IsUnit ((X.presheaf.germ x) f) :=
+  LocalRing.residue_ne_zero_iff_isUnit _
+
+@[simp]
+lemma evaluation_ne_zero_iff_mem_basicOpen (f : X.presheaf.obj (op U)) (x : U) :
```
erdOne, line 79: maybe `evaluation_eq_zero_iff_not_mem_basicOpen` as simp instead.

chrisflav: Instead or additionally?

erdOne: I thought lint was complaining that the LHS is not simpNF as `ne_eq` rewrites it to `\not _ = _`

chrisflav: No, it's only complaining for the `Γ` version, because it rewrites with `Γ_obj`.

erdOne: Ah okay. `evaluation_eq_zero_iff_not_mem_basicOpen` can still be added as simp lemma though. And then this lemma would be provable by simp.

chrisflav: Should I still keep the `evaluation_ne_zero_iff_mem_basicOpen` version?

erdOne: I kind of believe that `evaluation_eq_zero_iff_not_mem_basicOpen` and `evaluation_ne_zero_iff_mem_basicOpen` (and their `Γ`-counterparts) are the only `evaluation_*_zero_iff_*` lemmas worth keeping here, since `mem_basicOpen` should currently be our simpNF of this concept.

erdOne: Not 100% sure though. Will need to use this API more to find out, but this is my gut feeling.

chrisflav: I _thought_ I did use the `mem_maximalIdeal` versions somewhere in my application, but turns out I did not. So I am happy to remove them.

erdOne: Even if you did, I believe you should do that in two rewrites (or `A.trans B.symm`). It is more manageable for both mathlib and users if there is one canonical way to spell everything. This leads to n rewrite lemmas for n equivalent spellings instead of n^2, and the user will then always be encouraged to change their arguments into the canonical form, reliably knowing that this where all the API is.

chrisflav: Understood.

Mathlib/Geometry/RingedSpace/LocallyRingedSpace/ResidueField.lean
```diff
@@ -0,0 +1,113 @@
...
+
+If we interpret sections over `U` as functions of `X` defined on `U`, then this ring map
+corresponds to evaluation at `x`.
+-/
+def evaluation (x : U) :
+    X.presheaf.obj (op U) ⟶ X.residueField x :=
+  X.presheaf.germ x ≫ LocalRing.residue _
+
+/-- The global evaluation map from `Γ(X, ⊤)` to the residue field at `x`. -/
+def Γevaluation (x : X) : LocallyRingedSpace.Γ.obj (op X) ⟶ X.residueField x :=
```
erdOne, line 58:
```suggestion
def Γevaluation (x : X) : X.presheaf.obj (op ⊤) ⟶ X.residueField x :=
```
Does this make `Γevaluation_eq_zero_iff_not_mem_basicOpen` a valid simp lemma?

chrisflav: Indeed, good catch. Changed.

Mathlib/Geometry/RingedSpace/LocallyRingedSpace/ResidueField.lean
```diff
@@ -0,0 +1,113 @@
...
+corresponds to evaluation at `x`.
+-/
+def evaluation (x : U) : X.presheaf.obj (op U) ⟶ X.residueField x :=
+  X.presheaf.germ x ≫ LocalRing.residue _
+
+/-- The global evaluation map from `Γ(X, ⊤)` to the residue field at `x`. -/
+def Γevaluation (x : X) : X.presheaf.obj (op ⊤) ⟶ X.residueField x :=
+  X.evaluation ⟨x, show x ∈ ⊤ from trivial⟩
+
+@[simp]
```
joelriou, line 60: Do we really want this to be a simp lemma? (same for `Γevaluation_eq_zero_iff_not_mem_basicOpen`.)

erdOne: I think we should choose a simpNF. Do you think `X.evaluation x f = 0` would be better?

Mathlib/Geometry/RingedSpace/LocallyRingedSpace/ResidueField.lean
```diff
@@ -0,0 +1,113 @@
...
+
+lemma Γevaluation_ne_zero_iff_mem_basicOpen (x : X) (f : X.presheaf.obj (op ⊤)) :
+    X.Γevaluation x f ≠ 0 ↔ x ∈ X.toRingedSpace.basicOpen f :=
+  evaluation_ne_zero_iff_mem_basicOpen X ⟨x, show x ∈ ⊤ by trivial⟩ f
+
+variable {X Y : LocallyRingedSpace.{u}} (f : X ⟶ Y)
+
+/-- If `X ⟶ Y` is a morphism of locally ringed spaces and `x` a point of `X`, we obtain
+a morphism of residue fields in the other direction. -/
+def evaluationMap (x : X) : Y.residueField (f.val.base x) ⟶ X.residueField x :=
```
joelriou, line 83: Could you also include a lemma which relates `stalkMap f x` and `evaluationMap f x`, and lemmas `evaluationMap_id` and `evaluationMap_comp`?

joelriou: (Also, I am not sure about the name "evaluationMap"; do not you feel that `mapResidueField` would be better?)

chrisflav: `residueFieldMap` (in accordance with `stalkMap`) maybe?

Mathlib/Geometry/RingedSpace/LocallyRingedSpace/ResidueField.lean
joelriou: Ok, let us keep these simp lemmas.

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

chrisflav:
Thanks for the reviews!

PR #14304, opened 2024-06-30
[Merged by Bors] - chore(Order): remove almost all autoImplicit

grunweg:
In two files, only reduce the scope of it;
the other occurrences seem harder to remove.

---

### DIFF
--- a/Mathlib/Order/Atoms.lean
+++ b/Mathlib/Order/Atoms.lean
@@ -49,7 +49,7 @@ which are lattices with only two elements, and related ideas.
 
 -/
 
-variable {α β : Type*}
+variable {ι : Sort*} {α β : Type*}
 
 section Atoms
 
@@ -111,7 +111,7 @@ alias ⟨CovBy.is_atom, IsAtom.bot_covBy⟩ := bot_covBy_iff
 
 end PartialOrder
 
-theorem atom_le_iSup {ι : Sort*} {a : α} [Order.Frame α] (ha : IsAtom a) {f : ι → α} :
+theorem atom_le_iSup [Order.Frame α] {a : α} (ha : IsAtom a) {f : ι → α} :
     a ≤ iSup f ↔ ∃ i, a ≤ f i := by
   refine ⟨?_, fun ⟨i, hi⟩ => le_trans hi (le_iSup _ _)⟩
   show (a ≤ ⨆ i, f i) → _
@@ -201,7 +201,7 @@ alias ⟨CovBy.isCoatom, IsCoatom.covBy_top⟩ := covBy_top_iff
 
 end PartialOrder
 
-theorem iInf_le_coatom  {a : α} {ι : Sort*} [Order.Coframe α] (ha : IsCoatom a) {f : ι → α} :
+theorem iInf_le_coatom [Order.Coframe α] {a : α} (ha : IsCoatom a) {f : ι → α} :
     iInf f ≤ a ↔ ∃ i, f i ≤ a :=
   atom_le_iSup (α := αᵒᵈ) ha
 
--- a/Mathlib/Order/Atoms.lean
+++ b/Mathlib/Order/Atoms.lean
@@ -49,9 +49,6 @@ which are lattices with only two elements, and related ideas.
 
 -/
 
-set_option autoImplicit true
-
-
 variable {α β : Type*}
 
 section Atoms
@@ -114,7 +111,7 @@ alias ⟨CovBy.is_atom, IsAtom.bot_covBy⟩ := bot_covBy_iff
 
 end PartialOrder
 
-theorem atom_le_iSup [Order.Frame α] (ha : IsAtom a) {f : ι → α} :
+theorem atom_le_iSup {ι : Sort*} {a : α} [Order.Frame α] (ha : IsAtom a) {f : ι → α} :
     a ≤ iSup f ↔ ∃ i, a ≤ f i := by
   refine ⟨?_, fun ⟨i, hi⟩ => le_trans hi (le_iSup _ _)⟩
   show (a ≤ ⨆ i, f i) → _
@@ -204,7 +201,7 @@ alias ⟨CovBy.isCoatom, IsCoatom.covBy_top⟩ := covBy_top_iff
 
 end PartialOrder
 
-theorem iInf_le_coatom [Order.Coframe α] (ha : IsCoatom a) {f : ι → α} :
+theorem iInf_le_coatom  {a : α} {ι : Sort*} [Order.Coframe α] (ha : IsCoatom a) {f : ι → α} :
     iInf f ≤ a ↔ ∃ i, f i ≤ a :=
   atom_le_iSup (α := αᵒᵈ) ha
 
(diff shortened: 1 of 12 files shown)


Mathlib/Order/Atoms.lean
```diff
@@ -49,9 +49,6 @@ which are lattices with only two elements, and related ideas.
 
 -/
 
-set_option autoImplicit true
-
-
 variable {α β : Type*}
```
eric-wieser, line 52:
```suggestion
variable {ι : Sort*} {α β : Type*}
```

Mathlib/Order/Filter/AtTopBot.lean
```diff
@@ -1545,7 +1543,7 @@ theorem prod_atTop_atTop_eq [Preorder α] [Preorder β] :
 #align filter.prod_at_top_at_top_eq Filter.prod_atTop_atTop_eq
 
 -- Porting note: generalized from `SemilatticeSup` to `Preorder`
-theorem prod_atBot_atBot_eq [Preorder β₁] [Preorder β₂] :
+theorem prod_atBot_atBot_eq {β₁ β₂ : Type*} [Preorder β₁] [Preorder β₂] :
```
eric-wieser, line 1546: Let's just use `α` and `β` here to match the rest of the section?

Mathlib/Order/Filter/Basic.lean
```diff
@@ -3317,24 +3320,28 @@ theorem Set.MapsTo.tendsto {α β} {s : Set α} {t : Set β} {f : α → β} (h
...
 
+set_option autoImplicit true in
 theorem Filter.map_mapsTo_Iic_iff_mapsTo {m : α → β} :
     MapsTo (map m) (Iic <| 𝓟 s) (Iic <| 𝓟 t) ↔ MapsTo m s t := by
   rw [map_mapsTo_Iic_iff_tendsto, tendsto_principal_principal, MapsTo]
 
 alias ⟨_, Set.MapsTo.filter_map_Iic⟩ := Filter.map_mapsTo_Iic_iff_mapsTo
 
 -- TODO(Anatole): unify with the global case
+set_option autoImplicit true
```
eric-wieser, line 3345:
```suggestion
```

Mathlib/Order/Atoms.lean
```diff
@@ -204,7 +201,7 @@ alias ⟨CovBy.isCoatom, IsCoatom.covBy_top⟩ := covBy_top_iff
 
 end PartialOrder
 
-theorem iInf_le_coatom [Order.Coframe α] (ha : IsCoatom a) {f : ι → α} :
+theorem iInf_le_coatom  {a : α} {ι : Sort*} [Order.Coframe α] (ha : IsCoatom a) {f : ι → α} :
```
eric-wieser, line 204: Generally data comes after typeclasses:
```suggestion
theorem iInf_le_coatom [Order.Coframe α] {a : α} (ha : IsCoatom a) {f : ι → α} :
```

Mathlib/Order/Minimal.lean
```diff
@@ -315,7 +313,7 @@ theorem IsAntichain.maximals_lowerClosure (hs : IsAntichain (· ≤ ·) s) :
 
 section Image
 
-variable {f : α → β} {r : α → α → Prop} {s : β → β → Prop}
+variable {β : Type*} {f : α → β} {r : α → α → Prop} {s : β → β → Prop}
```
eric-wieser, line 316: Maybe hoist `beta` to the top of the file, if that works?

Mathlib/Order/RelIso/Basic.lean
```diff
@@ -265,8 +264,9 @@ theorem map_rel_iff (f : r ↪r s) {a b} : s (f a) (f b) ↔ r a b :=
   f.map_rel_iff'
 #align rel_embedding.map_rel_iff RelEmbedding.map_rel_iff
 
+set_option autoImplicit true in
 @[simp]
-theorem coe_mk : ⇑(⟨f, h⟩ : r ↪r s) = f :=
+theorem coe_mk {h : ∀ {a b : α}, s (f a) (f b) ↔ r a b} : ⇑(⟨f, h⟩ : r ↪r s) = f :=
```
eric-wieser, line 269:
```suggestion
theorem coe_mk {h} : ⇑(⟨f, h⟩ : r ↪r s) = f :=
```
is a good option here to avoid having to write anything verbose.

Mathlib/Order/Atoms.lean
grunweg: Thanks, TIL!

Mathlib/Order/Filter/AtTopBot.lean
grunweg: Done; and also re-used alpha and beta in a few more places.

Mathlib/Order/Minimal.lean
grunweg: It does; done!

Mathlib/Order/Filter/Basic.lean
```diff
@@ -3317,25 +3320,28 @@ theorem Set.MapsTo.tendsto {α β} {s : Set α} {t : Set β} {f : α → β} (h
...
+set_option autoImplicit true in
 theorem Filter.map_mapsTo_Iic_iff_mapsTo {m : α → β} :
     MapsTo (map m) (Iic <| 𝓟 s) (Iic <| 𝓟 t) ↔ MapsTo m s t := by
   rw [map_mapsTo_Iic_iff_tendsto, tendsto_principal_principal, MapsTo]
 
 alias ⟨_, Set.MapsTo.filter_map_Iic⟩ := Filter.map_mapsTo_Iic_iff_mapsTo
 
 -- TODO(Anatole): unify with the global case
-
+set_option autoImplicit true
```
sgouezel, line 3344:
```suggestion
set_option autoImplicit true in
```

grunweg: That option was meant to be global for the remaining file - but indeed, it can be removed as well; I just pushed a commit doing so.

Mathlib/Order/Filter/Basic.lean
```diff
@@ -1734,6 +1733,8 @@ instance : Trans ((· ≤ᶠ[l] ·) : (α → β) → (α → β) → Prop) (·
 
 end Preorder
 
+variable {l : Filter α}
```
riccardobrasca, line 1736: Maybe you add here `s` and `t`?

Mathlib/Order/Minimal.lean
```diff
@@ -91,30 +91,32 @@ theorem eq_of_mem_minimals (ha : a ∈ minimals r s) (hb : b ∈ s) (h : r b a)
   antisymm (ha.2 hb h) h
 #align eq_of_mem_minimals eq_of_mem_minimals
 
-set_option autoImplicit true
-
-theorem mem_maximals_iff : x ∈ maximals r s ↔ x ∈ s ∧ ∀ ⦃y⦄, y ∈ s → r x y → x = y := by
+theorem mem_maximals_iff {x : α} : x ∈ maximals r s ↔ x ∈ s ∧ ∀ ⦃y⦄, y ∈ s → r x y → x = y := by
```
riccardobrasca, line 94: I think `variable {x : \a}` is a good idea.

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Order/Filter/Basic.lean
grunweg: I've checked: the file uses s and t for various things; there's no consistent pattern.
I think it's more readable as-is. (Refactoring that file is a good follow-up idea, though I'd probably start by splitting the file...)

grunweg:
Thanks for the reviews, everybody - these were really helpful comments!
bors r+

PR #14314, opened 2024-07-01
[Merged by Bors] - chore: delete obsolete scripts

kim-em:
I appreciate this is somewhat scary, so happy to have careful review here (or to entertain other suggestions about how to proceed).

After this, I would like to institute a policy that every script in `scripts/` must have a description in `scripts/README.md`.


grunweg (commented):
Thanks for doing this; I agree this policy would be great to have. (In fact, mucking out that folder was on my list of things as well - I'm happy to see you got to it earlier!)

Let me go through the first scripts individually.
For one group, I immediately agree that those can be removed. All of them are specific to mathlib and are obsolete now that the port is done. Feel free to cherry-pick those in a separate PR and maintainer merge that on my behalf already.
- align_import generates the `align_import` commands: there is no need to re-generate these as the port is now done -> this seems safe to remove
- align.py: similar (though I'm not 100% sure what this tool does)
- `make_port_status` regenerates mathlib's port-status page: there's no need for that any more; `run_port_status` is similar
- `start_port` is also specific to mathlib -> seems safe to remove

I'd like to talk about fix-lints, fix-line-breaks and fix-comments more carefully.
- The functionality of `fix-line-breaks` is included in `lint_style.py` now. ~~ --- but the calling script is not adapted, AFAICT.~~ Edited later: this script was only called in `start-port.sh`.
- I am not sure about `fix_lints`. Would or could this be useful to somebody porting their Lean3 project? These still happen occasionally. (If the latter, this script should at least be documented appropriate in the wiki - so people looking for it can still find this without PR archeology.)
- Please don't remove `fix-comments.py` just yet (or let's have a longer discussion first); I think mathlib still has a about a hundred comments referring to mathlib3 names. (I worked on this in the spring, not all of this has landed yet; some was fixed in the mean-time. I partially reinvented that script, before realising it existed...; I would certainly prefer to run it on all of mathlib once to see if it finds anything.

I have not thought about check_silent.sh or benchmark.sh yet. The former's check does run in CI right now, correct? How is it implemented now? (If that makes the script obsolete, that's fine. Edited later: yes, both are obsolete, see @semorrison's comments.)

semorrison:
> * The functionality of `fix-line-breaks` is included in `lint_style.py` now --- but the calling script is not adapted, AFAICT.

`fix-line-breaks` is only called by `start_port.sh`, which is now obsolete. Do you agree?

grunweg:
> > * The functionality of `fix-line-breaks` is included in `lint_style.py` now --- but the calling script is not adapted, AFAICT.
> 
> `fix-line-breaks` is only called by `start_port.sh`, which is now obsolete. Do you agree?

Thanks for checking - I agree. You can fold that into #14346 as well.

Ruben-VandeVelde:
I believe I've run `scripts/fix-comments.py` not *that* long ago on a file where we'd forgotten it during the port, so I might keep that one.

semorrison:
Okay, I've restored `fix-comments.py`. I think the others can still go?

grunweg:
Yes, looks good to me. (I'll try to run `fix-comments.py` on all of mathlib in the coming weeks, and will propose deleting it afterwards.)
maintainer merge

jcommelin:
Thanks :tada:

bors merge

PR #14128, opened 2024-06-25
[Merged by Bors] - feat(Topology/Instances/EReal): lemmas related to liminf and limsup

pitmonticone:
Changes: 
- [x] Add `liminf_le_liminf`
- [x] Add `limsup_le_limsup`
- [x] Add `limsup_add_le_lt₂`
- [x] Add `limsup_add_bot_ne_top`
- [x] Add `limsup_add_le_add_limsup`
- [x] Add `ge_iff_le_forall_real_lt`
- [x] Add `liminf_add_ge_gt₂`
- [x] Add `liminf_add_top_ne_bot`
- [x] Add `add_liminf_le_liminf_add`
- [x] Add `limsup_le_iff`
- [x] Add `limsup_le_const_forall`
- [x] Add `const_le_limsup_forall`
- [x] Add `liminf_le_const_forall`
- [x] Add `const_le_liminf_forall`
- [x] Add `limsup_max`

Co-authored-by: @D-Thomine 

---

- [x] depends on: #14125

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14324, opened 2024-07-01
[Merged by Bors] - chore(Data/Real/EReal): `theorem`s to `lemma`s

pitmonticone:
This PR turns some theorems into lemmas as suggested [here](https://github.com/leanprover-community/mathlib4/pull/14128#discussion_r1660572911).

---


sgouezel:
bors r+
Thanks!
But no need to do that in other files: when we want to enforce this, we will write a program to turn all the theorems without a docstring into lemmas. In the meantime, just make sure that new statements you introduce respect the rule.

PR #14327, opened 2024-07-01
[Merged by Bors] - feat(Data.ZMod): More lemmas about `ZMod.val`

javra:
---

### DIFF
--- a/Mathlib/Data/Nat/Defs.lean
+++ b/Mathlib/Data/Nat/Defs.lean
@@ -1070,7 +1070,7 @@ lemma sub_mod_eq_zero_of_mod_eq (h : m % k = n % k) : (m - n) % k = 0 := by
 @[simp] lemma one_mod (n : ℕ) : 1 % (n + 2) = 1 :=
   Nat.mod_eq_of_lt (Nat.add_lt_add_right n.succ_pos 1)
 
-lemma ne_one_iff_one_mod : ∀ {n : ℕ}, n ≠ 1 ↔ 1 % n = 1
+lemma one_mod_eq_one : ∀ {n : ℕ}, 1 % n = 1 ↔ n ≠ 1
   | 0 | 1 | n + 2 => by simp
 
 lemma dvd_sub_mod (k : ℕ) : n ∣ k - k % n :=
--- a/Mathlib/Data/Nat/Defs.lean
+++ b/Mathlib/Data/Nat/Defs.lean
@@ -1070,8 +1070,8 @@ lemma sub_mod_eq_zero_of_mod_eq (h : m % k = n % k) : (m - n) % k = 0 := by
 @[simp] lemma one_mod (n : ℕ) : 1 % (n + 2) = 1 :=
   Nat.mod_eq_of_lt (Nat.add_lt_add_right n.succ_pos 1)
 
-lemma one_mod_of_ne_one : ∀ {n : ℕ}, n ≠ 1 → 1 % n = 1
-  | 0, _ | (n + 2), _ => by simp
+lemma one_mod_iff_ne_one : ∀ {n : ℕ}, n ≠ 1 ↔ 1 % n = 1
+  | 0 | 1 | n + 2 => by simp
 
 lemma dvd_sub_mod (k : ℕ) : n ∣ k - k % n :=
   ⟨k / n, Nat.sub_eq_of_eq_add (Nat.div_add_mod k n).symm⟩
--- a/Mathlib/Data/Nat/Defs.lean
+++ b/Mathlib/Data/Nat/Defs.lean
@@ -1070,7 +1070,7 @@ lemma sub_mod_eq_zero_of_mod_eq (h : m % k = n % k) : (m - n) % k = 0 := by
 @[simp] lemma one_mod (n : ℕ) : 1 % (n + 2) = 1 :=
   Nat.mod_eq_of_lt (Nat.add_lt_add_right n.succ_pos 1)
 
-lemma one_mod_iff_ne_one : ∀ {n : ℕ}, n ≠ 1 ↔ 1 % n = 1
+lemma ne_one_iff_one_mod : ∀ {n : ℕ}, n ≠ 1 ↔ 1 % n = 1
   | 0 | 1 | n + 2 => by simp
 
 lemma dvd_sub_mod (k : ℕ) : n ∣ k - k % n :=
--- a/Mathlib/Data/Nat/Defs.lean
+++ b/Mathlib/Data/Nat/Defs.lean
@@ -1073,6 +1073,9 @@ lemma sub_mod_eq_zero_of_mod_eq (h : m % k = n % k) : (m - n) % k = 0 := by
 lemma one_mod_eq_one : ∀ {n : ℕ}, 1 % n = 1 ↔ n ≠ 1
   | 0 | 1 | n + 2 => by simp
 
+@[deprecated (since := "2024-08-28")]
+lemma one_mod_of_ne_one  : ∀ {n : ℕ}, n ≠ 1 → 1 % n = 1 := one_mod_eq_one.mpr
+
 lemma dvd_sub_mod (k : ℕ) : n ∣ k - k % n :=
   ⟨k / n, Nat.sub_eq_of_eq_add (Nat.div_add_mod k n).symm⟩
 
--- a/Mathlib/Data/ZMod/Basic.lean
+++ b/Mathlib/Data/ZMod/Basic.lean
@@ -990,8 +990,6 @@ theorem val_ne_zero {n : ℕ} (a : ZMod n) : a.val ≠ 0 ↔ a ≠ 0 :=
 theorem val_pos_of_ne_zero {n : ℕ} {a : ZMod n} (h : a ≠ 0) : 0 < a.val :=
   Nat.pos_of_ne_zero <| (val_ne_zero a).mpr h
 
-
-
 theorem val_eq_one : ∀ {n : ℕ} [n.AtLeastTwo] (a : ZMod n), a.val = 1 ↔ a = 1
   | 0, ⟨hn⟩, _
   | 1, ⟨hn⟩, _ => by simp at hn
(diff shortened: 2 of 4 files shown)


javra:
Linter now gives me a warning to split `Data.ZMod.Basic`, we should probably do that soon. Also there are some instances of `[Fact (1 < n)]`, which maybe should be replaced by `[n.AtLeastTwo]`?

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -981,6 +987,20 @@ theorem val_eq_zero : ∀ {n : ℕ} (a : ZMod n), a.val = 0 ↔ a = 0
 theorem val_ne_zero {n : ℕ} (a : ZMod n) : a.val ≠ 0 ↔ a ≠ 0 :=
   (val_eq_zero a).not
 
+theorem val_pos_of_ne_zero {n : ℕ} {a : ZMod n} (h : a ≠ 0) : 0 < a.val := by
+  apply Nat.pos_of_ne_zero
+  intro h
+  rw [ZMod.val_eq_zero] at h
+  contradiction
```
Ruben-VandeVelde, line 994: This can be `Nat.pos_of_ne_zero <| (val_ne_zero a).mpr h`

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -981,6 +987,20 @@ theorem val_eq_zero : ∀ {n : ℕ} (a : ZMod n), a.val = 0 ↔ a = 0
...
+
+theorem val_eq_one : ∀ {n : ℕ} [n.AtLeastTwo] (a : ZMod n), a.val = 1 ↔ a = 1
+  | 0, ⟨hn⟩, _
+  | 1, ⟨hn⟩, _ => by simp at hn
+  | n + 2, _, _ => by
+    rw [Fin.ext_iff]
+    simp only [Fin.val_one]
+    exact Iff.rfl
```
Ruben-VandeVelde, line 1002: I'm surprised simp doesn't cover this. Are you forcing through a defeq issue?

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1011,6 +1031,17 @@ theorem neg_eq_self_iff {n : ℕ} (a : ZMod n) : -a = a ↔ a = 0 ∨ 2 * a.val
 theorem val_cast_of_lt {n : ℕ} {a : ℕ} (h : a < n) : (a : ZMod n).val = a := by
   rw [val_natCast, Nat.mod_eq_of_lt h]
 
+theorem val_cast_ZMod_lt {m : ℕ} [NeZero m] (n : ℕ) [NeZero n] (a : ZMod m) :
```
Ruben-VandeVelde, line 1034: Should ZMod not become zmod in names (should it be mentioned at all?)

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1011,6 +1031,17 @@ theorem neg_eq_self_iff {n : ℕ} (a : ZMod n) : -a = a ↔ a = 0 ∨ 2 * a.val
...
+    (a.cast : ZMod n).val < m := by
+  rcases m with (⟨⟩|⟨m⟩); · cases NeZero.ne 0 rfl
+  by_cases h : m < n
+  · rcases n with (⟨⟩|⟨n⟩); · simp at h
+    erw [val_cast_of_lt]
```
Ruben-VandeVelde, line 1039: Why do you need erw? That's a code smell

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1056,6 +1087,29 @@ theorem cast_cast_zmod_of_le {m n : ℕ} [hm : NeZero m] (h : m ≤ n) (a : ZMod
...
   rw [cast_eq_val, val_cast_eq_val_of_lt (a.val_lt.trans_le h), natCast_zmod_val]
 
+theorem val_pow {m n : ℕ} {a : ZMod n} [ilt : Fact (1 < n)] (h : a.val ^ m < n) :
+    (a ^ m).val = a.val ^ m := by
+  induction m with
+  | zero => simp [ZMod.val_one]
+  | succ m ih =>
+    have : a.val ^ m < n := by
+      by_cases ha : a = 0
+      · cases ha
```
Ruben-VandeVelde, line 1097: Are those `cases` just a less clear way of writing `subst`? If so, please use that

Mathlib/Data/ZMod/Basic.lean
javra: Not sure, the naming convention states
> When something named with `UpperCamelCase` is part of something named with `snake_case`, it is referenced in `lowerCamelCase`.

Mathlib/Data/ZMod/Basic.lean
javra: Yes, `Fin.ext_iff` doesnt get applied by `simp` because the variables are still in `ZMod (n + 2)` instead of `Fin (n + 2)`. Likewise, the `Iff.rfl` needs to be forced because it's `Fin.val` on one side and `ZMod.val` on the other.

Mathlib/Data/ZMod/Basic.lean
javra: ... but then, I don't know if `ZMod` really counts as camel case, and it's `..._zmod` everywhere else, so I'll adapt.

Mathlib/Data/ZMod/Basic.lean
javra: I don't think so, I'm pretty sure we do treat both cases separately.

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1056,6 +1081,29 @@ theorem cast_cast_zmod_of_le {m n : ℕ} [hm : NeZero m] (h : m ≤ n) (a : ZMod
...
```
YaelDillies, line 1091:
```suggestion
      obtain rfl | ha := eq_or_ne a 0
```

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -981,6 +987,14 @@ theorem val_eq_zero : ∀ {n : ℕ} (a : ZMod n), a.val = 0 ↔ a = 0
 theorem val_ne_zero {n : ℕ} (a : ZMod n) : a.val ≠ 0 ↔ a ≠ 0 :=
   (val_eq_zero a).not
 
+theorem val_pos_of_ne_zero {n : ℕ} {a : ZMod n} (h : a ≠ 0) : 0 < a.val :=
+  Nat.pos_of_ne_zero <| (val_ne_zero a).mpr h
+
+theorem val_eq_one : ∀ {n : ℕ} [n.AtLeastTwo] (a : ZMod n), a.val = 1 ↔ a = 1
```
YaelDillies, line 993:
```suggestion
theorem val_eq_one : ∀ {n : ℕ} (hn : n ≠ 1) (a : ZMod n), a.val = 1 ↔ a = 1
```
Use `ZMod.val_one` and `Nat.one_mod_of_ne_one` (which you should turn into an `Iff`)

javra: The statement is not true for `n = 0` because `ZMod.val` then is the absolut value.

Mathlib/Data/Nat/Defs.lean
```diff
@@ -1070,8 +1070,8 @@ lemma sub_mod_eq_zero_of_mod_eq (h : m % k = n % k) : (m - n) % k = 0 := by
 @[simp] lemma one_mod (n : ℕ) : 1 % (n + 2) = 1 :=
   Nat.mod_eq_of_lt (Nat.add_lt_add_right n.succ_pos 1)
 
-lemma one_mod_of_ne_one : ∀ {n : ℕ}, n ≠ 1 → 1 % n = 1
-  | 0, _ | (n + 2), _ => by simp
+lemma ne_one_iff_one_mod : ∀ {n : ℕ}, n ≠ 1 ↔ 1 % n = 1
```
YaelDillies, line 1073:
```suggestion
lemma one_mod_eq_one : ∀ {n : ℕ}, 1 % n = 1 ↔ n ≠ 1
```

Mathlib/Data/ZMod/Basic.lean
YaelDillies: Then
```suggestion
theorem val_eq_one : ∀ {n : ℕ} (hn : 1 < n) (a : ZMod n), a.val = 1 ↔ a = 1
```
`Nat.AtLeastTwo` should only be used for `OfNat.ofNat`

Ruben-VandeVelde (approved):
maintainer merge

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -981,6 +987,14 @@ theorem val_eq_zero : ∀ {n : ℕ} (a : ZMod n), a.val = 0 ↔ a = 0
 theorem val_ne_zero {n : ℕ} (a : ZMod n) : a.val ≠ 0 ↔ a ≠ 0 :=
   (val_eq_zero a).not
 
+theorem val_pos_of_ne_zero {n : ℕ} {a : ZMod n} (h : a ≠ 0) : 0 < a.val :=
+  Nat.pos_of_ne_zero <| (val_ne_zero a).mpr h
```
YaelDillies, line 991: Surely
```suggestion
theorem val_pos {n : ℕ} {a : ZMod n} : 0 < a.val ↔ 0 < a:= by simpa [pos_iff_ne_zero] using val_ne_zero _
```

javra: To my knowledge, there's no `LT` instance on `ZMod n` and I think that's a conscious choice

YaelDillies: Then
```suggestion
theorem val_pos {n : ℕ} {a : ZMod n} : 0 < a.val ↔ a ≠ 0 := by simpa [pos_iff_ne_zero] using val_ne_zero _
```

YaelDillies:
maintainer merge

mattrobball:
Should we be deprecated instead of deleting or did I miss something?

javra:
> Should we be deprecated instead of deleting or did I miss something?

The `Nat.one_mod_eq_one` lemma? Or `ZMod.val_pos_of_ne_zero`?

mattrobball:
`Nat.one_mod_of_ne_one` was deleted. Can you add it back as an alias and deprecate it?

javra:
Can't add it as an alias because only constants can be `alias`ed, but I can  re-add it as a separate lemma and deprecate it.

mattrobball:
Thanks. 

bors merge

PR #14332, opened 2024-07-01
mrb/unbundled alg from ord plus lean 2905

mattrobball:
- **first batch**
- **next batch**
- **next batch**
- **another batch**
- **another batch**
- **another batch**
- **another batch**
- **another batch**
- **more**
- **more**
- **final**
- **lint**
- **changes**


mattrobball:
!bench

PR #14335, opened 2024-07-01
[Merged by Bors] - chore (Module.LinearMap.End): avoid ordered algebra imports

mattrobball:
We can just specialize to the results about `Nat` and avoid importing `Algebra.Order.Ring.Nat`.

---


semorrison:
bors merge

PR #14340, opened 2024-07-01
[Merged by Bors] - feat: API for multiequalizers in the category of types

joelriou:
In this PR, we give a concrete characterization of limit multiforks in the category of types.

---

This is used in #14208


Mathlib/CategoryTheory/Limits/Shapes/Types.lean
```diff
@@ -929,4 +931,87 @@ lemma pushoutCocone_inl_eq_inr_iff_of_isColimit {c : PushoutCocone f g} (hc : Is
...
+        | .left i => s.val i
+        | .right j => I.fst j (s.val _)
+      property := by
+        rintro _ _ (_|_|r)
+        · rfl
+        · rfl
+        · exact (s.property r).symm }
+  invFun s :=
+    { val := fun i ↦ s.val (.left i)
+      property := fun r ↦ (s.property (.fst r)).trans (s.property (.snd r)).symm }
```
dagurtomas, line 965:
```suggestion
  toFun s :=
  { val := fun i ↦ match i with
      | .left i => s.val i
      | .right j => I.fst j (s.val _)
    property := by
      rintro _ _ (_|_|r)
      · rfl
      · rfl
      · exact (s.property r).symm }
  invFun s :=
  { val := fun i ↦ s.val (.left i)
    property := fun r ↦ (s.property (.fst r)).trans (s.property (.snd r)).symm }
```

dagurtomas: I don't know if we have a rule for this, feel free to ignore. I just think it looks slightly better this way.

dagurtomas (approved):
LGTM

Mathlib/CategoryTheory/Limits/Shapes/Types.lean
joelriou: Thanks for the reviews! but I do not think that we have any precise rule for `{...}`!

Mathlib/CategoryTheory/Limits/Shapes/Types.lean
```diff
@@ -929,4 +931,87 @@ lemma pushoutCocone_inl_eq_inr_iff_of_isColimit {c : PushoutCocone f g} (hc : Is
...
-end CategoryTheory.Limits.Types
+end Types
+
+section Multiequalizer
+
+variable (I : MulticospanIndex (Type u))
+
+/-- Given a multiequalizer diagram `I : MulticospanIndex (Type u)` in the category
+of types, this is a type which identifies to the sections of the
+functor `I.multicospan.sections`. -/
```
erdOne, line 942:
```suggestion
functor `I.multicospan`. -/
```
Also I'd call `I.multicospan` the diagram and `I : MulticospanIndex (Type u)` the index of that diagram.

joelriou: Thanks! I have changed the phrasing.

erdOne:
Thanks!
maintainer merge

semorrison:
bors merge

PR #14341, opened 2024-07-01
[Merged by Bors] - chore (Data.Rat.Lemmas): avoid importing `OrderedCommMonoid` 

mattrobball:
We can avoid importing `Algebra.Order.Group.Defs` in `Data.Rat.Lemmas` if we use `Nat` versions of some lemmas and slightly modify one proof.

---

We could also exile to `Rat.inv_intCast_den` (and `Rat.inv_natCast_den`) to `Data.Rat.Cast.Lemmas` since it doesn't seem to be used.


kbuzzard:
Thanks a lot!

bors merge

PR #14342, opened 2024-07-01
[Merged by Bors] - feat (RingTheory/Valuation): add class IsValExtension

jjdishere:
Add typeclass  `IsValExtension` and related lemmas.

Co-authored-by: Bichang Lei [leibichang58@ruc.edu.cn](mailto:leibichang58@ruc.edu.cn)

---

- [x] depends on : #13533 


faenuccio:
I understand the idea beneath the `IsValExtension` declaration, but why do you want it to be a class? Which kind of interactions do you expect with type class inference? Why don't you want it to be simply an assumption that you can add?

jjdishere:
> I understand the idea beneath the `IsValExtension` declaration, but why do you want it to be a class? Which kind of interactions do you expect with type class inference? Why don't you want it to be simply an assumption that you can add?

There are two main reasons that I want to make it a class.
1.  When working with concrete examples such as $\mathbb{C}_p/\mathbb{Q}_p$, or an arbitrary finite extension of local fields, the type class inference can infer the instance for these canonical valuations automatically. (This is the main reason.)
2. It expresses that vL is an extension of vK, so it is the setup for many definitions/theorems. For example, this can be the setup for higher ramification groups (the restriction of vL hardly equals vK due to normalization). Another example is that this can be used in the setup for Krasner's lemma, e.g. for $K^{alg}/K$ when $K$ is any rank one valued field. So making it a class saves explicit variables and enables notations like `G(vL/vK)^[u]`. (This is a relatively minor reason.)

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,164 @@
...
+The class `IsValExtension R A` states that the valuation of `A` is an extension of the valuation
+on `R`. More precisely, the valuation on `R` is equivlent to the comap of the valuation on `A`.
+-/
+class IsValExtension : Prop where
+  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
+  val_isEquiv_comap : vR.IsEquiv <| vA.comap (algebraMap R A)
+
+namespace IsValExtension
+
+section CoeLemma
```
faenuccio, line 64: Why this name of section? There is no `Coe` around... perhaps something with `AlgebraMap` would be more appropriate.

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,164 @@
...
+
+variable {ΓR ΓA ΓK: Type*} [LinearOrderedCommGroupWithZero ΓR] [LinearOrderedCommGroupWithZero ΓK]
+    [LinearOrderedCommGroupWithZero ΓA] {vR : Valuation R ΓR} {vK : Valuation K ΓK}
+    {vA : Valuation A ΓA} [IsValExtension vR vA] [IsValExtension vK vA]
+
+section mk'
+
+/--
+When `K` is a field, if the preimage of the valuation integers of `A` equals to the valuation
+integers of `K`, then the valuation on `A` is an extension of valuation on `K`.
```
faenuccio, line 103:
```suggestion
integers of `K`, then the valuation on `A` is an extension of the valuation on `K`.
```

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,164 @@
...
+    simp only [Algebra.id.map_eq_id, comap_id]
+    rfl
+
+end CoeLemma
+
+variable {ΓR ΓA ΓK: Type*} [LinearOrderedCommGroupWithZero ΓR] [LinearOrderedCommGroupWithZero ΓK]
+    [LinearOrderedCommGroupWithZero ΓA] {vR : Valuation R ΓR} {vK : Valuation K ΓK}
+    {vA : Valuation A ΓA} [IsValExtension vR vA] [IsValExtension vK vA]
+
+section mk'
```
faenuccio, line 99: Not sure this is the best name for the section, but given its length you can skip the sectioning, perhaps. Or did you want to insert the variables inside it?

faenuccio (commented):
Given the examples you have, I agree having a class seems a good idea. Thanks!

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: I'll reorganise the sections. The name `mk'` is really poor.

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,164 @@
...
+* [Bourbaki, Nicolas. *Commutative algebra*] Chapter VI §3, Valuations.
+* <https://en.wikipedia.org/wiki/Valuation_(algebra)#Extension_of_valuations>
+
+## Tags
+Valuation, Extension of Valuations
+
+-/
+open Valuation
+
+variable {R K A ΓR ΓK ΓA : Type*} [CommRing R] [Field K] [Ring A]
```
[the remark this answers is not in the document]
jjdishere: Thank you! Actually, I found that `K` is not used in this part. I'll delete it.

Mathlib/RingTheory/Valuation/ValExtension.lean
jjdishere: Thank you! This is a remaining section marker. When `IsValExt` is defined for `Valued`, these lemmas are coe lemma indeed. But not anymore for `IsValExt vR vA`.

faenuccio:
I would love @mariainesdff and @adamtopaz opinion on this PR.

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,160 @@
...
+  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
+  val_isEquiv_comap : vR.IsEquiv <| vA.comap (algebraMap R A)
+
+namespace IsValExtension
+
+section algebraMap
+
+variable [IsValExtension vR vA]
+
+-- @[simp] does not work because `vR` cannot be inferred from `R`.
```
adamtopaz, line 67: Because of issues like this, I think it would be worthwhile to experiment with making `vR` an `outParam`.

adamtopaz (commented):
Sorry that this PR has been sitting for so long! Overall, I think the design of `IsValExtension` is good. I do think it's worthwhile to experiment making the valuation of the base an `outParam`. Mathematically that roughly corresponds to the intuition that `vR` is "determined" by `vA`. I hope doing this would make some of the automation work better as well.

faenuccio:
> Sorry that this PR has been sitting for so long! Overall, I think the design of `IsValExtension` is good. I do think it's worthwhile to experiment making the valuation of the base an `outParam`. Mathematically that roughly corresponds to the intuition that `vR` is "determined" by `vA`. I hope doing this would make some of the automation work better as well.

Very nice point! @jjdishere , can you give it a try?

Mathlib/RingTheory/Valuation/ValExtension.lean
```diff
@@ -0,0 +1,160 @@
...
+variable {R A ΓR ΓA : Type*} [CommRing R] [Ring A]
+    [LinearOrderedCommMonoidWithZero ΓR] [LinearOrderedCommMonoidWithZero ΓA] [Algebra R A]
+    (vR : Valuation R ΓR) (vA : Valuation A ΓA)
+
+/--
+The class `IsValExtension R A` states that the valuation of `A` is an extension of the valuation
+on `R`. More precisely, the valuation on `R` is equivlent to the comap of the valuation on `A`.
+-/
+class IsValExtension : Prop where
+  /-- The valuation on `R` is equivlent to the comap of the valuation on `A` -/
```
faenuccio, line 58:
```suggestion
  /-- The valuation on `R` is equivalent to the comap of the valuation on `A` -/
```

jjdishere:
> Sorry that this PR has been sitting for so long! Overall, I think the design of `IsValExtension` is good. I do think it's worthwhile to experiment making the valuation of the base an `outParam`. Mathematically that roughly corresponds to the intuition that `vR` is "determined" by `vA`. I hope doing this would make some of the automation work better as well.

Thank you for your advice @adamtopaz . I have discussed with @faenuccio about this possible improvement but we’ve encountered an obstacle. The issue is that `vR` is only determined up to equivalent classes, even by `R`, `A` and `vA` together.

For example, when `R` is $$\mathbb{Q}_p$$, `A` is any finite extension of `R` with the discrete valuation,  we still have many choices of valuations `vR` satisfies `[IsValExtension vR vA]` e.g. The p-adic valuation sending `R` to `ℤₘ₀`, the p-adic valuation to `NNReal`, or the valuation sending $$p$$ to `(.ofAdd -n) : ℤₘ₀` instead of `(.ofAdd -1) : ℤₘ₀`.

So this `vR` and thus the target of the simp lemma cannot be determined until we explicitly fill in what `vR` is. Do you have any suggestions on how to proceed?

faenuccio:
maintainer merge

PR #14339, opened 2024-07-01
[Merged by Bors] - feat: the associator isomorphism for graded objects

joelriou:
In this PR, we construct the associator isomorphism `tensorObj (tensorObj X₁ X₂) X₃ ≅ tensorObj X₁ (tensorObj X₂ X₃)` for graded objects indexed by an additive monoid.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14343, opened 2024-07-01
[Merged by Bors] - feat(CategoryTheory): the triangle identity for the tensor product of graded objects

joelriou:
In this PR, we construct the left and right unitors for the tensor product of graded objects and we obtain the triangle identity.

---

After one more PR about the pentagon relation, we shall get the monoidal structure...

- [x] depends on: #14339


jcommelin:
Thanks :tada:

bors merge

PR #14344, opened 2024-07-01
[Merged by Bors] - chore(CategoryTheory/Limits/Shapes/Pullbacks): split into multiple files and add documentation

callesonne:
In this PR we split the file `CategoryTheory/Limits/Shapes/Pullbacks.lean` into 7 (!) smaller files and add documentation.

This contribution was inspired by the AIM workshop "Formalizing algebraic geometry" in June 2024.

---

This PR does not touch the code itself (except some lemma that I think had the wrong name by mistake, see the PR summary). In an upcoming PR I will golf & apply some fixes that I have noticed whilst going through this file, and in yet another PR I will add some API that I thought was missing whilst writing #14208


erdOne:
This is great!

joelriou:
It looks good to me. I know this can be a penance, but you should run `git blame` (probably more on mathlib3...) so as to identify slightly more precisely the main authors of the different chunks.

callesonne:
> It looks good to me. I know this can be a penance, but you should run `git blame` (probably more on mathlib3...) so as to identify slightly more precisely the main authors of the different chunks.

Yes! I was thinking about asking about this (also I won't add myself as author for most files in the final version of this PR)

callesonne:
> It looks good to me. I know this can be a penance, but you should run `git blame` (probably more on mathlib3...) so as to identify slightly more precisely the main authors of the different chunks.

Okay this gave me almost no information :')

erdOne:
Fantastic! Thanks a lot!
maintainer merge.

PR #14347, opened 2024-07-01
[Merged by Bors] - chore: rm scripts/check_silent.sh

kim-em:
Split out from #14314 as @grunweg wasn't certain of the impact.

We no longer use this script (verified by `grep`), instead relying on `lake`s `--wfail` option.


grunweg:
Makes sense, thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #14348, opened 2024-07-02
perf: disable the `unusedVariables` linter for speed

eric-wieser:
I doubt we want to merge this, but it shaves two minutes of the lint time.

---


eric-wieser:
!bench

PR #14355, opened 2024-07-02
[Merged by Bors] - chore(Mathlib/Init): remove autoImplicit

grunweg:
I know Mathlib/Init is mostly going to be removed --- but whatever *is* moved will benefit from being autoImplicit-correct.
(Besides, these files were easy to process, so one might as well do this.)

---


jcommelin:
Thanks :tada:

bors merge

PR #14357, opened 2024-07-02
[Merged by Bors] - chore(Data/Nat): refactor import of `Nat/Prime` by `Nat/Factors`

rwst:
Further reduce imports of full `Prime.lean`. Here, `Data.Nat.Factors` now imports only `Prime/Defs`.
For this, four lemmas had to be added to `Prime/Defs` from `Prime/Basic`. Many modules import
`Data.Nat.Factors`, so their dependencies on `Prime` theorems are now uncovered. Still, this
change will reduce the amount of what is imported, in general.

Motivation is a circular dependency that would prevent me from adding a lemma to `Prime/Basic`.

---


rwst:
!bench

rwst:
@tb65536 this PR would be necessary to be able to add #14049 to `Data/Nat/Prime/Basic` instead of `Archive`. You stated an interest in #14049. Could you please review?

semorrison:
bors merge

PR #11521, opened 2024-03-20
[Merged by Bors] - feat: add `fast_instance%` elaborator

eric-wieser:
Quoting the docstring:

`fast_instance% inst` takes an expression for a typeclass instance `inst`, and unfolds it into
constructor applications that leverage existing instances.
For instance, when used as
```lean
instance instSemiring : Semiring X := sorry
instance instRing : Ring X := fast_instance% Function.Injective.ring ..
```
this will define `instRing` as a nested constructor application that refers to `instSemiring`.
The advantage is then that `instRing.toSemiring` unifies almost immediately with `instSemiring`,
rather than having to break it down into smaller pieces.

Related to #7432

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14358, opened 2024-07-02
[Merged by Bors] - feat(RingTheory/AdicCompletion): comparison with tensor product

chrisflav:
We construct the natural map `AdicCompletion I R ⊗[R] M →ₗ[AdicCompletion I R] AdicCompletion I M` for any
`R`-module `M` and ideal `I` and show that it is an isomorphism for `M = R^n` and surjective if `M` is a finite `R`-module.

Co-authored-by: Judith Ludwig <ludwig@mathi.uni-heidelberg.de>

---

- [x] depends on: #14534

For context: In the next PR we conclude from this one that this above map is an isomorphism for `R` Noetherian and `M` finite and deduce from this that `AdicCompletion I R` is a flat `R`-algebra if `R` is Noetherian.


Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,86 @@
...
+Copyright (c) 2024 Judith Ludwig, Christian Merten. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Judith Ludwig, Christian Merten
+-/
+import Mathlib.LinearAlgebra.TensorProduct.Tower
+import Mathlib.LinearAlgebra.Pi
+
+/-!
+
+# Tensor product and products
```
riccardobrasca, line 11: Can you mention somewhere also the file `Mathlib.LinearAlgebra.TensorProduct.Prod`? (And maybe also the other way, mentioning this file there).

Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,86 @@
...
+Let `S` be an `R`-algebra, `N` an `S`-module and `ι` an index type. We then have a natural map
+
+`TensorProduct.piScalarRightHom`: `N ⊗[R] (ι → R) →ₗ[S] (ι → N)`
+
+In general, this is not an isomorphism, but if `ι` is finite, then it is
+and it is packaged as `TensorProduct.piScalarRight`.
+
+-/
+
+variable (R : Type*) [CommRing R]
```
riccardobrasca, line 24: Does something (at least the defs) work over semirings?

Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,86 @@
...
+variable (S : Type*) [CommRing S] [Algebra R S]
+variable (N : Type*) [AddCommMonoid N] [Module R N] [Module S N] [IsScalarTower R S N]
```
riccardobrasca, line 26:
```suggestion
variable (N : Type*) [AddCommGroup N] [Module R N] [Module S N] [IsScalarTower R S N]
```
Unless you move to semirings. Of course mathematically it is the same, but I think is more reasonable this way.

Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,86 @@
...
+variable (ι : Type*)
+
+open LinearMap TensorProduct
+
+private def TensorProduct.piScalarRightHomBil : N →ₗ[S] (ι → R) →ₗ[R] (ι → N) where
```
riccardobrasca, line 31: You can use `namespace TensorProduct` since a lot of declarations start like that.

Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,86 @@
...
+  simp [piScalarRightInv, Pi.single_apply, TensorProduct.ite_tmul]
+
+@[simp]
+lemma Pi.single_apply_smul (x : N) (i j : ι) :
+    (Pi.single i 1 : ι → R) j • x = (Pi.single i x : ι → N) j := by
+  simp [Pi.single_apply]
+
+/-- For any `R`-module `N` and finite index type `ι`, `N ⊗[R] (ι → R)` is canonically
+isomorphic to `ι → N`. -/
+noncomputable def TensorProduct.piScalarRight : N ⊗[R] (ι → R) ≃ₗ[S] (ι → N) :=
```
riccardobrasca, line 76: Can you add the relevant `apply` and `symm_apply` lemmas?

Mathlib/RingTheory/AdicCompletion/AsTensorProduct.lean
```diff
@@ -0,0 +1,150 @@
...
+import Mathlib.RingTheory.AdicCompletion.Exactness
+import Mathlib.RingTheory.TensorProduct.Basic
+
+/-!
+
+# Adic completion as tensor product
+
+In this file we examine properties of the natural map
+
+`AdicCompletion I R ⊗[R] M →ₗ[AdicCompletion I R] AdicCompletion I M`.
```
riccardobrasca, line 16:
```suggestion
`(AdicCompletion I R) ⊗[R] M →ₗ[AdicCompletion I R] AdicCompletion I M`.
```
Just to make it more readable here.

Mathlib/RingTheory/AdicCompletion/AsTensorProduct.lean
```diff
@@ -0,0 +1,150 @@
...
+  ext
+  simp
+
+section PiFintype
+
+/-
+In this section we show that `ofTensorProduct` is an isomorphism if `M = R^n`.
+-/
+
+attribute [-simp] _root_.smul_eq_mul Algebra.id.smul_eq_mul
```
riccardobrasca, line 81:
```suggestion
```
You can add `-smul_eq_mul, -Algebra.id.smul_eq_mul` in the `simpa` that break below to obtain the same result.

riccardobrasca:
I have the impression the file `AsTensorProduct` is quite slow.

!bench

Mathlib/LinearAlgebra/TensorProduct/Pi.lean
chrisflav: I added an `apply` lemma and a `symm_single` lemma as I think that is the most reasonable one I could think of (as the inverse is private).

chrisflav:
!bench

chrisflav:
Benchmark seems to be okay, but I agree the file is slow (and the last PR in this series will make it slower still). I think it's the module instance of  `AdicCompletion I M` over `AdicCompletion I R` that is slow, this was already visible in the algebra file and I think it's because of the many quotient rings involved. Do you think it is okay for now @riccardobrasca?
(In the future, might it help to introduce `IsAdicCompletion I M` which eliminates the explicit quotient construction?)

riccardobrasca:
> Benchmark seems to be okay, but I agree the file is slow (and the last PR in this series will make it slower still). I think it's the module instance of `AdicCompletion I M` over `AdicCompletion I R` that is slow, this was already visible in the algebra file and I think it's because of the many quotient rings involved. Do you think it is okay for now @riccardobrasca? (In the future, might it help to introduce `IsAdicCompletion I M` which eliminates the explicit quotient construction?)

Can you open a discussion on Zulip about the slowness? We surely want `IsAdicCompletion`, but also `AdicCompletion` (to complete a given ring), and maybe it is worth to start investigating now.

chrisflav:
> > Benchmark seems to be okay, but I agree the file is slow (and the last PR in this series will make it slower still). I think it's the module instance of `AdicCompletion I M` over `AdicCompletion I R` that is slow, this was already visible in the algebra file and I think it's because of the many quotient rings involved. Do you think it is okay for now @riccardobrasca? (In the future, might it help to introduce `IsAdicCompletion I M` which eliminates the explicit quotient construction?)
> 
> Can you open a discussion on Zulip about the slowness? We surely want `IsAdicCompletion`, but also `AdicCompletion` (to complete a given ring), and maybe it is worth to start investigating now.

Okay, I will do that after the weekend, since I will be away from a computer until Monday.

chrisflav:
!bench

chrisflav:
!bench

chrisflav:
!bench

chrisflav:
!bench

chrisflav:
!bench

chrisflav:
@riccardobrasca: The speed-up PR #14534 has been merged now (quick summary: currently no big speed-up, but when #11521 lands, this will also make the `AsTensorProduct` file fast). Anything else I should address for this PR?

riccardobrasca:
Thanks for having investigated this!

bors merge

chrisflav:
Thanks for the review!

PR #14360, opened 2024-07-02
add import_data

adomani:
---


adomani:
This PR served as an illustration of a new tab for the `Import summary` proposed in #14363.

It has now served its purpose.

PR #14363, opened 2024-07-02
[Merged by Bors] - feat: more import diff options

adomani:
This PR adds more import-diff information to the PR summary.  It introduces an extra collapsible tab containing the import change for *all* files for which there has been a change, sorted by decreasing change (first all the import increases and then all the decreases).

See #14360 for an example of how the new diff tab looks like.

This came up in [this Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/import.20changes.20bot.20on.20PRs/near/447893584).

---

### DIFF
--- a/.github/workflows/PR_summary.yml
+++ b/.github/workflows/PR_summary.yml
@@ -56,7 +56,11 @@ jobs:
         title="### PR summary"
 
         ## Import count comment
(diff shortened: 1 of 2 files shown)


semorrison:
bors d+

PR #14364, opened 2024-07-02
feat(Combinatorics/SimpleGraph): Add definition and coloring of `cycleGraph`

IvanRenison:
---


Parcly-Taxel:
Isn't this called a cycle graph?

IvanRenison:
Yes, you are right, it usually called like that. I will change it

PR #14366, opened 2024-07-02
[Merged by Bors] - feat(RingTheory/AdicCompletion): adic completion of Noetherian ring is flat

chrisflav:
We show that the adic completion of a Noetherian ring `R` is flat.

More precisely, we show that on finite modules over a Noetherian ring, tensoring a module `M` with the adic completion of `R` is the same as adically completing `M`. From this we conclude since adic completion on such modules is exact.

Co-authored-by: Judith Ludwig <ludwig@mathi.uni-heidelberg.de>

---

- [x] depends on: #14358


Mathlib/RingTheory/AdicCompletion/AsTensorProduct.lean
```diff
@@ -146,4 +158,211 @@ lemma ofTensorProduct_surjective_of_fg [Module.Finite R M] :
...
+def lTensorKerIncl : AdicCompletion I R ⊗[R] LinearMap.ker f →ₗ[AdicCompletion I R]
+    AdicCompletion I R ⊗[R] (ι → R) :=
+  AlgebraTensorModule.map LinearMap.id (LinearMap.ker f).subtype
+
+/- The second horizontal arrow in the top row. -/
+private def lTensorf :
+    AdicCompletion I R ⊗[R] (ι → R) →ₗ[AdicCompletion I R] AdicCompletion I R ⊗[R] M :=
+  AlgebraTensorModule.map LinearMap.id f
+
+private lemma if_exact : Function.Exact (LinearMap.ker f).subtype f := by
```
riccardobrasca, line 209: If you like golfing this can become a one liner I think.

Mathlib/RingTheory/AdicCompletion/AsTensorProduct.lean
```diff
@@ -146,4 +158,211 @@ lemma ofTensorProduct_surjective_of_fg [Module.Finite R M] :
...
+lemma tensor_map_id_left_eq_map :
+    (AlgebraTensorModule.map LinearMap.id f) =
+      (ofTensorProductEquivOfFiniteNoetherian I N).symm.toLinearMap ∘ₗ
+      map I f ∘ₗ
+      (ofTensorProductEquivOfFiniteNoetherian I M).toLinearMap := by
+  erw [ofTensorProduct_naturality I f]
+  ext x
+  simp
+
+lemma tensor_map_id_left_injective_of_injective (hf : Function.Injective f) :
```
riccardobrasca, line 348: `f` can be implicit here.

riccardobrasca (approved):
LGTM. My only question is whether it makes sense to keep on proving stuff about *the* completion rather than introducing a predicate `IsAdicCompletion`: we will surely want it, and you are just adding more work to adapt everything.

chrisflav:
> LGTM. My only question is whether it makes sense to keep on proving stuff about _the_ completion rather than introducing a predicate `IsAdicCompletion`: we will surely want it, and you are just adding more work to adapt everything.

I agree, this is the last PR in this series though. Also I think that the flatness can be rather easily obtained for `IsAdicCompletion` by just transferring via the isomorphism.

riccardobrasca:
bors d+

chrisflav:
Thanks for the review!

bors r+

PR #14367, opened 2024-07-02
[Merged by Bors] - chore: move iterate_prod_map to `Data.Prod.Basic`

D-Thomine:
The lemma [Function.iterate_prod_map](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Dynamics/PeriodicPts.html#Function.iterate_prod_map) is too fundamental to belong to a Dynamics folder.

It is moved to Data.Prod.Basic. The name of the lemma is changed from `iterate_prod_map` to `Prod.map_iterate`, which is more coherent with the usual naming (see `Continuous.iterate`, `pow_iterate`...) and thus more searchable.

---

### DIFF
--- a/Mathlib/Data/Prod/Basic.lean
+++ b/Mathlib/Data/Prod/Basic.lean
@@ -145,6 +145,8 @@ theorem map_iterate (f : α → α) (g : β → β) (n : ℕ) :
     (Prod.map f g)^[n] = Prod.map f^[n] g^[n] := by induction n <;> simp [*, Prod.map_comp_map]
 #align function.iterate_prod_map Prod.map_iterate
 
+@[deprecated (since := "2024-07-03")] alias iterate_prod_map := Prod.map_iterate
+
 theorem fst_surjective [h : Nonempty β] : Function.Surjective (@fst α β) :=
   fun x ↦ h.elim fun y ↦ ⟨⟨x, y⟩, rfl⟩
 #align prod.fst_surjective Prod.fst_surjective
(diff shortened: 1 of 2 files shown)


Mathlib/Data/Prod/Basic.lean
```diff
@@ -140,6 +140,10 @@ theorem map_id : Prod.map (@id α) (@id β) = id :=
   id_prod
 #align prod.map_id Prod.map_id
 
+@[simp]
+theorem map_iterate (f : α → α) (g : β → β) (n : ℕ) :
+    (Prod.map f g)^[n] = Prod.map f^[n] g^[n] := by induction n <;> simp [*, Prod.map_comp_map]
+
```
grunweg, line 146: Please keep the #align statement: you can change the second item to account for the rename.
```suggestion
#align function.iterate_prod_map Prod.map_iterate

grunweg: In addition, mathlib policy is that whenever you rename a lemma, you add an alias for the old name with a deprecation warning. (This way, code using the old name still compiles, but emits a warning pointing to the new name --- this makes for a much nicer experience in projects depending on mathlib.) Please add such an alias; the syntax is like so:
`@[deprecated (since := "2024-07-02")] alias Function.iterate_prod_map := Prod.map_iterate`.

grunweg (commented):
Thanks for your contribution to mathlib! I have two small comments. When you have made this change, please mark as `awaiting-review` again. Thanks!

Mathlib/Data/Prod/Basic.lean
D-Thomine: Done. Thanks for the information!

grunweg: Thanks - can you move the deprecation alias to the new file as well? I *think* this is a more natural place...
and I presume the dynamics file imports `Data/Prod/Basic` anyway, so this has no consequences for anybody importing `Dynamics`.

D-Thomine: Done! Just have to wait for the thing to build.

grunweg:
Thanks; this looks good to me now. Let's ask a maintainer to take a look.
maintainer merge

mattrobball:
!bench

mattrobball:
bors merge

PR #14373, opened 2024-07-03
[Merged by Bors] - feat(Algebra/Algebra/Basic) Extend scalars of linear maps along surjective ring homs.

erdOne:
---


semorrison:
Could you explain what this will be used for, either by expanding a doc string or in the PR description?

erdOne:
I'm not sure what I should add to the docstrings. My use case is exactly "If `R →+* S` is surjective, then `R`-linear maps are also `S`-linear."
An example is for a presentation `0 -> I -> R[X] -> S -> 0`, the map `I/I^2 -> S ⊗ Ω[R[X]⁄R]` is `R[X]`-linear, and this is used to show that it is `S` linear. Should I include this example anywhere?

semorrison:
bors merge

PR #14376, opened 2024-07-03
[Merged by Bors] - feat(Mathlib/Data/Matrix/Notation): delaborator for `!![a, b; c, d]` notation

Komyyy:
---


Mathlib/Data/Matrix/Notation.lean
```diff
@@ -114,6 +112,32 @@ macro_rules
...
+  let some (m, n) ← withNatValue em fun m => withNatValue en fun n => pure (some (m, n)) | failure
+  if m = 0 then
+    withAppArg <| do
+      guard <| (← getExpr).isAppOf ``vecEmpty
+      let commas := mkArray n (Syntax.atom .none ",")
+      `(!![$[,%$commas]*])
+  else
+    if n = 0 then
+      withAppArg <| do
+        let `(![$[![]%$evecs],*]) ← delab | failure
```
eric-wieser, line 134: Matching on `delab` like this feels a big fragile, but I don't think it's actually wrong.

eric-wieser (approved):
Nifty! I'm not super thrilled by the implementation matching on the output of a previous delab, but I also don't have a case in mind where it fails.

I'll let Kyle give it a final check.

Mathlib/Data/Matrix/Notation.lean
Komyyy: I don't know if this design pattern is wrong because I don't have much experience writing delaborator.
I think it is better to ask @kmill.

kmill: Matching on `delab` is at least done in the builtin delaborators.

Here's a refinement of the code, along with having it check that `pp.explicit` is false as well, since `DFunLike.coe` and `Matrix.of` have some implicit arguments.

```lean
@[delab app.DFunLike.coe]
def delabMatrixNotation : Delab := whenNotPPOption getPPExplicit <| whenPPOption getPPNotation <|
  withOverApp 6 do
    let mkApp3 (.const ``Matrix.of _) (.app (.const ``Fin _) em) (.app (.const ``Fin _) en) _ :=
      (← getExpr).appFn!.appArg! | failure
    let some m ← withNatValue em (pure ∘ some) | failure
    let some n ← withNatValue en (pure ∘ some) | failure
    withAppArg do
      if m = 0 then
        guard <| (← getExpr).isAppOfArity ``vecEmpty 1
        let commas := mkArray n (mkAtom ",")
        `(!![$[,%$commas]*])
      else
        if n = 0 then
          let `(![$[![]%$evecs],*]) ← delab | failure
          `(!![$[;%$evecs]*])
        else
          let `(![$[![$[$melems],*]],*]) ← delab | failure
          `(!![$[$[$melems],*];*])
```

Komyyy: Thank you @kmill! I have rewritten.

kmill:
Thanks!

bors r+

PR #14377, opened 2024-07-03
[Merged by Bors] - feat(RingTheory): Surjective on stalks ring homomorphisms.

erdOne:
---


Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,79 @@
...
+import Mathlib.AlgebraicGeometry.PrimeSpectrum.Basic
+
+/-!
+
+# Lemmas regarding the prime spectrum of tensor products
+
+## Main result
+- `PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks`:
+  If `R →+* T` is surjective on stalks,
+  then `Spec(S ⊗[R] T) → Spec S × Spec T` is an topological embedding.
```
dagurtomas, line 16:
```suggestion
  then `Spec(S ⊗[R] T) → Spec S × Spec T` is a topological embedding.
```

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,79 @@
...
+import Mathlib.AlgebraicGeometry.PrimeSpectrum.Basic
+
+/-!
+
+# Lemmas regarding the prime spectrum of tensor products
+
+## Main result
+- `PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks`:
+  If `R →+* T` is surjective on stalks,
+  then `Spec(S ⊗[R] T) → Spec S × Spec T` is a topological embedding.
```
jcommelin, line 16: Wait, don't we already have the statement that this is an iso? Or is it only an iso as raw types?

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,169 @@
...
+    RingHom.coe_comp] at this
+  exact this.of_comp
+
+open TensorProduct
+
+/--
+If `R → T` is surjective on stalks, and `J` is some prime of `T`,
+then every element `x` in `S ⊗[R] T` satisfies `(1 ⊗ rt)x = a ⊗ t` for some `r : R` and `rt ∉ J`.
+-/
+lemma exists_mul_eq_tmul_of_surjectiveOnStalks
```
dagurtomas, line 114: Here, using the namespace `SurjectiveOnStalks` might be useful for dot notation. For `of_surjective` and `of_isLocalization`, dot notation is useless, so I'm not sure we want them in the namespace. In any case you might want to just open the namespace at some point to avoid writing `SurjectiveOnStalks.` all the time.

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,169 @@
...
+namespace RingHom
+
+/--
+A ring homomorphism `R →+* S` is surjective on stalks if `R_p →+* S_q` is surjective for all pairs
+of primes `p = f⁻¹(q)`.
+-/
+def SurjectiveOnStalks (f : R →+* S) : Prop :=
+  ∀ (I : Ideal S) (_ : I.IsPrime), Function.Surjective (Localization.localRingHom _ I f rfl)
+
+lemma surjective_localRingHom_iff (P : Ideal S) [P.IsPrime] :
```
dagurtomas, line 30: Maybe this lemma deserves a docstring to explain its usage, since it's used a lot.

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,169 @@
...
+-/
+import Mathlib.RingTheory.Localization.AtPrime
+import Mathlib.RingTheory.TensorProduct.Basic
+
+/-!
+# Ring Homomorphisms surjective on stalks
+
+In this file, we prove some results on ring homomorphisms surjective on stalks, to be used in
+the development of immersions in algebraic geometry.
+
```
dagurtomas, line 14: Add some more details about what results are in this file?

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,169 @@
...
+  constructor
+  · intro H y
+    obtain ⟨a, ha⟩ := H (IsLocalization.mk' _ y (1 : P.primeCompl))
+    obtain ⟨a, t, rfl⟩ := IsLocalization.mk'_surjective (P.comap f).primeCompl a
+    rw [Localization.localRingHom_mk', IsLocalization.mk'_eq_iff_eq,
+      Submonoid.coe_one, one_mul, IsLocalization.eq_iff_exists P.primeCompl] at ha
+    obtain ⟨c, hc⟩ := ha
+    simp only [← mul_assoc] at hc
+    refine ⟨_, _, _, c.2, t.2, hc.symm⟩
+  · refine fun H y ↦ Localization.ind (fun ⟨y, t, h⟩ ↦ ?_) y
```
dagurtomas, line 42:
```suggestion
    exact ⟨_, _, _, c.2, t.2, hc.symm⟩
  · refine fun H y ↦ Localization.ind (fun ⟨y, t, h⟩ ↦ ?_) y
```

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,169 @@
...
+lemma SurjectiveOnStalks.baseChange
+    (hf : (algebraMap R T).SurjectiveOnStalks) :
+    (algebraMap S (S ⊗[R] T)).SurjectiveOnStalks := by
+  let g : T →+* S ⊗[R] T := Algebra.TensorProduct.includeRight.toRingHom
+  intros J hJ
+  rw [surjective_localRingHom_iff]
+  intro x
+  obtain ⟨t, r, a, ht, e⟩ := exists_mul_eq_tmul_of_surjectiveOnStalks hf x (J.comap g)
+    inferInstance
+  refine ⟨a, algebraMap _ _ r, 1 ⊗ₜ (r • t), ht, ?_, ?_⟩
```
dagurtomas, line 147:
```suggestion
  intro J hJ
  rw [surjective_localRingHom_iff]
  intro x
  obtain ⟨t, r, a, ht, e⟩ := exists_mul_eq_tmul_of_surjectiveOnStalks hf x (J.comap g)
    inferInstance
  exact ⟨a, algebraMap _ _ r, 1 ⊗ₜ (r • t), ht, ?_, ?_⟩ 
```

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,80 @@
...
+import Mathlib.RingTheory.SurjectiveOnStalks
+import Mathlib.AlgebraicGeometry.PrimeSpectrum.Basic
+
+/-!
+
+# Lemmas regarding the prime spectrum of tensor products
+
+## Main result
+- `PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks`:
+  If `R →+* T` is surjective on stalks,
```
dagurtomas, line 15: Could you add a reference to the file where "surjective on stalks" is defined?

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,80 @@
...
+open TensorProduct
+
+/-- The canonical map from `Spec(S ⊗[R] T)` to the cartesian product `Spec S × Spec T`. -/
+noncomputable
+def PrimeSpectrum.tensorProductTo (x : PrimeSpectrum (S ⊗[R] T)) :
+    PrimeSpectrum S × PrimeSpectrum T :=
+  ⟨comap (algebraMap _ _) x, comap Algebra.TensorProduct.includeRight.toRingHom x⟩
+
+lemma PrimeSpectrum.continuous_tensorProductTo : Continuous (tensorProductTo R S T) :=
+  (comap _).2.prod_mk (comap _).2
```
dagurtomas, line 32: Maybe this should just be term of type `C(_, _)`?

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,80 @@
...
+    Embedding (tensorProductTo R S T) := by
+  refine ⟨?_, fun p₁ p₂ e ↦
+    (embedding_tensorProductTo_of_surjectiveOnStalks_aux R S T hRT p₁ p₂ e).antisymm
+      (embedding_tensorProductTo_of_surjectiveOnStalks_aux R S T hRT p₂ p₁ e.symm)⟩
+  let g : T →+* S ⊗[R] T := Algebra.TensorProduct.includeRight.toRingHom
+  refine ⟨(continuous_tensorProductTo R S T).le_induced.antisymm ?_⟩
+  conv_rhs => rw [isBasis_basic_opens.eq_generateFrom]
+  rw [TopologicalSpace.le_generateFrom_iff_subset_isOpen]
+  rintro _ ⟨_, ⟨f, rfl⟩, rfl⟩
+  rw [Set.mem_setOf_eq, @isOpen_iff_forall_mem_open]
```
dagurtomas, line 67: Could this part be extracted to a lemma about when a topology `t` satisfies `t \le zariskiTopology`?

dagurtomas (commented):
Looks good, added some comments.

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
erdOne: I think I prefer unbundled morphisms more. Plus the main point of this function is that it is a topological embedding (under assumptions) and not merely continuous functions.

Mathlib/RingTheory/SurjectiveOnStalks.lean
erdOne: I don't think `exact` works here.

Mathlib/RingTheory/SurjectiveOnStalks.lean
erdOne: I adjusted the names according to your suggestions. I refrained from using `namespace` as only a few lemmas are in it.

Mathlib/RingTheory/SurjectiveOnStalks.lean
dagurtomas: You're right, I misread

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
dagurtomas: Yes, thanks

dagurtomas (approved):
LGTM

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,181 @@
...
+  intros I hI
+  have := hg I hI
+  rw [Localization.localRingHom_comp (I.comap (g.comp f)) (I.comap g) _ _ rfl _ rfl,
+    RingHom.coe_comp] at this
+  exact this.of_comp
+
+open TensorProduct
+
+/--
+If `R → T` is surjective on stalks, and `J` is some prime of `T`,
```
AntoineChambert-Loir, line 124: The docstring doesn't make sense.
`(1 ⊗ r • t) * x = a ⊗ t` for some `a : S`, `r : R` and `t: T` such that `r • t ∉ J`

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,181 @@
...
+then every element `x` in `S ⊗[R] T` satisfies `(1 ⊗ rt)x = a ⊗ t` for some `r : R` and `rt ∉ J`.
+-/
+lemma SurjectiveOnStalks.exists_mul_eq_tmul
+    (hf₂ : (algebraMap R T).SurjectiveOnStalks)
+    (x : S ⊗[R] T) (J : Ideal T) (hJ : J.IsPrime) :
+    ∃ (t : T) (r : R) (a : S), (r • t ∉ J) ∧
+      (1 : S) ⊗ₜ[R] (r • t) * x = a ⊗ₜ[R] t := by
+  induction' x using TensorProduct.induction_on with x₁ x₂ x₁ x₂ hx₁ hx₂
```
AntoineChambert-Loir, line 132: I had been told that one should now prefer `induction` to `induction'`.

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,181 @@
...
+variable {T} [CommRing T] [Algebra R T] [Algebra S T] [IsScalarTower R S T]
+variable {g : S →+* T} {f : R →+* S}
+
+namespace RingHom
+
+/--
+A ring homomorphism `R →+* S` is surjective on stalks if `R_p →+* S_q` is surjective for all pairs
+of primes `p = f⁻¹(q)`.
+-/
+def SurjectiveOnStalks (f : R →+* S) : Prop :=
```
AntoineChambert-Loir, line 32: This is not crucial, but below you denote prime ideals as `P`, maybe that would be clearer here.

Mathlib/Topology/Sets/Opens.lean
```diff
@@ -306,6 +306,12 @@ theorem IsBasis.isCompact_open_iff_eq_finite_iUnion {ι : Type*} (b : ι → Ope
     simp
   · exact hb'
 
+lemma IsBasis.le_iff {α} {t₁ t₂ : TopologicalSpace α}
+    {Us : Set (Opens α)} (hUs : IsBasis Us) : -- basis for t₂
```
AntoineChambert-Loir, line 310: What in the syntax makes us sure that it is a basis for `t₂`? Wouldn't it be safer to enter `@IsBasis α t₂ Us`?

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,78 @@
...
+def PrimeSpectrum.tensorProductTo (x : PrimeSpectrum (S ⊗[R] T)) :
+    PrimeSpectrum S × PrimeSpectrum T :=
+  ⟨comap (algebraMap _ _) x, comap Algebra.TensorProduct.includeRight.toRingHom x⟩
+
+lemma PrimeSpectrum.continuous_tensorProductTo : Continuous (tensorProductTo R S T) :=
+  (comap _).2.prod_mk (comap _).2
+
+variable (hRT : (algebraMap R T).SurjectiveOnStalks)
+
+lemma PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks_aux
```
AntoineChambert-Loir, line 36: `SurjectiveOnStalks` uses `(I : Ideal _) (_ : IsPrime I)`.  Here, you prefer `p : PrimeSpectrum _`. In what case do you prefer which?

erdOne: I only use `PrimeSpectrum` when I need the zarkisi topology.

AntoineChambert-Loir: Is it needed here?

AntoineChambert-Loir: Oh, I see. The `_aux` function proves half of injectivity, but  embedding is both injectivity and something about the topology.

Mathlib/Topology/Sets/Opens.lean
```diff
@@ -306,6 +306,12 @@ theorem IsBasis.isCompact_open_iff_eq_finite_iUnion {ι : Type*} (b : ι → Ope
     simp
   · exact hb'
 
+lemma IsBasis.le_iff {α} {t₁ t₂ : TopologicalSpace α}
```
AntoineChambert-Loir, line 309: (By the way, maybe a leaner syntax such as`IsBasis[t₂] Us` was preferable, I don't know the usages here.)

erdOne: I just looked for it as well but seems like it does not exist. This is the only occurrence of `@IsBasis` in mathlib so maybe it's not worth adding it.

AntoineChambert-Loir: Maybe that exists for `Opens`, anyway, there's no ambiguity here which is what matters.

Mathlib/RingTheory/SurjectiveOnStalks.lean
```diff
@@ -0,0 +1,185 @@
...
+    ⟨r, 1, 1, by simpa [← Ideal.eq_top_iff_one], by simpa [← Ideal.eq_top_iff_one], by simp [hr]⟩
+
+variable (S) in
+lemma surjectiveOnStalks_of_isLocalization [IsLocalization M S] :
+    SurjectiveOnStalks (algebraMap R S) := by
+  refine surjectiveOnStalks_of_exists_div fun s ↦ ?_
+  obtain ⟨x, s, rfl⟩ := IsLocalization.mk'_surjective M s
+  exact ⟨x, s, IsLocalization.map_units S s, IsLocalization.mk'_spec' S x s⟩
+
+lemma SurjectiveOnStalks.comp (hg : SurjectiveOnStalks g) (hf : SurjectiveOnStalks f) :
```
AntoineChambert-Loir, line 107: Besides composition, is it relevant to add the case of `id` or would the case of localization take care of it?

erdOne: `surjectiveOnStalks_of_surjective` should cover it quite easily.

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,78 @@
...
+# Lemmas regarding the prime spectrum of tensor products
+
+## Main result
+- `PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks`:
+  If `R →+* T` is surjective on stalks (see Mathlib/RingTheory/SurjectiveOnStalks.lean),
+  then `Spec(S ⊗[R] T) → Spec S × Spec T` is a topological embedding
+  (where `Spec S × Spec T` is the cartesian product with the product topology).
+-/
+
+variable (R S T) [CommRing R] [CommRing S] [Algebra R S]
```
riccardobrasca, line 20:
```suggestion
variable (R S T : Type*) [CommRing R] [CommRing S] [Algebra R S]
```
Is this the same?

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,78 @@
...
+  obtain ⟨t, r, a, ht, e⟩ := hRT.exists_mul_eq_tmul x
+    (p₂.asIdeal.comap g) inferInstance
+  have h₁ : a ⊗ₜ[R] t ∈ p₁.asIdeal := e ▸ p₁.asIdeal.mul_mem_left (1 ⊗ₜ[R] (r • t)) hxp₁
+  have h₂ : a ⊗ₜ[R] t ∉ p₂.asIdeal := e ▸ p₂.asIdeal.primeCompl.mul_mem ht hxp₂
+  rw [← mul_one a, ← one_mul t, ← Algebra.TensorProduct.tmul_mul_tmul] at h₁ h₂
+  have h₃ : t ∉ p₂.asIdeal.comap g := fun h ↦ h₂ (Ideal.mul_mem_left _ _ h)
+  have h₄ : a ∉ p₂.asIdeal.comap (algebraMap S (S ⊗[R] T)) :=
+    fun h ↦ h₂ (Ideal.mul_mem_right _ _ h)
+  replace h₃ : t ∉ p₁.asIdeal.comap g := by
+    rwa [show p₁.asIdeal.comap g = p₂.asIdeal.comap g from congr($h.2.1)]
```
riccardobrasca, line 52: What does the `$` do? (personal curiosity).

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
```diff
@@ -0,0 +1,78 @@
...
+
+lemma PrimeSpectrum.embedding_tensorProductTo_of_surjectiveOnStalks :
+    Embedding (tensorProductTo R S T) := by
+  refine ⟨?_, fun p₁ p₂ e ↦
+    (embedding_tensorProductTo_of_surjectiveOnStalks_aux R S T hRT p₁ p₂ e).antisymm
+      (embedding_tensorProductTo_of_surjectiveOnStalks_aux R S T hRT p₂ p₁ e.symm)⟩
+  let g : T →+* S ⊗[R] T := Algebra.TensorProduct.includeRight.toRingHom
+  refine ⟨(continuous_tensorProductTo ..).le_induced.antisymm (isBasis_basic_opens.le_iff.mpr ?_)⟩
+  rintro _ ⟨f, rfl⟩
+  rw [@isOpen_iff_forall_mem_open]
```
riccardobrasca, line 65: Is the `@` needed here?

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
erdOne: It is the syntax of `congr`. If `e` is a term of type `a = b`, `congr(f $e)` is a term of type `f a = f b`. Here it is `congr(($h).2.1)` Apparently lean is smart enough to figure out the dot notations.

Mathlib/AlgebraicGeometry/PrimeSpectrum/TensorProduct.lean
erdOne: Yes because the `TopologicalSpace` argument is not the one given by typeclass synthesis.

erdOne:
Thanks for the reviews!
bors merge

PR #14382, opened 2024-07-03
[Merged by Bors] - feat(Topology/Profinite): extending cones in `Profinite` and `LightProfinite`

dagurtomas:
Let `(Sᵢ)_{i : I}` be a family of finite sets indexed by a cofiltered category `I` and let `S` be
its limit in `Profinite`. Let `G` be a functor from `Profinite` to a category `C` and suppose that
`G` preserves the limit described above. Suppose further that the projection maps `S ⟶ Sᵢ` are
epimorphic for all `i`. Then `G.obj S` is isomorphic to a limit indexed by
`StructuredArrow S toProfinite` (see `Profinite.Extend.isLimitCone`).

We also provide the dual result for a functor of the form `G : Profiniteᵒᵖ ⥤ C`.

---


Mathlib/CategoryTheory/Limits/Cones.lean
```diff
@@ -304,6 +304,14 @@ theorem cone_iso_of_hom_iso {K : J ⥤ C} {c d : Cone K} (f : c ⟶ d) [i : IsIs
...
         w := fun j => (asIso f.hom).inv_comp_eq.2 (f.w j).symm }, by aesop_cat⟩⟩
 
+/-- Given an isomorphism of cones, produce an isomorphism of their points. -/
+@[simps]
+def ptIsoOfIso  {K : J ⥤ C} {c d : Cone K} (i : c ≅ d) : c.pt ≅ d.pt where
+  hom := i.hom.hom
+  inv := i.inv.hom
+  hom_inv_id := by simp [← Cone.category_comp_hom]
+  inv_hom_id := by simp [← Cone.category_comp_hom]
+
```
dagurtomas, line 314:
```suggestion
```

Mathlib/Topology/Category/Profinite/Extend.lean
```diff
@@ -0,0 +1,207 @@
...
+
+universe u w
+
+open CategoryTheory Limits FintypeCat Functor
+
+attribute [local instance] ConcreteCategory.instFunLike
+
+namespace Profinite
+
+variable {I : Type u} [Category.{u} I] [IsCofiltered I]
```
adamtopaz, line 34:
```suggestion
variable {I : Type u} [SmallCategory I] [IsCofiltered I]
```

Mathlib/Topology/Category/Profinite/Extend.lean
```diff
@@ -0,0 +1,207 @@
...
+  (functor c).op ⋙ StructuredArrow.toCostructuredArrow _ _
+
+-- We check that the opposite of the original diagram factors through `Profinite.Extend.functorOp`.
+example : functorOp c ⋙ CostructuredArrow.proj toProfinite.op ⟨c.pt⟩ ≅ F.op := Iso.refl _
+
+/--
+If the projection maps in the cone are epimorphic and the cone is limiting, then
+`Profinite.Extend.functor` is initial.
+-/
+lemma functor_initial (hc : IsLimit c) [∀ i, Epi (c.π.app i)] : Initial (functor c) := by
```
adamtopaz, line 82: I think this is also true with the weaker assumption that for all i there exists a j mapping to i such that `c.pi.app j` is epi, right? The way you stated it is the right way to do it from the point of view of the typeclass system, but we may want to think later about how to formulate such weaker assumptions.

Mathlib/Topology/Category/Profinite/Extend.lean
```diff
@@ -0,0 +1,207 @@
...
+end Extend
+
+open Extend
+
+section ProfiniteAsLimit
+
+variable (S : Profinite.{u})
+
+/-- A functor `StructuredArrow S toProfinite ⥤ Profinite` whose limit is isomorphic to `S`. -/
+abbrev diagram' : StructuredArrow S toProfinite ⥤ Profinite :=
```
adamtopaz, line 190: I think it would be better for this diagram to have type `StructuredArrow S toProfinite ⥤ FintypeCat`, and work with the composition with `toProfinite` when you need the diagram to take values in `Profinite`.

Mathlib/Topology/Category/LightProfinite/Extend.lean
```diff
@@ -0,0 +1,193 @@
...
+open Extend
+
+section LightProfiniteAsLimit
+
+variable (S : LightProfinite.{u})
+
+/--
+A functor `StructuredArrow S toLightProfinite ⥤ LightProfinite` whose limit is isomorphic to `S`.
+-/
+abbrev diagram' : StructuredArrow S toLightProfinite ⥤ LightProfinite :=
```
adamtopaz, line 176: Similarly here... I think this should take values in `FintypeCat`.

adamtopaz (commented):
Looks good! I had a few suggestions which are fairly minor. I assume you're preparing to relate `Profinite.extend` and friends to Kan extensions?

Mathlib/Topology/Category/Profinite/Extend.lean
dagurtomas: I was trying to make the naming consistent with `Profinite.diagram` from the `asLimit` file. There we have `Profinite.fintypeDiagram` taking values in `FintypeCat` and `Profinite.diagram` its postcomposition with `toProfinite`. Here, the `Profinite.fintypeDiagram'` would just be `StructuredArrow.proj _ _` so I didn't feel that it deserved a special name. But I can add it as another `abbrev`.

Mathlib/Topology/Category/Profinite/Extend.lean
dagurtomas: I guess so. I'll add a TODO in the docstring

adamtopaz:
Thanks!
bors r+

PR #14385, opened 2024-07-03
[Merged by Bors] - chore: remove ported tactic from Mathport syntax

adomani:
`move_add`, `move_mul` and `move_oper` are all available in `Mathlib`.

---

### DIFF
--- a/Mathlib/Mathport/Syntax.lean
+++ b/Mathlib/Mathport/Syntax.lean
@@ -253,12 +253,6 @@ syntax generalizingClause := " generalizing" (ppSpace ident)+
 
 /- E -/ syntax (name := wittTruncateFunTac) "witt_truncate_fun_tac" : tactic
 
-/- M -/ syntax (name := moveOp) "move_op " term:max ppSpace rwRule,+ (location)? : tactic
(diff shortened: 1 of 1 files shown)


grunweg (commented):
Thanks for doing this! move_add and move_mul look good to me - but wasn't `move_op` renamed to `move_oper`?

grunweg:
awaiting-author

adomani:
It is called `move_oper` now, yes, but where should I report that?

grunweg:
Good question! I don't know if there's a way to report that - if there isn't, I won't object to removing this as-is.

grunweg:
Should perhaps mathport be tweaked, to emit `move_oper` instead of `move_op`?

adomani:
I have not looked at mathport, but the only references to `move_op` that I could find in mathlib are the ones that I removed.  The tactic was almost never used anyway, and probably really never in its `move_op`/`move_oper` form, so I do not think that it is a big deal if mathport converts it incorrectly.

I am going to re-label as awaiting-review, and maybe someone else can give an opinion as well!

Ruben-VandeVelde (approved):
maintainer merge

semorrison:
bors merge

PR #14391, opened 2024-07-03
[Merged by Bors] - doc: fix some author fields in references

acmepjz:
---

### DIFF
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -608,7 +608,7 @@
 }
 
 @Book{            cabreragarciarodriguezpalacios2014,
-  author        = {Miguel {Cabrera Garc\'{\i}a} and \'Angel {Rodr\'{\i}guez
+  author        = {Miguel {Cabrera Garc\'ia} and \'Angel {Rodr\'iguez
                   Palacios}},
   title         = {{Non-associative normed algebras. Volume 1. The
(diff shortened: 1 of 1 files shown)


acmepjz:
I think that `\'{\i}` is also wrong, it should be `\'i`. Any opinions?

faenuccio:
I have removed the `awaiting-review` label according to the decision explained here
https://leanprover.zulipchat.com/#narrow/stream/113486-announce/topic/Changes.20to.20the.20.23queue/near/450176907

acmepjz:
> I have removed the `awaiting-review` label according to the decision explained here https://leanprover.zulipchat.com/#narrow/stream/113486-announce/topic/Changes.20to.20the.20.23queue/near/450176907

I see. Sorry for the inconvenience.

eric-wieser:
> I think that `\'{\i}` is also wrong (it does not render correctly in my WIP pure Lean bibtex diacritic parser [leanprover/doc-gen4#200](https://github.com/leanprover/doc-gen4/pull/200)), it should be `\'i`. Any opinions?

https://tex.stackexchange.com/q/409927/41112 suggests that the original is valid, but unnecessary in modern LaTeX.

bors merge

Thanks!

PR #14411, opened 2024-07-04
[Merged by Bors] - feat(Data/Multiset/Order): add Dershowitz-Manna Ordering and Theorem

haitian-yuki:
add Dershowitz-Manna Ordering and Theorem for Multisets

Co-authored-by: Malvin Gattinger <malvin@w4eg.eu>

---

- [x] depends on: #15645

https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Dershowitz-Manna.20Ordering.20theorem/near/434390710


riccardobrasca:
Can you please mark this PR `WIP` or `awaiting-review`? Thanks!

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+  split <;> simp_all
+
+lemma mul_mem_not_erase [DecidableEq α] {a a0: α} {M X : Multiset α}
+    (H : M = (a0 ::ₘ X).erase a) (hyp : ¬ a = a0) : a0 ∈ M := by
+  rw [H]
+  have : a0 ∈ (a0 ::ₘ X).erase a ↔ a0 ∈ (a0 ::ₘ X) := by
+    apply Multiset.mem_erase_of_ne
+    aesop
+  rw [this]
+  simp_all
```
Rida-Hamadani, line 93:
```suggestion
  rw [H, Multiset.mem_erase_of_ne fun h ↦ hyp <| id <| Eq.symm h, Multiset.mem_cons]
  tauto
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+  have := tsub_add_cancel_of_le h0
+  linarith
+
+lemma double_split {α} [dec : DecidableEq α]:
+    ∀ (M N P Q: Multiset α) ,  M + N = P + Q → N = N ∩ Q + (P - M)  := by
+  intros M N P Q h
+  ext x
+  rw [Multiset.count_add]
+  rw [Multiset.count_inter]
+  rw [Multiset.count_sub]
```
Rida-Hamadani, line 314:
```suggestion
  rw [Multiset.count_add, Multiset.count_inter, Multiset.count_sub]
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+  have H0 : M.count x + N.count x = P.count x + Q.count x := by
+    rw [Multiset.ext] at h
+    simp_all only [Multiset.mem_add, Multiset.count_add]
+  if l_u : M.count x ≤ P.count x then
+    have : N.count x ≥ Q.count x := by linarith
+    simp_all only [ge_iff_le, min_eq_right]
+    apply le_eq_sub (M.count x) (N.count x) (P.count x) (Q.count x)
+    · simp_all
```
Rida-Hamadani, line 322:
```suggestion
    · exact l_u
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+    · exact H0
+  else
+    simp_all only [not_le, gt_iff_lt]
+    have : Multiset.count x N ≤ Multiset.count x Q := by linarith
+    have:= le_of_lt l_u
```
Rida-Hamadani, line 327:
```suggestion
    have := le_of_lt l_u
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+      rw [this]
+      rw [add_assoc]
+    rw [this]
+    have : Z1 ∩ Z2 + (Y1 - X2) = Z2 ∩ Z1 + (Y1 - X2) := by
+      rw [Multiset.inter_comm]
+    rw [this]
+    rw [← double_split]
+    rw [add_comm]
+    rw [← N2_def]
+    rw [N1_def]
```
Rida-Hamadani, line 360:
```suggestion
    rw [this]
    rw [← double_split]
    rw [add_comm]
    rw [← N2_def]
    rw [N1_def]
```
Please put all rewrites in a single `rw` whenever possible. This comment applies in other places in the PR too.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+  simp_all
+
+lemma mem_erase_cons [DecidableEq α] {a0: α} {M : Multiset α} (_ : a0 ∈ M) :
+    M = M - {a0} + {a0} := by
+  rw [add_comm]
+  simp_all [Multiset.singleton_add, mul_singleton_erase]
+
+lemma cons_erase [DecidableEq α] {a a0: α} {M X : Multiset α}
+    (H : a ::ₘ M = X + {a0}) : M = X + {a0} - {a} := by
+  if hyp : a = a0 then
```
Rida-Hamadani, line 102: Please use `by_cases` instead of `if ... then`. This applies everywhere in the PR.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+                  have x_in_fy : x ∈ f y := by
+                    unfold_let f; simp; rw [← this]; exact ⟨x_in_X, x_lt_t⟩
+                  have x_notin_Xfy : x ∉ X - f y := by
+                    by_contra
+                    let neg_f : α → Multiset α := fun y' => X.filter (fun x => ¬ x < y')
+                    have : X - f y = neg_f y := by
+                      have fy_negfy_X : f y + neg_f y = X := by
+                        rw [Multiset.filter_add_not]
+                      rw [← fy_negfy_X]; simp
+                    have x_in_neg_fy : x ∈ neg_f y := by rw [this] at x_in; exact x_in
```
Rida-Hamadani, line 504:
```suggestion
                    have x_in_neg_fy : x ∈ neg_f y := this ▸ x_in
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+                    assumption
+                  have x_in_fy : x ∈ f y := by
+                    unfold_let f; simp; rw [← this]; exact ⟨x_in_X, x_lt_t⟩
+                  have x_notin_Xfy : x ∉ X - f y := by
+                    by_contra
+                    let neg_f : α → Multiset α := fun y' => X.filter (fun x => ¬ x < y')
+                    have : X - f y = neg_f y := by
+                      have fy_negfy_X : f y + neg_f y = X := by
+                        rw [Multiset.filter_add_not]
+                      rw [← fy_negfy_X]; simp
```
Rida-Hamadani, line 503:
```suggestion
                      have fy_negfy_X : f y + neg_f y = X := Multiset.filter_add_not _ _
                      rw [← fy_negfy_X, add_tsub_cancel_left]
```

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,552 @@
...
+                  have : t = y := by
+                    have : Y = newY + {y} := by
+                      unfold_let newY
+                      rw [add_comm, Multiset.singleton_add]
+                      simp [Multiset.cons_erase claim]
+                    rw [this] at t_in_Y
+                    rw [Multiset.mem_add] at t_in_Y
+                    have : t ∈ ( {y} : Multiset α) := by exact Or.resolve_left t_in_Y t_in_newY
+                    rw [← Multiset.mem_singleton]
+                    assumption
```
Rida-Hamadani, line 494:
```suggestion
                    rwa [← Multiset.mem_singleton]
```
You can replace `simp; assumption` and `rw [...]; assumption` by `simpa` and `rwa`.

Rida-Hamadani (commented):
Thank you! Some comments are general and need to be applied in other places in PR (instead of just committing the suggested portions).

Mathlib/Data/Multiset/Order.lean
haitian-yuki: Thanks! One thing I notice is that I can only combine them into two lines but not one and I thought they should be equivalent.

Rida-Hamadani: This is interesting! If I had to guess what is going on here: it is probably because `rw` tries `rfl` after the end of the rewrites, and in this case, there are 2 goals, the first goal is being closed by the `rfl` which is called at the end of the first `rw`. This can be verified by replacing the first `rw` with `rewrite`, which rewrites without trying `rfl` at the end. If an error occurred, then my guess above is correct.

haitian-yuki: Thanks! I think you are right :-)

This: 
`
rw [this,← double_split]
rw [add_comm, ← N2_def, N1_def]
`
works while

`
rewrite [this,← double_split]
rw [add_comm, ← N2_def, N1_def]
`
gives an error.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+  Our code here is inspired by their version of called `mOrd_wf` in the file
+  [MultisetList.v](https://github.com/fblanqui/color/blob/1.8.5/Util/Multiset/MultisetList.v).
+
+-/
+
+variable {α : Type*}
+
+/-- The standard Dershowitz–Manna ordering: -/
+inductive MultisetLT [DecidableEq α] [Preorder α] (M N : Multiset α) : Prop :=
+  | MLT : ∀ (X Y Z: Multiset α),
```
semorrison, line 38:
```suggestion
  | MLT : ∀ (X Y Z : Multiset α),
```
Spaces around colons. Also below.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+  | RedLt : ∀ (X Y:Multiset α) (a : α) ,
+    (M = X + Y) →
+    (N = X + {a}) →
+    (∀ y, y ∈ Y → y < a) → MultisetRedLt M N
+
+open Relation
+
+/-- MultisetLt is the transitive closure of MultisetRedLt and is equivalent to MultisetLT
+    (proved later). -/
+def MultisetLt [DecidableEq α][LT α] : Multiset α → Multiset α → Prop := TransGen MultisetRedLt
```
semorrison, line 56:
```suggestion
def MultisetLt [DecidableEq α] [LT α] : Multiset α → Multiset α → Prop := TransGen MultisetRedLt
```
Spaces between brackets.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+Copyright (c) 2024 Haitian Wang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Haitian Wang, Malvin Gattinger
+-/
+import Mathlib.Tactic.Linarith
+import Mathlib.Data.Multiset.Basic
+import Mathlib.Logic.Relation
+
+/-!
+# Dershowitz-Manna ordering
```
semorrison, line 11: I think "Dershowitz–Manna" should be in the filename, and possibly also in the name of the inductive.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+inductive MultisetLT [DecidableEq α] [Preorder α] (M N : Multiset α) : Prop :=
+  | MLT : ∀ (X Y Z: Multiset α),
+    Y ≠ ∅ →
+    M = Z + X →
+    N = Z + Y →
+    (∀ x ∈ X, ∃ y ∈ Y, x < y) → MultisetLT M N
+
+/-- MultisetRedLt is a special case of MultisetLT. The transitive closure of it is used to define
+an equivalent (proved later) version of the ordering. -/
+inductive MultisetRedLt [DecidableEq α][LT α] (M N : Multiset α) : Prop :=
```
semorrison, line 46: `LT` not `Lt`

semorrison: And in particular please don't differentiate two closely related definitions solely by capitalisation.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+    (∀ y, y ∈ Y → y < a) → MultisetRedLt M N
+
+open Relation
+
+/-- MultisetLt is the transitive closure of MultisetRedLt and is equivalent to MultisetLT
+    (proved later). -/
+def MultisetLt [DecidableEq α][LT α] : Multiset α → Multiset α → Prop := TransGen MultisetRedLt
+
+/-- AccM_1 defines the accessibility relation given MultisetRedLt. -/
+def AccM_1 [DecidableEq α][Preorder α] : Multiset α → Prop := Acc MultisetRedLt
```
semorrison, line 59: What do the numerical suffixes denote?

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+theorem redLt_LT [DecidableEq α] [Preorder α] (M N : Multiset α) :
+    MultisetRedLt M N → MultisetLT M N := by
+  intro hyp
+  rcases hyp with ⟨X, Y, a, M_def, N_def, ys_lt_a⟩
+  apply MultisetLT.MLT Y {a} X _ M_def N_def
+  · simpa
+  · simp
+
+/- Some useful lemmas about Multisets and the defined relations: -/
+lemma not_MultisetRedLt_0 [DecidableEq α] [LT α] (M: Multiset α) : ¬ MultisetRedLt M 0 := by
```
semorrison, line 71: Usually we use `zero` in the name, not `0`.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+  intro h
+  cases h with
+  | RedLt X Y a M nonsense _ =>
+    have contra : a ∈ (0 : Multiset α):= by
+      rw [nonsense]
+      simp_all only [Multiset.mem_add, Multiset.mem_singleton, or_true]
+    contradiction
+
+lemma mul_singleton_erase [DecidableEq α] {a : α} {M : Multiset α} :
```
semorrison, line 80: Please move to an earlier file (and double check it doesn't already exist).

semorrison:
Obviously my review is far from complete, but I'm not an expert here so hopefully someone else will come along, but these will give you something to start with.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+          apply mul_mem_not_erase
+          · change M = Multiset.erase (a0 ::ₘ X) a
+            rw [mul_singleton_erase, add_comm] at this0
+            exact this0
+          · exact hyp
+        apply mem_erase_cons
+        · exact a0M
+      exact H2
+
+lemma mord_wf_1 [DecidableEq α] [Preorder α] (a : α) (M0 : Multiset α)
```
semorrison, line 168: Needs a better name, I think.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+/- If `MultisetRedLt` is well-founded, then its transitive closure `MultisetLt` is also
+well-founded. -/
+lemma Lt_wf [DecidableEq α] [LT α]
+    (h : WellFounded (MultisetRedLt : Multiset α → Multiset α → Prop)) :
+    WellFounded (MultisetLt : Multiset α → Multiset α → Prop) := by
+  unfold MultisetLt
+  apply WellFounded.transGen
+  assumption
+
+lemma mem_leq_diff [DecidableEq α] : ∀ (M N: Multiset α), N ≤ M → M = M - N + N := by
```
semorrison, line 253: This and following need to move to an earlier file. 

Indeed, it may be best to make a separate PR with the basic Multiset material, so it can be reviewed properly separately.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+  intros M N h
+  rw [← Multiset.union_def, Multiset.eq_union_left]
+  exact h
+
+lemma le_sub_add {α} [dec : DecidableEq α]:
```
semorrison, line 258: Name should be something like `sub_add_of_le`.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+    ∀ (M N P : Multiset α) , N ≤ M → M - N + P = M + P - N  := by
+  intro M N P h
```
semorrison, line 260: Do proofs like this not work by something close to `ext; intros; simp; omega`? I guess you need to extract the componentwise consequence of the hypothesis too.

Mathlib/Data/Multiset/Order.lean
```diff
@@ -0,0 +1,513 @@
...
+      have y_in_Y1 : y ∈ Y1 := by
+        have : Y1 - X2 ≤ Y1 := by simp_all only [tsub_le_iff_right, le_add_iff_nonneg_right,
+        zero_le]
+        apply Multiset.mem_of_le
+        exact this
+        exact y_in
+      let _ := h y_in_Y1
+      aesop
+
+lemma direct_subset_red [dec : DecidableEq α] [Preorder α]
```
semorrison, line 373: The names all need to change, but this lemma needs to be named `X_of_Y` for the appropriate X and Y.

Mathlib/Data/Multiset/Order.lean
m4lvin: Maybe this can be replaced by using [`Multiset.sub_eq_fold_erase`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Multiset/Basic.html#Multiset.sub_eq_fold_erase) instead?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+  Our code here is inspired by their formalization and the theorem is called `mOrd_wf` in the file
+  [MultisetList.v](https://github.com/fblanqui/color/blob/1.8.5/Util/Multiset/MultisetOrder.v).
+
+-/
+
+variable {α : Type*}
+
+/-- The standard Dershowitz–Manna ordering: -/
+inductive MultisetDMLT [DecidableEq α] [Preorder α] (M N : Multiset α) : Prop :=
+  | DMLT : ∀ (X Y Z : Multiset α),
```
goens, line 38: not clear why this needs to be an inductive (with one constructor), couldn't it just be a def?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+    Y ≠ ∅ →
+    M = Z + X →
+    N = Z + Y →
+    (∀ x ∈ X, ∃ y ∈ Y, x < y) → MultisetDMLT M N
+
+/-- MultisetRedLT is a special case of MultisetDMLT. The transitive closure of it is used to define
+    an equivalent (proved later) version of the ordering. -/
+inductive MultisetRedLT [DecidableEq α] [LT α] (M N : Multiset α) : Prop :=
+  | RedLT : ∀ (X Y : Multiset α) (a : α) ,
```
goens, line 47: Also here, why is this a one-constructor inductive without self reference? Seems to complicate things, no?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+    (M = X + Y) →
+    (N = X + {a}) →
+    (∀ y, y ∈ Y → y < a) → MultisetRedLT M N
+
+open Relation
+
+/-- MultisetTransLT is the transitive closure of MultisetRedLT and is equivalent to MultisetDMLT
+    (proved later). -/
+def MultisetTransLT [DecidableEq α] [LT α] : Multiset α → Multiset α → Prop :=
```
goens, line 56: I'm not 100% sure exactly what the Mathlib convention is, but shouldn't the `Multiset` be a namespace instead of part of the name in CammelCase? i.e. `Multiset.transLT` and `Multiset.DMLT` above, etc?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+    left
+    constructor
+    · rw [h1]
+      have : X = (M - {a0} + {a}) := by
+        rw [add_comm, Multiset.singleton_add] at *
+        ext b
+        simp [Multiset.ext, Multiset.count_cons] at h0
+        by_cases h : b = a
+        · have := h0 b
+          aesop
```
goens, line 100: does `aesop?` work (not sure if you want to leave the `aesop` calls, in the same spirit as `simp` vs `simp only`)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+      simp_all
+    · simp_all
+  · exists (Y + (M - {a0}))
+    left
+    constructor
+    · rw [h1]
+      have : X = (M - {a0} + {a}) := by
+        rw [add_comm, Multiset.singleton_add] at *
+        ext b
+        simp [Multiset.ext, Multiset.count_cons] at h0
```
goens, line 97: could this be replaced with a `simp only` (using `simp`)?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+        by_cases h : b = a
+        · have := h0 b
+          aesop
+        · have := h0 b
+          simp [Multiset.sub_singleton]
+          aesop
```
goens, line 103: another instance of an `aesop` call that may be replaced with the result of `aesop?` (I think)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+    ∀ M, AccM M → AccM (a ::ₘ M) := by
+  unfold AccM
+  intros M h0
+  induction h0 with
+  | intro x wfH wfh2 =>
+    apply acc_cons
+    · simpa
+    · constructor
+      simpa
+    · simpa
```
goens, line 157: can these be replaced with something like `constructor <;> simpa`? (don't have an enviroment to test this right now)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+  intros y h
+  apply Acc.intro y
+  assumption
+
+/- If `MultisetRedLT` is well-founded, then its transitive closure `MultisetTransLT` is also
+well-founded. -/
+lemma transLT_wf [DecidableEq α] [LT α]
+    (h : WellFounded (MultisetRedLT : Multiset α → Multiset α → Prop)) :
+    WellFounded (MultisetTransLT : Multiset α → Multiset α → Prop) := by
+  unfold MultisetTransLT
```
goens, line 205: is this necessary? (would the `apply` below not work without it)?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+      have y_in_Y1 : y ∈ Y1 := by
+        have : Y1 - X2 ≤ Y1 := by
+          simp_all [tsub_le_iff_right, le_add_iff_nonneg_right,
+        zero_le]
+          simp only [Multiset.zero_le]
+        apply Multiset.mem_of_le
+        exact this
+        exact y_in
+      let _ := h y_in_Y1
+      aesop
```
goens, line 268: `aesop?`

goens (commented):
It looks like some proofs can be golfed a little bit more, and I'm not sure about the namings/namespaces (might be worth reading the [naming conventions](https://leanprover-community.github.io/contribute/naming.html) for that), that's the best feedback I can give so far (before a maintainer looks at it)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+
+In this file we define the _Dershowitz-Manna ordering_ on multisets.
+We prove that, given a well-founded partial order on the underlying set,
+the Dershowitz-Manna ordering defined over multisets is also well-founded.
+
+## Main results
+
+- `MultisetDMLT` :  the standard definition fo the `Dershowitz-Manna ordering`.
+- `dm_wf` :       the main theorem about the `Dershowitz-Manna ordering` being well-founded.
+- `TransLT_eq_DMLT` : two definitions of the `Dershowitz-Manna ordering` are equivalent.
```
YaelDillies, line 21: The alignment looks off 😅

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: Yes!

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: You are right. It is not necessary;-)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,409 @@
...
+
+open Relation
+
+/-- MultisetTransLT is the transitive closure of MultisetRedLT and is equivalent to MultisetDMLT
+    (proved later). -/
+def MultisetTransLT [DecidableEq α] [LT α] : Multiset α → Multiset α → Prop :=
+    TransGen MultisetRedLT
+
+/-- A shorthand notation: AccM defines the accessibility relation given MultisetRedLT. -/
+def AccM [DecidableEq α] [Preorder α] : Multiset α → Prop := Acc MultisetRedLT
```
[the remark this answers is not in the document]
haitian-yuki: You are right! The definition is redundant.

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: Yes, you are right! We changed it into a def now.

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: Sure! We have added the namespace 'Multiset' now;-)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+import Mathlib.Data.Multiset.Basic
+import Mathlib.Logic.Relation
+
+/-!
+# Dershowitz-Manna ordering
+
+In this file we define the _Dershowitz-Manna ordering_ on multisets.
+We prove that, given a well-founded partial order on the underlying set,
+the Dershowitz-Manna ordering defined over multisets is also well-founded.
+
```
YaelDillies, line 16: Can you quickly explain the definition of the order here?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+  Our code here is inspired by their formalization and the theorem is called `mOrd_wf` in the file
+  [MultisetList.v](https://github.com/fblanqui/color/blob/1.8.5/Util/Multiset/MultisetOrder.v).
+
+-/
+
+namespace Multiset
+
+variable {α : Type*}
+
+/-- The standard Dershowitz–Manna ordering: -/
```
YaelDillies, line 38: Did you forget to write something here?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+def DMLT [Preorder α] (M N : Multiset α) : Prop :=
+  ∃ (X Y Z : Multiset α),
+      Z ≠ ∅
+    ∧ M = X + Y
+    ∧ N = X + Z
+    ∧ (∀ y ∈ Y, ∃ z ∈ Z, y < z)
+
+/-- A special case of DMLT. The transitive closure of it is used to define
+    an equivalent (proved later) version of the ordering. -/
```
YaelDillies, line 47: We don't indent docstrings
```suggestion
an equivalent (proved later) version of the ordering. -/
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+def DMLT_singleton [LT α] (M N : Multiset α) : Prop :=
```
YaelDillies, line 48: Prop-valued things should be UpperCamelCased
```suggestion
def DMLTSingleton [LT α] (M N : Multiset α) : Prop :=
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+/-- A special case of DMLT. -/
+theorem dmlt_of_DMLT_singleton [Preorder α] (M N : Multiset α)
+    (h : DMLT_singleton M N) : DMLT M N := by
+  rcases h with ⟨X, Y, a, M_def, N_def, ys_lt_a⟩
+  use X, Y, {a}, by simp, M_def, N_def
+  · simpa
+
+lemma DMLT_singleton_insert [DecidableEq α] [LT α] {a : α} {M N : Multiset α}
+    (h : DMLT_singleton N (a ::ₘ M)) :
+    ∃ M', N = (a ::ₘ M') ∧ (DMLT_singleton M' M) ∨ (N = M + M') ∧ (∀ x : α, x ∈ M' → x < a) := by
```
YaelDillies, line 70:
```suggestion
    ∃ M', N = a ::ₘ M' ∧ DMLT_singleton M' M ∨ N = M + M' ∧ ∀ x ∈ M', x < a := by
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+        have this0: M = X + {a0} - {a} := by
+          rw [← h0, sub_singleton, erase_cons_head]
+        have a0M: a0 ∈ M := by
+          rw [this0, sub_singleton, mem_erase_of_ne]
+          rw [mem_add, mem_singleton]
+          · apply Or.inr
+            rfl
+          · exact fun h ↦ hyp (Eq.symm h)
+        rw [add_comm]
+        simp_all [singleton_add]
```
YaelDillies, line 115: Try to follow the line of sight rule. It says that the main branch of your code should be less indented
```suggestion
  exists (Y + (M - {a0}))
  left
  constructor
  · rw [h1]
    have : X = (M - {a0} + {a}) := by
      rw [add_comm, singleton_add] at *
      ext b
      simp only [ext, count_cons] at h0
      by_cases h : b = a
      · have := h0 b
        simp_all only [ite_true, ite_false, add_zero, sub_singleton, count_cons_self, ne_eq,
          not_false_eq_true, count_erase_of_ne]
      have := h0 b
      simp [sub_singleton]
      simp_all only [↓reduceIte, add_zero, ne_eq, not_false_eq_true, count_cons_of_ne]
      split at this
      next h_1 =>
        simp_all only [count_erase_self, add_tsub_cancel_right]
      next h_1 => simp_all only [add_zero, ne_eq, not_false_eq_true, count_erase_of_ne]
    subst this
    rw [add_comm]
    nth_rewrite 2 [add_comm]
    rw [singleton_add, add_cons]
  · unfold DMLT_singleton
    refine ⟨M - {a0}, Y, a0, ?_, ?_, h2⟩
    · change Y + (M - {a0}) = (M - {a0}) + Y
      rw [add_comm]
    · change M = M - {a0} + {a0}
      have this0 : M = X + {a0} - {a} := by
        rw [← h0, sub_singleton, erase_cons_head]
      have a0M : a0 ∈ M := by
        rw [this0, sub_singleton, mem_erase_of_ne]
        rw [mem_add, mem_singleton]
        · apply Or.inr
          rfl
        · exact fun h ↦ hyp (Eq.symm h)
      rw [add_comm]
      simp_all [singleton_add]
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+      (M = X + Y)
+    ∧ (N = X + {a})
+    ∧ (∀ y, y ∈ Y → y < a)
+
+open Relation
+
+/-- The transitive closure of DMLT_singleton and is equivalent to DMLT
+    (proved later). -/
+def TransLT [LT α] : Multiset α → Multiset α → Prop :=
+    TransGen DMLT_singleton
```
YaelDillies, line 59:
```suggestion
def TransLT [LT α] : Multiset α → Multiset α → Prop :=
  TransGen DMLT_singleton
```
or
```suggestion
def TransLT [LT α] : Multiset α → Multiset α → Prop := TransGen DMLT_singleton
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,405 @@
...
+    ∧ N = X + Z
+    ∧ (∀ y ∈ Y, ∃ z ∈ Z, y < z)
+
+/-- A special case of DMLT. The transitive closure of it is used to define
+    an equivalent (proved later) version of the ordering. -/
+def DMLT_singleton [LT α] (M N : Multiset α) : Prop :=
+  ∃ (X Y : Multiset α) (a : α) ,
+      (M = X + Y)
+    ∧ (N = X + {a})
+    ∧ (∀ y, y ∈ Y → y < a)
```
YaelDillies, line 52:
```suggestion
  ∃ X Y a,
      M = X + Y
    ∧ N = X + {a}
    ∧ ∀ y ∈ Y, y < a
```

YaelDillies (commented):
Thanks for the PR!

Am I right in saying that `DMLT_singleton` and `TransLT` are auxiliary to the proof? If so, can you make them as well as all lemmas about them `private`?

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: No. I was intended to show the definition comes after this. But now I removed the colon.

YaelDillies:
I mean replacing `def`/`lemma` by `private def`/`private lemma`

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -10,9 +10,12 @@ import Mathlib.Logic.Relation
...
 # Dershowitz-Manna ordering
 
-In this file we define the _Dershowitz-Manna ordering_ on multisets.
-We prove that, given a well-founded partial order on the underlying set,
-the Dershowitz-Manna ordering defined over multisets is also well-founded.
+In this file we define the _Dershowitz-Manna ordering_ on multisets. Specifically, for two multisets
+M and N over an underlying set S, M is smaller than N in the Dershowitz-Manna ordering if M can be
+obtained from N by replacing one or more elements in N by some finite number of elements from S,
+each of which is smaller (in the underling ordering over S) than one of the replaced elements from
+N. We prove that, given a well-founded partial order on the underlying set, the Dershowitz-Manna
```
YaelDillies, line 17:
```suggestion
`N`. We prove that, given a well-founded partial order on the underlying set, the Dershowitz-Manna
```
and everywhere else

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -10,9 +10,12 @@ import Mathlib.Logic.Relation
 /-!
 # Dershowitz-Manna ordering
 
-In this file we define the _Dershowitz-Manna ordering_ on multisets.
-We prove that, given a well-founded partial order on the underlying set,
-the Dershowitz-Manna ordering defined over multisets is also well-founded.
+In this file we define the _Dershowitz-Manna ordering_ on multisets. Specifically, for two multisets
+M and N over an underlying set S, M is smaller than N in the Dershowitz-Manna ordering if M can be
```
YaelDillies, line 14:
```suggestion
M and N in a partial order, M is smaller than N in the Dershowitz-Manna ordering if M can be
```
I don't think it's worth mentioning the underlying type, but it's definitely worth mentioning that it's ordered

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -35,7 +38,7 @@ namespace Multiset
 
 variable {α : Type*}
 
-/-- The standard Dershowitz–Manna ordering: -/
+/-- The standard Dershowitz–Manna ordering. -/
 def DMLT [Preorder α] (M N : Multiset α) : Prop :=
```
YaelDillies, line 42: This abbreviation seems a bit obscure. What about
```suggestion
def IsDershowitzMannaLT [Preorder α] (M N : Multiset α) : Prop :=
```

haitian-yuki: Sounds good! Thanks:-)

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,411 @@
...
+def IsDershowitzMannaLT [Preorder α] (M N : Multiset α) : Prop :=
+  ∃ (X Y Z : Multiset α),
+      Z ≠ ∅
+    ∧ M = X + Y
+    ∧ N = X + Z
+    ∧ (∀ y ∈ Y, ∃ z ∈ Z, y < z)
+
+/-- A special case of `IsDershowitzMannaLT`. The transitive closure of it is used to define
+an equivalent (proved later) version of the ordering. -/
+private def DMLTSingleton [LT α] (M N : Multiset α) : Prop :=
```
YaelDillies, line 52: Even if it's auxiliary, maybe we can give it a meaningful name as well?
```suggestion
private def OneStepLT [LT α] (M N : Multiset α) : Prop :=
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,411 @@
...
+            have : X + newZ + {z} = X + (newZ + {z}) := by apply add_assoc
+            rw [this, newZ_z_Z]
+            exact N_def
+          · unfold_let f; intro z z_in; simp at z_in; exact z_in.2
+        apply TransGen.single
+        exact this
+
+/-- `TransLT` and `IsDershowitzMannaLT` are equivalent. -/
+private lemma transLT_eq_dmLT [DecidableEq α] [Preorder α]
+    [DecidableRel (fun (x : α) (y: α) => x < y)] :
```
YaelDillies, line 381:
```suggestion
    [@DecidableRel α (· < ·)] :
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,411 @@
...
+      use W, U, {y}
+      simp_all
+    case tail _ aih bih =>
+      apply dmlt_trans _ _ _ _ bih
+      apply dmlt_of_DMLT_singleton
+      assumption
+  · apply transLT_of_dmLT
+
+/-- The desired theorem: If `LT.lt` is well-founded, then `IsDershowitzMannaLT` is well-founded. -/
+theorem IsDershowitzMannaLT.wf [DecidableEq α] [Preorder α]
```
YaelDillies, line 400:
```suggestion
theorem wellFounded_isDershowitzMannaLT [DecidableEq α] [Preorder α]
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,411 @@
...
+
+* [CoLoR](https://github.com/fblanqui/color), a Coq library on rewriting theory and termination.
+  Our code here is inspired by their formalization and the theorem is called `mOrd_wf` in the file
+  [MultisetList.v](https://github.com/fblanqui/color/blob/1.8.5/Util/Multiset/MultisetOrder.v).
+
+-/
+
+namespace Multiset
+
+variable {α : Type*}
```
YaelDillies, line 40:
```suggestion
variable {α : Type*} [Preorder α]
```
so that you don't have to add it to every single statement

haitian-yuki: Aha, that is helpful!

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,410 @@
...
+    ∧ ∀ y ∈ Y, y < a
+
+open Relation
+
+/-- The transitive closure of `OneStepLT` and is equivalent to `IsDershowitzMannaLT`
+(proved later). -/
+private def TransLT [LT α] : Multiset α → Multiset α → Prop := TransGen OneStepLT
+
+/-- A special case of `IsDershowitzMannaLT`. -/
+private lemma dmlt_of_DMLT_singleton (M N : Multiset α)
```
YaelDillies, line 65: Again, not crucial, because they are private, but can you rename your auxiliary lemmas to somewhat match the naming convention? Eg
```suggestion
private lemma isDershowitzMannaLT_of_oneStepLT {M N : Multiset α}
```

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,410 @@
...
+  intros a
+  apply acc_of_acc_lt
+  intros x _
+  apply wf_lt.induction x
+  intros y h
+  apply Acc.intro y
+  assumption
+
+/-- `IsDershowitzMannaLT` is transitive. -/
+private lemma dmlt_trans {α} [dec : DecidableEq α] [Preorder α] :
```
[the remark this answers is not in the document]
YaelDillies: It doesn't refer to any `private` definition, right? If so, then yes it might as well be public (and called the way you suggest)

YaelDillies:
I have just pushed a golf. I am running out of time today, but you should feel free to have a look for yourself at my commit and try golfing further using the tricks you will have learned. 😃

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -286,47 +228,28 @@ private lemma transLT_of_isDershowitzMannaLT [dec : DecidableEq α]
...
-          [empty_eq_zero, ne_eq, card_eq_zero, not_false_eq_true, ge_iff_le, card_erase_of_mem,
-          Nat.pred_eq_sub_one, tsub_pos_iff_lt]
-          apply this
-          · simp_all only [card_eq_zero, not_false_eq_true]
-          · simp_all only [not_false_eq_true]
-        rw [card_pos] at this
-        simp_all only [empty_eq_zero, ne_eq, not_false_eq_true]
+      have newZ_nonEmpty : newZ ≠ 0 := by
+        simp [newZ, ← sub_singleton, tsub_eq_zero_iff_le]
+        aesop
```
m4lvin, line 233: Thanks for the golfing! I am wondering here: Is it okay to leave `aesop` in for mathlib PRs and actually preferred in order to have shorter/readable proofs? In previous rounds we tried to always replace it by explicit proofs, but they can be lengthy :thinking:

YaelDillies: Oh no, using aesop is perfectly so long as performance is not too terrible

haitian-yuki:
> I have just pushed a golf. I am running out of time today, but you should feel free to have a look for yourself at my commit and try golfing further using the tricks you will have learned. 😃

Hi Thanks for showing some of the golfing tricks;-) @YaelDillies . I tried golfing a little bit more and maybe now it is in a better form. One thing that I am still unsure about is whether I am allowed to use aesop in the proof as long as it doesn't show `time run out` thing? Is that what you meant by `so long as performance is not too terrible`? Thanks!

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,294 @@
...
+private lemma acc_cons_of_acc (a : α)
+    (H : ∀ b, ∀ M, b < a → Acc OneStep M → Acc OneStep (b ::ₘ M)) :
+    ∀ M, Acc OneStep M → Acc OneStep (a ::ₘ M) := by
+  intros M h0
+  induction h0 with
+  | intro x wfH wfh2 =>
+    apply acc_cons
+    · simpa
+    · constructor; simpa
+    · simpa
```
Rida-Hamadani, line 157:
```suggestion
    · assumption
    · constructor; assumption
    · assumption
```
`simpa` is (kinda) `simp` followed by `assumption`, if `assumption` alone works then it is better. This might apply to other places too.

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,294 @@
...
+          let Y_lt_Z := Y_lt_Z y_in_Y
+          rcases Y_lt_Z with ⟨t, t_in_Z, y_lt_t⟩
+          use t
+          constructor
+          · by_cases t_in_newZ : t ∈ newZ
+            · exact t_in_newZ
+            · exfalso
+              have : t = z := by
+                have : Z = newZ + {z} := by
+                  unfold newZ
```
Rida-Hamadani, line 256:
```suggestion
```
This `unfold` is not needed, I suspect many other `unfold`s are not either, please check!

Rida-Hamadani (commented):
To answer the question regarding `aesop`, you can write `#count_heartbeats in` 1 line before a lemma to see how much of a difference using `aesop` changes, relevant scale for heartbeats is 200000.

Mathlib/Data/Multiset/DershowitzManna.lean
haitian-yuki: Yes, you are right! I replaced all the other replaceable `unfold ..` as well ;-)

YaelDillies:
Can you please merge master to get rid of the obscure CI error?

Mathlib/Data/Multiset/DershowitzManna.lean
```diff
@@ -0,0 +1,292 @@
...
+      Z ≠ ∅
+    ∧ M = X + Y
+    ∧ N = X + Z
+    ∧ ∀ y ∈ Y, ∃ z ∈ Z, y < z
+
+/-- `IsDershowitzMannaLT` is transitive. -/
+lemma IsDershowitzMannaLT.trans :
+    IsDershowitzMannaLT M N → IsDershowitzMannaLT N P → IsDershowitzMannaLT M P := by
+  classical
+  rintro ⟨X₁, Y₁, Z₁, hZ₁, rfl, rfl, hYZ₁⟩ ⟨X₂, Y₂, Z₂, hZ₂, hXZXY, rfl, hYZ₂⟩
```
m4lvin, line 56: The `hZ₁` here is unused and can be replaced with `_`. For some reason when copying the file into another project (where I use a slightly older Lean version) the linter warned about this, but did not do so here in CI :thinking:.

YaelDillies: You can even replace it with `-` to make sure it's not even introduced

YaelDillies:
I'm hoping to do a third golf pass at some point next week (I am completely submerged right now), then yes it should be good 😄

YaelDillies:
Sorry for the delay, but here's the final golf. Thank you for your patience!

maintainer merge

kim-em:
There's still an unresolved review comment [above](https://github.com/leanprover-community/mathlib4/pull/14411#discussion_r1835534865).

kim-em:
bors merge

bryangingechen:
bors r+
bors d+

PR #14413, opened 2024-07-04
[Merged by Bors] - feat(Algebra/Category/Ring/Colimits): category of (possibly non-commutative ring) has colimits

jjaassoonn:
We already have that category of commutative rings has colimits, the same construction works for category of rings.

---

### DIFF
--- a/Mathlib/Algebra/Category/Ring/Colimits.lean
+++ b/Mathlib/Algebra/Category/Ring/Colimits.lean
@@ -28,16 +28,16 @@ open CategoryTheory.Limits
 namespace RingCat.Colimits
 
 /-!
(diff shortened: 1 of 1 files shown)


erdOne:
If it is the same construction, can you show that `CommRingCat => RingCat` creates all colimits?

jjaassoonn:
> If it is the same construction, can you show that `CommRingCat => RingCat` creates all colimits?

I just realized I don't understand the maths proof:
We need that the colimit of commutative ring viewed as objects of `RingCat` is still a commutative ring. I am not sure if this true in general.

erdOne:
Ah yes you are right. 
maintainer merge

semorrison:
bors merge

PR #8612, opened 2023-11-24
[Merged by Bors] - feat(Data/Nat/Choose): Lucas' theorem

grhkm21:
This PR proves Lucas' theorem, which states that $\binom{n}{k} \equiv \prod_i \binom{\lfloor n / p^i \rfloor \pmod{p}}{\lfloor k / p^i \rfloor \pmod{p}} \equiv \prod_i \binom{n_i}{k_i} \pmod{p}$, where $n_i$ and $k_i$ are the $i$-th base-$p$ digits of $n$ and $k$ respectively. This is proven by looking at Freshman's dream closely + induction

PR #14416, opened 2024-07-04
[Merged by Bors] - refactor(CharP): make variables explicit + refactor

grhkm21:
Follows up on Yael's comment in #8612 by making `n` (in `x ^ (p ^ n)`) explicit.

---
- [x] depends on: #8612 Lucas' Theorem

CI should fail in places where the theorems are used, but I will fix them using the CI results.


grhkm21:
Only the final commit (36th) is from this PR, the rest is from #8612.

YaelDillies:
Why are you making this depend on #8612 and not the other way around?

grhkm21:
Because that one is tagged maintainer-merge so I thought I shouldn't touch it. Oh also the actual dependency part is I also refactor `add_pow_eq_add_pow_mod_mul_pow_add_pow_div`, if that's what you're asking

YaelDillies:
Oh, feel free to keep on touching PRs after they were maintainer merged. We're going to change the workflow soon to make it clearer that `maintainer-merge` is not a once and for all state.

Mathlib/Algebra/CharP/Basic.lean
```diff
@@ -54,92 +54,91 @@ end Commute
 
 section CommSemiring
 
-variable [CommSemiring R] {p : ℕ} {x y : R}
+variable [CommSemiring R] {p : ℕ} (x y : R) (n : ℕ)
```
YaelDillies, line 57:
```suggestion
variable [CommSemiring R] {p : ℕ} (hp : p.Prime) (x y : R) (n : ℕ)
include hp
```
so that `hp` stays before `x y n`

grhkm21: Oh yes, thank you, `include hp` works well. I also changed the section above.

YaelDillies (approved):
Thanks! 🚀

maintainer merge

Mathlib/Algebra/CharP/Basic.lean
```diff
@@ -21,9 +21,10 @@ variable {R : Type*}
 
 namespace Commute
 
-variable [Semiring R] {p : ℕ} {x y : R}
+variable [Semiring R] {p : ℕ} (hp : p.Prime) {x y : R}
+include hp
```
j-loreaux, line 25: I think even `section`ed or `namespace`d `include`s are advised against, in favor of `include ... in`, but it's possible I'm wrong here. Admittedly the sections are short, so maybe it's fine.

Mathlib/Algebra/CharP/Basic.lean
```diff
@@ -37,109 +38,109 @@ protected theorem add_pow_prime_pow_eq (hp : p.Prime) (h : Commute x y) (n : ℕ
...
     ∃ r, (x + y) ^ p = x ^ p + y ^ p + p * r :=
   ⟨_, h.add_pow_prime_eq hp⟩
 
 end Commute
 
 section CommSemiring
 
-variable [CommSemiring R] {p : ℕ} {x y : R}
+variable [CommSemiring R] {p : ℕ} (hp : p.Prime) (x y : R) (n : ℕ)
+include hp
```
j-loreaux, line 59: ditto

Mathlib/Algebra/CharP/Basic.lean
j-loreaux: I checked in the reviewer stream on Zulip, and indeed I think we want to only have `include ... in`, so please either change it to that, or else back to explicitly writing the hypothesis in each statement.

YaelDillies: I don't think there is such a consensus, though?

vihdzp: The reason using `include` like this is advised against is that it's really easy to accidentally include a hypothesis in a theorem that doesn't need it.

vihdzp:
Are you planning to return to this PR? If not I can take this over.

Mathlib/Algebra/CharP/Basic.lean
YaelDillies: Yes but all theorems here are about that hypothesis! And there is a linter against unused hypotheses anyway

YaelDillies:
I disagree with the need to remove the `include`.

maintainer merge

fpvandoorn:
We don't have a strong policy against `include` (yet?) and I think an `include` directly after the variable command can be fine in certain cases (if it's not omitted later).

bors merge

PR #14417, opened 2024-07-04
[Merged by Bors] - fix: handle duplicate constants in shake

digama0:
Fixes an issue [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/check.20for.20unused.20imports.20doesn't.20stop/near/449087555), due to constants being declared multiple times (which was not previously legal).

---

### DIFF
--- a/Mathlib.lean
+++ b/Mathlib.lean
@@ -1,3 +1,4 @@
+import Mathlib.A
 import Mathlib.Algebra.AddConstMap.Basic
 import Mathlib.Algebra.AddTorsor
 import Mathlib.Algebra.Algebra.Basic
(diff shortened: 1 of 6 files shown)


fpvandoorn:
If not too hard, a test would be nice.

bors d+

digama0:
@fpvandoorn We don't have any test harness setup for shake, and I'm not sure how one would do it. I can do some spurious changes here as a test, though.

digama0:
Test [failed successfully](https://github.com/leanprover-community/mathlib4/actions/runs/9882804958/job/27296364113)

semorrison:
bors merge

PR #14419, opened 2024-07-04
[Merged by Bors] - feat(Logic/Basic): add `simp` lemma for `@Exists.choose _ (· = a) _`

urkud:
and for the symmetric case.

---


eric-wieser:
!bench

eric-wieser:
bors d+

Let's check this doesn't fire unexpectedly often and impact performance (and if it does, let's have it without `simp`)

urkud:
No significant changes, so merging.
bors merge

PR #14425, opened 2024-07-04
[Merged by Bors] - chore: Golf `Finset.exists_subsuperset_card_eq`

YaelDillies:
Thanks Bhavik!

---

### DIFF
--- a/Mathlib/Data/Finset/Card.lean
+++ b/Mathlib/Data/Finset/Card.lean
@@ -623,15 +623,11 @@ set `u` of size `n` which is both a superset of `s` and a subset of `t`. -/
 lemma exists_subsuperset_card_eq (hst : s ⊆ t) (hsn : s.card ≤ n) (hnt : n ≤ t.card) :
     ∃ u, s ⊆ u ∧ u ⊆ t ∧ card u = n := by
   classical
-  obtain ⟨k, hk⟩ := Nat.le.dest hnt
-  clear hnt
-  induction' k with k ih generalizing t
-  · exact ⟨t, hst, Subset.rfl, hk.symm⟩
-  obtain ⟨a, ha⟩ : (t \ s).Nonempty := by rw [← card_pos, card_sdiff hst]; omega
-  replace hst : s ⊆ t.erase a := subset_erase.2 ⟨hst, (mem_sdiff.1 ha).2⟩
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Finset/Card.lean
```diff
@@ -623,15 +623,11 @@ set `u` of size `n` which is both a superset of `s` and a subset of `t`. -/
...
-  obtain ⟨k, hk⟩ := Nat.le.dest hnt
-  clear hnt
-  induction' k with k ih generalizing t
-  · exact ⟨t, hst, Subset.rfl, hk.symm⟩
-  obtain ⟨a, ha⟩ : (t \ s).Nonempty := by rw [← card_pos, card_sdiff hst]; omega
-  replace hst : s ⊆ t.erase a := subset_erase.2 ⟨hst, (mem_sdiff.1 ha).2⟩
-  replace hk : n + k = card (erase t a) := by rw [card_erase_of_mem (mem_sdiff.1 ha).1]; omega
-  obtain ⟨u, hsu, hut, hu⟩ := ih hst hk
-  exact ⟨u, hsu, hut.trans (erase_subset ..), hu⟩
+  refine Nat.decreasingInduction' ?_ hnt ⟨t, by simp [hst]⟩
```
eric-wieser, line 626: This bugged me, so I've made a start on https://github.com/leanprover-community/mathlib4/pull/14431

YaelDillies: Nerdsniping Eric is the best kind of nerdsnipe 😝

rwst (approved):
LGTM

Mathlib/Data/Finset/Card.lean
b-mehta: This was the best possible outcome of this golf :D

b-mehta (approved):
lgtm but I wrote it, so someone else should review too.

eric-wieser:
I'd prefer to block this on #14431, to motivate review there once I finish up in an hour or so

eric-wieser (approved):
Nevermind, I can't actually make `decreasingInduction'` work with `induction` after all, it's not an induction principle in the usual sense because the inductive step depends on the original value.

bors merge

PR #14428, opened 2024-07-04
[Merged by Bors] - feat(AlgebraicGeometry/Limits): Spec preserves finite coproducts

erdOne:
---

- [x] depends on: #14429


Mathlib/AlgebraicGeometry/OpenImmersion.lean
```diff
@@ -186,6 +186,24 @@ instance basic_open_isOpenImmersion {R : CommRingCat.{u}} (f : R) :
     exact Spec_map_localization_isIso R (Submonoid.powers f) x
 #align algebraic_geometry.Scheme.basic_open_IsOpenImmersion AlgebraicGeometry.Scheme.basic_open_isOpenImmersion
 
+instance {R} [CommRing R] (f : R) :
+    IsOpenImmersion (Spec.map (CommRingCat.ofHom (algebraMap R (Localization.Away f)))) :=
+  basic_open_isOpenImmersion (R := .of R) f
+
+lemma IsOpenImmersion.of_isLocalization {R S} [CommRing R] [CommRing S]
+    [Algebra R S] (f : R) [IsLocalization.Away f S] :
+    IsOpenImmersion (Spec.map (CommRingCat.ofHom (algebraMap R S))) := by
```
alreadydone, line 195: Haha, I was about to ask whether we have this :) A challenge (which I don't need) is to prove the same for localization at a f.g. submonoid.

erdOne: I don't think it is that hard actually. The same proof shows that the stalks are isomorphic and the underlying map is a topological embedding. To show that the image is open, we just need to show that it is the intersection of the basic open of the generators, which can't be that hard in lean right...?

alreadydone: Yeah probably not too hard. I was thinking about using induction and transitivity of open immersion and of localization.

jcommelin (commented):
Thanks :tada:

bors merge

erdOne:
good to go now.

mattrobball:
!bench

erdOne:
!bench

erdOne:
!bench

erdOne:
I think this is the best I can do. @mattrobball any thoughts?

mattrobball:
- Did you check there are no particular pain points (eg a tactic or instance synthesis that is mainly responsible)?
- Did the file get 40% longer?

erdOne:
1. There are two defs that aren't instant but the slowness seems to be spread evenly through tactics.
2. The file length is almost 2x
So I think it's fine?

mattrobball:
Yeah

mattrobball:
bors merge

PR #14430, opened 2024-07-04
[Merged by Bors] - refactor(AlgebraicGeometry): Rework `Morphisms/Basic` to phase away from TFAEs

erdOne:
Previously the contents of `Morphism/Basic` were all implementation detail that needed to be copied over to each new morphism class. In this PR we clean up the file and promote it into a proper interface of the API. We also phase away from TFAEs in favor of easier to use iff lemmas. We introduce the following two interfaces:

## `IsLocalAtTarget`

- `AlgebraicGeometry.IsLocalAtTarget`: We say that `IsLocalAtTarget P` for
`P : MorphismProperty Scheme` if
1. `P` respects isomorphisms.
2. If `P` holds for `f : X ⟶ Y`, then `P` holds for `f ∣_ U` for any open `U` of `Y`.
3. If `P` holds for `f ∣_ U` for an open cover `U` of `Y`, then `P` holds for `f`.

For a morphism property `P` local at the target and `f : X ⟶ Y`, we provide these API lemmas:

- `AlgebraicGeometry.IsLocalAtTarget.of_isPullback`:
    `P` is preserved under pullback along open immersions.
- `AlgebraicGeometry.IsLocalAtTarget.restrict`:
    `P f → P (f ∣_ U)` for an open `U` of `Y`.
- `AlgebraicGeometry.IsLocalAtTarget.iff_of_iSup_eq_top`:
    `P f ↔ ∀ i, P (f ∣_ U i)` for a family `U i` of open sets covering `Y`.
- `AlgebraicGeometry.IsLocalAtTarget.iff_of_openCover`:
    `P f ↔ ∀ i, P (𝒰.pullbackHom f i)` for `𝒰 : Y.openCover`.

## `HasAffineProperty`

- `AlgebraicGeometry.HasAffineProperty`:
  `HasAffineProperty P Q` is a type class asserting that `P` is local at the target,
  and over affine schemes, it is equivalent to `Q : AffineTargetMorphismProperty`.

For `HasAffineProperty P Q` and `f : X ⟶ Y`, we provide these API lemmas:

- `AlgebraicGeometry.HasAffineProperty.of_isPullback`:
    `P` is preserved under pullback along open immersions from affine schemes.
- `AlgebraicGeometry.HasAffineProperty.restrict`:
    `P f → Q (f ∣_ U)` for affine `U` of `Y`.
- `AlgebraicGeometry.HasAffineProperty.iff_of_iSup_eq_top`:
    `P f ↔ ∀ i, Q (f ∣_ U i)` for a family `U i` of affine open sets covering `Y`.
- `AlgebraicGeometry.HasAffineProperty.iff_of_openCover`:
    `P f ↔ ∀ i, P (𝒰.pullbackHom f i)` for affine open covers `𝒰` of `Y`.
- `AlgebraicGeometry.HasAffineProperty.stableUnderBaseChange_mk`:
    If `Q` is stable under affine base change, then `P` is stable under arbitrary base change.

---


Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +206,256 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+  · introv H
+    rintro ⟨U, hU : IsAffineOpen U⟩; dsimp
+    haveI : IsAffine _ := hU.preimage_of_isIso e.hom
+    rw [morphismRestrict_comp, P.cancel_right_of_respectsIso]
+    exact H ⟨(Opens.map e.hom.val.base).obj U, hU.preimage_of_isIso e.hom⟩
+
+class HasAffineProperty (P : MorphismProperty Scheme) where
+  protected affineProperty : AffineTargetMorphismProperty
+  isLocal_affineProperty : affineProperty.IsLocal
+  protected eq_targetAffineLocally' : P = targetAffineLocally affineProperty
```
chrisflav, line 275: I think this is a very good idea.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +206,256 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
-    TFAE
-      [targetAffineLocally P f,
-        ∃ (𝒰 : Scheme.OpenCover.{u} Y) (_ : ∀ i, IsAffine (𝒰.obj i)),
-          ∀ i : 𝒰.J, P (pullback.snd : (𝒰.pullbackCover f).obj i ⟶ 𝒰.obj i),
-        ∀ (𝒰 : Scheme.OpenCover.{u} Y) [∀ i, IsAffine (𝒰.obj i)] (i : 𝒰.J),
-          P (pullback.snd : (𝒰.pullbackCover f).obj i ⟶ 𝒰.obj i),
-        ∀ {U : Scheme} (g : U ⟶ Y) [IsAffine U] [IsOpenImmersion g],
-          P (pullback.snd : pullback f g ⟶ U),
-        ∃ (ι : Type u) (U : ι → Opens Y.carrier) (_ : iSup U = ⊤) (hU' : ∀ i, IsAffineOpen (U i)),
-          ∀ i, @P _ _ (f ∣_ U i) (hU' i)] := by
```
chrisflav, line 230: I agree that it is more useful to have the equivalences as separate lemmas. Maybe this TFAE is still good as an overview of all characterizations though, but this might also just be a docstring.

Mathlib/AlgebraicGeometry/Morphisms/QuasiCompact.lean
```diff
@@ -151,113 +126,58 @@ theorem isCompact_basicOpen (X : Scheme) {U : Opens X} (hU : IsCompact (U : Set
...
+    CompactSpace X ↔ QuasiCompact (terminal.from X) := by
+  rw [HasAffineProperty.iff_of_isAffine (P := @QuasiCompact)]
 #align algebraic_geometry.compact_space_iff_quasi_compact AlgebraicGeometry.compactSpace_iff_quasiCompact
 
-theorem QuasiCompact.affine_openCover_iff {X Y : Scheme.{u}} (𝒰 : Scheme.OpenCover.{u} Y)
-    [∀ i, IsAffine (𝒰.obj i)] (f : X ⟶ Y) :
-    QuasiCompact f ↔ ∀ i, CompactSpace (pullback f (𝒰.map i) : _) :=
-  quasiCompact_eq_affineProperty.symm ▸ QuasiCompact.affineProperty_isLocal.affine_openCover_iff f 𝒰
-#align algebraic_geometry.quasi_compact.affine_open_cover_iff AlgebraicGeometry.QuasiCompact.affine_openCover_iff
-
```
chrisflav, line 230: Since all of the specialized auxiliary reduction lemmas are removed (which I think is very good), maybe we can add the new code (i.e. `rw [HasAffineProperty.iff_of_openCover (P := @QuasiCompact) 𝒰]`)  for a reduction like this in some docstring as an example for the usage of the `HasAffineProperty` API.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -68,16 +68,76 @@ noncomputable section
...
+-/
+class IsLocalAtTarget (P : MorphismProperty Scheme) : Prop where
+  /-- `P` respects isomorphisms. -/
+  respectsIso : P.RespectsIso
+  /-- If `P` holds for `f : X ⟶ Y`, then `P` holds for `f ∣_ U` for any `U`. -/
+  restrict : ∀ {X Y : Scheme} (f : X ⟶ Y) (U : Opens Y), P f → P (f ∣_ U)
+  /-- If `P` holds for `f ∣_ U` for an open cover `U` of `Y`, then `P` holds for `f`.  -/
+  of_sSup_eq_top :
+    ∀ {X Y : Scheme.{u}} (f : X ⟶ Y) (Us : Set (Opens Y)), sSup Us = ⊤ →
+      (∀ U ∈ Us, P (f ∣_ U)) → P f
```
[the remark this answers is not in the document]
erdOne: I also like the `OpenCover` version, but because of universe issue, we would need to show `OpenCover.{u} => restrict open => OpenCover.{v}` for both of these conditions.
If you think the `OpenCover` version is better I can change it.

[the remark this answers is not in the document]
chrisflav: I think I prefer the `OpenCover` version while adding the set-theoretic versions as alternative constructors.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+  · introv H
+    rintro ⟨U, hU : IsAffineOpen U⟩; dsimp
+    haveI : IsAffine _ := hU.preimage_of_isIso e.hom
+    rw [morphismRestrict_comp, P.cancel_right_of_respectsIso]
+    exact H ⟨(Opens.map e.hom.val.base).obj U, hU.preimage_of_isIso e.hom⟩
+
+/-- `IsLocalAtTarget P Q` is a type class asserting that `P` is local at target, and over affine
+  schemes, it is equivalent to `Q : AffineTargetMorphismProperty`. -/
+class HasAffineProperty (P : MorphismProperty Scheme)
+    (Q : outParam AffineTargetMorphismProperty) : Prop where
```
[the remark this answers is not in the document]
chrisflav: Is the `outParam` justified here? It might be reasonable to have multiple `HasAffineProperty P Q` for fixed `P` and varying `Q`. In this case the lemma `of_isLocalAtTarget` below could be an instance as well.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+variable (P : MorphismProperty Scheme) {Q} [HasAffineProperty P Q]
+variable {X Y : Scheme.{u}} {f : X ⟶ Y}
+
+instance (Q : AffineTargetMorphismProperty) [Q.IsLocal] :
+    HasAffineProperty (targetAffineLocally Q) Q :=
+  ⟨inferInstance, rfl⟩
+
+lemma eq_targetAffineLocally : P = targetAffineLocally Q := eq_targetAffineLocally'
+
+/-- Every property local at the target can be associated with a affine target property.
```
chrisflav, line 316:
```suggestion
/-- Every property local at the target can be associated with an affine target property.
```

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+    rw [morphismRestrict_comp, P.cancel_left_of_respectsIso]
+    exact H U
+  · introv H
+    rintro ⟨U, hU : IsAffineOpen U⟩; dsimp
+    haveI : IsAffine _ := hU.preimage_of_isIso e.hom
+    rw [morphismRestrict_comp, P.cancel_right_of_respectsIso]
+    exact H ⟨(Opens.map e.hom.val.base).obj U, hU.preimage_of_isIso e.hom⟩
+
+/-- `IsLocalAtTarget P Q` is a type class asserting that `P` is local at target, and over affine
+  schemes, it is equivalent to `Q : AffineTargetMorphismProperty`. -/
```
chrisflav, line 299: This seems to be a bit off, since `P` is not required to be local at the target in the definition. Maybe this is clearer?
```suggestion
/-- `HasAffineProperty P Q` is a type class asserting that over affine schemes `P` is equivalent to the local `Q : AffineTargetMorphismProperty`. -/
```

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+  exact forall_congr' fun i ↦ Q.arrow_mk_iso_iff
+    (morphismRestrictOpensRange f _)
+
+theorem morphsimRestrict (U : Opens Y) (H : P f) : P (f ∣_ U) := by
+  letI := isLocal_affineProperty P
+  rw [eq_targetAffineLocally P] at H ⊢
+  intros V
+  rw [Q.arrow_mk_iso_iff (morphismRestrictRestrict _ _ _)]
+  exact H ⟨_, V.2.image_of_isOpenImmersion _⟩
+
```
chrisflav, line 404: I don't think this lemma is necessary, as it can be obtained from `IsLocalAtTarget.restrict`.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+  of_basicOpenCover {_ Y} _ _ _ hs := IsLocalAtTarget.of_iSup_eq_top _
+    (((isAffineOpen_top Y).basicOpen_union_eq_self_iff _).mpr hs)
 
 /-- A `P : AffineTargetMorphismProperty` is stable under base change if `P` holds for `Y ⟶ S`
 implies that `P` holds for `X ×ₛ Y ⟶ X` with `X` and `S` affine schemes. -/
-def StableUnderBaseChange (P : AffineTargetMorphismProperty) : Prop :=
+def AffineTargetMorphismProperty.StableUnderBaseChange (P : AffineTargetMorphismProperty) : Prop :=
   ∀ ⦃X Y S : Scheme⦄ [IsAffine S] [IsAffine X] (f : X ⟶ S) (g : Y ⟶ S),
     P g → P (pullback.fst f g)
 #align algebraic_geometry.affine_target_morphism_property.stable_under_base_change AlgebraicGeometry.AffineTargetMorphismProperty.StableUnderBaseChange
```
chrisflav, line 265: Side note: Should this be defined in terms of `IsPullback` mimicking `MorphismProperty.StableUnderBaseChange`?

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+  rw [iff_of_openCover (P := P) (Scheme.openCoverOfIsIso.{0} (𝟙 Y))]
+  trans Q (pullback.snd f (𝟙 _))
+  · exact ⟨fun H => H PUnit.unit, fun H _ => H⟩
+  rw [← Category.comp_id (pullback.snd _ _), ← pullback.condition,
+    Q.cancel_left_of_respectsIso]
+#align algebraic_geometry.affine_target_morphism_property.is_local.affine_target_iff AlgebraicGeometry.HasAffineProperty.iff_of_isAffine
+
+instance (𝒰 : X.AffineOpenCover) (i) : IsAffine (𝒰.openCover.obj i) :=
+  inferInstanceAs (IsAffine (Spec (𝒰.obj i)))
+
```
chrisflav, line 418: Can you please move this to `Mathlib.AlgebraicGeometry.Cover.Open`?

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+      exact (opensRange_ιOpens _).symm
+    rw [← Q.cancel_left_of_respectsIso (𝒰.pullbackCoverAffineRefinementObj f _).inv,
+      𝒰.pullbackCoverAffineRefinementObj_inv_pullbackHom]
+    exact of_isPullback (.of_hasPullback _ _) this
+#align algebraic_geometry.affine_target_morphism_property.is_local.target_affine_locally_is_local AlgebraicGeometry.HasAffineProperty.instIsLocalAtTarget
+
+theorem pullback_fst_of_right (hP' : Q.StableUnderBaseChange)
+    {X Y S : Scheme} (f : X ⟶ S) (g : Y ⟶ S) [IsAffine S] (H : Q g) :
+    P (pullback.fst f g) := by
+  letI := isLocal_affineProperty P
```
chrisflav, line 442: Since this is not to be used outside this file: Can this be `private`?

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
-  have := hP.1
+  have : e.hom ≫ pullback.fst _ _ = X.affineCover.pullbackHom (pullback.fst _ _) i := by
+    simp [e, Scheme.OpenCover.pullbackHom]
+  rw [← this, Q.cancel_left_of_respectsIso]
+  apply hP' _ _
+  exact H
+#align algebraic_geometry.affine_target_morphism_property.is_local.target_affine_locally_pullback_fst_of_right_of_stable_under_base_change AlgebraicGeometry.HasAffineProperty.pullback_fst_of_right
+
+theorem stableUnderBaseChange_mk (hP' : Q.StableUnderBaseChange) :
+    P.StableUnderBaseChange :=
```
chrisflav, line 454:
```suggestion
theorem stableUnderBaseChange_mk (hQ : Q.StableUnderBaseChange) :
    P.StableUnderBaseChange :=
```
Maybe `stableUnderBaseChange_of` is a better name?

chrisflav (commented):
I had a look at the big file. Will review the rest tomorrow.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
erdOne: I changed it to `stableUnderBaseChange`.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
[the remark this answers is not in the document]
erdOne: So that by rewriting with API lemmas you get the `AffineTargetMorphismProperty` directly instead of some `MorphismProperty.affineProperty`. But in spirit, it is still a typeclass on `P`, so `Q` should be an `outParam`.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
 3. If `P` holds for `f ∣_ Y.basicOpen r` for all `r` in a spanning set of the global sections,
   then `P` holds for `f`.
 -/
-structure AffineTargetMorphismProperty.IsLocal (P : AffineTargetMorphismProperty) : Prop where
+class AffineTargetMorphismProperty.IsLocal (P : AffineTargetMorphismProperty) : Prop where
   /-- `P` as a morphism property respects isomorphisms -/
-  RespectsIso : P.toProperty.RespectsIso
+  respectsIso : P.toProperty.RespectsIso
```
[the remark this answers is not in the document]
erdOne: I don't think this will ever fire so I refrained from adding it.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -156,347 +233,248 @@ instance targetAffineLocally_respectsIso (P : AffineTargetMorphismProperty)
...
+    rintro ⟨U, hU : IsAffineOpen U⟩; dsimp
+    haveI : IsAffine _ := hU.preimage_of_isIso e.hom
+    rw [morphismRestrict_comp, P.cancel_right_of_respectsIso]
+    exact H ⟨(Opens.map e.hom.val.base).obj U, hU.preimage_of_isIso e.hom⟩
+
+/-- `IsLocalAtTarget P Q` is a type class asserting that `P` is local at target, and over affine
+  schemes, it is equivalent to `Q : AffineTargetMorphismProperty`. -/
+class HasAffineProperty (P : MorphismProperty Scheme)
+    (Q : outParam AffineTargetMorphismProperty) : Prop where
+  isLocal_affineProperty : Q.IsLocal
```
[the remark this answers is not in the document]
erdOne: I think `isLocal_affineProperty` is the right name according to the naming convention. Also this will never fire in practice so I don't think we should have `:= by infer_instance` as it gives worse error messages.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
```diff
@@ -68,22 +84,93 @@ noncomputable section
...
+      ∀ {X Y : Scheme.{u}} (f : X ⟶ Y) {ι : Type u} (U : ι → Opens Y), iSup U = ⊤ →
+        (∀ i, P (f ∣_ U i)) → P f) :
+    IsLocalAtTarget P := by
+  refine ⟨inferInstance, fun {X Y} f 𝒰 ↦ ⟨?_, fun H ↦ of_sSup_eq_top f _ 𝒰.iSup_opensRange ?_⟩⟩
+  · exact fun H i ↦ (P.arrow_mk_iso_iff (morphismRestrictOpensRange f _)).mp (restrict _ _ H)
+  · exact fun i ↦ (P.arrow_mk_iso_iff (morphismRestrictOpensRange f _)).mpr (H i)
+
+variable {P} [hP : IsLocalAtTarget P]
+variable {X Y U V : Scheme.{u}} {f : X ⟶ Y} {g : U ⟶ Y} [IsOpenImmersion g] (𝒰 : Y.OpenCover)
+variable {iV : V ⟶ X} {f' : V ⟶ U} (h : IsPullback iV f' f g)
```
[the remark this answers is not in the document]
erdOne: I changed the names a bit.

Mathlib/AlgebraicGeometry/Morphisms/Basic.lean
erdOne: I changed the docstring.

Mathlib/AlgebraicGeometry/Morphisms/Constructors.lean
```diff
@@ -56,136 +56,157 @@ instance AffineTargetMorphismProperty.diagonal_respectsIso (P : AffineTargetMorp
...
+
+lemma _root_.CategoryTheory.IsPullback.of_id_fst {C} [Category C] {X Y : C} (f : X ⟶ Y) :
+    IsPullback (𝟙 _) f f (𝟙 _) := IsPullback.of_horiz_isIso ⟨by simp⟩
+
+lemma _root_.CategoryTheory.IsPullback.of_id_snd {C} [Category C] {X Y : C} (f : X ⟶ Y) :
+    IsPullback f (𝟙 _) (𝟙 _) f := IsPullback.of_vert_isIso ⟨by simp⟩
+
+theorem HasAffineProperty.of_affineProperty_diagonal
+    (P) {Q} [HasAffineProperty P Q] {X Y} {f : X ⟶ Y} [IsAffine Y] :
+    Q.diagonal f ↔ P.diagonal f := by
```
chrisflav, line 126:
```suggestion
theorem HasAffineProperty.affineProperty_diagonal_iff_diagonal
    (P) {Q} [HasAffineProperty P Q] {X Y} {f : X ⟶ Y} [IsAffine Y] :
    Q.diagonal f ↔ P.diagonal f := by
```

Mathlib/AlgebraicGeometry/Morphisms/Constructors.lean
```diff
@@ -56,136 +56,157 @@ instance AffineTargetMorphismProperty.diagonal_respectsIso (P : AffineTargetMorp
...
+    refine diagonal_of_openCover_diagonal_affineProperty (targetAffineLocally Q) f
+      (Y.openCoverOfSuprEqTop _ (((isAffineOpen_top Y).basicOpen_union_eq_self_iff _).mpr hs)) ?_
+    intro i
+    exact (Q.diagonal.arrow_mk_iso_iff
+      (morphismRestrictEq _ (by simp) ≪≫ morphismRestrictOpensRange _ _)).mp (hs' i)
+#align algebraic_geometry.affine_target_morphism_property.is_local.diagonal AlgebraicGeometry.HasAffineProperty.diagonal_affineProperty_isLocal
+
+theorem HasAffineProperty.targetAffineLocally_diagonal_affineProperty
+    (P) {Q} [HasAffineProperty P Q] :
+    targetAffineLocally Q.diagonal = P.diagonal := by
```
chrisflav, line 165: I think this should be `private` as it should only be used to construct `HasAffineProperty P.diagonal Q.diagonal`.

Mathlib/AlgebraicGeometry/Morphisms/Constructors.lean
```diff
@@ -56,136 +56,157 @@ instance AffineTargetMorphismProperty.diagonal_respectsIso (P : AffineTargetMorp
...
+    (pullback.fst (f := f) (g := 𝟙 Y)), pullback.condition, Category.comp_id] at hf
+  let 𝒰 := X.affineCover.pushforwardIso (inv (pullback.fst (f := f) (g := 𝟙 Y)))
+  have (i) : IsAffine (𝒰.obj i) := by dsimp [𝒰]; infer_instance
+  exact HasAffineProperty.diagonal_of_openCover P f (Scheme.openCoverOfIsIso (𝟙 _))
+    (fun _ ↦ 𝒰) (fun _ _ _ ↦ hf _ _)
+
+theorem HasAffineProperty.affineProperty_diagonal (P) {Q} [HasAffineProperty P Q] :
+    Q.diagonal = AffineTargetMorphismProperty.of P.diagonal := by
+  ext X Y f _
+  exact of_affineProperty_diagonal P
```
chrisflav, line 139: I think this is not needed as it follows from the `HasAffineProperty` instance below and the general `HasAffineProperty.iff` (if you want you can state the `[HasAffineProperty P Q] : Q = AffineTargetMorphismProperty.of P` separately from the `HasAffineProperty.iff` lemma).

Mathlib/AlgebraicGeometry/Morphisms/Constructors.lean
```diff
@@ -56,136 +56,157 @@ instance AffineTargetMorphismProperty.diagonal_respectsIso (P : AffineTargetMorp
...
+
+-- Doesn't build without the `IsAffine` instance but the linter complains
+@[nolint unusedHavesSuffices]
+instance HasAffineProperty.diagonal_affineProperty_isLocal
+    {Q : AffineTargetMorphismProperty} [Q.IsLocal] :
+    Q.diagonal.IsLocal where
+  respectsIso := inferInstance
+  to_basicOpen {X Y} _ f r hf :=
+    have : IsAffine (Y ∣_ᵤ Y.basicOpen r) := (isAffineOpen_top Y).basicOpen r
```
chrisflav, line 148:
```suggestion
instance HasAffineProperty.diagonal_affineProperty_isLocal
    {Q : AffineTargetMorphismProperty} [Q.IsLocal] :
    Q.diagonal.IsLocal where
  respectsIso := inferInstance
  to_basicOpen {X Y} _ f r hf :=
```
With the newly added global instance, it no longer complains.

Mathlib/AlgebraicGeometry/Morphisms/Constructors.lean
```diff
@@ -240,17 +261,15 @@ lemma topologically_propertyIsLocalAtTarget
...
     exact hP₂ f.val.base U.carrier hf
   · intro X Y f ι U hU hf
-    apply hP₃ f.val.base U hU
-    · exact f.val.base.continuous
-    · intro i
-      rw [← morphismRestrict_val_base]
-      exact hf i
+    apply hP₃ f.val.base U hU f.val.base.continuous fun i ↦ ?_
+    rw [← morphismRestrict_val_base]
+    exact hf i
```
chrisflav, line 272: I can't comment on unchanged code, but I think that the name of this lemma should be changed to `topologically_isLocalAtTarget`.

Mathlib/AlgebraicGeometry/Morphisms/FiniteType.lean
```diff
@@ -83,16 +83,9 @@ theorem LocallyOfFiniteType.source_openCover_iff {X Y : Scheme.{u}} (f : X ⟶ Y
   locallyOfFiniteType_eq.symm ▸ RingHom.finiteType_is_local.source_openCover_iff f 𝒰
 #align algebraic_geometry.locally_of_finite_type.source_open_cover_iff AlgebraicGeometry.LocallyOfFiniteType.source_openCover_iff
 
-theorem LocallyOfFiniteType.openCover_iff {X Y : Scheme.{u}} (f : X ⟶ Y)
-    (𝒰 : Scheme.OpenCover.{u} Y) :
-    LocallyOfFiniteType f ↔ ∀ i, LocallyOfFiniteType (pullback.snd f (𝒰.map i)) :=
-  locallyOfFiniteType_eq.symm ▸ RingHom.finiteType_is_local.is_local_affineLocally.openCover_iff f 𝒰
-#align algebraic_geometry.locally_of_finite_type.open_cover_iff AlgebraicGeometry.LocallyOfFiniteType.openCover_iff
-
```
chrisflav, line 91: Can we also get rid of `LocallyOfFiniteType.affine_openCover_iff` and `LocallyOfFiniteType.source_openCover_iff`? Maybe we can introduce some `HasAffineRingProperty` extending `HasAffineProperty` (this can and should of course go in a follow-up PR).

Mathlib/AlgebraicGeometry/Morphisms/QuasiCompact.lean
chrisflav: What did come out of this?

chrisflav (commented):
As I said earlier, I really like this change.

Mathlib/AlgebraicGeometry/Morphisms/FiniteType.lean
erdOne: Yes that's on my plan but probably next PR.

Mathlib/AlgebraicGeometry/Morphisms/QuasiCompact.lean
erdOne: I think the newly added docstring of the file is enough?

chrisflav (approved):
LGTM, thanks a lot!

erdOne:
Thanks for the swift review!

mattrobball:
!bench

jcommelin (commented):
Thanks :tada:

bors merge

mattrobball:
Let's wait for the benchmarking to finish. AG is a finicky area currently. 

bors r-

mattrobball:
Seems minor. I will see if I can track that down later. 

bors merge

PR #14431, opened 2024-07-04
[Merged by Bors] - feat: a replacement for `Nat.leRecOn'` that works with `induction`

eric-wieser:
Before this PR, we had:
* `theorem Nat.le.rec`: induct on `a ≤ b`, but using an annoying `a.le b` spelling
* `theorem Nat.le_induction`: induct on `a ≤ b`
* `def Nat.leRecOn'`, which is `Nat.le.rec` that works in `Sort*` but does not work with the `induction` keyword. There are no lemmas about this definition.
* `def Nat.leRecOn`, a weaker version of `Nat.leRecOn'` that also doesn't work with `induction`. There are lots of lemmas about this definition

This PR introduces `Nat.leRec`, which is a `Sort*`-valued version of `Nat.le.rec`, with a signature that coincides in the `Prop` case.
Where appropriate, the lemmas about `Nat.leRecOn` are recovered as special cases of lemmas about this new definition.
`Nat.leRecOn'` is then deprecated, as it is strictly less useful than `Nat.leRec`.

This also changes the signature of `Nat.decreasingInduction` to be usable with the `induction` tactic, with no deprecated alias since there are very few uses.

`Nat.decreasingInduction'` is left alone, as it is unclear to me how to state it in a way understood by `induction`.

---

For reviewers: what should the names of the inductive cases be?

For the base case, we could use `base`/`refl`/`self`.
For the inductive case, we could use `step`/`next`/`succ`.

Note that the existing principles are already inconsistent here.

### DIFF
--- a/Mathlib/Data/Nat/Defs.lean
+++ b/Mathlib/Data/Nat/Defs.lean
@@ -886,6 +886,9 @@ lemma leRecOn'_succ {n} {motive : (m : ℕ) → n ≤ m → Sort*}
     lhs
     rw [leRecOn', Or.by_cases, dif_pos h1]
 
+lemma leRecOn'_succ' {n} {motive : (m : ℕ) → n ≤ m → Sort*} (next self) :
+    (leRecOn' (motive := motive) next self (le_succ _)) = next _ self := by
+  rw [leRecOn'_succ, leRecOn'_self]
 /-- Recursion starting at a non-zero number: given a map `C k → C (k + 1)` for each `k`,
 there is a map from `C n` to each `C m`, `n ≤ m`. For a version where the assumption is only made
 when `k ≥ n`, see `Nat.leRecOn'`. -/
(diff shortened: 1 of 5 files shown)


YaelDillies:
> For the base case, we could use `base`/`refl`/`self`.
> For the inductive case, we could use `step`/`next`/`succ`.

What about `refl` and `le_succ_of_le`? Those are the names of the corresponding proofs of `a ≤ b`

Mathlib/Data/Nat/Defs.lean
```diff
@@ -854,52 +854,103 @@ lemma rec_add_one {C : ℕ → Sort*} (h0 : C 0) (h : ∀ n, C n → C (n + 1))
     Nat.rec (motive := C) h0 h 1 = h 0 h0 := rfl
 
 /-- Recursion starting at a non-zero number: given a map `C k → C (k+1)` for each `k ≥ n`,
-there is a map from `C n` to each `C m`, `n ≤ m`. -/
+there is a map from `C n` to each `C m`, `n ≤ m`.
+
+This is a version of `Nat.le.rec` that works for `Sort`. -/
```
nomeata, line 859: If compatibility with `indiction` is a feature, maybe mention that in the docstring? Maybe even with an example?

Mathlib/Data/Nat/Defs.lean
```diff
@@ -854,52 +854,103 @@ lemma rec_add_one {C : ℕ → Sort*} (h0 : C 0) (h : ∀ n, C n → C (n + 1))
...
+lemma leRec_self {n} {motive : (m : ℕ) → n ≤ m → Sort*}
+    (self : motive n le_rfl)
+    (next : ∀ ⦃k⦄ (h : n ≤ k), motive k h → motive (k + 1) (le_succ_of_le h)) :
+    (leRec (motive := motive) self next le_rfl : motive n le_rfl) = self := by
+  cases n <;> simp [leRec, Or.by_cases, dif_neg]
+
+@[simp]
+lemma leRec_succ {n} {motive : (m : ℕ) → n ≤ m → Sort*}
+    (next : ∀ ⦃k⦄ (h : n ≤ k), motive k h → motive (k + 1) (le_succ_of_le h))
+    (self : motive n le_rfl)
```
rosborn, line 885: Is there a reason why `self` is after `next` here?

eric-wieser: Nice catch, this got missed in my cleanup pass that chose the sensible order.

Mathlib/Data/Nat/Defs.lean
eric-wieser: Done

eric-wieser:
> What about `refl` and `le_succ_of_le`?

Done.

YaelDillies (approved):
I am reviewing from a phone, but it looks good

maintainer merge

eric-wieser:
maintainer merge

since Yael's comment seemed to be ignored

riccardobrasca:
Thanks!

bors merge

PR #14432, opened 2024-07-04
[Merged by Bors] - feat(ContinuousFunctionalCalculus): When `cfc` is applied to a function that maps zero to zero, it is equal to `cfcₙ`

dupuisf:
This PR shows that `cfcₙ f a = cfc f a` whenever `f` maps zero to zero (and the usual continuity conditions apply).

This required refactoring the proof that a unital CFC implies a non-unital one, since we need reuse the objects that were defined locally inside the proof.

---

### DIFF
--- a/Mathlib/Topology/ContinuousFunction/NonUnitalFunctionalCalculus.lean
+++ b/Mathlib/Topology/ContinuousFunction/NonUnitalFunctionalCalculus.lean
@@ -667,10 +667,10 @@ instance ContinuousFunctionalCalculus.toNonUnital : NonUnitalContinuousFunctiona
   predicate_zero := cfc_predicate_zero R
   exists_cfc_of_predicate _ ha :=
     ⟨cfcₙHom_of_cfcHom R ha,
-            closedEmbedding_cfcₙHom_of_cfcHom ha,
-            cfcHom_id ha,
-            cfcₙHom_of_cfcHom_map_quasispectrum ha,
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/ContinuousFunction/NonUnitalFunctionalCalculus.lean
```diff
@@ -600,57 +600,102 @@ variable [TopologicalRing R] [ContinuousStar R] [Ring A] [StarRing A] [Topologic
 variable [Algebra R A] [ContinuousFunctionalCalculus R p]
 variable [h_cpct : ∀ a : A, CompactSpace (spectrum R a)]
 
+variable (R) in
+/-- The non-unital functional calculus obtained by restricting a unital calculus to functions
+that map zero to zero. -/
```
j-loreaux, line 605:
```suggestion
/-- The non-unital continuous functional calculus obtained by restricting a unital calculus
to functions that map zero to zero. This is an auxiliary definition and is not
intended for use outside this file. The equality between the non-unital and unital
calculi in this case is encoded in the lemma `cfcₙ_eq_cfc`. -/
```

j-loreaux (commented):
This issue comes up a lot (i.e., where we want to build the functional calculus and then immediately prove that it's equal to the one we just constructed). I think we want a new tool for this, but that's an issue for a separate PR. That way we would avoid all these auxiliary declarations floating around.

j-loreaux:
bors merge

PR #14437, opened 2024-07-05
[Merged by Bors] - feat: `#min_imports in` command

adomani:
This PR introduces the `#min_imports in` command.

Writing `#min_imports in stx` scans the syntax `stx` to find a collection of minimal imports that should be sufficient for `stx` to make sense.

If `stx` is a command, then it also elaborates `stx`.  If `stx` produces a declaration, then
it also finds the imports implied by the declaration.

Unlike the related `#find_home`, this command takes into account notation and tactic information.

For instance, you can write
```lean
import Mathlib

#min_imports in
lemma hi (n : ℕ) : n = n := by extract_goal; rfl
-- import Mathlib.Tactic.ExtractGoal
-- import Mathlib.Tactic.Lemma
-- import Mathlib.Init.Data.Nat.Notation
```
and you can see that it notices `lemma`, `ℕ` and `extract_goal`.

---


grunweg:
This looks exciting :tada: So, just to clarify: can I run this on an entire file like #find_home?

adomani:
No, currently it is "declaration" based.

adomani:
Actually, I do not think that `#find_home` can be run on a file either: `#find_home` takes an identifier as input (which is legal also for `#min_imps`).  Moreover, `#min_imps` takes any `command` or `term` and does its thing.

So, you can write
```lean
#min_imps
lemma hi (n : ℕ) : n = n := by extract_goal; rfl
-- import Mathlib.Tactic.ExtractGoal
-- import Mathlib.Tactic.Lemma
-- import Mathlib.Init.Data.Nat.Notation
```
and you can see that it notices `lemma`, `ℕ` and `extract_goal`.

semorrison:
Happy to try this out in the wild. 

bors d+

adomani:
bors merge

kmill:
Something that occurred to me that would also take into account elaborators is doing the following:

1. save (and clear) the info trees with `Lean.Elab.getResetInfoTrees`
2. elaborate the command
3. recursively walk through the info trees collecting (1) the `elaborator` fields, (2) for every `Syntax` each of the `kind`s using your `getSyntaxNodeKinds`, and (3) for each TermInfo node, all the constants appearing in the `expr` and `expectedType?` fields.
4. try to resolve all of these names, and find which modules they come from, like the code you already have, to compute imports.
5. restore the info trees by prepending them to the new list of info trees.

adomani:
@kmill, I am happy that you mention this!  I had also thought of using the `InfoTrees`, but I still feel a little intimidated by them, so I resorted to the more "explicit" approach, thinking that it would have been easier.

However, seeing now the hoops that I am going through to retrieve the correct name for nameless instances, I am going back to thinking that the `InfoTrees` might be the way to go.

Currently, what I am really struggling with is this example:
```
import Mathlib.Data.Sym.Sym2.Init -- the actual minimal import
import Aesop.Frontend.Attribute   -- the import that `#min_imports in` suggests

import Mathlib.Tactic.MinImports

#min_imports in
@[aesop (rule_sets := [Sym2]) [safe [constructors, cases], norm]]
inductive Rel (α : Type) : α × α → α × α → Prop
  | refl (x y : α) : Rel _ (x, y) (x, y)
  | swap (x y : α) : Rel _ (x, y) (y, x)
```
The `aesop` `rule_sets` elude my command.

Do you think that `InfoTrees` would pick up on this import?

kmill:
Re InfoTree wrangling, since you're not going to do any meta calculations (you just want constant names), you can really just scan through all the Info nodes and ignore things like context updates.

Picking up things like aesop rulesets seems complicated. Maybe aesop could install some fake infotree nodes with references to rulesets just to help your tool? Or maybe these infotree nodes are already present since aesop might use the elaborator to resolve them? (Every elabTerm results in some infotree nodes.)

adomani:
Ok, I will experiment with `InfoTrees`.  I would like to understand them better anyway, since they would be useful for the "beginners linter" flagging `Nat.sub`: I want that linter to be aware of hypotheses `a \le b` whenever `b - a` is present and this requires "persisting" fvars across the whole syntax tree.  Or at least, this is what I think!

Also, there are very few `aesop` `rule_sets` throughout mathlib, and I am happy to live without the tool parsing them correctly.  Besides, after one file that has been adjusted by hand, every file that imports it no longer has the issue.

PR #14445, opened 2024-07-05
[Merged by Bors] - chore (Data.Fintype.Basic): de-simp `Fintype.univ_ofSubsingleton`

mattrobball:
In principle, `Fintype.univ_ofSubsingleton` could be a `simp` theorem but it applies to any occurence of `univ` and requires unification of the (possibly very complex) `Fintype` instances. It appears to not be needed in `simp`. So we remove the attribute. We get some performance gains as a benefit. 

---

See the [Zulip topic](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60simp.60.20vs.20subsingleton.20instances) for an example of it causing problems in the wild.

### DIFF
--- a/Mathlib/CategoryTheory/Preadditive/Mat.lean
+++ b/Mathlib/CategoryTheory/Preadditive/Mat.lean
@@ -341,7 +341,7 @@ def isoBiproductEmbedding (M : Mat_ C) : M ≅ ⨁ fun i => (embedding C).obj (M
     rw [Finset.sum_apply, Finset.sum_apply, Finset.sum_eq_single i]; rotate_left
     · intro b _ hb
       dsimp
-      simp only [Finset.sum_singleton]
+      erw [Finset.sum_singleton]
       rw [dif_neg hb.symm, zero_comp]
     · intro h
       simp at h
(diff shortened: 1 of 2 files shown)


mattrobball:
!bench

Mathlib/CategoryTheory/Preadditive/Mat.lean
```diff
@@ -341,7 +341,7 @@ def isoBiproductEmbedding (M : Mat_ C) : M ≅ ⨁ fun i => (embedding C).obj (M
     rw [Finset.sum_apply, Finset.sum_apply, Finset.sum_eq_single i]; rotate_left
     · intro b _ hb
       dsimp
-      simp only [Finset.sum_singleton]
+      erw [Finset.sum_singleton]
```
b-mehta, line 344: Can we use `rw [Finset.univ_ofSubsingleton, Finset.sum_singleton]` here? I'd like to reduce usage `erw` if possible.

mattrobball: Yes indeed! I tried turning it on as an instance here so assumed something else was going on. But it is good. Thanks.

semorrison:
bors merge

PR #14447, opened 2024-07-05
[Merged by Bors] - doc: add reassoc docstring

kbuzzard:
Add docstring for `@[reassoc]` attribute.

---

Note that, in contrast to when I'm adding maths docstrings, here I know far less about what I'm talking about (in particular I can't read the source code to verify that what I'm saying corresponds to what is written). I'm making this PR because of a general frustration about attributes not being documented properly. This PR makes the docstring added here appear when hovering over `@[reassoc]`. My sources are the [module docstring](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Tactic/CategoryTheory/Reassoc.html) for the attribute, and the comment [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/reassoc.20and.20simp/near/449310703) by Markus Himmel. I also fix a couple of typos in the module docstring.

### DIFF
--- a/Mathlib/Tactic/CategoryTheory/Reassoc.lean
+++ b/Mathlib/Tactic/CategoryTheory/Reassoc.lean
@@ -12,9 +12,9 @@ import Mathlib.Lean.Meta.Simp
 
 Adding `@[reassoc]` to a lemma named `F` of shape `∀ .., f = g`,
 where `f g : X ⟶ Y` in some category
-will create a new lemmas named `F_assoc` of shape
+will create a new lemma named `F_assoc` of shape
 `∀ .. {Z : C} (h : Y ⟶ Z), f ≫ h = g ≫ h`
-but with the conclusions simplified used the axioms for a category
+but with the conclusions simplified using the axioms for a category
 (`Category.comp_id`, `Category.id_comp`, and `Category.assoc`).
 
 This is useful for generating lemmas which the simplifier can use even on expressions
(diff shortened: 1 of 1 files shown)


joelriou:
It looks good to me. If you would like, you may also mention that the combination `@[simp, reassoc]` is possible. This can be useful for lemmas of the form `f = a ≫ b`, i.e. no composition in the LHS, but we may occasionally want to apply the `_assoc` version in the other direction: replacing `a ≫ b ≫ h` by `f ≫ h`.

Mathlib/Tactic/CategoryTheory/Reassoc.lean
```diff
@@ -48,7 +48,24 @@ but with compositions fully right associated and identities removed.
...
+So, for example, if the conclusion of `F` is `a ≫ b = g` then
+the conclusion of `F_assoc` will be `a ≫ (b ≫ h) = g ≫ h` (note that `≫` reassociates
+to the right so the brackets will not appear in the statement).
+
+This attribute is useful for generating lemmas which the simplifier can use even on expressions
+that are already right associated.
+
+Note that if you want both the lemma and the reassociated lemma to be
+`simp` lemmas, you should tag the lemma `@[reassoc (attr := simp)]`.
+The variants `@[simp, reassoc]` and `@[reassoc, simp]` on a lemma `F` will tag `F` with `@[simp]`,
```
joelriou, line 66: In my understanding, the two variants are just the same:
```suggestion
The variant `@[simp, reassoc]` on a lemma `F` will tag `F` with `@[simp]`,
```

joelriou:
Thanks!

bors d+

PR #14448, opened 2024-07-05
[Merged by Bors] - feat(ContinuousFunctionalCalculus): Define the real log based on the CFC

dupuisf:
This PR defines `CFC.log` as `cfc Real.log`, and shows its basic properties, such as the fact that it's the inverse of `NormedSpace.exp ℝ`. Along the way, we also show that the exponential defined via the CFC is equal to `NormedSpace.exp`, which is defined via power series.

---


Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/ExpLog.lean
```diff
@@ -0,0 +1,194 @@
...
++ Relate `CFC.log` to `rpow`, `zpow`, `sqrt`, `inv`.
+-/
+
+open NormedSpace
+
+section general_exponential
+variable {𝕜 : Type*} {α : Type*} [RCLike 𝕜] [TopologicalSpace α] [CompactSpace α]
+
+lemma NormedSpace.exp_continuousMap_eq (f : C(α, 𝕜)) :
+    exp 𝕜 f = (⟨exp 𝕜 ∘ f, Continuous.comp exp_continuous f.continuous⟩ : C(α, 𝕜)) := by
```
dupuisf, line 47: It feels like this lemma should belong with `NormedSpace.exp`, but none of those files have the right imports. `#find_home` tells me to put it in `Mathlib.Analysis.NormedSpace.Spectrum`, but that feels even more out of place than just leaving it here.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/ExpLog.lean
```diff
@@ -0,0 +1,195 @@
...
+    obtain ⟨z, hz⟩ := hx
+    rw [← hz.2]
+    have : 0 < r • z := (smul_pos_iff_of_pos_left hr).mpr (ha₂ z hz.1)
+    exact ne_of_gt this
+  rw [log, ← cfc_smul_id (S := ℝ) (R := ℝ) r a ha₁, ← cfc_comp Real.log (r • ·) a ha₁ ha₂'', log]
+  have hmain : Set.EqOn (Real.log ∘ (r • ·)) (fun z => Real.log r + Real.log z) (spectrum ℝ a) := by
+    intro x hx
+    simp only [smul_eq_mul, Function.comp_apply]
+    exact Real.log_mul (ne_of_gt hr) <| ne_of_gt (ha₂ x hx)
+  rw [cfc_congr hmain, cfc_const_add _ a _ ha₁ ha₂']
```
j-loreaux, line 174:
```suggestion
lemma log_smul {r : ℝ} (a : A) (ha₂ : ∀ x ∈ spectrum ℝ a, 0 < x) (hr : 0 < r)
    (ha₁ : IsSelfAdjoint a := by cfc_tac) :
    log (r • a) = algebraMap ℝ A (Real.log r) + log a := by
  have : ∀ x ∈ spectrum ℝ a, x ≠ 0 := by peel ha₂ with x hx h; exact h.ne'
  have ha₂' : ContinuousOn Real.log (spectrum ℝ a) := by fun_prop (disch := assumption)
  have ha₂'' : ContinuousOn Real.log ((r • ·) '' spectrum ℝ a)  := by fun_prop (disch := aesop)
  rw [log, ← cfc_smul_id (R := ℝ) r a ha₁, ← cfc_comp Real.log (r • ·) a ha₁ ha₂'', log]
  calc
    _ = cfc (fun z => Real.log r + Real.log z) a :=
      cfc_congr (Real.log_mul hr.ne' <| ne_of_gt <| ha₂ · ·)
    _ = _ := by rw [cfc_const_add _ a _ ha₁ ha₂']
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/ExpLog.lean
```diff
@@ -0,0 +1,195 @@
...
+  have ha₂'' : ContinuousOn Real.log ((· ^ n) '' spectrum ℝ a)  := by
+    refine ContinuousOn.mono Real.continuousOn_log fun x hx => ?_
+    rw [Set.mem_compl_singleton_iff]
+    rw [Set.mem_image] at hx
+    obtain ⟨z, hz⟩ := hx
+    rw [← hz.2]
+    exact ne_of_gt (pow_pos (ha₂ z hz.1) n)
+  rw [log, ← cfc_pow_id (R := ℝ) a n ha₁, ← cfc_comp Real.log (· ^ n) a ha₁ ha₂'', log]
+  have hmain : Real.log ∘ (· ^ n) = fun z => n • Real.log z := by ext; simp
+  rw [hmain, cfc_smul n Real.log a ha₂']
```
j-loreaux, line 192:
```suggestion
lemma log_pow (n : ℕ) (a : A) (ha₂ : ∀ x ∈ spectrum ℝ a, 0 < x)
    (ha₁ : IsSelfAdjoint a := by cfc_tac) : log (a ^ n) = n • log a := by
  have : ∀ x ∈ spectrum ℝ a, x ≠ 0 := by peel ha₂ with x hx h; exact h.ne'
  have ha₂' : ContinuousOn Real.log (spectrum ℝ a) := by fun_prop (disch := assumption)
  have ha₂'' : ContinuousOn Real.log ((· ^ n) '' spectrum ℝ a)  := by fun_prop (disch := aesop)
  rw [log, ← cfc_pow_id (R := ℝ) a n ha₁, ← cfc_comp' Real.log (· ^ n) a ha₂'', log]
  simp_rw [Real.log_pow, nsmul_eq_smul_cast ℝ n, cfc_const_mul (n : ℝ) Real.log a ha₂']
```

Mathlib/Topology/ContinuousFunction/FunctionalCalculus.lean
```diff
@@ -429,6 +429,12 @@ lemma cfc_add (f g : R → R) (hf : ContinuousOn f (spectrum R a) := by cfc_cont
     congr
   · simp [cfc_apply_of_not_predicate a ha]
 
+lemma cfc_const_add (r : R) (a : A) (f : R → R) (ha : p a)
+    (hf : ContinuousOn f (spectrum R a) := by cfc_cont_tac) :
```
j-loreaux, line 433: I think it makes sense to have `cfc_add_const` too for convenience.
```suggestion
lemma cfc_const_add (r : R) (f : R → R) (a : A)
    (hf : ContinuousOn f (spectrum R a) := by cfc_cont_tac) (ha : p a := by cfc_tac) :
```

j-loreaux (commented):
It's very annoying that `aesop` won't prove `(∀ x ∈ spectrum ℝ a, 0 < x) → ∀ x ∈ spectrum ℝ a, x ≠ 0`, but I also don't think we can make `ne_of_gt` or its synonyms into an `aesop` lemma (they would match too much, and they are very `unsafe`). It would be nice if we had some automation to handle this.

It would also be nice to have ways to customize the `fun_prop` part of `cfc_cont_tac` after the fact, maybe with some options? We should think about this. Because it would be good to add the necessary details (like the `x ≠ 0` bit above), and then set an option to make `cfc_cont_tac` do what we need it to, without forcing ourselves to manually apply `fun_prop` with special dischargers.

dupuisf:
> It's very annoying that `aesop` won't prove `(∀ x ∈ spectrum ℝ a, 0 < x) → ∀ x ∈ spectrum ℝ a, x ≠ 0`, but I also don't think we can make `ne_of_gt` or its synonyms into an `aesop` lemma (they would match too much, and they are very `unsafe`). It would be nice if we had some automation to handle this.
> 
> It would also be nice to have ways to customize the `fun_prop` part of `cfc_cont_tac` after the fact, maybe with some options? We should think about this. Because it would be good to add the necessary details (like the `x ≠ 0` bit above), and then set an option to make `cfc_cont_tac` do what we need it to, without forcing ourselves to manually apply `fun_prop` with special dischargers.

Yes, we should think about this at some point. I think it's an instance of a much more general problem though: how do we automatize goals like `ContinuousOn` on an interval? If there are function compositions, we need a way to map the interval through the function composition and to show that some interval is contained in another. I don't think we have any tactic at the moment that can do this well.

j-loreaux:
bors merge

PR #14393, opened 2024-07-03
[Merged by Bors] - chore (Algebra.Order.Ring.Defs): split file and unbundle results 

mattrobball:
All of the theorems in `Algebra.Order.Ring.Defs` currently take bundled ordered algebraic typeclasses, eg `OrderedSemiring` as paramaters unecessarily. We unbundle the results keeping the algebra and order data/theorems separate and adding the mixin compatbility typeclasses as needed and place the theorems in `Algebra.Order.Ring.Unbundled.Basic`. As `Algebra.Order.Ring.Defs` and upstream resuls already allow us to pass from bundled classes to unbundled classes appropriately, we lose no uusability by doing this. We lessen dependecies and gain future flexibility.

---

- [x] depends on #14390 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14451, opened 2024-07-05
[Merged by Bors] - chore (Algebra.Order.Field.Defs): split off results about unbundled ordered algebra

mattrobball:
We should be able to use these results about unbundled ordered algebra without importing the heavy machinery of the bundled typeclasses. 

---

- [x] depends on: #14393 


mattrobball:
!bench

semorrison:
bors merge

PR #14456, opened 2024-07-05
[Merged by Bors] - feat(RingTheory/TwoSidedIdeal): add lattice structure to two-sided-ideals

jjaassoonn:
---


Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,93 @@
...
+  bot := { ringCon := ⊥ }
+
+lemma mem_bot {x : R} : x ∈ (⊥ : TwoSidedIdeal R) ↔ x = 0 :=
+  Iff.rfl
+
+instance : CompleteLattice (TwoSidedIdeal R) :=
+  Function.Injective.completeLattice (f := TwoSidedIdeal.ringCon (R := R))
+    ringCon_injective (fun _ _ => rfl) (fun _ _ => rfl)
+      (fun _ => sSup_image) (fun _ => sInf_image) rfl rfl
+
```
jjaassoonn, line 54: This actually causes instance diamond: we now have two partial orders on `TwoSidedIdeals`: one from `SetLike` instances; the other from `Lattice` structure. I need some help on solving this.

jjaassoonn: One potential solution is to write this instance by hand.

jjaassoonn:
I didn't use `Function.Injective.completeLattice` is to avoid instance diamond between `SetLike.instPartialOrder` and `CompleteLattice.instPartialOrder`

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,113 @@
...
+/-!
+# The complete lattice structure on two-sided ideals
+-/
+
+namespace TwoSidedIdeal
+
+variable (R : Type*) [NonUnitalNonAssocRing R]
+
+instance : Sup (TwoSidedIdeal R) where
+  sup := fun I J => { ringCon := I.ringCon ⊔ J.ringCon }
```
eric-wieser, line 18: nit:
```suggestion
  sup I J := { ringCon := I.ringCon ⊔ J.ringCon }
```

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -43,6 +43,9 @@ section NonUnitalNonAssocRing
 
 variable {R : Type*} [NonUnitalNonAssocRing R] (I : TwoSidedIdeal R)
 
+lemma ringCon_injective : Function.Injective (TwoSidedIdeal.ringCon (R := R)) := by
```
eric-wieser, line 46: Could you add the `Equiv` too, or maybe an `OrderIso`?

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,113 @@
...
+  bot := { ringCon := ⊥ }
+
+lemma mem_bot {x : R} : x ∈ (⊥ : TwoSidedIdeal R) ↔ x = 0 :=
+  Iff.rfl
+
+instance : CompleteLattice (TwoSidedIdeal R) where
+  __ := (inferInstance : SemilatticeSup (TwoSidedIdeal R))
+  __ := (inferInstance : SemilatticeInf (TwoSidedIdeal R))
+  __ := (inferInstance : CompleteSemilatticeSup (TwoSidedIdeal R))
+  __ := (inferInstance : CompleteSemilatticeInf (TwoSidedIdeal R))
```
eric-wieser, line 109: I suspect you can drop these and Lean works it out automatically?

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,113 @@
...
+
+    refine sup_le (fun x y h => ?_) (fun x y h => ?_) <;> rw [rel_iff] at h ⊢ <;> rw [mem_mk']
+    exacts [⟨_, ⟨h, ⟨0, ⟨zero_mem _, add_zero _⟩⟩⟩⟩, ⟨0, ⟨zero_mem _, ⟨_, ⟨h, zero_add _⟩⟩⟩⟩]
+
+  · rintro ⟨y, ⟨hy, ⟨z, ⟨hz, rfl⟩⟩⟩⟩; exact add_mem _ (mem_sup_left hy) (mem_sup_right hz)
+
+end sup
+
+instance : Inf (TwoSidedIdeal R) where
+  inf := fun I J => { ringCon := I.ringCon ⊓ J.ringCon }
```
eric-wieser, line 63: For each of these instances, could you add the obvious `(I ⊓ J).ringCon = I.ringCon ⊓ J.ringCon`-style lemma?

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
jjaassoonn: done

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
jjaassoonn: unfortunately not. Lean complains
```
fields missing: 'inf_le_left', 'inf_le_right', 'le_inf', 'le_sSup', 'sSup_le', 'sInf_le', 'le_sInf'
```

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
jjaassoonn: done

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,135 @@
...
+-/
+
+namespace TwoSidedIdeal
+
+variable (R : Type*) [NonUnitalNonAssocRing R]
+
+instance : Sup (TwoSidedIdeal R) where
+  sup I J := { ringCon := I.ringCon ⊔ J.ringCon }
+
+lemma sup_ringCon (I J : TwoSidedIdeal R) : (I ⊔ J).ringCon = I.ringCon ⊔ J.ringCon := rfl
```
jjaassoonn, line 20: I am not sure if these should be tagged with `simp`. Ideally, when people use `TwoSidedIdeal`, they shouldn't be seeing `TwoSidedIdeal.ringCon` all the time. What do you think @eric-wieser?

Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
```diff
@@ -73,8 +73,22 @@ def coeOrderEmbedding : TwoSidedIdeal R ↪o Set R where
   inj' := SetLike.coe_injective
   map_rel_iff' {I J} := ⟨fun (h : (I : Set R) ⊆ (J : Set R)) _ h' => h h', fun h _ h' => h h'⟩
 
-lemma le_iff {I J : TwoSidedIdeal R} : I ≤ J ↔ (I : Set R) ⊆ (J : Set R) :=
-  coeOrderEmbedding.map_rel_iff.symm
+lemma le_iff {I J : TwoSidedIdeal R} : I ≤ J ↔ (I : Set R) ⊆ (J : Set R) := Iff.rfl
+
+/-- Two-sided-ideals corresponds to congruence relation of a ring. -/
```
kbuzzard, line 78:
```suggestion
/-- Two-sided-ideals corresponds to congruence relations on a ring. -/
```

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,135 @@
...
+      ⟨0, ⟨zero_mem _, ⟨0, ⟨zero_mem _, zero_add _⟩⟩⟩⟩
+      (by rintro _ _ ⟨x, ⟨hx, ⟨y, ⟨hy, rfl⟩⟩⟩⟩ ⟨a, ⟨ha, ⟨b, ⟨hb, rfl⟩⟩⟩⟩;
+          exact ⟨x + a, ⟨add_mem _ hx ha, ⟨y + b, ⟨add_mem _ hy hb, by abel⟩⟩⟩⟩)
+      (by rintro _ ⟨x, ⟨hx, ⟨y, ⟨hy, rfl⟩⟩⟩⟩
+          exact ⟨-x, ⟨neg_mem _ hx, ⟨-y, ⟨neg_mem _ hy, by abel⟩⟩⟩⟩)
+      (by rintro r _ ⟨x, ⟨hx, ⟨y, ⟨hy, rfl⟩⟩⟩⟩
+          exact ⟨_, ⟨mul_mem_left _ _ _ hx, ⟨_, ⟨mul_mem_left _ _ _ hy, mul_add _ _ _ |>.symm⟩⟩⟩⟩)
+      (by rintro r _ ⟨x, ⟨hx, ⟨y, ⟨hy, rfl⟩⟩⟩⟩
+          exact ⟨_, ⟨mul_mem_right _ _ _ hx, ⟨_, ⟨mul_mem_right _ _ _ hy, add_mul _ _ _ |>.symm⟩⟩⟩⟩)
+
```
kbuzzard, line 53: Why the blank lines?

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,135 @@
...
+instance : SupSet (TwoSidedIdeal R) where
+  sSup s := { ringCon := sSup $ TwoSidedIdeal.ringCon '' s }
+
+lemma sSup_ringCon (S : Set (TwoSidedIdeal R)) :
+    (sSup S).ringCon = sSup (TwoSidedIdeal.ringCon '' S) := rfl
+
+lemma iSup_ringCon {ι : Type*} (I : ι → TwoSidedIdeal R) :
+    (⨆ i, I i).ringCon = ⨆ i, (I i).ringCon := by
+  simp only [iSup, sSup_ringCon]; congr!; ext; simp
```
kbuzzard, line 86:
```suggestion
  simp only [iSup, sSup_ringCon]; congr; ext; simp
```
(it's just `apply congr_arg`)

kbuzzard (approved):

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
jjaassoonn: I though it could increase readibility as the above is the construction of some two-sided-ideal

jjaassoonn: removed blank line

riccardobrasca:
I am busy until Wednesday, but I can have a look later.

jjaassoonn:
> I am busy until Wednesday, but I can have a look later.

No rush! Thank you very much!

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,132 @@
...
+instance : SemilatticeSup (TwoSidedIdeal R) where
+  le_sup_left I J :=  by rw [ringCon_le_iff]; exact le_sup_left
+  le_sup_right I J := by rw [ringCon_le_iff]; exact le_sup_right
+  sup_le I J K h1 h2 := by rw [ringCon_le_iff] at h1 h2 ⊢; exact sup_le h1 h2
+
+section sup
+
+variable {R}
+
+lemma mem_sup_left {I J : TwoSidedIdeal R} {x : R} :
```
riccardobrasca, line 31: Why not having `(hx : x \in I)`?

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
```diff
@@ -0,0 +1,132 @@
...
+    x ∈ I → x ∈ I ⊔ J :=
+  fun h => (show I ≤ I ⊔ J from le_sup_left) h
+
+lemma mem_sup_right {I J : TwoSidedIdeal R} {x : R} :
```
riccardobrasca, line 35: Same.

riccardobrasca (approved):
LGTM, thanks!

@eric-wieser there is still a question for you :)

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
jjaassoonn: done

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
jjaassoonn: done

j-loreaux:
I have a big question: why aren't we just pulling the entire `CompleteLattice` structure (aside from the `LE`, which comes via the `SetLike` instance) through the order isomorphism all at once? Is that not easily doable?

jjaassoonn:
> I have a big question: why aren't we just pulling the entire `CompleteLattice` structure (aside from the `LE`, which comes via the `SetLike` instance) through the order isomorphism all at once? Is that not easily doable?

Because pulling back the order of ringcon results in a diamond: the “smaller than” from the pulling back is not definitionally equal to that from setlike instance

j-loreaux:
Right, but I'm talking about leaving the existing order alone and pulling the rest back.

j-loreaux:
Sounds like we want a `CompleteLattice.replacePartialOrder` declaration marked as `reducible` for things like this. But that's for a separate PR, so I'll leave this alone.

Mathlib/RingTheory/TwoSidedIdeal/Lattice.lean
j-loreaux: I agree that people shouldn't really see this, but if they do, I think having this as a `simp` lemma is fine.

j-loreaux:
I addressed the comment directed at Eric (I think satisfactorily), and given that there are two other approving reviews:

bors merge

PR #14462, opened 2024-07-05
[Merged by Bors] - chore(TwoSidedIdeal/Basic): apply forgotten review comments

jjaassoonn:
After being delegated, I forgot to apply the suggestions from #13902 by @jcommelin ....

---

### DIFF
--- a/Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
+++ b/Mathlib/RingTheory/TwoSidedIdeal/Basic.lean
@@ -96,7 +96,7 @@ instance : AddSubgroupClass (TwoSidedIdeal R) R where
   add_mem := @add_mem _ _
   neg_mem := @neg_mem _ _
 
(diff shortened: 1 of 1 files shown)


grunweg:
Thanks for following up! I have just one tiny question; would `maintainer merge` with that addressed.
Did you think about this part of `jcommelin`'s comment?
> Then you can use it in theproofs of sub_mem and nsmul_mem etc [...] this isn't very important

jjaassoonn:
> Thanks for following up! I have just one tiny question; would `maintainer merge` with that addressed. Did you think about this part of `jcommelin`'s comment?
> 
> > Then you can use it in theproofs of sub_mem and nsmul_mem etc [...] this isn't very important

Thanks for pointing this out. I have applied this suggestion as well!

riccardobrasca:
Thanks!

bors merge

PR #14465, opened 2024-07-05
[Merged by Bors] - chore: postpone Field past Algebra/Defs

kim-em:
---


Mathlib/Algebra/Algebra/Field.lean
```diff
@@ -0,0 +1,59 @@
+/-
```
grunweg, line 1: IIUC, this just copies the old module docstring. What's mathlib policy here?
(If anybody wants to do archeology: [this mathlib3 blame](https://github.com/leanprover-community/mathlib/blame/master/src/algebra/algebra/basic.lean), starting at line 184, should be helpful.)

grunweg (commented):
Thanks for doing this - this looks good to me, with one question about the module docstring.
maintainer merge

mattrobball:
bors merge

PR #14467, opened 2024-07-06
[Merged by Bors] - doc: add definition of upper/lower set to UpperSet, LowerSet

kbuzzard:
Explanations are there in the docstring to `IsUpperSet` and `IsLowerSet` but there's no reason they can't be here as well (indeed `UpperSet` came up in a Zulip discussion and the current docstring is in some sense content-free; perhaps the term is standard amongst order theorists but I didn't know what it meant and it's not hard to explain). 

---

### DIFF
--- a/Mathlib/Order/UpperLower/Basic.lean
+++ b/Mathlib/Order/UpperLower/Basic.lean
@@ -511,12 +511,7 @@ section LE
 
 variable [LE α]
 
-/--
-The type of upper sets of an order.
-
-An upper set in an order is a subset such that any element greater than one of its members is
-also a member. Also called up-set, upward-closed set.
--/
+@[inherit_doc IsUpperSet]
 structure UpperSet (α : Type*) [LE α] where
   /-- The carrier of an `UpperSet`. -/
   carrier : Set α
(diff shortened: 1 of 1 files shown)


Mathlib/Order/UpperLower/Basic.lean
```diff
@@ -511,15 +511,25 @@ section LE
 
 variable [LE α]
 
-/-- The type of upper sets of an order. -/
+/--
+The type of upper sets of an order.
+
+An upper set in an order is a subset such that any element greater than one of its members is
```
adomani, line 517: I forget the actual syntax for it, but this might be one of the rare cases where `extend_doc` could be useful.  Maybe you can use `@[inherit_doc IsUpperSet]` on `UpperSet` and the `extend_doc UpperSet before "..."

adomani: I just tested and this has the same effect as your change:
```lean
@[inherit_doc IsUpperSet]
structure UpperSet (α : Type*) [LE α] where
  /-- The carrier of an `UpperSet`. -/
  carrier : Set α
  /-- The carrier of an `UpperSet` is an upper set. -/
  upper' : IsUpperSet carrier
#align upper_set UpperSet

extend_docs UpperSet before "The type of upper sets of an order."

@[inherit_doc IsLowerSet]
structure LowerSet (α : Type*) [LE α] where
  /-- The carrier of a `LowerSet`. -/
  carrier : Set α
  /-- The carrier of a `LowerSet` is a lower set. -/
  lower' : IsLowerSet carrier
#align lower_set LowerSet

extend_docs LowerSet before "The type of lower sets of an order."
```

and moreover you do not have to worry about synchronising the doc-strings!

jcommelin (commented):
bors d+

YaelDillies (approved):
I agree with Damiano's comment. Otherwise LGTM

Mathlib/Order/UpperLower/Basic.lean
kbuzzard: >  this has the same effect as your change:

It doesn't because I cunningly removed a redundant `alpha` from the IsUpperSet docstring ;-) But I'm happy to put it back.

PR #14470, opened 2024-07-06
[Merged by Bors] - chore(Algebra/Lie): prerequisites for Lie's theorem

jcommelin:
---


Mathlib/Algebra/Lie/Solvable.lean
```diff
@@ -192,6 +192,12 @@ theorem derivedSeries_map_eq (k : ℕ) (h : Function.Surjective f) :
   · simp only [derivedSeries_def, map_bracket_eq f h, ih, derivedSeriesOfIdeal_succ]
 #align lie_ideal.derived_series_map_eq LieIdeal.derivedSeries_map_eq
 
+theorem derivedSeries_eq_top (n : ℕ) (h : derivedSeries R L 1 = ⊤) : derivedSeries R L n = ⊤ := by
+  rw [derivedSeries_def]
+  induction' n with n ih
+  · simp only [Nat.zero_eq, derivedSeriesOfIdeal_zero]
+  · rwa [derivedSeriesOfIdeal_succ, ih]
```
ocfnash, line 199: How about:
```suggestion
theorem derivedSeries_succ_eq_top_iff (n : ℕ) :
    derivedSeries R L (n + 1) = ⊤ ↔ derivedSeries R L 1 = ⊤ := by
  simp only [derivedSeries_def]
  induction' n with n ih; · simp
  rw [derivedSeriesOfIdeal_succ]
  refine ⟨fun h ↦ ?_, fun h ↦ by rwa [ih.mpr h]⟩
  rw [← ih, eq_top_iff]
  conv_lhs => rw [← h]
  exact LieSubmodule.lie_le_right _ _

theorem derivedSeries_eq_top (n : ℕ) (h : derivedSeries R L 1 = ⊤) :
    derivedSeries R L n = ⊤ := by
  cases n
  · rfl
  · rwa [derivedSeries_succ_eq_top_iff]
```

Mathlib/Algebra/Lie/Solvable.lean
```diff
@@ -211,6 +217,13 @@ instance isSolvableAdd {I J : LieIdeal R L} [hI : IsSolvable R I] [hJ : IsSolvab
   exact ⟨⟨k + l, LieIdeal.derivedSeries_add_eq_bot hk hl⟩⟩
 #align lie_algebra.is_solvable_add LieAlgebra.isSolvableAdd
 
+theorem derivedSeries_ne_top_of_solvable [IsSolvable R L] [Nontrivial L] :
+    derivedSeries R L 1 ≠ ⊤ := by
```
ocfnash, line 221: I think it is slightly more useful to prove:
```suggestion
theorem derivedSeries_lt_top_of_solvable [IsSolvable R L] [Nontrivial L] :
    derivedSeries R L 1 < ⊤ := by
```
since it is _slightly_ more easier to pass from `I < J` to `I ≠ J` using `LT.lt.ne` than vice-versa (when `J = ⊤`).

Mathlib/Algebra/Lie/Submodule.lean
```diff
@@ -79,6 +79,10 @@ instance coeSubmodule : CoeOut (LieSubmodule R L M) (Submodule R M) :=
   ⟨toSubmodule⟩
 #align lie_submodule.coe_submodule LieSubmodule.coeSubmodule
 
+instance instCanLiftSubmoduleLieSubmodule : CanLift (Submodule R M) (LieSubmodule R L M) (·)
```
ocfnash, line 82: How about:
```suggestion
instance instCanLiftSubmoduleLieSubmodule : CanLift (Submodule R M) (LieSubmodule R L M) (↑)
```

ocfnash (approved):
Thanks!

bors d+

Mathlib/Algebra/Lie/Submodule.lean
jcommelin: Those are not syntactically equal, for some reason. And I think the `.toSubmodule` version provided by `(·)` has better simp properties.

We should probably fix the coe at some point.

PR #14471, opened 2024-07-06
[Merged by Bors] - feat(MeasureTheory): method of exhaustion

RemyDegenne:
If `μ, ν` are two measures with `ν` s-finite, then there exists a set `s` such that `μ` is sigma-finite on `s`, and for all sets `t ⊆ sᶜ`, either `ν t = 0` or `μ t = ∞`. The technique used to obtain this result is what Halmos calls the "method of exhaustion".

We use this to prove that if `μ ≪ ν` and `ν` is s-finite, then `μ` is s-finite.

This PR also moves the definition of `sigmaFiniteSet` to the new file and refactors it with the new definition `sigmaFiniteSetWRT`.

---


Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,207 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Rémy Degenne
+-/
+import Mathlib.MeasureTheory.Measure.Typeclasses
+
+/-!
+# Method of exhaustion
+
+If `μ, ν` are two measures with `ν` finite, then there exists a set `s` such that
+`μ` is sigma-finite on `s`, and for all sets `t ⊆ sᶜ`, either `ν t = 0` or `μ t = ∞`.
```
sgouezel, line 12: I would expect that this statement generalizes readily when nu is S-finite (because it is then equivalent to a finite measure). What about defining `sigmaFiniteSetWRT` without any assumptions on nu as: if there exists a measurable set s such that (mu.restrict s is sigma-finite and for all t in s^c, either nu t = 0 or mu t = \infty), then pick such a set s, else pick the empty set. Then prove that for finite measures, it does what it should do (following exactly your proof), and then deduce the same for S-finite nu?

RemyDegenne:
@sgouezel I generalized the definition of `sigmaFiniteSetWRT` in the way you suggested.
I have added another change to the PR: I changed the definition of `sigmaFiniteSet` to use the new `sigmaFiniteSetWRT` and moved the whole thing to the new Exhaustion file.

RemyDegenne:
The import change bot is saying "Too many changes (249)!" but it looks like that simply means that 249 files now import the new Exhaustion file (since it is imported by `WithDensity`).

semorrison:
> The import change bot is saying "Too many changes (249)!" but it looks like that simply means that 249 files now import the new Exhaustion file (since it is imported by `WithDensity`).

Oh, we should definitely change that bot so the message is more clearly a failure to display the differences, target than objecting to the PR!

@adomani?

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,335 @@
...
+for `C` the supremum of `ν s` over all measurable sets `s` with `μ.restrict s` sigma-finite,
+`ν (μ.sigmaFiniteSetGE ν n) ≥ C - 1/n`. -/
+def Measure.sigmaFiniteSetGE (μ ν : Measure α) [IsFiniteMeasure ν] (n : ℕ) : Set α :=
+  (exists_isSigmaFiniteSet_measure_ge μ ν n).choose
+
+lemma measurableSet_sigmaFiniteSetGE (μ ν : Measure α) [IsFiniteMeasure ν] (n : ℕ) :
+    MeasurableSet (μ.sigmaFiniteSetGE ν n) :=
+  (exists_isSigmaFiniteSet_measure_ge μ ν n).choose_spec.1
+
+instance (n : ℕ) : SigmaFinite (μ.restrict (μ.sigmaFiniteSetGE ν n)) :=
```
sgouezel, line 128: Since this construction is purely local to this file, I wouldn't register it as a global instance, to avoid cluttering instance search. Can you make it a local instance instead?

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,335 @@
...
+variable [IsFiniteMeasure ν]
+
+/-! We prove that the condition in the definition of `sigmaFiniteSetWRT` is true for finite
+measures. Since every s-finite measure is absolutely continuous with respect to a finite measure,
+the condition will then also be true for s-finite measures. -/
+
+/-- Let `C` be the supremum of `ν s` over all measurable sets `s` such that `μ.restrict s` is
+sigma-finite. `C` is finite since `ν` is a finite measure. Then there exists a measurable set `t`
+with `μ.restrict t` sigma-finite such that `ν t ≥ C - 1/n`. -/
+lemma exists_isSigmaFiniteSet_measure_ge (μ ν : Measure α) [IsFiniteMeasure ν] (n : ℕ) :
```
sgouezel, line 90: The assumptions `(μ ν : Measure α) [IsFiniteMeasure ν]` show up in this lemma and all the next ones. Can you make them into a `variable` line?

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,335 @@
...
+`ν (μ.sigmaFiniteSetWRT ν)` has maximal measure among such sets.
+This is an auxiliary definition for `Measure.sigmaFiniteSetWRT`. -/
+def Measure.sigmaFiniteSetWRT' (μ ν : Measure α) [IsFiniteMeasure ν] : Set α :=
+  ⋃ n, μ.sigmaFiniteSetGE ν n
+
+lemma measurableSet_sigmaFiniteSetWRT' (μ ν : Measure α) [IsFiniteMeasure ν] :
+    MeasurableSet (μ.sigmaFiniteSetWRT' ν) :=
+  MeasurableSet.iUnion (measurableSet_sigmaFiniteSetGE _ _)
+
+instance [IsFiniteMeasure ν] : SigmaFinite (μ.restrict (μ.sigmaFiniteSetWRT' ν)) := by
```
sgouezel, line 164: ditto, make it a local instance?

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,335 @@
...
+  · refine (le_iSup (f := fun _ ↦ _)
+      (inferInstance : SigmaFinite (μ.restrict (μ.sigmaFiniteSetWRT' ν)))).trans ?_
+    exact le_iSup₂ (f := fun s _ ↦ ⨆ (_ : SigmaFinite (μ.restrict s)), ν s) (μ.sigmaFiniteSetWRT' ν)
+      (measurableSet_sigmaFiniteSetWRT' μ ν)
+  · exact le_of_tendsto' (tendsto_measure_sigmaFiniteSetGE μ ν)
+      (fun _ ↦ measure_mono (Set.subset_iUnion _ _))
+
+/-- For all measurable sets `s` in `(μ.sigmaFiniteSetWRT ν)ᶜ`, if `ν s ≠ 0` then `μ s = ∞`. -/
+lemma measure_eq_top_of_subset_compl_sigmaFiniteSetWRT'
+    (hs : MeasurableSet s) (hs_subset : s ⊆ (μ.sigmaFiniteSetWRT' ν)ᶜ) (hνs : ν s ≠ 0) :
```
sgouezel, line 209: Once you've proved this one, can you drop the `MeasurableSet` assumption, by arguing that the measure is the inf of the measure of measurable sets containing them (which you can take inside `(μ.sigmaFiniteSetWRT' ν)ᶜ` as it's measurable)?

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
```diff
@@ -0,0 +1,335 @@
...
+    · exact le_iSup₂ (f := fun s _ ↦ ⨆ (_ : SigmaFinite (μ.restrict _)), ν s)
+        (μ.sigmaFiniteSetWRT' ν ∪ s) ((measurableSet_sigmaFiniteSetWRT' μ ν).union hs)
+  exact h_lt.not_le h_le
+
+end IsFiniteMeasure
+
+section SFinite
+
+lemma measure_eq_top_of_subset_compl_sigmaFiniteSetWRT [SFinite ν]
+    (hs : MeasurableSet s) (hs_subset : s ⊆ (μ.sigmaFiniteSetWRT ν)ᶜ) (hνs : ν s ≠ 0) :
```
sgouezel, line 234: If my comment above works fine, then you should be able to remove `hs` here, and then the lemma ` measure_eq_top_of_absolutelyContinuous_of_subset_compl_sigmaFiniteSetWRT` becomes redundant, right?

adomani:
Oh, I had not thought of the possible interpretation that the bot was rejecting the PR!  I introduced the limit on number of changed files, since otherwise, if the message is too long, then GitHub will just fail to post it.

I'll update the message, but still leave on the stop-gag, to avoid corrupting the `PR summary` message.

Mathlib/MeasureTheory/Decomposition/Exhaustion.lean
RemyDegenne: I'd rather like to keep all explicit arguments with the lemma definitions.

RemyDegenne:
I don't know how to turn instances local. When I add the `local` keyword in front of them, the theorem vs def linter complains.

adomani:
This is exactly the reason the linter complains: as far as I know, there is no way to make an instance local.  There is a lean4 issue about this.  I'll look for it and edit this comment.

RemyDegenne:
Thanks for the info! I'll turn them into lemmas.

sgouezel (commented):
bors d+
Thanks!

PR #14474, opened 2024-07-06
[Merged by Bors] - chore: homogenize TODO format

mo271:
This makes sure that the by far most common way of declaring TODOs in module doc comments is also used by the few that deviated from that. This makes sure that it is easier to find those.

Drive by: fix number of `#` for some other headings, to make doc strings more uniform.

---

### DIFF
--- a/Mathlib/Algebra/Group/NatPowAssoc.lean
+++ b/Mathlib/Algebra/Group/NatPowAssoc.lean
@@ -28,7 +28,7 @@ We also produce the following instances:
 
 - `NatPowAssoc` for Monoids, Pi types and products.
 
-## Todo
+## TODO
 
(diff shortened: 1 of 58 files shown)


mo271:
There is also 50 occurrences of "Future Work", which I suppose could also be replaced with "TODO", but that seems more controversial, so I leave it for now...

rwst (approved):
Exactly what is described, except the very last change, which leaves three ###s.  LGTM

mo271:
> Exactly what is described, except the very last change, which leaves three ###s. LGTM

For this one, the three ###s make sense, because there are multiple TODOs in different levels.

eric-wieser:
bors merge

Thanks!

PR #14478, opened 2024-07-06
[Merged by Bors] - feat(RingTheory/Unramified): Unramified + free => finite

erdOne:
---

### DIFF
--- a/Mathlib/RingTheory/Unramified/Finite.lean
+++ b/Mathlib/RingTheory/Unramified/Finite.lean
@@ -3,7 +3,6 @@ Copyright (c) 2024 Andrew Yang. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Andrew Yang
 -/
-import Mathlib.RingTheory.EssentialFiniteness
 import Mathlib.RingTheory.Ideal.IdempotentFG
 import Mathlib.RingTheory.Unramified.Derivations
 import Mathlib.RingTheory.Flat.Stability
--- a/Mathlib/RingTheory/Unramified/Finite.lean
+++ b/Mathlib/RingTheory/Unramified/Finite.lean
@@ -44,7 +44,6 @@ A finite-type `R`-algebra `S` is (formally) unramified iff there exists a `t : S
 theorem iff_exists_tensorProduct [EssFiniteType R S] :
     FormallyUnramified R S ↔ ∃ t : S ⊗[R] S,
       (∀ s, ((1 : S) ⊗ₜ[R] s - s ⊗ₜ[R] (1 : S)) * t = 0) ∧ TensorProduct.lmul' R t = 1 := by
-  have := KaehlerDifferential.ideal_fg R S
   rw [iff_subsingleton_kaehlerDifferential, KaehlerDifferential,
     Ideal.cotangent_subsingleton_iff, Ideal.isIdempotentElem_iff_of_fg _
       (KaehlerDifferential.ideal_fg R S)]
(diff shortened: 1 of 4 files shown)


Mathlib/RingTheory/Unramified/Finite.lean
```diff
@@ -0,0 +1,277 @@
...
+      intro i
+      simp_rw [Finsupp.sum, TensorProduct.smul_tmul', ← TensorProduct.sum_tmul]
+      congr 1
+      exact b.total_repr _
+    trans (x ⊗ₜ 1) * elem R S
+    · simp_rw [this, hf, Finsupp.sum, Finset.mul_sum, TensorProduct.tmul_mul_tmul, one_mul]
+    · rw [← one_tmul_mul_elem, hf, finite_of_free_aux]
+      rfl
+    · intro _; simp
+    · intro _; simp
```
kbuzzard, line 212:
```suggestion
    · intro; simp
    · intro; simp
```
You don't need them so I'm not sure there's any point in putting them, like you wouldn't put `apply foo _`.

kbuzzard (approved):
Very well done indeed! I've left a bunch of mostly minor comments.

Mathlib/RingTheory/Unramified/Finite.lean
```diff
@@ -0,0 +1,277 @@
...
+variable (R S)
+
+/-- An unramified free algebra is finitely generated. Iversen I.2.8 -/
+lemma finite_of_free [Module.Free R S] : Module.Finite R S := by
+  classical
+  let I := Module.Free.ChooseBasisIndex R S
+  -- Let `bᵢ` be an `R`-basis of `S`.
+  let b : Basis I R S := Module.Free.chooseBasis R S
+  -- Let `∑ₛ fᵢ ⊗ bᵢ : S ⊗[R] S` (summing over some finite `s`) be an element such that
+  -- `∑ₛ fᵢbᵢ = 1` and `∀ s, sfᵢ ⊗ bᵢ = aᵢ ⊗ sfᵢ` which exists since `S` is unramified over `R`.
```
erdOne, line 156:
```suggestion
  -- `∑ₛ fᵢbᵢ = 1` and `∀ x : S, xfᵢ ⊗ bᵢ = aᵢ ⊗ xfᵢ` which exists since `S` is unramified over `R`.
```

Mathlib/RingTheory/Unramified/Finite.lean
```diff
@@ -0,0 +1,278 @@
...
+variable (R S) in
+/--
+A finite-type `R`-algebra `S` is (formally) unramified iff there exists a `t : S ⊗[R] S` satisfying
+1. `t` annihilates every `1 ⊗ s - s ⊗ 1`.
+2. the image of `t` is `1` under the map `S ⊗[R] S → S`.
+See `Algebra.FormallyUnramified.iff_exists_tensorProduct`.
+This is the choice of such a `t`.
+-/
+noncomputable
+def elem : S ⊗[R] S :=
```
riccardobrasca, line 94: Do you think this element will be used outside this file? Asking because in that case it is probably better to prove everything for a generic `t` such that blah blah, rather than only for the one given by `choose`.

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/RingTheory/Unramified/Finite.lean
erdOne: Maybe, but I don't think the element itself is important as long as it exists.

riccardobrasca: Yes sure, but this is exactly my point: your lemmas don't apply to element with the same property. Anyway it's not very important, we can always fix this later.

erdOne:
bors merge

PR #14479, opened 2024-07-06
[Merged by Bors] - feat (Algebra/Order/AddTorsor) : Antidiagonals for SMul and VAdd, with finiteness properties

ScottCarnahan:
We introduce antidiagonals for scalar multiplication and vector addition, and show that they are finite for partially well-ordered input sets.  We also add some results involving minima.  This will be used in future PRs on Hahn modules and vertex operators.

- [x] depends on: #14732 

---


ScottCarnahan:
Given the increase in imports, should I make a new file instead?

Mathlib/Data/Set/SMulAntidiagonal.lean
```diff
@@ -0,0 +1,145 @@
...
+    {t : Set P} (hs : s.IsPWO) (ht : t.IsPWO) : IsPWO (s • t) := by
+  rw [← @image_smul_prod]
+  exact (hs.prod ht).image_of_monotone (monotone_fst.SMul monotone_snd)
+
+@[to_additive]
+theorem IsWF.SMul [LinearOrder G] [LinearOrder P] [SMul G P] [IsOrderedCancelSMul G P] {s : Set G}
+    {t : Set P} (hs : s.IsWF) (ht : t.IsWF) : IsWF (s • t) :=
+  (hs.isPWO.SMul ht.isPWO).isWF
+
+-- _root_ seems to be needed here, and I have no idea why.
```
mattrobball, line 135: It is the naming. `IsWF.SMul` now exists in the same namespace as `IsWF.min_SMul`. 

If the naming was kept as is, convention would be `IsWF.smul` but I think `Set.isWF_smul` is better. Also, for similar declarations here.

Mathlib/Algebra/Order/AddTorsor.lean
```diff
@@ -76,6 +75,19 @@ instance [OrderedCommMonoid G] : IsOrderedSMul G G where
   smul_le_smul_left _ _ := mul_le_mul_left'
   smul_le_smul_right _ _ := mul_le_mul_right'
 
+@[to_additive]
+theorem IsOrderedSMul.smul_le_smul [Preorder G] [Preorder P] [SMul G P] [IsOrderedSMul G P]
+    {a b : G} {c d : P} (hab : a ≤ b) (hcd : c ≤ d) : a • c ≤ b • d :=
+  (IsOrderedSMul.smul_le_smul_left _ _ hcd _).trans (IsOrderedSMul.smul_le_smul_right _ _ hab _)
+
+/-- The vector sum of two monotone functions is monotone. -/
```
semorrison, line 83: Wrong doc string?

semorrison:
bors d+

ScottCarnahan:
Thanks for the reviews!
bors r+

PR #14480, opened 2024-07-06
feat: the minImports linter

adomani:
An extension of #14437 that runs the import minimization on a sequence of commands.

```lean
#min_imports start
#check ℤ
lemma ohMy : True := by extract_goal; change? _; trivial
until
-- import Mathlib.Tactic.Change
-- import Mathlib.Tactic.ExtractGoal
-- import Mathlib.Tactic.Lemma
-- import Mathlib.Init.Data.Int.Basic
```

---

- [ ] depends on: #14437, where the main command `#min_imports in` is defined.


adomani:
Closing in favour of #14840.

PR #14482, opened 2024-07-06
[Merged by Bors] - feat(RingTheory/Flat): a module is flat iff tensoring preserves exact sequences

jjaassoonn:
---

- [ ] depends on: #14520


Mathlib/RingTheory/Flat/CategoryTheory.lean
```diff
@@ -0,0 +1,147 @@
...
+  preserves {c} hc := by
+    let ι : c.pt ⟶ X := c.π.app .zero
+    have mono0 : Mono ι :=
+      { right_cancellation := fun {Z} g h H => by
+          let c' : Limits.Cone (Limits.parallelPair f 0) :=
+          ⟨Z, ⟨fun | .zero => h ≫ ι | .one => 0, by rintro _ _ (⟨j⟩|⟨j⟩) <;> simpa [ι] using H⟩⟩
+
+          rw [hc.uniq c' g, hc.uniq c' h] <;>
+          rintro (⟨j⟩|⟨j⟩) <;> simpa [ι] using H }
+    have exact0 : Exact ι f := by
```
joelriou, line 56: Could you use https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Homology/ShortComplex/Exact.html instead? (It is likely that the notion from `Algebra.Homology.Exact` will be removed eventually. Actually, your PR reminds me that I should remove it...)

jjaassoonn: I am not very proficient in the homological algebra part of library. Can you check if you like the changes now please?

jjaassoonn:
!bench

Mathlib/RingTheory/Flat/CategoryTheory.lean
jjaassoonn: The PR has gotten quite big after the change, if necessary I will split it.

jjaassoonn:
!bench

Mathlib/Algebra/Category/ModuleCat/Abelian.lean
```diff
@@ -128,4 +129,30 @@ theorem exact_iff : Exact f g ↔ LinearMap.range f = LinearMap.ker g := by
 set_option linter.uppercaseLean3 false in
 #align Module.exact_iff ModuleCat.exact_iff
 
+theorem shortComplex_exact (H : LinearMap.range f = LinearMap.ker g) :
```
joelriou, line 132: It seems to me that these lemmas are already in the file `Algebra.Homology.ShortComplex.ModuleCat`.

joelriou: (If this is needed, note also that in #14520, I introduce a lemma which relates `Function.Exact` and  `ShortComplex.Exact` in the category of modules.)

jjaassoonn: Thanks! That will save may rewrites `LinearMap.exact_iff`

Mathlib/RingTheory/Flat/CategoryTheory.lean
```diff
@@ -0,0 +1,193 @@
...
+      (DFunLike.ext _ _ h.apply_apply_eq_zero)) (ModuleCat.shortComplex_exact _ _ $ Eq.symm $
+        LinearMap.exact_iff |>.1 $ h)
+    rw [LinearMap.exact_iff, Eq.comm]
+    rw [ModuleCat.iff_shortComplex_exact] at H
+    convert H⟩
+
+open scoped MonoidalCategory in
+set_option maxHeartbeats 400000 in
+noncomputable instance [flat : Flat R M] {X Y : ModuleCat.{u} R} (f : X ⟶ Y) :
+    Limits.PreservesLimit (Limits.parallelPair f 0) (tensorLeft M) where
```
joelriou, line 94: After you have obtained `iff_lTensor_preserves_shortComplex_exact`, I believe that the rest should be obtained from a (mission) API about additive functors between abelian categories: their preserves exactness iff they preserves both finite limits and colimits. Then, I think we should develop this API for general characterizations of left exact/right exact/exact functors between abelian categories in terms of their behaviour on exact short complexes.

joelriou: A suggestion is to keep in this PR only the beginning of the file until `iff_rTensor_preserves_shortComplex_exact`. Then, depending on how you feel about the general API I am suggesting above, you may do it, or I may do it if you prefer.

jjaassoonn: I will try to open a PR about this suggestion this week!

jjaassoonn: I have removed everything else except `iff_rTensor_preserves_shortComplex_exact ` and above. I think I will wait for #14520 to golf the proof a bit more

jjaassoonn: > A suggestion is to keep in this PR only the beginning of the file until `iff_rTensor_preserves_shortComplex_exact`. Then, depending on how you feel about the general API I am suggesting above, you may do it, or I may do it if you prefer.

I would want to use this as an opportunity to learn more about the homological part of the library, but my code would probably not be optimal. If you think it would be more efficient to open the PR yourself, I wouldn't mind either.

joelriou: Please do it!

Mathlib/RingTheory/Flat/CategoryTheory.lean
```diff
@@ -0,0 +1,69 @@
...
+
+lemma rTensor_shortComplex_exact [Flat R M] (C : ShortComplex $ ModuleCat R) (hC : C.Exact) :
+    C.map (tensorRight M) |>.Exact := by
+  rw [moduleCat_exact_iff_function_exact] at hC ⊢
+  exact rTensor_exact M hC
+
+lemma iff_lTensor_preserves_shortComplex_exact :
+    Flat R M ↔
+    ∀ (C : ShortComplex $ ModuleCat R) (_ : C.Exact), (C.map (tensorLeft M) |>.Exact) :=
+  ⟨fun _ _=> lTensor_shortComplex_exact _ _, fun H => iff_lTensor_exact.2 $
```
joelriou, line 52:
```suggestion
  ⟨fun _ _ => lTensor_shortComplex_exact _ _, fun H => iff_lTensor_exact.2 $
```

Mathlib/RingTheory/Flat/CategoryTheory.lean
```diff
@@ -0,0 +1,69 @@
...
+    fun _ _ _ _ _ _ _ _ _ f g h =>
+      moduleCat_exact_iff_function_exact _ |>.1 $
+      H (.mk (ModuleCat.ofHom f) (ModuleCat.ofHom g)
+        (DFunLike.ext _ _ h.apply_apply_eq_zero))
+          (moduleCat_exact_iff_function_exact _ |>.2 h)⟩
+
+lemma iff_rTensor_preserves_shortComplex_exact :
+    Flat R M ↔
+    ∀ (C : ShortComplex $ ModuleCat R) (_ : C.Exact), (C.map (tensorRight M) |>.Exact) :=
+  ⟨fun _ _=> rTensor_shortComplex_exact _ _, fun H => iff_rTensor_exact.2 $
```
joelriou, line 62:
```suggestion
  ⟨fun _ _ => rTensor_shortComplex_exact _ _, fun H => iff_rTensor_exact.2 $
```

joelriou:
Thanks!

bors merge

PR #14483, opened 2024-07-06
[Merged by Bors] - feat: monoid action on finprod

kbuzzard:
---

Is adding this import OK? If not then it seems to me that this lemma needs to go by itself in a new file.


Mathlib/GroupTheory/GroupAction/BigOperators.lean
```diff
@@ -67,4 +68,9 @@ theorem Finset.smul_prod {r : α} {f : γ → β} {s : Finset γ} :
   map_prod (MulDistribMulAction.toMonoidHom β r) f s
 #align finset.smul_prod Finset.smul_prod
 
+theorem smul_finprod [Finite γ] {f : γ → β} (r : α) :
```
eric-wieser, line 71: I think you want to state this for `γ : Sort*`?

kbuzzard: ha ha that might be the case but then I can't prove it :-)

eric-wieser: I think you then have to use `cases nonempty_fintype (PLift γ)`, and you can probably work out the rest by looking for other uses of that pattern.

eric-wieser:
bors merge

Thanks!

PR #14491, opened 2024-07-07
[Merged by Bors] - chore: soften `set_option` linter message

adomani:
[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60pp.2Emvars.60.20forbidden.3F/near/449666047)

---

### DIFF
--- a/Mathlib/Tactic/Linter/Style.lean
+++ b/Mathlib/Tactic/Linter/Style.lean
@@ -65,8 +65,8 @@ def setOptionLinter : Linter where run := withSetOptionIn fun stx => do
           Linter.logLint linter.setOption head
             m!"Setting options starting with '{"', '".intercalate (forbidden.map (·.toString))}' \
                is only intended for development and not for final code. \
-               If you intend to submit this contribution to the Mathlib project, please remove
-               'set_option {name}'."
+               If you intend to submit this contribution to the Mathlib project, \
+               please remove 'set_option {name}'."
 
 initialize addLinter setOptionLinter
 
--- a/Mathlib/Tactic/Linter/Style.lean
+++ b/Mathlib/Tactic/Linter/Style.lean
@@ -61,7 +61,10 @@ def setOptionLinter : Linter where run := withSetOptionIn fun stx => do
     if let some (head) := stx.find? is_set_option then
       if let some (name) := parse_set_option head then
         if #[`pp, `profiler, `trace, `debug].contains name.getRoot then
-          Linter.logLint linter.setOption head m!"Forbidden set_option `{name}`; please remove"
+          Linter.logLint linter.setOption head
+            m!"'set_option {name}' is only intended for development: \
+               unless you remove this 'set_option' or you silence the linter, \
+               `Mathlib` CI will flag this use."
 
 initialize addLinter setOptionLinter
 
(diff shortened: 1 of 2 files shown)


semorrison:
I think this text is fine. An alternative would be 

"This `set_option` command should not be committed as part of a PR. Before submitting a PR, please remove the `set_option`. (It will cause the PR to be rejected in CI.)

semorrison:
bors d+

kmill:
Another suggestion: "Setting options starting with 'pp', 'profiler', 'trace', or 'debug' is not allowed in contributions to the Mathlib project. Before submitting a PR, please remove this 'set_option' command."

grunweg:
Thanks for doing this; the new message is clearer. Let me join the word-smithing bikeshed: what do you think about
`set_option {name}' is only intended for development, not for final code: therefore, it is not allowed in contributions to the mathlib project.`

I arrived at this phrasing because
- it does not name the list of all options (I don't view this list as exhaustive; I imagine it could go stale quickly)*
- it avoids the jargon-y terms "PR" or "CI" (which are not quite the point imho: a direct push to mathlib master with this option would be just as undesirable)

I am not fully happy with this mentioning `mathlib` at all (in downstream projects, this is slightly misleading). One *could* have different errors for mathlib and downstream:
- in mathlib, take on of the proposed phrasings
- downstream, just say "please consider removing this option"

* For instance, I have toyed with the long-term goal of also making it outlaw `autoImplicit` in most of mathlib (i.e., enforcing what is currently said in code review): this is a different kind of forbidden option.

adomani:
I just pushed yet another version of the message:
```
warning: Setting options starting with 'debug', 'pp', 'profiler', 'trace' is only intended
for development and not for final code. If you intend to submit this contribution to the
Mathlib project, please remove 'set_option trace.profiler.output'.
note: this linter can be disabled with `set_option linter.setOption false`
```

I like explicitly mentioning what the rules of the linter are, hence the message contains `'debug', 'pp', 'profiler', 'trace'`.  This list is embedded in the code in such a way that the natural way of modifying it will also update automatically the message, maintaining it in sync.

I agree with the desire to avoid PR and CI in the message, but I think that mentioning `Mathlib` is important, since when submitting there the command should be removed.

I like mentioning explicitly also what exact `set_option` was used and the "please remove ...` does that.

Mathlib/Tactic/Linter/Style.lean
```diff
@@ -58,10 +58,15 @@ def setOptionLinter : Linter where run := withSetOptionIn fun stx => do
...
-          Linter.logLint linter.setOption head m!"Forbidden set_option `{name}`; please remove"
+    if let some head := stx.find? is_set_option then
+      if let some name := parse_set_option head then
+        let forbidden := [`debug, `pp, `profiler, `trace]
+        if forbidden.contains name.getRoot then
+          Linter.logLint linter.setOption head
+            m!"Setting options starting with '{"', '".intercalate (forbidden.map (·.toString))}' \
+               is only intended for development and not for final code. \
+               If you intend to submit this contribution to the Mathlib project, please remove
+               'set_option {name}'."
```
adomani, line 69:
```suggestion
               If you intend to submit this contribution to the Mathlib project, \
               please remove 'set_option {name}'."
```

PR #14492, opened 2024-07-07
[Merged by Bors] - doc: fix diagram formatting

Ruben-VandeVelde:
---

### DIFF
--- a/Mathlib/CategoryTheory/Limits/Shapes/Pullback/Pasting.lean
+++ b/Mathlib/CategoryTheory/Limits/Shapes/Pullback/Pasting.lean
@@ -46,11 +46,13 @@ variable (h₁ : i₁ ≫ g₁ = f₁ ≫ i₂) (h₂ : i₂ ≫ g₂ = f₂ ≫
 
 /-- Given
 
+```
 X₁ - f₁ -> X₂ - f₂ -> X₃
 |          |          |
 i₁         i₂         i₃
-∨          ∨          ∨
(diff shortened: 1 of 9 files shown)


grunweg:
Thanks for going over these! The changes seem plausible --- but I have no way to verify. Is there any way to look at the new documention, for instance, to check that the diagrams render better now?

Mathlib/CategoryTheory/Limits/Shapes/Pullback/Pasting.lean
```diff
@@ -46,11 +46,13 @@ variable (h₁ : i₁ ≫ g₁ = f₁ ≫ i₂) (h₂ : i₂ ≫ g₂ = f₂ ≫
 
 /-- Given
 
+```
 X₁ - f₁ -> X₂ - f₂ -> X₃
```
eric-wieser, line 50: Perhaps worth doing the horizontal arrows too

eric-wieser (approved):
bors d+

Thanks!

Mathlib/CategoryTheory/Limits/Shapes/Pullback/Pasting.lean
Ruben-VandeVelde: Note that the `f`s are inside the arrow, so I'm not sure whether that'd be more or less readable

Ruben-VandeVelde:
bors merge

PR #14497, opened 2024-07-07
[Merged by Bors] - fix: app unexpander for Opposite.op and pp_nodot for Opposite.unop

TwoFX:
Also adds a test that `op`, `unop` and also `Functor.map` (since that was broken for a long time as well) are pretty-printed correctly.

---


test/CategoryTheory/Notation.lean
eric-wieser, line 1: Perhaps `PrettyPrinting.lean`?

eric-wieser (approved):
bors d+

Just an optional rename suggestion.

PR #14463, opened 2024-07-05
[Merged by Bors] - fix(SetTheory/Game/Basic): Fix the definition of surreal inverse

hwatheod:
fix the definition of surreal inverse

---
Details in this [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Surreal.20inverse.20formula.20in.20mathlib4.20is.20incorrect).  This PR is split out of #14498 which contains the full proof that this corrected definition is indeed a multiplicative inverse.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14498, opened 2024-07-07
feat(SetTheory/Surreal/Division): define the field structure on the Surreal numbers

hwatheod:
define the field structure on the Surreal numbers

---

- [x] depends on: #14463
- [x] depends on: #14600 
- [x] depends on: #14517 

In this PR, we prove that the inverse defined in `SetTheory/Game/Basic` is indeed a multiplicative inverse. The bulk of the proof is in the new file `SetTheory/Surreal/Division`. Right now, the proof is a bit repetitive as there are usually many similar cases which differ only in small places like Left vs Right. Suggestions for how to improve the proof are welcome.


Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,853 @@
...
+
+lemma pos_num_eq_normalization :
+    Surreal.mk x x_num = Surreal.mk (normalization x) (num_of_normalization_num x_num x_pos) := by
+  simp only [mk_eq_mk]
+  rw [equiv_def]
+  rcases x with ⟨xl, xr, xL, xR⟩
+  simp only [normalization, insertLeft]
+  constructor <;> (
+    rw [le_def]
+    simp
```
eric-wieser, line 118: This is a non-terminal simp, and should be squeezed to a `simp only`

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,757 @@
...
+variable {x : PGame} (x_num : x.Numeric)
+
+namespace Surreal.Division
+section Positive
+
+variable (x_pos : 0 < x)
+
+/-! ### A few random positivity lemmas -/
+
+/-- This applies to an arbitrary `LinearCommOrderedRing` and probably should go elsewhere. -/
```
semorrison, line 73: Indeed!

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,757 @@
...
+  · exact x_pos
+  · exact numeric_zero
+  · exact x_num
+
+/-! ### Normalization of a positive numeric game -/
+
+/-- If `x` is a positive numeric game, then keeping only the positive Left options then inserting
+  0 as a Left option, results in an equal game. This game is called the normalization of `x`.
+-/
+def normalization (x : PGame) : PGame :=
```
semorrison, line 93: I've forgotten my ONAG: if this is useful beyond setting up the field structure, perhaps this could preemptively be moved to its own file? Otherwise fine as is.

semorrison:
Looking pretty good structurally, but I haven't read the proofs at all. 

Would someone from amongst @timotree3, @FR-vdash-bot, @vihdzp, who have been thinking about games and surreals recently, be able to give a review? 

Particularly useful would be checking that the proofs look reasonable, and perhaps suggesting names for lemmas. (I'm a little uncomfortable with `eq1` etc and `onag` in names, but perhaps just making these private is good enough, or even the fact that they are already well namespaced.)

Mathlib/SetTheory/Surreal/Division.lean
hwatheod: I don't see this normalization used in ONAG beyond setting up the field structure.

vihdzp:
I must admit that I've never actually read ONAG past the first few pages. My interest of games in Lean was motivated foremost by the fact their API is very similar to that of ordinals (in fact, ordinals are just games when you remove either left or right options!), which is a topic I'm much more familiar in. That said, I really don't want this PR to rot over the following months. I'll try reading up to surreal division throughout the next week or so, and if no one else has stepped in by then, I'll give a proper PR review.

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1619,7 +1619,7 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
 /-- The pregame constructed by inserting `x'` as a new left option into x. -/
 def insertLeft (x x' : PGame.{u}) : PGame :=
   match x with
-  | mk xl xr xL xR => mk (xl ⊕ PUnit) xr (Sum.elim xL fun _ => x') xR
```
vihdzp, line 1622: Can we use `Option xl` instead?

Mathlib/SetTheory/Surreal/Multiplication.lean
```diff
@@ -531,4 +531,16 @@ noncomputable instance : LinearOrderedCommRing Surreal where
   mul_pos := by rintro ⟨x⟩ ⟨y⟩; exact x.2.mul_pos y.2
   decidableLE := Classical.decRel _
 
+lemma one_def : 1 = mk 1 numeric_one := by rfl
```
vihdzp, line 534:
```suggestion
lemma one_def : 1 = mk 1 numeric_one := rfl
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+  fun j => lt_trans x_pos (Numeric.lt_moveRight x_num j)
+
+/-! ### Normalization of a positive numeric game -/
+
+/-- If `x` is a positive numeric game, then keeping only the positive Left options then inserting
+  0 as a Left option, results in an equal game. This game is called the normalization of `x`.
+-/
+def normalization (x : PGame) : PGame :=
+  match x with
+  | ⟨_, r, L, R⟩ => insertLeft ⟨{i // 0 < L i}, r, fun i => L i, R⟩ 0
```
vihdzp, line 89:
```suggestion
def normalization : PGame → PGame
  | ⟨_, r, L, R⟩ => insertLeft ⟨{i // 0 < L i}, r, fun i => L i, R⟩ 0
```

vihdzp: I think it might be useful to separate out `⟨{i // 0 < L i}, r, fun i => L i, R⟩` into another auxiliary definition, maybe `normalization'`. That way you can separate the proofs of `normalization'` and `normalization` being numbers, and equivalent to the numbers they normalize. Somewhere along those lines, you'd use
```
theorem insertLeft_congr_left {x y z : PGame} (h : x ≈ y) : insertLeft x z ≈ insertLeft y z := by
  sorry
```
which is currently missing from the library.

Note that `Numeric.normalization'` does not require `0 < x`.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+open SetTheory
+open PGame
+open Equiv
+
+variable {x : PGame} (x_num : x.Numeric)
+
+namespace Surreal.Division
+section Positive
+
+variable (x_pos : 0 < x)
```
vihdzp, line 69: It's not good practice to have conditions on variables. You might then miss cases when they're not strictly necessary. For instance, `Numeric.normalization` only requires `0 ≤ x` (though I guess it does make sense to instead require `0 < x`).

vihdzp (changes requested):
Initial remarks on `normalize`.

Mathlib/SetTheory/Surreal/Division.lean
vihdzp: Actually, my `normalization'` isn't always equivalent to the original game, consider {0 | }. So this might not have been such a good suggestion.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      right
+      use j
+  · constructor
+    · intro i
+      left
+      left
+      use i
+    · intro j
+      right
+      use j
```
vihdzp, line 140:
```suggestion
lemma eq_normalization_of_pos:
    mk x x_num = mk (normalization x) (num_of_normalization_num x_num x_pos) := by
  rw [mk_eq_mk]
   -- we show: x ≈ x.insertLeft 0 ≈ normalization x
  apply trans <| symm <| insertLeft_equiv_of_lf x_pos.lf
  rw [equiv_def]
  rcases x with ⟨xl, xr, xL, xR⟩
  simp only [normalization, insertLeft]
  constructor <;> (
    rw [le_def]
    simp only [leftMoves_mk, moveLeft_mk, Sum.exists, Sum.elim_inl, Subtype.exists,
      Sum.elim_inr, exists_const, Sum.forall, le_refl, or_true, IsEmpty.exists_iff,
      or_false, implies_true, and_true, moveRight_mk]
  ) <;>
  constructor <;>
  intro i
  · left
    cases' le_or_gf (xL i) 0 with h h
    · right; exact h
    · left; use i, h.lt numeric_zero (x_num.moveLeft _)
  · right; use i
  · left; left; use i
  · right; use i
```
It might be possible to golf this further, `le_def` is usually strictly worse than `le_iff_forall_lf` or `le_iff_forall_lt`.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+lemma inv'_numeric (ih_xl : ∀ i, 0 < x.moveLeft i → (x.moveLeft i).inv'.Numeric)
+    (ih_xr : ∀ j, (x.moveRight j).inv'.Numeric) :
+    (∀ i, (x.inv'.moveLeft i).Numeric) ∧ (∀ j, (x.inv'.moveRight j).Numeric) := by
+  constructor <;> (
+    intro i
+    rcases x with ⟨xl, xr, xL, xR⟩
+    exact invVal_numeric ih_xl ih_xr x_num i
+  )
+
+lemma inv'_numeric_left (ih_xl : ∀ i, 0 < x.moveLeft i → (x.moveLeft i).inv'.Numeric)
```
vihdzp, line 179: Is it really worth it having these separate left and right statements if they can be so easily gotten out of `inv'_numeric`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+  during the main proof.
+
+-/
+
+/-- Given `y''`, return `(x', x'_inv, y')` -/
+def components {l r} {L : l → PGame} {R : r → PGame}
+    (h3 : ∀ i, 0 < L i → (L i).inv'.Numeric)
+    (h4 : ∀ j, (R j).inv'.Numeric) (h5 : (PGame.mk l r L R).Numeric) {b : Bool}
+    (x_pos : 0 < PGame.mk l r L R)
+    (i'' : InvTy {i // 0 < L i} r b) :=
```
vihdzp, line 200:
```suggestion
    (i'' : InvTy {i // 0 < L i} r b) : Surreal × Surreal × Surreal :=
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+
+/-! ### Helper lemmas for the main proof
+
+  These lemmas do the heavy lifting within the proof of the main result `onag_1_10`.
+  They are stated with additional hypotheses which will be satisfied by induction
+  during the main proof.
+
+-/
+
+/-- Given `y''`, return `(x', x'_inv, y')` -/
```
vihdzp, line 195: Can you specify what these variables represent in the docstring?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+    (mk (L i) (h5.2.1 i), mk (L i).inv' (h3 i i.2),
+      mk ((PGame.mk l r L R).inv'.moveRight j) (by apply inv'_numeric_right <;> tauto))
+  | InvTy.right₁ i j =>
+    (mk (L i) (h5.2.1 i), mk (L i).inv' (h3 i i.2),
+      mk ((PGame.mk l r L R).inv'.moveLeft j) (by apply inv'_numeric_left <;> tauto))
+  | InvTy.right₂ i j =>
+    (mk (R i) (h5.2.2 i), mk (R i).inv' (h4 i),
+      mk ((PGame.mk l r L R).inv'.moveRight j) (by apply inv'_numeric_right <;> tauto))
+
+/-- The identity `1 - x*y'' = (1 - x * y')*(x' - x)/x'` -/
```
vihdzp, line 216: This, if perhaps not in this preliminary form, seems like a potentially useful equality later down the line. So I think it makes sense to try and come up with a better name. Maybe `one_neg_div_option`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+lemma eq1 {l r} {L : l → PGame} {R : r → PGame}
+    (h3 : ∀ i, 0 < L i → (L i).inv'.Numeric)
+    (h4 : ∀ j, (R j).inv'.Numeric) (h5 : (PGame.mk l r L R).Numeric) {b : Bool}
+    (x_pos : 0 < PGame.mk l r L R)
+    (inv_l : ∀ (i : { i // 0 < L i }), mk (L i) (h5.2.1 i) * mk (L i).inv' (h3 i i.2) = 1)
+    (inv_r : ∀ j, mk (R j) (h5.2.2 j) * mk (R j).inv' (h4 j) = 1)
+    (i'' : InvTy {i // 0 < L i} r b) :
+    let (x', x'_inv, y') := components h3 h4 h5 x_pos i''
```
vihdzp, line 224: This is a really unwieldy type signature. Can you put that in a separate def with a docstring explaining what's going on?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+    let x := mk (PGame.mk l r L R) h5
+    let y'' := match b with
+      | false => mk ((PGame.mk l r L R).inv'.moveLeft i'')
+        (by apply inv'_numeric_left <;> tauto)
+      | true => mk ((PGame.mk l r L R).inv'.moveRight i'')
+        (by apply inv'_numeric_right <;> tauto)
+    match i'' with
+    | InvTy.zero => true
+    | _ => 1 - x*y'' = (1 - x * y')*(x' - x)*x'_inv := by
```
vihdzp, line 233:
```suggestion
    | _ => 1 - x * y'' = (1 - x * y') * (x' - x) * x'_inv := by
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+lemma eq1 {l r} {L : l → PGame} {R : r → PGame}
+    (h3 : ∀ i, 0 < L i → (L i).inv'.Numeric)
+    (h4 : ∀ j, (R j).inv'.Numeric) (h5 : (PGame.mk l r L R).Numeric) {b : Bool}
+    (x_pos : 0 < PGame.mk l r L R)
+    (inv_l : ∀ (i : { i // 0 < L i }), mk (L i) (h5.2.1 i) * mk (L i).inv' (h3 i i.2) = 1)
+    (inv_r : ∀ j, mk (R j) (h5.2.2 j) * mk (R j).inv' (h4 j) = 1)
+    (i'' : InvTy {i // 0 < L i} r b) :
+    let (x', x'_inv, y') := components h3 h4 h5 x_pos i''
+    let x := mk (PGame.mk l r L R) h5
+    let y'' := match b with
```
vihdzp, line 226: This could be turned into a def too.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      | false => mk ((PGame.mk l r L R).inv'.moveLeft i'')
+        (by apply inv'_numeric_left <;> tauto)
+      | true => mk ((PGame.mk l r L R).inv'.moveRight i'')
+        (by apply inv'_numeric_right <;> tauto)
+    match i'' with
+    | InvTy.zero => true
+    | _ => 1 - x*y'' = (1 - x * y')*(x' - x)*x'_inv := by
+  cases i'' <;> simp only [components]
+  all_goals
```
vihdzp, line 235: If I understand correctly, what this inductive proof does is
a) take care of all numericity hypotheses first
b) cast your values into elements of `Surreal*`
c) delegate all the actual calculations to `field_simp; ring`
Can you add a few short comments in the proof explaining this?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+  These lemmas do the heavy lifting within the proof of the main result `onag_1_10`.
+  They are stated with additional hypotheses which will be satisfied by induction
+  during the main proof.
+
+-/
+
+/-- Given `y''`, return `(x', x'_inv, y')` -/
+def components {l r} {L : l → PGame} {R : r → PGame}
+    (h3 : ∀ i, 0 < L i → (L i).inv'.Numeric)
+    (h4 : ∀ j, (R j).inv'.Numeric) (h5 : (PGame.mk l r L R).Numeric) {b : Bool}
```
vihdzp, line 198: Maybe something like this?
```suggestion
    (hnum_L : ∀ i, 0 < L i → (L i).inv'.Numeric)
    (hnum_R : ∀ j, (R j).inv'.Numeric) (hnum : (PGame.mk l r L R).Numeric) {b : Bool}
```
Really, anything more descriptive would be nice. Alternatively, you could turn these hypothesis into `def`s like in the surreal multiplication proof.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+    specialize inv_l i
+    have : IsUnit (mk (L i) (h5.2.1 i)) := by rw [isUnit_iff_exists_inv]; exact ⟨_, inv_l⟩
+    lift (mk (L i) (h5.2.1 i)) to Surrealˣ using this with s
+    have : mk (L i).inv' (h3 i i.2) = ↑s⁻¹ := Eq.symm (Units.inv_eq_of_mul_eq_one_right inv_l)
+  all_goals
+    rw [this]
+    field_simp
+    ring
+
+lemma onag_1_10_i' {l r} {L : l → PGame} {R : r → PGame}
```
vihdzp, line 259: Most of the same comments on the previous proof apply here. Maybe this could be named `one_between_mul_options`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      apply mul_neg_of_neg_of_pos
+      · apply mul_neg_of_neg_of_pos
+        · exact ih
+        · simp only [sub_pos, mk_lt_mk]
+          apply Numeric.lt_moveRight
+          exact h5
+      · exact ihr_pos j
+
+/-- The identity `x' * y + x * y' - x' * y' = 1 + x' * (y - y'')` -/
+lemma eq2 {l r} {L : l → PGame} {R : r → PGame}
```
vihdzp, line 335: `mul_option_eq_one_add_mul`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+    specialize inv_l i
+    have : IsUnit (mk (L i) (h5.2.1 i)) := by rw [isUnit_iff_exists_inv]; exact ⟨_, inv_l⟩
+    lift (mk (L i) (h5.2.1 i)) to Surrealˣ using this with s
+    have : mk (L i).inv' (h3 i i.2) = ↑s⁻¹ := Eq.symm (Units.inv_eq_of_mul_eq_one_right inv_l)
+  all_goals
+    rw [this]
+    field_simp
+    ring
+
+lemma onag_1_10_iii_left' {l r} {L : l → PGame} {R : r → PGame}
```
vihdzp, line 385: `mul_option_lt_one`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      rw [this]
+      simp only [add_lt_iff_neg_left]
+      apply mul_neg_of_pos_of_neg
+      · exact right_pos_of_pos h5 x_pos i
+      · apply sub_neg.mpr
+        have := mk_lt_mk_moveRight inv_numeric (InvTy.right₂ i j)
+        conv_rhs at this => simp [inv']
+        exact this
+
+lemma onag_1_10_iii_right' {l r} {L : l → PGame} {R : r → PGame}
```
vihdzp, line 445: `one_lt_mul_option`?

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      apply Left.mul_pos
+      · exact right_pos_of_pos h5 x_pos i
+      · apply sub_pos.mpr
+        have := mk_moveLeft_lt_mk inv_numeric (InvTy.left₁ i j)
+        conv_rhs at this => simp [inv']
+        exact this
+
+/-! ### The main lemma -/
+
+lemma onag_1_10 :
```
vihdzp, line 507: `numeric_inv'_and_inv_surreal`? Honestly it's probably fine to just name this one `main`.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+
+    have onag_1_10_iv : x * x.inv' ≈ 1 := by
+      suffices mk x x_num * mk x.inv' onag_1_10_ii = 1 by
+        simp only [← mk_mul, one_def, mk_eq_mk] at this
+        exact this
+      rwa [pos_num_eq_normalization x_num x_pos]
+
+    exact ⟨onag_1_10_ii, onag_1_10_iv⟩
+
+lemma onag_1_10_ii : (x.inv').Numeric := (onag_1_10 x_num x_pos).1
```
vihdzp, line 633:
```suggestion
lemma Numeric.inv' : (x.inv').Numeric := (onag_1_10 x_num x_pos).1
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+
+lemma onag_1_10_iv : x * x.inv' ≈ 1 := (onag_1_10 x_num x_pos).2
```
vihdzp, line 635:
```suggestion
lemma mul_inv'_self : x * x.inv' ≈ 1 := (onag_1_10 x_num x_pos).2
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+
+end Positive
+
+lemma inv_surreal (hx : ¬ x ≈ 0) : x * x⁻¹ ≈ 1 := by
```
vihdzp, line 639:
```suggestion
protected lemma mul_inv_self (hx : ¬ x ≈ 0) : x * x⁻¹ ≈ 1 := by
```

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+    simp only [quot_neg_mul, quot_mul_neg] at this ⊢
+    exact this
+
+end Surreal.Division
+
+namespace SetTheory.PGame
+open Surreal.Division
+
+lemma Numeric.inv (x_num : x.Numeric) : x⁻¹.Numeric := by
+  rcases lf_or_equiv_or_gf x 0 with neg | zero | pos
```
vihdzp, line 664: Is it simpler to use `lt_or_equiv_or_gt`?

vihdzp (commented):
I haven't really delved into the main proof just yet. I think there's a lot that can be improved in readability beforehand.

Mathlib/SetTheory/Surreal/Division.lean
```diff
@@ -0,0 +1,714 @@
...
+      · apply Numeric.add
+        · exact numeric_one
+        · apply Numeric.mul
+          · apply Numeric.sub
+            · first | exact h5.2.1 i | exact h5.2.2 i
+            · exact h5
+          · exact ih
+      · first | exact h4 i | exact h3 i i.2
+
+lemma inv'_numeric (ih_xl : ∀ i, 0 < x.moveLeft i → (x.moveLeft i).inv'.Numeric)
```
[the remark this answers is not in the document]
vihdzp: This name is a bit too close to that of `Numeric.inv'`, so maybe something like `numeric_inv'_of_numeric_options` is better.

vihdzp:
I opened #15779 which should help golf down the results on `normalization` a bit.

Mathlib/SetTheory/Game/PGame.lean
vihdzp: My PR #15779 does just that.

Mathlib/SetTheory/Surreal/Division.lean
hwatheod: With `protected`, this doesn't work - then Lean cannot find this `mul_inv_self` when it is used later in this file. I removed the `protected` and it works.

vihdzp: The `protected` keyword makes it so that you have to specify the namespace explicitly, i.e. `Surreal.mul_inv_self`. That way, the name of this lemma can correspond to the analogous [`mul_inv_self`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#mul_inv_self) in the root namespace without clashing with it.

hwatheod: According to [here](https://github.com/leanprover-community/mathlib4/blob/74ed7693458847faae657f3489f6d1d228607280/Mathlib/Algebra/Group/Defs.lean#L1073) `mul_inv_self` is deprecated in favor of `mul_inv_cancel` so perhaps I should use `mul_inv_cancel` here also ?

hwatheod: I changed `mul_inv_self` to `mul_inv_cancel`, added `protected` keyword, and also renamed `mul_inv'_self` to `mul_inv'_cancel`.

PR #14499, opened 2024-07-07
[Merged by Bors] - Removes non-printable characters U+200b and U+200c from docstrings

adomasbaliuka:
Mathlib contains a few non-printable unicode characters in some docstrings. I'm not sure why these were put in (some of them seem to have been ported from Mathlib 3 and it's hard to trace back where they were added first).

The characters removed here are
- U+200b (​zero width space) 
- U+200c (zero-width non-joiner)

It might make sense to remove them, see the [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Whitelist.20for.20Unicode.3F).

If you review this PR, please also comment on whether you think a linter concerned with Unicode would be helpful. (e.g., warn on any *new* unicode characters being added to Mathlib, or, e.g., warn on any unicode characters being added that don't have a way to type using `\<combination>`)

---

### DIFF
--- a/Mathlib/Data/List/Defs.lean
+++ b/Mathlib/Data/List/Defs.lean
@@ -453,7 +453,7 @@ def map₂Right' (f : Option α → β → γ) (as : List α) (bs : List β) : L
 
 
 /-- Left-biased version of `List.map₂`. `map₂Left f as bs` applies `f` to each pair
-`aᵢ ∈ as` and `bᵢ ‌∈ bs`. If `bs` is shorter than `as`, `f` is applied to `none`
+`aᵢ ∈ as` and `bᵢ ∈ bs`. If `bs` is shorter than `as`, `f` is applied to `none`
 for the remaining `aᵢ`.
 
 ```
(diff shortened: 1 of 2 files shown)


Mathlib/Data/List/Defs.lean
```diff
@@ -472,7 +472,7 @@ def map₂Left (f : α → Option β → γ) : List α → List β → List γ
 #align list.map₂_left List.map₂Left
 
 /-- Right-biased version of `List.map₂`. `map₂Right f as bs` applies `f` to each
-pair `aᵢ ∈ as` and `bᵢ ‌∈ bs`. If `as` is shorter than `bs`, `f` is applied to
+pair `aᵢ ∈ as` and `bᵢ ∈ bs`. If `as` is shorter than `bs`, `f` is applied to
```
eric-wieser, line 475: These ones came from https://github.com/leanprover-community/mathlib/pull/4512

eric-wieser (approved):
bors merge

Thanks!

I think a good start would be a linter against invisible unicode characters; I think banning ones not in vscode would be disruptive, since synchronizing releases would be awkward.

PR #14489, opened 2024-07-07
[Merged by Bors] - feat(Function/LpSpace): continuity of `indicatorConstLp` in the set

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14500, opened 2024-07-07
[Merged by Bors] - feat(LpSpace/ContinuousCompMeasurePreserving): new file

urkud:
Prove that the composition of an `L^p` function
with a measure preserving continuous function
is continuous in both arguments.

---

- [x] depends on: #14487
- [ ] depends on: #14489
- [ ] depends on: #14592


Mathlib/MeasureTheory/Function/LpSpace/ContinuousCompMeasurePreserving.lean
```diff
@@ -0,0 +1,165 @@
...
+Authors: Yury Kudryashov
+-/
+import Mathlib.MeasureTheory.Function.SimpleFuncDenseLp
+import Mathlib.MeasureTheory.Measure.Regular
+
+/-!
+# Continuity of `MeasureTheory.Lp.compMeasurePreserving`
+
+In this file we prove that the composition of an `L^p` function `g`
+with a continuous measure-preservingg map `f` is continuous in both arguments.
```
semorrison, line 13:
```suggestion
with a continuous measure-preserving map `f` is continuous in both arguments.
```

semorrison:
bors d+

PR #14502, opened 2024-07-08
[Merged by Bors] - feat: Lemma for `fderiv` of scalar function

adomasbaliuka:
Adds lemma for simplifying the Fréchet derivative when applied to a function maping scalars to scalars. 

This (currently named `fderiv_deriv'`) is a special case of `fderiv_deriv`. Neither of `simp`, `fun_prop`, `aesop`, `exact?` , `apply?` currently seems to make progress on it.

Possibly, it should be instead stated as `(fderiv 𝕜 f x : 𝕜 → 𝕜) = ((deriv f x) * ·)` if that makes it more likely to be applied by `simp`?

I think it should be marked `@[simp]` because one would basically always prefer the right hand side to the left hand side of the equality. Does that make sense?

I also think `fderiv_deriv` should be marked `@[simp]`, though maybe there is a reason why it isn't. Marking it such requires a change to the proof of `deriv_fderiv`, which was using a `simp` rather than a `simp only` (not a [non-terminal simp](https://leanprover-community.github.io/glossary.html#non-terminal-simp) but still to be avoided, I would assume?). Maybe other things are also likely to break, however, if those are added to simp?

---

### DIFF
--- a/Mathlib/Analysis/Calculus/Deriv/Basic.lean
+++ b/Mathlib/Analysis/Calculus/Deriv/Basic.lean
@@ -489,12 +489,6 @@ lemma fderiv_deriv' {f : 𝕜 → 𝕜} {x y : 𝕜} : (fderiv 𝕜 f x : 𝕜 �
   simp only [ContinuousLinearMap.smulRight_apply, ContinuousLinearMap.one_apply, smul_eq_mul]
   ring
 
-@[simp]
-lemma fderiv_deriv'' {f : 𝕜 → 𝕜} {x : 𝕜} : (fderiv 𝕜 f x : 𝕜 → 𝕜) = ((deriv f x) * ·) := by
-  ext
-  simp_all only [fderiv_deriv']
-
-
 theorem norm_deriv_eq_norm_fderiv : ‖deriv f x‖ = ‖fderiv 𝕜 f x‖ := by
   simp [← deriv_fderiv]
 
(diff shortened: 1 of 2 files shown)


grunweg:
Thanks for your contribution to mathlib! I guess you'd like code to be reviewed - hence I have labelled it `awaiting-review`. (You can read more about this and related labels [here](https://leanprover-community.github.io/contribute/index.html#lifecycle-of-a-pr).

Mathlib/Analysis/Calculus/Deriv/Basic.lean
```diff
@@ -481,6 +481,12 @@ theorem fderiv_deriv : (fderiv 𝕜 f x : 𝕜 → F) 1 = deriv f x :=
 theorem deriv_fderiv : smulRight (1 : 𝕜 →L[𝕜] 𝕜) (deriv f x) = fderiv 𝕜 f x := by simp [deriv]
 #align deriv_fderiv deriv_fderiv
 
+@[simp]
+lemma fderiv_deriv' {f : 𝕜 → 𝕜} {x y : 𝕜} : (fderiv 𝕜 f x : 𝕜 → 𝕜) y = (deriv f x) * y := by
```
fpvandoorn, line 485:
```suggestion
lemma fderiv_eq_deriv_mul {f : 𝕜 → 𝕜} {x y : 𝕜} : (fderiv 𝕜 f x : 𝕜 → 𝕜) y = deriv f x * y := by
```

fpvandoorn:
This lemma looks good to me. I would like a different name though (we want to avoid primes, and also the other two names are not great).

fpvandoorn:
Actually, can you add a version for `f : 𝕜 → F` where the right-hand-side is `y \bu deriv f x`?

fpvandoorn:
And it makes sense to me to make `fderiv_deriv` a simp-lemma. Feel free to do that. That might need some fixes in other places.

adomasbaliuka:
I don't really understand why adding new `simp` lemmas breaks a lot proofs. How to judge if adding a `simp` lemma makes the situation overall better or worse?

fpvandoorn:
Thanks!

bors merge

PR #14504, opened 2024-07-08
[Merged by Bors] - feat(Topology/Sequences): preservation of sequential compactness under continuous functions 

janemms:
Add four lemmas about preservation of sequential compactness under (sequentially) continuous functions. 
[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Sequential.20compactness.20under.20sequentially.20cont.20functions/near/449031782)

---

### DIFF
--- a/Mathlib/Topology/Sequences.lean
+++ b/Mathlib/Topology/Sequences.lean
@@ -277,6 +277,41 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
 
 end FirstCountableTopology
 
+section Image
+
+/-- Sequential compactness of sets is preserved under sequentially continuous functions. -/
+theorem IsSeqCompact.image {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y] (f : X → Y)
+    (f_cont : SeqContinuous f) {K : Set X} (K_cpt : IsSeqCompact K) :
+    IsSeqCompact (f '' K) := by
+  intro ys ys_in_fK
+  choose xs xs_in_K fxs_eq_ys using fun n ↦ (ys_in_fK n)
+  simp only [Set.mem_image, exists_exists_and_eq_and]
+  obtain ⟨a, a_in_K, phi, phi_mono, xs_phi_lim⟩ := K_cpt xs_in_K
+  refine ⟨a, a_in_K, phi, phi_mono, ?_⟩
+  exact Filter.Tendsto.congr (fun x ↦ fxs_eq_ys (phi x)) (f_cont xs_phi_lim)
+
+/-- The range of sequentially continuous function on a sequentially compact space is sequentially
+compact. -/
+theorem isSeqCompact_range {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    [SeqCompactSpace X] (f : X → Y) (hf : SeqContinuous f) :
+    IsSeqCompact (Set.range f) := by
+  rw [← Set.image_univ]
+  exact IsSeqCompact.image f hf ((seqCompactSpace_iff X).mp ‹SeqCompactSpace X›)
+
+/-- Sequential compactness of sets is preserved under continuous functions. -/
+theorem IsSeqCompact.image_of_continuous {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    (f : X → Y) (f_cont : Continuous f) {K : Set X} (K_cpt : IsSeqCompact K) :
+    IsSeqCompact (f '' K) :=
+  IsSeqCompact.image f (Continuous.seqContinuous f_cont) K_cpt
+
+/-- The range of continuous function on a sequentially compact space is sequentially compact. -/
+theorem SeqCompactSpace.range_of_continuous {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    [SeqCompactSpace X] (f : X → Y) (f_cont : Continuous f) :
+    IsSeqCompact (Set.range f) :=
+  isSeqCompact_range f (Continuous.seqContinuous f_cont)
+
+end Image
+
 end SeqCompact
 
 section UniformSpaceSeqCompact
--- a/Mathlib/Topology/Sequences.lean
+++ b/Mathlib/Topology/Sequences.lean
@@ -302,7 +302,7 @@ theorem isSeqCompact_range [SeqCompactSpace X] (f_cont : SeqContinuous f) :
 theorem IsSeqCompact.image_of_continuous (f_cont : Continuous f) {K : Set X}
     (K_cpt : IsSeqCompact K) :
     IsSeqCompact (f '' K) :=
-  IsSeqCompact.image (Continuous.seqContinuous f_cont) K_cpt
+  K_cpt.image f_cont.seqContinuous
 
 /-- The range of continuous function on a sequentially compact space is sequentially compact. -/
 theorem isSeqCompact_range_of_continuous [SeqCompactSpace X] (f_cont : Continuous f) :
--- a/Mathlib/Topology/Sequences.lean
+++ b/Mathlib/Topology/Sequences.lean
@@ -243,7 +243,7 @@ theorem IsSeqCompact.subseq_of_frequently_in {s : Set X} (hs : IsSeqCompact s) {
 
 theorem SeqCompactSpace.tendsto_subseq [SeqCompactSpace X] (x : ℕ → X) :
     ∃ (a : X) (φ : ℕ → ℕ), StrictMono φ ∧ Tendsto (x ∘ φ) atTop (𝓝 a) :=
-  let ⟨a, _, φ, mono, h⟩ := seq_compact_univ fun n => mem_univ (x n)
+  let ⟨a, _, φ, mono, h⟩ := isSeqCompact_univ fun n => mem_univ (x n)
   ⟨a, φ, mono, h⟩
 
 section FirstCountableTopology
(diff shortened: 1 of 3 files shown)


Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+  exact IsSeqCompact.image f_cont ((seqCompactSpace_iff X).mp ‹SeqCompactSpace X›)
+
+/-- Sequential compactness of sets is preserved under continuous functions. -/
+theorem IsSeqCompact.image_of_continuous (f_cont : Continuous f) {K : Set X}
+    (K_cpt : IsSeqCompact K) :
+    IsSeqCompact (f '' K) :=
+  IsSeqCompact.image (Continuous.seqContinuous f_cont) K_cpt
+
+/-- The range of continuous function on a sequentially compact space is sequentially compact. -/
+theorem SeqCompactSpace.range_of_continuous [SeqCompactSpace X] (f_cont : Continuous f) :
```
urkud, line 335:
```suggestion
theorem isSeqCompact_range_of_continuous [SeqCompactSpace X] (f_cont : Continuous f) :
```

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+theorem isSeqCompact_range [SeqCompactSpace X] (f_cont : SeqContinuous f) :
+    IsSeqCompact (Set.range f) := by
+  rw [← Set.image_univ]
+  exact IsSeqCompact.image f_cont ((seqCompactSpace_iff X).mp ‹SeqCompactSpace X›)
+
+/-- Sequential compactness of sets is preserved under continuous functions. -/
+theorem IsSeqCompact.image_of_continuous (f_cont : Continuous f) {K : Set X}
+    (K_cpt : IsSeqCompact K) :
+    IsSeqCompact (f '' K) :=
+  IsSeqCompact.image (Continuous.seqContinuous f_cont) K_cpt
```
dagurtomas, line 332:
```suggestion
  K_cpt.image f_cont.seqContinuous
```

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+
+/-- The range of continuous function on a sequentially compact space is sequentially compact. -/
+theorem isSeqCompact_range_of_continuous [SeqCompactSpace X] (f_cont : Continuous f) :
+    IsSeqCompact (Set.range f) :=
+  isSeqCompact_range (Continuous.seqContinuous f_cont)
```
dagurtomas, line 337:
```suggestion
  isSeqCompact_range f_cont.seqContinuous
```

dagurtomas (approved):

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
 
 end FirstCountableTopology
 
+section Image
+
+variable {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y] {f : X → Y}
```
urkud, line 309: Some variables are already in scope.
```suggestion
variable [TopologicalSpace Y] {f : X → Y}
```

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+
+/-- Sequential compactness of sets is preserved under sequentially continuous functions. -/
+theorem IsSeqCompact.image (f_cont : SeqContinuous f) {K : Set X} (K_cpt : IsSeqCompact K) :
+    IsSeqCompact (f '' K) := by
+  intro ys ys_in_fK
+  choose xs xs_in_K fxs_eq_ys using fun n ↦ (ys_in_fK n)
+  simp only [Set.mem_image, exists_exists_and_eq_and]
+  obtain ⟨a, a_in_K, phi, phi_mono, xs_phi_lim⟩ := K_cpt xs_in_K
+  refine ⟨a, a_in_K, phi, phi_mono, ?_⟩
+  exact Filter.Tendsto.congr (fun x ↦ fxs_eq_ys (phi x)) (f_cont xs_phi_lim)
```
urkud, line 319: Minor golfing suggestion:
```suggestion
  choose xs xs_in_K fxs_eq_ys using ys_in_fK
  obtain ⟨a, a_in_K, phi, phi_mono, xs_phi_lim⟩ := K_cpt xs_in_K
  refine ⟨f a, mem_image_of_mem f a_in_K, phi, phi_mono, ?_⟩
  exact (f_cont xs_phi_lim).congr fun x ↦ fxs_eq_ys (phi x)
```

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,40 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+
+/-- The range of sequentially continuous function on a sequentially compact space is sequentially
+compact. -/
+theorem isSeqCompact_range [SeqCompactSpace X] (f_cont : SeqContinuous f) :
+    IsSeqCompact (Set.range f) := by
+  rw [← Set.image_univ]
+  exact IsSeqCompact.image f_cont ((seqCompactSpace_iff X).mp ‹SeqCompactSpace X›)
+
+/-- Sequential compactness of sets is preserved under continuous functions. -/
+theorem IsSeqCompact.image_of_continuous (f_cont : Continuous f) {K : Set X}
```
urkud, line 329: I don't think that we need `_of_continuous` lemmas. It's easy enough to use `h.image f_cont.seqContinuous`.

kkytola: Let me comment briefly, since I think I was the one who suggested to @janemms the addition of the spellings with the assumption `Continuous f`.

I of course agree that instead of applying `IsSeqCompact.image_of_continuous` or `isSeqCompact_range_of_continuous`, it will be very easy to apply `IsSeqCompact.image` or `isSeqCompact_range` by transforming a continuity hypothesis to a sequential continuity hypothesis with dot notation.

The main reason why I thought that the versions `IsSeqCompact.image_of_continuous` or `isSeqCompact_range_of_continuous` would be useful is discoverability. For example Loogle would not find the main lemma `IsSeqCompact.image` with key words `IsSeqCompact` and `Continuous` (the users would need to generalize the search term to `SeqContinuous` by themselves.) Likewise, `exact?` would not find the lemmas, if the context only has continuity hypothesis (so a user would need to understand to add `have f_seqcont := f_cont.seqContinuous` before the search succeeds). Of course a sharp user will understand that only the general version should exist in Mathlib, but if generalizing the hypothesis does not even occur to the user, this will be easy to miss. It is different from cases where the generalized hypotheses are of type class form, since then `exact?` will do the appropriate generalization on behalf of the user.

I don't have strong opinions on which one is better, just wanted to present also the argument in favor of the choice to include the alternative versions as well.

urkud: About `loogle`: I would search for `IsSeqCompact (_ '' _)`. About `exact?`: if `f` is given by a formula, then you need something like `continuity` anyway, so `exact?` won't work. Should we ask on Zulip for more opinions?

urkud: E.g., we don't add `Continuous` versions of the lemmas assuming `Measurable` or `AEMeasurable`.

kkytola: To me either choice is fine. And I'm sure if your opinion after hearing the discoverability argument is clear, then that is almost surely the right thing to do!

(Regarding your last example, however, I find the existence of both `Measurable` and `AEMeasurable` versions of some lemmas is more analogous to the current `Continuous` and `SeqContinuous` duplication question. I think a user will not be surprised that measurability needs to be derived from continuity explicitly. But a user might not realize they should derive a.e.-measurability from measurability, so having both is good.)

Moreover, `SeqContinuous` and `IsSeqCompact` are probably going to be rather rare (much more so than `Measurable` and `AEMeasurable`). So I don't think the discoverability difficulties would affect very many users in any case. And I fully trust your judgement and experience about library design!

janemms: After discussing this with @YaelDillies, I decided to remove the lemmas assuming continuity.

Mathlib/Topology/Sequences.lean
```diff
@@ -304,6 +304,28 @@ theorem CompactSpace.tendsto_subseq [CompactSpace X] (x : ℕ → X) :
...
+  obtain ⟨a, a_in_K, phi, phi_mono, xs_phi_lim⟩ := K_cpt xs_in_K
+  refine ⟨f a, mem_image_of_mem f a_in_K, phi, phi_mono, ?_⟩
+  exact (f_cont xs_phi_lim).congr fun x ↦ fxs_eq_ys (phi x)
+
+/-- The range of sequentially continuous function on a sequentially compact space is sequentially
+compact. -/
+theorem IsSeqCompact.range [SeqCompactSpace X] (f_cont : SeqContinuous f) :
+    IsSeqCompact (Set.range f) := by
+  rw [← Set.image_univ]
+  exact IsSeqCompact.image f_cont ((seqCompactSpace_iff X).mp ‹SeqCompactSpace X›)
```
YaelDillies, line 325: Can you take the opportunity to rename `SeqCompactSpace.seq_compact_univ` to `IsSeqCompact.univ` so that this becomes
```suggestion
  simpa using IsSeqCompact.univ.image f_cont
```

YaelDillies (approved):
Thanks! 🚀

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14490, opened 2024-07-07
[Merged by Bors] - feat(CategoryTheory/Adjunction): left adjoint is faithful iff unit is mono, etc.

dagurtomas:
We prove the full Lemma 4.5.13 in Riehl's *Category Theory in Context*, characterizing when a left/right adjoint is full, resp. faithful, resp fully faithful in terms of the unit/counit being various types of epi/mono/iso. Earlier we only had the statements for fully faithful functors.

---
This can be used to show that the functor from topological spaces to condensed sets is faithful, see #14455 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14455, opened 2024-07-05
[Merged by Bors] - feat:(Condensed): the functor from `TopCat` to `CondensedSet` is a faithful right adjoint.

dagurtomas:
Also adds some lemmas which improve automation in condensed mathematics.  

---
In #14514 we define compactly generated spaces and show that this functor becomes fully faithful when restricted to them.

- [x] depends on: #14490 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14506, opened 2024-07-08
[Merged by Bors] - feat(LightCondensed): the functor from `TopCat` to `LightCondSet` is a faithful right adjoint

dagurtomas:
---

- [x] depends on: #14455 
- [x] depends on: #14490 


Mathlib/Condensed/Light/Basic.lean
```diff
@@ -33,3 +33,35 @@ Light condensed sets. Because `LightProfinite` is an essentially small category,
...
+  apply Sheaf.hom_ext
+  ext
+  exact h _
+
+end LightCondensed
+
+namespace LightCondSet
+
+-- Note: `simp` can prove this when stated for `Condensed C` for a concrete category `C`.
+-- However, it doesn't seem to see through the abbreviation `CondensedSet`
```
adamtopaz, line 61: This comment should be fixed (same for the one about `LightCondMod`)

Mathlib/Condensed/Light/TopCatAdjunction.lean
```diff
@@ -0,0 +1,129 @@
...
+
+universe u
+
+open LightCondensed LightCondSet CategoryTheory
+
+attribute [local instance] ConcreteCategory.instFunLike
+
+variable (X : LightCondSet.{u})
+
+/-- Auxiliary defiinition to define the topology on `X(*)` for a light condensed set `X`. -/
```
adamtopaz, line 24:
```suggestion
/-- Auxiliary definition to define the topology on `X(*)` for a light condensed set `X`. -/
```

Mathlib/Condensed/Light/TopCatAdjunction.lean
```diff
@@ -0,0 +1,129 @@
...
+private def LightCondSet.coinducingCoprod :
+    (Σ (i : (S : LightProfinite.{u}) × X.val.obj ⟨S⟩), i.fst) →
+      X.val.obj ⟨LightProfinite.of PUnit⟩ :=
+  fun ⟨⟨S, i⟩, s⟩ ↦ X.val.map (S.const s).op i
+
+/-- Let `X` be a light condensed set. We define a topology on `X(*)` as the quotient topology of
+all the maps from light profinite sets `S` to `X(*)`, corresponding to elements of `X(S)`.
+In other words, the topology coinduced by the map `LightCondSet.coinducingCoprod` above. -/
+local instance : TopologicalSpace (X.val.obj ⟨LightProfinite.of PUnit⟩) :=
```
adamtopaz, line 38: Can you give this instance a name please? Since it's stated only as a local instance, we may need to refer to it explicitly elsewhere.

Mathlib/Condensed/Light/TopCatAdjunction.lean
```diff
@@ -0,0 +1,129 @@
...
+  TopologicalSpace.coinduced (coinducingCoprod X) inferInstance
+
+/-- The object part of the functor `LightCondSet ⥤ TopCat`  -/
+def LightCondSet.toTopCat : TopCat.{u} := TopCat.of (X.val.obj ⟨LightProfinite.of PUnit⟩)
+
+namespace LightCondSet
```
adamtopaz, line 44: I think the namespace management could be improved a bit. There are a few declarations above that could be moved into in the `LightCondSet` namespace here.

Mathlib/Condensed/Light/TopCatAdjunction.lean
```diff
@@ -0,0 +1,129 @@
...
+@[simps val_app val_app_apply]
+def topCatAdjunctionUnit (X : LightCondSet.{u}) : X ⟶ X.toTopCat.toLightCondSet where
+  val := {
+    app := fun S x ↦ {
+      toFun := fun s ↦ X.val.map (S.unop.const s).op x
+      continuous_toFun := by
+        suffices ∀ (i : (T : LightProfinite.{u}) × X.val.obj ⟨T⟩),
+          Continuous (fun (a : i.fst) ↦ X.coinducingCoprod ⟨i, a⟩) from this ⟨_, _⟩
+        rw [← continuous_sigma_iff]
+        apply continuous_coinduced_rng }
```
adamtopaz, line 106: I think it would be better to extract the `app` field as a separate definition in this case, and the naturality as a lemma. This way we could have better control over the `simps` lemmas that are generated.

adamtopaz (changes requested):
Thanks for this work! I had just a few relatively minor comments.

Mathlib/Condensed/Light/TopCatAdjunction.lean
dagurtomas: The `val_app` simp lemma is crucial to prove naturality of the unit. It's pretty ugly, of the form 
```lean
  X.topCatAdjunctionUnit.val.app S x =
    { toFun := fun s ↦ X.val.map (S.unop.const s).op x, continuous_toFun := ⋯ }
```
I tried extracting the `app` field as a separate definition, but then this lemma wouldn't fire, unless I stated it explicitly

dagurtomas: Should I try to make it work without this lemma?

adamtopaz: Oh, in that case let's just leave it as is. It would probably work if you tagged the `app` definition with `simp` (not `simps`), but that's not ideal.

adamtopaz (approved):
LGTM!

bors r+

adamtopaz:
> @dagurtomas there is nothing wrong with this PR, we are trying to finally get rid of `align` and so we are not merging PRs for a couple of hours. Feel free to merge it once we've done (or if we just gave up for today...)
> 
> bors d+

Oops! Sorry I should have read zulip first :-/

semorrison:
bors merge

PR #14513, opened 2024-07-08
[Merged by Bors] - chore: add test file to import everything

kim-em:
---

### DIFF
--- a/test/import_all.lean
+++ b/test/import_all.lean
@@ -4,12 +4,11 @@ import Qq
 import Batteries
 import Aesop
 import ProofWidgets
-import ImportGraph
 import Mathlib
 
 /-!
 Verify that Mathlib and all its upstream dependencies can be simultaneously imported.
 
-We don't `import Cli` here as that is only used by scripts,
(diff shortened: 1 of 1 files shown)


grunweg:
Good idea! (Out of curiosity: were there any bugs in the past caught by this? Or is this just a precaution?)

As a follow-up, would you like to ensure this stays in sync with the lakefile, somehow - or is that an unimportant concern? Feel free to maintainer merge on my behalf once you thought about this.

grunweg (approved):

semorrison:
Yes, it would be nice to stay in sync, but I coudn't think of an easy way to do this --- we don't want to simply scrape the manifest, because some dependencies (e.g. `Cli` and `ImportGraph`) are not used in Mathlib proper so may not have been built by the time the test runs.

I think that dependencies change slowly enough, and the test is not critical, that we can hope for the best here, at least for now.

grunweg:
Sounds reasonable - so let me
maintainer merge

j-loreaux:
bors merge

PR #14517, opened 2024-07-08
[Merged by Bors] - feat(SetTheory/Game/PGame): inserting an option a game

hwatheod:
inserting an option into a game

---
This PR is split out of #14498. Here, we define the operation of inserting a left or right option into a game, and prove a few standard results about it.

### DIFF
--- a/Mathlib/SetTheory/Game/PGame.lean
+++ b/Mathlib/SetTheory/Game/PGame.lean
@@ -1909,7 +1909,7 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
 /-- The pregame constructed by inserting `x'` as a new left option into x. -/
 def insertLeft (x x' : PGame.{u}) : PGame :=
   match x with
-  | mk xl xr xL xR => mk (Sum xl PUnit) xr (Sum.elim xL fun _ => x') xR
+  | mk xl xr xL xR => mk (xl ⊕ PUnit) xr (Sum.elim xL fun _ => x') xR
 
 /-- A new left option cannot hurt Left. -/
 lemma le_insertLeft (x x' : PGame) : x ≤ insertLeft x x' := by
@@ -1956,7 +1956,7 @@ lemma insertLeft_equiv_of_lf {x x' : PGame} (h : x' ⧏ x) : insertLeft x x' ≈
 /-- The pregame constructed by inserting `x'` as a new right option into x. -/
 def insertRight (x x' : PGame.{u}) : PGame :=
   match x with
-  | mk xl xr xL xR => mk xl (Sum xr PUnit) xL (Sum.elim xR fun _ => x')
+  | mk xl xr xL xR => mk xl (xr ⊕ PUnit) xL (Sum.elim xR fun _ => x')
 
 theorem neg_insertRight_neg (x x' : PGame.{u}) : (-x).insertRight (-x') = -x.insertLeft x' := by
   cases x
--- a/Mathlib/SetTheory/Game/PGame.lean
+++ b/Mathlib/SetTheory/Game/PGame.lean
@@ -1958,47 +1958,26 @@ def insertRight (x x' : PGame.{u}) : PGame :=
   match x with
   | mk xl xr xL xR => mk xl (Sum xr PUnit) xL (Sum.elim xR fun _ => x')
 
+theorem neg_insertRight_neg (x x' : PGame.{u}) : (-x).insertRight (-x') = -x.insertLeft x' := by
+  cases x
+  cases x'
+  dsimp [insertRight, insertLeft]
+  congr! with (i | j)
+
+theorem neg_insertLeft_neg (x x' : PGame.{u}) : (-x).insertLeft (-x') = -x.insertRight x' := by
+  rw [← neg_eq_iff_eq_neg, ← neg_insertRight_neg, neg_neg, neg_neg]
+
 /-- A new right option cannot hurt Right. -/
 lemma insertRight_le (x x' : PGame) : insertRight x x' ≤ x := by
-  rw [le_def]
-  constructor
-  · intro i
-    left
-    rcases x with ⟨xl, xr, xL, xR⟩
-    simp only [leftMoves_mk, insertRight, moveLeft_mk]
-    use i
-  · intro j
-    right
-    rcases x with ⟨xl, xr, xL, xR⟩
-    simp only [insertRight, rightMoves_mk, moveRight_mk, Sum.exists, Sum.elim_inl]
-    left
-    use j
+  rw [← neg_le_neg_iff, ← neg_insertLeft_neg]
+  exact le_insertLeft _ _
 
 /-- Adding a gift horse right option does not change the value of `x`. A gift horse right option is
  a game `x'` with `x ⧏ x'`. It is called "gift horse" because it seems like Right has gotten the
  "gift" of a new option, but actually the value of the game did not change. -/
 lemma equiv_insertRight_of_lf {x x' : PGame} (h : x ⧏ x') : x ≈ insertRight x x' := by
-  rw [equiv_def]
-  constructor
-  · rw [le_def]
-    constructor
-    · intro i
-      rcases x with ⟨xl, xr, xL, xR⟩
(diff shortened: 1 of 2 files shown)


Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,106 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
...
+    left
+    use i
+  · intro j
+    right
+    rcases x with ⟨xl, xr, xL, xR⟩
+    simp only [rightMoves_mk, moveRight_mk, insertLeft]
+    use j
+
+/-- Adding a gift horse left option does not change the value of x. -/
+lemma insertLeft_giftHorse (x : PGame) (x' : PGame) (h : x' ⧏ x) : x ≈ insertLeft x x' := by
```
eric-wieser, line 1935: I think the usual naming would be something more like
```suggestion
lemma equiv_insertLeft_of_lf {x x' : PGame} (h : x' ⧏ x) : x ≈ insertLeft x x' := by
```

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,106 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
...
+    simp only [insertLeft, leftMoves_mk, moveLeft_mk, Sum.exists, exists_const]
+    left
+    use i
+  · intro j
+    right
+    rcases x with ⟨xl, xr, xL, xR⟩
+    simp only [rightMoves_mk, moveRight_mk, insertLeft]
+    use j
+
+/-- Adding a gift horse left option does not change the value of x. -/
```
eric-wieser, line 1934: Can you explain "gift horse" in this docstring, since it has not been mentioned before?

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,110 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
       ⟩
 #align pgame.lt_iff_sub_pos SetTheory.PGame.lt_iff_sub_pos
 
+/-! ### Inserting an option -/
+
+/-- The pregame constructed by inserting x' as a new left option into x. -/
```
eric-wieser, line 1909:
```suggestion
/-- The pregame constructed by inserting `x'` as a new left option into `x`. -/
```
since these are variable names, and this renders as markdown in vscode / the web docs. Same throughout your docstrings.

eric-wieser (commented):
I think this is about as far as I can review this, as someone not that familiar in the area. The style and naming looks good now, thanks. I'll have a quick go at shortening the proofs, but probably someone else should give this the final review to check if any obvious lemmas are missing.

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,102 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
...
+  · intro j
+    right
+    rcases x with ⟨xl, xr, xL, xR⟩
+    simp only [rightMoves_mk, moveRight_mk, insertLeft]
+    use j
+
+/-- Adding a gift horse left option does not change the value of `x`. A gift horse left option is
+ a game `x'` with `x' ⧏ x`. It is called "gift horse" because it seems like Left has gotten the
+ "gift" of a new option, but actually the value of the game did not change. -/
+lemma equiv_insertLeft_of_lf {x x' : PGame} (h : x' ⧏ x) : x ≈ insertLeft x x' := by
```
semorrison, line 1933: Flip conclusion and rename `insertLeft_equiv_of_lf`, and similarly below.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -202,6 +202,32 @@ theorem numeric_of_isEmpty_rightMoves (x : PGame) [IsEmpty x.RightMoves]
   Numeric.mk (fun _ => isEmptyElim) H isEmptyElim
 #align pgame.numeric_of_is_empty_right_moves SetTheory.PGame.numeric_of_isEmpty_rightMoves
 
+/-- Inserting a smaller numeric left option into a numeric results in a numeric. -/
+theorem numeric_of_insertLeft_numeric {x x' : PGame} (x_num : x.Numeric) (x'_num : x'.Numeric)
```
semorrison, line 206: Name is wrong. Should be `insertLeft_numeric_of_numeric`, but just `insertLeft_numeric` is better. 

Similarly below.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -202,6 +202,32 @@ theorem numeric_of_isEmpty_rightMoves (x : PGame) [IsEmpty x.RightMoves]
   Numeric.mk (fun _ => isEmptyElim) H isEmptyElim
 #align pgame.numeric_of_is_empty_right_moves SetTheory.PGame.numeric_of_isEmpty_rightMoves
 
+/-- Inserting a smaller numeric left option into a numeric results in a numeric. -/
```
semorrison, line 205:
```suggestion
/-- Inserting a smaller numeric left option into a numeric game results in a numeric game. -/
```

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,87 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
...
+  | mk xl xr xL xR => mk xl (Sum xr PUnit) xL (Sum.elim xR fun _ => x')
+
+theorem neg_insertRight_neg (x x' : PGame.{u}) : (-x).insertRight (-x') = -x.insertLeft x' := by
+  cases x
+  cases x'
+  dsimp [insertRight, insertLeft]
+  congr! with (i | j)
+
+theorem neg_insertLeft_neg (x x' : PGame.{u}) : (-x).insertLeft (-x') = -x.insertRight x' := by
+  rw [← neg_eq_iff_eq_neg, ← neg_insertRight_neg, neg_neg, neg_neg]
```
eric-wieser, line 1968: I added these two theorems, since they make the proofs much shorter below.

hwatheod:
I moved `insertLeft_numeric` and `insertRight_numeric` to a position after `Numeric.neg` so I could use `Numeric.neg` to write the proof of `insertRight_numeric` in terms of `insertLeft_numeric` (like @eric-wieser did for the other file)

Mathlib/SetTheory/Game/PGame.lean
```diff
@@ -1904,6 +1904,87 @@ theorem lt_iff_sub_pos {x y : PGame} : x < y ↔ 0 < y - x :=
       ⟩
 #align pgame.lt_iff_sub_pos SetTheory.PGame.lt_iff_sub_pos
 
+/-! ### Inserting an option -/
+
+/-- The pregame constructed by inserting `x'` as a new left option into x. -/
+def insertLeft (x x' : PGame.{u}) : PGame :=
+  match x with
+  | mk xl xr xL xR => mk (xl ⊕ PUnit) xr (Sum.elim xL fun _ => x') xR
```
eric-wieser, line 1912: Any reason to not instead use
```suggestion
  | mk xl xr xL xR => mk (Option xl) xr (Option.elim xL fun _ => x') xR
```
or whatever the spelling is that works?

eric-wieser: This nicely avoids having any useless `_ : Unit` variables floating around

hwatheod: I'm not sure about this. First, the proof of `insertLeft_numeric` seems to get a little more complicated, since `simp` doesn't seem to work very well with `Option` values and I have to do an additional layer of `constructor` / `rcases`. But also, perhaps more significantly, this wouldn't generalize to the case where we want to add several options at once, such as a whole family of options `xl' -> PGame`, in which case we definitely need a `xl ⊕ xl'`. I think we would want to do that later in the development of the theory.

semorrison:
bors merge

Issue #13998, opened 2024-06-20
duplicated definition of the pullback functor between over categories


emilyriehl:
Today @sinhp and I discovered that there are two definitions of the pullback functor between over categories:

(i) one called `Over.pullback` that appears in [Limits.Over](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Limits/Over.html).

(ii) one called `Over.baseChange` that appears in [Adjunction.Over](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Adjunction/Over.html) and

`Over.pullback` is proven to be a right adjoint in `Over.mapPullbackAdj`, while `Over.baseChange` was proven to be a right adjoint in `Over.mapAdjunction` --- my first mathlib PR!

The comments to the `Limits.Over` file mention a TODO: If `C` has binary products, then `forget X : Over X ⥤ C` has a right adjoint. This appears as `Over.forgetAdjStar` in the `Over.Adjunction` file.

@sinhp has discussed this with @jcommelin and @joelriou who confirm that the correct place for this material is in the adjunction folder. But I have no clue which names, constructions, etc are best.

jcommelin:
Closed by #14519

PR #14519, opened 2024-07-08
[Merged by Bors] - deduplicate pullback and baseChange

sinhp:
Deduplication of `Over.pullback` and `Over.baseChange`.  `Over.baseChange` has been renamed to `Over.pullback` and `simp` lemmas added to `Over.pullback` and `Over.mapPullbackAdj`.  

---


emilyriehl:
This addresses issue #13998.

emilyriehl:
@sinhp maybe this no longer needs to be a draft?

sinhp:
> @sinhp maybe this no longer needs to be a draft?

indeed, just made it "ready for review".

semorrison:
@sinhp, @emilyriehl, just noting that this needs to have a `git merge origin/master` and the conflicts resolved before it can be reviewed.

emilyriehl:
Thanks for explaining how to fix this @semorrison. Should be ready to review now.

semorrison:
My only concern here is that you've somehow pulled in more imports in `Mathlib.CategoryTheory.Adjunction.Over` (+24 transitive imports) and a few more imports into downstream files `Mathlib.CategoryTheory.Localization.FiniteProducts`, `Mathlib.CategoryTheory.Limits.Shapes.Diagonal`, `Mathlib.CategoryTheory.MorphismProperty.Limits`, according to https://github.com/leanprover-community/mathlib4/pull/14519#issuecomment-2214559936.

It's not critical, but we do like to avoid import creep.

I had a look at what has changed using
```
lake exe graph --from Mathlib.CategoryTheory.Category.Basic --to Mathlib.CategoryTheory.Adjunction.Over --reduce p.pdf
```
before and after this PR. `Mathlib.CategoryTheory.Mates` pulls in `Mathlib.CategoryTheory.Conj`, which then for its second half (but I think not for mates?) needs `Mathlib.CategoryTheory.Endomorphism` which then pulls in a bunch of algebra files.

If we want to fix this (either before or after merging this PR), I think we should try to split `Conj`, moving its second half to a new file, and hopefully `Mates` doesn't need to import that.

semorrison:
Looking more, yes, `Conj` can be split into a first half called `HomCongr`, and then leave the second half as `Conj`.

If @sinhp or @emilyriehl would be game to do this (not necessarily immediately) I'm otherwise happy to merge here. If this doesn't sound fun I'm happy to merge in any case, and hopefully we can get to this split later.

emilyriehl:
@semorrison I've never tried to split a file, but I'd be happy to figure it out, because I'm sure I'll learn something in the process. (I'll ask on the zulip for help.) I'd prefer to merge this now to preempt future merge conflicts. I'll send you a new PR next week (or possibly the week after) to address the imports issue.

semorrison:
bors merge

PR #14515, opened 2024-07-08
[Merged by Bors] - feat: the (covariant) long exact sequence of Ext

joelriou:
We obtain the long exact sequence of `Ext`-groups `Ext X S.X₁ n₀ → Ext X S.X₂ n₀ → Ext X S.X₃ n₀ → Ext X S.X₁ n₁ → Ext X S.X₂ n₁ → Ext X S.X₃ n₁` when `n₀ + 1 = n₁` and `S` is a short exact short complex in an abelian category.

---

- [x] depends on: #14496
- [x] depends on: #14520
- [x] depends on: #14793 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14520, opened 2024-07-08
[Merged by Bors] - feat: more API for Function.Exact

joelriou:
If two diagrams involving additive maps are isomorphic, one is exact (in the sense of `Function.Exact`) iff the other is. In this PR, we also relate `Function.Exact` and `ShortComplex.Exact` in the categories of abelian groups and modules over a ring.

---

This shall be used in the (draft) PR #14515 which constructs the covariant long exact sequence of `Ext`-groups: even though this is phrased in terms of `ShortComplex.Exact`, proofs use `Function.Exact` because I need to change universes.


Mathlib/Algebra/Homology/ShortComplex/Ab.lean
```diff
@@ -120,6 +121,8 @@ lemma ab_exact_iff_range_eq_ker : S.Exact ↔ S.f.range = S.g.ker := by
   · intro h
     rw [h]
 
+variable {S}
+
 lemma ShortExact.ab_injective_f (hS : S.ShortExact) :
```
jjaassoonn, line 126: Should this be called `ShortExact.addCommGrp_injective_f`?

jjaassoonn: Oops

AntoineChambert-Loir:
I have no serious comment to make except questions about the way mathlib is supposed to handle namespaces. The multiple invocations of `_root_.` (lines 78ff of Exact.lean) make me feel one should organize the code slightly differently (for readability, and also to minimize risks of giving functions an incorrect/inconsistent namespace), but I have no idea of what to do. (It has to do with the organization of a file in  `section`s, and the behaviour of `namespace`, habits seem to differ.)

joelriou:
> I have no serious comment to make except questions about the way mathlib is supposed to handle namespaces. The multiple invocations of `_root_.` (lines 78ff of Exact.lean) make me feel one should organize the code slightly differently

I had just replicated the same organization which was already in this file in the `LinearMap` part. I have slightly reorganized the file so that there are no longer `_root_`.

Mathlib/Algebra/Exact.lean
```diff
@@ -40,57 +40,152 @@ def Exact [Zero P] : Prop := ∀ y, g y = 0 ↔ y ∈ Set.range f
...
+lemma addMonoidHom_comp_eq_zero (h : Exact f g) : g.comp f = 0 :=
+  DFunLike.coe_injective h.comp_eq_zero
+
+section
+
+variable {X₁ X₂ X₃ Y₁ Y₂ Y₃ : Type*} [AddCommMonoid X₁] [AddCommMonoid X₂] [AddCommMonoid X₃]
+  [AddCommMonoid Y₁] [AddCommMonoid Y₂] [AddCommMonoid Y₃]
+  (e₁ : X₁ ≃+ Y₁) (e₂ : X₂ ≃+ Y₂) (e₃ : X₃ ≃+ Y₃)
+  {f₁₂ : X₁ →+ X₂} {f₂₃ : X₂ →+ X₃} {g₁₂ : Y₁ →+ Y₂} {g₂₃ : Y₂ →+ Y₃}
+  (comm₁₂ : g₁₂.comp e₁.toAddMonoidHom = e₂.toAddMonoidHom.comp f₁₂)
```
riccardobrasca, line 112:
```suggestion
  (comm₁₂ : g₁₂.comp e₁ = e₂.toAddMonoidHom.comp f₁₂)
```
Not sure if it is nicer, but it works. As you prefer.

Mathlib/Algebra/Exact.lean
```diff
@@ -40,57 +40,152 @@ def Exact [Zero P] : Prop := ∀ y, g y = 0 ↔ y ∈ Set.range f
...
+  (comm₂₃ : g₂₃.comp e₂.toAddMonoidHom = e₃.toAddMonoidHom.comp f₂₃)
+
+lemma of_ladder_addEquiv_of_exact (H : Exact f₁₂ f₂₃) : Exact g₁₂ g₂₃ := by
```
riccardobrasca, line 115: Should the name mention some `comm` stuff?

Mathlib/Algebra/Exact.lean
```diff
@@ -40,57 +40,152 @@ def Exact [Zero P] : Prop := ∀ y, g y = 0 ↔ y ∈ Set.range f
...
   Exact.of_comp_of_mem_range (congrArg DFunLike.coe h1) h2
 
-lemma _root_.LinearMap.exact_of_comp_of_mem_range
+lemma exact_of_comp_of_mem_range
     (h1 : g ∘ₗ f = 0) (h2 : ∀ x, g x = 0 → x ∈ range f) : Exact f g :=
   exact_of_comp_eq_zero_of_ker_le_range h1 h2
 
+end LinearMap
+
+lemma LinearEquiv.conj_exact_iff_exact (e : N ≃ₗ[R] N') :
```
riccardobrasca, line 177:
```suggestion
variable (f g) in
lemma LinearEquiv.conj_exact_iff_exact (e : N ≃ₗ[R] N') :
```

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Algebra/Exact.lean
joelriou: I would think that this was the intent of `ladder` in the original name of the `LinearMap` version, which I have basically duplicated here for `AddMonoidHom`.

joelriou:
Thanks!

bors merge

PR #14521, opened 2024-07-08
[Merged by Bors] - chore(Data/Stream/Init): use `ℕ` instead of `Nat` consistently

grunweg:
The file currently uses a mix of these: use the nice unicode character, since we can.

---


grunweg:
That is a reasonable argument, in principle. Note that the file is currently inconsistent about this.

Komyyy (changes requested):
This PR looks reasonable!
A further request from me is that `Stream.Defs` should also be edited.

grunweg:
> This PR looks reasonable! A further request from me is that `Stream.Defs` should also be edited.

Done!

Komyyy (approved):

Komyyy:
@grunweg Thank you!

j-loreaux:
I asked Kim and they were fine with this.

bors merge

PR #14523, opened 2024-07-08
chore: use new `HashMap` in `allNamesByModule`

kim-em:
Note this is a PR to `nightly-testing`, since this isn't available on `master` yet, but I think it can be reviewed normally.


semorrison:
@adomani the declarations diff appears to be incorrect, here. Do we need to disable this for PRs to something other than `master`?

adomani:
The script already takes an optional commit that defaults to master.  I would have to look into how to make the action detect the target branch and the script can run with that instead of master, I suppose.

I won't have time to look at it today, as I'm about to have dinner!

PR #14524, opened 2024-07-08
[Merged by Bors] - feat(Limits/Shapes/Pullback): Add API and golf proofs

callesonne:
This is PR adds some API to pullbacks, and golfs some proofs in those files. 

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14526, opened 2024-07-08
[Merged by Bors] - feat(Limits/Shapes/Pullback): Give pullback.fst and friends explicit parameters

callesonne:
This PR add changes `pullback.fst`, `pullback.fst`, `pushout.inr` and `pushout.inl` to have explicit parameters.

---

- [x] depends on: #14524


joelriou:
In general, compared to other mathlib users, I am not a huge fan of implicit parameters, but even I got used to using implicit parameters in this case. Then, I am not sure we want to proceed in this direction.

joelriou:
Ok. Thanks!

bors merge

joelriou:
Please merge it when it passes CI:

bors d+

PR #14527, opened 2024-07-08
[Merged by Bors] - chore: remove `@[simp]` from `List.replicate_succ`

kim-em:
I may be tilting at windmills, but I've removed `@[simp]` from `List.replicate_succ` in Lean, and I think it results in slightly better simp normal forms in some situations. We don't yet have automatic post-Mathlib testing of simp normal forms, but I hope we will eventually.

In the meantime, this removes an inconsistency by cancelling Mathlib's downstream addition of `@[simp]` to `List.replicate.succ`. Mathlib doesn't mind much.


jcommelin:
Thanks :tada:

bors merge

PR #14529, opened 2024-07-08
[Merged by Bors] - chore(Data/Set): golf and add decidability of insert

b-mehta:
Adds the insert version, golfs existing versions using `inferInstanceAs`.
Makes the lemmas all consistent about using `Decidable` vs `DecidablePred`.

These decidability instances are useful in downstream applications. 

---


Ruben-VandeVelde (approved):
maintainer merge

urkud:
bors merge

PR #14390, opened 2024-07-03
[Merged by Bors] - chore (Algebra.Order.Monoid.Canonical.Defs): split off `ExistsAddOfLE` and `ExistsMulOfLE`

mattrobball:
The type classes `ExistsAddOfLE` and `ExistsMulOfLE` are unbundled versions of canonical ordering. To use them properly and avoid bundled ordered algebraic classes, we should put them in a separate file.

---

- [x] depends on: #14380

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14531, opened 2024-07-08
[Merged by Bors] - chore (Algebra.Order.WithTop): split into unbundled and bundled 

mattrobball:
Much of the results in `Algebra.Order.Monoid.WithTop` only depend on unbundled ordered algebra instance parameters. We split those results out into a new file `Algera.Order.Monoid.Unbundled.WithTop` leaving the facts about bundled ordered monoids.

---

- [x] depends on : #14390 


riccardobrasca:
LGTM. Do you know why there are declarations that changed?

bprs d+

riccardobrasca:
Ops

bors d+

mattrobball:
> LGTM. Do you know why there are declarations that changed?

I couldn't figure why these got reported. I checked they were just moved. Maybe a bug with two declarations in the same file having the same name after stripping namespaces.

mattrobball:
bors merge

PR #14533, opened 2024-07-08
[Merged by Bors] - style(DedekindDomain/AdicValuation): fix some snake_case to lowerCamelCase

faenuccio:
In some theorem names, `int_valuation`  was still used instead of `intValuation`, and this PR fixes this.

### DIFF
--- a/Mathlib/RingTheory/DedekindDomain/AdicValuation.lean
+++ b/Mathlib/RingTheory/DedekindDomain/AdicValuation.lean
@@ -29,14 +29,14 @@ We define the completion of `K` with respect to the `v`-adic valuation, denoted
     `v.adicCompletion`.
 
 ## Main results
-- `IsDedekindDomain.HeightOneSpectrum.int_valuation_le_one` : The `v`-adic valuation on `R` is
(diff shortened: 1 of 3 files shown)


grunweg:
Thanks for pursuing this clean-up? Would you also like to rename `int_valuation_exists_uniformizer` and `int_valuation_le_pow_iff_dvd`?
Can you also add deprecation aliasses for the renamed theorems? (There is a script somewhere to generate these automatically, though I'm not sure what the current status is. @adomani might know, or you can ask on zulip.)

faenuccio:
OK done, I have discussed with @adomani  (thanks!!!) and have added the deprecation aliases using the `deprecate to` tool. I had to manually remove the namespace from the aliases lest they would not work (but now they do).

adomani:
Ah, thanks for the report on the namespace!  I will try to take a look, but I'm not sure when I'll have the time!

mariainesdff:
Thanks!

mariainesdff:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14534, opened 2024-07-08
[Merged by Bors] - chore(RingTheory/AdicCompletion): manual instances

chrisflav:
Instead of inferring `AddCommGroup`, `Module`, etc. instances for `AdicCompletion I M` from being a submodule, we explicitly spell out the definitions to improve performance.

Also marks some definitions as `noncomputable` to save time spent on compilation.

See https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Adic.20completion.20is.20slow.

---


chrisflav:
!bench

chrisflav:
!bench

Mathlib/RingTheory/AdicCompletion/Basic.lean
```diff
@@ -348,28 +374,38 @@ def AdicCauchySequence : Type _ := { f : ℕ → M // IsAdicCauchy I M f }
 
 namespace AdicCauchySequence
 
-/-- The type of `I`-adic cauchy sequences is a submodule of the product `ℕ → M`. -/
-def submodule : Submodule R (ℕ → M) where
```
mattrobball, line 352: These are still valuable I think

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -50,20 +50,33 @@ def transitionMapₐ {m n : ℕ} (hmn : m ≤ n) :
...
-/-- `AdicCompletion I R` is a subring of `∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)`. -/
-def subring : Subring (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
-  Subalgebra.toSubring (subalgebra I)
-
-instance : CommRing (AdicCompletion I R) :=
-  inferInstanceAs <| CommRing (subring I)
-
-instance : Algebra R (AdicCompletion I R) :=
-  inferInstanceAs <| Algebra R (subalgebra I)
+instance : Mul (AdicCompletion I R) where
```
mattrobball, line 53: Can you make the data instances `irreducible`?

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -50,20 +50,33 @@ def transitionMapₐ {m n : ℕ} (hmn : m ≤ n) :
...
+  mul x y := ⟨x.val * y.val, by simp [x.property, y.property]⟩
+
+instance : One (AdicCompletion I R) where
+  one := ⟨1, by simp⟩
+
+instance : CommRing (AdicCompletion I R) where
```
mattrobball, line 59: `Function.Injective.commRing` might help

chrisflav: I changed it, not entirely sure its better now though.

Mathlib/RingTheory/AdicCompletion/Algebra.lean
chrisflav: Done.

chrisflav:
!bench

chrisflav:
!bench

Mathlib/RingTheory/AdicCompletion/Algebra.lean
chrisflav: I did another benchmark comparing the version with `Function.Injective.commRing` and the manual one: http://speed.lean-fro.org/mathlib4/compare/04e26275-26be-4d08-8dd1-ffeb4e23bd0e/to/e7b27246-a3e6-496a-b552-ff4b45c7236e?hash2=55cf90705c3509ab5da127b7ecd4db0e147af667
It seems the manual one is significantly faster, so I suggest we leave it at this.

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -59,11 +64,35 @@ def subalgebra : Subalgebra R (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
 def subring : Subring (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
   Subalgebra.toSubring (subalgebra I)
 
-instance : CommRing (AdicCompletion I R) :=
-  inferInstanceAs <| CommRing (subring I)
-
-instance : Algebra R (AdicCompletion I R) :=
-  inferInstanceAs <| Algebra R (subalgebra I)
+@[irreducible]
```
mattrobball, line 67: I am not sure if the `irreducible` attribute does anything on an instance. The pattern is 
```suggestion
@[irreducible]
def mul (x y : AdicCompletion I R) : AdicCompletion I R := ⟨x.val * y.val, by simp [x.property, y.property]⟩

instance instMul : Mul (AdicCompletion I R) where 
  mul := mul 
```
so Lean can never unfold `mul`

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -59,11 +64,35 @@ def subalgebra : Subalgebra R (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
...
+  mul_assoc x y z := Subtype.ext <| mul_assoc x.val y.val z.val
+  one_mul x := Subtype.ext <| one_mul x.val
+  mul_one x := Subtype.ext <| mul_one x.val
+  zsmul := zsmulRec
+  add_left_neg x := Subtype.ext <| add_left_neg x.val
+  sub_eq_add_neg x y := Subtype.ext <| sub_eq_add_neg x.val y.val
+  mul_comm x y := Subtype.ext <| mul_comm x.val y.val
+
+instance : Algebra R (AdicCompletion I R) where
+  toFun r := ⟨algebraMap R (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) r, by simp⟩
```
mattrobball, line 89: Lean might try to unfold `toFun` unnecessarily during unification checks. You could try hoisting it out as a separate declaration.

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -59,11 +64,35 @@ def subalgebra : Subalgebra R (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
...
+
+instance : CommRing (AdicCompletion I R) where
+  left_distrib x y z := Subtype.ext <| left_distrib x.val y.val z.val
+  right_distrib x y z := Subtype.ext <| right_distrib x.val y.val z.val
+  zero_mul x := Subtype.ext <| zero_mul x.val
+  mul_zero x := Subtype.ext <| mul_zero x.val
+  mul_assoc x y z := Subtype.ext <| mul_assoc x.val y.val z.val
+  one_mul x := Subtype.ext <| one_mul x.val
+  mul_one x := Subtype.ext <| mul_one x.val
+  zsmul := zsmulRec
```
mattrobball, line 83: I also wonder if there is a better syntactical spelling for `nsmul` and `zsmul` fields. Being data, they could also be made irreducible to avoid unfolding.

Mathlib/RingTheory/AdicCompletion/Algebra.lean
chrisflav: Should I do this for every data field? Also `zero`, `one`, etc.?

mattrobball: I think it is a worth a test.

chrisflav:
!bench

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -215,6 +336,8 @@ instance : IsScalarTower R (AdicCompletion I R) (AdicCompletion I M) where
     ext n
     rw [smul_eval, val_smul, val_smul, smul_eval, smul_assoc]
 
+unseal smul' in
+unseal mul in
 /-- A priori `AdicCompletion I R` has two `AdicCompletion I R`-module instances.
 Both agree definitionally. -/
 example : module I = @Algebra.toModule (AdicCompletion I R)
```
chrisflav, line 343: I think the two `unseal`s here are a problem: In the other PR in the `AsTensorProduct` file (https://github.com/leanprover-community/mathlib4/blob/abf5c5635c9bde7cefa76b2c18405067b604e089/Mathlib/RingTheory/AdicCompletion/AsTensorProduct.lean#L133-L138), I had to add the same `unseal`s because Lean can no longer identify these two instances as equal. Is there anything I can do about this? Or does it mean we can't make `mul` and `smul'` irreducible?

chrisflav:
I now have benchmarks of making everything irreducible: The difference is insignificant, only significant change is in the `Algebra` file which gets slightly slower by making everything irreducible:
http://speed.lean-fro.org/mathlib4/compare/e7b27246-a3e6-496a-b552-ff4b45c7236e/to/a50cfe65-06b9-429a-b2ed-ef196015bb23?hash1=55cf90705c3509ab5da127b7ecd4db0e147af667

But of course, in the slowest lemma of the `AsTensorProduct` file I currently have two `unseal`s, see above. @mattrobball

mattrobball:
Let me actually open up the file. I will need some time though

Mathlib/RingTheory/AdicCompletion/Algebra.lean
```diff
@@ -59,11 +64,61 @@ def subalgebra : Subalgebra R (∀ n, R ⧸ (I ^ n • ⊤ : Ideal R)) :=
...
+@[irreducible]
+def mul (x y : AdicCompletion I R) : AdicCompletion I R :=
+  ⟨x.val * y.val, by simp [x.property, y.property]⟩
+
+instance : Mul (AdicCompletion I R) where
+  mul := mul I
+
+section
+
+unseal mul
```
mattrobball, line 85: For future reference, this partially defeats the purpose of `irreducible`. You can do a more targeted `unseal X in` in a proof step I believe. Or you can wrap your term in `with_unfolding_all` to achieve a similar effect

chrisflav: I thought that the `unseal` here is confined to this section? But you are saying that even this is too big of a scope?

mattrobball: You are correct, it is. But with the definition open, Lean will try to do extra work there. Less than overall, but still can be avoided. 

The pattern is usually: 
- make `@[irreducible] def X` 
- develop minimal necessary API for `X` either `unseal`ing or `with_all_unfolding`
- never `unseal` again 

So in general, I would never see an `unseal X` unless it ended in an `in`

mattrobball:
```lean
add_smul r s x := by
    dsimp 
    ext n
    simp only [coe_eval, smul_eval, map_add, add_smul, val_add]
```
seems to help around L250. 

Also, why were the `simp` attributes removed?

Mathlib/RingTheory/AdicCompletion/Algebra.lean
chrisflav: I see. But in the case above where I need `unseal` to see that the two `AdicCompletion I R`-module instances on `AdicCompletion I R` are defeq, I seem to be bound to use `unseal` whenever they come up. Is this correct or can I make this work somehow?

mattrobball: If you doing it by hand, then yes you are stuck with that. If it is happening in instance synthesis, it is much worse.

chrisflav:
> ```lean
> add_smul r s x := by
>     dsimp 
>     ext n
>     simp only [coe_eval, smul_eval, map_add, add_smul, val_add]
> ```
> 
> seems to help around L250.
> 
> Also, why were the `simp` attributes removed?

L250 of which file? And which `simp` attributes?

mattrobball:
Sorry, I've only opened `Algebra`

chrisflav:
> Sorry, I've only opened `Algebra` and I can't add comments on unchanged code :(

I am counting heartbeats there with `count_heartbeats in` and adding the `dsimp` increases heartbeats (very sligthly) from 25125 to 25250. Am I missing something?

mattrobball:
Sorry I was on the wrong declaration: `AdicCompletion.module` at L312

chrisflav:
> Sorry I was on the wrong declaration: `AdicCompletion.module` at L312

Now I am confused, at which version are you looking? In the current one, the `Algebra` file only has 274 lines (Note that I reverted the two commits making everything `irreducible`).

mattrobball:
It is out-dated but `AdicCompletion.module` didn't change its name right?

chrisflav:
> It is out-dated but `AdicCompletion.module` didn't change its name right?

Ah yes, no it did not change its name, but then adding the `dsimp` does not seem to make a significant difference to `count_heartbeats in`.

mattrobball:
Oh well, it was minor I think anyway

chrisflav:
So should I go for a final benchmark?

mattrobball:
There is a good chance it won't make it in without using the `Function.Injective` constructors. We should test that fully first

chrisflav:
I had tried out the version with `Function.Injective` constructors above, but it was significantly slower. See this comparison: http://speed.lean-fro.org/mathlib4/compare/04e26275-26be-4d08-8dd1-ffeb4e23bd0e/to/e7b27246-a3e6-496a-b552-ff4b45c7236e?hash2=55cf90705c3509ab5da127b7ecd4db0e147af667

The first run is with `Function.Injective` constructors and the second one without.

chrisflav:
Anything else I should test?

chrisflav:
!bench

mattrobball:
What's the effect on `AsTensorProduct`?

chrisflav:
Benchmark is upcoming. My local one looks very good.

chrisflav:
Benchmark is there. This is the comparsion:

http://speed.lean-fro.org/mathlib4/compare/e7b27246-a3e6-496a-b552-ff4b45c7236e/to/74e2d180-1ac7-47c1-9b6e-c7589ce1cd90?hash1=abf5c5635c9bde7cefa76b2c18405067b604e089

First commit: manual instances
Second commit: `Function.Injective` constructors with `fast_instance%`.

I think this looks promising, so I'll revert the commits adding the `fast_instance%` stuff and we add them back in later when the #11521 lands. What do you think @mattrobball ?

mattrobball:
Make a dependent PR with the `fast_instance` insertions also so we can quickly put them back when that PR lands

chrisflav:
> Also please make a separate dependent PR with the `fast_instance` insertions also so we can quickly put them back when that PR lands

Since #11521 currently has merge conflicts, what should I do to create the dependent PR? Maybe from #11521 the actual `fast_instance%` code should be split of, otherwise it will collect merge conflicts all the time.

mattrobball:
I forgot it wasn't just the single file. I'll split it later today.

chrisflav:
Anything else I should do for this PR to be merged?

mattrobball:
I don't think so. Thanks for your patience.

bors merge

chrisflav:
Thanks a lot for the guidance and for helping analyzing this!

PR #14535, opened 2024-07-08
[Merged by Bors] - feat(Analysis/Fourier/ZMod): more API for discrete Fourier transform

loefflerd:
Define the discrete Fourier transform as a `LinearEquiv`, rather than just a map; and prove the Fourier inversion formula, as well as a bunch of more routine lemmas. 

---

Remarks:
- I had to drop my plan of making the DFT be *by definition* an instance of the general Fourier-transform machinery, because I wanted to make it work for functions valued in any vector space; and mathlib's definition of "integration" only works when the target space is complete, even if the integral is over a finite discrete set. So now we define it by hand, & prove compatibility with the non-discrete FT if completeness holds.
- For some lemmas I wasn't sure whether it was best to formulate the statement with the functions "fully applied" or not. E.g. should `dft_comp_neg` be
```lemma dft_comp_neg (Φ : ZMod N → E) : 𝓕 (fun j ↦ Φ (-j)) = fun k ↦ 𝓕 Φ (-k)```
or
```lemma dft_comp_neg (Φ : ZMod N → E) (k : ZMod N) : 𝓕 (fun j ↦ Φ (-j)) k = 𝓕 Φ (-k)```
? 

- [x] depends on: #14495 

### CODE REVIEW

**Comment by MichaelStollBayreuth** [comment]
I think it would make sense to define DFT with more general targets (modules over a ring that contains a primitive $n$th root of unity). This would be relevant for applications in Computer Algebra (fast multiplication algorithms), for example. (Sorry for not mentioning this earlier on Zulip.)

**Comment by loefflerd** [comment]
Thanks for looking at this! I'm also at a conference this week, so there's no rush.

Regarding extensions to more general coefficients: do you have a suggestion for how the API should look? I'd rather avoid a situation where users need to specify an explicit choice of coefficient ring, & root of unity in it, every time they invoke any lemmas about the DFT.

**Comment by MichaelStollBayreuth** [comment]
I think we can define a quite general version of DFT for functions `Φ : R → M`, where `R` is a finite ring, `M` is a module over another ring `R'`, and we fix an additive character `ψ : R → R'`. The relevant condition on `ψ` to have interesting results should then be that it is primitive as an additive character (and that `n : R'` is invertible, or rather, `n : Module.End R' M` is invertible, for the linear equivalence). One can then still define `𝓕 Φ` to be `dft stdAddChar Φ`, where `Φ : ZMod N → E`. For convenience, some results may need to be restated in this context (to make use of properties of `stdAddChar` and/or the complex numbers that may not be true in general), but I think having the general version available is worth the extra effort.

BTW (and slightly off-topic), I think the correct definition of "`ζ : R` is a primitive `n`th root of unity" is "ζ ^ n = 1 and for all 0 < k < n, 1 - ζ ^ k is invertible" (one can restrict `k` to the maximal proper divisors of `n`, of course). With this definition, images of primitive `n`th roots of unity under ring homomorphisms are still primitive `n`th roots of unity, which is not in general the case with the current definition, which just says that `ζ` has order `n` in the multiplicative monoid. Note that this definition also implies that `n` is invertible in `R`.

File: Mathlib/Analysis/Fourier/ZMod.lean [hunk 1]
```diff
@@ -10,53 +10,184 @@ import Mathlib.NumberTheory.DirichletCharacter.GaussSum
...
+-/
 
-/-- The discrete Fourier transform on `ℤ / N ℤ` (with the counting measure) -/
-noncomputable def dft (Φ : ZMod N → ℂ) (k : ZMod N) : ℂ :=
-  Fourier.fourierIntegral toCircle Measure.count Φ k
+/--
+The discrete Fourier transform on `ℤ / N ℤ` (with the counting measure). This definition is
+private because it is superseded by the bundled `LinearEquiv` version.
+-/
+private noncomputable def aux_dft (Φ : ZMod N → E) (k : ZMod N) : E :=
```
YaelDillies, line 45:
```suggestion
private noncomputable def auxDFT (Φ : ZMod N → E) (k : ZMod N) : E :=
```

Mathlib/Analysis/Fourier/ZMod.lean
```diff
@@ -10,53 +10,184 @@ import Mathlib.NumberTheory.DirichletCharacter.GaussSum
...
+lemma dft_apply (Φ : ZMod N → E) (k : ZMod N) :
+    𝓕 Φ k = ∑ j : ZMod N, stdAddChar (-(j * k)) • Φ j :=
+  rfl
 
-end ZMod
+lemma dft_def (Φ : ZMod N → E) :
+    𝓕 Φ = fun k ↦ ∑ j : ZMod N, stdAddChar (-(j * k)) • Φ j :=
+  rfl
+
+lemma inv_dft_apply (Ψ : ZMod N → E) (k : ZMod N) :
```
YaelDillies, line 107: I would suggest
```suggestion
lemma invDFT_apply (Ψ : ZMod N → E) (k : ZMod N) :
```
to avoid confusing with `(dft Φ)⁻¹`

loefflerd:
Thanks Yaël for the comments, I have made the modifications as you suggested.

YaelDillies (approved):
Thanks!

maintainer merge

PR #14540, opened 2024-07-08
[Merged by Bors] - chore: remove scripts/mk_all.sh

grunweg:
It has since been rewritten in Lean, and is unused.

---


grunweg:
CC @semorrison adomani

adomani:
My main reason for wanting to keep this script was that Floris had problems using `lake exe mk_all` on windows.  However, I think that those issues were only marginally related to the lake script and they appear to have been solved.  I use the script regularly and CI uses is as well.  I would just like some confirmation from some MacOs user that it works as intended for them and then I'm happy to remove the shell script!

semorrison:
Worked for me on macos.

bors merge

Ruben-VandeVelde:
I'm sad to see this - I'm still using it regularly

adomani:
Rubén, why do you not use `lake exe mk_all`?

PR #14542, opened 2024-07-08
feat(GroupTheory/Perm/Support): Redefine `support` to be a `Set`, removing finiteness requirements.

wrenna-robson:
This refactors GroupTheory/Perm/Support.lean, and then makes minimally consequential fixes to others statements and proofs as a result.

---

This was a challenging PR to get to pass CI. There is absolutely no need that I can see that the definition of `Equiv.Perm.support` needs to have a finiteness condition, but because of the way it is currently uses, that assumption is everywhere, including in some deeply complex proofs.

As there is often a need to refer to the cardinality of the support, I've added in an appropriate definition for this - this definition is noncomputable but has a computable form under the same conditions under which it was previously used. I don't think this is an ideal way of doing this but I think a more full refactoring will take more time and thinking than I have.

In general I think we should avoid excessive decidable equality and finiteness conditions creeping in when they are not necessary, but this specific PR was motivated by a need to use the notion of support for a permutation whose support is finite but whose underlying domain of action is not.


Mathlib/GroupTheory/Perm/Cycle/Concrete.lean
```diff
@@ -197,23 +199,26 @@ variable [Fintype α] [DecidableEq α] (p : Equiv.Perm α) (x : α)
 until looping. That means when `f x = x`, `toList f x = []`.
 -/
 def toList : List α :=
-  (List.range (cycleOf p x).support.card).map fun k => (p ^ k) x
+  (List.range (cycleOf p x).support.toFinset.card).map fun k => (p ^ k) x
```
YaelDillies, line 202:
```suggestion
  (List.range (cycleOf p x).supportCard).map fun k => (p ^ k) x
```

Mathlib/GroupTheory/Perm/Cycle/Type.lean
```diff
@@ -45,24 +45,25 @@ variable [DecidableEq α]
...
 def cycleType (σ : Perm α) : Multiset ℕ :=
-  σ.cycleFactorsFinset.1.map (Finset.card ∘ support)
+  σ.cycleFactorsFinset.1.map (fun f => f.support.toFinset.card)
 
 theorem cycleType_def (σ : Perm α) :
-    σ.cycleType = σ.cycleFactorsFinset.1.map (Finset.card ∘ support) :=
+    σ.cycleType = σ.cycleFactorsFinset.1.map (fun f => f.support.toFinset.card) :=
   rfl
 
+
```
YaelDillies, line 54:
```suggestion
```

Mathlib/GroupTheory/Perm/Cycle/Type.lean
```diff
@@ -45,24 +45,25 @@ variable [DecidableEq α]
...
+    σ.cycleType = s.1.map (fun f => f.supportCard) := by
+  simp_rw [cycleType_def, supportCard_compute]
   congr
   rw [cycleFactorsFinset_eq_finset]
   exact ⟨h1, h2, h0⟩
 
 theorem cycleType_eq {σ : Perm α} (l : List (Perm α)) (h0 : l.prod = σ)
     (h1 : ∀ σ : Perm α, σ ∈ l → σ.IsCycle) (h2 : l.Pairwise Disjoint) :
-    σ.cycleType = l.map (Finset.card ∘ support) := by
+    σ.cycleType = l.map (fun f => f.supportCard) := by
```
YaelDillies, line 66:
```suggestion
    σ.cycleType = l.map supportCard := by
```

Mathlib/GroupTheory/Perm/Support.lean
```diff
@@ -17,8 +17,12 @@ In the following, `f g : Equiv.Perm α`.
 * `Equiv.Perm.Disjoint`: two permutations `f` and `g` are `Disjoint` if every element is fixed
   either by `f`, or by `g`.
   Equivalently, `f` and `g` are `Disjoint` iff their `support` are disjoint.
-* `Equiv.Perm.IsSwap`: `f = swap x y` for `x ≠ y`.
+* `Equiv.Perm.IsSwapOn f x y`: `f` maps `x` to `y`, `y` to `x`, and fixes all other elements.
+* `Equiv.Perm.IsSwap`: `f` transposes some distinct elements `x` and `y`, and
+  fixes all other elements.
 * `Equiv.Perm.support`: the elements `x : α` that are not fixed by `f`.
+* `Equiv.Perm.supportCard`: the cardinality of `f.support` when it is finite. Noncomputable but
```
YaelDillies, line 24: Can you please rename `supportCard` to `cardSupport`?

Mathlib/GroupTheory/Perm/Support.lean
```diff
@@ -166,111 +146,188 @@ theorem Disjoint.mul_apply_eq_iff {σ τ : Perm α} (hστ : Disjoint σ τ) {a
...
-
 end Disjoint
 
 section IsSwap
 
-variable [DecidableEq α]
+/-- `f.IsSwapOn x y` indicates that the permutation `f` is a transposition of the two elements `x`
+and `y`. Note that this means that `f` might be trivial, if `x = y`. -/
+def IsSwapOn (f : Perm α) (x y : α) : Prop :=
+  (∀ {z}, z ≠ x → z ≠ y → f z = z) ∧ f x = y ∧ f y = x
```
YaelDillies, line 156: Can you make this a separate PR?

Mathlib/GroupTheory/Perm/Support.lean
```diff
@@ -166,111 +146,188 @@ theorem Disjoint.mul_apply_eq_iff {σ τ : Perm α} (hστ : Disjoint σ τ) {a
...
-theorem not_mem_support {x : α} : x ∉ f.support ↔ f x = x := by simp
+section support
 
-theorem coe_support_eq_set_support (f : Perm α) : (f.support : Set α) = { x | f x ≠ x } := by
-  ext
-  simp
+variable {f g : Perm α}
+
+/-- The `Set` of nonfixed points of a permutation. -/
+def support (f : Perm α) : Set α := { x | f x ≠ x }
```
YaelDillies, line 299: As explained on Zulip, can you please keep the finset-valued support around (along with its API, which does **not** include the lemmas that mention it just to talk about its membership as there it can be replaced by the new `support`), maybe named `supportFinset`?

And to help with reviewing can you make a first PR that renames `support` to `supportFinset` everywhere? This will create a lot of conflicts with the current PR, but you can systematically solve them by taking the change from this PR (except for the ones that remove the `supportFinset` API)

Mathlib/GroupTheory/Perm/Support.lean
linesthatinterlace: Will do - there was a good reason I did it here I think but I can't remember what it was and I can surely disentangle it.

PR #14544, opened 2024-07-08
[Merged by Bors] - feat: star algebra homomorphisms between non-unital C⋆-algebras are contractive

j-loreaux:
This generalizes the existing results for unital algebras to non-unital algebras by passing to the unitization within the proof.

We have to add a `StarModule` hypothesis to recover the unital case, but this is minor.

---


Mathlib/Analysis/NormedSpace/Star/Spectrum.lean
```diff
@@ -122,48 +122,64 @@ theorem selfAdjoint.val_re_map_spectrum [StarModule ℂ A] (a : selfAdjoint A) :
...
+variable [FunLike F A B] [NonUnitalAlgHomClass F ℂ A B] [NonUnitalStarAlgHomClass F ℂ A B]
+
+open Unitization
+
+/-- A non-unital star algebra homomorphism of complex C⋆-algebras is norm contractive. -/
+lemma nnnorm_apply_le (φ : F) (a : A) : ‖φ a‖₊ ≤ ‖a‖₊ := by
+  have h (ψ : Unitization ℂ A →⋆ₐ[ℂ] Unitization ℂ B) (x : Unitization ℂ A) :
+      ‖ψ x‖₊ ≤ ‖x‖₊ := by
+    suffices ∀ s : Unitization ℂ A, IsSelfAdjoint s → ‖ψ s‖₊ ≤ ‖s‖₊ by
+      set_option maxSynthPendingDepth 2 in -- see: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/weird.20type.20class.20synthesis.20error/near/421224482
```
dupuisf, line 141: In that Zulip thread, the issue is apparently the fact that we can't play with reducibility and so we're stuck. But this was back in February, and now we have `seal` and `unseal` -- do you think we could revisit the issue?

j-loreaux: I've refactored so that these `synthPending`s are no longer necessary (although it's not my favorite solution, as things are a bit brittle). I don't see how we can use `seal` / `unseal` to help us here, but I've asked Anne on Zulip and pinged you in the message.

dupuisf:
Thanks!

bors r+

PR #14548, opened 2024-07-09
[Merged by Bors] - feat(CategoryTheory/GlueData): New constructor for glue data.

erdOne:
---
- [x] depends on: #14524


joelriou:
It looks good to me. Could you merge with master after #14524 is merged?

PR #14556, opened 2024-07-09
[Merged by Bors] - chore: remove Data/UnionFind

grunweg:
Its implementation has been improved in `lean4lean` and `batteries`; the mathlib implementation can be removed in favour of these.

Searching github for 'import Mathlib.Data.UnionFind' only yields forks of mathlib (either explicit or manual); hence this seems safe to remove without a deprecation period.

---

Search query used: https://github.com/search?q=lang%3ALean+%22import+Mathlib.Data.UnionFind%22&type=code&p=1


semorrison:
bors d+

grunweg:
Thanks for the quick review!
bors r+

PR #14557, opened 2024-07-09
[Merged by Bors] - fix: make the allowed unused tactics extensible

adomani:
This PR allows to change the tactics that the unused tactic linter ignores.

The main use-case is allowing `done` at the end of proofs, which is very useful for teaching purposes.  See for instance [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Disable.20.60linter.2EunusedTactic.60.20warning.20for.20.60done.60).

---

### DIFF
--- a/Mathlib/Tactic/Linter/UnusedTactic.lean
+++ b/Mathlib/Tactic/Linter/UnusedTactic.lean
@@ -94,10 +94,12 @@ initialize allowedRef : IO.Ref (HashSet SyntaxNodeKind) ←
     |>.insert `change?
     |>.insert `«tactic#adaptation_note_»
 
-/-- `#allow_unused_tactic` takes an input a space-separated list of identifiers and extends
-the tactics that are allowed by the unused tactic linter to not modify goals using the
-input identifiers.
-For instance, you can allow `done` and `skip` to not emit warning using
+/-- `#allow_unused_tactic` takes an input a space-separated list of identifiers.
+These identifiers are then allowed by the unused tactic linter:
+even if these tactics do not modify goals, there will be no warning emitted.
+Note: for this to work, these identifiers should be the `SyntaxNodeKind` of each tactic.
+
+For instance, you can allow `done` and `skip` tactics using
 ```lean
 #allow_unused_tactic Lean.Parser.Tactic.done Lean.Parser.Tactic.skip
 ```
--- a/Mathlib/Tactic/Linter/UnusedTactic.lean
+++ b/Mathlib/Tactic/Linter/UnusedTactic.lean
@@ -99,7 +99,7 @@ These identifiers are then allowed by the unused tactic linter:
 even if these tactics do not modify goals, there will be no warning emitted.
 Note: for this to work, these identifiers should be the `SyntaxNodeKind` of each tactic.
 
-For instance, you can allow `done` and `skip` tactics using
+For instance, you can allow the `done` and `skip` tactics using
 ```lean
 #allow_unused_tactic Lean.Parser.Tactic.done Lean.Parser.Tactic.skip
 ```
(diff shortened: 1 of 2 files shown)


test/UnusedTactic.lean
```diff
@@ -22,3 +22,23 @@ set_option linter.unusedTactic true in
 example : True := by
   congr
   constructor
+
+section allowing_more_unused_tactics
+--  test that allowing more unused tactics has the desired effect of silencing the linter
```
grunweg, line 27: Perhaps also add a test mentioning one of these tactics *before* the modification?

Mathlib/Tactic/Linter/UnusedTactic.lean
```diff
@@ -66,31 +66,46 @@ namespace UnusedTactic
...
+    |>.insert `Mathlib.Tactic.Propose.propose'
+    |>.insert `Lean.Parser.Tactic.traceState
+    |>.insert `Mathlib.Tactic.tacticMatch_target_
+    |>.insert `change?
+    |>.insert `«tactic#adaptation_note_»
+
+/-- `#allow_unused_tactic` takes an input a space-separated list of identifiers and extends
+the tactics that are allowed by the unused tactic linter to not modify goals using the
+input identifiers.
+For instance, you can allow `done` and `skip` to not emit warning using
```
grunweg, line 100: Two wording suggestions:
- The main sentence is understandable, but hard to parse - can this be simplified? Here's an attempt:
`#allow_unused_tactic` takes as input a space-separated list of identifiers. These identifier also are allowed by the unused tactic linter: even if these tactics do not modify goals, there will be no warning emitted. Note: for this to work, these identifiers should be the `SyntaxNodeKind` of each tactic.
- What do you think about this rephrasing for the example? "For instance, you can allow the `done` and `skip` tactics using..."

grunweg (commented):
This change seems well-motivated; thank you for doing this!
I have two wording suggestions on the docs. I'm seeing IO.Ref for the first time (and it's undocumented...), so somebody else should review this part carefully. The remainder looks good to me.

grunweg:
PS. I would call this a "feature" PR; this might be a matter of taste :-)

test/UnusedTactic.lean
adomani: Thanks, I added `done` to the previous `example`!

Mathlib/Tactic/Linter/UnusedTactic.lean
adomani: Thanks, I incorporated your suggestions, with some very minor adaptations!

adomani:
@eric-wieser, I know that you have some reservations on `IO.Ref`: this is just a heads-up that I am using one here!  :smile:

Mathlib/Tactic/Linter/UnusedTactic.lean
```diff
@@ -94,10 +94,12 @@ initialize allowedRef : IO.Ref (HashSet SyntaxNodeKind) ←
...
-/-- `#allow_unused_tactic` takes an input a space-separated list of identifiers and extends
-the tactics that are allowed by the unused tactic linter to not modify goals using the
-input identifiers.
-For instance, you can allow `done` and `skip` to not emit warning using
+/-- `#allow_unused_tactic` takes an input a space-separated list of identifiers.
+These identifiers are then allowed by the unused tactic linter:
+even if these tactics do not modify goals, there will be no warning emitted.
+Note: for this to work, these identifiers should be the `SyntaxNodeKind` of each tactic.
+
+For instance, you can allow `done` and `skip` tactics using
```
grunweg, line 102:
```suggestion
For instance, you can allow the `done` and `skip` tactics using

grunweg:
Thanks, looks good to me. Let's have a more qualified pair of eyes, perhaps Eric?
maintainer merge

jcommelin:
Thanks :tada: 

If CI passes, please remove the label `awaiting-CI` and merge this yourself, by adding a comment `bors r+`.

bors d+

PR #14559, opened 2024-07-09
[Merged by Bors] - chore(PULL_REQUEST_TEMPLATE.md): add template for moves and deletions

jcommelin:
---

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,6 +24,18 @@ If this PR depends on other PRs, please list them below this comment,
 using the following format:
 - [ ] depends on: #abc [optional extra text]
 - [ ] depends on: #xyz [optional extra text]
+
+If you are moving or deleting declarations,
+please use the following template above the `---` line.
+
+Moves:
+- Vector.* -> Mathlib.Vector.*
+- ...
+
+Deletions:
+- Nat.bit1_add_bit1
+- ...
+
 -->
 
 [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/from-referrer/)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,17 +16,8 @@ To indicate co-authors, include lines at the bottom of the commit message
 
 Co-authored-by: Author Name <author@email.com>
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,6 +24,18 @@ If this PR depends on other PRs, please list them below this comment,
...
+If you are moving or deleting declarations,
+please use the following template above the `---` line.
+
+Moves:
+- Vector.* -> Mathlib.Vector.*
+- ...
+
+Deletions:
+- Nat.bit1_add_bit1
+- ...
```
eric-wieser, line 37: Maybe put this above the `depends on` section, since this goes above the `---` but depends on goes below.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,6 +16,17 @@ To indicate co-authors, include lines at the bottom of the commit message
 
 Co-authored-by: Author Name <author@email.com>
 
+If you are moving or deleting declarations, please include lines at the bottom of the commit message
```
adomani, line 19:
```suggestion
If you are moving or deleting declarations, please include these lines at the bottom of the commit message
```

semorrison:
Is the intention that a bot will fill this in?

jcommelin:
Maybe not directly filling in, but certainly the intention is that a bot will generate a first draft that can be copy-pasta-edited.

adomani:
In particular, the pairs of declarations that `decl_diff` matches, are good candidates for `Moves` and the ones that have an unmatched `-` are good candidates for `Deletions`.

adomani:
I looked briefly into this and I am only seeing hacky ways of getting a PR template that depends on the PR itself.

As far as I understand, GitHub simply reads the file that this PR is editing and uses that as a template.  So, this file would have had to be edited between pushing and creating a PR.  For instance, the very first action upon pushing could be to compute the decl_diff, modify the PR_template file and then have it ready by the time you actually open the PR!

I wonder if not git-indexing this file and simply regenerating it on every push would give the "personalised touch" that we want.

semorrison:
I think if we develop a bot that can contribute answers here, it should just run after the PR is open. We can add a note to the template saying "You don't need to fill this in immediately: a bot will edit the PR top comment shortly with a first approximation, and you should check that it looks right."

semorrison:
bors merge

PR #14564, opened 2024-07-09
[Merged by Bors] - feat(Order.LiminfLimsup): add limsup_max

D-Thomine:
Add characterization of limsup and limsup of max.

---

This PR adds to Order.LiminfLimsup:

- Characterizations of the limsup and liminf in Conditionally complete linear orders: `limsup_le_iff`, `le_limsup_iff`, `le_liminf_iff`, `liminf_le_iff`.
- Identities such as limsup max = max limsup: `limsup_max`, `liminf_min`. Some lemmas about bounded/cobounded functions to handle limsup max.
- Same identities with supremum/infimum indexed over a Finset: `limsup_finset_sup`, `liminf_finset_inf`, `limsup_finset_sup'`, `liminf_finset_inf'`.
- ~Some identities defined on Conditionally complete linear orders now have versions on Complete linear orders. The autocomplete tactic doesn't work very well, so in practice they can be very helpful. Example: `limsup_le_limsup'`.~

This file is in dire need of a refactor. It is getting too long ; the linter fails because the file has reached almost 1800 lines (I am not sure if it makes sense to disable the flag as long as we don't go over the limit of 1900 lines). There are many less immediate issues which would warrant some extensive work.

---


pitmonticone:
Hi @D-Thomine, updated `style-exceptions.txt` as discussed [here](https://github.com/mseri/BET/issues/55#issuecomment-2217706009) and [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Porting.20TopologicalEntropy.20into.20Mathlib/near/450168092).

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -617,6 +617,10 @@ theorem limsup_le_limsup {α : Type*} [ConditionallyCompleteLattice β] {f : Fil
   limsSup_le_limsSup hu hv fun _ => h.trans
 #align filter.limsup_le_limsup Filter.limsup_le_limsup
 
+theorem limsup_le_limsup' {α} [CompleteLattice β] {f : Filter α} {u v : α → β} (h : u ≤ᶠ[f] v) :
```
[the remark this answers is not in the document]
D-Thomine: My feeling is that autoparam works well with e.g. the `exact` tactic, but not the `apply` tactic. But, yes, I agree that the right solution is to fix autoparam instead of adding duplicate lemmas (and then delete the duplicate lemmas).

I have a dozen of examples of autoparam not firing. I will try to write some minimal examples for Zulip.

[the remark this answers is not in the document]
D-Thomine: For my issues with the automatic fill-in: I've opened a [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/239415-metaprogramming-.2F-tactics/topic/isBoundedDefault.20not.20triggering/near/450437091).

In the meantime, these specialized versions are removed. Better to work now with a wonky tool which will, hopefully, be fixed, than to have to delete lemmas (with all the downstream effects) later.

D-Thomine:
Opening this PR for review. Apart from some golfing:

- @pitmonticone has added a style exception so that the linter stops complaining about the length of the file (thanks!).
- I've removed the duplicate lemmas which gave simpler versions of the theorem on Complete lattices / Complete linear orders.
- I've modified some lemmas about IsBoundedUnder/IsCoboundedUnder for finite sup/inf. One lemmas about the sup was not useful, while I've added two lemmas about the inf which are.

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1533,3 +1579,161 @@ theorem OrderIso.liminf_apply {γ} [ConditionallyCompleteLattice β] [Conditiona
...
+    (h₄ : f.IsBoundedUnder (· ≤ ·) v := by isBoundedDefault) :
+    limsup (fun a ↦ max (u a) (v a)) f = max (limsup u f) (limsup v f) := by
+  have bddmax := IsBoundedUnder.sup h₃ h₄
+  have cobddmax := isCoboundedUnder_le_max (v := v) (Or.inl h₁)
+  apply le_antisymm
+  · refine (limsup_le_iff cobddmax bddmax).2 (fun b hb ↦ ?_)
+    have hu := eventually_lt_of_limsup_lt (lt_of_le_of_lt (le_max_left _ _) hb) h₃
+    have hv := eventually_lt_of_limsup_lt (lt_of_le_of_lt (le_max_right _ _) hb) h₄
+    refine mem_of_superset (inter_mem hu hv) (fun _ ↦ ?_)
+    simp
```
sgouezel, line 1614:
```suggestion
    exact mem_of_superset (inter_mem hu hv) (fun _ ↦ by simp)
```

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1533,3 +1579,161 @@ theorem OrderIso.liminf_apply {γ} [ConditionallyCompleteLattice β] [Conditiona
...
+    f.IsBoundedUnder (· ≤ ·) (fun a ↦ sup' s hs F a) := by
+  choose! m hm using h
+  use sup' s hs m
+  simp only [eventually_map] at hm ⊢
+  rw [← eventually_all_finset s] at hm
+  refine hm.mono fun a h ↦ ?_
+  simp only [Finset.sup'_apply, sup'_le_iff]
+  exact fun i i_s ↦ le_trans (h i i_s) (le_sup' m i_s)
+
+theorem isCoboundedUnder_le_finset_sup' [LinearOrder β] {f : Filter α} {F : ι → α → β}
```
sgouezel, line 1649: For symmetry, what about a `sup` version for this one also?

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1533,3 +1579,161 @@ theorem OrderIso.liminf_apply {γ} [ConditionallyCompleteLattice β] [Conditiona
...
+    {s : Finset ι} (hs : s.Nonempty) (h : ∃ i ∈ s, f.IsCoboundedUnder (· ≥ ·) (F i)) :
+    f.IsCoboundedUnder (· ≥ ·) (fun a ↦ inf' s hs F a) :=
+  isCoboundedUnder_le_finset_sup' (β := βᵒᵈ) hs h
+
+theorem limsup_finset_sup [ConditionallyCompleteLinearOrderBot β] {f : Filter α}
+    {F : ι → α → β} {s : Finset ι}
+    (h₁ : ∀ i ∈ s, f.IsCoboundedUnder (· ≤ ·) (F i))
+    (h₂ : ∀ i ∈ s, f.IsBoundedUnder (· ≤ ·) (F i)) :
+    limsup (fun a ↦ sup s F a) f = sup s (fun i ↦ limsup (F i) f) := by
+  rcases eq_or_neBot f with (rfl | _)
```
sgouezel, line 1680: instead of doing twice the same proof, is it possible to deduce this one from the `sup'` version, separating the cases where `s` is empty or not?

sgouezel (commented):
Looks good, thanks!

Mathlib/Order/LiminfLimsup.lean
D-Thomine: It is already in the file, under the name `isCobounded_le_of_bot` :)

Mathlib/Order/LiminfLimsup.lean
D-Thomine: Nice idea, it works very well (and made me notice an imprecision in some statements : I now write `limsup (fun a ↦ sup s (fun i ↦ F i a)) f` instead of `limsup (fun a ↦ sup s F a) f`).

Mathlib/Order/LiminfLimsup.lean
D-Thomine: Done.

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1533,3 +1579,144 @@ theorem OrderIso.liminf_apply {γ} [ConditionallyCompleteLattice β] [Conditiona
...
+
+theorem isBoundedUnder_ge_finset_inf [LinearOrder β] [OrderTop β] {f : Filter α} {F : ι → α → β}
+    {s : Finset ι} (h : ∀ i ∈ s, f.IsBoundedUnder (· ≥ ·) (F i)) :
+    f.IsBoundedUnder (· ≥ ·) (fun a ↦ inf s (fun i ↦ F i a)) :=
+  isBoundedUnder_le_finset_sup (β := βᵒᵈ) h
+
+theorem limsup_finset_sup' [ConditionallyCompleteLinearOrder β] {f : Filter α}
+    {F : ι → α → β} {s : Finset ι} (hs : s.Nonempty)
+    (h₁ : ∀ i ∈ s, f.IsCoboundedUnder (· ≤ ·) (F i))
+    (h₂ : ∀ i ∈ s, f.IsBoundedUnder (· ≤ ·) (F i)) :
```
sgouezel, line 1677:
```suggestion
    (h₁ : ∀ i ∈ s, f.IsCoboundedUnder (· ≤ ·) (F i) := by exact fun i hi ↦ by isBoundedDefault)
    (h₂ : ∀ i ∈ s, f.IsBoundedUnder (· ≤ ·) (F i) := by exact fun i hi ↦ by isBoundedDefault) :
```
Same thing below.

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1533,3 +1579,144 @@ theorem OrderIso.liminf_apply {γ} [ConditionallyCompleteLattice β] [Conditiona
...
+      intro i i_s
+      apply eventually_lt_of_limsup_lt _ (h₂ i i_s)
+      exact lt_of_le_of_lt (Finset.le_sup' (f := fun i ↦ limsup (F i) f) i_s) hb
+    · simp only [mem_iInter, mem_setOf_eq, Finset.sup'_apply, sup'_lt_iff, imp_self, implies_true]
+  · apply Finset.sup'_le hs (fun i ↦ limsup (F i) f)
+    refine fun i i_s ↦ limsup_le_limsup (eventually_of_forall (fun a ↦ ?_)) (h₁ i i_s) bddsup
+    simp only [Finset.sup'_apply, le_sup'_iff]
+    use i, i_s
+
+theorem limsup_finset_sup [ConditionallyCompleteLinearOrderBot β] {f : Filter α}
```
sgouezel, line 1701: Is it possible instead to state this assuming `[ConditionallyCompleteLinearOrder β] [OrderBot β]` instead? In this way, it would be also dualizable to a liminf version, which is not possible currently.

D-Thomine: This requires a tweak to another file. The lemma `csSup_univ` required a `ConditionallyCompleteLinearOrderBot`, but its assumptions can be weakened to `ConditionallyCompleteLinearOrder` + `OrderBot`. With this modification, which should not break anything else, this works.

Mathlib/Order/LiminfLimsup.lean
D-Thomine: Nice idea. Tested, and in works!

sgouezel:
bors r+
Thanks!

PR #14571, opened 2024-07-09
feat(RingTheory.MvPowerSeries.Order) Order of multivariate power series

AntoineChambert-Loir:
Define order of power series.

TODO : refactor the case of `PowerSeries`

Co-authored-by @mariainesdff 

---


AntoineChambert-Loir:
Closed because I wanted to change its name and restrict its scope, and opened another one by accident.

PR #14579, opened 2024-07-09
[Merged by Bors] - feat(Algebra/Polynomial/Bivariate): add lemmas for bivariate polynomials

Multramate:
---

### DIFF
--- a/Mathlib/Algebra/Polynomial/Bivariate.lean
+++ b/Mathlib/Algebra/Polynomial/Bivariate.lean
@@ -41,14 +41,87 @@ abbrev CC (r : R) : R[X][Y] := C (C r)
 lemma evalEval_C (x y : R) (p : R[X]) : (C p).evalEval x y = p.eval x := by
   rw [evalEval, eval_C]
 
+lemma evalEval_CC (x y : R) (p : R) : (CC p).evalEval x y = p := by
+  rw [evalEval_C, eval_C]
+
+lemma evalEval_zero (x y : R) : (0 : R[X][Y]).evalEval x y = 0 := by
+  simp only [evalEval, eval_zero]
+
+lemma evalEval_one (x y : R) : (1 : R[X][Y]).evalEval x y = 1 := by
+  simp only [evalEval, eval_one]
+
+lemma evalEval_ofNat (x y : R) (n : ℕ) [n.AtLeastTwo] :
+    (OfNat.ofNat n : R[X][Y]).evalEval x y = OfNat.ofNat n := by
+  simp only [evalEval, eval_ofNat]
+
+lemma evalEval_natCast (x y : R) (n : ℕ) : (n : R[X][Y]).evalEval x y = n := by
+  simp only [evalEval, eval_natCast]
+
 lemma evalEval_X (x y : R) : X.evalEval x y = y := by
   rw [evalEval, eval_X, eval_C]
 
+lemma evalEval_add (x y : R) (p q : R[X][Y]) :
+    (p + q).evalEval x y = p.evalEval x y + q.evalEval x y := by
+  simp only [evalEval, eval_add]
+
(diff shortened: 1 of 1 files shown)


Multramate:
These are not strictly necessary, because we can just rewrite `evalEval` into `eval` (or into `evalEvalRingHom`), and apply the corresponding lemmas of `eval` (or of `evalEvalRingHom`), but are API that are "nice to have".

I noted, when doing this, that there are a few missing gaps in the API of `eval`, but I'll not touch them here.

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -41,15 +41,88 @@ abbrev CC (r : R) : R[X][Y] := C (C r)
 lemma evalEval_C (x y : R) (p : R[X]) : (C p).evalEval x y = p.eval x := by
   rw [evalEval, eval_C]
 
+lemma evalEval_CC (x y : R) (p : R) : (CC p).evalEval x y = p := by
+  rw [evalEval_C, eval_C]
+
+lemma evalEval_zero (x y : R) : (0 : R[X][Y]).evalEval x y = 0 := by
+  simp only [evalEval, eval_zero]
+
+lemma evalEval_one (x y : R) : (1 : R[X][Y]).evalEval x y = 1 := by
```
semorrison, line 50: I am surprised that most of the lemmas here are not `@[simp]`!

Multramate: I guess they should be...

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -41,15 +41,101 @@ abbrev CC (r : R) : R[X][Y] := C (C r)
...
+@[simp]
+lemma evalEval_natCast (x y : R) (n : ℕ) : (n : R[X][Y]).evalEval x y = n := by
+  simp only [evalEval, eval_natCast]
+
+@[simp]
 lemma evalEval_X (x y : R) : X.evalEval x y = y := by
   rw [evalEval, eval_X, eval_C]
 
+@[simp]
+lemma evalEval_add (x y : R) (p q : R[X][Y]) :
```
semorrison, line 70: Ugh, this is not a simp lemma for MvPolynomial, but is for Polynomial. Given that inconsistency, I don't mind what happens here, but if you feel like looking into that later, that would be wonderful. :-)

Mathlib/Algebra/Polynomial/Bivariate.lean
```diff
@@ -41,15 +41,101 @@ abbrev CC (r : R) : R[X][Y] := C (C r)
...
+lemma evalEval_zero (x y : R) : (0 : R[X][Y]).evalEval x y = 0 := by
+  simp only [evalEval, eval_zero]
+
+@[simp]
+lemma evalEval_one (x y : R) : (1 : R[X][Y]).evalEval x y = 1 := by
+  simp only [evalEval, eval_one]
+
+@[simp]
+lemma evalEval_ofNat (x y : R) (n : ℕ) [n.AtLeastTwo] :
+    (OfNat.ofNat n : R[X][Y]).evalEval x y = OfNat.ofNat n := by
```
semorrison, line 58: Is it necessary to write OfNat.ofNat here, rather than just n?

Ruben-VandeVelde, line 58: Does this work without no_index?

Multramate: What is `no_index`?

Multramate: I copied the corresponding lemma for `eval`. Without the `Ofnat.ofNat` this is the same as `evalEval_natCast`.

jcommelin (commented):
Thanks :tada:

bors merge

PR #14580, opened 2024-07-09
doc (Nat/Nth): add some docstrings, fix typos

rwst:
Fix 2 typos in the module docstrings, add docstrings to 37 lemmas.

---

### DIFF
--- a/Mathlib/Data/Nat/Nth.lean
+++ b/Mathlib/Data/Nat/Nth.lean
@@ -23,10 +23,10 @@ and provides lemmas that deal with this function and its connection to `Nat.coun
 
 ## Main results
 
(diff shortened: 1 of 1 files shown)


faenuccio:
I have removed the `awaiting-review` label according to the decision explained here
https://leanprover.zulipchat.com/#narrow/stream/113486-announce/topic/Changes.20to.20the.20.23queue/near/450176907

eric-wieser (commented):
The typo fixes look good, but most of these docstrings seem to me to fall in the same camp as `i++ // add one to i`, and we therefore don't want them.

Feel free to discuss more on Zulip if you disagree; but it would be great to merge the typo fixes by themselves before that.

rwst:
Will reopen with the typos fixes.

PR #14582, opened 2024-07-09
[Merged by Bors] - feat(Combinatorics/SimpleGraph): define `edist` and rewrite `dist`

Rida-Hamadani:
Define `edist` as the extended distance between two vertices of a simple graph, belonging to `ENat`, and rewrite `dist` as the coercion of `edist` to `Nat` using `toNat`. Rewrite the API to meet these new definitions.

---


Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
+/--
+The extended distance between two vertices is the length of the shortest walk between them.
+It is `⊤` if no such walk exists.
+-/
+noncomputable def edist (u v : V) : ℕ∞ :=
+  ⨅ w : G.Walk u v, w.length
+
+variable {G} {u v w : V}
+
+protected theorem Reachable.exists_walk_of_edist (hr : G.Reachable u v) :
```
kmill, line 55: The dist theorems don't seem to follow the mathlib naming convention. Would you mind updating them and these edist theorems? I'm thinking this could be `Reachable.exists_walk_length_eq_edist` (and similar for `Connected` and the pre-existing dist versions).

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
+
+/--
+The extended distance between two vertices is the length of the shortest walk between them.
+It is `⊤` if no such walk exists.
+-/
+noncomputable def edist (u v : V) : ℕ∞ :=
+  ⨅ w : G.Walk u v, w.length
+
+variable {G} {u v w : V}
+
```
kmill, line 54: I think it would be good to include
```lean
theorem edist_eq_sInf : G.edist u v = sInf (Set.range fun w : G.Walk u v ↦ (w.length : ℕ∞)) := rfl
```
so that it's possible to rewrite without trying to get things to unfold in the right way.

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
+theorem edist_comm : G.edist u v = G.edist v u := by
+  have {u v : V} (h : G.Reachable u v) : G.edist u v ≤ G.edist v u := by
+    obtain ⟨p, hp⟩ := h.symm.exists_walk_of_edist
+    rw [← hp, ← Walk.length_reverse]
+    apply edist_le
+  by_cases h : G.Reachable u v
+  · apply le_antisymm (this h) (this h.symm)
+  · have h' : ¬G.Reachable v u := fun h' => absurd h'.symm h
+    simp [h, h', edist_eq_top_of_not_reachable]
```
kmill, line 102: Now that we're using ENat, we don't need to reason by cases. First, we should have this theorem in Connectivity:
```lean
theorem Walk.reverse_surjective : Function.Surjective (Walk.reverse : G.Walk u v → _) :=
  Function.RightInverse.surjective Walk.reverse_reverse
```
And then we can do some rewriting:
```suggestion
theorem edist_comm : G.edist u v = G.edist v u := by
  rw [edist_eq_sInf, ← Set.image_univ, ← Set.image_univ_of_surjective Walk.reverse_surjective,
    ← Set.image_comp, Set.image_univ, Function.comp_def]
  simp_rw [Walk.length_reverse, ← edist_eq_sInf]
```
I feel like there's a missing iInf reindexing lemma (do you know @b-mehta ?)

kmill, line 102: Now that we're using ENat, we don't need to reason by cases. First, we should have this theorem in Connectivity:
```lean
lemma Walk.reverse_surjective : Function.Surjective (Walk.reverse : G.Walk u v → _) :=
  Function.RightInverse.surjective Walk.reverse_reverse
```
And then we can do some rewriting:
```suggestion
theorem edist_comm : G.edist u v = G.edist v u := by
  rw [edist_eq_sInf, ← Set.image_univ, ← Set.image_univ_of_surjective Walk.reverse_surjective,
    ← Set.image_comp, Set.image_univ, Function.comp_def]
  simp_rw [Walk.length_reverse, ← edist_eq_sInf]
```
I feel like there's a missing iInf reindexing lemma. I'd rather be able to say "reindex using this surjective function". (Do you know @b-mehta ?)

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
+The extended distance between vertices is equal to `1` if and only if these vertices are adjacent.
+-/
+theorem edist_eq_one_iff_adj : G.edist u v = 1 ↔ G.Adj u v := by
+  have ne_top_of_eq_one {n : ℕ∞} (h : n = 1) : n ≠ ⊤ := by
+    rw [← ENat.coe_toNat_eq_self, h]; rfl
+  refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
+  · let ⟨w, hw⟩ := exists_walk_of_edist_ne_top <| ne_top_of_eq_one h
+    rw [h, Nat.cast_eq_one] at hw
+    exact w.adj_of_length_eq_one <| hw
+  · exact ge_antisymm (ENat.one_le_iff_pos.mpr <| pos_edist_of_ne h.ne) <| edist_le h.toWalk
```
kmill, line 118: Changes:
* Use `obtain` instead of `let` since it's a proof (`have` would work too). There's no difference in this case, but `let` signals "I care about the definition"
* Rather than define a sub-lemma, it's possible to use `simp` (or `decide`) to handle the neq. This is a slight abuse of `rw`'s ability to unify metavariables, and more principled would be `obtain ⟨w, hw⟩ := exists_walk_of_edist_ne_top (by rw [h]; simp : G.edist u v ≠ ⊤)`.
* `le_antisymm` instead of `ge_antisymm` since the operands are in terms of le, for clarity.

```suggestion
theorem edist_eq_one_iff_adj : G.edist u v = 1 ↔ G.Adj u v := by
  refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
  · obtain ⟨w, hw⟩ := exists_walk_of_edist_ne_top <| by rw [h]; simp
    rw [h, Nat.cast_eq_one] at hw
    exact w.adj_of_length_eq_one hw
  · exact le_antisymm (edist_le h.toWalk) (ENat.one_le_iff_pos.mpr <| pos_edist_of_ne h.ne)
```

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
   obtain ⟨p, hp⟩ := h.symm.exists_walk_of_dist
   rw [← hp, ← Walk.length_reverse]
   apply dist_le
 
-theorem dist_comm {u v : V} : G.dist u v = G.dist v u := by
+theorem dist_comm : G.dist u v = G.dist v u := by
   by_cases h : G.Reachable u v
   · apply le_antisymm (dist_comm_aux h) (dist_comm_aux h.symm)
   · have h' : ¬G.Reachable v u := fun h' => absurd h'.symm h
     simp [h, h', dist_eq_zero_of_not_reachable]
```
kmill, line 205: We can delete `dist_comm_aux` and make use of `edist_comm`:
```suggestion
theorem dist_comm : G.dist u v = G.dist v u := by
  by_cases h : G.Reachable u v
  · rw [dist, edist_comm, ← dist]
  · have h' : ¬G.Reachable v u := fun h' => absurd h'.symm h
    simp [h, h', dist_eq_zero_of_not_reachable]
```

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,190 @@ variable {V : Type*} (G : SimpleGraph V)
...
   (Reachable.of_dist_ne_zero h).exists_walk_of_dist
 
 /- The distance between vertices is equal to `1` if and only if these vertices are adjacent. -/
-theorem dist_eq_one_iff_adj {u v : V} : G.dist u v = 1 ↔ G.Adj u v := by
+theorem dist_eq_one_iff_adj : G.dist u v = 1 ↔ G.Adj u v := by
   refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
   · let ⟨w, hw⟩ := exists_walk_of_dist_ne_zero <| ne_zero_of_eq_one h
     exact w.adj_of_length_eq_one <| h ▸ hw
   · have : h.toWalk.length = 1 := Walk.length_cons _ _
     exact ge_antisymm (h.reachable.pos_dist_of_ne h.ne) (this ▸ dist_le _)
```
kmill, line 225: Again, let's make use of the edist lemma:
```suggestion
theorem dist_eq_one_iff_adj : G.dist u v = 1 ↔ G.Adj u v := by
  rw [dist, ENat.toNat_eq_iff, ENat.coe_one, edist_eq_one_iff_adj]
  decide
```

kmill (commented):
Thanks! Overall it looks good. Some suggestions below:

Mathlib/Combinatorics/SimpleGraph/Metric.lean
Rida-Hamadani: Thank you.  I've updated the naming of the lemmas that don't follow the naming convention, but I find naming lemmas confusing as there are lots of exceptions and subtleties.  Please let me know if the names are good now or if any need changing.

kmill: Sorry, I meant to say *just* these four theorems needed changing. Could you revert your last commit please?

Rida-Hamadani: Sure!

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
```diff
@@ -372,6 +372,9 @@ theorem reverse_reverse {u v : V} (p : G.Walk u v) : p.reverse.reverse = p := by
   | cons _ _ ih => simp [ih]
 #align simple_graph.walk.reverse_reverse SimpleGraph.Walk.reverse_reverse
 
+theorem reverse_surjective {u v : V} : Function.Surjective (reverse : G.Walk u v → _) :=
```
b-mehta, line 375: I think Involutive might be the correct thing to have here, maybe Bijective / Surj / Inj could be given from that but imo Involutive is enough

Mathlib/Data/ENat/Lattice.lean
```diff
@@ -45,7 +45,19 @@ lemma coe_iSup : BddAbove (range f) → ↑(⨆ i, f i) = ⨆ i, (f i : ℕ∞)
 @[norm_cast] lemma coe_iInf [Nonempty ι] : ↑(⨅ i, f i) = ⨅ i, (f i : ℕ∞) :=
   WithTop.coe_iInf (OrderBot.bddBelow _)
 
-variable {s : Set ℕ∞}
+lemma iInf_toNat : (⨅ i, (f i : ℕ∞)).toNat = ⨅ i, f i := by
+  cases isEmpty_or_nonempty ι
+  · simp [iInf_coe_eq_top.mpr ‹_›]
```
b-mehta, line 50: I think a simp lemma saying `⨅ i, (f i : ℕ∞) = ⊤` in the case that iota is empty would be sensible, and also simplify this bit of the proof

Mathlib/Combinatorics/SimpleGraph/Metric.lean
b-mehta: Hmm not sure, although if we have bijectivity it could be even easier. Will think about this.

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,176 @@ variable {V : Type*} (G : SimpleGraph V)
...
+
+protected theorem Reachable.exists_walk_length_eq_edist (hr : G.Reachable u v) :
+    ∃ p : G.Walk u v, p.length = G.edist u v :=
+  csInf_mem <| Set.range_nonempty_iff_nonempty.mpr hr
+
+protected theorem Connected.exists_walk_length_eq_edist  (hconn : G.Connected) (u v : V) :
+    ∃ p : G.Walk u v, p.length = G.edist u v :=
+  (hconn u v).exists_walk_length_eq_edist
+
+theorem edist_le (p : G.Walk u v) :
```
b-mehta, line 65: Maybe an alias for Walk.edist_le could be nice

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,176 @@ variable {V : Type*} (G : SimpleGraph V)
...
+
+@[simp]
+theorem edist_eq_zero_iff :
+    G.edist u v = 0 ↔ u = v := by
+  apply Iff.intro <;> simp [edist, ENat.iInf_eq_zero]
+
+theorem edist_self : edist G v v = 0 :=
+  edist_eq_zero_iff.mpr rfl
+
+theorem pos_edist_of_ne (hne : u ≠ v) :
```
b-mehta, line 77: alias `Ne.edist_pos` maybe?

also I think this should be edist_pos_of_ne

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,176 @@ variable {V : Type*} (G : SimpleGraph V)
...
+    0 < G.edist u v :=
+  pos_iff_ne_zero.mpr <| edist_eq_zero_iff.ne.mpr hne
+
+lemma edist_eq_top_of_not_reachable (h : ¬G.Reachable u v) :
+    G.edist u v = ⊤ := by
+  simp [edist, not_reachable_iff_isEmpty_walk.mp h]
+
+theorem Reachable.of_edist_ne_top (h : G.edist u v ≠ ⊤) :
```
b-mehta, line 85: I wouldn't use `Reachable.` here, since dot notation won't apply. `reachable_of_edist_ne_top` would be better

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,176 @@ variable {V : Type*} (G : SimpleGraph V)
...
+  simp_rw [Walk.length_reverse, ← edist_eq_sInf]
+
+lemma exists_walk_of_edist_ne_top (h : G.edist u v ≠ ⊤) :
+    ∃ p : G.Walk u v, p.length = G.edist u v :=
+  (Reachable.of_edist_ne_top h).exists_walk_length_eq_edist
+
+/--
+The extended distance between vertices is equal to `1` if and only if these vertices are adjacent.
+-/
+theorem edist_eq_one_iff_adj : G.edist u v = 1 ↔ G.Adj u v := by
```
b-mehta, line 108: Might be a good simp lemma

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,176 @@ variable {V : Type*} (G : SimpleGraph V)
...
   apply dist_le
 #align simple_graph.connected.dist_triangle SimpleGraph.Connected.dist_triangle
 
-private theorem dist_comm_aux {u v : V} (h : G.Reachable u v) : G.dist u v ≤ G.dist v u := by
-  obtain ⟨p, hp⟩ := h.symm.exists_walk_of_dist
-  rw [← hp, ← Walk.length_reverse]
-  apply dist_le
-
-theorem dist_comm {u v : V} : G.dist u v = G.dist v u := by
+theorem dist_comm : G.dist u v = G.dist v u := by
```
b-mehta, line 190: you can actually do this whole proof with just `rw [dist, dist, edist_comm]`

b-mehta (approved):

Mathlib/Combinatorics/SimpleGraph/Metric.lean
kmill: Probably like this:
```
protected alias Walk.edist_le := edist_le
```

Mathlib/Combinatorics/SimpleGraph/Metric.lean
Rida-Hamadani: Should I also change the `dist` versions of the lemmas and alias them too?

kmill: Are you sure about this alias Bhavik? This is putting a graph-specific theorem in the global Ne namespace. We could have `Ne.simpleGraph_walk_edist_pos` or something like that, but otherwise I'd suggest not having such an alias.

Mathlib/Combinatorics/SimpleGraph/Connectivity.lean
Rida-Hamadani: I'm facing an issue here: `Function.Involutive` expects the function to have the same domain type and range type, but for reverse, the domain is of type `G.Walk u v` while its range is of type `G.Walk v u`.

kmill: Oh yeah, that's a fundamental issue. How about for now we settle for `reverse_surjective` for now? If you want you can also throw in `reverse_injective` and `reverse_bijective`.

There's a way to state that it's involutive, but what comes to mind involves sigma types and isn't very nice:
```
theorem Walk.reverse_involutive :
    Function.Involutive
      (Sigma.map Prod.swap (fun _ => Walk.reverse) : (p : V × V) × G.Walk p.1 p.2 → (p : V × V) × G.Walk p.1 p.2) := by
  intro ⟨u, v⟩
  simp [Sigma.map]
```

b-mehta: Oh right, good point! Yeah don't do involutive

Mathlib/Combinatorics/SimpleGraph/Metric.lean
b-mehta: Oh true, `Ne.edist_pos` is probably too non-specific to be global

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,179 @@ variable {V : Type*} (G : SimpleGraph V)
...
+  apply edist_le
+
+theorem edist_comm : G.edist u v = G.edist v u := by
+  rw [edist_eq_sInf, ← Set.image_univ, ← Set.image_univ_of_surjective Walk.reverse_surjective,
+    ← Set.image_comp, Set.image_univ, Function.comp_def]
+  simp_rw [Walk.length_reverse, ← edist_eq_sInf]
+
+lemma exists_walk_of_edist_ne_top (h : G.edist u v ≠ ⊤) :
+    ∃ p : G.Walk u v, p.length = G.edist u v :=
+  (reachable_of_edist_ne_top h).exists_walk_length_eq_edist
```
b-mehta, line 105:
```
lemma exists_walk_of_edist_eq_coe {k : ℕ} (h : G.edist u v = k) :
    ∃ p : G.Walk u v, p.length = k :=
  have : G.edist u v ≠ ⊤ := by rw [h]; exact ENat.coe_ne_top _
  have ⟨p, hp⟩ := exists_walk_of_edist_ne_top this
  ⟨p, Nat.cast_injective (hp.trans h)⟩
```
this seems helpful, I think.

Also a statement that edist isn't top iff the vertices are reachable (as an iff) would be good

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -43,105 +41,179 @@ variable {V : Type*} (G : SimpleGraph V)
...
+
+lemma edist_eq_top_of_not_reachable (h : ¬G.Reachable u v) :
+    G.edist u v = ⊤ := by
+  simp [edist, not_reachable_iff_isEmpty_walk.mp h]
+
+theorem reachable_of_edist_ne_top (h : G.edist u v ≠ ⊤) :
+    G.Reachable u v :=
+  not_not.mp <| edist_eq_top_of_not_reachable.mt h
+
+protected theorem Connected.edist_triangle (hconn : G.Connected) :
```
b-mehta, line 91: This doesn't need connectedness since we're using edist!
```
protected theorem edist_triangle : G.edist u w ≤ G.edist u v + G.edist v w := by
  rcases eq_or_ne (G.edist u v) ⊤ with huv | huv
  case inl => simp [huv]
  case inr =>
    rcases eq_or_ne (G.edist v w) ⊤ with hvw | hvw
    case inl => simp [hvw]
    case inr =>
      obtain ⟨p, hp⟩ := exists_walk_of_edist_ne_top huv
      obtain ⟨q, hq⟩ := exists_walk_of_edist_ne_top hvw
      rw [←hp, ←hq, ←Nat.cast_add, ←Walk.length_append]
      exact edist_le _
 ```

Comment (kmill) [reply Mathlib/Combinatorics/SimpleGraph/Metric.lean:91]: It's probably not worth developing all this theory, but it occurred to me that we could pull back infinity to Walk to have "extended walks", along with the "does not exist" walk. Then `edist_triangle` could be proved without any cases, because it's already been done in `EWalk.length_append`.
```lean
def EWalk (u v : V) := Option (G.Walk u v)

def EWalk.length : G.EWalk u v → ℕ∞ := Option.map Walk.length

def EWalk.append (p : G.EWalk u v) (q : G.EWalk v w) : G.EWalk u w :=
  Walk.append <$> p <*> q

def EWalk.length_append {p : G.EWalk u v} {q : G.EWalk v w} :
    (p.append q).length = p.length + q.length := by
  cases p <;> cases q <;> simp [EWalk.append, EWalk.length] <;> rfl

theorem exists_ewalk_length_eq_edist :
    ∃ p : G.EWalk u v, p.length = G.edist u v := by
  sorry
```

Rida-Hamadani: It is a great idea! I'll keep it in mind to potentially work on.
I think it would be nicer to move `Walk`s from `Connectivity.lean` into their own file (in the `Connectivity` folder) before attempting that, what do you think?

b-mehta: I think that change does sound valuable - Reachable and Connected can be in Connectivity.lean; Walks and Paths can be in Walk.lean

Rida-Hamadani: #15171 does this.

b-mehta:
Nice work!

bors merge

PR #14586, opened 2024-07-09
[Merged by Bors] - fix(nlinarith): make `findSquares` deterministic

eric-wieser:
Previously the order in which the assumptions were added was determined by `Expr` hashes, which while deterministic are not part of the user's mental model.

This matters (see [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/nlinarith.20is.20affected.20by.20.60example.60s/near/449994710)) because it then affects the order in which metavariables are unified.

As a bonus(?), this now means that squares are now found up to reducible defeq.

---

### DIFF
--- a/Mathlib/Tactic/Linarith/Preprocessing.lean
+++ b/Mathlib/Tactic/Linarith/Preprocessing.lean
@@ -350,11 +350,8 @@ def nlinarithExtras : GlobalPreprocessor where
       let si ← ls.foldrM (fun h s' => do findSquares s' (← instantiateMVars (← inferType h)))
         HashSet.empty
       si.toList.mapM fun (i, is_sq) => return ((← get).atoms[i]!, is_sq)
-    let new_es ← s.filterMapM fun (e, is_sq) => do
-      try
-        mkAppM (if is_sq then ``sq_nonneg else ``mul_self_nonneg) #[e]
-      catch _ =>
-        pure none
+    let new_es ← s.filterMapM fun (e, is_sq) =>
+      observing? <| mkAppM (if is_sq then ``sq_nonneg else ``mul_self_nonneg) #[e]
     let new_es ← compWithZero.globalize.transform new_es
     trace[linarith] "nlinarith preprocessing found squares"
     trace[linarith] "{s}"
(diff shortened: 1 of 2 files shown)


Mathlib/Tactic/Linarith/Preprocessing.lean
```diff
@@ -306,9 +307,9 @@ section nlinarith
 /--
 `findSquares s e` collects all terms of the form `a ^ 2` and `a * a` that appear in `e`
 and adds them to the set `s`.
-A pair `(a, true)` is added to `s` when `a^2` appears in `e`,
-and `(a, false)` is added to `s` when `a*a` appears in `e`.  -/
-partial def findSquares (s : HashSet (Expr × Bool)) (e : Expr) : MetaM (HashSet (Expr × Bool)) :=
+A pair `(i, true)` is idded to `s` when `atoms[i]^2` appears in `e`,
```
llllvvuu, line 310:
```suggestion
A pair `(i, true)` is added to `s` when `atoms[i]^2` appears in `e`,
```

Mathlib/Tactic/Linarith/Preprocessing.lean
```diff
@@ -340,15 +345,16 @@ This preprocessor is typically run last, after all inputs have been canonized.
...
-      HashSet.empty
-    let new_es ← s.foldM (fun new_es (⟨e, is_sq⟩ : Expr × Bool) =>
-      ((do
-        let p ← mkAppM (if is_sq then ``sq_nonneg else ``mul_self_nonneg) #[e]
-        pure <| p::new_es) <|> pure new_es)) ([] : List Expr)
+    -- find the squares in `AtomM` to ensure deterministic behavior
+    let s ← AtomM.run .reducible do
+      let si ← ls.foldrM (fun h s' => do findSquares s' (← instantiateMVars (← inferType h)))
+        HashSet.empty
+      si.toList.mapM fun (i, is_sq) => return ((← get).atoms[i]!, is_sq)
```
kmill, line 352: To make this more deterministic, could we sort `si.toList`? Otherwise it comes out in whatever order the hash turns out to be. Let's sort by the lex order of the pairs.

eric-wieser: How about using an `RBSet` instead, so that we can do the sorting as we go?

kmill: Sure, seems reasonable.

kmill:
Looks good, thanks!

bors d+

eric-wieser:
bors merge

PR #14591, opened 2024-07-09
[Merged by Bors] - feat(Topology/Algebra/Order/LiminfLimsup): generalize map liminf and map limsup lemmas

kkytola:
Generalize map liminf and map limsup lemmas. Instead of assuming boundedness in both directions, it suffices to assume boundedness in one direction and coboundedness in the other direction.

---

This generalization resulted from [a review comment](https://github.com/leanprover-community/mathlib4/pull/13938#discussion_r1649744441) of @sgouezel asking for the strenghtening of a specific ENNReal lemma about liminfs.


Mathlib/Topology/Algebra/Order/LiminfLimsup.lean
```diff
@@ -347,70 +347,109 @@ theorem Antitone.map_limsSup_of_continuousAt {F : Filter R} [NeBot F] {f : R →
...
 `Filter.limsup` to the `Filter.liminf` of the images (if the filter is bounded from above and
-below). -/
+frequently bounded from below). -/
 theorem Antitone.map_limsup_of_continuousAt {f : R → S} (f_decr : Antitone f) (a : ι → R)
     (f_cont : ContinuousAt f (F.limsup a))
     (bdd_above : F.IsBoundedUnder (· ≤ ·) a := by isBoundedDefault)
-    (bdd_below : F.IsBoundedUnder (· ≥ ·) a := by isBoundedDefault) :
+    (cobdd : F.IsCoboundedUnder (· ≤ ·) a := by isBoundedDefault) :
     f (F.limsup a) = F.liminf (f ∘ a) :=
+<<<<<<< HEAD
```
semorrison, line 356: Oops, merge conflicts committed.

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1517,3 +1517,98 @@ theorem liminf_finset_inf [ConditionallyCompleteLinearOrder β] [OrderTop β] {f
   limsup_finset_sup (β := βᵒᵈ) h₁ h₂
 
 end MinMax
+
+section frequently_bounded
+
+variable {R S : Type*} {F : Filter R} [LinearOrder R] [LinearOrder S]
+
+/-- For nontrivial filters in linear orders, coboundedness for `≤` implies frequently boundedness
```
j-loreaux, line 1525: I think "frequent boundedness" might make sense, but adverbs can't really modify nouns like this. Perhaps a complete rewrite of the sentence would be an alternative. Applies here and throughout.

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1517,3 +1517,98 @@ theorem liminf_finset_inf [ConditionallyCompleteLinearOrder β] [OrderTop β] {f
...
+from below. -/
+lemma IsCobounded.frequently_ge [NeBot F] (cobdd : IsCobounded (· ≤ ·) F) :
+    ∃ l, ∃ᶠ x in F, l ≤ x := by
```
j-loreaux, line 1528: Isn't the best result stronger than this? That is, can't you get `\forall\^f l in atTop, ...`? Or am I missing something obvious?

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1517,3 +1517,98 @@ theorem liminf_finset_inf [ConditionallyCompleteLinearOrder β] [OrderTop β] {f
...
+    exact tbot this
+  refine ⟨t', ?_⟩
+  intro ev
+  specialize ht t' (by filter_upwards [ev] with _ h using (not_le.mp h).le)
+  apply lt_irrefl t' <| lt_of_lt_of_le ht' ht
+
+/-- For nontrivial filters in linear orders, coboundedness for `≥` implies frequently boundedness
+from above. -/
+lemma IsCobounded.frequently_le [NeBot F] (cobdd : IsCobounded (· ≥ ·) F) :
+    ∃ u, ∃ᶠ x in F, x ≤ u :=
```
j-loreaux, line 1543: Same here

Mathlib/Order/LiminfLimsup.lean
```diff
@@ -1517,3 +1517,98 @@ theorem liminf_finset_inf [ConditionallyCompleteLinearOrder β] [OrderTop β] {f
...
+  IsCobounded.frequently_ge (R := Rᵒᵈ) cobdd
+
+/-- In linear orders, frequently boundedness from below implies coboundedness for `≤`. -/
+lemma isCobounded_le_of_frequently_ge {l : R} (freq_ge : ∃ᶠ x in F, l ≤ x) :
```
j-loreaux, line 1547: We're now putting stuff like this in the `IsCobounded` namespace. It allows for dot notation (where there is no text preceding the dot and the expected type is known).

j-loreaux (commented):
Thanks! This is a nice generalization.

Mathlib/Order/LiminfLimsup.lean
kkytola: I guess `\forall\^f l in atBot, ...` is possible, but in my opinion it seems to hide the usually desired existential (a value of a frequent lower bound) under a less useful quantifier. I believe in almost all use cases one would then have to use `Eventually.exists` + `Exists.choose`.

Unless you think that the generalization of the existential is valuable for some use cases, I'm tempted to keep the current form.

Mathlib/Order/LiminfLimsup.lean
kkytola: I now at least temporarily added `IsCobounded.eventually_atTop_frequently_le`, although I am not sure of its usefulness, so I would be happy to remove it.

Mathlib/Order/LiminfLimsup.lean
kkytola: Thanks! This made me realize that I had made a mistake in the namespaces, because they need to be further under the `Filter` namespace (so `Filter.IsCobounded` etc.). I did not right away get the dot notation working, however, but that is probably my ignorance of the syntax and from my point of view the naming in `Filter.IsCobounded` namespace would be ok in any case. I changed the naming.

Mathlib/Order/LiminfLimsup.lean
j-loreaux: Note: you *reversed* the `atTop` and `atBot` from what I intended. But it's possible what I wanted just isn't true. In that case, you can remove the new lemmas you added.

kkytola: Oh, I thought `atTop` was a misprint and should have been `atBot`. The `atTop` spelling is just false: it would say that a cobounded sequence has arbitrarily large frequent lower bounds (the constant sequence of real zeroes is a counterexample).

The combination of lemmas `IsCobounded.frequently_ge` + `IsCobounded.of_frequently_ge` says that a (nontrivial) filter is cobounded (in the `≤` direction) if and only if it has _some_ frequent lower bound. (In particular this makes it obvious that the constant sequence is cobounded and gives the counterexample to arbitrarily large frequent lower bounds.) Maybe an "iff" version would be worth including? I think I ended up wanting that in the follow-up PR.

So I remove the superficially stronger phrasing now. I indeed cannot imagine any use cases for it.

j-loreaux: Thanks for the explanation. Obviously I was not thinking clearly.

j-loreaux:
bors merge

PR #9734, opened 2024-01-14
[Merged by Bors] - feat: Binary entropy

adomasbaliuka:
Define the (Shannon) binary and q-ary entropy functions.

The binary entropy function is used in information theory; for example, in Shannon's theorems.

---

- [x] depends on: #14594
- [x] depends on: #15828
- [x] depends on: #15829

The lemmas are mostly "trivial", especially using `Mathlib.Analysis.SpecialFunctions.Log.NegMulLog`.
This is the kind of application that `NegMulLog` was made for.

If Mathlib is to include Shannon's theorems in the future (already formalized in Coq), this might hopefully save someone a few minutes.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14593, opened 2024-07-09
[Merged by Bors] - feat: Lemmas `DifferentiableAt.add_iff_left`, `DifferentiableAt.add_iff_right`

adomasbaliuka:
Given a function differentiable at a point, the result of adding a second function is differentiable (there) iff the second function is differentiable (there).

---
The part that's new is: at a point, the sum of a differentiable and a non-differentiable function is non-differentiable.
The other direction of the `iff` is already covered by `DifferentiableAt.add`.

Lemmas are marked `@[simp]` because other `iff` lemmas in that file are marked as such.

Needed for #9734 

### DIFF
--- a/Mathlib/Analysis/Calculus/FDeriv/Add.lean
+++ b/Mathlib/Analysis/Calculus/FDeriv/Add.lean
@@ -559,6 +559,30 @@ theorem DifferentiableOn.sub (hf : DifferentiableOn 𝕜 f s) (hg : Differentiab
     DifferentiableOn 𝕜 (fun y => f y - g y) s := fun x hx => (hf x hx).sub (hg x hx)
 #align differentiable_on.sub DifferentiableOn.sub
 
+@[simp]
+lemma DifferentiableOn.add_iff_left (hg : DifferentiableOn 𝕜 g s) :
+    DifferentiableOn 𝕜 (fun y => f y + g y) s ↔ DifferentiableOn 𝕜 f s := by
+  refine ⟨fun h ↦ ?_, fun hf ↦ hf.add hg⟩
+  simpa only [add_sub_cancel_right] using h.sub hg
+
+@[simp]
+lemma DifferentiableOn.add_iff_right (hg : DifferentiableOn 𝕜 f s) :
+    DifferentiableOn 𝕜 (fun y => f y + g y) s ↔ DifferentiableOn 𝕜 g s := by
+  simp only [add_comm (f _), hg.add_iff_left]
+
+@[simp]
+lemma DifferentiableOn.sub_iff_left (hg : DifferentiableOn 𝕜 g s) :
+    DifferentiableOn 𝕜 (fun y => f y - g y) s ↔ DifferentiableOn 𝕜 f s := by
+  refine ⟨fun h ↦ ?_, fun hf ↦ hf.sub hg⟩
+  simpa only [sub_add_cancel] using h.add hg
+
+@[simp]
+lemma DifferentiableOn.sub_iff_right (hg : DifferentiableOn 𝕜 f s) :
+    DifferentiableOn 𝕜 (fun y => f y - g y) s ↔ DifferentiableOn 𝕜 g s := by
+  refine ⟨fun h ↦ ?_, fun hf ↦ hg.sub hf⟩
+  simpa only [sub_sub_cancel_left, differentiableOn_neg_iff] using h.sub hg
+
+
 @[simp, fun_prop]
 theorem Differentiable.sub (hf : Differentiable 𝕜 f) (hg : Differentiable 𝕜 g) :
     Differentiable 𝕜 fun y => f y - g y := fun x => (hf x).sub (hg x)
--- a/Mathlib/Analysis/Calculus/FDeriv/Add.lean
+++ b/Mathlib/Analysis/Calculus/FDeriv/Add.lean
@@ -534,13 +534,8 @@ theorem DifferentiableAt.sub (hf : DifferentiableAt 𝕜 f x) (hg : Differentiab
 @[simp]
 lemma DifferentiableAt.add_iff_left (hg : DifferentiableAt 𝕜 g x) :
     DifferentiableAt 𝕜 (fun y => f y + g y) x ↔ DifferentiableAt 𝕜 f x := by
-  constructor <;> intro h
-  · have f_eq_sum_sub_g: f = (fun y => f y + g y) - g := by
-      ext
-      simp only [Pi.sub_apply, add_sub_cancel_right]
-    rw [f_eq_sum_sub_g]
-    exact DifferentiableAt.sub h hg
-  · simp_all only [DifferentiableAt.add]
+  refine ⟨fun h ↦ ?_, fun hf ↦ hf.add hg⟩
+  simpa using h.sub hg
 
 @[simp]
 lemma DifferentiableAt.add_iff_right (hg : DifferentiableAt 𝕜 f x) :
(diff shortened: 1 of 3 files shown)


urkud:
Let's test if `@[simp]` causes troubles.

urkud:
!bench

Mathlib/Analysis/Calculus/FDeriv/Add.lean
```diff
@@ -531,6 +531,23 @@ theorem DifferentiableAt.sub (hf : DifferentiableAt 𝕜 f x) (hg : Differentiab
...
+@[simp]
+lemma DifferentiableAt.add_iff_left (hg : DifferentiableAt 𝕜 g x) :
+    DifferentiableAt 𝕜 (fun y => f y + g y) x ↔ DifferentiableAt 𝕜 f x := by
+  constructor <;> intro h
+  · have f_eq_sum_sub_g: f = (fun y => f y + g y) - g := by
+      ext
+      simp only [Pi.sub_apply, add_sub_cancel_right]
+    rw [f_eq_sum_sub_g]
+    exact DifferentiableAt.sub h hg
+  · simp_all only [DifferentiableAt.add]
```
urkud, line 543: Not tested, but should work:
```suggestion
  refine ⟨fun h ↦ ?_, fun hf ↦ hf.add hg⟩
  simpa using h.sub hg
```

Mathlib/Analysis/Calculus/FDeriv/Add.lean
```diff
@@ -531,6 +531,23 @@ theorem DifferentiableAt.sub (hf : DifferentiableAt 𝕜 f x) (hg : Differentiab
...
+
+@[simp]
+lemma DifferentiableAt.add_iff_right (hg : DifferentiableAt 𝕜 f x) :
+    DifferentiableAt 𝕜 (fun y => f y + g y) x ↔ DifferentiableAt 𝕜 g x := by
+  rw [show (fun y ↦ f y + g y) = (fun y ↦ g y + f y) by ext; rw [add_comm]]
+  exact hg.add_iff_left
```
urkud, line 549:
```suggestion
  simp only [add_comm (f _), hg.add_iff_left]
```

urkud:
@YaelDillies Could you please remind me what's our convention about `left`/`right` in lemmas like these?

urkud:
LGTM but I don't remember what's our convention about `left`/`right` here, so delegating to Yael who probably remembers it.
bors d=@YaelDillies

YaelDillies (commented):
`binOp_iff_left` should be of the form `P (binOp f g) \iff P f`, and `binOp_iff_right` should be of the form `P (binOp f g) \iff P g`.

Can you add the corresponding `sub` lemmas?

adomasbaliuka:
> Can you add the corresponding `sub` lemmas?

Now the linter complains that old lemmas can be proved by simp. What to do?
- Do not mark `sub_iff_*` as `@[simp]`
- Remove the `simp` mark from the old lemmas (which concern the special case with a constant)
- ?

YaelDillies:
Remove the simp from the old lemmas (/deprecate the old lemmas since they are superseded?)

Can you also add the `DifferentiableOn` version of your new lemmas?

Mathlib/Analysis/Calculus/FDeriv/Add.lean
```diff
@@ -624,7 +670,7 @@ theorem DifferentiableAt.sub_const (hf : DifferentiableAt 𝕜 f x) (c : F) :
   (hf.hasFDerivAt.sub_const c).differentiableAt
 #align differentiable_at.sub_const DifferentiableAt.sub_const
 
-@[simp]
+@[deprecated DifferentiableAt.sub_iff_left]
```
YaelDillies, line 673: Please add `(since := ... )` clauses to your deprecation. Note that we have a `deprecated alias` code snippet to help you with this (although currently it doesn't do exactly what you want here)
```suggestion
@[deprecated DifferentiableAt.sub_iff_left (since := 2024-07-11)]
```

Also please golf the proof of the lemmas you deprecate so that people know how to replace the deprecated lemmas by the non-deprecated ones

Mathlib/Analysis/Calculus/FDeriv/Add.lean
```diff
@@ -670,32 +670,32 @@ theorem DifferentiableAt.sub_const (hf : DifferentiableAt 𝕜 f x) (c : F) :
   (hf.hasFDerivAt.sub_const c).differentiableAt
 #align differentiable_at.sub_const DifferentiableAt.sub_const
 
-@[deprecated DifferentiableAt.sub_iff_left]
+@[deprecated DifferentiableAt.sub_iff_left (since := "2024-07-11")]
 theorem differentiableAt_sub_const_iff (c : F) :
-    DifferentiableAt 𝕜 (fun y => f y - c) x ↔ DifferentiableAt 𝕜 f x := by
-  simp only [sub_eq_add_neg, differentiableAt_add_const_iff]
+    DifferentiableAt 𝕜 (fun y => f y - c) x ↔ DifferentiableAt 𝕜 f x :=
+  DifferentiableAt.sub_iff_left (differentiableAt_const _)
```
YaelDillies, line 676: If you didn't know about this, this is called dot notation and is the entire point of naming lemmas about `FancyProp` like `FancyProp.add`, `FancyProp.neg`, ...
```suggestion
  (differentiableAt_const _).sub_iff_left 
```

adomasbaliuka: Thanks. I did know about this but just "copy pasted" without thinking.

Mathlib/Analysis/Calculus/FDeriv/Add.lean
```diff
@@ -531,16 +531,85 @@ theorem DifferentiableAt.sub (hf : DifferentiableAt 𝕜 f x) (hg : Differentiab
...
+@[simp]
+lemma DifferentiableAt.add_iff_right (hg : DifferentiableAt 𝕜 f x) :
+    DifferentiableAt 𝕜 (fun y => f y + g y) x ↔ DifferentiableAt 𝕜 g x := by
+  simp only [add_comm (f _), hg.add_iff_left]
+
+@[simp]
+lemma DifferentiableAt.sub_iff_left (hg : DifferentiableAt 𝕜 g x) :
+    DifferentiableAt 𝕜 (fun y => f y - g y) x ↔ DifferentiableAt 𝕜 f x := by
+  refine ⟨fun h ↦ ?_, fun hf ↦ hf.sub hg⟩
+  simpa only [sub_add_cancel] using h.add hg
```
YaelDillies, line 549:
```suggestion
  simp [sub_eq_add_neg, hg]
```
should work, hopefully. If it doesn't, please make it work by adding `DifferentiableAt 𝕜 (fun y => -f y) ↔ DifferentiableAt 𝕜 f x` as a simp lemma

adomasbaliuka: Does work. Replaced by `simp only` instead because I'm guessing it's faster and more stable to future changes.

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/Analysis/Calculus/FDeriv/Add.lean
YaelDillies: Faster, a bit. More stable to future changes, not necessarily since `simp only` requires you to name lemmas which might be renamed in the future

adomasbaliuka: > `simp only` requires you to name lemmas which might be renamed in the future
I would think renaming lemmas is much easier than figuring out which new `simp` lemma is now preventing a proof from working and why.

YaelDillies: In my experience, not necessarily, no. Further, renamings happen much more often than terminal simp calls not working anymore

jcommelin (commented):
bors merge

PR #14600, opened 2024-07-10
[Merged by Bors] - feat(SetTheory/Surreal/Basic): add some lemmas for Surreal.mk

hwatheod:
add some lemmas for Surreal.mk

---
Split out of #14498 

This contains all the lemmas in #14278 plus a few more. I was unable to reopen that one (I think I messed something up by doing a force push), so I am creating a new one.

### DIFF
--- a/Mathlib/SetTheory/Surreal/Basic.lean
+++ b/Mathlib/SetTheory/Surreal/Basic.lean
@@ -338,21 +338,13 @@ lemma mk_lt_mk {x y : PGame.{u}} {hx hy} : mk x hx < mk y hy ↔ x < y := Iff.rf
 
 lemma zero_lt_mk {x : PGame.{u}} {hx} : 0 < mk x hx ↔ 0 < x := Iff.rfl
 
-/-- Same as moveLeft_lt, but for `Surreal` instead of `PGame` -/
+/-- Same as `moveLeft_lt`, but for `Surreal` instead of `PGame` -/
 theorem mk_moveLeft_lt_mk {x : PGame} (o : Numeric x) (i) :
-    Surreal.mk (x.moveLeft i) (by
-      rw [numeric_def] at o
-      exact o.2.1 i
-    ) < Surreal.mk x o := by
-  exact Numeric.moveLeft_lt o i
+    Surreal.mk (x.moveLeft i) (Numeric.moveLeft o i) < Surreal.mk x o := Numeric.moveLeft_lt o i
 
-/-- Same as lt_moveRight, but for `Surreal` instead of `PGame` -/
+/-- Same as `lt_moveRight`, but for `Surreal` instead of `PGame` -/
 theorem mk_lt_mk_moveRight {x : PGame} (o : Numeric x) (j) :
-    Surreal.mk x o < Surreal.mk (x.moveRight j) (by
-      rw [numeric_def] at o
-      exact o.2.2 j
-    ) := by
-  exact Numeric.lt_moveRight o j
+    Surreal.mk x o < Surreal.mk (x.moveRight j) (Numeric.moveRight o j) := Numeric.lt_moveRight o j
 
 /-- Addition on surreals is inherited from pre-game addition:
 the sum of `x = {xL | xR}` and `y = {yL | yR}` is `{xL + y, x + yL | xR + y, x + yR}`. -/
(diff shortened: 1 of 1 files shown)


Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -301,6 +301,10 @@ instance : One Surreal :=
 instance : Inhabited Surreal :=
   ⟨0⟩
 
+lemma mk_eq_mk {x y : PGame.{u}} {hx hy} : mk x hx = mk y hy ↔ x ≈ y := Quotient.eq
```
semorrison, line 304: Aren't these in the wrong file? Nothing about surreals here.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -324,10 +328,32 @@ instance instLE : LE Surreal :=
...
+
 instance instLT : LT Surreal :=
   ⟨lift₂ (fun x y _ _ => x < y) fun _ _ _ _ hx hy => propext (lt_congr hx hy)⟩
 #align surreal.has_lt Surreal.instLT
 
+lemma mk_lt_mk {x y : PGame.{u}} {hx hy} : mk x hx < mk y hy ↔ x < y := Iff.rfl
+
+lemma zero_lt_mk {x : PGame.{u}} {hx} : 0 < mk x hx ↔ 0 < x := Iff.rfl
+
+/-- Same as moveLeft_lt, but for `Surreal` instead of `PGame` -/
```
semorrison, line 341:
```suggestion
/-- Same as `moveLeft_lt`, but for `Surreal` instead of `PGame` -/
```

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -358,6 +384,14 @@ instance orderedAddCommGroup : OrderedAddCommGroup Surreal where
   nsmul := nsmulRec
   zsmul := zsmulRec
 
+lemma add_def {x y: PGame} (hx: x.Numeric) (hy: y.Numeric):
```
semorrison, line 387: Spaces before colons, please, throughout.

semorrison, line 387: These lemmas are all misnamed, I think. They will not actually let you rewrites `x + y` for `x y : Surreal`, which is what they lemma with this name must do. 

Suggest flipping sides and renaming `mk_add`, etc.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -358,6 +384,14 @@ instance orderedAddCommGroup : OrderedAddCommGroup Surreal where
...
+    Surreal.mk x hx + Surreal.mk y hy = Surreal.mk (x + y) (hx.add hy) := by rfl
+
+lemma sub_def {x y: PGame} (hx: x.Numeric) (hy: y.Numeric):
+    Surreal.mk x hx - Surreal.mk y hy = Surreal.mk (x - y) (hx.sub hy) := by rfl
+
+lemma zero_def : mk 0 numeric_zero = 0 := by rfl
```
semorrison, line 393: This one, if flipped, can keep its name.

Mathlib/SetTheory/Surreal/Basic.lean
```diff
@@ -324,10 +328,32 @@ instance instLE : LE Surreal :=
...
+lemma mk_lt_mk {x y : PGame.{u}} {hx hy} : mk x hx < mk y hy ↔ x < y := Iff.rfl
+
+lemma zero_lt_mk {x : PGame.{u}} {hx} : 0 < mk x hx ↔ 0 < x := Iff.rfl
+
+/-- Same as moveLeft_lt, but for `Surreal` instead of `PGame` -/
+theorem mk_moveLeft_lt_mk {x : PGame} (o : Numeric x) (i) :
+    Surreal.mk (x.moveLeft i) (by
+      rw [numeric_def] at o
+      exact o.2.1 i
+    ) < Surreal.mk x o := by
```
eric-wieser, line 346: This inner `by` is probably worth giving its own theorem. Presumably it would be `Numeric.moveLeft`?

eric-wieser: In fact, I think the theorem already exists :)

semorrison:
bors merge

PR #14601, opened 2024-07-10
[Merged by Bors] - chore: make `Algebra.cast` reducible

Command-Master:
Make `Algebra.cast` `@[reducible]`

---


eric-wieser (approved):
Nice, this looks like a good solution in the short term to not having to think about which spelling is being used.

Mathlib/Analysis/InnerProductSpace/TwoDim.lean
```diff
@@ -489,7 +489,7 @@ theorem kahler_swap (x y : E) : o.kahler x y = conj (o.kahler y x) := by
   have : ∀ r : ℝ, Complex.ofReal' r = @RCLike.ofReal ℂ _ r := fun r => rfl
   simp only [kahler_apply_apply]
   rw [real_inner_comm, areaForm_swap]
-  simp [this]
+  simp [-Complex.coe_algebraMap, this]
```
Vierkantor, line 492: Unfortunately GitHub doesn't let me make a suggestion here, but here's an alternative proof that doesn't rely on going against the `simp` direction:

```lean
theorem kahler_swap (x y : E) : o.kahler x y = conj (o.kahler y x) := by
  simp only [kahler_apply_apply]
  rw [real_inner_comm, areaForm_swap]
  simp [Complex.conj_ofReal]
```

Question for someone who knows more of the analysis side: should we make `Complex.conj_ofReal` a `simp` lemma?

Mathlib/Analysis/InnerProductSpace/TwoDim.lean
```diff
@@ -526,7 +526,7 @@ theorem kahler_comp_rightAngleRotation' (x y : E) :
 @[simp]
 theorem kahler_neg_orientation (x y : E) : (-o).kahler x y = conj (o.kahler x y) := by
   have : ∀ r : ℝ, Complex.ofReal' r = @RCLike.ofReal ℂ _ r := fun r => rfl
-  simp [kahler_apply_apply, this]
+  simp [kahler_apply_apply, this, -Complex.coe_algebraMap]
```
Vierkantor, line 529: Similarly:

```lean
theorem kahler_neg_orientation (x y : E) : (-o).kahler x y = conj (o.kahler x y) := by
  simp [kahler_apply_apply, Complex.conj_ofReal]
```

Command-Master:
!bench

Vierkantor:
Looks good to me, thank you!

bors merge

Vierkantor:
Also: good idea to run a benchmark :)

PR #14603, opened 2024-07-10
feat: degree is invariant under graph isomorphism

awueth:
---

Mathlib has the definition `SimpleGraph.Iso.mapNeighborSet` which is an equivalence between neighbor sets induced by an isomorphism. Would it be beneficial to add the same equivalence for `neighborFinset`?

### DIFF
--- a/Mathlib/Combinatorics/SimpleGraph/Maps.lean
+++ b/Mathlib/Combinatorics/SimpleGraph/Maps.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Hunter Monroe, Kyle Miller
 -/
 import Mathlib.Combinatorics.SimpleGraph.Dart
+import Mathlib.Combinatorics.SimpleGraph.Finite
 import Mathlib.Data.FunLike.Fintype
 
 /-!
@@ -670,5 +671,18 @@ noncomputable def overFinIso (hc : Fintype.card V = n) : G ≃g G.overFin hc := 
   use Fintype.equivFinOfCardEq hc; simp [overFin]
 
 end Finite
+
+section degree
+
+variable (v) [Fintype (G.neighborSet v)] (f : G ≃g G')
+
+instance : Fintype (G'.neighborSet (f v)) :=
+  Fintype.ofEquiv ↑(G.neighborSet v) (f.mapNeighborSet v)
+
+theorem degree_map_eq : G'.degree (f v) = G.degree v := by
+  apply eq_comm.1 (Finset.card_bijective f (RelIso.bijective f) _)
+  simp_rw [mem_neighborFinset, f.map_adj_iff, implies_true]
+
+end degree
 
 end SimpleGraph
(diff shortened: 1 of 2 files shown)


semorrison:
If you look at the "PR summary" comment, and click on "Import changes for modified files", you'll see that this change significantly adds to the imports of this file. Can you refactor so this result goes in a different (possibly new?) file?

And yes, to the question in the PR description!

Thanks.

awueth:
I could move it to `Mathlib.Combinatorics.SimpleGraph.Finite` where `degree` lives but then I would need to import `Mathlib.Combinatorics.SimpleGraph.Maps` there. Is this better or should I create a new file?

semorrison:
My preference is a new file, e.g. `Map/Finite`, but I am more averse to imports than most. I'm happy to merge if you move it there, but also happy if another maintainer is happy to merge as is.

jcommelin (commented):
I suggest merging this, and we can split later, in an organized fashion.

awueth:
Apologies for not answering, I don't mind making a new file and adding the equivalent theorems for finite Neighbourhoods.

semorrison:
@awueth, up to you! Please add `awaiting-author` or `WIP` if you'd like to work on this more, and otherwise we'll keep reviewing as is.

Mathlib/Combinatorics/SimpleGraph/Maps.lean
```diff
@@ -671,4 +672,17 @@ noncomputable def overFinIso (hc : Fintype.card V = n) : G ≃g G.overFin hc :=
 
 end Finite
 
+section degree
+
+variable (v) [Fintype (G.neighborSet v)] (f : G ≃g G')
+
+instance : Fintype (G'.neighborSet (f v)) :=
+  Fintype.ofEquiv ↑(G.neighborSet v) (f.mapNeighborSet v)
```
urkud, line 680: This creates a diamond with the existing instance that gives `Fintype (G.neighborSet _)` from `Fintype G` and some `Decidable*`.

Mathlib/Combinatorics/SimpleGraph/Maps.lean
```diff
@@ -671,4 +672,17 @@ noncomputable def overFinIso (hc : Fintype.card V = n) : G ≃g G.overFin hc :=
...
+
+theorem degree_map_eq : G'.degree (f v) = G.degree v := by
```
urkud, line 682: IMHO, this should assume `Fintype` for both `neighborSet`s, so that it works for any instance.

Mathlib/Combinatorics/SimpleGraph/Maps.lean
awueth: What is a diamond and why is it problematic?

Mathlib/Combinatorics/SimpleGraph/Maps.lean
awueth: Should I remove the instance on line 679 then? Doesn't adding Fintype as an assumption for both `neighborSet`s make the theorem less general?

awueth:
I moved the contents of this pull request to a new file now and added `apply_mem_neighborFinset_iff` and `mapNeighborFinset`. I could create a `NeighborFinset` counterpart for any theorem in `SimpleGraph.Maps` about `NeighborSet`, but I'm not sure whether that is useful.

PR #13841, opened 2024-06-15
[Merged by Bors] - feat:(CategoryTheory): `C ⥤ D` is enriched over `C ⥤ Type`

mckoen:
This PR shows that `C ⥤ D` is enriched over `C ⥤ Type max v' v u` where `C : Type u` with `[Category.{v} C]` and `[Category.{v'} D]`.  To do this, a hom-object is constructed for functors `F G : C ⥤ D`. Applies in particular to simplicial objects being enriched over simplicial sets.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13710, opened 2024-06-11
[Merged by Bors] - feat(CategoryTheory): `MonoidalClosed` instance on functors to `Type`

mckoen:
This PR defines an instance of `MonoidalClosed` on `C ⥤ Type max w v u` where `C : Type u` with `[Category.{v} C]`, and `w` is an arbitrary universe. An explicit internal hom functor is constructed, and is proved to be a right adjoint to `tensorLeft`.

---

- [x] depends on: #13841
<!-- The text above the `---` will become the commit message when your
PR is merged. Please leave a blank newline before the `---`, otherwise
GitHub will format the text above it as a title.

To indicate co-authors, include lines at the bottom of the commit message
(that is, before the `---`) using the following format:

Co-authored-by: Author Name <author@email.com>

Any other comments you want to keep out of the PR commit should go
below the `---`, and placed outside this HTML comment, or else they
will be invisible to reviewers.

If this PR depends on other PRs, please list them below this comment,
using the following format:
- [ ] depends on: #abc [optional extra text]
- [ ] depends on: #xyz [optional extra text]

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14608, opened 2024-07-10
feat(CategoryTheory): sheaf categories are cartesian closed

dagurtomas:
This PR contains some partial progress to more explicit witnesses to the fact that the category of functors to types is cartesian closed. I'm not actively working on this for the time being, but this code might be useful if someone decides to pick it up. 

---


Mathlib/CategoryTheory/Sites/CartesianClosed.lean
```diff
@@ -0,0 +1,35 @@
...
+
+noncomputable section
+
+open CategoryTheory Limits
+
+variable {C : Type*} [Category C] (J : GrothendieckTopology C) (A : Type*) [Category A]
+
+instance [HasWeakSheafify J A] : Reflective (sheafToPresheaf J A) where
+  map_surjective := (fullyFaithfulSheafToPresheaf _ _).map_surjective
+  map_injective := (fullyFaithfulSheafToPresheaf _ _).map_injective
```
joelriou, line 24: This instance should be moved to the file `Sites.Sheafification`.
```suggestion
```

Mathlib/CategoryTheory/Sites/CartesianClosed.lean
```diff
@@ -0,0 +1,35 @@
...
+  adj := sheafificationAdjunction _ _
+
+instance [HasSheafify J A] :  PreservesLimitsOfShape (Discrete (WalkingPair))
+    (reflector (sheafToPresheaf J A)) :=
+  inferInstanceAs (PreservesLimitsOfShape _ (presheafToSheaf _ _))
```
joelriou, line 29: This should also be moved to `Sites.Sheafification`:
```suggestion
instance [HasSheafify J A] : PreservesFiniteLimits (reflector (sheafToPresheaf J A)) :=
  inferInstanceAs (PreservesFiniteLimits (presheafToSheaf _ _))
```

Mathlib/CategoryTheory/Sites/CartesianClosed.lean
```diff
@@ -0,0 +1,35 @@
...
+
+instance [HasFiniteProducts A] : HasFiniteProducts (Cᵒᵖ ⥤ A) := ⟨inferInstance⟩
```
joelriou, line 31: This should be moved to `CategoryTheory.Limits.FunctorCategory`.

dagurtomas:
The reason these instances are in the wrong file is that I wanted to avoid changing imports. But I'll try to move them and banchmark

Mathlib/CategoryTheory/Limits/Yoneda.lean
```diff
@@ -222,4 +222,4 @@ end Functor
 
 end CategoryTheory
 
-assert_not_exists AddCommMonoid
+-- assert_not_exists AddCommMonoid
```
dagurtomas, line 225: Keeping this gave the following error:

```
Declaration AddCommMonoid is not allowed to be imported by this file.
It is defined in Mathlib.Algebra.Group.Defs,
  which is imported by Mathlib.Algebra.Group.Basic,
  which is imported by Mathlib.Algebra.Group.Embedding,
  which is imported by Mathlib.Data.Finset.Image,
  which is imported by Mathlib.Data.Fintype.Basic,
  which is imported by Mathlib.Data.Fintype.Card,
  which is imported by Mathlib.CategoryTheory.FinCategory.AsType,
  which is imported by Mathlib.CategoryTheory.Limits.Shapes.FiniteLimits,
  which is imported by Mathlib.CategoryTheory.Limits.Shapes.FiniteProducts,
  which is imported by Mathlib.CategoryTheory.Limits.FunctorCategory,
  which is imported by this file.

These invariants are maintained by `assert_not_exists` statements, and exist in order to ensure that "complicated" parts of the library are not accidentally introduced as dependencies of "simple" parts of the library.
```

I have no idea what its purpose was, but the error is clearly caused by changing imports

dagurtomas: I could revert the import change in `Limits.FunctorCategory`, and create a new file with the declarations about finite limits in functor categories?

dagurtomas: Having read the docstring of `assert_not_exists`, I think I definitely should revert this import change and create a new file instead.

Mathlib/Condensed/Light/CartesianClosed.lean
```diff
@@ -0,0 +1,33 @@
...
+
+variable {C : Type u} [SmallCategory C]
+
+instance : CartesianClosed (LightProfinite.{u}ᵒᵖ ⥤ Type u) := by
+  -- We need to consider `LightProfinite` as a small category:
+  let e : LightProfinite.{u}ᵒᵖ ⥤ Type u ≌ (SmallModel.{u, u, u+1} LightProfinite.{u})ᵒᵖ ⥤ Type u :=
+    Functor.asEquivalence ((whiskeringLeft _ _ _).obj (equivSmallModel _).op.inverse)
+  -- Now we conclude because the category of functors from a small category to a cartesian closed
+  -- category is cartesian closed.
+  exact cartesianClosedOfEquiv e.symm
```
joelriou, line 29: It seems to me this may intersect slightly with #13841 or #13710.

Also, it would be good to show the existence of suitable adjoint functors (the result ` Presheaf.isLeftAdjoint_of_preservesColimits` used in `CategoryTheory.Closed.Types`) under more general assumptions, so that it may be applied to condensed or light condensed sets.

Mathlib/CategoryTheory/Limits/Yoneda.lean
joelriou: Indeed, creating a new file it the best solution.

Mathlib/Condensed/Light/CartesianClosed.lean
dagurtomas: I wasn't aware of these two PRs. I'm not sure exactly what you mean by generalizing the result `Presheaf.isLeftAdjoint_of_preservesColimits`, but it seems to me that #13710 is proving that `C ⥤ Type max w v u` is cartesian closed, which can be obtained by generalizing what I do for `CompHaus` as follows:

```lean
instance {C : Type u} [Category.{v} C] : CartesianClosed (C ⥤ Type (max u v w)) :=
  let e : C ⥤ Type (max u v w) ≌
    (ULiftHom.{max u v w} (ULift.{max u v w} C)) ⥤ Type (max u v w) :=
      Functor.asEquivalence ((whiskeringLeft _ _ _).obj
        (ULift.equivalence.trans ULiftHom.equiv).inverse)
  cartesianClosedOfEquiv e.symm
```

It might be useful to have the more explicit approach of #13710, I don't know.

dagurtomas: Oh I see, #13710 has something to do with `ChosenFiniteProducts`, so it's a bit different, and I guess the point there is to be explicit somehow.

joelriou: You proof uses `Presheaf.isLeftAdjoint_of_preservesColimits` (though not directly). I feel it would be cleaner if a more general (universe-wise) version of this was used more directly here instead of transferring the `CartesianClosed` instances (which presumably gives no control or understanding on the right adjoint functors that are part of the data).

dagurtomas: I see your point, but I think in most cases one doesn't need any "control or understanding" of the right adjoints other than the fact that they are right adjoints. Similarly to that one doesn't usually need any definitional properties of limits/colimits other than the formal ones, etc...

But I'll see if I can generalize universes in `Presheaf.isLeftAdjoint_of_preservesColimits` at some point, I won't have that much Lean time in the coming weeks though.

Mathlib/CategoryTheory/Limits/FunctorCategoryFinite.lean
```diff
@@ -0,0 +1,26 @@
...
+
+variable {C : Type*} [Category C] {K : Type*} [Category K]
+
+instance [HasFiniteLimits C] : HasFiniteLimits (K ⥤ C) := ⟨fun _ ↦ inferInstance⟩
+
+instance [HasFiniteProducts C] : HasFiniteProducts (K ⥤ C) := ⟨inferInstance⟩
+
+instance [HasFiniteColimits C] : HasFiniteColimits (K ⥤ C) := ⟨fun _ ↦ inferInstance⟩
+
+instance [HasFiniteCoproducts C] : HasFiniteCoproducts (K ⥤ C) := ⟨inferInstance⟩
```
joelriou, line 26: Note: in a near future, this may make CI unhappy :-)
```suggestion
instance [HasFiniteCoproducts C] : HasFiniteCoproducts (K ⥤ C) := ⟨inferInstance⟩

end CategoryTheory.Limits
```

Mathlib/CategoryTheory/Sites/Sheafification.lean
```diff
@@ -72,6 +72,15 @@ def sheafificationAdjunction [HasWeakSheafify J A] :
 instance [HasWeakSheafify J A] : (presheafToSheaf J A).IsLeftAdjoint :=
   ⟨_, ⟨sheafificationAdjunction J A⟩⟩
 
+instance [HasWeakSheafify J A] : Reflective (sheafToPresheaf J A) where
+  map_surjective := (fullyFaithfulSheafToPresheaf _ _).map_surjective
+  map_injective := (fullyFaithfulSheafToPresheaf _ _).map_injective
```
joelriou, line 77:
```suggestion
```

Mathlib/CategoryTheory/Sites/Sheafification.lean
```diff
@@ -72,6 +72,15 @@ def sheafificationAdjunction [HasWeakSheafify J A] :
...
+  adj := sheafificationAdjunction _ _
+
+instance [HasSheafify J A] :  PreservesLimitsOfShape (Discrete (WalkingPair))
+    (reflector (sheafToPresheaf J A)) :=
+  inferInstanceAs (PreservesLimitsOfShape _ (presheafToSheaf _ _))
```
joelriou, line 82:
```suggestion
instance [HasSheafify J A] : PreservesFiniteLimitsOfShape (reflector (sheafToPresheaf J A)) :=
  inferInstanceAs (PreservesFiniteLimits (presheafToSheaf _ _))
```

dagurtomas:
I'm abandoning this for the time being, generalising universes turned out to be more work than expected. Another possible approach would be to define the right adjoint explicitly for presheaf categories (also started here, full of sorries)

PR #12498, opened 2024-04-28
[Merged by Bors] - feat(LinearAlgebra/TensorProduct/Submodule): add some linear maps induced by multiplication for submodules

acmepjz:
... used in the definition of linearly disjointness.

Let `R` be a commutative ring, `S` be an `R`-algebra (not necessarily commutative).
Let `M` and `N` be `R`-submodules in `S`.

- `Submodule.mulMap`: the natural `R`-linear map `M ⊗[R] N →ₗ[R] S`
  induced by the multiplication in `S`, whose image is `M * N` (`Submodule.mulMap_range`).

- `Submodule.mulMap'`: the natural map `M ⊗[R] N →ₗ[R] M * N`
  induced by multiplication in `S`, which is surjective (`Submodule.mulMap'_surjective`).

- `Submodule.lTensorOne`, `Submodule.rTensorOne`: the natural isomorphism between
  `i(R) ⊗[R] N` and `N`, resp. `M ⊗[R] i(R)` and `M`, induced by multiplication in `S`,
  here `i : R → S` is the structure map. They generalize `TensorProduct.lid`
  and `TensorProduct.rid`, as `i(R)` is not necessarily isomorphic to `R`.

  Note that we use `⊥ : Subalgebra R S` instead of `1 : Submodule R S`, since the map
  `R →ₗ[R] (1 : Submodule R S)` is not defined directly in mathlib yet.

- `Submodule.mulLeftMap`: if `m : ι → M` is a family of elements, then there is an `R`-linear map
  from `ι →₀ N` to `S` which maps `{ n_i }` to the sum of `m_i * n_i`.

- `Submodule.mulRightMap`: if `n : ι → N` is a family of elements, then there is an `R`-linear map
  from `ι →₀ M` to `S` which maps `{ m_i }` to the sum of `m_i * n_i`.

---

discussion: <https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2312498.20some.20linear.20maps.20induced.20by.20mul.2E.20for.20submodules>

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14611, opened 2024-07-10
[Merged by Bors] - feat(LinearAlgebra/TensorProduct/Subalgebra): some linear maps induced by multiplication for subalgebras

acmepjz:
This is the follow-up of the PR #12498. It is used in the definition of linearly disjointness (#9651).

---


Mathlib/LinearAlgebra/TensorProduct/Subalgebra.lean
```diff
@@ -0,0 +1,203 @@
...
+
+namespace Subalgebra
+
+variable {R : Type u} {S : Type v}
+
+section Semiring
+
+variable [CommSemiring R] [Semiring S] [Algebra R S]
+
+variable (A : Subalgebra R S)
```
eric-wieser, line 45: This generalizes to `A : Type` where `[IsScalarTower R A S]` or similar, right?

acmepjz: Maybe yes; I'll check it later.

acmepjz: I'm afraid that this is not true, for example `R=A=Z, S=F_p`. You may require that the map `A -> S` is injective. But in my opinion this is the same as `A : Subalgebra R S`. Another problem is that this isomorphism is defined in terms of `Submodule.lTensorOne`, which has no such generalizations (yet).

erdOne:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/LinearAlgebra/TensorProduct/Subalgebra.lean
```diff
@@ -0,0 +1,203 @@
...
+
+/-- If `A` is a subalgebra of `S/R`, there is the natural `R`-algebra isomorphism between
+`i(R) ⊗[R] A` and `A` induced by multiplication in `S`, here `i : R → S` is the structure map.
+This generalizes `Algebra.TensorProduct.lid` as `i(R)` is not necessarily isomorphic to `R`. -/
+def lTensorBot : (⊥ : Subalgebra R S) ⊗[R] A ≃ₐ[R] A := by
```
eric-wieser, line 50: Should this not be called `lTensorOne` to match the submodule version? That matches the pattern of naming defs after their behavior not their type.

bors r-


bors d+

acmepjz: Originally, `Submodule.lTensorOne` is about `1 : Submodule R S`. But according to discussion https://github.com/leanprover-community/mathlib4/pull/12498#discussion_r1586280439 it is changed to `⊥ : Subalgebra R S` (which as a submodule is equal to `1 : Submodule R S`) since it makes the statement and proof a little bit easier.

But for subalgebras, we don't have `1 : Subalgebra R S` but only `⊥ : Subalgebra R S`, so in my opinion it's better to talk about `lTensorBot` since it is for `⊥ : Subalgebra R S`. What's your opinion?

acmepjz: Sure, will be added in a few minutes.

PR #14617, opened 2024-07-10
[Merged by Bors] - fix: add `assumption` to autoparam `isBoundedDefault`

adomani:
Maybe, `applyc` in lean3 also tried `assumption`, so I added it here!

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/239415-metaprogramming-.2F-tactics/topic/isBoundedDefault.20not.20triggering)

---

### DIFF
--- a/Mathlib/Order/LiminfLimsup.lean
+++ b/Mathlib/Order/LiminfLimsup.lean
@@ -412,7 +412,8 @@ macro "isBoundedDefault" : tactic =>
     | apply isCobounded_le_of_bot
     | apply isCobounded_ge_of_top
     | apply isBounded_le_of_top
-    | apply isBounded_ge_of_bot)
+    | apply isBounded_ge_of_bot
+    | assumption)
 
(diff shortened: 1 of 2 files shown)


sgouezel:
Since it seems to improve the situation, let's get this in!
bors r+

D-Thomine:
I am not completely confident with the third test case.

`example : limsup u f ≤ limsup v f := by apply limsup_le_limsup h`

already worked with the previous version. What failed was

`example : limsup u f ≤ limsup v f := by apply limsup_le_limsup; exact h`

which required two more arguments.

sgouezel:
Indeed, 
```
  apply limsup_le_limsup
  exact h
```
still fails. I've already sent to bors, so the PR will be merged soon, but it doesn't solve the issue.

Note that `refine limsup_le_limsup ?_` instead of `apply limsup_le_limsup` works as a workaround.

adomani:
Yes, I agree.  Another workaround is
```
  apply (limsup_le_limsup)
```
and this reminds something that has changed recently.  Let me see if I can find it!

PR #14628, opened 2024-07-10
[Merged by Bors] - feat(to_additive): unfold lemmas generated by `simp`

fpvandoorn:
* This removes a bunch of hacky code
* Thanks to @kmill for the idea
* Also fix a bug in the implementation of `@[to_additive (attr := to_additive, x)]`
* The removed test tested for an implementation detail. The rest of the test file (and Mathlib) still passes, which shows that it still works as desired.

---


kmill (approved):
Looks good to me. The best kind of feature is one that's implemented by mostly deleting code!

For due diligence, I'll mention that I think the whole auxLemma business is about trying to avoid re-typechecking proof terms, so unfolding auxLemmas defeats this optimization, however I don't believe that simp's auxLemmas tend to be very complicated.

eric-wieser:
What's the "feat" here? Is there any visible behavior change, or is this just internal cleanup?

eric-wieser:
!bench

semorrison:
bors merge

PR #14632, opened 2024-07-10
[Merged by Bors] - chore(Analysis/Calculus/FDeriv/Extend): use Lean 4 naming scheme

dupuisf:
Most of the declarations in this file are still named according to the Lean 3 naming scheme.

---


eric-wieser (approved):
@sgouezel, can you take a final look?

sgouezel:
bors r+
Thanks!

PR #14633, opened 2024-07-10
[Merged by Bors] - chore: protect `QuadraticForm.map_zero`

eric-wieser:
---


dupuisf (commented):
Thanks!

bors r+

PR #14644, opened 2024-07-11
chore: remove `bit0` and `bit1`

jcommelin:
This PR replaces `bit0 x` resp. `bit1 x` with:
- `2 * x` resp. `2 * x + 1` if possible, or
- `2 • x` resp. `2 • x + 1` if possible, or
- `x + x` resp. `x + x + 1` as fallback.

The versions `Num.bit0` and `Num.bit1` (as well as the `PosNum` and `ZNum` versions) are left alone.
All lemmas that mention `bit0`/`bit1` have been adapted. A follow-up PR can remove these lemmas.

Deletions:
- bit0
- bit1

---

- See #14744 for a PR on top of this, that attempts to fix CI and remove many deprecated lemmas.
- See #14745 for a PR that combines these two PRs. That is probably the PR that should be merged in the end.


Mathlib/Algebra/AddTorsor.lean
```diff
@@ -467,20 +467,20 @@ theorem pointReflection_involutive (x : P) : Involutive (pointReflection x : P 
 set_option linter.deprecated false
```
Ruben-VandeVelde, line 467: I'm guessing all of those can go now

jcommelin: I would like to keep this diff as small as possible. And I also think that many of these lemmas can be removed or renamed in a second pass.

jcommelin:
Closing, as #14745 is the version that shall be merged.

PR #9163, opened 2023-12-20
[Merged by Bors] - feat(MeasureTheory/Function): stronger version of Vitali's convergence theorem

igorkhavkine:
Prove a stronger version of Vitali's convergence theorem, which does not require an `IsFiniteMeasure` hypothesis. Instead it uses a new uniform tightness hypothesis. The hypothesis `UnifTight` is implemented closely following the structure of the existing `UnifIntegrable`.

---

- [x] depends on: #14570
- [x] depends on: #14646
- [x] depends on: #14954

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14646, opened 2024-07-11
[Merged by Bors] - feat(MeasureTheory/Function): Uniform tightness of functions in Lp

igorkhavkine:
Uniform tightness is a hypothesis necessary for the general version of the Vitali Convergence Theorem proven in #9163.

---

- [x] depends on: #14853


Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,85 @@
...
+
+section UnifTight
+
+/- This follows closely the `UnifIntegrable` section
+from `MeasureTheory.Functions.UniformIntegrable`.-/
+
+variable {f g : ι → α → β} {p : ℝ≥0∞}
+
+
+/-- Definition of being Uniformly Tight.
```
YaelDillies, line 57: I don't see the point of this sentence

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,85 @@
...
+
+section VitaliConvergence
+
+variable {μ : Measure α} {p : ℝ≥0∞}
+
+variable {f : ℕ → α → β} {g : α → β}
+
+/-- **Vitali's convergence theorem** (general version).
+A sequence of functions `f` converges to `g` in Lp
+if and only if it is uniformly integrable, uniformly tight and to `g` in measure. -/
```
YaelDillies, line 77:
```suggestion
if and only if it is uniformly integrable, uniformly tight and converges to `g` in measure. -/
```

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,85 @@
...
+theorem tendstoInMeasure_iff_tendsto_Lp (hp : 1 ≤ p) (hp' : p ≠ ∞)
+    (hf : ∀ n, Memℒp (f n) p μ) (hg : Memℒp g p μ) :
+    TendstoInMeasure μ f atTop g ∧ UnifIntegrable f p μ ∧ UnifTight f p μ
+      ↔ Tendsto (fun n => snorm (f n - g) p μ) atTop (𝓝 0) :=
+  sorry
```
YaelDillies, line 82: What's this sorry for?

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,85 @@
...
+  exists some measurable set `s` with finite measure such that the Lp-norm of
+  `f i` restricted to `sᶜ` is smaller than `ε` for all `i`.
+
+# Main results
+
+* `MeasureTheory.unifTight_finite`: a finite sequence of Lp functions is uniformly
+  tight.
+* `MeasureTheory.tendsto_Lp_notFinite_of_tendsto_ae`: a sequence of Lp functions which is uniformly
+  integrable and uniformly tight converges in Lp if it converge almost everywhere.
+* `MeasureTheory.tendstoInMeasure_notFinite_iff_tendsto_Lp`: Vitali convergence theorem:
```
YaelDillies, line 31: This lemma doesn't exist

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,85 @@
...
+* `MeasureTheory.UnifTight`:
+  A sequence of functions `f` is uniformly tight in `L^p` if for all `ε > 0`, there
+  exists some measurable set `s` with finite measure such that the Lp-norm of
+  `f i` restricted to `sᶜ` is smaller than `ε` for all `i`.
+
+# Main results
+
+* `MeasureTheory.unifTight_finite`: a finite sequence of Lp functions is uniformly
+  tight.
+* `MeasureTheory.tendsto_Lp_notFinite_of_tendsto_ae`: a sequence of Lp functions which is uniformly
```
YaelDillies, line 29: Neither does this one

igorkhavkine:
Just to clarify, how self contained should this PR be (given that it should be subsumed by #9163)? I could just remove anything not directly pertaining to the UnifTight definition, including the formalization of the statement of the general Vitali Convergence Theorem (it's sorry-ed here because the proof takes up all of #9163).

YaelDillies:
It should be self-contained and only pertain to uniform tightness

igorkhavkine:
OK, I've made the PR as much self contained as I think it makes sense.

YaelDillies (commented):
Sorry, I'm being picky here, but hopefully it teaches you how to best split a PR, which will make both your and the reviewers' lives better.

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,44 @@
...
+
+/-- A sequence of functions `f` is uniformly tight in `L^p` if for all `ε > 0`, there
+exists some measurable set `s` with finite measure such that the Lp-norm of
+`f i` restricted to `sᶜ` is smaller than `ε` for all `i`. -/
+def UnifTight {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) : Prop :=
+  ∀ ⦃ε : ℝ≥0∞⦄, 0 < ε → ∃ s : Set α, μ s ≠ ∞ ∧ ∀ i, snorm (sᶜ.indicator (f i)) p μ ≤ ε
+
+end UnifTight
+
+-- main results and proofs to be filled in
```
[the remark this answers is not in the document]
igorkhavkine: Sorry, I guess I really didn't understand what you meant by self-contained. Can you clarify whether the this PR should reproduce the contents of `section UnifTight` in the original PR, or just the smaller `namespace UnifTight`?

<probably uninformed opinion>It seems to me that we are getting dangerously close to creating pull requests for the sake of creating pull requests. Whatever logical structure can be created at the level of the hierarchy of PRs, it seems to me, is already present in the original UnifTight.lean source file, at the level of code sections and namespaces.</opinion>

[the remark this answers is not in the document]
YaelDillies: The problem with #9163 as it stands is that its PR title does not in any way indicate that it's introducing a definition of uniform tightness. So either:
* You make the title of #9163 be about uniform tightness
* You split off the definition of uniform tightness (and anything that is more relevant to uniform tightness than the extension of Vitali convergence) to a separate PR (this one).

Depending on how you title the PR, having everything in one PR implies either
* that uniform tightness is mathematically used exclusively as an input to the more general version of Vitali convergence
* that the more general version of Vitali convergence is API to the definition of uniform tightness

I definitely disagree with the first, and probably also disagrees with the second.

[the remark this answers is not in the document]
YaelDillies: Pragmatically, I would expect the current PR to contain as much as possible from `MeasureTheory.Function.UnifTight` in #9163 that doesn't require changes to other files. Hopefully that represents the bulk of the file, meaning that we can get most of it merged quickly while other dependencies of #9163 are getting reviewed. Then #9163 will contain not much more than the extension of Vitali convergence.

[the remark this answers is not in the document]
igorkhavkine: OK, thanks. That clarifies what you were after. I will try to do the split later today.

> Depending on how you title the PR, having everything in one PR implies either
> 
>     * that uniform tightness is mathematically used exclusively as an input to the more general version of Vitali convergence
> 
>     * that the more general version of Vitali convergence is API to the definition of uniform tightness
> 
> I definitely disagree with the first, and probably also disagrees with the second.

While you are certainly right at the level of mathematics, arguably, at the level of Mathlib *right now* either one of these possibilities is as good as true. Perhaps I'm being naive here, but it's hard for me to see how the PR hierarchy (which AFAIK disappears into the aether as soon as all the PRs are merged) is relevant to the variation in the domain of use of tightness vs Vitali in Mathlib *in the future*. If there's some point of principle here about structuring PRs, perhaps it could be explained better to beginners. //opinion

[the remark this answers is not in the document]
YaelDillies: Well, you definitely want the very few people who care about uniform tightness or Vitali convergence to comment on your PRs and having a PR title that advertises your PR as relevant is a great first step. Note that, even if **you** think topics A and B are very closed, there will be people who care about A but not B, or B but not A. I can't count anymore the number of PRs which I should have reviewed but missed because they had a misleading title/a title about other changes that were lumped with the ones I cared about.

On a more long-termy note, commits are part of the history of Mathlib forever, so it's definitely worth spending *some* time thinking about how to present things to future git archeologists. Eg early PRs like https://github.com/leanprover-community/mathlib/pull/327 are really hard to understand and contain much less documentation than they should have. Splitting PRs really helps here.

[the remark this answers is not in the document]
igorkhavkine: > Pragmatically, I would expect the current PR to contain as much as possible from `MeasureTheory.Function.UnifTight` in #9163 that doesn't require changes to other files. Hopefully that represents the bulk of the file, meaning that we can get most of it merged quickly while other dependencies of #9163 are getting reviewed. Then #9163 will contain not much more than the extension of Vitali convergence.

Alright, probably this version of the PR is closer to what you were expecting. It still spans multiple files, due to the rearrangements that Yury had done earlier.

If you find some code/comment formatting changes that you would like to make, just go ahead and suggest them. I'm quite indifferent about them and I doubt that I would find all of them myself.

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,213 @@
...
+    (hg : Memℒp g p μ) (hfg : Tendsto (fun n => snorm (f n - g) p μ) atTop (𝓝 0)) :
+    UnifTight f p μ := by
+  have : f = (fun _ => g) + fun n => f n - g := by ext1 n; simp
+  rw [this]
+  refine UnifTight.add ?_ ?_ (fun _ => hg.aestronglyMeasurable)
+      fun n => (hf n).1.sub hg.aestronglyMeasurable
+  · exact unifTight_const hp' hg
+  · exact unifTight_of_tendsto_Lp_zero hp' (fun n => (hf n).sub hg) hfg
+
+end VitaliConvergence
```
YaelDillies, line 212: I would leave this section out until #9163

YaelDillies (commented):
Yes, thanks. This looks much better. This PR now clearly works towards a specific goal (namely defining uniform tightness and proving basic properties)

Mathlib/MeasureTheory/Integral/Lebesgue.lean
```diff
@@ -1432,6 +1432,11 @@ theorem lintegral_add_compl (f : α → ℝ≥0∞) {A : Set α} (hA : Measurabl
   rw [← lintegral_add_measure, Measure.restrict_add_restrict_compl hA]
 #align measure_theory.lintegral_add_compl MeasureTheory.lintegral_add_compl
 
+theorem setLintegral_compl {f : α → ℝ≥0∞} {s : Set α} (hsm : MeasurableSet s)
+    (hfs : ∫⁻ x in s, f x ∂μ ≠ ∞) :
+    ∫⁻ x in sᶜ, f x ∂μ = ∫⁻ x, f x ∂μ - ∫⁻ x in s, f x ∂μ := by
+  rw [← lintegral_add_compl (μ := μ) f hsm, ENNReal.add_sub_cancel_left hfs]
```
YaelDillies, line 1438: Can you add the `setIntegral` version as well?

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,181 @@
...
+from `MeasureTheory.Functions.UniformIntegrable`.-/
+
+variable {f g : ι → α → β} {p : ℝ≥0∞}
+
+
+/-- A sequence of functions `f` is uniformly tight in `L^p` if for all `ε > 0`, there
+exists some measurable set `s` with finite measure such that the Lp-norm of
+`f i` restricted to `sᶜ` is smaller than `ε` for all `i`. -/
+def UnifTight {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) : Prop :=
+  ∀ ⦃ε : ℝ≥0∞⦄, 0 < ε → ∃ s : Set α, μ s ≠ ∞ ∧ ∀ i, snorm (sᶜ.indicator (f i)) p μ ≤ ε
```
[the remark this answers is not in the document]
igorkhavkine: This was Yury's design decision (cf [this commit](https://github.com/leanprover-community/mathlib4/pull/9163/commits/d4b82684179c3fd3eeba5883fb1dfea98c240c19) and [comment](https://github.com/leanprover-community/mathlib4/pull/9163#issuecomment-1940535054)). I was originally following the analogy with `UnifIntegrable` that had `⦃ε : ℝ⦄`. Changing this will create some propagation.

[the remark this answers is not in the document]
YaelDillies: @urkud, can we have some background explanation?

Mathlib/MeasureTheory/Integral/Lebesgue.lean
igorkhavkine: Done (now in `MeasureTheory.Integral.SetIntegral`). After a cursory search, I found only [one place](https://github.com/leanprover-community/mathlib4/blob/d419680bd7d6378744b29e71e3e161a996e82178/Mathlib/Probability/Martingale/Convergence.lean#L404) in Mathlib where `setIntegral_compl` could be usefully dropped in.

Mathlib/MeasureTheory/Function/UnifTight.lean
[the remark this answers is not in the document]
urkud: I think that we should have both versions. I don't care which one is the definition and which one is a lemma. Probably, I was wrong and the `NNReal` version should be the definition.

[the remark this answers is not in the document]
YaelDillies: I think that's a good idea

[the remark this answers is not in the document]
igorkhavkine: > I think that's a good idea

I don't know. This sounds like a doubling of the API in size, for both versions of the uniform tightness condition. Is there any example in Mathlib where something like that is actually done (in an economical way, if possible)?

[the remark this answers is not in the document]
YaelDillies: Why would it double the size of the API? All it's going to add is a lemma saying that the definitions in terms of `NNReal` and `ENNReal` agree.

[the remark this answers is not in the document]
igorkhavkine: Sorry, I'm probably missing the point again. If there are multiple versions of `UnifTight` (`\eps` being `Real`, `NNReal` or `ENNReal`), then what is the version to be used in any lemma/theorem that relies on `UnifTight`? Naively, one would add all the corresponding versions, increasing the API by a factor of 2 or 3.

If no that, then one version of `UnifTight` should be primary and only that one used in the API. So which version should that be?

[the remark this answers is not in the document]
YaelDillies: When several definitions A and B are equivalent, you don't write `def A`, `def B`, `lemma _ : A \iff B` but `def A`, `lemma _ : A \iff B`.

Here the definition should be in terms of `NNReal` and you can have a lemma stating the equivalence with the `ENNReal` version.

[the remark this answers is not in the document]
igorkhavkine: OK, one last comment before the decision is made. The type of `ε` already came up already early on (cf [this comment](https://github.com/leanprover-community/mathlib4/pull/9163/files#r1493852124)). The one argument I can see in favor of `ε : ENNReal` is that in all the proofs `ε` is always compared with `ENNReal` because that is the target type of [docs#MeasureTheory.snorm](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Function/LpSeminorm/Basic.html#MeasureTheory.snorm), which makes some internal manipulations easier.

The tradeoffs are:
* `ε : Real`: mathematically natural, annoying casts in proofs
* `ε : NNReal`: still mathematically natural, still some (but less annoying) casts in proofs
* `ε : ENNReal`: always check `∞` as a special case (unnatural?), no casts in proofs

So, final decision?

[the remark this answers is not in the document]
YaelDillies: I already knew this tradeoff, so I'm happy with `NNReal` 😁

Mathlib/MeasureTheory/Function/LpSeminorm/Basic.lean
```diff
@@ -1155,6 +1167,25 @@ theorem _root_.Continuous.memℒp_top_of_hasCompactSupport
   apply memℒp_top_of_bound ?_ C (Filter.eventually_of_forall hC)
   exact (hf.stronglyMeasurable_of_hasCompactSupport h'f).aestronglyMeasurable
 
+section UnifTight
+
+/-- A single function that is `Memℒp f p μ` is tight with respect to `μ`. -/
+theorem Memℒp.exists_snorm_indicator_compl_lt {β : Type*} [NormedAddCommGroup β] (hp_top : p ≠ ∞)
+    {f : α → β} (hf : Memℒp f p μ) {ε : ℝ≥0} (hε : ε ≠ 0) :
```
YaelDillies, line 1174: Why did you change `ENNReal` to `NNReal` here? It was more general before (even if trivially)

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,218 @@
...
+variable {f g : ι → α → β} {p : ℝ≥0∞}
+
+
+/-- A sequence of functions `f` is uniformly tight in `L^p` if for all `ε > 0`, there
+exists some measurable set `s` with finite measure such that the Lp-norm of
+`f i` restricted to `sᶜ` is smaller than `ε` for all `i`. -/
+def UnifTight {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) : Prop :=
+  ∀ ⦃ε : ℝ≥0⦄, 0 < ε → ∃ s : Set α, μ s ≠ ∞ ∧ ∀ i, snorm (sᶜ.indicator (f i)) p μ ≤ ε
+
+theorem unifTight_ennreal_iff {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) :
```
YaelDillies, line 55:
```suggestion
theorem unifTight_iff_ennreal {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) :
```
looks more accurate

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,218 @@
...
+    have hε := ENNReal.toNNReal_pos heε.ne.symm heε_top
+    have hεeε := coe_toNNReal heε_top
+    obtain ⟨s, hμs, hfε⟩ := hut hε
+    use s, hμs; intro i
+    exact (hfε i).trans_eq hεeε
+  · intro hut ε hε
+    have heε := ENNReal.coe_pos.mpr hε
+    exact hut heε
+
+theorem unifTight_real_iff {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) :
```
YaelDillies, line 71: and similarly 
```suggestion
theorem unifTight_iff_real {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) :
```

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,218 @@
...
+    UnifTight g p μ := by
+  intro ε hε
+  obtain ⟨s, hμs, hfε⟩ := hf hε
+  refine ⟨s, hμs, fun n => (le_of_eq <| snorm_congr_ae ?_).trans (hfε n)⟩
+  filter_upwards [hfg n] with x hx
+  simp only [indicator, mem_compl_iff, ite_not, hx]
+
+end UnifTight
+
+/-- If two functions agree a.e., one is tight iff the other is tight. -/
```
YaelDillies, line 144:
```suggestion
/-- If two functions agree a.e., then one is tight iff the other is tight. -/
```
just for clarity

Mathlib/MeasureTheory/Function/LpSeminorm/Basic.lean
igorkhavkine: ¯\_(ツ)_/¯ I guess with that philosophy, there was no reason to change `UnifTight` to use `ENNReal` in the first place. I changed it here because it made some of the proofs I was reworking easier. Right now this doesn't contradict any convention for what type `ε` and `δ` in `Analysis` or `MeasureTheory`, since I don't see any unified convention.

YaelDillies: > I guess with that philosophy, there was no reason to change `UnifTight` to use `ENNReal` in the first place

I mean... yes, I agree? But that doesn't go against my comment above: Have easy to prove definitions, hard to prove theorems.

igorkhavkine: OK, I can understand that. Apologies for being dense, but as I'm still learning about different Mathlib philosophies and conventions it'd be nice to hear some general guidelines or explanations along with patch change suggestions. Understanding some general guidelines better I think would also help to smoothly update the parent PR.

YaelDillies: That's why I keep on commenting here 😃

There's a lot of unwritten community knowledge that you will learn by just asking. We all wish it were written down, but nobody has done that yet. So please keep on asking questions.

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,218 @@
...
+/-- A sequence of functions `f` is uniformly tight in `L^p` if for all `ε > 0`, there
+exists some measurable set `s` with finite measure such that the Lp-norm of
+`f i` restricted to `sᶜ` is smaller than `ε` for all `i`. -/
+def UnifTight {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) : Prop :=
+  ∀ ⦃ε : ℝ≥0⦄, 0 < ε → ∃ s : Set α, μ s ≠ ∞ ∧ ∀ i, snorm (sᶜ.indicator (f i)) p μ ≤ ε
+
+theorem unifTight_iff_ennreal {_ : MeasurableSpace α} (f : ι → α → β) (p : ℝ≥0∞) (μ : Measure α) :
+    UnifTight f p μ ↔ ∀ ⦃ε : ℝ≥0∞⦄, 0 < ε → ∃ s : Set α,
+      μ s ≠ ∞ ∧ ∀ i, snorm (sᶜ.indicator (f i)) p μ ≤ ε := by
+  constructor
```
YaelDillies, line 58:
```suggestion
  simp [ENNReal.forall_ennreal, UnifTight]
```

Mathlib/MeasureTheory/Function/UnifTight.lean
```diff
@@ -0,0 +1,218 @@
...
+    (hf : ∀ i, Memℒp (f i) p μ) : UnifTight f p μ := fun ε hε ↦ by
+  obtain ⟨n, hn⟩ := Finite.exists_equiv_fin ι
+  set g : Fin n → α → β := f ∘ hn.some.symm
+  have hg : ∀ i, Memℒp (g i) p μ := fun _ => hf _
+  obtain ⟨s, hμs, hfε⟩ := unifTight_fin hp_top hg hε
+  refine ⟨s, hμs, fun i => ?_⟩
+  specialize hfε (hn.some i)
+  unfold_let g at hfε
+  simp_rw [Function.comp_apply, Equiv.symm_apply_apply] at hfε
+  assumption
```
YaelDillies, line 214:
```suggestion
  simpa [g] using hfε (hn.some i)
```
or something like that

Mathlib/MeasureTheory/Integral/Lebesgue.lean
YaelDillies: Yeah, looks like a good idea to use it there. Mind opening a separate PR for `setIntegral_compl`/`setLintegral_compl` and the potential use you found?

YaelDillies:
Can you merge master to reduce the diff?

igorkhavkine:
> Can you merge master to reduce the diff?

Done. The build is marked as failed because of a linter complaint, which I'm not sure what to do about.

Also, I was undecided about choosing a proof for `unifTight_iff_real`. Please have a look.

YaelDillies:
> The build is marked as failed because of a linter complaint, which I'm not sure what to do about.

You need to run `lake exe lint_style --update`

> I was undecided about choosing a proof for `unifTight_iff_real`.

I like the second one better. Also note that you can(/should?) incorporate the `exact` into the `refine`.

igorkhavkine:
> > The build is marked as failed because of a linter complaint, which I'm not sure what to do about.
> 
> You need to run `lake exe lint_style --update`

That reproduces the same error locally, I think. Anyway the linter seems to be suggesting splitting MeasureTheory/Integral/Lebesgue.lean because it has grown too large. This patch adds about 50 lines to that file, which could potentially be moved elsewhere, but the file would still remain large. If a similar amount of code would be added to it on another occasion, the same error would come up. Perhaps it requires some higher level decisions about potentially splitting the file in a fine grained way. I don't feel qualified to decide that.

YaelDillies:
@grunweg, what are we doing wrong? Shouldn't `lake exe lint_style --update` update the lint style file?

grunweg:
> @grunweg, what are we doing wrong? Shouldn't `lake exe lint_style --update` update the lint style file?

That is implemented in #14697, which needs to be reviewed :-) Right now, the `--update` option does not quite do that yet. (I'm silently changing the behaviour of --update in that PR, on the grounds that the current behaviour is not what you want.)

YaelDillies:
@igorkhavkine, can you run `scripts/update-style-exceptions.sh`

grunweg:
That last commit was unexpected... I just pushed a fix to the style exceptions file.

YaelDillies (approved):
This looks good to me but I need a measure theorist to sign off.

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14648, opened 2024-07-11
[Merged by Bors] - chore(Data.Complex): add `@[simp]` to `Complex.conj_ofReal`

Vierkantor:
This came up in #14601: it seems this `simp` lemma is missing, and as a consequence the `simp` set wouldn't converge.

Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Should.20docs.23Complex.2Econj_ofReal.20be.20.60.40.5Bsimp.5D.60.3F

---

### DIFF
--- a/Mathlib/Data/Complex/Basic.lean
+++ b/Mathlib/Data/Complex/Basic.lean
@@ -569,6 +569,7 @@ theorem conj_im (z : ℂ) : (conj z).im = -z.im :=
   rfl
 #align complex.conj_im Complex.conj_im
 
+@[simp]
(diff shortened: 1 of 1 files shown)


j-loreaux:
Why were there no runs to compare against?

Vierkantor:
The benchmark bot isn't caught up with the master branch. I picked a random finished commit and that seems to indicate no regressions.

rwst:
!bench

j-loreaux:
bors merge

PR #14649, opened 2024-07-11
[Merged by Bors] - feat(Algebra/Homology): support of embeddings of complex shapes

joelriou:
In this PR, given embeddings of complex shapes, we define predicates for homological complexes that are supported (resp. strictly supported) on/outside this embedding. In future PRs, this shall be used in order to define cochain complexes that are cohomologically `≤` or `≥` a certain integer. 

---


semorrison:
bors merge

PR #14650, opened 2024-07-11
[Merged by Bors] - feat(Algebra/Homology): boundaries of embeddings of complex shapes

joelriou:
In this PR, we define the lower/upper boundaries of embeddings of complex shapes. These notions shall be important when constructing canonical truncations of homological complexes.

---


semorrison:
bors merge

PR #14651, opened 2024-07-11
[Merged by Bors] - refactor: give `edist_dist` a default value

fpvandoorn:
---


grunweg (commented):
Looks reasonable to me, thanks for cleaning this up! It's interesting to see how diverse the other manual proofs were. Do you think benchmarking this change is useful (very possibly not!)?

maintainer merge

fpvandoorn:
There shouldn't be any performance impact

eric-wieser (approved):
bors merge

Thanks!

PR #14654, opened 2024-07-11
feat(runLinter): allow only running certain linters

grunweg:
This adds a script `run_linter`, which is a fork of batteries' `runLinter` with extra features:
- rewrite as a CLI app, so --help messages etc. come for free
- add a `--list` option, which prints the list of linters it was about to run, and exists
- add `--select`, `--ignore` and `--only` options which ensure certain linters are always run, never resp. that only a certain list of linters are run. (`add` takes priority over `exclude` or `only`.)
- add an `update_only_remove` flag which only runs linters mentioned in the `nolints` file: this only removes entries (by design), but is often much faster.
- renamed to `run_linter`: actually, per the [zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Naming.20convention/near/450896930), this should be `run-linter` --- and one could as well use the opportunity to rename this to e.g. `env-linter`. Asked [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Configuring.2Fmodernizing.20runLinter).

Intended use cases
1. Run `update_only_remove` when updating the nolints file. In fact, it is so fast that it could even run during regular CI, after the normal `lint` step...
2. Run the mathlib linters locally. The simpNF linters are slow enough (for a reason, but still) that doing so is currently annoying. On my notebook, they can easily take half an hour, during which my notebook is otherwise hardly usable.
3. You're developing a new linter and only want to run *that* one (as there was no change about the other ones).

Probably, the nicest result would be this PR landing in batteries (perhaps in pieces) instead of here.

---

Two CLI warts I'm unhappy about.
- how can I have flag names with a hyphen? Currently, I have to write `--print_linters`, which is less pretty.
- many options take an `Array String`: how can I pass this to the executable? This is not obvious to me...


scripts/run_linter.lean
```diff
@@ -0,0 +1,167 @@
...
+unsafe def runLinter : Cmd := `[Cli|
+  lint_style VIA runLinterCli; ["0.0.1"]
+  "Runs the linters on all declarations in the given module (or `Mathlib` by default)."
+
+  FLAGS:
+    only_linters : Array String;  "Only run these named linters"
+    exclude_linters : Array String; "Do not run these named linters"
+    add_linters : Array String; "Run these linters *in addition* to the default set\n\
+      Takes priority over the --only_linters or --exclude_linters flags."
+    print_linters; "Print the list of all discovered/configured linters and exit"
```
rosborn, line 153: Remove `_linters` from these options. I would recommend using `--only`, `--ignore`, `--select`, and `--list` for these options (`include` and `exclude` are usually for files to lint or avoid linting). The priority should be on the last option provided. So `--select foo --ignore foo` would ignore, but `--ignore foo --select foo` would run `foo`.

grunweg: Thanks for the feedback! I like your renaming suggestions, I'll do that in a moment.

I'm not sure regarding the order of options matter (my current instinct is to complain about contradictory information). Do you have any examples in mind which are order-sensitive --- or cases where specifying contradictory information is actually useful?

rosborn: It is useful for when you create a shell alias `alias mylint="lake exe run_linter --only foo"` and then using the alias `mylint --select bar` will work as expected.

PR #14655, opened 2024-07-11
[Merged by Bors] - doc(Analysis/Convex): Update todo about extreme and exposed sets

YaelDillies:
These were badly outdated:
1. We now have intrinsic closure/interior/frontier.
2. There is indeed some more stuff in the `sperner_again` branch of mathlib3, which I have now ported to LeanCamCombi, but it is not so interesting.

---


semorrison:
bors merge

PR #14656, opened 2024-07-11
[Merged by Bors] - feat(CategoryTheory): Cat-valued functors induced by passing to Over/Under categories

emilyriehl:
This PR redefines but does not rename the natural isomorphisms `mapId` and `mapComp` for the composition functors associated to over or under categories. Originally these were defined using `isoRefl`. Now instead these are defined using `eqToIso` applied to equalities between the appropriate functors, which are also added here. These equalities are also used to define functors `T ⥤ Cat` and `Tᵒᵖ ⥤ Cat` sending an object `X` to `Over X` and `Under X` respectively.

---

The advantage of defining the natural isomorphisms this way is that the data is easier to use (read: easier to get rid of) in the context of the Beck-Chevalley isomorphisms using for instance coherences involving Mates. In particular, in a future PR, I plan to redefine the corresponding isomorphisms to the pullback functors as `conjugateEquiv` of the natural isomorphisms here. The reason this is not done now is those definitions are currently in the process of being moved by #14519. I thought it best to keep the two PRs independent, but if there is a better practice please let me know.

Finally, should anything here be tagged with "simp"? I still don't get how that works.

### DIFF
--- a/Mathlib/CategoryTheory/Comma/Over.lean
+++ b/Mathlib/CategoryTheory/Comma/Over.lean
@@ -182,20 +182,82 @@ theorem map_obj_hom : ((map f).obj U).hom = U.hom ≫ f :=
 theorem map_map_left : ((map f).map g).left = g.left :=
   rfl
 #align category_theory.over.map_map_left CategoryTheory.Over.map_map_left
+end
 
-variable (Y)
+section coherences
+/--
+This section proves various equalities between functors that
+demonstrate, for instance, that over categories assemble into a
+functor `mapFunctor : T ⥤ Cat`.
+
+These equalities between functors are then converted to natural
+isomorphisms using `eqToIso`. Such natural isomorphisms could be
+obtained directly using `Iso.refl` but this method will have
+better computational properties, when used, for instance, in
+developing the theory of Beck-Chevalley transformations.
+-/
 
 /-- Mapping by the identity morphism is just the identity functor. -/
-def mapId : map (𝟙 Y) ≅ 𝟭 _ :=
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -182,20 +182,75 @@ theorem map_obj_hom : ((map f).obj U).hom = U.hom ≫ f :=
...
+    simp
+
+/-- The natural isomorphism arising from `mapComp_eq`. -/
+def mapComp {X Y Z : T} (f : X ⟶ Y) (g : Y ⟶ Z) :
+    map (f ≫ g) ≅ (map f) ⋙ (map g) := eqToIso (mapComp_eq f g)
 #align category_theory.over.map_comp CategoryTheory.Over.mapComp
 
-end
+/-- The functor defined by the over categories.-/
+def mapFunctor : T ⥤ Cat where
```
semorrison, line 247:
```suggestion
@[simps] def mapFunctor : T ⥤ Cat where
```
and similarly for `Under`. This will automatically generate simp lemmas so e.g. `mapFunctor.map f` become `map f`.

semorrison, line 247: I think you need a `variable (T) in` before this declaration, otherwise `T` remains an implicit argument and it is inconvenient to specify what `T` is at the call site.

semorrison:
If you're happy making the suggested changes, you can merge afterwards. If you'd like more review, just remove the `delegated` label.

bors d+

PR #14659, opened 2024-07-11
[Merged by Bors] - feat(FieldTheory/*): prove coprimality of polynomials in terms of roots

Multramate:
---

### DIFF
--- a/Mathlib/Algebra/Polynomial/RingDivision.lean
+++ b/Mathlib/Algebra/Polynomial/RingDivision.lean
@@ -750,6 +750,13 @@ theorem comp_eq_zero_iff : p.comp q = 0 ↔ p = 0 ∨ p.eval (q.coeff 0) = 0 ∧
       Or.rec (fun h => by rw [h, zero_comp]) (fun h => by rw [h.2, comp_C, h.1, C_0]) h
 #align polynomial.comp_eq_zero_iff Polynomial.comp_eq_zero_iff
 
+lemma aeval_ne_zero_of_isCoprime [CommSemiring R] [Nontrivial S] [Semiring S] [Algebra R S]
+    {p q : R[X]} (h : IsCoprime p q) (s : S) : ¬(aeval s p = 0 ∧ aeval s q = 0) := by
+  rintro ⟨hp, hq⟩
+  rcases h with ⟨_, _, h⟩
+  apply_fun aeval s at h
+  simp only [map_add, map_mul, map_one, hp, hq, mul_zero, add_zero, zero_ne_one] at h
+
 theorem isCoprime_X_sub_C_of_isUnit_sub {R} [CommRing R] {a b : R} (h : IsUnit (a - b)) :
     IsCoprime (X - C a) (X - C b) :=
   ⟨-C h.unit⁻¹.val, C h.unit⁻¹.val, by
--- a/Mathlib/Algebra/Polynomial/RingDivision.lean
+++ b/Mathlib/Algebra/Polynomial/RingDivision.lean
@@ -751,11 +751,11 @@ theorem comp_eq_zero_iff : p.comp q = 0 ↔ p = 0 ∨ p.eval (q.coeff 0) = 0 ∧
 #align polynomial.comp_eq_zero_iff Polynomial.comp_eq_zero_iff
 
 lemma aeval_ne_zero_of_isCoprime [CommSemiring R] [Nontrivial S] [Semiring S] [Algebra R S]
-    {p q : R[X]} (h : IsCoprime p q) (s : S) : ¬(aeval s p = 0 ∧ aeval s q = 0) := by
-  rintro ⟨hp, hq⟩
+    {p q : R[X]} (h : IsCoprime p q) (s : S) : aeval s p ≠ 0 ∨ aeval s q ≠ 0 := by
+  by_contra! hpq
   rcases h with ⟨_, _, h⟩
   apply_fun aeval s at h
-  simp only [map_add, map_mul, map_one, hp, hq, mul_zero, add_zero, zero_ne_one] at h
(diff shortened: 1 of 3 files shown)


Multramate:
Note that `Polynomial.isCoprime_iff_aeval_ne_zero` and `Polynomial.isCoprime_iff_aeval_ne_zero_of_isAlgClosed` are essentially saying the same thing: two polynomials are coprime iff they have no common roots in any algebraic extension. Note also that the typeclasses `[CommRing A] [IsDomain A] [Algebra K A]` in the former can be replaced with the weaker assumptions `[Nontrivial A] [Semiring A] [Algebra F A]`, but the stronger assumption might be more usable. They should all be equivalent, but for some reason I wasn't able to prove one from the other directly, so any help is very much appreciated.

Mathlib/Algebra/Polynomial/RingDivision.lean
```diff
@@ -750,6 +750,13 @@ theorem comp_eq_zero_iff : p.comp q = 0 ↔ p = 0 ∨ p.eval (q.coeff 0) = 0 ∧
       Or.rec (fun h => by rw [h, zero_comp]) (fun h => by rw [h.2, comp_C, h.1, C_0]) h
 #align polynomial.comp_eq_zero_iff Polynomial.comp_eq_zero_iff
 
+lemma aeval_ne_zero_of_isCoprime [CommSemiring R] [Nontrivial S] [Semiring S] [Algebra R S]
+    {p q : R[X]} (h : IsCoprime p q) (s : S) : ¬(aeval s p = 0 ∧ aeval s q = 0) := by
```
eric-wieser, line 754: Stating this (and the other lemmas) with an `Or` feels more reasonable to me; not-least because this spelling isn't simp-normal.

Multramate: I'll change to that, but will keep the naming. Note that this just adds a layer of `not_and_or` everywhere.

Mathlib/Algebra/Polynomial/RingDivision.lean
```diff
@@ -750,6 +750,13 @@ theorem comp_eq_zero_iff : p.comp q = 0 ↔ p = 0 ∨ p.eval (q.coeff 0) = 0 ∧
       Or.rec (fun h => by rw [h, zero_comp]) (fun h => by rw [h.2, comp_C, h.1, C_0]) h
 #align polynomial.comp_eq_zero_iff Polynomial.comp_eq_zero_iff
 
+lemma aeval_ne_zero_of_isCoprime [CommSemiring R] [Nontrivial S] [Semiring S] [Algebra R S]
```
riccardobrasca, line 753: What about a name that allows dot notation?

Mathlib/FieldTheory/IsAlgClosed/Basic.lean
```diff
@@ -208,6 +208,17 @@ theorem IntermediateField.eq_bot_of_isAlgClosed_of_isAlgebraic {k K : Type*} [Fi
   obtain ⟨y, hy⟩ := IsAlgClosed.algebraMap_surjective_of_isIntegral (k := k) (⟨x, hx⟩ : L)
   exact ⟨y, congr_arg (algebraMap L K) hy⟩
 
+lemma Polynomial.isCoprime_iff_aeval_ne_zero_of_isAlgClosed (K : Type v) [Field K] [IsAlgClosed K]
```
riccardobrasca, line 211: Can you generalize this without assuming `K` to be algebraically closed but taking `a` in an algebraic closure? (We can also keep this version of course)

riccardobrasca (approved):

Mathlib/FieldTheory/IsAlgClosed/Basic.lean
Multramate: You mean something like this?
```
lemma Polynomial.isCoprime_iff_aeval_ne_zero_of_isAlgClosure (K : Type v) [Field K] [Algebra k K] [IsAlgClosure k K]
    (p q : k[X]) : IsCoprime p q ↔ ∀ a : K, aeval a p ≠ 0 ∨ aeval a q ≠ 0 :=
  @Polynomial.isCoprime_iff_aeval_ne_zero_of_isAlgClosed _ _ _ _ (IsAlgClosure.alg_closed k) _ _ _
```
That seems to be a specialisation rather than a generalisation?

Unless you mean the fixed `AlgebraicClosure`?

Mathlib/Algebra/Polynomial/RingDivision.lean
Multramate: Do you think `IsCoprime.aeval_ne_zero` is better than `Polynomial.aeval_ne_zero_of_isCoprime`?

Mathlib/FieldTheory/IsAlgClosed/Basic.lean
riccardobrasca: How are you avoiding that coprimality is preserved by field extensions?

Mathlib/Algebra/Polynomial/RingDivision.lean
riccardobrasca: Mmm, you're right, we also want `Polynomial`. Let's keep as it is then.

Mathlib/FieldTheory/IsAlgClosed/Basic.lean
Multramate: I'm not quite sure what you mean. We also have `Polynomial.isCoprime_iff_aeval_ne_zero` which is the version that only talks about field extensions, but I don't know how to make the proof work for algebraic closures.

riccardobrasca: I am sorry, you're lemma already says what I wanted.

PR #14660, opened 2024-07-11
[Merged by Bors] - feat: Intercalate a simplex between a point and a neighborhood

YaelDillies:
... in finite dimensions.

---

- [x] depends on: #14666
- [x] depends on: #14682
- [x] depends on: #14863

This will be used to show that convex functions are continuous.


Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -153,3 +155,51 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
+  wlog hx : x = 0
+  · obtain ⟨t, ht⟩ := this (s := -x +ᵥ s) (by simpa using vadd_mem_nhds (-x) hs) rfl
+    use x +ᵥ t
+    simpa [subset_set_vadd_iff, mem_vadd_set_iff_neg_vadd_mem, convexHull_vadd, interior_vadd]
+      using ht
+  subst hx
+  obtain ⟨b⟩ := exists_affineBasis_of_finiteDimensional
+    (ι := Fin (finrank ℝ E + 1)) (k := ℝ) (P := E) (by simp)
+  obtain ⟨ε, hε, hεs⟩ := Metric.mem_nhds_iff.1 hs
+  set u : Finset E := -Finset.univ.centroid ℝ b +ᵥ Finset.univ.image b
```
eric-wieser, line 178: I think a comment might be useful here and perhaps one other place in the proof

YaelDillies: Yes, that's exactly what's happening

YaelDillies: I've now added comments throughout

Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -153,3 +155,56 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
 #align is_connected_set_of_same_ray_and_ne_zero isConnected_setOf_sameRay_and_ne_zero
+
+end SeminormedAddCommGroup
+
+section NormedAddCommGroup
+variable [NormedAddCommGroup E] [NormedSpace ℝ E] [FiniteDimensional ℝ E] {s : Set E} {x : E}
+
+/-- We can intercalate a polyhedron between a point and one of its neighborhoods. -/
+lemma exists_mem_interior_convexHull_finset (hs : s ∈ 𝓝 x) :
+    ∃ t : Finset E, x ∈ interior (convexHull ℝ t : Set E) ∧ convexHull ℝ t ⊆ s := by
```
eric-wieser, line 166: Can you add the auxiliary lemma "we can intercalate a simplex between a point and one of its neighborhoods"?

YaelDillies: I've started working on it but it's quite a lot harder because of missing API. See #14682

Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -133,3 +135,58 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
+  -- The strategy is now to find an arbitrary maximal spanning simplex (aka an affine basis)...
+  obtain ⟨b⟩ := exists_affineBasis_of_finiteDimensional
+    (ι := Fin (finrank ℝ E + 1)) (k := ℝ) (P := E) (by simp)
+  -- ... translate it to contain the origin...
+  set c : AffineBasis (Fin (finrank ℝ E + 1)) ℝ E := -Finset.univ.centroid ℝ b +ᵥ b
+  have hc₀ : 0 ∈ interior (convexHull ℝ (range c) : Set E) := by
+    simpa [c, convexHull_vadd, interior_vadd, range_add, Pi.vadd_def, mem_vadd_set_iff_neg_vadd_mem]
+      using b.centroid_mem_interior_convexHull
+  have hcnorm :
+      range c ⊆ closedBall 0 (Finset.univ.sup' Finset.univ_nonempty (fun i ↦ ‖c i‖) + 1) := by
```
eric-wieser, line 165: Nit: add
```
let normc := Finset.univ.sup' Finset.univ_nonempty (fun i ↦ ‖c i‖)
```
or similar since you repeat this term four times and it's long.

Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -133,3 +135,58 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
+      range_subset_iff, norm_smul]
+    simpa [ε', hε₀.ne', range_subset_iff, ← mul_div_right_comm (ε / 2), div_le_iff hc',
+      mul_le_mul_left hε₀] using hcnorm
+  refine ⟨d, ?_, ?_⟩
+  · simpa [d, Pi.smul_def, range_smul, interior_smul₀, convexHull_smul, zero_mem_smul_set_iff,
+      hε'.ne']
+  calc
+    convexHull ℝ (range d) ⊆ closedBall 0 (ε / 2) := convexHull_min hdnorm (convex_closedBall ..)
+    _ ⊆ ball 0 ε := closedBall_subset_ball (by linarith)
+    _ ⊆ s := hεs
```
eric-wieser, line 190: super-nit: I'd argue for a `·` on this calc

Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -133,3 +135,58 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
+      Finset.le_sup'_iff, forall_mem_range]
+    exact fun i ↦ ⟨i, by simp⟩
+  -- ... and finally scale it to fit inside the neighborhood `s`.
+  obtain ⟨ε, hε, hεs⟩ := Metric.mem_nhds_iff.1 hs
+  set ε' : ℝ := ε / 2 / (Finset.univ.sup' Finset.univ_nonempty (fun i ↦ ‖c i‖) + 1)
+  have hc' : 0 < Finset.univ.sup' Finset.univ_nonempty (fun i ↦ ‖c i‖) + 1 := by
+    have : 0 ≤ Finset.univ.sup' Finset.univ_nonempty (fun i ↦ ‖c i‖) :=
+      Finset.le_sup'_of_le _ (Finset.mem_univ 0) (norm_nonneg _)
+    positivity
+  have hε' : 0 < ε' := by dsimp [ε']; positivity
```
eric-wieser, line 176: This works without the `dsimp`

Mathlib/Analysis/Convex/Normed.lean
```diff
@@ -133,3 +135,58 @@ theorem isConnected_setOf_sameRay_and_ne_zero {x : E} (hx : x ≠ 0) :
...
+
+end SeminormedAddCommGroup
+
+section NormedAddCommGroup
+variable [NormedAddCommGroup E] [NormedSpace ℝ E] [FiniteDimensional ℝ E] {s : Set E} {x : E}
+
+/-- We can intercalate a simplex between a point and one of its neighborhoods. -/
+lemma exists_mem_interior_convexHull_affineBasis (hs : s ∈ 𝓝 x) :
+    ∃ b : AffineBasis (Fin (finrank ℝ E + 1)) ℝ E,
+      x ∈ interior (convexHull ℝ (range b)) ∧ convexHull ℝ (range b) ⊆ s := by
```
eric-wieser, line 147: Is there some stronger statement along the lines of "and the points of the simplex lie on the `interior` or `frontier` of `s`"?

eric-wieser (approved):
bors d+

Thanks!

Mathlib/Analysis/Convex/Normed.lean
YaelDillies: I guess you are right, yes

Mathlib/Analysis/Convex/Normed.lean
YaelDillies: Because of [`interior_mem_nhds`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Basic.html#interior_mem_nhds), this lemma is easily strengthenable to `...  ∧ convexHull ℝ (range b) ⊆ interior s`. I don't think I should be making that a lemma though as the current is syntactically more general and the stronger one is easily derived

YaelDillies:
bors merge

PR #14665, opened 2024-07-12
[Merged by Bors] - chore(Logic/Equiv/Fin): Tidy tuple equivalences

YaelDillies:
The names here were all over the place and there were missing versions. Harmonise the names and add the missing versions.

---

Before I fix downstream uses, I would like to know whether people think my names are good or not. I see two minor issues with them:
* ~~What I am naming `fooEquiv` is the `Equiv` such that `fooEquiv.invFun = foo`, not `fooEquiv.toFun = foo`.~~ Fixed by turning the equivs around
* The rest of the library contains things like `MeasurableEquiv.finFoo` rather than `Fin.fooMeasurableEquiv`.

### DIFF
--- a/Mathlib/Data/Fin/Tuple/Basic.lean
+++ b/Mathlib/Data/Fin/Tuple/Basic.lean
@@ -272,7 +272,7 @@ theorem comp_tail {α : Type*} {β : Type*} (g : α → β) (q : Fin n.succ → 
 
 theorem le_cons [∀ i, Preorder (α i)] {x : α 0} {q : ∀ i, α i} {p : ∀ i : Fin n, α i.succ} :
     q ≤ cons x p ↔ q 0 ≤ x ∧ tail q ≤ p :=
-  forall_fin_succ.trans <| and_congr Iff.rfl <| forall_congr' fun j ↦ by simp [tail]
+  forall_iff_succ.trans <| and_congr Iff.rfl <| forall_congr' fun j ↦ by simp [tail]
 
 theorem cons_le [∀ i, Preorder (α i)] {x : α 0} {q : ∀ i, α i} {p : ∀ i : Fin n, α i.succ} :
     cons x p ≤ q ↔ x ≤ q 0 ∧ p ≤ tail q :=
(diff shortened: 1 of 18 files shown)


Mathlib/Data/Fin/Tuple/Basic.lean
```diff
@@ -803,9 +803,16 @@ def succAboveCases {α : Fin (n + 1) → Sort u} (i : Fin (n + 1)) (x : α i)
...
+  ⟨fun h ↦ ⟨h _, fun _ ↦ h _⟩, fun h ↦ cases h.1 h.2⟩
+
+lemma forall_iff_castSucc {P : Fin (n + 1) → Prop} :
+    (∀ i, P i) ↔ P (last n) ∧ ∀ i, P (castSucc i) :=
+  ⟨fun h ↦ ⟨h _, fun _ ↦ h _⟩, fun h ↦ lastCases h.1 h.2⟩
+
+theorem forall_iff_succAbove {P : Fin (n + 1) → Prop} (p : Fin (n + 1)) :
+    (∀ i, P i) ↔ P p ∧ ∀ i, P (p.succAbove i) :=
+  ⟨fun h ↦ ⟨h _, fun _ ↦ h _⟩, fun h ↦ succAboveCases p h.1 h.2⟩
 #align fin.forall_iff_succ_above Fin.forall_iff_succAbove
```
eric-wieser, line 816: Could you add the `Exists` versions too?

semorrison:
Could you list some examples of what you're referring to with "The rest of the library contains things like MeasurableEquiv.finFoo rather than Fin.fooMeasurableEquiv."?

YaelDillies:
I mean `MeasurableEquiv.piFinSuccAbove`

Mathlib/Data/Fin/Tuple/Basic.lean
```diff
@@ -724,9 +724,44 @@ def succAboveCases {α : Fin (n + 1) → Sort u} (i : Fin (n + 1)) (x : α i)
...
-theorem forall_iff_succAbove {p : Fin (n + 1) → Prop} (i : Fin (n + 1)) :
-    (∀ j, p j) ↔ p i ∧ ∀ j, p (i.succAbove j) :=
-  ⟨fun h ↦ ⟨h _, fun _ ↦ h _⟩, fun h ↦ succAboveCases i h.1 h.2⟩
+/-- This is a duplicate of `Fin.forall_fin_succ` in Core. We should upstream the name change. -/
+lemma forall_iff_succ {P : Fin (n + 1) → Prop} : (∀ i, P i) ↔ P 0 ∧ ∀ i, P (succ i) :=
+  ⟨fun h ↦ ⟨h _, fun _ ↦ h _⟩, fun h ↦ cases h.1 h.2⟩
+
+/-- This is a duplicate of `Fin.exists_fin_succ` in Core. We should upstream the name change. -/
+lemma exists_iff_succ {P : Fin (n + 1) → Prop} : (∃ i, P i) ↔ P 0 ∨ ∃ i, P (succ i) where
+  mp := by
```
b-mehta, line 733: I think we should give an alias here - together with the same comment - rather than re-proving the lemma. Same above.

b-mehta:
bors merge

PR #14673, opened 2024-07-12
[Merged by Bors] - feat(GroupTheory/GroupAction): Add theorem for `smul` of `toConjAct`

Louis-Le-Grand:
Add the theorem `toConjAct_smul` as discussed in #13930.

This shows that `toConjAct g • h` is equal to `g * h * g⁻¹`.

---


eric-wieser (approved):
bors merge

Thanks!

PR #14674, opened 2024-07-12
[Merged by Bors] - feat(Algebra/Group/SemiconjBy): relating SemiconjBy to orderOf

edegeltje:
this PR adds `SemiconjBy.orderOf_eq`, proving that semiconjugates have the same order.
it also adds `SemiconjBy.eq_one_iff`, which is used to prove the previous statement.

relevant zulip thread:
https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/group.20problem

Co-authored-by: Ruben Van de Velde

---

### DIFF
--- a/Mathlib/Algebra/Group/Semiconj/Basic.lean
+++ b/Mathlib/Algebra/Group/Semiconj/Basic.lean
@@ -63,5 +63,9 @@ variable [Group G] {a x y : G}
 #align semiconj_by.zpow_right SemiconjBy.zpow_right
 #align add_semiconj_by.zsmul_right AddSemiconjBy.zsmul_right
 
+variable (a) in
+@[to_additive] lemma eq_one_iff (h : SemiconjBy a x y): x = 1 ↔ y = 1 := by
+  rw [← conj_eq_one_iff (a := a) (b := x), h.eq, mul_inv_cancel_right]
+
 end Group
 end SemiconjBy
--- a/Mathlib/GroupTheory/OrderOfElement.lean
+++ b/Mathlib/GroupTheory/OrderOfElement.lean
@@ -1347,8 +1347,8 @@ lemma charP_of_prime_pow_injective (R) [Ring R] [Fintype R] (p n : ℕ) [hp : Fa
 
 namespace SemiconjBy
 
-variable [Group G] (a:G) {x y : G} (h:SemiconjBy a x y) in
-@[to_additive] lemma orderOf_eq : orderOf x = orderOf y := by
+@[to_additive]
+lemma orderOf_eq [Group G] (a : G) {x y : G} (h : SemiconjBy a x y) : orderOf x = orderOf y := by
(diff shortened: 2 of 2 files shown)


Mathlib/Algebra/Group/Semiconj/Basic.lean
```diff
@@ -63,5 +63,9 @@ variable [Group G] {a x y : G}
 #align semiconj_by.zpow_right SemiconjBy.zpow_right
 #align add_semiconj_by.zsmul_right AddSemiconjBy.zsmul_right
 
+variable (a) in
+@[to_additive] lemma eq_one_iff (h : SemiconjBy a x y): x = 1 ↔ y = 1 := by
```
urkud, line 67: Why do you want `a` to be an explicit variable here?

Mathlib/GroupTheory/OrderOfElement.lean
```diff
@@ -1344,3 +1344,13 @@ lemma charP_of_prime_pow_injective (R) [Ring R] [Fintype R] (p n : ℕ) [hp : Fa
   obtain rfl : i = n := hR i hi $ by rw [← Nat.cast_pow, CharP.cast_eq_zero]
   assumption
 #align char_p_of_prime_pow_injective charP_of_prime_pow_injective
+
+namespace SemiconjBy
+
+variable [Group G] (a:G) {x y : G} (h:SemiconjBy a x y) in
+@[to_additive] lemma orderOf_eq : orderOf x = orderOf y := by
```
urkud, line 1351:
```suggestion
@[to_additive]
lemma orderOf_eq [Group G] (a : G) {x y : G} (h : SemiconjBy a x y) : orderOf x = orderOf y := by
```

Mathlib/Algebra/Group/Semiconj/Basic.lean
edegeltje: because you can prevent the occurance of a data-valued goal, in the case of using this lemma with `rw` without `h`. It still leaves `y` unspecified, but this is at least a partial fix for that problem.

edegeltje: if you know of a better formulation to allow users to be able to prevent data-valued goals when using this lemma, please do suggest it.

urkud: E.g., Lean 4 allows `rw [SemiconjBy.eq_one_iff (a := a) (y := y)]`. If `SemiconjBy` is not available in the environment, then I think that a better style is to add `have H : SemiconjBy a x y` first.

urkud: Anyway, it's a minor issue. You can choose either option, then merge. Thanks!
bors d+

semorrison:
The delegated label didn't appear, so trying again.

bors d+

PR #14684, opened 2024-07-12
[Merged by Bors] - feat(GroupTheory/Coset): add equivalences of fibers of monoid homomorphisms

Multramate:
---

### DIFF
--- a/Mathlib/GroupTheory/Coset.lean
+++ b/Mathlib/GroupTheory/Coset.lean
@@ -813,8 +813,8 @@ namespace MonoidHom
 
 variable [Group α] {H : Type*} [Group H]
 
-/-- An equivalence between any fiber of a `MonoidHom` and its kernel. -/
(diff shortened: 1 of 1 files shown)


Mathlib/GroupTheory/Coset.lean
```diff
@@ -809,6 +809,54 @@ theorem card_comap_dvd_of_injective (K : Subgroup H) (f : α →* H)
 
 end Subgroup
 
+namespace MonoidHom
+
+variable [Group α] {H : Type*} [Group H]
+
+/-- An equivalence between any fiber of a `MonoidHom` and its kernel. -/
```
semorrison, line 816:
```suggestion
/-- An equivalence between any non-empty fiber of a `MonoidHom` and its kernel. -/
```
(I think the empty sets sitting over points not in the range count as fibers too?)

semorrison: This issue also occurs below.

semorrison:
bors d+

PR #14686, opened 2024-07-12
feat(AlgebraicGeometry/Grassmannian): define the Grassmannian scheme

smorel394:
# The Grassmannian scheme

Define the Grassmannian scheme by gluing affine charts.
We fix a commutative ring `K`, a free `K`-module of finite type `V` and two natural numbers
`r` and `c`. The scheme we define should parametrize surjective `K`-linear maps
`V →ₗ[K] (Fin r → K)`, assuming that `V` is of rank `r + c`. We actually define the scheme
without assuming the condition on the rank of `V`, but it is empty unless the rank of `V` is
`r + c`.

Main definitions:
* `Grassmannian.glueData K V r c`: the `AlgebraicGeometry.Scheme.GlueData` defining the
Grassmannian scheme.
* `Grassmannian K V r c`: the Grassmannian scheme, defined as
`(Grassmannian.glueData K V r c).glued`.
* `Grassmannian.structMorphism K V r c`: the structural morphism from `Grassmannian K V r c`
to `Spec K`.

# Implementation

We use as index type for the charts the type `Basis (Fin (r + c)) K V` (so this is empty unless
`V` is free of rank `r + c`). All the charts are the same and equal to the affine space with
coordinates indexed by `Fin c × Fin r`, that is, to `Spec (MvPolynomial (Fin c × Fin r) K)`.
The idea is that, for `i` in `Basis (Fin (r + c)) K V`, the corresponding chart will
parametrize all surjective `K`-linear maps `φ : V →ₗ[K] (Fin r → K)` that become isomorphisms
when restricted to the `K`-submodule generated by the first `r` vectors of the basis `i`.
To get the point of the chart corresponding to `φ`, we take the matrix of `φ` in the basis `i` of
`V` and the canonical basis of `Fin r → K`, we divide it on the right by its top `r × r`
square submatrix (which is invertible by assumption), and we taken the botton `c × r`
submatrix.

This is almost the usual description of the Grassmannian by charts, with three differences:
* We consider the Grassmannian parametrizing `r`-dimensional quotients of `V` instead of
`r`-dimensional subspaces of `V`, because this is more natural when working over a general
ring (or scheme).
* In the usual description, we fix a basis of `V` and index the chart by its subsets `I` of
cardinality `r`. Here, to avoid making a choice, we instead index charts by the set of
bases of `V` and always choose the subset `I` to consist of the first `r` vectors.
* Instead of working with `FiniteDimensional.finrank K V - r`, which would cause technical
trouble because of the way subtraction works on `ℕ`, we introduce the codimension `c` as an
auxiliary variable, and our constructions are only interesting when `r + c` is equal to
`FiniteDimensional.finrank K V`.

# Why is this a WIP
* There a bunch of lemmas in the file `AlgebraicGeometry/Grassmannian/Lemmas.lean` that either should not be necessary or should be put in another PR for some other part of mathlib.
* The proofs in `AlgebraicGeometry/Grassmannian/Basic.lean` are probably too complicated, and the names suck.

# Notes.

This contribution was created as part of the AIM workshop "Formalizing algebraic geometry" in
June 2024.

---

- [x] depends on: #14711


Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
```diff
@@ -0,0 +1,245 @@
...
+universe u v w
+
+section
+
+/-- If we give ourselves a scheme gluing data `GD`, and if we have morphisms from the charts
+`GD.U i` of the gluing data to a scheme `Y` that are compatible with the transition morphisms
+between charts, then we get a morphism from the glued scheme `GD.glued` to `Y`. This is an easy
+consequence of the definition of `GD.glued` as a `Multicoequalized`, but is included here for
+convenience.-/
+def AlgebraicGeometry.Scheme.GlueData.glueMorphisms (GD : Scheme.GlueData)
```
erdOne, line 39: The intended way is to use [`AlgebraicGeometry.Scheme.GlueData.openCover`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Gluing.html#AlgebraicGeometry.Scheme.GlueData.openCover) and [`AlgebraicGeometry.Scheme.OpenCover.glueMorphisms`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Gluing.html#AlgebraicGeometry.Scheme.OpenCover.glueMorphisms). Do they cause any inconvenience?

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
```diff
@@ -0,0 +1,245 @@
...
+
+end
+
+section
+
+open Function
+
+variable {X Y : Scheme}
+
+lemma SchemeIsoToBijective (f : X ≅ Y) : Bijective f.hom.val.base := by
```
erdOne, line 78: The current way is
`example (f : X ⟶ Y) [IsIso f] : Bijective f.val.base := f.homeomorph.bijective`
Not sure if we should provide this as a separate lemma.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
```diff
@@ -0,0 +1,245 @@
...
+
+variable (R S T U : Type u) [CommRing R] [CommRing S] [CommRing T] [CommRing U]
+  [Algebra R S] [Algebra R T] [Algebra R U] [Algebra S U] [IsScalarTower R S U] [Algebra T U]
+  [IsScalarTower R T U]
+  (f g : R)
+
+/-- If `S` is an `R`-algebra, this is the morphism of schemes from `Spec S` to `Spec R` defined
+by the structural map from `R` to `S`.-/
+noncomputable
+nonrec abbrev Spec.algebraMap : Spec (.of S) ⟶ Spec (.of R) :=
```
erdOne, line 97: Do you think this is useful elsewhere?

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
```diff
@@ -0,0 +1,245 @@
...
+variable {X Y : Scheme}
+
+lemma SchemeIsoToBijective (f : X ≅ Y) : Bijective f.hom.val.base := by
+  change Bijective ((Scheme.forgetToTop ⋙ forget TopCat).mapIso f).toEquiv
+  exact Equiv.bijective _
+
+end
+
+section
+-- The code in this section was written by Andrew Yang.
```
erdOne, line 85: A better way to do this is to allow [`CommRingCat.pushoutCocone`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Category/Ring/Constructions.html#CommRingCat.pushoutCocone) to accept `Algebra` instances. I'll try to open a PR later today.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
smorel394: I don't know. It was in your code so I kept it, but I haven't used it myself.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
smorel394: Probably not, I just didn't know the way you mentioned.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
smorel394: Yes. When you construct a scheme by gluing and want to define a morphism knowing its "restrictions" to the charts, the gluing conditions that will naturally appear on these restrictions are the the ones given as condition `hf` in my lemma `AlgebraicGeometry.Scheme.GlueData.glueMorphisms`. I tried to rewrite the lemma's proof in the following way:
```
def AlgebraicGeometry.Scheme.GlueData.glueMorphisms (GD : Scheme.GlueData)
    {Y : Scheme} (f : (i : GD.J) → GD.U i ⟶ Y) (hf : ∀ (i j : GD.J),
    GD.f i j ≫ (f i) = GD.t i j ≫ GD.f j i ≫ (f j)) :
    GD.glued ⟶ Y := by
  refine Scheme.OpenCover.glueMorphisms GD.openCover f ?_
  intro i j
  simp? [Scheme.GlueData.openCover, hf]
```
but this does not close the goal. This means that I have to do something nontrivial to prove the condition that the lemma `Scheme.OpenCover.glueMorphisms` expects from condition `hf`. I only tried for a few minutes, but I could not figure it yet.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
smorel394: Concerning this lemma, I also realized that I wasn't using it anymore in the Grassmannian code! So I removed it.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
erdOne: I see. I agree it is useful.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
erdOne: #14711

smorel394: Thanks!

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
smorel394: This will disappear once your PR on the pullback of affine schemes is merged, so I'll mark the conversation as resolved.

Mathlib/AlgebraicGeometry/Grassmannian/Lemmas.lean
```diff
@@ -0,0 +1,154 @@
...
+    {Y : Scheme} (f : (i : GD.J) → GD.U i ⟶ Y) (hf : ∀ (i j : GD.J),
+    GD.f i j ≫ (f i) = GD.t i j ≫ GD.f j i ≫ (f j)) :
+    GD.glued ⟶ Y := by
+  refine Limits.Multicoequalizer.desc _ Y f ?_
+  simp only [GlueData.diagram_l, GlueData.diagram_left, GlueData.diagram_right, Prod.forall,
+    GlueData.diagram_fstFrom, GlueData.diagram_fst, GlueData.diagram_sndFrom, GlueData.diagram_snd,
+    Category.assoc]
+  exact hf
+
+@[simp, reassoc]
```
joelriou, line 48: This way, the `_assoc` version is also a simp lemma:
```suggestion
@[reassoc (attr := simp)]
```

joneugster:

PR #14689, opened 2024-07-12
[Merged by Bors] - docs: fix references.bib

acmepjz:
This PR removes `\url` and `\_` commands in `references.bib`. The existence of them breaks https://github.com/leanprover/doc-gen4/pull/209.

- The first is `howpublished = {\url{...}}`. I think it's better to use the standard `url` field. Meanwhile `howpublished` field is preserved since some bibtex parser complains about missing that field.
- There are some `\_` in `url` and `doi` fields. I think it's better to replace them with `_`; there are already some `_` in `url` and `doi` fields of other bibitems.
- There is `$K\_X$`, but which, after checking the original article, should be `$K_X$`.

---

### DIFF
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -144,7 +144,8 @@
   author        = {Jeremy Avigad and Leonardo de Moura and Soonho Kong},
   title         = {{T}heorem {P}roving in {L}ean},
   year          = {2017},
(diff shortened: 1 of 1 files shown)


bryangingechen:
Thanks!
bors r+

PR #14692, opened 2024-07-13
[Merged by Bors] - docs(Data/Finset/Basic): fix docstring referring to other files

acmepjz:
This PR fixes docstring referring to other files. Originally they were `Mathlib.XXX.XXX`, but according to doc-gen4, to make them work correctly, they should be `Mathlib/XXX/XXX.lean`.

---


bryangingechen:
Thanks!
bors r+

PR #14698, opened 2024-07-13
chore(Integral/Lebesgue): move lemmas about σ-finite measures

urkud:
... to a new file

---
I'm looking for better split points, but we can already move these lemmas.


Mathlib/MeasureTheory/Integral/Lebesgue/SigmaFinite.lean
```diff
@@ -0,0 +1,201 @@
...
+/-- If the Lebesgue integral of a function is bounded by some constant on all sets with finite
+measure in a sub-σ-algebra and the measure is σ-finite on that sub-σ-algebra, then the integral
+over the whole space is bounded by that same constant. Version for a measurable function.
+See `lintegral_le_of_forall_fin_meas_le'` for the more general `AEMeasurable` version. -/
+theorem lintegral_le_of_forall_fin_meas_le_of_measurable (hm : m ≤ m0) [SigmaFinite (μ.trim hm)]
+    (C : ℝ≥0∞) {f : α → ℝ≥0∞} (hf_meas : Measurable f)
+    (hf : ∀ s, MeasurableSet[m] s → μ s ≠ ∞ → ∫⁻ x in s, f x ∂μ ≤ C) : ∫⁻ x, f x ∂μ ≤ C := by
+  have : ∫⁻ x in univ, f x ∂μ = ∫⁻ x, f x ∂μ := by simp only [Measure.restrict_univ]
+  rw [← this]
+  refine univ_le_of_forall_fin_meas_le hm C hf fun S hS_meas hS_mono => ?_
```
urkud, line 73: The rest of the proof is about `∫⁻ x in ⋃ n, S n, f x = ⨆ n, ∫⁻ x in S n, f x`. I'm going to move it to a separate lemma, then this file becomes much shorter, so I'm going to look for another way to split the original file.

PR #14699, opened 2024-07-13
[Merged by Bors] - feat(RingTheory/Derivation): define `Differential` and `DifferentialAlgebra` typeclasses

Command-Master:
Define a typeclass `Differential` of `CommRing`s with a chosen `Derivation`, a notation for that derivation, and a `DifferentialAlgebra` for `Algerba`s on differential rings whose `algebraMap` commutes with the derivation.

---


Command-Master:
Is there a problem with the instance I add?

erdOne:
Do you plan to add more results about this? Please move it into a new file in that case.

Command-Master:
> Do you plan to add more results about this? Please move it into a new file in that case.

A few more, proving that `AlgHom`s commute with the derivative in some cases. What's a good name for this file?

erdOne:
`RingTheory/Derivation/DifferentialRing` or `RingTheory/DifferentialRing` both makes sense IMO.

Mathlib/Topology/ContinuousFunction/UniqueCFC.lean
```diff
@@ -96,6 +96,7 @@ variable {A : Type*} [TopologicalSpace A] [Ring A] [StarRing A] [Algebra ℝ A]
 
 namespace StarAlgHom
 
+set_option synthInstance.maxHeartbeats 21000 in
```
erdOne, line 99: Is this still necessary?

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,83 @@
...
+differential algebras, which are algebras of differential rings where the algebra map commutes
+with the derivative.
+
+This file also defines the x′ notation for the derivative of x in a differential ring.
+-/
+
+/--
+A `CommRing` with a bundled `Derivation` to itself.
+-/
+class CommDifferentialRing (R : Type*) extends CommRing R where
```
erdOne, line 21: Are there any non-commutative `CommDifferentialRing`s?

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,83 @@
...
+  /-- The `Derivation` assosiated with the ring. -/
+  deriv : Derivation ℤ R R
+
+@[inherit_doc]
+postfix:max "′" => CommDifferentialRing.deriv
```
erdOne, line 26:
```suggestion
scoped[DifferentialRing] postfix:max "′" => CommDifferentialRing.deriv

open scoped DifferentialRing
```

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,83 @@
...
+
+export DifferentialAlgebra (deriv_algebraMap)
+
+instance (A : Type*) [CommDifferentialRing A] : DifferentialAlgebra A A where
+  deriv_algebraMap _ := rfl
+
+/--
+Transfer a `CommDifferentialRing` instance accross a `RingEquiv`.
+-/
+def CommDifferentialRing.equiv {R R2 : Type*} [CommRing R] [CommDifferentialRing R2] (h : R ≃+* R2):
```
erdOne, line 56:
```suggestion
@[reducible]
def CommDifferentialRing.equiv {R R2 : Type*} [CommRing R] [CommDifferentialRing R2] (h : R ≃+* R2):
```
see library note [reducible non-instances].

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,83 @@
...
+  deriv := Derivation.mk' (h.symm.toAddMonoidHom.toIntLinearMap ∘ₗ
+    CommDifferentialRing.deriv.toLinearMap ∘ₗ h.toAddMonoidHom.toIntLinearMap) (by simp)
+
+@[norm_cast]
+lemma algebraMap.coe_deriv {A : Type*} {B : Type*} [CommDifferentialRing A] [CommDifferentialRing B]
+    [DifferentialAlgebra A B] (a : A) :
+    (a′ : A) = (a : B)′ :=
+  (DifferentialAlgebra.deriv_algebraMap _).symm
+
+
```
erdOne, line 67:
```suggestion
```

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,83 @@
...
+/--
+Transfer a `DifferentialAlgebra` instance accross a `AlgEquiv`.
+-/
+def DifferentialAlgebra.equiv {A : Type*} [CommDifferentialRing A]
```
erdOne, line 71:
```suggestion
@[reducible]
def DifferentialAlgebra.equiv {A : Type*} [CommDifferentialRing A]
```

Mathlib/RingTheory/Derivation/DifferentialRing.lean
Command-Master: I'm not aware of any, the existing definition of Derivation requires commutativity, and Wikipedia defines it with commutativity, but I don't know differential algebra well enough to tell for sure.

Mathlib/Topology/ContinuousFunction/UniqueCFC.lean
Command-Master: I tried removing it, let's see if it works

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,103 @@
...
+A differential algebra is an `Algebra` where the derivation commutes with `algebraMap`.
+-/
+class DifferentialAlgebra (A B : Type*) [CommDifferentialRing A] [CommDifferentialRing B]
+    extends Algebra A B where
+  deriv_algebraMap : ∀ a : A, (algebraMap A B a)′ = algebraMap A B a′
+
+export DifferentialAlgebra (deriv_algebraMap)
+
+/--
+A differential ring `A` and an algebra over it `B` share constants is all
```
erdOne, line 53:
```suggestion
A differential ring `A` and an algebra over it `B` share constants if all
```
Also I would suggest the following instead
```
class CommDifferentialRing.ContainConstants (A B : Type*) [CommRing A] [CommDifferentialRing B]
    [Algebra A B] : Prop where
  /-- If the derivative of x is 0, then it's in the range of `algberaMap A B`. -/
  protected mem_range_of_deriv_eq_zero {x : B} (h : x′ = 0) : x ∈ (algebraMap A B).range
```

Mathlib/RingTheory/Derivation/DifferentialRing.lean
erdOne: Would you ever want to consider multiple differential-ring structures on the same ring (maybe multiple partial derivatives)? If so, I would suggest `class DifferentialRing (R : Type*) [CommRing R] where` instead.

Command-Master:
How would `class DifferentialRing (R : Type*) [CommRing R] where` help for having multiple derivatives?

Mathlib/RingTheory/Derivation/DifferentialRing.lean
erdOne: You could have multiple `(d1 d2 : DifferentialRing R)` lying around, like how mathlib deals with multiple topologies on a type. Also you wouldn't be able to declare differential fields with this bundled approach.

Command-Master: I would (and do, in #14860), they just extend both `CommDifferentialRing` and `Field`

Command-Master: I tried to change it, what do you think? It's nicer in some ways, but it also requires writing more typeclass assumptions

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,99 @@
...
+
+instance (A : Type*) [CommRing A] [Differential A] : Differential.ContainConstants A A where
+  mem_range_of_deriv_eq_zero {x} _ := ⟨x, rfl⟩
+
+/--
+Transfer a `Differential` instance accross a `RingEquiv`.
+-/
+@[reducible]
+def Differential.equiv {R R2 : Type*} [CommRing R] [CommRing R2] [Differential R2]
+    (h : R ≃+* R2) : Differential R :=
```
jcommelin, line 80: Could you please turn the `2` into a subscript? Type `\2` in VSCode to get the unicode subscript version.

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,99 @@
...
+
+instance (A : Type*) [CommRing A] [Differential A] : DifferentialAlgebra A A where
+  deriv_algebraMap _ := rfl
+
+instance (A : Type*) [CommRing A] [Differential A] : Differential.ContainConstants A A where
+  mem_range_of_deriv_eq_zero {x} _ := ⟨x, rfl⟩
+
+/--
+Transfer a `Differential` instance accross a `RingEquiv`.
+-/
```
jcommelin, line 77:
```suggestion
/-- Transfer a `Differential` instance accross a `RingEquiv`. -/
```

Mathlib/RingTheory/Derivation/DifferentialRing.lean
```diff
@@ -0,0 +1,99 @@
...
+/-!
+# Differential and Algebras
+
+This file defines derivations from a commutative ring to itself as a typeclass, which lets us
+use the x′ notation for the derivative of x.
+-/
+
+/--
+A derivation from a ring to itself, as a typeclass.
+-/
```
jcommelin, line 17:
```suggestion
/-- A derivation from a ring to itself, as a typeclass. -/
```

jcommelin (commented):
My apologies for the delay in reviewing. Holiday season etc...

I think this looks good now.

bors merge

PR #14701, opened 2024-07-13
[Merged by Bors] - feat(Algebra/Polynomial/Basic): transfer some `Finsupp` smul lemmas to polynomials

Command-Master:
Add `sum_smul_index'` and `smul_sum` for polynomials.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14702, opened 2024-07-13
[Merged by Bors] - feat(Algebra/Polynomial/Derivation): coefficient-wise derivation on a polynomial

Command-Master:
Define `Derivation.coeffwise`, which applies the derivation to the a polynomial coefficient-wise.

---

- [x] depends on: #14701


erdOne (commented):
```lean
/--
The `R`-derivation from `A[X]` to `M[X]` which applies the derivative to each
of the coefficients.
-/
def coeffwise : Derivation R A[X] (PolynomialModule A M) where
  __ := (PolynomialModule.map A d.toLinearMap).comp
    PolynomialModule.equivPolynomial.symm.toLinearMap
  map_one_eq_zero' := show (Finsupp.single 0 1).mapRange (d : A → M) d.map_zero = 0 by simp
  leibniz' p q := by
    dsimp 
    induction p using Polynomial.induction_on' with
    | h_add => simp only [add_mul, map_add, add_smul, smul_add, add_add_add_comm, *]
    | h_monomial n a =>
      induction q using Polynomial.induction_on' with
      | h_add => simp only [mul_add, map_add, add_smul, smul_add, add_add_add_comm, *]
      | h_monomial m b =>
        refine Finsupp.ext fun i ↦ ?_
        dsimp [PolynomialModule.equivPolynomial, PolynomialModule.map]
        simp only [toFinsupp_mul, toFinsupp_monomial, AddMonoidAlgebra.single_mul_single]
        show d _ = _ + _
        erw [Finsupp.mapRange.linearMap_apply, Finsupp.mapRange.linearMap_apply]
        rw [Finsupp.mapRange_single, Finsupp.mapRange_single]
        erw [PolynomialModule.monomial_smul_single, PolynomialModule.monomial_smul_single]
        simp only [AddMonoidAlgebra.single_apply, apply_ite d, leibniz, map_zero, coeFn_coe,
          PolynomialModule.single_apply, ite_add_zero, add_comm m n]

@[simp]
lemma coeffwise_apply (p : A[X]) (i) :
    d.coeffwise p i = d (coeff p i) := rfl

@[simp]
lemma coeffwise_monomial (n) (x : A) :
    d.coeffwise (monomial n x) = .single A n (d x) := Finsupp.ext fun _ ↦ by
  simp [coeff_monomial, apply_ite d, PolynomialModule.single_apply]

@[simp]
lemma coeffwise_X :
    d.coeffwise (X : A[X]) = 0 := Finsupp.ext fun _ ↦ by
  simp [← monomial_one_one_eq_X]

theorem apply_eval_eq (x : A) (p : A[X]) :
    d (eval x p) = PolynomialModule.eval x (d.coeffwise p) + eval x (derivative p) • d x := by
  induction p using Polynomial.induction_on' with
  | h_add => simp_all only [eval_add, map_add, add_smul]; abel
  | h_monomial n a =>
    simp only [eval_monomial, leibniz, leibniz_pow, coeffwise_monomial,
      PolynomialModule.eval_single, derivative_monomial, ← mul_smul, nsmul_eq_smul_cast A]
    rw [add_comm]
    congr 2
    ring
```
This seems like a better approach. Note that you would need to relax the typeclass arguments on `PolynomialModule.map` to make this work.
By the way it would be nice if you can identify the missing lemmas and remove those `erw`s.

Command-Master:
How could the typeclass parameters of `PolynomialModule.map` be relaxed? Its domain is `PolynomialModule R M`, which requires `CommRing R`. You'd have to relax all of `PolynomialModule`.

erdOne:
You're right. Then that's the thing to do if you really care about semirings.

Command-Master:
I added the lemmas. I'm not sure about changing the definition - it makes `coeffwise_apply` definitional, and is a bit cleaner, but loses `R` being a semiring.

erdOne:
I don't think that matters much, unless you absolutely need semirings. Someone else who needs semirings may generalise `PolynomialModule` altogether.

Mathlib/Algebra/Polynomial/Derivation.lean
```diff
@@ -5,6 +5,7 @@ Authors: Kevin Buzzard, Richard Hill
 -/
 import Mathlib.Algebra.Polynomial.AlgebraMap
 import Mathlib.Algebra.Polynomial.Derivative
+import Mathlib.Algebra.Polynomial.Module.Basic
```
semorrison, line 8: I think you should create a new file for this material. One shouldn't need to know about `PolynomialModule` just to talk about the basic material here.

Mathlib/Algebra/Polynomial/Derivation.lean
```diff
@@ -148,4 +151,87 @@ theorem comp_aeval_eq (d : Derivation R A M) (f : R[X]) :
...
+section coeffwise
+
+variable {R A M : Type*} [CommRing R] [CommRing A] [Algebra R A] [AddCommGroup M]
+  [Module A M] [Module R M] [IsScalarTower R A M] (d : Derivation R A M) (a : A)
+
+/--
+The `R`-derivation from `A[X]` to `M[X]` which applies the derivative to each
+of the coefficients.
+-/
+def coeffwise : Derivation R A[X] (PolynomialModule A M) where
```
semorrison, line 165: Could we have a more standard make?

Either `mapCoeffs` or `Derivation.polynomialModule`?

Command-Master: Renamed to `mapCoeffs`.

Mathlib/Algebra/Polynomial/Derivation.lean
erdOne: I don't really understand this concern. `PolynomialModule` is a basic file and the imports of that file are already imported by `Polynomial/Derivation`. Anyways splitting is still nice but I think it's confusing to have `Polynomial/Derivation` and `Derivation/Polynomial`. Can you give it a better name?

Command-Master: Is there anything you suggest?

erdOne: Not sure. Maybe `Polynomial/Derivation/MapCoeffs` and the original file to `Polynomial/Derivation/Basic`

Command-Master: I'm not sure if `Polynomial/Derivation` should be a directory, is there anything else which could be placed there?

Command-Master: Perhaps the original file could be moved to `RingTheory/Derivation/Polynomial` and this will be `RingTheory/Derivation/MapCoeffs`?

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -0,0 +1,98 @@
...
+
+@[simp]
+lemma mapCoeffs_X :
+    d.mapCoeffs (X : A[X]) = 0 := by simp [← monomial_one_one_eq_X]
+
+@[simp]
+lemma mapCoeffs_C (x : A) :
+    d.mapCoeffs (C x) = .single A 0 (d x) := by simp [← monomial_zero_left]
+
+variable {K M' : Type*} [CommRing K] [Algebra R K] [Algebra A K]
```
jcommelin, line 67: Could you please rename `K` to `B`? To me, `K` suggests that it is a field.

Note that, in VSCode, if you place your cursor on `K` and hit <F2> then you can rename it once, and it will update all other occurences of `K` for you.

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -0,0 +1,98 @@
...
+    [AddCommGroup M'] [Module K M'] [Module R M'] [Module A M']
+
+theorem apply_aeval_eq' (d2 : Derivation R K M') (f : M →ₗ[A] M')
```
jcommelin, line 70: Would it make sense to rename `d2` to `d'`, in analogy with `M` and `M'`?

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -0,0 +1,98 @@
...
+namespace Derivation
+
+variable {R A M : Type*} [CommRing R] [CommRing A] [Algebra R A] [AddCommGroup M]
+  [Module A M] [Module R M] [IsScalarTower R A M] (d : Derivation R A M) (a : A)
+
+/--
+The `R`-derivation from `A[X]` to `M[X]` which applies the derivative to each
+of the coefficients.
+-/
+def mapCoeffs : Derivation R A[X] (PolynomialModule A M) where
```
YaelDillies, line 28: This name is ambiguous in that one could also derive the a multivariate polynomial coefficient-wise. What about something like this?
```suggestion
def polynomialCoeffs : Derivation R A[X] (PolynomialModule A M) where
```

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -0,0 +1,98 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Daniel Weber
+-/
+import Mathlib.RingTheory.Derivation.Basic
+import Mathlib.Algebra.Polynomial.Module.Basic
+
+/-!
+# Derivations of univariate polynomials
+
+In this file we prove properties about `R`-derivations of `Polynomial A`.
```
YaelDillies, line 12: I feel like this is a bit of a blank statement. Eg you are not proving anything about the obvious derivation on polynomials which sends `X ^ n` to `n • X ^ (n - 1)`. Can you expand?

Mathlib/RingTheory/Derivation/MapCoeffs.lean
Command-Master: Should I also change the file name, or is it ok?

Command-Master: Thinking about it a bit more, I don't love `polynomialCoeffs` - it feels weird to write `D.polynomialCoeffs p`

YaelDillies: I would personally just call it `Derivation.Polynomial`, but I don't know 🤷

Command-Master: Perhaps just `Derivation.poly`? Or is that too confusing?

YaelDillies: In absolute, it might not be confusing. But here it would be confusing to the extent that existing files about `Polynomial` have `Polynomial` in their name, not just `Poly`.

Command-Master: Perhaps `Polynomial.differentiateCoeffs`? Also that's still quite long

Command-Master: This was originally in the same file as that derivation (`Algebra/Polynomial/Derivation`), but I moved it to this file

YaelDillies: Oh I see. I guess it can go back there

Command-Master: @semorrison @erdOne could we get your opinions on this?

Command-Master: In #14967 I also import `RingTheory.Derivation.DifferentialRing` to this file, so I think it's good to keep as a separate file

Command-Master: How bad do you think this name is? I'm unable to think of any alternative I like

YaelDillies: I guess it's fine, but let's not resolve this conversation in case someone thinks of a better name

YaelDillies (approved):
maintainer merge

riccardobrasca:
Thanks!

bors merge

Command-Master:
bors merge

Ruben-VandeVelde:
maintainer merge

riccardobrasca:
bors merge

PR #14704, opened 2024-07-13
feat(FieldTheory): define typeclass for simple extensions, and prove some properties

Command-Master:
Define `SimpleExtension F K`, which says that `K` is a simple field extension of `F`, and show that together with transcendentality it implies `IsFractionRing F[X] K`.

---

- [x] depends on: #14710 


Mathlib/FieldTheory/SimpleExtension.lean
```diff
@@ -0,0 +1,120 @@
...
+open Polynomial IntermediateField algebraMap
+
+/--
+A typeclass for K being a simple extension of F, F⟮k⟯ = K.
+-/
+class SimpleExtension (F K : Type*) [CommSemiring F] [Field K] [Algebra F K] where
+  /-- The element of K which is adjoined to F. -/
+  k : K
+  /-- Every value in K is a fraction of two polynomials with coefficients in F, evaluated at k. -/
+  is_adjoin' : ∀ (x : K), ∃ (r s : Polynomial F), x = (aeval k) r / (aeval k) s
```
acmepjz, line 29: I think `class IsSimpleExtension ... : Prop` is better? Your current definition is a `Type` which carries a data `k` which is the generator of `K` over `F`. The disadvantage is that different `k` gives different instance `SimpleExtension F K`, which in my opinion complicates things.

Mathlib/FieldTheory/SimpleExtension.lean
```diff
@@ -0,0 +1,120 @@
...
+section CommRing
+
+variable (F K : Type*) [CommRing F] [Nontrivial F] [IsDomain F] [Field K] [Algebra F K]
+    [se : SimpleExtension F K]
+
+/--
+A typeclass for K being a transcendental simple extension of F.
+-/
+class TranscendentalExtension : Prop where
+  is_transcendental : Transcendental F se.k
```
acmepjz, line 64: I think your definition means purely transcendental of transcendental degree one, right? But from your name `TranscendentalExtension` neither purely transcendental nor of transcendental degree one can be inferred, which may be misleading.

Mathlib/FieldTheory/SimpleExtension.lean
```diff
@@ -0,0 +1,120 @@
+/-
+Copyright (c) 2024 Daniel Weber. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Daniel Weber
+-/
+import Mathlib.FieldTheory.PrimitiveElement
+import Mathlib.RingTheory.Henselian
+import Mathlib.Algebra.Lie.OfAssociative
```
acmepjz, line 8: Why are these two imports needed in that file? It seems that simple extension does not use Henselian or Lie algebra.

Command-Master: I don't know, that's what #min_imports said, but indeed shake complains about it.

acmepjz: You can try to remove these two imports and see if it still compiles. (The answer is yes.)

Mathlib/FieldTheory/SimpleExtension.lean
Command-Master: My use for this `SimpleExtension` is for Liouville's theorem, where I extend this with the existence of a `v` in `F` such that `k' / k = v'` (an exponential extension) or `k' = v' / v` (a logarithmic extension). Would there be a good way to do that with `IsSimpleExtension`?

Mathlib/FieldTheory/SimpleExtension.lean
Command-Master: Maybe it's better to just create `Algebra.IsTranscendental` as not `Algebra.IsAlgebraic` which implies this?

Mathlib/FieldTheory/SimpleExtension.lean
acmepjz: I'm not sure. But from your code `Algebra F[X] K` seems that the generator `k` needs to be gives, otherwise this instance cannot be defined. So maybe your `SimpleExtension` with data is better...

I think you should clearly state in the docstring that `SimpleExtension` is with a generator given.

Mathlib/FieldTheory/SimpleExtension.lean
acmepjz: I mean maybe the name could be changed to `PurelyTranscendentalOfDegOne`

Mathlib/FieldTheory/SimpleExtension.lean
Command-Master: I edited the docstring, is it better now?

Mathlib/FieldTheory/SimpleExtension.lean
Command-Master: I removed this typeclass and replaced this assumptions with `Algebra.Transcendental`, introduced in #14710 .

Mathlib/FieldTheory/SimpleExtension.lean
Command-Master: Done.

tb65536:
Having this be a typeclass with noncanonical data seems a bit strange to me. Is there a reason you choice this approach over a plain `structure` or a data-less predicate or adding an assumption `F⟮k⟯ = ⊤` to the `IsFractionRing F[X] K` result?

Command-Master:
> Having this be a typeclass with noncanonical data seems a bit strange to me. Is there a reason you choice this approach over a plain `structure` or a data-less predicate or adding an assumption `F⟮k⟯ = ⊤` to the `IsFractionRing F[X] K` result?

What do you mean by "adding an assumption `F⟮k⟯ = ⊤` to the `IsFractionRing F[X] K` result"? Without having noncanonical data I can't get an `Algebra F[X] K` instance

tb65536:
> What do you mean by "adding an assumption `F⟮k⟯ = ⊤` to the `IsFractionRing F[X] K` result"? Without having noncanonical data I can't get an `Algebra F[X] K` instance

I just mean something like
```
theorem toIsFractionRing [Algebra.Transcendental F K] (k : K) (h : F⟮k⟯ = ⊤) :
    IsFractionRing F[X] K :=
```

acmepjz:
Is this PR still active? It needs to be migrate to fork and PR workflow.

joneugster:

PR #14713, opened 2024-07-13
[Merged by Bors] - fix: don't create new branches (and PRs) in update_dependencies.yml

bryangingechen:
The bot has been [creating a bunch of PRs](https://github.com/leanprover-community/mathlib4/issues?q=is%3Aopen+label%3Aauto-merge-after-CI+chore%3A+update+Mathlib+dependencies). This is because the current workflow [adds a timestamp suffix to the branch name](https://github.com/leanprover-community/mathlib4/blob/0512d468dedbde988a3759fcb4118c07e1cd3185/.github/workflows/update_dependencies.yml), which is discouraged in the README of the [create-pull-request action that we're using](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#alternative-strategy---always-create-a-new-pull-request-branch) (precisely because it leads to the creation of lots of PRs).

After this PR, the bot will always use a branch named `update-dependencies-bot-use-only` and hence should only create a new PR if the previous one has been merged. Note that the action force-pushes to this branch, so it's possible that the bot can end up canceling the bors queue. If this turns out to be an issue, we could add a step which checks if the `ready-to-merge` label is present, but I've put this off for now.

Summary of [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/.22update.20mathlib.20dependencies.22.20PRs) (link requires access to the mathlib reviewers channel on Zulip).

---

### DIFF
--- a/.github/workflows/update_dependencies.yml
+++ b/.github/workflows/update_dependencies.yml
@@ -29,11 +29,9 @@ jobs:
       - name: Update dependencies
         run: lake update
 
-      - name: Generate branch name
-        id: generate_branch_name
+      - name: Generate PR title
         run: |
           timestamp=$(date -u +"%Y-%m-%d-%H-%M")
-          echo "branch_name=update-dependencies-$timestamp" >> "$GITHUB_ENV"
           echo "pr_title=chore: update Mathlib dependencies $(date -u +"%Y-%m-%d")" >> "$GITHUB_ENV"
 
       - name: Create Pull Request
@@ -41,7 +39,7 @@ jobs:
         with:
           token: "${{ secrets.NIGHTLY_TESTING }}"
           commit-message: "chore: update Mathlib dependencies ${{ env.timestamp }}"
-          branch: "${{ env.branch_name }}"
+          branch: "for-bot-use-only-update-dependencies"
           base: master
           title: "${{ env.pr_title }}"
           body: "This PR updates the Mathlib dependencies."
--- a/.github/workflows/update_dependencies.yml
+++ b/.github/workflows/update_dependencies.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Generate PR title
         run: |
-          timestamp=$(date -u +"%Y-%m-%d-%H-%M")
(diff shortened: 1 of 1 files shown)


.github/workflows/update_dependencies.yml
```diff
@@ -29,19 +29,17 @@ jobs:
...
-          timestamp=$(date -u +"%Y-%m-%d-%H-%M")
-          echo "branch_name=update-dependencies-$timestamp" >> "$GITHUB_ENV"
+          echo "timestamp=$(date -u +"%Y-%m-%d-%H-%M")" >> "$GITHUB_ENV"
           echo "pr_title=chore: update Mathlib dependencies $(date -u +"%Y-%m-%d")" >> "$GITHUB_ENV"
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
         with:
           token: "${{ secrets.NIGHTLY_TESTING }}"
           commit-message: "chore: update Mathlib dependencies ${{ env.timestamp }}"
```
bryangingechen, line 41: It looks like `env.timestamp` was always blank because `timestamp` wasn't being saved to `$GITHUB_ENV`.

semorrison:
The bot that reports failures on zulip is expecting the branch name to begin (rather than end) with `update-dependencies`.

.github/workflows/update_dependencies.yml
```diff
@@ -29,19 +29,17 @@ jobs:
...
-          branch: "${{ env.branch_name }}"
+          branch: "for-bot-use-only-update-dependencies"
```
bryangingechen, line 42:
```suggestion
          # this branch is referenced in update_dependencies_zulip.yml
          branch: "update-dependencies-bot-use-only"
```

semorrison:
bors merge

bryangingechen:
The issue where the bot pushes to the same branch and cancels bors batches can be seen here #16286. The problem is that this ends up canceling the bors batches that the PR is part of and delaying all the other PRs there. I'll try and address this by implementing the "ready-to-merge" check later.

PR #14715, opened 2024-07-13
feat:(Combinatorics/SimpleGraph): connected `SimpleGraph` forms a `MetricSpace`

Rida-Hamadani:
---

- [ ] depends on: #14582 


Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -152,14 +243,38 @@ theorem Walk.isPath_of_length_eq_dist {u v : V} (p : G.Walk u v) (hp : p.length
...
     ∃ (p : G.Walk u v), p.IsPath ∧ p.length = G.dist u v := by
-  obtain ⟨p, h⟩ := hconn.exists_walk_of_dist u v
+  obtain ⟨p, h⟩ := hconn.exists_walk_length_eq_dist  u v
   exact ⟨p, p.isPath_of_length_eq_dist h, h⟩
 
+end dist
+
+section Topology
+
+def Verts {V : Type*} (_ : SimpleGraph V) := V
```
kmill, line 260: I don't think this is a good idea. I know I suggested it, but I don't think it should be included in mathlib. The problem is that now `Verts G` is another way to refer to the vertex type, which will create issues.

We should build a bridge to Quiver if we really want this.

Rida-Hamadani: Thanks for the review! I will convert this to a draft and work on it.

PR #14721, opened 2024-07-14
[Merged by Bors] - refactor(Order/Minimal): change minimality to a predicate

apnelson1:
This PR changes the notion of minimality from being a set with an unbundled relation `r` (i.e. having type signature `minimals (r : α → α → Prop) (s : Set α) : Set α`) to being a predicate with a bundled order relation in a typeclass:  `Minimal [LE α] (P : α → Prop) : α → Prop`.
It also does the same for maximality. 

This is hopefully an improvement, for a few reasons : 
* ergonomic API with dot notation (eg `Minimal.eq_of_le`, `Minimal.not_lt`), 
* no need to specify a particular relation explicitly (no use of `minimals`/`maximals` for a relation that was not defeq to `LE.le` currently appears anywhere in mathlib),
* the potential in future to unify ad hoc notions of min/maximality appearing in various locations in mathlib (eg `zorn_preorder`, `Sylow.is_maximal'` to name a couple of random examples). 

Since this change removes the definitions `minimals` and `maximals` and all their API, it required a total overhaul of the contents of `Order.Minimal`, and necessarily required updating all uses of `maximals`/`minimals` elsewhere. There were not all too many of these; this PR is designed to be the smallest it could be while making this change. 

The changes required outside `Order.Minimal` were: 

- redefinition of `Ideal.minimalPrimes` in `RingTheory/Ideal/MinimalPrime` to use `Minimal` rather than `minimals`, with the necessary minor changes to proofs that used the old definition, including in `AlgebraicGeometry/PrimeSpectrum/Basic`.
- redefinition of `irreducibleComponents` in `Topology/Irreducible` to use `Minimal` rather than `minimals`, with necessary minor changes to proofs that used the old definition. 
- updating of the maximality used in the definition of a matroid in `Data/Matroid/Basic` from `maximals` to `Maximal`, with the necessary changes to everywhere in `Data/Matroid/*` that a matroid was specified using the definition directly. 
- the necessary changes to two lemmas in `Data/Finset/Lattice` that used `minimals`/`maximals` in the statements. 

I attempted to clean up the API in `Order.Minimal` to work smoothly with the new definitions, but numerous judgment calls were needed; I removed a few lemmas about `maximals/minimals` that were used nowhere, and seemed unlikely to be useful with the new definition.

Comment is welcome!

---


Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+
+/-- If `P` is up-closed, then maximal elements satisfying `P` are exactly the globally maximal
+elements satisfying `P`. -/
+theorem maximal_iff_isMax (hP : ∀ ⦃x y⦄, P y → y ≤ x → P x) : Maximal P x ↔ P x ∧ IsMax x :=
+  ⟨fun h ↦ ⟨h.prop, fun _ h' ↦ h.le_of_ge (hP h.prop h') h'⟩, fun h ↦ ⟨h.1, fun _ _  h' ↦ h.2 h'⟩⟩
+
+theorem Minimal.and_right (h : Minimal P x) (hQ : Q x) : Minimal (fun x ↦ (P x ∧ Q x)) x :=
+  ⟨⟨h.prop, hQ⟩, fun _ hy ↦ h.le_of_le hy.1⟩
+
+theorem Minimal.and_left (h : Minimal P x) (hQ : Q x) : Minimal (fun x ↦ (Q x ∧ P x)) x :=
```
YaelDillies, line 117: Can you first show the more general
```suggestion
theorem Minimal.mono (h : Minimal P x) (h : Q ≤ P) (hQ : Q x) : Minimal Q x :=
```

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+theorem Minimal.or (h : Minimal (fun x ↦ P x ∨ Q x) x) : Minimal P x ∨ Minimal Q x := by
+  obtain ⟨(h | h), hmin⟩ := h
+  · exact .inl ⟨h, fun y hy hyx ↦ hmin (Or.inl hy) hyx⟩
+  exact .inr ⟨h, fun y hy hyx ↦ hmin (Or.inr hy) hyx⟩
+
+theorem Maximal.or (h : Maximal (fun x ↦ P x ∨ Q x) x) : Maximal P x ∨ Maximal Q x :=
+  Minimal.or (α := αᵒᵈ) h
+
+theorem minimal_and_iff_right_of_imp (hPQ : ∀ ⦃x⦄, P x → Q x) :
+    Minimal (fun x ↦ P x ∧ Q x) x ↔ (Minimal P x) ∧ Q x := by
```
YaelDillies, line 141: This lemma is really weird. Without even knowing what the definition of `Minimal` is, the LHS simplifies to
```suggestion
    Minimal P x ↔ (Minimal P x) ∧ Q x := by
```
Maybe just drop it?

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+  ⟨⟨hQ, h.prop⟩, fun _ hy ↦ h.le_of_ge hy.2⟩
+
+@[simp] theorem minimal_eq_iff : Minimal (· = y) x ↔ x = y := by
+  simp (config := {contextual := true}) [Minimal]
+
+@[simp] theorem maximal_eq_iff : Maximal (· = y) x ↔ x = y := by
+  simp (config := {contextual := true}) [Maximal]
+
+theorem Minimal.or (h : Minimal (fun x ↦ P x ∨ Q x) x) : Minimal P x ∨ Minimal Q x := by
+  obtain ⟨(h | h), hmin⟩ := h
```
YaelDillies, line 133: Unnecessary brackets
```suggestion
  obtain ⟨h | h, hmin⟩ := h
```

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+@[simp] theorem maximal_ge_iff : Maximal (y ≤ ·) x ↔ y ≤ x ∧ IsMax x :=
+  minimal_le_iff (α := αᵒᵈ)
+
+@[simp] theorem minimal_lt_iff : Minimal (· < y) x ↔ x < y ∧ IsMin x :=
+  minimal_iff_isMin (fun _ _ h h' ↦ h'.trans_lt h)
+
+@[simp] theorem maximal_gt_iff : Maximal (y < ·) x ↔ y < x ∧ IsMax x :=
+  minimal_lt_iff (α := αᵒᵈ)
+
+theorem not_minimal_iff (hx : P x) : ¬ Minimal P x ↔ ∃ y, y < x ∧ P y := by
```
YaelDillies, line 193: Can you make `not_minimal_iff` be the (currently missing) preorder version, and `not_minimal_iff_exists_lt` be this lemma?

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+  rw [← not_iff_not, not_not, not_exists]
+  simp only [Minimal, hx, true_and, lt_iff_le_not_le, not_and, and_imp, not_imp_not]
+  tauto
+
+theorem not_maximal_iff (hx : P x) : ¬ Maximal P x ↔ ∃ y, x < y ∧ P y :=
+  not_minimal_iff (α := αᵒᵈ) hx
+
+theorem exists_of_not_minimal (hx : P x) (h : ¬ Minimal P x) : ∃ y, y < x ∧ P y :=
+  (not_minimal_iff hx).1 h
```
YaelDillies, line 202:
```suggestion
alias ⟨exists_of_not_minimal, _⟩ :=not_minimal_iff
```

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+theorem maximal_iff_eq (hy : P y) (hP : ∀ ⦃x⦄, P x → x ≤ y) : Maximal P x ↔ x = y :=
+  minimal_iff_eq (α := αᵒᵈ) hy hP
+
+@[simp] theorem minimal_ge_iff : Minimal (y ≤ ·) x ↔ x = y :=
+  minimal_iff_eq rfl.le fun _ ↦ id
+
+@[simp] theorem maximal_le_iff : Maximal (· ≤ y) x ↔ x = y :=
+  maximal_iff_eq rfl.le fun _ ↦ id
+
+theorem minimal_iff_minimal_of_imp_of_forall (hPQ : ∀ ⦃x⦄, Q x → P x)
```
YaelDillies, line 253:
```suggestion
theorem minimal_congr_of_imp_of_forall (hPQ : ∀ ⦃x⦄, Q x → P x)
```
maybe? I'm not too attached

Mathlib/Order/Minimal.lean
```diff
@@ -1,507 +1,645 @@
...
+  minimal_iff_minimal_of_imp_of_forall (α := αᵒᵈ) hPQ h
+
+end PartialOrder
+
+section Subset
+
+variable {P : Set α → Prop} {s t : Set α}
+
+theorem Minimal.eq_of_superset (h : Minimal P s) (ht : P t) (hts : t ⊆ s) : s = t :=
+  h.eq_of_ge ht hts
```
YaelDillies, line 273: It's annoying that those lemmas don't apply to finset

apnelson1: I think this is out of scope - nothing about Finset was in the original file.

Mathlib/Order/Minimal.lean
apnelson1: I think I prefer this way.

Mathlib/Order/Minimal.lean
apnelson1: In the abstract, it looks weird, but the intention here is that the lemma provides a motive for rewriting that would otherwise be annoying to write explicitly, say if you have an goal of the form `_ ∧ Q x ↔ Minimal (fun x ↦ P x ∧ Q x) x` and you want to reduce it to `Q x → ((Minimal P x) ↔ _)`, where `Q` and `P` have long statements.

apnelson1: (note that `rw [and_iff_left_of_imp]` wouldn't work in that example because of the binder, and `simp_rw [and_iff_left_of_imp _]` would require an annoying proof term if `P` and `Q` are complicated. )

Mathlib/Order/Minimal.lean
apnelson1: (I've added a TODO)

Mathlib/Order/Minimal.lean
YaelDillies: Is `R x ∧ Q x ↔ Minimal (fun x ↦ P x ∧ Q x)` really such a common goal?

apnelson1: I'm not making a point purely on the basis of that kind of goal, but that kind of thing has certainly come up for me.

YaelDillies: I would like to see some concrete examples. I suspect there's a better(/more general?) lemma to be had

apnelson1: I think I've answered your question (it provides a motive for a rewrite), and I don't think it makes sense to ask for concrete examples when this is by definition new API. 

If you have an abstract lemma in mind, suggest it!

YaelDillies: You are free to provide non-minimal concrete examples deep down in one of your repos if that makes it easier for you!

apnelson1: Any decision to remove or keep these lemmas is marginally impactful at best, and given that I earnestly anticipate them to be somewhat useful at least in my case, I'd rather just move forward with the PR than continue this discussion.

Mathlib/Data/Finset/Lattice.lean
```diff
@@ -1814,19 +1814,17 @@ section minimal
...
+    Maximal P s ↔ P s ∧ ∀ x ∉ s, ¬ P (insert x s) := by
+  simp only [Maximal, and_congr_right_iff]
+  exact fun _ ↦ ⟨fun h x hxs hx ↦ hxs <| h hx (subset_insert _ _) (mem_insert_self x s),
+    fun h t ht hst x hxt ↦ by_contra fun hxs ↦ h x hxs (hP ht (insert_subset hxt hst))⟩
+
+theorem minimal_iff_forall_diff_singleton (hP : ∀ ⦃s t⦄, P t → t ⊆ s → P s) :
+    Minimal P s ↔ P s ∧ ∀ x ∈ s, ¬ P (s.erase x) :=
+  ⟨fun h ↦ ⟨h.prop, fun x hxs hx ↦ by simpa using h.le_of_le hx (erase_subset _ _) hxs⟩,
+    fun h ↦ ⟨h.1, fun t ht hts x hxs ↦ by_contra fun hxt ↦
+      h.2 x hxs <| hP ht (subset_erase.2 ⟨hts, hxt⟩)⟩⟩
```
YaelDillies, line 1827:
```suggestion
    Minimal P s ↔ P s ∧ ∀ x ∈ s, ¬ P (s.erase x) where
  mp h := ⟨h.prop, fun x hxs hx ↦ by simpa using h.le_of_le hx (erase_subset _ _) hxs⟩
  mpr h := ⟨h.1, fun t ht hts x hxs ↦ by_contra fun hxt ↦
    h.2 x hxs <| hP ht (subset_erase.2 ⟨hts, hxt⟩)⟩
```

Mathlib/Data/Matroid/Basic.lean
```diff
@@ -458,9 +458,9 @@ theorem eq_of_base_iff_base_forall {M₁ M₂ : Matroid α} (hE : M₁.E = M₂.
       fun hB ↦ (h <| hB.subset_ground.trans_eq hE.symm).2 hB⟩
   ext <;> simp [hE, M₁.indep_iff', M₂.indep_iff', h']
 
-theorem base_compl_iff_mem_maximals_disjoint_base (hB : B ⊆ M.E := by aesop_mat) :
-    M.Base (M.E \ B) ↔ B ∈ maximals (· ⊆ ·) {I | I ⊆ M.E ∧ ∃ B, M.Base B ∧ Disjoint I B} := by
-  simp_rw [mem_maximals_setOf_iff, and_iff_right hB, and_imp, forall_exists_index]
+theorem base_compl_iff_maximal_disjoint_base (hB : B ⊆ M.E := by aesop_mat) :
+    M.Base (M.E \ B) ↔ Maximal (fun I ↦ I ⊆ M.E ∧ ∃ B, M.Base B ∧ Disjoint I B) B := by
+  simp_rw [maximal_iff, and_iff_right hB,  and_imp, forall_exists_index]
```
YaelDillies, line 463:
```suggestion
  simp_rw [maximal_iff, and_iff_right hB, and_imp, forall_exists_index]
```

Mathlib/Data/Matroid/Basic.lean
```diff
@@ -566,16 +566,15 @@ theorem Base.eq_of_subset_indep (hB : M.Base B) (hI : M.Indep I) (hBI : B ⊆ I)
   let ⟨B', hB', hB'I⟩ := hI.exists_base_superset
   hBI.antisymm (by rwa [hB.eq_of_subset_base hB' (hBI.trans hB'I)])
 
-theorem base_iff_maximal_indep : M.Base B ↔ M.Indep B ∧ ∀ I, M.Indep I → B ⊆ I → B = I := by
-  refine ⟨fun h ↦ ⟨h.indep, fun _ ↦ h.eq_of_subset_indep ⟩, fun ⟨h, h'⟩ ↦ ?_⟩
+theorem base_iff_maximal_indep : M.Base B ↔ Maximal M.Indep B := by
+  rw [maximal_subset_iff]
+  refine ⟨fun h ↦ ⟨h.indep, fun _ ↦ h.eq_of_subset_indep⟩, fun ⟨h, h'⟩ ↦ ?_⟩
```
YaelDillies, line 571: When doing a refactor, adding unnecessary changes is *not* helping. I understand that your PR is motivated by applications in matroid theory, but I don't need to see the changes in an already big diff to be convinced.

You have already added them here, so I won't ask you to revert, but this is typically the kind of changes that can/should be left to a separate PR.

Mathlib/Data/Matroid/Basic.lean
```diff
@@ -837,10 +829,8 @@ theorem Basis.not_basis_of_ssubset (hI : M.Basis I X) (hJI : J ⊂ I) : ¬ M.Bas
 
 theorem Indep.subset_basis_of_subset (hI : M.Indep I) (hIX : I ⊆ X) (hX : X ⊆ M.E := by aesop_mat) :
     ∃ J, M.Basis J X ∧ I ⊆ J := by
-  obtain ⟨J, ⟨(hJ : M.Indep J),hIJ,hJX⟩, hJmax⟩ := M.maximality X hX I hI hIX
-  use J
-  rw [and_iff_left hIJ, basis_iff, and_iff_right hJ, and_iff_right hJX]
-  exact fun K hK hJK hKX ↦ hJK.antisymm (hJmax ⟨hK, hIJ.trans hJK, hKX⟩ hJK)
+  obtain ⟨J, hJ, hJmax⟩ := M.maximality X hX I hI hIX
+  exact ⟨J, ⟨⟨hJmax, hX⟩, hJ⟩⟩
```
YaelDillies, line 833:
```suggestion
  exact ⟨J, ⟨hJmax, hX⟩, hJ⟩
```

Mathlib/Data/Matroid/Dual.lean
```diff
@@ -67,50 +67,46 @@ section dual
...
-    subst this
-
-    refine subset_diff.mpr ⟨hJX, by_contra (fun hne ↦ ?_)⟩
-    obtain ⟨e, heJ, heB'⟩ := not_disjoint_iff.mp hne
+    by_contra hdj'
+    obtain ⟨e, heJ, heB'⟩ := not_disjoint_iff.mp hdj'
     obtain (heB'' | ⟨-,heX⟩ ) := hB₁I heB'
     · exact hdj.ne_of_mem heJ heB'' rfl
     exact heX (hJX heJ)
+
```
YaelDillies, line 109:
```suggestion
```

Mathlib/Data/Matroid/IndepAxioms.lean
```diff
@@ -97,8 +97,8 @@ structure IndepMatroid (α : Type*) where
   (Indep : Set α → Prop)
   (indep_empty : Indep ∅)
   (indep_subset : ∀ ⦃I J⦄, Indep J → I ⊆ J → Indep I)
-  (indep_aug : ∀⦃I B⦄, Indep I → I ∉ maximals (· ⊆ ·) {I | Indep I} →
-      B ∈ maximals (· ⊆ ·) {I | Indep I} → ∃ x ∈ B \ I, Indep (insert x I))
+  (indep_aug : ∀⦃I B⦄, Indep I → ¬ Maximal Indep I →
```
YaelDillies, line 100:
```suggestion
  (indep_aug : ∀ ⦃I B⦄, Indep I → ¬ Maximal Indep I →
```

Mathlib/Data/Matroid/IndepAxioms.lean
```diff
@@ -167,23 +160,19 @@ namespace IndepMatroid
...
-    simp only [mem_maximals_iff, mem_setOf_eq, not_and, not_forall, exists_prop,
-      exists_and_left, iff_true_intro hI, true_imp_iff] at hImax hBmax
-    obtain ⟨I', hI', hII', hne⟩ := hImax
-    obtain ⟨e, heI', heI⟩ := exists_of_ssubset (hII'.ssubset_of_ne hne)
-    have hins : Indep (insert e I) := indep_subset hI' (insert_subset heI' hII')
-    obtain (heB | heB) := em (e ∈ B)
+    obtain ⟨e, heI, hins⟩ := exists_insert_of_not_maximal indep_subset hI hImax
+    by_cases heB : e ∈ B
     · exact ⟨e, ⟨heB, heI⟩, hins⟩
+
```
YaelDillies, line 166:
```suggestion
```
(I assume those are typos? since a refactor is not really the place to orthogonally prettify proofs)

Mathlib/RingTheory/Ideal/MinimalPrime.lean
```diff
@@ -36,16 +36,16 @@ variable {R S : Type*} [CommSemiring R] [CommSemiring S] (I J : Ideal R)
 
 /-- `I.minimalPrimes` is the set of ideals that are minimal primes over `I`. -/
 protected def Ideal.minimalPrimes : Set (Ideal R) :=
-  minimals (· ≤ ·) { p | p.IsPrime ∧ I ≤ p }
+  {p | Minimal (fun p ↦ p.IsPrime ∧ I ≤ p) p}
```
YaelDillies, line 39: There are two `p` in context
```suggestion
  {p | Minimal (fun q ↦ q.IsPrime ∧ I ≤ q) p}
```

Mathlib/Order/Minimal.lean
YaelDillies, line 1: Sorry, but this diff is unreviewable as is 😢. I cannot assert whether the former API was preserved. Can you please split this PR into
1. A PR solely adding new API and reproving the existing API using it. This demonstrates that the new API is at least as powerful as the old one.
2. A PR replacing all uses of the old API by the new API
3. A PR deprecating the old API

Possibly, PRs 2 and 3 can be made into a single one, but I insist that PR 1 should be separate.

apnelson1: You are departing in tone from what one would normally consider a 'review'. But I will make a PR of the form (1.) as you suggest. 

Also, for the benefit of anyone doing a refactor, I think that guidelines like your comment should be available in advance somewhere easily accessible to anyone making a PR. I think that in this case, such a small refactor could easily be done in one PR without adverse consequences; learning that things should be done a certain way as a matter of policy shouldn't be so ad hoc.

YaelDillies: They certainly do not *have* to be this way, but my experience is that you get much better feedback (both from Lean and from the reviewers) when you do things the way I explain above. Indeed, I basically came up with that method of proceeding and I don't think anyone but me has seriously used it in mathlib refactors yet.

Mathlib/AlgebraicGeometry/PrimeSpectrum/Basic.lean
```diff
@@ -623,8 +624,10 @@ lemma vanishingIdeal_irreducibleComponents :
     vanishingIdeal '' (irreducibleComponents <| PrimeSpectrum R) =
     minimalPrimes R := by
   rw [irreducibleComponents_eq_maximals_closed, minimalPrimes_eq_minimals,
-    ← minimals_swap, ← vanishingIdeal_isClosed_isIrreducible, image_minimals_of_rel_iff_rel]
-  exact fun s t hs _ ↦ vanishingIdeal_anti_mono_iff hs.1
+    image_antitone_setOf_maximal (fun s t hs _ ↦ (vanishingIdeal_anti_mono_iff hs.1).symm),
+    show Ideal.IsPrime = fun x ↦ x ∈ {x | Ideal.IsPrime x} from rfl,
```
j-loreaux, line 628: This `show` is not pretty. You could do: `← funext (@Set.mem_setOf_eq _ · Ideal.IsPrime)`, but I think it would be better to add the `rfl` lemma:
```lean
import Mathlib.Data.Set.Basic

example {α : Type u} {p : α → Prop} :
    (fun a ↦ a ∈ {x : α | p x}) = p :=
  rfl
```

Mathlib/Data/Matroid/Dual.lean
```diff
@@ -67,50 +67,46 @@ section dual
...
+    obtain rfl : B₁ = B' := by
       refine hB₁.eq_of_subset_indep hB'.indep (fun e he ↦ ?_)
       refine (hB₁I he).elim (fun heB'' ↦ ?_) (fun h ↦ h.1)
       refine (em (e ∈ X)).elim (fun heX ↦ hI' (Or.inl ⟨heB'', heX⟩)) (fun heX ↦ hIB' ?_)
-      refine hI.mem_of_insert_indep ⟨hB₁.subset_ground he, heX⟩
-        (hB₁.indep.subset (insert_subset he ?_))
-      refine (subset_union_of_subset_right (subset_diff.mpr ⟨hIB',?_⟩) _).trans hI'B₁
-      exact disjoint_of_subset_left hI.subset disjoint_sdiff_left
+      refine hI.mem_of_insert_indep ⟨hB₁.subset_ground he, heX⟩ ?_
+      exact hB₁.indep.subset (insert_subset he (subset_union_right.trans hI'B₁))
```
j-loreaux, line 103: why are there random blank lines floating around here? And why are some added?

j-loreaux (commented):
I've left a few minor comments, but overall, I think this is okay. That said, I have not thoroughly vetted the new API in comparison to the old one. My feeling is that it seems to work with what we have (in that nothing breaks), and that in the worst case, we can add some things later that are missing. On the other hand, this is not my area of the library, so I won't merge at this time.

Mathlib/AlgebraicGeometry/PrimeSpectrum/Basic.lean
apnelson1: I opted for the former, because it looks to me like `mem_setOf_eq` should really be phrased as the lemma you suggest in the first place, and it wouldn't be the right place here to make that change.

Mathlib/Data/Matroid/IndepAxioms.lean
```diff
@@ -273,20 +263,22 @@ theorem _root_.Matroid.existsMaximalSubsetProperty_of_bdd {P : Set α → Prop}
...
     rwa [ncard_def, heq, ENat.toNat_coe]
-    -- have := (hP Y hY).2
-  obtain ⟨Y, hY, hY'⟩ := Finite.exists_maximal_wrt' ncard _ hfin ⟨I, hI, rfl.subset, hIX⟩
-  refine ⟨Y, hY, fun J ⟨hJ, hIJ, hJX⟩ (hYJ : Y ⊆ J) ↦ (?_ : J ⊆ Y)⟩
-  have hJfin := finite_of_encard_le_coe (hP J hJ)
-  refine (eq_of_subset_of_ncard_le hYJ ?_ hJfin).symm.subset
-  rw [hY' J ⟨hJ, hIJ, hJX⟩ (ncard_le_ncard hYJ hJfin)]
+  obtain ⟨Y, ⟨hY, hIY, hYX⟩, hY'⟩ :=
+    Finite.exists_maximal_wrt' ncard _ hfin ⟨I, hI, rfl.subset, hIX⟩
+
```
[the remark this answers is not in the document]
apnelson1: I've tried to revert to the way things were. I tend to include blank lines to separate ideas in proofs, and when I was rewriting to work with the new API, I didn't faithfully preserve where the blanks were.

apnelson1:
> I've left a few minor comments, but overall, I think this is okay. That said, I have not thoroughly vetted the new API in comparison to the old one. My feeling is that it seems to work with what we have (in that nothing breaks), and that in the worst case, we can add some things later that are missing. On the other hand, this is not my area of the library, so I won't merge at this time.

Thanks for looking at this - I am sorry reviewing wasn't easier!

Mathlib/Order/Minimal.lean
```diff
@@ -1,472 +1,661 @@
...
 
-* `maximals r s`: Maximal elements of `s` with respect to `r`.
-* `minimals r s`: Minimal elements of `s` with respect to `r`.
+* `Minimal P x`: `x` is minimal satisfying `P`.
+* `Maximal P x`: `x` is maximal satisfying `P`.
+
+## Implementation Details
+
+This file underwent a refactor from a version where minimality and maximality were defined using
+sets rather than predicates, and with an unbundled order relation rather than a `LE` instance.
```
b-mehta, line 24: This is the main change that worries me here (all the others look great), in particular this forbids us from (conveniently) talking about minimality with respect to a relation that isn't the default ordering on the type... For the moment, perhaps it's enough to add a TODO saying that this case might need further consideration?

Mathlib/Data/Matroid/IndepAxioms.lean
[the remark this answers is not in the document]
j-loreaux: @apnelson1 if you feel there is a separate enough flow ‌/ idea in the proof which is not otherwise captured by the structure of the proof, that's a sign that you should add a *comment* instead of a blank line.

Mathlib/Order/Minimal.lean
apnelson1: Thanks - I've added this TODO to the docstring.

b-mehta:
bors merge

PR #14722, opened 2024-07-14
[Merged by Bors] - chore(CategoryTheory/DiscreteCategory): remove discrete_fun_ma_eq_id, mv …

mo271:
…discrete_functor_obj_eq_as

Seems like `CategoryTheory.FreeMonoidalCategory.discrete_functor_map_eq_id` is an exact duplicate of `CategoryTheory.Discrete.functor_map`, except it was a simp lemma.

### DIFF
--- a/Mathlib/CategoryTheory/DiscreteCategory.lean
+++ b/Mathlib/CategoryTheory/DiscreteCategory.lean
@@ -183,9 +183,17 @@ theorem functor_obj {I : Type u₁} (F : I → C) (i : I) :
   rfl
 #align category_theory.discrete.functor_obj CategoryTheory.Discrete.functor_obj
 
+@[simp]
 theorem functor_map {I : Type u₁} (F : I → C) {i : Discrete I} (f : i ⟶ i) :
     (Discrete.functor F).map f = 𝟙 (F i.as) := by aesop_cat
 #align category_theory.discrete.functor_map CategoryTheory.Discrete.functor_map
+#align category_theory.free_monoidal_category.discrete_functor_map_eq_id CategoryTheory.Discrete.functor_map
+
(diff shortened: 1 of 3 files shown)


Mathlib/CategoryTheory/DiscreteCategory.lean
```diff
@@ -183,9 +183,17 @@ theorem functor_obj {I : Type u₁} (F : I → C) (i : I) :
   rfl
 #align category_theory.discrete.functor_obj CategoryTheory.Discrete.functor_obj
 
+@[simp]
```
grunweg, line 186: This was simp 1100 in the other file - did you mean to change this?

mo271: Yes, this change was intentional. I suspect the 1100 was only added to avoid the `simp can prove this` lint, but reordering this also solves it.

semorrison:
bors merge

Ruben-VandeVelde:
Is there a reason we don't have a deprecated alias?

mo271:
> Is there a reason we don't have a deprecated alias?

no, I simply forgot to do it. Will fix it later today...

mo271:
@Ruben-VandeVelde I added the deprecation aliases in #14791, please have a look if this is the correct way to do it. (I specified the complete name, I added the alias in the place they were moved to. Not sure if it would be better to add them in a place where they were moved from. I couldn't find documentation about this.

grunweg:
I think adding it where you moved it to makes sense. Lemmas *moving* is one source of upgrade conflicts users currently have to just live with. (There's a draft deprecation policy; I'll try to ensure this detail is included there.)

PR #14676, opened 2024-07-12
[Merged by Bors] - refactor: improvements to lint_style

grunweg:
To prepare for #14273; pulled out for easy of reviewing.
Commits can be reviewed individually.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14724, opened 2024-07-14
[Merged by Bors] - fix(lint_style): only print style errors once overall

grunweg:
The previous code meant that once an error was found, *all* errors so far would be printed after checking *every module*, leading to hundreds of repetitions of the same warning.

---


bryangingechen:
Thanks!
bors r+

grunweg:
Thanks for the really fast review!
For posterity: this regressed in #14676.

PR #14726, opened 2024-07-14
[Merged by Bors] - chore(Tactic): reduce use of Implicit, part 2

grunweg:
---

### DIFF
--- a/Mathlib/Tactic/ToAdditive.lean
+++ b/Mathlib/Tactic/ToAdditive.lean
@@ -475,9 +475,6 @@ structure Config : Type where
   existing : Option Bool := none
   deriving Repr
 
---set_option autoImplicit true
---variable [Monad M] [MonadOptions M] [MonadEnv M]
-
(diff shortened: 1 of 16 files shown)


Mathlib/Tactic/ToAdditive.lean
```diff
@@ -477,7 +475,8 @@ structure Config : Type where
   existing : Option Bool := none
   deriving Repr
 
-variable [Monad M] [MonadOptions M] [MonadEnv M]
+--set_option autoImplicit true
+--variable [Monad M] [MonadOptions M] [MonadEnv M]
```
eric-wieser, line 479: Delete?

eric-wieser (approved):

Mathlib/Tactic/ToAdditive.lean
grunweg: Indeed!

jcommelin:
Thanks :tada:

bors merge

PR #14730, opened 2024-07-14
[Merged by Bors] - feat(RingTheory/Derivation/Basic): add `coe_restrictScalars` lemma for derivations

Command-Master:
Add `coe_restrictScalars` lemma for derivations.

---

### DIFF
--- a/Mathlib/RingTheory/Derivation/Basic.lean
+++ b/Mathlib/RingTheory/Derivation/Basic.lean
@@ -358,6 +358,9 @@ protected def restrictScalars (d : Derivation S A M) : Derivation R A M where
   toLinearMap := d.toLinearMap.restrictScalars R
 #align derivation.restrict_scalars Derivation.restrictScalars
 
+@[simp]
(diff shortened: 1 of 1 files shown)


Mathlib/RingTheory/Derivation/Basic.lean
```diff
@@ -358,6 +358,9 @@ protected def restrictScalars (d : Derivation S A M) : Derivation R A M where
   toLinearMap := d.toLinearMap.restrictScalars R
 #align derivation.restrict_scalars Derivation.restrictScalars
 
+@[simp]
+lemma coe_restrictScalars (d : Derivation S A M) : ⇑(d.restrictScalars R) = ⇑d := rfl
```
eric-wieser, line 362: Often we don't make these lemmas `simp`, since that means they can end up applying before the `Derivation`-specific lemmas have a chance.

Can you add `restrictScalars_apply` too, and make that `simp` instead?

Command-Master: Done.

eric-wieser (approved):
bors merge

Thanks!

PR #14731, opened 2024-07-14
feat: the repeated typeclass assumption linter

adomani:
[#mathlib4 > recyclable variables @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/recyclable.20variables/near/451167963) also [#mathlib4 > Normed modules @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Normed.20modules/near/512242167)

---


leanprover-community-bot-assistant:

Mathlib/Tactic/Linter/DependentTypeclass.lean
```diff
@@ -0,0 +1,121 @@
...
+`d` using `varDecls`.
+Otherwise it returns `none`.
+-/
+def getMinInstsFor (d : TSyntax ``Lean.Parser.Term.bracketedBinder)
+    (varDecls : Array (TSyntax ``Lean.Parser.Term.bracketedBinder)) :
+      CommandElabM (Option Expr) := do
+  let sc ← getScope
+  let newName ← liftCoreM <| mkFreshUserName `hello
+  let stxId ← `(declId| $(mkIdent newName))
+  let tac ← `(tacticSeq| first | infer_instance | sorry)
```
mattrobball, line 51: This seems like a very large hammer for the task of finding instances. 

What do you want to run continuously versus in CI? 

Looking at the Zulip conversation, I would guess, for running continuously, it would be enough if we look at `LocalInstances` and check for overlapping projections to some data field. 

For CI, I think we can have a flag that is more aggressive so that people don't do `variable [Field \R]` for example. 


What do you think?

PR #14732, opened 2024-07-14
[Merged by Bors] - refactor (Algebra/Order/AddTorsor) : Add multiplicative versions of everything, use to_additive

ScottCarnahan:
This PR adds multiplicative versions of the existing Prop-valued mixin classes relating vector addition to order.  Also, we rename:

-  `OrderedVAdd` to `IsOrderedVAdd`
-  `CancelVAdd` to `IsCancelVAdd`
-  `OrderedCancelVAdd` to `IsOrderedCancelVAdd`

---

### DIFF
--- a/Mathlib/Algebra/Order/AddTorsor.lean
+++ b/Mathlib/Algebra/Order/AddTorsor.lean
@@ -12,7 +12,8 @@ This file defines ordered scalar multiplication and vector addition, and proves 
 In the additive case, a motivating example is given by the additive action of `ℤ` on subsets of
 reals that are closed under integer translation.  The order compatibility allows for a treatment of
 the `R((z))`-module structure on `(z ^ s) V((z))` for an `R`-module `V`, using the formalism of Hahn
-series.
+series.  In the multiplicative case, a standard example is the action of non-negative rationals on
+an ordered field.
 
 ## Implementation notes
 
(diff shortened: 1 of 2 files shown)


Mathlib/Algebra/Order/AddTorsor.lean
```diff
@@ -44,65 +53,95 @@ open Function
...
-  vadd_le_vadd_left _ _ := add_le_add_left
-  vadd_le_vadd_right _ _ h a := add_le_add_right h a
+@[deprecated (since := "2024-07-15")] alias OrderedVAdd := IsOrderedVAdd
 
-instance OrderedVAdd.toCovariantClassLeft [LE G] [LE P] [VAdd G P] [OrderedVAdd G P] :
-    CovariantClass G P (· +ᵥ ·) (· ≤ ·) where
-  elim := fun a _ _ bc ↦ OrderedVAdd.vadd_le_vadd_left _ _ bc a
+/-- An ordered scalar multiplication is a bi-monotone scalar multiplication. -/
+@[to_additive]
+class IsOrderedSMul (G P : Type*) [LE G] [LE P] [SMul G P] : Prop where
```
eric-wieser, line 64: How does this differ from the existing [`OrderedSMul`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/Module/OrderedSMul.html#OrderedSMul)?

ScottCarnahan: They seem to be almost completely different:

- `OrderedSMul` is about strict inequality, while `IsOrderedSMul` is about `LE`
- `OrderedSMul` requires `G` to be an ordered semiring and `P` to have `OrderedAddCommMonoid` structure.  However, the algebraic structures do not appear in the defining `Prop`s.
- `OrderedSMul` conditions only trigger for strictly positive elements of `G`, while `IsOrderedSMul` more or less requires `G` to be made of non-negative elements.  I'm really only interested in the additive version, where this condition is meaningless.

I am a little concerned about the near collision in names, but the `OrderedSMul` module docstring suggests that it may be deprecated at some point in the future.

eric-wieser: Maybe @YaelDillies has thoughts

YaelDillies: To explain, this is the same distinction as between `OrderedCommMonoid` and `OrderedCommMonoidWithZero`.

I indeed hope `OrderedSMul` will go away soon, but it's a bit more complicated than I had originally hoped

semorrison:
bors merge

eric-wieser:
bors r-

I don't think we should merge this without some docstring disambiguating `OrderedSMul` and `IsOrderedSMul`

ScottCarnahan:
Sorry, I had not picked up the suggestion underlying your initial comment.

eric-wieser:
bors d+

No worries, the suggestion was mainly "let's get more input from Yael to decide what to do next"

PR #14735, opened 2024-07-14
[Merged by Bors] - feat(Algebra/Polynomial/AlgebraMap): generalize one of the ring in `map_aeval_eq_aeval_map` to `Semiring`

acmepjz:
... it is not needed to be commutative.

---


eric-wieser (approved):
bors merge

Thanks!

PR #14736, opened 2024-07-14
[Merged by Bors] - feat(Measure/WithDensity): drop measurability assumption

urkud:
- Show that for an s-finite measure, any function is equivalent to a
  measurable function in the sense of Lebesgue integrals over subsets.
- Use it to drop measurability assumption
  and add an instance `Measure.withDensity.instSFinite`.

---


RemyDegenne:
Thanks!
bors r+

PR #14737, opened 2024-07-15
[Merged by Bors] - feat(CstarRing): weaken the definition of `CstarRing` from an equality to an inequality

dupuisf:
This PR weakens the definition of `CstarRing` from `‖x⋆ * x‖ = ‖x‖ * ‖x‖` to `‖x‖ * ‖x‖ ≤ ‖x⋆ * x‖`. The "weaker" condition is then shown to be equivalent to the original one.

---


semorrison:
bors merge

PR #14738, opened 2024-07-15
[Merged by Bors] - feat(Topology/ENNReal): add `finset_sum_iSup`

urkud:
Moves:
- finset_sum_iSup_nat -> finsetSum_iSup_of_monotone

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14740, opened 2024-07-15
[Merged by Bors] - feat(Integral/Lebesgue): add `setLIntegral_biUnion_of_directed`

urkud:
also drop some unneeded measurability assumptions.

---

- [x] depends on: #14738

### DIFF
--- a/Mathlib/MeasureTheory/Function/SimpleFunc.lean
+++ b/Mathlib/MeasureTheory/Function/SimpleFunc.lean
@@ -1083,6 +1083,15 @@ theorem restrict_lintegral_eq_lintegral_restrict (f : α →ₛ ℝ≥0∞) {s :
   rw [f.restrict_lintegral hs, lintegral_restrict]
 #align measure_theory.simple_func.restrict_lintegral_eq_lintegral_restrict MeasureTheory.SimpleFunc.restrict_lintegral_eq_lintegral_restrict
 
+theorem lintegral_restrict_iUnion_of_directed {ι : Type*} [Countable ι]
+    (f : α →ₛ ℝ≥0∞) {s : ι → Set α} (hd : Directed (· ⊆ ·) s) (μ : Measure α) :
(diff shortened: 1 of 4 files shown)


Mathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -1083,6 +1083,15 @@ theorem restrict_lintegral_eq_lintegral_restrict (f : α →ₛ ℝ≥0∞) {s :
   rw [f.restrict_lintegral hs, lintegral_restrict]
 #align measure_theory.simple_func.restrict_lintegral_eq_lintegral_restrict MeasureTheory.SimpleFunc.restrict_lintegral_eq_lintegral_restrict
 
+theorem lintegral_restrict_iUnion_of_directed {ι : Type*} [Countable ι]
+    (f : α →ₛ ℝ≥0∞) {s : ι → Set α} (hd : Directed (· ⊆ ·) s) (μ : Measure α) :
+    f.lintegral (μ.restrict (⋃ i, s i)) = ⨆ i, f.lintegral (μ.restrict (s i)) := by
+  simp only [lintegral, Measure.restrict_iUnion_apply_eq_iSup hd (measurableSet_preimage ..),
+    ENNReal.mul_iSup]
+  refine finsetSum_iSup fun i j ↦ (hd i j).imp fun k ⟨hik, hjk⟩ ↦ fun a ↦ ?_
+  -- TODO (PR #14739) make `gcongr` close this goal
```
semorrison, line 1092: Very minor, but I think it's better to use full URLs rather than numbers, as they are both clickable by humans and unambiguously parseable by scripts (which may also have to deal with links to other repos).

semorrison:
bors d+

Mathlib/MeasureTheory/Function/SimpleFunc.lean
urkud:
```suggestion
  -- TODO https://github.com/leanprover-community/mathlib4/pull/14739 make `gcongr` close this goal
```

PR #14744, opened 2024-07-15
remove deprecated lemmas and simp tags

jcommelin:
---

This PR is stacked on top of #14644. See #14745 for the total PR against master.


riccardobrasca:
LGTM to me.

PR #14745, opened 2024-07-15
[Merged by Bors] - chore: remove bit0/bit1 and associated lemmas

jcommelin:
This PR replaces `bit0 x` resp. `bit1 x` with:
- `2 * x` resp. `2 * x + 1` if possible, or
- `2 • x` resp. `2 • x + 1` if possible, or
- `x + x` resp. `x + x + 1` as fallback.

The versions `Num.bit0` and `Num.bit1` (as well as the `PosNum` and `ZNum` versions) are left alone.
All lemmas that mention `bit0`/`bit1` have been adapted. A follow-up PR can remove these lemmas.

Deletions:
- bit0
- bit1

---

This PR combines #14644 and #14744. It might be easier to review those separately.
But I think that this current PR is the one that should be merged in the end.


riccardobrasca:
LGTM thanks! (I don't understand the mergify error).

bors d+

PR #14746, opened 2024-07-15
[Merged by Bors] - chore: separate dependent typeclass assumptions

adomani:
Flagged by the linter at #14731.

If you are reviewing, there is a challenge: sectioning off `IsDomain` from `IsDedekindDomain` means that a lemma that before this PR had both in scope and actually really used `IsDomain` has suddenly become "worse", since now it only has `IsDedekindDomain` in scope.

I am thinking about how to automatically address this challenge, but for the moment, this is what there is!  :smile: 

---


urkud:
Can you generate doc pages for both versions (or otherwise pretty-print all theorems in the file) and compare them?

mattrobball (approved):

mattrobball:
Thanks! 

bors delegate+

PR #14748, opened 2024-07-15
[Merged by Bors] - feat(AlgebraicGeometry): Define preimmersions.

erdOne:
---


semorrison:
There is apparently a massive import increase to Mathlib.Geometry.RingedSpace.SheafedSpace. can this be avoided?

erdOne:
Every downstream file that imports `Mathlib.Geometry.RingedSpace.SheafedSpace` also imports `Mathlib.RingedSpace.Stalks`, so this change only affects that single file. IMO stalks are fundamental to sheafed spaces, so this import is permissible from a library design POV. But if you insist I could split the new additions into a new file.

joelriou:
Thanks!

bors merge

grunweg:
Can you fix the build, please? Thanks!

erdOne:
Should be good to go now.

jcommelin:
bors merge

PR #14642, opened 2024-07-11
[Merged by Bors] - chore(Mathlib/Init/ZeroOne): rename `Mathlib.Init.ZeroOne` to `Mathlib.Algebra.Group.ZeroOne`

Komyyy:
---

- [x] depends on: #14750

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14750, opened 2024-07-15
[Merged by Bors] - chore(Mathlib/Tactic/ToAdditive/Frontend): isolate the `to_additive` attribute implementation in a new file `Tactic.ToAdditive.Frontend` (2/2)

Komyyy:
Ideally, we want to move the `attribute [to_additive]` commands in `Init.ZeroOne` into `Tactic.ToAdditive` without any other changes.
However, the `to_additive` attribute is initialized in the file and cannot be included in the same file.
So it would be necessary to isolate the original `to_additive` attribute implementation in a new file `Tactic.ToAdditive.Frontend`, and create an new file `Tactic.ToAdditive` as a placeholder for the `attribute [to_additive]` commands in `Init.ZeroOne`.
But if we put these two changes in the same PR, git cannot recognizes that `Tactic.ToAdditive` is renamed, so we split the PR.

---

- [x] depends on: #14816


jcommelin:
Dear @Komyyy I still don't understand why the file `Tactic/ToAdditive` needs to be renamed in the same PR that touches `Mathlib/Init/ZeroOne`. I would like to have a PR that does only the changes to `Tactic/ToAdditive`.

Komyyy:
@jcommelin
Ideally, the `attribute [to_additive]` commands in `Init.ZeroOne` should be moved into `Tactic.ToAdditive` without any other changes.
However, the `to_additive` attribute is initialized in the file and cannot be included in the same file. So I thought it would be necessary to isolate the original `to_additive` attribute implementation in a new file `Tactic.ToAdditive.Frontend` in the same PR as the PR to which the commands in that file are moved.

jcommelin:
I think you can do the "isolation" part before moving the commands in `Init.ZeroOne`. Or am I missing something?

Komyyy:
@jcommelin
I made this PR do only isolation part and moved the part moving the commands in `Init.ZeroOne` to #14642.

jcommelin:
Thanks. This is now an atomic change.
I assume it will pass CI, but let's wait to be sure:

bors d+

YaelDillies (approved):
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #14751, opened 2024-07-15
[Merged by Bors] - feat(RingTheory/Localization): more API and kernels of localized maps

chrisflav:
This PR adds various API lemmas on localization. In particular it computes the kernel of a localized map as the localization of the kernel.

---


erdOne:
Thanks!
maintainer merge

Mathlib/RingTheory/Ideal/Maps.lean
```diff
@@ -908,3 +908,14 @@ lemma coe_ideal_map (I : Ideal A) :
     Ideal.map f I = Ideal.map (f : A →+* B) I := rfl
 
 end AlgHom
+
+namespace Algebra
+
+variable {R : Type*} [CommSemiring R] (S : Type*) [Semiring S] [Algebra R S]
+
+@[simps!]
+def idealMap (I : Ideal R) : I →ₗ[R] I.map (algebraMap R S) :=
```
riccardobrasca, line 917: The linter will ask for a documentation here.

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/RingTheory/Ideal/Maps.lean
chrisflav: That got lost in copy pasting. Thanks!

chrisflav:
Thanks for the reviews!

PR #14754, opened 2024-07-15
[Merged by Bors] - chore(Logic/Function/Defs): move file out of Init

jcommelin:
---


Mathlib/Tactic/DeriveTraversable.lean
```diff
@@ -4,6 +4,9 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Simon Hudon
 -/
 import Mathlib.Control.Traversable.Lemmas
+import Lean.Elab.Match
```
grunweg, line 7: I'm assuming these were transitively imported before?

grunweg (commented):
One small question - otherwise, looks good to me. Thanks for doing this!
maintainer merge

Mathlib/Tactic/DeriveTraversable.lean
jcommelin: Yeah, this is the result of shake doing its thing

PR #14433, opened 2024-07-05
[Merged by Bors] - feat(MetricSpace/Ultra): IsUltrametricDist and basic facts on open/closed sets in such spaces

pechersky:
TODO in future PRs
- add "all triangles are isosceles" in such normed spaces
- totally disconnected space
- give instances of this to Z_p, Q_p, etc

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14768, opened 2024-07-15
[Merged by Bors] - feat(Analysis/Normed/{Group,Field}/Ultra): Nonarchimedean norms from ultrametrics

pechersky:
---

- [x] depends on: #14433
- [x] depends on: #15271
- [x] depends on: #15284


Mathlib/Topology/MetricSpace/Ultra/TotallyDisconnected.lean
```diff
@@ -0,0 +1,33 @@
...
+    intro t _ ht' y hy z hz
+    contrapose! ht'
+    intro H
+    specialize H (ball y (dist y z / 2)) (ball y (dist y z / 2))ᶜ
+      isOpen_ball (isClosed_ball _ _).isOpen_compl
+    simp only [Set.union_compl_self, Set.subset_univ, Set.inter_compl_self, Set.inter_empty,
+      Set.not_nonempty_empty, imp_false, true_implies] at H
+    refine H ⟨y, hy, mem_ball.mpr ?_⟩ ⟨z, hz, ?_⟩
+    · simpa using ht'
+    · simp [dist_comm, dist_nonneg]
```
loefflerd, line 33: I find this proof a bit hard to read, not least because of starting a big tactic block inside the argument to a lemma  Can you rephrase it slightly and maybe add a comment or two to explain what's going on? For a start, you can get rid of the first few lines by starting with `rw [totallyDisconnectedSpace_iff]`, i.e.
```
instance {X : Type*} [MetricSpace X] [IsUltrametricDist X] : TotallyDisconnectedSpace X := by
  rw [totallyDisconnectedSpace_iff]
  intro t _ ht' y hy z hz
  contrapose! ht'
  intro H
  specialize H (ball y (dist y z / 2)) (ball y (dist y z / 2))ᶜ
    isOpen_ball (isClosed_ball _ _).isOpen_compl
  simp only [Set.union_compl_self, Set.subset_univ, Set.inter_compl_self, Set.inter_empty,
    Set.not_nonempty_empty, imp_false, true_implies] at H
  refine H ⟨y, hy, mem_ball.mpr ?_⟩ ⟨z, hz, ?_⟩
  · simpa using ht'
  · simp [dist_comm, dist_nonneg]
```
works, but further simplification is surely possible.

It feels to me like the "core" of this argument can be factored out, something like "a space is totally disconnected if for any distinct x, y there is a clopen containing x but not y". Maybe it would be useful to make this explicit.

loefflerd (commented):
I've left some comments, focussing on the last couple of files (which is where there is most overlap with my own PR). 

I wonder if it would be best to separate this out into two PR's: one for the bits which don't involve suprema (which are much more self-contained, localised in just two files, and which I would be happy to recommend merging very soon once the comments above are dealt with) and a second one for the results about finite sums / products, which involve lots of lemmas spread across many files and where I feel we should think rather more carefully about how best to formulate things.

Mathlib/Topology/MetricSpace/Ultra/TotallyDisconnected.lean
loefflerd: > It feels to me like the "core" of this argument can be factored out, something like "a space is totally disconnected if (iff?) for any distinct x, y there is a clopen containing x but not y". Maybe it would be useful to make this explicit.

Wait, that is in the library already. So one can golf this down to
```lean
instance {X : Type*} [MetricSpace X] [IsUltrametricDist X] : TotallyDisconnectedSpace X := by
  refine (totallyDisconnectedSpace_iff X).mpr (isTotallyDisconnected_of_isClopen_set fun x y h ↦ ?_)
  obtain ⟨r, hr, hr'⟩ := exists_between (dist_pos.mpr h)
  refine ⟨_, IsUltrametricDist.isClopen_ball x r, ?_, ?_⟩
  · simp only [mem_ball, dist_self, hr]
  · simp only [mem_ball, dist_comm, not_lt, hr'.le]
```
(Note the use of "exists_between" rather than explicitly halving the distance, this is a useful trick.)

Mathlib/Topology/MetricSpace/Ultra/Normed.lean
```diff
@@ -0,0 +1,346 @@
...
+@[to_additive]
+lemma isUltrametricDist_of_forall_nnnorm_mul_le_max_nnnorm
+    (h : ∀ x y : S', ‖x * y‖₊ ≤ max ‖x‖₊ ‖y‖₊) : IsUltrametricDist S' :=
+  isUltrametricDist_of_forall_norm_mul_le_max_norm h
+
+lemma isUltrametricDist_of_isNonarchimedean_nnnorm {S' : Type*} [SeminormedAddGroup S']
+    (h : IsNonarchimedean ((↑) ∘ (nnnorm : S' → ℝ≥0))) : IsUltrametricDist S' :=
+  isUltrametricDist_of_forall_nnnorm_add_le_max_nnnorm h
+
+@[to_additive]
```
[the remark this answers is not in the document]
pechersky: I hope to use these lemmas when discussing norms of coefficients of power series, where there is an obvious order, and thus List is appropriate.

I totally understand the pain around `iSup` -- but it is the best way of expressing the element of the collection that is the max across it. And I have the associated lemmas in the other file proving this element is in the collection. These lemmas give the strongest result, against which all existing `(c)iSup` lemmas can be used. And I have the `Finset.sup` versions as well. Proving the list lemma from the finset lemma would have to go through a deduplication-doesn't-affect-result lemma which I don't currently have.

In any case, I'll extract out these lemmas into a separate PR.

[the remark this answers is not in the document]
pechersky: Oh! And one more reason to have it over `List` -- no commutativity means the `prod` result can only be stated about `List`.

Mathlib/Topology/MetricSpace/Ultra/Normed.lean
```diff
@@ -0,0 +1,346 @@
+/-
+Copyright (c) 2024 Yakov Pechersky. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yakov Pechersky
+-/
+import Mathlib.Analysis.Normed.Field.Basic
```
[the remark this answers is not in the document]
pechersky: Unfortunately, that's where `NormedDivisionRing` is defined.

Mathlib/Topology/MetricSpace/Ultra/TotallyDisconnected.lean
pechersky: Will move to a separate PR.

pechersky: #17297

pechersky:
@loefflerd is there a reason you marked this as blocked by the totally disconnected PR? they seem orthogonal.

loefflerd:
> @loefflerd is there a reason you marked this as blocked by the totally disconnected PR? they seem orthogonal.

Yes, there's a reason – just not a good reason! Sorry; I mistakenly thought this PR included the other one. I'll remove the flag.

Mathlib/Analysis/Normed/Group/Ultra.lean
```diff
@@ -0,0 +1,175 @@
...
+lemma _root_.Finset.Nonempty.norm_prod_le_sup'_norm {s : Finset ι} (hs : s.Nonempty) (f : ι → M) :
+    ‖∏ i ∈ s, f i‖ ≤ s.sup' hs (‖f ·‖) := by
+  simp only [Finset.le_sup'_iff]
+  refine Finset.Nonempty.cons_induction ?_ ?_ hs
+  · simp
+  rintro a s ha _ ⟨b, hb, IH⟩
+  simp only [Finset.prod_cons, Finset.mem_cons, exists_eq_or_imp]
+  refine (le_total ‖∏ i ∈ s, f i‖ ‖f a‖).imp ?_ ?_ <;> intro h
+  · exact (norm_mul_le_max _ _).trans (by simp [h])
+  · exact ⟨b, hb, (norm_mul_le_max _ _).trans (by simpa [h] using IH)⟩
```
loefflerd, line 148: You can write this in a more "lean4" syntax as follows:
```lean
lemma _root_.Finset.Nonempty.norm_prod_le_sup'_norm {s : Finset ι} (hs : s.Nonempty) (f : ι → M) :
    ‖∏ i ∈ s, f i‖ ≤ s.sup' hs (‖f ·‖) := by
  simp only [Finset.le_sup'_iff]
  induction hs using Finset.Nonempty.cons_induction with
  | singleton j => simp only [Finset.mem_singleton, Finset.prod_singleton, exists_eq_left, le_refl]
  | cons j t hj _ IH =>
      simp only [Finset.prod_cons, Finset.mem_cons, exists_eq_or_imp]
      rcases le_total ‖∏ i ∈ t, f i‖ ‖f j‖ with h | h
      · exact .inl <| (norm_mul_le_max _ _).trans (max_eq_left h).le
      · rcases IH with ⟨b, hb, IH⟩
        exact .inr ⟨b, hb, (norm_mul_le_max _ _).trans <| (max_eq_right h).le.trans IH⟩
```
(One line longer than the old version, since there's no equivalent of `rintro` in the new induction syntax... 🏌️ )

Mathlib/Analysis/Normed/Group/Ultra.lean
```diff
@@ -0,0 +1,175 @@
...
+
+/-- Nonarchimedean norm of a product is less than or equal the norm of any term in the product. -/
+@[to_additive "Nonarchimedean norm of a sum is less than or equal the norm of any term in the sum."]
```
loefflerd, line 151: This sounds like you're proving `≤ Inf`, not `≤ Sup` ("any" is annoyingly ambiguous); rephrase?

Mathlib/Analysis/Normed/Group/Ultra.lean
```diff
@@ -0,0 +1,175 @@
...
+  rw [← Finset.sup'_eq_sup hs, ← NNReal.coe_le_coe,
+    Finset.comp_sup'_eq_sup'_comp hs toReal coe_mono.map_sup]
+  simpa only [coe_nnnorm', Function.comp_apply, Finset.le_sup'_iff] using
+    hs.norm_prod_le_sup'_norm f
+
+/-- Nonarchimedean norm of a product is less than or equal the norm of any term in the product. -/
+@[to_additive "Nonarchimedean norm of a sum is less than or equal the norm of any term in the sum."]
+lemma _root_.Fintype.nnnorm_prod_le_sup_univ_norm [Fintype ι] (f : ι → M) :
+    ‖∏ i, f i‖₊ ≤ Finset.univ.sup (‖f ·‖₊) := by
+  simpa using Finset.univ.nnnorm_prod_le_sup_nnnorm f
```
loefflerd, line 165: Are this lemma and the next needed? By definition, `∏ x, f x` is just syntactic sugar for `Finset.prod Finset.univ f`, so these lemmas are special cases of the ones above. (If you squeeze the simpa, it's not actually using any lemmas at all.)

Mathlib/Analysis/Normed/Group/Ultra.lean
```diff
@@ -0,0 +1,175 @@
...
+/-- Nonarchimedean norm of a product is less than or equal the norm of any term in the product. -/
+@[to_additive "Nonarchimedean norm of a sum is less than or equal the norm of any term in the sum."]
+lemma _root_.Finset.nnnorm_prod_le_sup_nnnorm (s : Finset ι) (f : ι → M) :
+    ‖∏ i ∈ s, f i‖₊ ≤ s.sup (‖f ·‖₊) := by
+  rcases s.eq_empty_or_nonempty with rfl|hs
+  · simp
+  rw [← Finset.sup'_eq_sup hs, ← NNReal.coe_le_coe,
+    Finset.comp_sup'_eq_sup'_comp hs toReal coe_mono.map_sup]
+  simpa only [coe_nnnorm', Function.comp_apply, Finset.le_sup'_iff] using
+    hs.norm_prod_le_sup'_norm f
```
loefflerd, line 159: slight cleanup / golf:
```suggestion
  rcases s.eq_empty_or_nonempty with rfl | hs
  · simp only [Finset.prod_empty, nnnorm_one', Finset.sup_empty, bot_eq_zero', le_refl]
  · simpa only [← Finset.sup'_eq_sup hs, Finset.le_sup'_iff, coe_le_coe, coe_nnnorm']
      using hs.norm_prod_le_sup'_norm f
```

Mathlib/Analysis/Normed/Group/Ultra.lean
```diff
@@ -0,0 +1,175 @@
...
+lemma _root_.Fintype.nnnorm_prod_le_sup_univ_norm [Fintype ι] (f : ι → M) :
+    ‖∏ i, f i‖₊ ≤ Finset.univ.sup (‖f ·‖₊) := by
+  simpa using Finset.univ.nnnorm_prod_le_sup_nnnorm f
+
+/-- Nonarchimedean norm of a product is less than or equal the norm of any term in the product. -/
+@[to_additive "Nonarchimedean norm of a sum is less than or equal the norm of any term in the sum."]
+lemma _root_.Fintype.norm_prod_le_sup'_univ_norm [Nonempty ι] [Fintype ι] (f : ι → M) :
+    ‖∏ i, f i‖ ≤ Finset.univ.sup' Finset.univ_nonempty (‖f ·‖) := by
+  simpa using Finset.univ_nonempty.norm_prod_le_sup'_norm f
+
```
loefflerd, line 172:
```suggestion

/--
Generalised ultrametric triangle inequality for nonempty finite products in groups with an
ultrametric norm.
-/
@[to_additive "Generalised ultrametric triangle inequality for nonempty finite sums in additive
groups with an ultrametric norm ."]
lemma norm_prod_le_of_forall_le_of_nonempty {s : Finset ι} (hs : s.Nonempty) {f : ι → M} {C : ℝ}
    (hC : ∀ i ∈ s, ‖f i‖ ≤ C) : ‖∏ i ∈ s, f i‖ ≤ C :=
  (hs.norm_prod_le_sup'_norm f).trans (Finset.sup'_le hs _ hC)

/--
Generalised ultrametric triangle inequality for finite products in groups with an ultrametric
norm.
-/
@[to_additive "Generalised ultrametric triangle inequality for finite sums in additive groups
with an ultrametric norm."]
lemma norm_prod_le_of_forall_le_of_nonneg {s : Finset ι} {f : ι → M} {C : ℝ}
    (h_nonneg : 0 ≤ C) (hC : ∀ i ∈ s, ‖f i‖ ≤ C) : ‖∏ i ∈ s, f i‖ ≤ C := by
  lift C to NNReal using h_nonneg
  exact (s.nnnorm_prod_le_sup_nnnorm f).trans (Finset.sup_le <| mod_cast hC)
  
```

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,63 @@
...
+
+lemma norm_add_one_le_max_norm_one (x : K) :
+    ‖x + 1‖ ≤ max ‖x‖ 1 := by
+  simpa using norm_add_le_max x 1
+
+lemma nnnorm_add_one_le_max_nnnorm_one (x : K) :
+    ‖x + 1‖₊ ≤ max ‖x‖₊ 1 := by
+  simpa using norm_add_le_max x 1
+
+lemma nnnorm_natCast_le_one (n : ℕ) :
```
loefflerd, line 39: In these last four lemmas, `K` should be an explicit argument.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,63 @@
...
+
+variable {K : Type*} [NormedDivisionRing K] [IsUltrametricDist K]
+
+lemma norm_add_one_le_max_norm_one (x : K) :
+    ‖x + 1‖ ≤ max ‖x‖ 1 := by
+  simpa using norm_add_le_max x 1
+
+lemma nnnorm_add_one_le_max_nnnorm_one (x : K) :
+    ‖x + 1‖₊ ≤ max ‖x‖₊ 1 := by
+  simpa using norm_add_le_max x 1
```
loefflerd, line 37: `simpa using` isn't using anything here.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,63 @@
...
+lemma nnnorm_natCast_le_one (n : ℕ) :
+    ‖(n : K)‖₊ ≤ 1 := by
+  induction n with
+  | zero => simp
+  | succ n hn => simpa [hn] using nnnorm_add_one_le_max_nnnorm_one (n : K)
+
+lemma norm_natCast_le_one (n : ℕ) :
+    ‖(n : K)‖ ≤ 1 := by
+  rw [← Real.toNNReal_le_toNNReal_iff zero_le_one]
+  simpa using nnnorm_natCast_le_one n
```
loefflerd, line 48: the rewrites and simpas aren't needed, `:= nnnorm_natCast_le_one n` works.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,63 @@
...
+lemma nnnorm_intCast_le_one (z : ℤ) :
+    ‖(z : K)‖₊ ≤ 1 := by
+  induction z
+  · simpa using nnnorm_natCast_le_one _
+  · simpa only [Int.cast_negSucc, Nat.cast_one, nnnorm_neg] using nnnorm_natCast_le_one _
+
+lemma norm_intCast_le_one (z : ℤ) :
+    ‖(z : K)‖ ≤ 1 := by
+  rw [← Real.toNNReal_le_toNNReal_iff zero_le_one]
+  simpa using nnnorm_intCast_le_one z
```
loefflerd, line 59: ... and again here.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,63 @@
...
+
+ultrametric, nonarchimedean
+-/
+open Metric NNReal
+
+namespace IsUltrametricDist
+
+section DivisionRing
+
+variable {K : Type*} [NormedDivisionRing K] [IsUltrametricDist K]
```
loefflerd, line 29:
```suggestion
variable {K : Type*}  [SeminormedRing K] [NormOneClass K] [IsUltrametricDist K]
```
All the lemmas work fine in this more general context (meaning they apply to `ℤ_[p]`). This will of course mean renaming the file, and moving it to `Analysis/Normed/Ring`.

Mathlib/Analysis/Normed/Group/Ultra.lean
loefflerd: (feel free to ignore this if you prefer and I'll keep these lemmas in the other PR instead).

Mathlib/Analysis/Normed/Field/Ultra.lean
pechersky: How so? It needs to call `(nn)norm_one`.

pechersky: I'll use defeq with the norm lemma.

Mathlib/Analysis/Normed/Group/Ultra.lean
pechersky: We have a variety of lemmas that match the Fintype prod/sum syntactically, like `Fintype.prod_sum`. I make these basic lemmas to make life easier for the downstream consumer.

Mathlib/Analysis/Normed/Ring/Ultra.lean
```diff
@@ -0,0 +1,71 @@
...
+
+* `norm_intCast_le_one`:
+  the norm of the image of an integer in the division ring is always less than or equal to one
+
+## Implementation details
+
+A `[NormedRing R]` only assumes a submultiplicative norm. The weakest ring-like structure
+that has a bundled multiplicative norm is `[NormedDivisionRing K]`.
+Since the statements below hold in a multiplicative norm context, we can state them
+in an unbundled fashion using `[NormOneClass R]`.
```
loefflerd, line 25: I don't agree with the statements in this docstring. The lemmas all hold _without_ assuming any multiplicativity, or even submultiplicativity, of the norm; in fact they don't use the existence of multiplication on R at all!

In fact one can actually prove all these lemmas only assuming `{R : Type*} [SeminormedAddGroup R] [One R] [NormOneClass R] [IsUltrametricDist R]`. But one has to give the typeclass machinery a little help in order to get it to recognise that there is a coercion from N or Z to R. I'm not sure this extra generality is worth the bother, so I guess it's easiest to use the weakest pre-existing typeclass that implies both `[SeminormedAddGroup R]` and `[AddGroupWithOne R]`. But division structure is certainly irrelevant.

Mathlib/Analysis/Normed/Group/Ultra.lean
loefflerd: I feel that's a slightly different situation. For `Fintype.prod_sum`, the more general lemmas `Finset.prod_univ_sum` and `Finset.prod_sum` are both slightly more awkward to state than the fintype version, so if you used the finset lemmas applied to `Finset.univ`, you would be left with a `Finset.attach` or `Fintype.piFinset` to be tidied away afterwards. In the case at hand, the statement of the Fintype lemma is literally identical to the Finset lemma applied to `univ` (even at the very weak transparency level used by `rw`); so I don't think there's any actual gain for the downstream consumer.

Mathlib/Analysis/Normed/Ring/Ultra.lean
pechersky: Fair, I'll modify the contents here to discuss NormOneClass without mentioning multiplicativity. However, it's my understanding that one doesn't really get NormOneClass on its own that often -- that instead, it occurs on rings that _do_ have multiplicative norms, like the examples discussed (Z_p). So I wanted to give a connection to that and explain why NormedRing isn't appropriate here, even though it has Norm and Ring in the name.

pechersky: I guess the important counterexample for my assertion about multiplicativity and NormOneClass are CLMs.

pechersky:
@loefflerd Thank you for all the feedback! It's clarifying and brings a lot of polish.

loefflerd:
You're welcome, it's been a pleasure to work together on this. Let's declare it fully polished now :smile:

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #14576, opened 2024-07-09
[Merged by Bors] - feat(MeasureSpace): add `ae_smul_measure_eq`

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14772, opened 2024-07-15
[Merged by Bors] - chore(Function/EssSup): use `ae_smul_measure_eq`

urkud:
---

- [ ] depends on: #14576


RemyDegenne:
bors r+
Thanks!

PR #14781, opened 2024-07-15
[Merged by Bors] - feat(Order/Zorn): change zorn to use Maximal API

apnelson1:
We change statements of Zorn's lemma to use the new `Maximal` API from #14721. The new definitions eliminate the need for separate `PartialOrder` and `Preorder` versions of Zorn's lemma, and slightly simplify most of its applications. 

The new statements of Zorn aren't defeq to the old ones, so a good number of files are touched, but very few theorem statements outside `Order.Zorn` itself are actually changed - most changes are small modifications of existing proofs. 

- [x] depends on: #14721

---


Mathlib/Analysis/Convex/StoneSeparation.lean
```diff
@@ -78,14 +78,16 @@ theorem not_disjoint_segment_convexHull_triple {p q u v x y z : E} (hz : z ∈ s
 theorem exists_convex_convex_compl_subset (hs : Convex 𝕜 s) (ht : Convex 𝕜 t) (hst : Disjoint s t) :
     ∃ C : Set E, Convex 𝕜 C ∧ Convex 𝕜 Cᶜ ∧ s ⊆ C ∧ t ⊆ Cᶜ := by
   let S : Set (Set E) := { C | Convex 𝕜 C ∧ Disjoint C t }
-  obtain ⟨C, hC, hsC, hCmax⟩ :=
+  -- obtain ⟨C, hC, hsC, hCmax⟩ :=
```
j-loreaux, line 81:
```suggestion
```

Mathlib/Data/Matroid/IndepAxioms.lean
```diff
@@ -162,6 +162,7 @@ namespace IndepMatroid
     obtain ⟨e, heI, hins⟩ := exists_insert_of_not_maximal indep_subset hI hImax
     by_cases heB : e ∈ B
     · exact ⟨e, ⟨heB, heI⟩, hins⟩
+
```
j-loreaux, line 165: no need to unnecessarily increase the diff
```suggestion
```

Mathlib/Data/Set/Basic.lean
```diff
@@ -199,6 +199,11 @@ theorem forall_in_swap {p : α → β → Prop} : (∀ a ∈ s, ∀ (b), p a b)
 theorem mem_setOf {a : α} {p : α → Prop} : a ∈ { x | p x } ↔ p a :=
   Iff.rfl
 
+/-- This lemma is intended for use with `rw` where a membership predicate is needed,
+hence the explicit argument and the equality in the reverse direction from normal.
+See also `Set.mem_setOf_eq` for the reverse direction applied to an argument. -/
+theorem eq_mem_setOf (p : α → Prop) : p = (· ∈ {a | p a}) := rfl
```
j-loreaux, line 205: can you merge master so that this doesn't show up in the diff?

Mathlib/MeasureTheory/Covering/Vitali.lean
```diff
@@ -74,7 +74,8 @@ theorem exists_disjoint_subfamily_covering_enlargment (B : ι → Set α) (t : S
   let T : Set (Set ι) := { u | u ⊆ t ∧ u.PairwiseDisjoint B ∧
     ∀ a ∈ t, ∀ b ∈ u, (B a ∩ B b).Nonempty → ∃ c ∈ u, (B a ∩ B c).Nonempty ∧ δ a ≤ τ * δ c }
   -- By Zorn, choose a maximal family in the good set `T` of disjoint families.
-  obtain ⟨u, uT, hu⟩ : ∃ u ∈ T, ∀ v ∈ T, u ⊆ v → v = u := by
+
```
j-loreaux, line 77:
```suggestion
```

Mathlib/MeasureTheory/Covering/Vitali.lean
```diff
@@ -85,17 +86,19 @@ theorem exists_disjoint_subfamily_covering_enlargment (B : ι → Set α) (t : S
     obtain ⟨c, cu, ac, hc⟩ : ∃ c, c ∈ u ∧ (B a ∩ B c).Nonempty ∧ δ a ≤ τ * δ c :=
       (UT uU).2.2 a hat b hbu hab
     exact ⟨c, ⟨u, uU, cu⟩, ac, hc⟩
+  obtain ⟨u, hu⟩ := hzorn
+
```
j-loreaux, line 90:
```suggestion
```

Mathlib/MeasureTheory/Measure/Haar/Unique.lean
```diff
@@ -709,7 +709,8 @@ theorem measure_isHaarMeasure_eq_smul_of_isEverywherePos [LocallyCompactSpace G]
     exact ⟨closure k, hk.closure, isClosed_closure, mem_of_superset hmem subset_closure⟩
   have one_k : 1 ∈ k := mem_of_mem_nhds k_mem
   let A : Set (Set G) := {t | t ⊆ s ∧ PairwiseDisjoint t (fun x ↦ x • k)}
-  obtain ⟨m, mA, m_max⟩ : ∃ m ∈ A, ∀ a ∈ A, m ⊆ a → a = m := by
+  obtain ⟨m, m_max⟩ : ∃ m, Maximal (· ∈ A) m := by
+  -- obtain ⟨m, mA, m_max⟩ : ∃ m ∈ A, ∀ a ∈ A, m ⊆ a → a = m := by
```
j-loreaux, line 713:
```suggestion
```

Mathlib/Order/PrimeSeparator.lean
```diff
@@ -63,7 +63,8 @@ theorem prime_ideal_of_disjoint_filter_ideal (hFI : Disjoint (F : Set α) (I : S
             ⟨le_trans (hcS hJ).2.1 (le_sSup hJ), fun J hJ ↦ (hcS hJ).2.2⟩⟩
 
   -- Thus, by Zorn's lemma, we can pick a maximal ideal J in S.
-  obtain ⟨Jset, ⟨Jidl, IJ, JF⟩, ⟨_, Jmax⟩⟩ := zorn_subset_nonempty S chainub I IinS
+  obtain ⟨Jset, _, hmax⟩ := zorn_subset_nonempty S chainub I IinS
+  obtain ⟨Jidl, IJ, JF⟩ := hmax.prop
```
j-loreaux, line 67: I see this pattern a lot in this PR. Should we just have an additional lemma besides `zorn_subset_nonempty` that does the second step deconstruction too?

Mathlib/Order/PrimeSeparator.lean
```diff
@@ -94,8 +95,9 @@ theorem prime_ideal_of_disjoint_filter_ideal (hFI : Disjoint (F : Set α) (I : S
   have J₂J : ↑J₂ ≠ Jset := ne_of_mem_of_not_mem' a₂J₂ ha₂
 
   -- Therefore, since J is maximal, we must have Jᵢ ∉ S.
-  have J₁S : ↑J₁ ∉ S := fun h => J₁J (Jmax J₁ h (le_sup_left : J ≤ J₁))
-  have J₂S : ↑J₂ ∉ S := fun h => J₂J (Jmax J₂ h (le_sup_left : J ≤ J₂))
+
```
j-loreaux, line 98:
```suggestion
```

Mathlib/Order/Zorn.lean
```diff
@@ -95,98 +98,67 @@ section Preorder
...
 
-theorem zorn_preorder (h : ∀ c : Set α, IsChain (· ≤ ·) c → BddAbove c) :
-    ∃ m : α, ∀ a, m ≤ a → a ≤ m :=
+theorem zorn_le (h : ∀ c : Set α, IsChain (· ≤ ·) c → BddAbove c) : ∃ m : α, IsMax m :=
   exists_maximal_of_chains_bounded h le_trans
 
-theorem zorn_nonempty_preorder [Nonempty α]
-    (h : ∀ c : Set α, IsChain (· ≤ ·) c → c.Nonempty → BddAbove c) : ∃ m : α, ∀ a, m ≤ a → a ≤ m :=
+theorem zorn_le_nonempty [Nonempty α]
+    (h : ∀ c : Set α, IsChain (· ≤ ·) c → c.Nonempty → BddAbove c) : ∃ m : α, IsMax m :=
```
j-loreaux, line 105: I can see the appeal of leaving it in `IsMax` form to access API, but it felt like most of the time this condition was deconstructed in the proof anyway. Should we at least have both versions?

Mathlib/Topology/Algebra/Semigroup.lean
```diff
@@ -43,16 +44,16 @@ theorem exists_idempotent_of_compact_t2_of_continuous_mul_left {M} [Nonempty M]
...
       · refine ⟨N_closed.inter ((T1Space.t1 m).preimage (continuous_mul_left m)), ?_, ?_⟩
         · rwa [← scaling_eq_self] at hm
         · rintro m'' ⟨mem'', eq'' : _ = m⟩ m' ⟨mem', eq' : _ = m⟩
           refine ⟨N_mul _ mem'' _ mem', ?_⟩
           rw [Set.mem_setOf_eq, mul_assoc, eq', eq'']
       apply Set.inter_subset_left
-    -- Thus `m * m = m` as desired.
     rw [← absorbing_eq_self] at hm
     exact hm.2
+
```
j-loreaux, line 56: And I'm not sure why the comment line two lines above this got deleted.
```suggestion
```

j-loreaux (commented):
Looks relatively reasonable, but here are some comments.

Mathlib/Order/Zorn.lean
apnelson1: It's more the abstraction unifying `PartialOrder` and `Preorder` that I thought was an improvement. Does two versions mean three versions?

j-loreaux: Are there places where you are actually making use of `IsMax` API? Or are you always destructing it? If the latter, I suggest switching it back to the forall statement because it means it's just getting in the way.

apnelson1: My point is that there are two possible forall statements - the one with `le -> le` for preorders and the other with `le -> eq` for partial orders. The first is more general but less convenient and the second vice versa. Using `IsMax` abstracts away the problem, and stops us having to pick one of the two in theorem statements.

j-loreaux: Oh, I completely misunderstood! Thanks for the explanation. This is great as is then!

j-loreaux:
LGTM now.

maintainer merge

mattrobball:
!bench

mattrobball:
bors merge

PR #14785, opened 2024-07-16
[Merged by Bors] - chore: remove superfluous `ge_iff_le`

mo271:
These were likely introduced by calling simp?, but are not necessary.

This is fixed in https://github.com/leanprover/lean4/pull/4567, so now might be a good time for cleaning them up.

---

### DIFF
--- a/Archive/Imo/Imo2008Q2.lean
+++ b/Archive/Imo/Imo2008Q2.lean
@@ -53,7 +53,7 @@ theorem imo2008_q2a (x y z : ℝ) (h : x * y * z = 1) (hx : x ≠ 1) (hy : y ≠
   have hn_ne_zero : n ≠ 0 := by contrapose! hx; field_simp; assumption
   have hmn_ne_zero : m + n ≠ 0 := by contrapose! hz; field_simp; linarith
   have hc_sub_sub : c - (c - m - n) = m + n := by abel
-  rw [← sub_nonneg]
+  rw [ge_iff_le, ← sub_nonneg]
   convert sq_nonneg ((c * (m ^ 2 + n ^ 2 + m * n) - m * (m + n) ^ 2) / (m * n * (m + n)))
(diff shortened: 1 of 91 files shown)


Mathlib/Data/List/Chain.lean
```diff
@@ -453,9 +453,7 @@ theorem Chain'.cons_of_le [LinearOrder α] {a : α} {as m : List α}
         apply hmas
         apply le_of_lt
         exact (List.lt_iff_lex_lt _ _).mp (List.lt.head _ _ hh)
-      · simp only [List.cons.injEq] at hmas
-        rw [ge_iff_le, le_iff_lt_or_eq]
-        exact Or.inr hmas.1
+      · simp_all only [List.cons.injEq, le_refl]
```
mo271, line 456: this is the only place, where I manually simplified the proof further. All others are regular expression replacements.

Archive/Imo/Imo2008Q2.lean
```diff
@@ -53,7 +53,7 @@ theorem imo2008_q2a (x y z : ℝ) (h : x * y * z = 1) (hx : x ≠ 1) (hy : y ≠
   have hn_ne_zero : n ≠ 0 := by contrapose! hx; field_simp; assumption
   have hmn_ne_zero : m + n ≠ 0 := by contrapose! hz; field_simp; linarith
   have hc_sub_sub : c - (c - m - n) = m + n := by abel
-  rw [ge_iff_le, ← sub_nonneg]
+  rw [← sub_nonneg]
```
Ruben-VandeVelde, line 56: This one is actually necessary, because the theorem statement incorrectly uses ge

mo271: thanks, added it back

riccardobrasca:
Nice!

bors merge

PR #14789, opened 2024-07-16
[Merged by Bors] - refactor(GroupTheory/PGroup,Sylow): Replace Fintype.card with Nat.card

tb65536:
A lot of the group theory library still needs to be switched over from Fintype.card to Nat.card. This PR switches over GroupTheory/PGroup.lean and GroupTheory/Sylow (there was enough dependence between the two that it was simplest to do them together).

---

### DIFF
--- a/Mathlib/GroupTheory/Perm/Cycle/Type.lean
+++ b/Mathlib/GroupTheory/Perm/Cycle/Type.lean
@@ -470,6 +470,7 @@ theorem rotate_length : rotate v n = v :=
 
 end VectorsProdEqOne
 
+-- TODO: Make make the `Finite` version of this theorem the default
 /-- For every prime `p` dividing the order of a finite group `G` there exists an element of order
 `p` in `G`. This is known as Cauchy's theorem. -/
 theorem _root_.exists_prime_orderOf_dvd_card {G : Type*} [Group G] [Fintype G] (p : ℕ)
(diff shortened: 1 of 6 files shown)


Mathlib/GroupTheory/Perm/Cycle/Type.lean
```diff
@@ -514,6 +514,15 @@ theorem _root_.exists_prime_addOrderOf_dvd_card {G : Type*} [AddGroup G] [Fintyp
 
 attribute [to_additive existing] exists_prime_orderOf_dvd_card
 
+/-- For every prime `p` dividing the order of a finite group `G` there exists an element of order
+`p` in `G`. This is known as Cauchy's theorem. -/
+@[to_additive]
+theorem _root_.exists_prime_orderOf_dvd_card' {G : Type*} [Group G] [Finite G] (p : ℕ)
```
jcommelin, line 520: Should this eventually become the default? If so, then please leave a TODO on this decl and `exists_prime_orderOf_dvd_card`.

Mathlib/GroupTheory/Sylow.lean
```diff
@@ -309,36 +303,32 @@ instance [hp : Fact p.Prime] [Finite (Sylow p G)] : IsPretransitive G (Sylow p G
...
       exact ⟨H.mpr rfl, fun R h => Subtype.ext (Sylow.ext (H.mp h))⟩⟩
 
 variable (p) (G)
 
 /-- A generalization of **Sylow's third theorem**.
   If the number of Sylow `p`-subgroups is finite, then it is congruent to `1` modulo `p`. -/
-theorem card_sylow_modEq_one [Fact p.Prime] [Fintype (Sylow p G)] :
-    card (Sylow p G) ≡ 1 [MOD p] := by
+theorem card_sylow_modEq_one [Fact p.Prime] [Finite (Sylow p G)] :
+    Nat.card (Sylow p G) ≡ 1 [MOD p] := by
```
jcommelin, line 319: Is the file already in a shape where `open Fintype` can be removed?

Mathlib/GroupTheory/Perm/Cycle/Type.lean
tb65536: Sure, sounds good.

Mathlib/GroupTheory/Sylow.lean
tb65536: Yes (whereas PGroup still has an annoying proof that benefits from `open Fintype`).

jcommelin:
Thanks :tada:

bors merge

PR #14793, opened 2024-07-16
[Merged by Bors] - feat(Algebra/Homology): the Ext class of a short exact sequence

joelriou:
In this PR, we construct the class in `Ext S.X₃ S.X₁ 1` that is attached to a short exact short complex `S` in an abelian category.

---


Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
```diff
@@ -303,8 +317,107 @@ lemma homAddEquiv_apply (α : Ext X Y n) : homAddEquiv α = α.hom := rfl
...
+    dsimp [W, S']
+    infer_instance
+  have : HasSmallLocalizedHom.{w} W S'.X₃ K := by
+    rw [Localization.hasSmallLocalizedHom_iff_target W S'.X₃ w hw]
+    dsimp [W, S']
+    apply Localization.hasSmallLocalizedHom_of_hasSmallLocalizedShiftedHom₀ (M := ℤ)
+  have : HasSmallLocalizedShiftedHom.{w} W ℤ K S'.X₁ := by
+    rw [Localization.hasSmallLocalizedShiftedHom_iff_source.{w} W ℤ w hw S'.X₁]
+    dsimp [W, S']
+    infer_instance
```
erdOne, line 369: Is it possible to promote these to instances?

Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
```diff
@@ -303,8 +317,107 @@ lemma homAddEquiv_apply (α : Ext X Y n) : homAddEquiv α = α.hom := rfl
...
+/-- The class in `Ext S.X₃ S.X₁ 1` that is attached to a short exact
+short complex `S` in an abelian category. -/
+noncomputable def extClass : Ext.{w} S.X₃ S.X₁ 1 := by
+  let W := (HomologicalComplex.quasiIso C (ComplexShape.up ℤ))
+  let S' := S.map (CochainComplex.singleFunctor C 0)
+  have hS' : S'.ShortExact := hS.map_of_exact (HomologicalComplex.single _ _ _)
+  let K := CochainComplex.mappingCone S'.f
+  let w : K ⟶ S'.X₃ := CochainComplex.mappingCone.descShortComplex S'
+  have hw : W w := CochainComplex.mappingCone.quasiIso_descShortComplex hS'
+  let δ : K ⟶ S'.X₁⟦(1 : ℤ)⟧ := (CochainComplex.mappingCone.triangle S'.f).mor₃
```
erdOne, line 358: Is it possible to use local notations instead? Then you wouldn't need to copy-paste these in every proof.

Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
joelriou: No, we cannot, because the proofs depend on a property of `S'` (i.e. that it is short-exact).

Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
joelriou: Thanks for the suggestion, which I shall apply soon.

semorrison:
Perhaps the new material belongs in a new file? Happy to defer here.

joelriou:
> Perhaps the new material belongs in a new file? Happy to defer here.

Thanks for the suggestion. `Ext` becomes `Ext.Basic`, and the definition of the class in `Ext^1` is in the file `Ext.ExtClass`.

jcommelin (commented):
Thanks :tada:

bors merge

PR #14794, opened 2024-07-16
[Merged by Bors] - chore(scripts/create-adaptation-pr.sh): make script robust against changes on disk

jcommelin:
This script needs to change branches regularly.
By default shell scripts are not robust against changes to the script on disk.
So if different branches have different versions of the script,
unexpected things may happen.

By wrapping the script in

    {
    # script content
    exit
    }

we protect against such problems.

---


semorrison:
bors merge

PR #14799, opened 2024-07-16
feat(Analysis/MeanInequalities): Weighted QM-AM inequality

luigi-massacci:
The inequality between the quadratic and arithmetic mean as a consequence of Jensen's inequality. Weighted version for positive real valued functions from a `Finset`. Fix notation for sums, switching to `∈` in place of `in`. 

---

### DIFF
--- a/Mathlib/Analysis/MeanInequalities.lean
+++ b/Mathlib/Analysis/MeanInequalities.lean
@@ -326,7 +326,6 @@ section ArithMeanLEQuadMean
 
 namespace Real
 
-
 /-- **AM-QM inequality**: The arithmetic mean is less than or equal to the quadratic mean, weighted
 version for real-valued nonnegative functions. -/
 theorem arith_mean_le_quad_mean_weighted (w z : ι → ℝ) (hw : ∀ i ∈ s, 0 ≤ w i)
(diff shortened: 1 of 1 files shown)


acmepjz:
What about the generalization `∑ i ∈ s, w i * z i ≤ Real.rpow (∑ i ∈ s, w i * Real.rpow (z i) p)) (1 / p)` for any `1 ≤ p`? Furthermore, the right hand side increases as `p` increases.

YaelDillies:
> the right hand side increases as `p` increases.

Yes, this is the one statement I really want

vihdzp:
> What about the generalization `∑ i ∈ s, w i * z i ≤ Real.rpow (∑ i ∈ s, w i * Real.rpow (z i) p) (1 / p)` for any `1 ≤ p`? Furthermore, the right hand side increases as `p` increases.

It would be a nice project to define the generalized weighted mean for any exponent `p : EReal`. For `p = 0` the mean is the geometric mean, for `p = ⊥` it's the minimum, and for `p = ⊤` it's the maximum. That would subsume pretty much everything in this file.

YaelDillies:
Note that this generalised mean is basically [the compact Lp norm](https://github.com/YaelDillies/LeanAPAP/blob/master/LeanAPAP/Prereqs/LpNorm/Compact.lean).

vihdzp:
I think these are generalizations in different directions, since the generalized mean also allows for negative exponents, and for weights given to each variable.

YaelDillies:
The negative exponents I agree with, but the general weights version is [here](https://github.com/YaelDillies/LeanAPAP/blob/master/LeanAPAP/Prereqs/NNLpNorm.lean).

luigi-massacci:
Hello! Sorry for not replying, I was kinda busy. I formalised it like this because I wanted the IMO compendium version, but I think it's quite reasonable to have the version @acmepjz suggested, the proof needs almost no adjustments anyway, I think it might still be worthwhile to have a named lemma for the p = 2 case (even if it's just literally specifying p = 2 in the general case)

luigi-massacci:
I'll get to it when I can spare the time

PR #14800, opened 2024-07-16
[Merged by Bors] - feat(CategoryTheory/Sites): definition of sheaf cohomology

joelriou:
---


Mathlib/Algebra/Homology/DerivedCategory/Ext.lean
```diff
@@ -303,8 +303,54 @@ lemma homAddEquiv_apply (α : Ext X Y n) : homAddEquiv α = α.hom := rfl
...
+@[simps!]
+noncomputable def postcomp (β : Ext Y Z n) (X : C) {a b : ℕ} (h : a + n = b) :
+    Ext X Y a →+ Ext X Z b :=
+  AddMonoidHom.mk' (fun α => α.comp β h) (by simp)
+
+/-- The precomposition `Ext Y Z a →+ Ext X Z b` with `α : Ext X Y n` when `n + a = b`. -/
+@[simps!]
+noncomputable def precomp (α : Ext X Y n) (Z : C) {a b : ℕ} (h : n + a = b) :
+    Ext Y Z a →+ Ext X Z b :=
+  AddMonoidHom.mk' (fun β => α.comp β h) (by simp)
```
erdOne, line 316: I think it would be useful to package them as a bilinear map `Ext Y Z a →+ Ext X Y b →+ Ext X Z c` instead, as we have some API for them (e.g. `AddMonoidHom.flip`)

joelriou: I believe that a definition like `covariantSequence` https://github.com/leanprover-community/mathlib4/pull/14515/files#diff-54c3222a1a7565f433c9e06be02de04bc9070035460c229c3f05ca8a4fee701dR133-R138 is clearer using `postcomp` and the dot notation (and soon, there will be the dual construction with `precomp`), rather than some bilinear map evaluated at an element, and half of the times, it will have to be flipped. The best I could try may be defining the bilinear map, but continue using `postcomp` and `precomp` (which would then be abreviations).

joelriou: I have added a new definition `bilinearComp` and made `post/precomp` abbreviations.

joelriou: Thanks for the suggestion!

Mathlib/CategoryTheory/Sites/SheafCohomology/Basic.lean
```diff
@@ -0,0 +1,84 @@
...
+We also define `Sheaf.cohomologyPresheaf F n : Cᵒᵖ ⥤ AddCommGrp`
+which is the presheaf which sends `U` to the `n`th `Ext`-group
+from the free abelian sheaf generated by the presheaf
+of sets `yoneda.obj U` to `F`.
+
+## TODO
+* if `U` is a terminal object of `C`, define a bijection
+`(F.cohomologyPresheaf n).obj (Opposite.op U) ≃ Sheaf.H F n`.
+* if `U : C`, define a bijection
+`(F.cohomologyPresheaf n).obj (Opposite.op U) ≃ Sheaf.H (F.over U) n`.
```
erdOne, line 30:
```suggestion
* if `U` is a terminal object of `C`, define an isomorphism
`(F.cohomologyPresheaf n).obj (Opposite.op U) ≃+ Sheaf.H F n`.
* if `U : C`, define an isomorphism
`(F.cohomologyPresheaf n).obj (Opposite.op U) ≃+ Sheaf.H (F.over U) n`.
```
Or maybe group isomorphism.

erdOne:
Thanks!
maintainer merge

semorrison:
bors merge

PR #14805, opened 2024-07-16
[Merged by Bors] - chore(Tactic): reduce autoImplicit, part 4

grunweg:
After this, the only remaining uses are `Ring/Basic` (which is requires more work, hence shall get its own PR), `Sat/FromLRAT` and four single declarations in `ToExpr`.

---


semorrison:
bors merge

PR #14808, opened 2024-07-16
[Merged by Bors] - feat (RingTheory/Binomial): Chu-Vandermonde identity

ScottCarnahan:
A general version of Chu-Vandermonde: If `r` and `s` are commuting elements of a binomial ring, then `choose (r + s) k` equals the sum over `ij in antidiagonal k` of `choose r ij.1 * choose s ij.2`.

---

### DIFF
--- a/Mathlib/RingTheory/Binomial.lean
+++ b/Mathlib/RingTheory/Binomial.lean
@@ -344,6 +344,8 @@ open Polynomial
 
 variable {R : Type*}
 
+section
+
 /-- The binomial coefficient `choose r n` generalizes the natural number `choose` function,
   interpreted in terms of choosing without replacement. -/
 def choose [AddCommGroupWithOne R] [Pow R ℕ] [BinomialRing R] (r : R) (n : ℕ) : R :=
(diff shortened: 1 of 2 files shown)


Mathlib/RingTheory/Binomial.lean
```diff
@@ -450,6 +452,54 @@ theorem choose_add_smul_choose [NatPowAssoc R] (r : R) (n k : ℕ) :
   rw [choose_smul_choose (r + k) (n + k) k (Nat.le_add_left k n), Nat.add_sub_cancel,
     add_sub_cancel_right]
 
+end
+
+/-- Pochhammer version of Chu-Vandermonde identity -/
+theorem descPochhammer_smeval_add [Ring R] {r s : R} (k : ℕ) (h: Commute r s) :
+    (descPochhammer ℤ k).smeval (r + s) = ∑ ij ∈ Finset.HasAntidiagonal.antidiagonal k,
```
semorrison, line 459:
```suggestion
    (descPochhammer ℤ k).smeval (r + s) = ∑ ij ∈ antidiagonal k,
```
and `open Finset` before the theorem. (This also lets you drop various `Finset.` if you prefer.)

Mathlib/RingTheory/Binomial.lean
```diff
@@ -450,6 +452,54 @@ theorem choose_add_smul_choose [NatPowAssoc R] (r : R) (n k : ℕ) :
...
+    Nat.choose k ij.1 * ((descPochhammer ℤ ij.1).smeval r * (descPochhammer ℤ ij.2).smeval s) := by
+  induction k with
+  | zero => simp [descPochhammer_zero, Finset.Nat.antidiagonal_zero, Finset.sum_singleton]
```
semorrison, line 462:
```suggestion
  | zero => simp
```
These are all already `@[simp]` lemmas, so I'm not sure it's helpful to add the explicitly unless you are using `simp only`.

semorrison:
bors d+

PR #14811, opened 2024-07-16
[Merged by Bors] - docs: fix bibitem of `Proofs from THE BOOK`

acmepjz:
It was erroneously marked as `Article`, but it in in fact a book. The bibitem is corrected from the corresponding zbMath entry.

---


semorrison:
bors merge

PR #14814, opened 2024-07-16
[Merged by Bors] - feat(Polynomial/AlgebraMap): add `mapAlgHom`

quangvdao:
Add `Polynomial.mapAlgHom` as an algebra homomorphism between two polynomial rings, extending `Polynomial.mapRingHom`. Provide supporting theorems for `mapAlgHom`.

Moved `Polynomial.eval₂AlgHom'` to earlier in `AlgebraMap.lean` , so that I can provide a theorem linking `mapAlgHom` and `eval₂AlgHom'`, without having to put `mapAlgHom` stuff after `aeval`.

---

### DIFF
--- a/Mathlib/Algebra/Polynomial/AlgebraMap.lean
+++ b/Mathlib/Algebra/Polynomial/AlgebraMap.lean
@@ -195,6 +195,7 @@ theorem mapAlgHom_comp (C : Type z) [Semiring C] [Algebra R C] (f : B →ₐ[R] 
   simp [AlgHom.comp_algebraMap, map_map]
   congr
 
+/-- If `A` and `B` are isomorphic as `R`-algebras, then so are their polynomial rings -/
 def mapAlgEquiv (f : A ≃ₐ[R] B) : Polynomial A ≃ₐ[R] Polynomial B :=
   AlgEquiv.ofAlgHom (mapAlgHom f.toAlgHom) (mapAlgHom f.symm.toAlgHom) (by simp) (by simp)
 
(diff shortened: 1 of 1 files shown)


Mathlib/Algebra/Polynomial/AlgebraMap.lean
```diff
@@ -173,6 +165,56 @@ def eval₂AlgHom' (f : A →ₐ[R] B) (b : B) (hf : ∀ a, Commute (f a) b) : A
...
+theorem mapAlgHom_comp (C : Type z) [Semiring C] [Algebra R C] (f : B →ₐ[R] C) (g : A →ₐ[R] B) :
+    (mapAlgHom f).comp (mapAlgHom g) = mapAlgHom (f.comp g) := by
+  apply AlgHom.ext
+  intro x
+  simp [AlgHom.comp_algebraMap, map_map]
+  congr
+
+/-- If `A` and `B` are isomorphic as `R`-algebras, then so are their polynomial rings -/
+def mapAlgEquiv (f : A ≃ₐ[R] B) : Polynomial A ≃ₐ[R] Polynomial B :=
+  AlgEquiv.ofAlgHom (mapAlgHom f.toAlgHom) (mapAlgHom f.symm.toAlgHom) (by simp) (by simp)
```
semorrison, line 200: This needs analogues of the 4 theorems above.

acmepjz: Do we have a stronger result `Polynomial A ≃ₐ[Polynomial R] Polynomial B`?

quangvdao: I can add in those results. I am not sure how to name them though. `mapAlgHom'` and `mapAlgEquiv'` is an option, but they are not very descriptive.

erdOne: I think they could replace `mapAlgHom`. If you need the `R`-alghom version, use `AlgHom.restrictScalars`.

acmepjz:
Sorry but I think I made a mistake. The instance `Algebra (Polynomial R) (Polynomial A)` cannot be inferred automatically. And that should not be added to mathlib in my opinion, since potentially the intermediate X is not necessary map to X.

quangvdao:
@acmepjz: yeah, I agree. So let's just go ahead and add `mapAlgHom`, `mapAlgEquiv` the way I currently write it?

semorrison:
bors merge

PR #11149, opened 2024-03-04
[Merged by Bors] - chore(CategoryTheory/MonoidalCategory): split the monoidal composition

yuma-mizuno:
I realized that the monoid composition `⊗≫` can be defined without depending on the free monoid category, and that it is still useful.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14815, opened 2024-07-17
[Merged by Bors] - chore(CategoryTheory/Bicategory): split the bicategorical composition

yuma-mizuno:
Similar to #11149.

---


semorrison:
bors merge

PR #14821, opened 2024-07-17
[Merged by Bors] - chore(workflows/nightly_detect_failure): log messages

jcommelin:
The friendly bot repeats itself for no good reason.
We start logging messages to debug the issue.

---


adomani:
LGTM!

What I had in mind was printing to shell in which the action takes place.  However also printing to the actual message has the same effect!

There is an `or` in the condition: would this give you information regardless of which branch of the `or` is true?

jcommelin:
No, it only prints messages if there are actually messages to print. But this should always be true, since the Zulip thread has been nonempty for decades :smiley:

adomani:
Ok, I understand now!

If you nevertheless want to remove the precondition on line 206, I would still

maintainer merge

this PR!  :smile:

mattrobball:
bors merge

PR #14825, opened 2024-07-17
[Merged by Bors] - feat: missing Nat.card lemmas

PatrickMassot:
---

### DIFF
--- a/Mathlib/SetTheory/Cardinal/Finite.lean
+++ b/Mathlib/SetTheory/Cardinal/Finite.lean
@@ -94,6 +94,21 @@ theorem card_eq_of_bijective (f : α → β) (hf : Function.Bijective f) : Nat.c
   card_congr (Equiv.ofBijective f hf)
 #align nat.card_eq_of_bijective Nat.card_eq_of_bijective
 
+protected theorem bijective_iff_injective_and_card [Fintype β] (f : α → β) :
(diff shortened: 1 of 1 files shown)


Mathlib/SetTheory/Cardinal/Finite.lean
```diff
@@ -94,6 +94,21 @@ theorem card_eq_of_bijective (f : α → β) (hf : Function.Bijective f) : Nat.c
   card_congr (Equiv.ofBijective f hf)
 #align nat.card_eq_of_bijective Nat.card_eq_of_bijective
 
+protected theorem bijective_iff_injective_and_card [Fintype β] (f : α → β) :
```
apnelson1, line 97: This should be true with `Nat.card β ≤ Nat.card α`, right? And it's probably is mostly used in the `.mpr` direction to prove bijectivity. Would the version with `≤` be more convenient?

PatrickMassot: I really copied the Fintype version here, except that I removed the `[Fintype α]`. The issue with your version is that the conclusion is less strong in the `.mp` direction. So I think your version does not really make sense as an `iff`. Having the `.mpr` version with weaker assumption as a separate lemma would make sense.

PatrickMassot: I added your lemma (only the meaningful implication).

semorrison:
bors merge

PR #14826, opened 2024-07-17
[Merged by Bors] - chore: Split RingTheory.IntegralClosure

riccardobrasca:
We create a new folder `RingTheory.IntegralClosure` and we move the relevant files (also doing some splitting).

---


erdOne:
I don't really understand the difference between `Basic` and `Defs`, and the fact that `Defs` is more basic than `Basic` is confusing to me. IMHO that file should be split up into integral elements / integral algebras / integral closures instead.

riccardobrasca:
> I don't really understand the difference between `Basic` and `Defs`, and the fact that `Defs` is more basic than `Basic` is confusing to me. IMHO that file should be split up into integral elements / integral algebras / integral closures instead.

`Defs` is more fundamental than `Basic` everywhere, isn't it? Anyway let me see if I can split the folder into three subfolders.

riccardobrasca:
What do you think if this further splitting? (I still have to fix the docs)

erdOne:
I like the split, thanks! I still don't understand the `def` / `basic` split though (I learnt of it just now). It seems to me like a hack on low-hierarchical files to reduce imports, which IMO is not a problem here. This breaks the "when I need to find API lemmas, just `Go to Definition`" workflow that I (and maybe others?) rely on.

mattrobball:
bors merge

PR #14829, opened 2024-07-17
[Merged by Bors] - chore: Split and move RingTheory.Ideal.LocalRing

riccardobrasca:
Currently, [LocalRing](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Ideal/LocalRing.html#LocalRing) is defined in the (long) file `Mathlib.RingTheory.Ideal.LocalRing`. We also have a folder `Mathlib.RingTheory.LocalRing` that contains only one (advanced) file `Module`.

The file `Mathlib.RingTheory.Ideal.LocalRing` is moved an splitted in several files. The new structure is
```
├── RingTheory.LocalRing
│   ├── MaximalIdeal
│   │   ├── Defs
│   │   ├── Basic
│   ├── RingHom
│   │   ├── Defs
│   │   ├── Basic
│   ├── ResidueField
│   │   ├── Defs
│   │   ├── Basic
│   ├── Defs
│   ├── Basic
└── ...
```

---


erdOne:
Similarly, I think a split between local rings / local ring homs / residue fields would be more reasonable IMHO.

BTW thanks a lot for doing this! I created the `LocalRing` folder with the intention of moving the local ring file inside but never got around to it.

Mathlib/RingTheory/LocalRing/Defs.lean
```diff
@@ -0,0 +1,31 @@
...
+
+/-!
+
+# Local rings
+
+Define local rings as commutative rings having a unique maximal ideal.
+
+## Main definitions
+
+* `LocalRing`: A predicate on commutative semirings, stating that for any pair of elements that
```
jcommelin, line 19:
```suggestion
* `LocalRing`: A predicate on semirings, stating that for any pair of elements that
```

Mathlib/RingTheory/LocalRing/Defs.lean
```diff
@@ -0,0 +1,31 @@
...
+  adds up to `1`, one of them is a unit. This is shown to be equivalent to the condition that there
+  exists a unique maximal ideal.
```
jcommelin, line 21:
```suggestion
  adds up to `1`, one of them is a unit. In the commutative case this is shown to be equivalent
  to the condition that there exists a unique maximal ideal, see
  `LocalRing.of_unique_max_ideal` and `LocalRing.maximal_ideal_unique`.
```

Mathlib/RingTheory/LocalRing/ResidueField/Basic.lean
```diff
@@ -0,0 +1,199 @@
...
+  show Inhabited (R ⧸ maximalIdeal R) from inferInstance
+
+noncomputable instance ResidueField.field : Field (ResidueField R) :=
+  Ideal.Quotient.field (maximalIdeal R)
+#align local_ring.residue_field.field LocalRing.ResidueField.field
+
+/-- The quotient map from a local ring to its residue field. -/
+def residue : R →+* ResidueField R :=
+  Ideal.Quotient.mk _
+#align local_ring.residue LocalRing.residue
```
jcommelin, line 43: These are all defining data. If it doesn't increase imports, I think it would be good to have these in `Defs.lean`.

jcommelin (commented):
Thanks for taking on this split!

Mathlib/RingTheory/LocalRing/ResidueField/Basic.lean
riccardobrasca: I've moved all of them.  `ResidueField.field` needs `LocalRing.MaximalIdeal.Basic` instead of `LocalRing.MaximalIdeal.Defs`, but I think this is fine (I can move it back if we want to only import `LocalRing.MaximalIdeal.Defs`).

jcommelin:
bors d+

PR #14832, opened 2024-07-17
[Merged by Bors] - feat: add `orthogonalComplement_eq_orthogonalComplement`

Samyak-DT:
Included a lemma that asserts that two vector submodules of a vector space in a `RClike field' are equal if and only if their orthogonal complements are equal 

---

### DIFF
--- a/Mathlib/Analysis/InnerProductSpace/Projection.lean
+++ b/Mathlib/Analysis/InnerProductSpace/Projection.lean
@@ -845,6 +845,14 @@ theorem Submodule.isCompl_orthogonal_of_completeSpace [HasOrthogonalProjection K
   ⟨K.orthogonal_disjoint, codisjoint_iff.2 Submodule.sup_orthogonal_of_completeSpace⟩
 #align submodule.is_compl_orthogonal_of_complete_space Submodule.isCompl_orthogonal_of_completeSpace
 
+theorem eq_iff_orthogonalComplement_eq {L : Submodule 𝕜 E}[HasOrthogonalProjection K]
+    [HasOrthogonalProjection L] : K = L ↔ Kᗮ = Lᗮ := by
+   constructor
+   · exact fun a ↦ congrArg Submodule.orthogonal a
+   · intro H
+     rw [← (Submodule.orthogonal_orthogonal K), ← (Submodule.orthogonal_orthogonal) L]
+     exact congrArg Submodule.orthogonal H
+
 @[simp]
 theorem Submodule.orthogonal_eq_bot_iff [HasOrthogonalProjection K] : Kᗮ = ⊥ ↔ K = ⊤ := by
   refine ⟨?_, fun h => by rw [h, Submodule.top_orthogonal_eq_bot]⟩
--- a/Mathlib/Analysis/InnerProductSpace/Projection.lean
+++ b/Mathlib/Analysis/InnerProductSpace/Projection.lean
@@ -845,13 +845,11 @@ theorem Submodule.isCompl_orthogonal_of_completeSpace [HasOrthogonalProjection K
   ⟨K.orthogonal_disjoint, codisjoint_iff.2 Submodule.sup_orthogonal_of_completeSpace⟩
 #align submodule.is_compl_orthogonal_of_complete_space Submodule.isCompl_orthogonal_of_completeSpace
 
-theorem eq_iff_orthogonalComplement_eq {L : Submodule 𝕜 E}[HasOrthogonalProjection K]
-    [HasOrthogonalProjection L] : K = L ↔ Kᗮ = Lᗮ := by
-   constructor
-   · exact fun a ↦ congrArg Submodule.orthogonal a
-   · intro H
-     rw [← (Submodule.orthogonal_orthogonal K), ← (Submodule.orthogonal_orthogonal) L]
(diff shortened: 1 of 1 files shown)


Mathlib/Analysis/InnerProductSpace/Projection.lean
```diff
@@ -845,6 +845,14 @@ theorem Submodule.isCompl_orthogonal_of_completeSpace [HasOrthogonalProjection K
   ⟨K.orthogonal_disjoint, codisjoint_iff.2 Submodule.sup_orthogonal_of_completeSpace⟩
 #align submodule.is_compl_orthogonal_of_complete_space Submodule.isCompl_orthogonal_of_completeSpace
 
+theorem eq_iff_orthogonalComplement_eq {L : Submodule 𝕜 E}[HasOrthogonalProjection K]
+    [HasOrthogonalProjection L] : K = L ↔ Kᗮ = Lᗮ := by
```
j-loreaux, line 849: I would reverse the order (which forces a rename), and then it can be a `simp` lemma. It can't be a `simp` lemma without reverse the order because then it would just keep adding `ᗮ` repeatedly.
```suggestion
@[simp]
theorem orthogonalComplement_eq_orthogonalComplement {L : Submodule 𝕜 E} [HasOrthogonalProjection K]
    [HasOrthogonalProjection L] : Kᗮ = Lᗮ ↔ K = L := by
```

Mathlib/Analysis/InnerProductSpace/Projection.lean
```diff
@@ -845,6 +845,14 @@ theorem Submodule.isCompl_orthogonal_of_completeSpace [HasOrthogonalProjection K
...
+   constructor
+   · exact fun a ↦ congrArg Submodule.orthogonal a
+   · intro H
+     rw [← (Submodule.orthogonal_orthogonal K), ← (Submodule.orthogonal_orthogonal) L]
+     exact congrArg Submodule.orthogonal H
```
j-loreaux, line 854: With the above suggested change, you can do this for the proof (otherwise you have to reverse the order of the arguments):
```
  ⟨fun h ↦ by simpa using congr(Submodule.orthogonal $(h)),
    fun h ↦ congr(Submodule.orthogonal $(h))⟩
```
A few notes about this proof:
1. `congr()` is a *really* cool way of building equalities from existing ones. `congrArg` (or `congr_arg`) is the way of doing it by hand, but after you try to prove `a + b = c + d` from `h₁ : a = c` and `h₂ : b = d` using `congr_arg` alone (not allowing yourself `rw`), you will realize the full power of `congr`, which can do this simply as `congr($(h₁) + $(h₂))`.
2. `simpa using <term>` is nice because it allows you to supply a term which only needs to agree with the goal up to simplification. So if Lean can simplify this term and the goal to the same thing, it closes the goal.
3. While I see what you were doing with the backwards rewriting, it can sort of be a sign (not always) that you're doing something in a non-optimal way if you are reverse-rewriting `simp` lemmas. At the very least, if you try to `simp` after you undo your work. So, instead, it can sometimes be helpful to build a *more* complicated term, and then simplify it, which is what I've done here. In the first line, `h : Kᗮ = Lᗮ`, then the `congr(Submodule.orthogonal $(h)) : Kᗮᗮ = Lᗮᗮ`, and the `simpa` uses `Submodule.orthogonal_orthogonal` under the hood to remove the doubled perps to yield `K = L`, as desired.

This last one is a pretty common trick, and I think it's closer to what you might do on paper.

j-loreaux (commented):
Thanks for this! It does indeed appear to be missing, as you can verify with [Loogle](https://loogle.lean-lang.org/?q=%3FKᗮ+%3D+%3FLᗮ)

Mathlib/Analysis/InnerProductSpace/Projection.lean
Samyak-DT: Thank you for the helpful comments!

j-loreaux:
@Samyak-DT make sure when you are finished addressing comments on PRs that you remove the `awaiting-author` label in order to guarantee the PR will get reviewed again.

Mathlib/Analysis/InnerProductSpace/Projection.lean
```diff
@@ -845,6 +845,12 @@ theorem Submodule.isCompl_orthogonal_of_completeSpace [HasOrthogonalProjection K
   ⟨K.orthogonal_disjoint, codisjoint_iff.2 Submodule.sup_orthogonal_of_completeSpace⟩
 #align submodule.is_compl_orthogonal_of_complete_space Submodule.isCompl_orthogonal_of_completeSpace
 
+@[simp]
+theorem orthogonalComplement_eq_orthogonalComplement {L : Submodule 𝕜 E} [HasOrthogonalProjection K]
+    [HasOrthogonalProjection L] : Kᗮ = Lᗮ ↔ K = L := by
+    exact ⟨fun h ↦ by simpa using congr(Submodule.orthogonal $(h)),
```
j-loreaux, line 851:
```suggestion
    [HasOrthogonalProjection L] : Kᗮ = Lᗮ ↔ K = L :=
  ⟨fun h ↦ by simpa using congr(Submodule.orthogonal $(h)),
```

PR #14833, opened 2024-07-17
[Merged by Bors] - doc: fix module documentation for `AntivaryOn`

GanjinZero:
Fix `MonovaryOn f g s`: `f` antivaries with `g` on `s`. to `AntivaryOn f g s`: `f` monovaries with `g` on `s`.

---


j-loreaux:
@GanjinZero thanks for this change. Just so you know, our workflow is a bit different than most repos, and we normally don't accept PRs from forks. Instead, you generally need to follow the instructions [here](https://leanprover-community.github.io/contribute/index.html) to obtain write access to the repo.

In this case, since it is a change for documentation only, I'll just merge it.

bors merge.

j-loreaux:
bors merge

GanjinZero:
> @GanjinZero thanks for this change. Just so you know, our workflow is a bit different than most repos, and we normally don't accept PRs from forks. Instead, you generally need to follow the instructions [here](https://leanprover-community.github.io/contribute/index.html) to obtain write access to the repo.
> 
> In this case, since it is a change for documentation only, I'll just merge it.
> 
> bors merge.

Got it, will follow instructions next time.

PR #14820, opened 2024-07-17
[Merged by Bors] - feat: prove that a meromorphic function is eventually analytic

Command-Master:
Prove that a meromorphic function at a point is analytic on a punctured neighborhood of it, and that a meromorphic function on a set is analytic on a codiscrete subset of it.

---

- [x] depends on: #14837

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14837, opened 2024-07-17
[Merged by Bors] - feat(Topology/DiscreteSubset): prove lemmas about the codiscrete filter, and define codiscreteWithin

Command-Master:
Prove lemmas about the codiscrete filter.

---

### DIFF
--- a/Mathlib/Topology/DiscreteSubset.lean
+++ b/Mathlib/Topology/DiscreteSubset.lean
@@ -104,4 +104,41 @@ def Filter.codiscrete (X : Type*) [TopologicalSpace X] : Filter X where
     simp_rw [← isClosed_compl_iff, isClosed_and_discrete_iff] at hU hV ⊢
     exact fun x ↦ compl_inter U V ▸ sup_principal ▸ disjoint_sup_right.mpr ⟨hU x, hV x⟩
 
+lemma mem_codiscrete {S : Set X} :
+    S ∈ codiscrete X ↔ IsOpen S ∧ DiscreteTopology ↑Sᶜ := Iff.rfl
+
+lemma mem_codiscrete' {S : Set X} :
+    S ∈ codiscrete X ↔ ∀ x, Disjoint (𝓝[≠] x) (𝓟 Sᶜ) := by
+  rw [mem_codiscrete, ← isClosed_compl_iff, isClosed_and_discrete_iff]
+
+lemma mem_codiscrete_subtype {S : Set X} {U : Set S} :
+    U ∈ codiscrete S ↔ ∀ x ∈ S, ¬AccPt x (𝓟 (Subtype.val '' Uᶜ)) := by
+  simp only [mem_codiscrete', disjoint_principal_right, compl_compl, Subtype.forall, AccPt,
+    not_neBot, inf_principal_eq_bot]
+  congr! with x hx
+  constructor
+  · intro h
+    rw [nhdsWithin_subtype] at h
+    simp only [mem_comap] at h
+    obtain ⟨t, ht1, ht2⟩ := h
+    simp [mem_nhdsWithin] at ht1 ⊢
+    obtain ⟨u, hu1, hu2, hu3⟩ := ht1
+    use u, hu1, hu2
+    intro v hv
+    simp only [mem_compl_iff, mem_image, Subtype.exists, exists_and_right, exists_eq_right,
+      not_exists, not_not]
+    intro hv2
+    apply ht2
+    simp only [mem_preimage]
+    apply hu3
+    simpa [hv2]
+  · intro h
+    have : Tendsto Subtype.val (𝓝[≠] ⟨x, hx⟩) (𝓝[≠] x) :=
+      tendsto_nhdsWithin_of_tendsto_nhds_of_eventually_within _
+      continuous_subtype_val.continuousWithinAt <| eventually_mem_nhdsWithin.mono (by simp)
+    rw [tendsto_def] at this
+    convert this _ h
+    ext
+    simp
+
 end codiscrete_filter
--- a/Mathlib/Topology/DiscreteSubset.lean
+++ b/Mathlib/Topology/DiscreteSubset.lean
@@ -117,7 +117,7 @@ lemma mem_codiscreteWithin_accPt {S T : Set X} :
   simp only [mem_codiscreteWithin, disjoint_iff, AccPt, not_neBot]
 
 /-- In any topological space, we form a filter from the supremum of all punctured neighborhoods. -/
-def Filter.codiscrete (X : Type*) [TopologicalSpace X] : Filter X := codiscreteWithin Set.univ
+abbrev Filter.codiscrete (X : Type*) [TopologicalSpace X] : Filter X := codiscreteWithin Set.univ
 
 lemma mem_codiscrete {S : Set X} :
     S ∈ codiscrete X ↔ ∀ x, Disjoint (𝓝[≠] x) (𝓟 Sᶜ) := by
--- a/Mathlib/Topology/DiscreteSubset.lean
+++ b/Mathlib/Topology/DiscreteSubset.lean
@@ -31,7 +31,7 @@ see `IsClosed.tendsto_coe_cofinite_iff`.
 ## Co-discrete open sets
 
 In a topological space the sets which are open with discrete complement form a filter. We
-formalise this as `Filter.codiscrete`.
+formalise this as `Filter.codiscrete`. This is also the supremum of all punctured neighborhoods.
 
 -/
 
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/DiscreteSubset.lean
```diff
@@ -104,4 +104,41 @@ def Filter.codiscrete (X : Type*) [TopologicalSpace X] : Filter X where
...
 
+lemma mem_codiscrete {S : Set X} :
+    S ∈ codiscrete X ↔ IsOpen S ∧ DiscreteTopology ↑Sᶜ := Iff.rfl
+
+lemma mem_codiscrete' {S : Set X} :
+    S ∈ codiscrete X ↔ ∀ x, Disjoint (𝓝[≠] x) (𝓟 Sᶜ) := by
+  rw [mem_codiscrete, ← isClosed_compl_iff, isClosed_and_discrete_iff]
+
+lemma mem_codiscrete_subtype {S : Set X} {U : Set S} :
+    U ∈ codiscrete S ↔ ∀ x ∈ S, ¬AccPt x (𝓟 (Subtype.val '' Uᶜ)) := by
```
j-loreaux, line 115: Is this situation really coming up that frequently? We normally try to avoid situations like `{S : Set X} {U : Set S}`.

Command-Master: I don't know how common this is in general, but when working with analytic and meromorphic functions on a set `S` I believe that `codiscrete S` is the best way to state "almost everywhere" (see #14820, for example).

j-loreaux: I guess I see a number of issues here:

1. I think the *definition* of `Filter.codiscrete` should be migrated to `mem_codiscrete'`, as it plays better with our API, and avoids a `CoeSort` in the definition.
2. It's weird that you phrase `mem_codiscrete'` with `Disjoint` but `mem_codiscrete_subtype` with `¬ AccPt`
3. This is the big one. I would prefer that we *almost never* talk about `U ∈ codiscrete S` where `{S : Set X} {U : Set S}`. Instead, I think we should create a *new* definition `codiscreteWithin`, much like `nhdsWithin`. I will describe this now.

The definition of `t ∈ codiscrete X` is (equivalent to) `Disjoint (⨆ x : X, 𝓝[≠] x) (𝓟 tᶜ)`. I propose that we have, for `s : Set X`, `codiscreteWithin s X` where `t ∈ codiscreteWithin s X` means `Disjoint (⨆ x ∈ s, 𝓝[≠] x) (𝓟 tᶜ)`. Note, I have not thought about whether there is a nice filter-level description of this (as with `𝓝[s] x := 𝓝 x ⊓ (𝓟 s)`).

Note that if `t : Set s`, then `t ∈ codiscrete s ↔ (Subtype.val '' tᶜ) ∈ codiscreteWithin s X`. Probably all of this should be discussed on Zulip. But the advantage of my approach is that it allows everything to stay on the top level.

Command-Master: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/codiscrete.20within.20a.20set

j-loreaux: @Command-Master given the (albeit small) consensus from Zulip, can you please implement the following suggestions? (you may split this into multiple PRs to make review easier.)

1. (Optional, but would be my preference if it's not too much work, and it makes some of what follows slightly easier too) redefine `Filter.codiscrete X : Filter X := ⨆ x : X, 𝓝[≠] x`
2. Define for `s : Set X`, a new `Filter.codiscreteWithin s X : Filter X := ⨆ x ∈ s, 𝓝[≠] x`.
3. Develop the basic API for this new filter (e.g., `codiscreteWithin Set.univ X = codiscrete X`, or more generally, I think `(codiscreteWithin · X)` can be made into a `CompleteLatticeHom` pretty easily (if I'm not mistaken).

In the above you are welcome to include the equivalent conditions in terms of subtypes, but I think we should try to avoid them when / where possible.

Command-Master: I'm not sure if `⨆ x ∈ s, 𝓝[≠] x` is a correct definition for `codiscreteWithin` - using it, `S ∈ codiscreteWithin T` means that `Sᶜ` has no accumulation points within `T`, but I think the correct definition should mean that `T \ S` has no accumulation points within `T` (this also agrees with `codiscrete` on the subtype), that is `⨆ x ∈ S, 𝓝[S \ {x}] x`.
Also, shouldn't `X` be implicit in `codiscreteWithin`? Finally, what do you think about having `codiscrete X = codiscreteWithin Set.univ X` directly as the definition?

j-loreaux: Sure, if that's the version equivalent to the subtype for `codiscrete`, then let's go with that. Sorry I got it wrong before.

Yes `X` should be implicit.

I'm tempted to keep them separate, since the definition is cleaner, but perhaps an `abbrev` with the right API lemmas is okay. That's a question for another reviewer / Zulip (but you don't have to wait forever on Zulip, sorry it took so long before).

Mathlib/Topology/DiscreteSubset.lean
```diff
@@ -91,17 +92,75 @@ theorem isClosed_and_discrete_iff {S : Set X} :
...
+    apply hu3
+    simpa [hv2]
+  · intro h
+    have : Tendsto Subtype.val (𝓝[≠] ⟨x, hx⟩) (𝓝[≠] x) :=
+      tendsto_nhdsWithin_of_tendsto_nhds_of_eventually_within _
+      continuous_subtype_val.continuousWithinAt <| eventually_mem_nhdsWithin.mono (by simp)
+    rw [tendsto_def] at this
+    convert this _ h
+    ext
+    simp
```
j-loreaux, line 164:
```suggestion
    U ∈ codiscrete S ↔ (↑) '' U ∈ codiscreteWithin S := by
  simp [mem_codiscrete, disjoint_principal_right, compl_compl, Subtype.forall,
    mem_codiscreteWithin]
  congr! with x hx
  constructor
  · rw [nhdsWithin_subtype, mem_comap]
    rintro ⟨t, ht1, ht2⟩
    rw [mem_nhdsWithin] at ht1 ⊢
    obtain ⟨u, hu1, hu2, hu3⟩ := ht1
    refine ⟨u, hu1, hu2, fun v hv ↦ ?_⟩
    simpa using fun hv2 ↦ ⟨hv2, ht2 <| hu3 <| by simpa [hv2]⟩
  · suffices Tendsto (↑) (𝓝[≠] (⟨x, hx⟩ : S)) (𝓝[≠] x) by convert tendsto_def.mp this _; ext; simp
    exact tendsto_nhdsWithin_of_tendsto_nhds_of_eventually_within _
      continuous_subtype_val.continuousWithinAt <| eventually_mem_nhdsWithin.mono (by simp)
```

Mathlib/Topology/DiscreteSubset.lean
```diff
@@ -91,17 +92,75 @@ theorem isClosed_and_discrete_iff {S : Set X} :
...
-  univ_sets := ⟨isOpen_univ, compl_univ.symm ▸ Subsingleton.discreteTopology⟩
-  sets_of_superset := by
-    intro U V hU hV
-    simp_rw [← isClosed_compl_iff, isClosed_and_discrete_iff] at hU ⊢
-    exact fun x ↦ (hU x).mono_right (principal_mono.mpr <| compl_subset_compl.mpr hV)
-  inter_sets := by
-    intro U V hU hV
-    simp_rw [← isClosed_compl_iff, isClosed_and_discrete_iff] at hU hV ⊢
-    exact fun x ↦ compl_inter U V ▸ sup_principal ▸ disjoint_sup_right.mpr ⟨hU x, hV x⟩
+/-- We can create a filter of sets with no accumulation points inside a subset. -/
```
j-loreaux, line 95:
```suggestion
/-- The filter of sets with no accumulation points inside a set `S : Set X`, implemented
as the supremum over all punctured neighborhoods within `S`.  -/
```

Mathlib/Topology/DiscreteSubset.lean
```diff
@@ -91,17 +92,75 @@ theorem isClosed_and_discrete_iff {S : Set X} :
...
+      simp only [mem_compl_iff, mem_diff, not_and, not_not] at this
+      exact this hx.2.1
+    apply h
+    simp_all
+
+lemma mem_codiscreteWithin_accPt {S T : Set X} :
+    S ∈ codiscreteWithin T ↔ ∀ x ∈ T, ¬AccPt x (𝓟 (T \ S)) := by
+  simp only [mem_codiscreteWithin, disjoint_iff, AccPt, not_neBot]
+
+/-- In any topological space, we form a filter from the supremum of all punctured neighborhoods. -/
```
j-loreaux, line 120:
```suggestion
/-- In any topological space, the open sets with with discrete complement form a filter,
defined as the supremum of all punctured neighborhoods.

See `Filter.mem_codiscrete'` for the equivalence. -/
```

j-loreaux (commented):
I'm pretty happy with this now. Thanks for doing this work!

Please feel free to merge after you address these issues below. Make sure you wait for CI to finish successfully.

bors d+

PR #14838, opened 2024-07-17
[Merged by Bors] - feat(Order/Interval/Finset/Nat): `Icc_insert_succ_left`, `Icc_insert_succ_right`

jsm28:
Add two more lemmas about inserting an element just before or after an interal in `ℕ`:

```lean
lemma Icc_insert_succ_left (h : a ≤ b) : insert a (Icc (a + 1) b) = Icc a b := by

lemma Icc_insert_succ_right (h : a ≤ b + 1) : insert (b + 1) (Icc a b) = Icc a (b + 1) := by
```

---

### DIFF
--- a/Mathlib/Order/Interval/Finset/Nat.lean
+++ b/Mathlib/Order/Interval/Finset/Nat.lean
@@ -193,6 +193,16 @@ theorem Ico_insert_succ_left (h : a < b) : insert a (Ico a.succ b) = Ico a b := 
   rw [Ico_succ_left, ← Ioo_insert_left h]
 #align nat.Ico_insert_succ_left Nat.Ico_insert_succ_left
 
+lemma Icc_insert_succ_left (h : a ≤ b) : insert a (Icc (a + 1) b) = Icc a b := by
+  ext x
+  simp
+  omega
+
+lemma Icc_insert_succ_right (h : a ≤ b + 1) : insert (b + 1) (Icc a b) = Icc a (b + 1) := by
+  ext x
+  simp
+  omega
+
 theorem image_sub_const_Ico (h : c ≤ a) :
     ((Ico a b).image fun x => x - c) = Ico (a - c) (b - c) := by
   ext x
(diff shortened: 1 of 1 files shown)


rwst (approved):
LGTM. Good to have.

Mathlib/Order/Interval/Finset/Nat.lean
```diff
@@ -193,6 +193,16 @@ theorem Ico_insert_succ_left (h : a < b) : insert a (Ico a.succ b) = Ico a b :=
   rw [Ico_succ_left, ← Ioo_insert_left h]
 #align nat.Ico_insert_succ_left Nat.Ico_insert_succ_left
 
+lemma Icc_insert_succ_left (h : a ≤ b) : insert a (Icc (a + 1) b) = Icc a b := by
+  ext x
+  simp
```
mattrobball, line 198: Can you squeeze this `simp` by replacing it with the output of `simp?`? We want to avoid non-terminal `simp`'s because they are difficult to maintain.

Mathlib/Order/Interval/Finset/Nat.lean
```diff
@@ -193,6 +193,16 @@ theorem Ico_insert_succ_left (h : a < b) : insert a (Ico a.succ b) = Ico a b :=
...
+  omega
+
+lemma Icc_insert_succ_right (h : a ≤ b + 1) : insert (b + 1) (Icc a b) = Icc a (b + 1) := by
+  ext x
+  simp
```
mattrobball, line 203: Same here.

Mathlib/Order/Interval/Finset/Nat.lean
jsm28: I didn't think of `omega` as a rigid tactic (thus, I didn't consider this to be one of the places where a non-terminal `simp` is actually a problem), but I've made that change.

Mathlib/Order/Interval/Finset/Nat.lean
jsm28: Likewise.

Ruben-VandeVelde (approved):
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14839, opened 2024-07-17
test: feat: the minImports linter

adomani:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14840, opened 2024-07-17
[Merged by Bors] - feat: the `minImports` linter

adomani:
Introducing the "linter form" of `#min_imports in`.  The linter is called `linter.minImports` and flags every command that increases the imports-so-far in the file.  When it reaches the end of the file, it also reports which current imports are unneeded and which imports are missing to minimize the current imports.

Here is an example.

```lean
import Mathlib.Tactic.Linter.MinImports
-- `import Mathlib.Data.Int.Notation` flagged as missing
import Mathlib.Data.Nat.Notation   -- flagged as unneeded
import Mathlib.Data.Int.Defs       -- flagged as unneeded

-- Imports increased to [Lean.Parser.Command]
set_option linter.minImports true

-- Imports increased to [Mathlib.Tactic.Linter.MinImports]
#reset_min_imports

-- Imports increased to [Mathlib.Data.Int.Notation, Mathlib.Tactic.Linter.MinImports]
#guard (0 : ℤ) = 0
```

Test PR at #14839.

---


Mathlib/Tactic/Linter/MinImports.lean
```diff
@@ -0,0 +1,83 @@
...
+def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.minImports o
+
+@[inherit_doc Mathlib.Linter.linter.minImports]
+def minImportsLinter : Linter where run := withSetOptionIn fun stx => do
+    unless linter.minImports.get (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    let prevImports ← minImportsRef.get
+    --if stx.isOfKind ``Parser.Command.eoi then logInfo m!"{(← getEnv).imports.map (·.module)}"
```
semorrison, line 66: Could this commented out code be either removed or explained in comments before merging?

Mathlib/Tactic/MinImports.lean
```diff
@@ -59,34 +132,32 @@ def getAllImports {m : Type → Type} [Monad m] [MonadResolveName m] [MonadEnv m
...
-
-/-- `getIrredundantImports cmd` takes a `Syntax` input `cmd`.
-It returns the `NameSet` consisting of a minimal collection of module names whose transitive
+      let new := match env.getModuleIdxFor? t with
+        | some t => (hm.find? t).get!
+        | none   => .anonymous -- instead of `getMainModule`, we omit the current module
+      if !fins.contains new then fins := fins.insert new
+  return fins.erase .anonymous
+
+/-- `getIrredundantImports env importNames` takes an `Environment` and a `NameSet` as inputs.
```
semorrison, line 141: Is this functionality not already provided in the `import-graph` repo?

If so, please use, if not, ideally eventually this material could move there, but no hurry.

semorrison:
I love the test file!

Mathlib/Tactic/MinImports.lean
adomani: The closest that I could find in `ImportGraph` is `Lean.Environment.findRedundantImports`, which is the function that I am using in `env.findRedundantImports`.  This is indeed really close: it determines precisely the imports that I want to exclude!  I could not find the command that I defined, since `ImportGraph` proceeds to define some `down`/`upstream` of, containing all imports above or below and, for `#min_imports`, the commands are already mixed up with the `Expr` information with all the declarations in the environment.

I am happy to move this and most of the new `#min_imports` functionality to `ImportGraph`, but I would rather do that later, after testing it in mathlib: I imagine that quite a few improvements will be necessary and doing it within mathlib should have a quicker turnaround and involve fewer inter-repo messing around.

Mathlib/Tactic/Linter/MinImports.lean
adomani: I am planning to improve this message: besides warning on each import-bumping declaration, I am also going to make the command tag redundant/missing imports when it parses `eoi`.  This commented out code was simply a debugging message that I have not yet rewritten.

adomani: I just pushed a version where the command emits a report upon reaching `eoi` or `#exit`.

Mathlib/Tactic/Linter/MinImports.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-! # The `minImports` linter
+
+The `minImports` linter incrementally computes the minimal imports needed for each file to build.
+Whenever it detects that a new command requires an increase in the (transitive) imports that it
+computed so far, it emits a warning mentioning the bigger minimal imports.
+
+Unlike the related `#min_imports` command, the linter takes into account notation and tactic
+information.
+It also works incrementally, providing information that it better suited, for instance, to split
```
jcommelin, line 17: This sentence has a grammar mistake around "that it better suited", but I'm not sure in which way to fix it.

test/MinImports.lean
```diff
@@ -69,3 +75,59 @@ import Mathlib.Data.Nat.Notation
...
+-- finally, we leave the linter "on", to give it a chance to parse `eoi`
+-- and to silence it, we make sure to put in a command for each import
+
+/--
+warning: Imports increased to
+[Init.Guard, Lean.Parser.Command, Mathlib.Data.Int.Notation]
+note: this linter can be disabled with `set_option linter.minImports false`
+-/
+#guard_msgs in
+set_option linter.minImports true
```
jcommelin, line 115: Will `set_option linter.minImports true` always cause the next command to get a warning?
Should there be a command that sets `minImportsRef` to whatever is currently needed, without getting a warning from *this* linter. (It will be caught during PR review by the linter that flags commands.)

Mathlib/Tactic/Linter/MinImports.lean
adomani: Corrected the mistake and explained a little better.

test/MinImports.lean
adomani: This is a little tricky: the linter knows what are the imports-so-far, only if it has been diligently increasing its "import counter" with every command that it saw.  A switch like the one that you are suggesting would mean that the linter should have updated its counter also when it was "off".

I think that this feature is not for this linter in its current form.  If you want to know about whether a specific command increases imports, then you activate the linter at the start of the file and see if this command gets flagged.  If you want to know the imports of just the current command, then you can use `#min_imports in`.  Does that make sense?

adomani: What I have not yet decided, is whether the linter should consider `set_option`s as something that increases imports.  In some sense, it currently does and does not: it notices that a `set_option` is a command and adds `Lean.Parser.Command` to the imports, but it does not scan the state for what option it is, so it does not find the file where the option is defined.

adomani: Re-reading your comment, it could also be interpreted as: can we make sure that the first command after `set_option linter.minImports true` is *never* flagged and considered like a "reference" command.  This could be done: for instance the `minImportsRef` could be an `Option NameSet`, starting with `none` and, when it changes from `none` to `some ...`, the linter says nothing, but a change from `some ...` to `some (something else)` makes the linter speak.

Mathlib/Tactic/Linter/MinImports.lean
```diff
@@ -0,0 +1,109 @@
...
+      let explicitImportsInFile : NameSet :=
+        .fromArray (((← getEnv).imports.map (·.module)).erase `Init) Name.quickCmp
+      let newImps := importsSoFar.diff explicitImportsInFile
+      let currentlyUnneededImports := explicitImportsInFile.diff importsSoFar
+      -- we read the current file, to do a custom parsing of the imports:
+      -- this is a hack to obtain some `Syntax` information for the `import X` commands
+      let fil ← IO.FS.readFile (← getFileName)
+      for i in currentlyUnneededImports do
+        -- looking for the position of `import i`, so we split at ` i` and
+        -- compute the length of the string until `...import| i`
```
semorrison, line 77: These should all be `"`, not backticks.

Mathlib/Tactic/Linter/MinImports.lean
```diff
@@ -0,0 +1,109 @@
...
+        -- format the imports prepending `import ` to each module name
+        let withImport := (newImps.toArray.qsort Name.lt).map (s!"import {·}")
+        -- create a syntax node supported, likely on the first imported module name
+        let firstImport : Syntax := match fil.splitOn "import " with
+          | a::_::_ => .ofRange ⟨⟨a.length⟩, ⟨a.length + "import".length⟩⟩
+          | _ => .ofRange ⟨⟨0⟩, ⟨19⟩⟩
+        logWarningAt firstImport m!"-- missing imports\n{"\n".intercalate withImport.toList}"
+    let id ← getId stx
+    let newImports := getIrredundantImports (← getEnv) (← getAllImports stx id)
+    let tot := (newImports.append importsSoFar) --.erase `Lean.Parser.Command
```
semorrison, line 96: Stray inscrutable comment here?

semorrison:
I'm happy (excited!) to try this out in production.

bors d+

PR #14844, opened 2024-07-17
feat(Dynamics.Flow): add a notion of invariance under a map

D-Thomine:
The file had a notion of invariance under a family of maps or a monoid action.

Actions of Nat are determined by a single action. ~We give a simpler version `IsMapInvariant` of invariance in this case, and prove its equivalence with the more general one.~ This is the new lemma `mapsTo_self_iff_isinvariant_nat`

I did some additional cleaning (made some variables implicit and added a `mem` lemma for easier use).

~Note: The current state of Dynamics.Flow is not very convincing, as it is half about invariance and half about flows. It should probably be split in two files (and the section about invariance expanded). In the meantime, this PR is a quick fix for another project.~

---

### DIFF
--- a/Mathlib/Dynamics/Flow.lean
+++ b/Mathlib/Dynamics/Flow.lean
@@ -49,12 +49,8 @@ theorem isInvariant_iff_image {ϕ : τ → α → α} {s : Set α} :
   simp_rw [IsInvariant, mapsTo']
 #align is_invariant_iff_image isInvariant_iff_image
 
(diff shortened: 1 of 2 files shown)


urkud:
There are different notions that compete for this name. E.g., another one is `f ⁻¹' s = s`.

D-Thomine:
@urkud : Fair enough. I'm not completely satisfied with the current state, but I see your point about avoiding redundant definitions.

I've removed the new definition, and kept only some basic stuff:

- implicit variables in lemmas
- a `mem` lemma for invariance
- a simpler criterion for Nat actions (but reformulated with MapsTo).

PR #14850, opened 2024-07-17
[Merged by Bors] - chore: replace 'import Batteries' with finer imports

kim-em:


mattrobball:
bors delegate+

semorrison:
bors merge

semorrison:
bors merge

semorrison:
bors merge

.github/workflows/bors.yml
```diff
@@ -248,7 +248,10 @@ jobs:
         uses: liskin/gh-problem-matcher-wrap@v3
         with:
           linters: gcc
-          run: lake test
+          run: |
+            # make sure everything is available for import_all.lean
```
grunweg, line 252: There's no file `import_all.lean` in mathlib (and I don't find the relation to `mk_all` obvious; that doesn't require building batteries). Would you like to clarify this comment?
(The same applies to the three other copies of this comment.)

semorrison: `tests/import_all.lean`. I'll update the comment.

semorrison:
bors merge

.github/workflows/bors.yml
grunweg: :bulb: Thanks!

PR #14852, opened 2024-07-17
[Merged by Bors] - feat: generalize Siegel's lemma to `A = 0`

digama0:
The Wikipedia statement of Siegel's lemma did not contain the hypothesis `A != 0` in the lean formalization. A slight variant of the statement is still true if `A = 0`. The original statement is recovered under the assumption `A != 0`.

Co-authored-by: Michail Karatarakis <40603357+mkaratarakis@users.noreply.github.com>

---


fbarroero:
The Wikipedia [page](https://en.wikipedia.org/wiki/Siegel%27s_lemma) actually says that the coefficients of the linear system are not all zero, so the hypothesis `A!=0` is actually there.
In any case, I see no harm in this generalisation, so I am ok with that.

riccardobrasca:
Thanks!

bors merge

PR #14854, opened 2024-07-18
[Merged by Bors] - feat(SetTheory/Cardinal/Finite): More missing `Nat.card` lemmas

tb65536:
Building on #14825, this PR adds some missing lemmas relating `Nat.card` and surjectivity.

---


dupuisf (commented):
Thanks!

bors r+

PR #14857, opened 2024-07-18
[Merged by Bors] - feat(RingTheory/Derivation/Basic): add lemmas about the derivation of a fraction

Command-Master:
Add lemmas about the derivation of a fraction in a field.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14860, opened 2024-07-18
[Merged by Bors] - feat(Mathlib/FieldTheory): define differential fields

Command-Master:
Define logarithmic derivatives and prove some properties.

---

- [x] depends on: #14857 
- [x] depends on: #14699


Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,146 @@
...
+  unfold logd
+  field_simp
+  ring
+
+lemma logd_sub (h₁ : a ≠ 0) (h₂ : b ≠ 0) : logd a - logd b = logd (a / b) := by
+  unfold logd
+  field_simp [Derivation.leibniz_div]
+  ring
+
+lemma nat_mul_logd (a : ℕ) : a * logd b = logd (b ^ a) := by
```
YaelDillies, line 50: I would turn these three lemmas around

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,146 @@
...
+  | succ a h2 =>
+  rw [Nat.cast_add, Nat.cast_one, add_mul, one_mul, h2, logd_add, pow_succ] <;>
+  simp [h]
+
+lemma logd_eq_zero_iff : logd a = 0 ↔ a′ = 0 :=
+  ⟨fun h ↦ by simp only [logd, div_eq_zero_iff] at h; rcases h with h|h <;> simp [h],
+  fun h ↦ by unfold logd at *; simp [h]⟩
+
+lemma Multiset.logd_sum (ι : Type*) (s : Multiset ι) (f : ι → R) (h : ∀ x ∈ s, f x ≠ 0) :
+    (s.map (fun x ↦ logd (f x))).sum = logd (s.map f).prod := by
```
YaelDillies, line 64:
```suggestion
lemma logd_multisetProd {ι : Type*} (s : Multiset ι) (h : ∀ x ∈ s, f x ≠ 0) :
    logd (s.map f).prod = (s.map fun x ↦ logd (f x)).sum := by
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,146 @@
...
+    apply logd_add
+    simp [h]
+    apply Multiset.prod_ne_zero
+    all_goals simp_all
+
+lemma logd_sum (ι : Type*) (s : Finset ι) (f : ι → R) (h : ∀ x ∈ s, f x ≠ 0) :
+    ∑ x ∈ s, logd (f x) = logd (∏ x ∈ s, f x) := Multiset.logd_sum _ _ _ h
+
+lemma logd_sum_zero (ι : Type*) (s : Finset ι) (f : ι → R) (h : ∀ x ∈ s, f x = 0) :
+    ∑ x ∈ s, logd (f x) = logd (∏ x ∈ s, f x) := by
```
YaelDillies, line 79: I would turn these two around as well

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,146 @@
...
+/--
+A field extension `K / F` is Liouville if, whenever an element a ∈ F can be written as
+`a = v + ∑ cᵢ * logd uᵢ` for `v, cᵢ, uᵢ ∈ K` and `cᵢ` constant, it can also be written in that
+way with `v, cᵢ, uᵢ ∈ F`.
+-/
+class IsLiouville (F : Type u) (K : Type*) [Field F] [Field K] [Differential F] [Differential K]
+    [Algebra F K] [DifferentialAlgebra F K] : Prop where
+  is_liouville (a : F) (ι : Type u) [Fintype ι] (c : ι → F) (hc : ∀ x, (c x)′ = 0)
+    (u : ι → K) (v : K) (h : a = ∑ x, c x * logd (u x) + v′) :
+    ∃ (ι' : Type u) (_ : Fintype ι') (c' : ι' → F) (_ : ∀ x, (c' x)′ = 0)
```
YaelDillies, line 105:
```suggestion
    ∃ (ι' : Type) (_ : Fintype ι') (c' : ι' → F) (_ : ∀ x, (c' x)′ = 0)
```
since all fintypes are `0`-small. First thing you should do is show that this is equivalent to `∃ (ι' : Type u) (_ : Fintype ι') (c' : ι' → F) (_ : ∀ x, (c' x)′ = 0), ...` for any universe `u`. It will save you from the universe gymnastic you're currently performing in `IsLiouville.trans`

YaelDillies (commented):
Don't you think that the log derivative and Liouville extensions should live in separate files? Certainly, I have heard of the first but never of the second

Mathlib/FieldTheory/Differential/Basic.lean
Command-Master: Should I also have `(ι : Type)` as the argument?

YaelDillies: Yes, for the same reason

Command-Master:
> Don't you think that the log derivative and Liouville extensions should live in separate files? Certainly, I have heard of the first but never of the second

That makes sense, I'll move it.
By the way, Liouville extensions are something I came up with for field extensions which satisfy (a slight generalization of) Liouville's theorem, they don't appear in the literature that I know of. Perhaps I should add a comment about that.

YaelDillies:
Ah yes, definitely mention that!

Mathlib.lean
```diff
@@ -2606,6 +2606,7 @@ import Mathlib.FieldTheory.Adjoin
 import Mathlib.FieldTheory.AxGrothendieck
 import Mathlib.FieldTheory.Cardinality
 import Mathlib.FieldTheory.ChevalleyWarning
+import Mathlib.FieldTheory.Differential.Basic
 import Mathlib.FieldTheory.Extension
 import Mathlib.FieldTheory.Finite.Basic
 import Mathlib.FieldTheory.Finite.GaloisField
```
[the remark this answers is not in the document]
Command-Master: bot fix style

Command-Master:
bot fix style

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,146 @@
...
+-/
+
+open Differential algebraMap
+
+variable {R : Type*} [Field R] [Differential R] (a b : R)
+
+/--
+The logarithmic derivative of a is a′ / a.
+-/
+def logd : R := a′ / a
```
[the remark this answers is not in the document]
Command-Master: logDeriv is already used in `Mathlib.Analysis.Calculus.LogDeriv`, so I'm changing it back to `logd`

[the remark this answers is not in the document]
YaelDillies: Oh, but that's good because they are morally the same object. What it means instead is that your API should be namespaced (under `Derivation` probably?)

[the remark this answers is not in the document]
Command-Master: I think `Differential` is a better namespace

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,92 @@
...
+-/
+import Mathlib.Algebra.EuclideanDomain.Field
+import Mathlib.RingTheory.Derivation.DifferentialRing
+import Mathlib.RingTheory.LocalRing.Basic
+import Mathlib.Tactic.FieldSimp
+
+/-!
+# Differential Fields
+
+This file defines the logarithmic derivative `logDeriv` and proves properties of it.
```
YaelDillies, line 14:
```suggestion
This file defines the logarithmic derivative `Differential.logDeriv` and proves properties of it.
```
and can you add a sentence about how this differs from `logDeriv`?

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,92 @@
...
+lemma logDeriv_zero : logDeriv (0 : R) = 0 := by
+  unfold logDeriv
+  simp
+
+@[simp]
+lemma logDeriv_one : logDeriv (1 : R) = 0 := by
+  unfold logDeriv
+  simp
+
+lemma logDeriv_mul (h₁ : a ≠ 0) (h₂ : b ≠ 0) : logDeriv (a * b) = logDeriv a + logDeriv b := by
```
YaelDillies, line 38:
```suggestion
lemma logDeriv_mul (ha : a ≠ 0) (hb : b ≠ 0) : logDeriv (a * b) = logDeriv a + logDeriv b := by
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,92 @@
...
+  unfold logDeriv
+  field_simp
+  ring
+
+lemma logDeriv_div (h₁ : a ≠ 0) (h₂ : b ≠ 0) : logDeriv (a / b) = logDeriv a - logDeriv b := by
+  unfold logDeriv
+  field_simp [Derivation.leibniz_div]
+  ring
+
+lemma logDeriv_npow (a : ℕ) : logDeriv (b ^ a) = a * logDeriv b := by
```
YaelDillies, line 48:
```suggestion
lemma logDeriv_pow (n : ℕ) : logDeriv (a ^ n) = n * logDeriv n := by
```
and yet again probably should be a simp lemma

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,92 @@
...
+  by_cases h : b = 0
+  · cases a <;> simp [h]
+  induction a with
+  | zero => simp
+  | succ a h2 =>
+  rw [Nat.cast_add, Nat.cast_one, add_mul, one_mul, ← h2, pow_succ, logDeriv_mul] <;>
+  simp [h]
```
YaelDillies, line 55:
```suggestion
  induction a with
  | zero => simp
  | succ a h2 =>
    obtain rfl | hb := eq_or_ne b 0
    · simp
    · rw [Nat.cast_add, Nat.cast_one, add_mul, one_mul, ← h2, pow_succ, logDeriv_mul] <;>
      simp [hb]
```
is slightly more efficient in terms of case splits

Mathlib/FieldTheory/Differential/Liouville.lean
YaelDillies, line 1: As previously stated, I would rather have this file be a separate PR. I'm not sure I am competent to review it and its content is significantly different from the other new file

Mathlib/FieldTheory/Differential/Basic.lean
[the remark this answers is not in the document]
YaelDillies: `Differential` might be a bit unclear given that it also applies to the other `logDeriv`

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,94 @@
...
+
+/--
+The logarithmic derivative of a is a′ / a.
+-/
+def logDeriv : R := a′ / a
+
+@[simp]
+lemma logDeriv_zero : logDeriv (0 : R) = 0 := by
+  unfold logDeriv
+  simp
```
YaelDillies, line 32: Nitpick, but the idiomatic pattern is rather
```suggestion
  simp [logDeriv]
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,94 @@
...
+  field_simp
+  ring
+
+lemma logDeriv_div (ha : a ≠ 0) (hb : b ≠ 0) : logDeriv (a / b) = logDeriv a - logDeriv b := by
+  unfold logDeriv
+  field_simp [Derivation.leibniz_div]
+  ring
+
+@[simp]
+lemma logDeriv_pow (n : ℕ) : logDeriv (b ^ n) = n * logDeriv b := by
```
YaelDillies, line 50:
```suggestion
lemma logDeriv_pow (n : ℕ) (a : R) : logDeriv (a ^ n) = n * logDeriv a := by
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,94 @@
...
+  induction n with
+  | zero => simp
+  | succ n h2 =>
+    obtain rfl | hb := eq_or_ne b 0
+    · simp
+    · rw [Nat.cast_add, Nat.cast_one, add_mul, one_mul, ← h2, pow_succ, logDeriv_mul] <;>
+      simp [hb]
+
+lemma logDeriv_eq_zero_iff : logDeriv a = 0 ↔ a′ = 0 :=
```
YaelDillies, line 59:
```suggestion
lemma logDeriv_eq_zero : logDeriv a = 0 ↔ a′ = 0 :=
```
since it's easily inferred that it's an iff

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+import Mathlib.Algebra.EuclideanDomain.Field
+import Mathlib.RingTheory.Derivation.DifferentialRing
+import Mathlib.RingTheory.LocalRing.Basic
+import Mathlib.Tactic.FieldSimp
+
+/-!
+# Differential Fields
+
+This file defines the logarithmic derivative `Differential.logDeriv` and proves properties of it.
+This is defined algebraiclly, compared to `logDeriv` which is analytical.
```
YaelDillies, line 15:
```suggestion
This is defined algebraically, compared to `logDeriv` which is analytical.
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+def logDeriv : R := a′ / a
+
+@[simp]
+lemma logDeriv_zero : logDeriv (0 : R) = 0 := by
+  simp [logDeriv]
+
+@[simp]
+lemma logDeriv_one : logDeriv (1 : R) = 0 := by
+  unfold logDeriv
+  simp
```
YaelDillies, line 36:
```suggestion
  simp [logDeriv]
```

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+    rw [← h₂]
+    apply logDeriv_mul
+    simp [h]
+    apply Multiset.prod_ne_zero
+    all_goals simp_all
+
+lemma logDeriv_prod (ι : Type*) (s : Finset ι) (f : ι → R) (h : ∀ x ∈ s, f x ≠ 0) :
+    logDeriv (∏ x ∈ s, f x) = ∑ x ∈ s, logDeriv (f x) := logDeriv_multisetProd _ h
+
+lemma logDeriv_prod_of_zero (ι : Type*) (s : Finset ι) (f : ι → R) (h : ∀ x ∈ s, f x = 0) :
```
YaelDillies, line 77:
```suggestion
lemma logDeriv_prod_of_eq_zero (ι : Type*) (s : Finset ι) (f : ι → R) (h : ∀ x ∈ s, f x = 0) :
```
for clarity, maybe

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/FieldTheory/Differential/Basic.lean
```diff
@@ -0,0 +1,92 @@
...
+
+lemma logDeriv_multisetProd {ι : Type*} (s : Multiset ι) {f : ι → R} (h : ∀ x ∈ s, f x ≠ 0) :
+    logDeriv (s.map f).prod = (s.map fun x ↦ logDeriv (f x)).sum := by
+  induction s using Multiset.induction_on
+  · simp
+  · rename_i h₂
+    simp only [Function.comp_apply, Multiset.map_cons, Multiset.sum_cons, Multiset.prod_cons]
+    rw [← h₂]
+    apply logDeriv_mul
+    simp [h]
```
semorrison, line 69: Please convert this to a simp only

semorrison:
bors d+

Mathlib/FieldTheory/Differential/Basic.lean
Command-Master: This isn't nonterminal, there are simply multiple goals, so I structured the proof

PR #14862, opened 2024-07-18
[Merged by Bors] - chore(workflows/nightly_detect_failure): only compare initial parts of messages

jcommelin:
Comparing entire messages does not seem to work.
Maybe because Zulip removed the trailing newline?
Comparing the first 160 chars should be enough.

---


adomani:
You seem to have merge conflicts with the other attempt to print more.  Is it possible to print to a shell what `messages` is?  I think that looking at that may be helpful.

jcommelin:
@adomani I don't exactly know what you mean with printing to a shell.

Please feel free to push changes to this PR

adomani:
What I meant is what you already did with the `print` command: I just could not find where it was printing and was confused.

I agree that the issue seems to be some form of removal of trailing whitespace that makes the two strings different, even though they are not.  Let's try the comparison of 160 characters!

maintainer merge

(Note that I also added to the script a "confirmation print", in case the strings do match.)

semorrison:
bors merge

PR #14867, opened 2024-07-18
[Merged by Bors] - feat(CategoryTheory/Sites): constructors for Mayer-Vietoris squares

joelriou:
In this PR, we give constructors for Mayer-Vietoris squares, which allows the construction of these squares in the case of topological spaces.

---

- [x] depends on: #14959
- [x] depends on: #14957
- [x] depends on: #14951
- [x] depends on: #14890
- [x] depends on: #14935 
- [x] depends on: #14936


Mathlib/CategoryTheory/Sites/MayerVietorisSquare.lean
```diff
@@ -52,10 +51,28 @@ namespace CategoryTheory
 
 open Limits Opposite
 
+variable {C : Type u} [Category.{v} C]
+  {J : GrothendieckTopology C} [HasWeakSheafify J (Type v)]
+
+lemma Sheaf.isPullback_square_op_map_yoneda_presheafToSheaf_yoneda_iff
```
jcommelin, line 57: Should this even be a simp lemma?

jcommelin (commented):
LGTM

bors d+

Mathlib/CategoryTheory/Sites/MayerVietorisSquare.lean
joelriou: Thanks for the suggestion!

joelriou:
Thanks for the review!

bors merge

PR #14875, opened 2024-07-18
[Merged by Bors] - feat (Data.Finsupp.Weight) : general API for weights and size of finsupp

AntoineChambert-Loir:
We collect various definitions and lemma that are presently used for multivariate polynomials to define/study their degree / weighted degree in a single file 
`Mathlib.Data.Finsupp.Weight``

The reason is that this API will also be used for multivariate power series.

We have tried for some generality, but it is not the maximal one.

Adds a`SMulPosMono ℕ M` instance for any `OrderedAddCommMonoid M`

Co-authored-by María Inés de Frutos Fernández @mariainesdff

Moves:
- `MvPolynomial.degree -> Finsupp.degree`
- `MvPolynomial.weightedDegree -> Finsupp.weight`
- `MvPolynomial.degree_eq_zero_iff -> Finsupp.degree_eq_zero_iff`
- `MvPolynomial.weightedDegree_one -> Finsupp.degree_eq_weight_one`
- `MvPolynomial.weight_eq_zero_iff_eq_zero_iff -> Finsupp.weight_eq_zero_iff_eq_zero`

---


Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+
+section AddCommMonoid
+
+variable [AddCommMonoid M]
+/-- The `weight` of the finitely supported function `s : σ →₀ ℕ`
+with respect to `w : σ → M` is the sum `∑(s i)•(w i)`. -/
+noncomputable def weight : (σ →₀ ℕ) →+ M :=
+  (Finsupp.total σ M ℕ w).toAddMonoidHom
+
+theorem weight_apply (f : σ →₀ ℕ) :
```
riccardobrasca, line 27: I guess `simps` can generate this.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+    weight w f = Finsupp.sum f (fun i c => c • w i) := rfl
+
+/-- A weight function is nontorsion if its values are not torsion. -/
+class NonTorsionWeight (w : σ → M) : Prop where
+  eq_zero_of_smul_eq_zero {n : ℕ} {s : σ} (h : n • w s = 0)  : n = 0
```
riccardobrasca, line 32:
```suggestion
  eq_zero_of_smul_eq_zero {n : ℕ} {s : σ} (h : n • w s = 0) : n = 0
```

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+    w s ≠ 0 := fun h ↦ by
+  rw [← one_smul ℕ (w s)] at h
+  apply Nat.zero_ne_one.symm
+  exact NonTorsionWeight.eq_zero_of_smul_eq_zero h
+
+end AddCommMonoid
+
+section OrderedAddCommMonoid
+
+theorem Nat.le_weight (w : σ → ℕ) (s : σ) (hs : w s ≠ 0) (f : σ →₀ ℕ) :
```
riccardobrasca, line 51: What about just `le_weight`? I don't get the logic of the name here.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+
+variable [OrderedAddCommMonoid M] (w : σ → M)
+
+instance : SMulPosMono ℕ M := ⟨
+  fun b hb m m' h ↦ by
+    rw [← Nat.add_sub_of_le h, add_smul]
+    exact le_add_of_nonneg_right (nsmul_nonneg hb (m' - m)) ⟩
+
+variable {w} in
+theorem le_weight_of_nonneg' (hw : ∀ s, 0 ≤ w s) (s : σ) (f : σ →₀ ℕ) (hs : f s ≠ 0) :
```
riccardobrasca, line 72: Here `s` can be implicit, but as you prefer.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+    exact le_add_of_nonneg_right <| Finset.sum_nonneg <|
+      fun i _ ↦ nsmul_nonneg (hw i) (f i)
+
+end OrderedAddCommMonoid
+
+section CanonicallyOrderedAddCommMonoid
+
+variable {M : Type*} [CanonicallyOrderedAddCommMonoid M] (w : σ → M)
+
+theorem le_weight' (s : σ) (f : σ →₀ ℕ) (hs : f s ≠ 0) :
```
riccardobrasca, line 90: Same here, a bunch of variables can be implicit.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,131 @@
...
+    w s ≤ weight w f :=
+  le_weight_of_nonneg' w (fun _ ↦ zero_le _) s f hs
+
+/-- If `M` is a `CanonicallyOrderedAddCommMonoid`, then the `weightedHomogeneousComponent`
+  of weighted degree `0` of a polynomial is its constant coefficient. -/
```
riccardobrasca, line 95: I get the point, but there are no polynomials in the statement.

riccardobrasca (commented):
Can you add deprecation alias for the four declaration you removed? Those are

```
- degree
- weightedDegree
- weightedDegree_apply
- weightedDegree_one
```

Also, we are trying to write this in the PR message. From the [contribute](https://leanprover-community.github.io/contribute/index.html) page:



    If you are moving or deleting declarations, please include these lines at the bottom of the commit message (that is, before the ---) using the following format:

Moves:

    Vector.* -> Mathlib.Vector.*
    ...

Deletions:

    Nat.bit1_add_bit1
    ...

Mathlib/Data/Finsupp/Weight.lean
AntoineChambert-Loir: because it's only for Nat (in general the weight can be AddCommMonoid-valued)

Mathlib/Data/Finsupp/Weight.lean
AntoineChambert-Loir: It does not generate the same result, it is nice to go way down to `Finsupp.sum`

Mathlib/Data/Finsupp/Weight.lean
AntoineChambert-Loir: Yes, I also made `f` implicit

Mathlib/Data/Finsupp/Weight.lean
AntoineChambert-Loir: Anyway, I changed it to `le_weight`

AntoineChambert-Loir:
Addition of the `deprecation alias`es had a naughty side effect that `Finsupp.degree` and `MvPolynomial.degree` are confused one to another when both namespaces are open.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,196 @@
...
+    weight w f = Finsupp.sum f (fun i c => c • w i) := rfl
+
+@[deprecated weight_apply (since := "2024-07-20")]
+alias _root_.MvPolynomial.weightedDegree_apply := weight_apply
+
+/-- A weight function is nontorsion if its values are not torsion. -/
+class NonTorsionWeight (w : σ → M) : Prop where
+  eq_zero_of_smul_eq_zero {n : ℕ} {s : σ} (h : n • w s = 0)  : n = 0
+
+/-- Without zero divisors, nonzero weight is a NonTorsionWeight -/
```
riccardobrasca, line 91:
```suggestion
/-- Without zero divisors, nonzero weight is a `NonTorsionWeight` -/
```

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,196 @@
...
+  · simp only [not_mem_support_iff] at h
+    rw [h]
+    apply zero_le
+
+variable [OrderedAddCommMonoid M] (w : σ → M)
+
+instance : SMulPosMono ℕ M := ⟨
+  fun b hb m m' h ↦ by
+    rw [← Nat.add_sub_of_le h, add_smul]
+    exact le_add_of_nonneg_right (nsmul_nonneg hb (m' - m)) ⟩
```
riccardobrasca, line 126:
```suggestion
instance : SMulPosMono ℕ M :=
  ⟨fun b hb m m' h ↦ by
    rw [← Nat.add_sub_of_le h, add_smul]
    exact le_add_of_nonneg_right (nsmul_nonneg hb (m' - m))⟩
```
I think is the preferred style.

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,196 @@
...
+
+variable {w} in
+theorem le_weight_of_nonneg' (hw : ∀ s, 0 ≤ w s) {s : σ} {f : σ →₀ ℕ} (hs : f s ≠ 0) :
```
riccardobrasca, line 129:
```suggestion
theorem le_weight_of_ne_zero (hw : ∀ s, 0 ≤ w s) {s : σ} {f : σ →₀ ℕ} (hs : f s ≠ 0) :
```

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,196 @@
...
+    exact le_add_of_nonneg_right <| Finset.sum_nonneg <|
+      fun i _ ↦ nsmul_nonneg (hw i) (f i)
+
+end OrderedAddCommMonoid
+
+section CanonicallyOrderedAddCommMonoid
+
+variable {M : Type*} [CanonicallyOrderedAddCommMonoid M] (w : σ → M)
+
+theorem le_weight' {s : σ} {f : σ →₀ ℕ} (hs : f s ≠ 0) :
```
riccardobrasca, line 147:
```suggestion
theorem le_weight_of_ne_zero' {s : σ} {f : σ →₀ ℕ} (hs : f s ≠ 0) :
```

Mathlib/Data/Finsupp/Weight.lean
```diff
@@ -0,0 +1,196 @@
...
+alias _root_.MvPolynomial.degree := degree
+
+lemma degree_eq_zero_iff (d : σ →₀ ℕ) : degree d = 0 ↔ d = 0 := by
+  simp only [degree, Finset.sum_eq_zero_iff, Finsupp.mem_support_iff, ne_eq, Decidable.not_imp_self,
+    DFunLike.ext_iff, Finsupp.coe_zero, Pi.zero_apply]
+
+@[deprecated degree_eq_zero_iff (since := "2024-07-20")]
+alias _root_.MvPolynomial.degree_eq_zero_iff := degree_eq_zero_iff
+
+theorem degree_zero : degree (0 : σ →₀ ℕ) = 0 := by rw [degree_eq_zero_iff]
```
riccardobrasca, line 182: I remember you already answered, but why not `simp`? Anyway not very important.

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Data/Finsupp/Weight.lean
AntoineChambert-Loir: Let me add it…

riccardobrasca:
There is now a linter against `refine'`.

PR #14878, opened 2024-07-18
[Merged by Bors] - fix: handle nameless instances in `#min_imports in`

adomani:
This PR makes `#min_imports in` process better "nameless" instances, as well as attributes in declarations.

---


Mathlib/Tactic/MinImports.lean
```diff
@@ -37,20 +37,65 @@ def getVisited (env : Environment) (decl : Name) : NameSet :=
 
 /-- `getId stx` takes as input a `Syntax` `stx`.
 If `stx` contains a `declId`, then it returns the `ident`-syntax for the `declId`.
+If `stx` is a nameless instance, then it also tries to fetch the `ident` for the instance.
 Otherwise it returns `stx` itself. -/
-def getId (stx : Syntax) : Syntax :=
+def getId (stx : Syntax) : CommandElabM Syntax := do
+  -- if the command contains a `declId`, we use the implied `ident`
```
grunweg, line 43:
```suggestion
  -- If the command contains a `declId`, we use the implied `ident`.

File: Mathlib/Tactic/MinImports.lean [hunk 2]
```diff
@@ -37,20 +37,65 @@ def getVisited (env : Environment) (decl : Name) : NameSet :=
...
-    | none => stx
-
-/--`getAllImports cmd` takes a `Syntax` input `cmd` and returns the `NameSet` of all the
-module names that are implied by the `SyntaxNodeKinds` and the identifiers contained in `cmd`. -/
-def getAllImports {m : Type → Type} [Monad m] [MonadResolveName m] [MonadEnv m]
-    (cmd : Syntax) (dbg? : Bool := false) :
-    m NameSet := do
+    | some declId => return declId[0]
+    | none =>
+      -- otherwise, there is the possibility that the command is a nameless `instance`
```
grunweg, line 47: What do you think?
```suggestion
      -- Otherwise, there the command could be a nameless `instance`.

File: Mathlib/Tactic/MinImports.lean [hunk 3]
```diff
@@ -37,20 +37,65 @@ def getVisited (env : Environment) (decl : Name) : NameSet :=
...
+      match stx.find? (·.isOfKind ``Lean.Parser.Command.instance) with
+                | none => return stx
```
grunweg, line 49: That's some heavy indentation. From my side, four to six spaces less would also do.
(Not sure if that's in the style guide. I try to indent match patters by two spaces.)

Mathlib/Tactic/MinImports.lean
```diff
@@ -37,20 +37,65 @@ def getVisited (env : Environment) (decl : Name) : NameSet :=
...
+                | some stx => do
+                  -- if it is a nameless `instance`, we retrieve the autogenerated name
```
grunweg, line 51: Also optional: 
```suggestion
                  -- If it is a nameless `instance`, we retrieve the autogenerated name.
```

or
```suggestion

```suggestion
                  -- In that case, we retrieve the autogenerated name.

File: Mathlib/Tactic/MinImports.lean [hunk 5]
```diff
@@ -37,20 +37,65 @@ def getVisited (env : Environment) (decl : Name) : NameSet :=
...
+* if `cmd` adds a declaration `d` to the environment, then also all the module names implied by `d`.
+The argument `id` is optional and is expected to be the internally generated name of a "nameless"
+`instance`.
+-/
+def getAllImports (cmd : Syntax) (id : Syntax := default) (dbg? : Bool := false) :
+    CommandElabM NameSet := do
   let env ← getEnv
   --let nm ← liftCoreM do realizeGlobalConstNoOverloadWithInfo (getId cmd)
-  let ts := (getSyntaxNodeKinds cmd).append <| getVisited env (getId cmd).getId
+  -- we put together the implied declaration names, the `SyntaxNodeKinds`, the attributes
```
grunweg, line 94:
```suggestion
  -- We collect the implied declaration names, the `SyntaxNodeKinds` and the attributes.
```
That said, reading this comment leaves me wondering "which attributes" or "that attributes of what"...

Mathlib/Tactic/MinImports.lean
```diff
@@ -100,8 +144,11 @@ syntax "#min_imports in" term : command
 
 elab_rules : command
   | `(#min_imports in $cmd:command)=> do
+    -- in case `cmd` is a "nameless" `instance`, we produce its name
```
grunweg, line 147: (Can you elaborate why this is important, if that's brief?)
```suggestion
    -- In case `cmd` is a "nameless" `instance`, we produce its name.
    -- It is important that this is collected *before* adding the declaration to the environment.

File: test/MinImports.lean [hunk 7]
```diff
@@ -5,6 +5,12 @@ import Mathlib.Data.Nat.Notation
 import Mathlib.Data.Int.Notation
 import Mathlib.Tactic.NormNum.Basic
 
+/-- info: import Mathlib.Algebra.Ring.Nat -/
+#guard_msgs in
+#min_imports in
+-- if `#min_imports` were parsing just the syntax, the imports would be `Mathlib.Algebra.Ring.Defs`
```
grunweg, line 11:
```suggestion
-- If `#min_imports` were parsing just the syntax, the imports would be `Mathlib.Algebra.Ring.Defs`.

**Review by grunweg — COMMENTED** [review]
Nice find! I gave it a cursory read, just style tweaks so far.

Does this fix all bugs of "not enough imports" so far? (In any case, looks like useful progress.)

File: Mathlib/Tactic/MinImports.lean (hunk 2 above)
Comment (adomani) [reply Mathlib/Tactic/MinImports.lean:47]:
```suggestion
      -- Otherwise, the command could be a nameless `instance`.
```

Mathlib/Tactic/MinImports.lean
adomani: Oh, this is a left-over from when the `match` was on the previous line!

Mathlib/Tactic/MinImports.lean
adomani: These are the attributes of the declaration, e.g. `@[fun_prop] ...`.  Should I say more about this?

test/MinImports.lean
adomani: I added an explanation.

adomani:
> Does this fix all bugs of "not enough imports" so far? (In any case, looks like useful progress.)

I am not sure: I found these bugs by converting the command into a linter and bulk-processing a lot of mathlib files.  However, in the conversion, I still do not know how to handle in the linter the "nameless `instance`" issue.  So, the *linter* still shows differences, but I am not sure what would happen if I actually tested the *command*.

Mathlib/Tactic/MinImports.lean
semorrison: I think this is okay.

PR #14881, opened 2024-07-18
[Merged by Bors] - chore(Filter.Germ): split off bundled ordered algebra instances 

mattrobball:
We split `Order.Filter.Germ` into `Basic` and `OrderedMonoid` to allow for explicitly importing bundled ordered algebra instances when desired.

---


Mathlib/Order/Filter/Germ/OrderedMonoid.lean
```diff
@@ -0,0 +1,44 @@
...
+Copyright (c) 2020 Yury G. Kudryashov. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yury G. Kudryashov, Abhimanyu Pallavi Sudhir
+-/
+import Mathlib.Algebra.Module.Pi
+import Mathlib.Algebra.Order.Monoid.Canonical.Defs
+import Mathlib.Order.Filter.Germ.Basic
+
+/-!
+# Ordered monoid instances on germ of a function at a filter
```
grunweg, line 11:
```suggestion
# Ordered monoid instances on the space of germs of a function at a filter

File: Mathlib/Order/Filter/Germ/OrderedMonoid.lean [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+
+For each of the following structures we prove that if `β` has this structure, then so does
```
grunweg, line 13: You could also say
"We prove that if `β` has a `OrderedCancelCommMonoid` or `OrderedCancelAddCommMonoid` structure, respectively, then so does the space `Germ l β`."

Which one is better depends on whether you think this file will gain further structures in the future...

grunweg (commented):
Two small comments, otherwise looks good to me!

Mathlib/Order/Filter/Germ/OrderedMonoid.lean
mattrobball: I see the advantage to this vis-a-vis robustness but there is also value in declaratively. I think I'll leave it for now. Thanks.

grunweg:
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors d+

PR #12533, opened 2024-04-30
[Merged by Bors] - feat: `IsHomeomorph` predicate

peabrainiac:
Define an `IsHomeomorph` predicate for maps between topological spaces to complement the existing bundled `Homeomorph` API.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14882, opened 2024-07-18
[Merged by Bors] - feat(AlgebraicGeometry): Being an isomorphism is local at the target.

erdOne:
---

### DIFF
--- a/Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
+++ b/Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
@@ -1,0 +1,45 @@
+/-
+Copyright (c) 2022 Andrew Yang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Andrew Yang
+-/
+import Mathlib.AlgebraicGeometry.Morphisms.OpenImmersion
+import Mathlib.Topology.IsLocalHomeomorph
+
+/-!
+
+# Being an isomorphism is local at the target
+
+-/
+
+open CategoryTheory MorphismProperty
+
+namespace AlgebraicGeometry
+
+lemma isomorphisms_eq_isOpenImmersion_inf_surjective :
+    isomorphisms Scheme = (@IsOpenImmersion ⊓ @Surjective : MorphismProperty Scheme) := by
+  ext
+  exact (isIso_iff_isOpenImmersion _).trans
+    (and_congr Iff.rfl ((TopCat.epi_iff_surjective _).trans (surjective_iff _).symm))
+
+lemma isomorphisms_eq_stalkwise :
+    isomorphisms Scheme = (isomorphisms TopCat).inverseImage Scheme.forgetToTop ⊓
+      stalkwise (fun f ↦ Function.Bijective f) := by
+  rw [isomorphisms_eq_isOpenImmersion_inf_surjective, isOpenImmersion_eq_inf,
+    surjective_eq_topologically, inf_right_comm]
+  congr 1
+  ext X Y f
+  exact ⟨fun H ↦ inferInstanceAs (IsIso (TopCat.isoOfHomeo
+    (H.1.1.toHomeomeomorph_of_surjective H.2)).hom), fun (_ : IsIso f.1.base) ↦
+    let e := (TopCat.homeoOfIso <| asIso f.1.base); ⟨e.openEmbedding, e.surjective⟩⟩
+
+instance : IsLocalAtTarget (isomorphisms Scheme) :=
+  isomorphisms_eq_isOpenImmersion_inf_surjective ▸ inferInstance
+
+instance : HasAffineProperty (isomorphisms Scheme) fun X Y f _ ↦ IsAffine X ∧ IsIso (f.app ⊤) := by
+  convert HasAffineProperty.of_isLocalAtTarget (isomorphisms Scheme) with X Y f hY
+  exact ⟨fun ⟨_, _⟩ ↦ (arrow_mk_iso_iff (isomorphisms _) (arrowIsoSpecΓOfIsAffine f)).mpr
+    (inferInstanceAs (IsIso (Spec.map (f.app ⊤)))),
+    fun (_ : IsIso f) ↦ ⟨isAffine_of_isIso f, inferInstance⟩⟩
+
+end AlgebraicGeometry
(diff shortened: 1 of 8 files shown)


Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
```diff
@@ -0,0 +1,45 @@
...
+lemma isomorphisms_eq_isOpenImmersion_inf_surjective :
+    isomorphisms Scheme = (@IsOpenImmersion ⊓ @Surjective : MorphismProperty Scheme) := by
+  ext
+  exact (isIso_iff_isOpenImmersion _).trans
+    (and_congr Iff.rfl ((TopCat.epi_iff_surjective _).trans (surjective_iff _).symm))
+
+lemma isomorphisms_eq_stalkwise :
+    isomorphisms Scheme = (isomorphisms TopCat).inverseImage Scheme.forgetToTop ⊓
+      stalkwise (fun f ↦ Function.Bijective f) := by
+  rw [isomorphisms_eq_isOpenImmersion_inf_surjective, isOpenImmersion_eq_inf,
```
chrisflav, line 28: I think we should have `IsHomeomorphism` and then state this in terms of it, i.e. `isomorphisms Scheme = topologically IsHomeomorphism ⊓ stalkwise (fun f ↦ Function.Bijective f)`.

Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.lean
```diff
@@ -0,0 +1,147 @@
...
+
+section Surjective
+
+variable {X Y Z : Scheme.{u}} (f : X ⟶ Y) (g : Y ⟶ Z)
+
+/-- A morphism of schemes is surjective if the underlying map is. -/
+@[mk_iff]
+class Surjective : Prop where
+  surj : Function.Surjective f.1.base
+
```
chrisflav, line 56: I am wondering if we should agree on a convention, whether we write `f.val.base` or `f.1.base` or even `f.1.1`. I personally prefer the first version as I think it is the most readable one.

Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.lean
```diff
@@ -0,0 +1,147 @@
...
+of the underlying map, including
+
+- `Injective`
+- `Surjective`
+- `IsOpenMap`
+- `IsClosedMap`
+- `Embedding`
+- `OpenEmbedding`
+- `ClosedEmbedding`
+
```
chrisflav, line 23:
```suggestion
```

Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.lean
```diff
@@ -0,0 +1,147 @@
...
+-/
+import Mathlib.Topology.LocalAtTarget
+import Mathlib.AlgebraicGeometry.Morphisms.Constructors
+
+/-!
+
+## Properties on the underlying functions of morphisms of schemes.
+
+This file includes various results on properties of morphisms of schemes that come from properties
+of the underlying map, including
```
chrisflav, line 14:
```suggestion
of the underlying map of topological spaces, including
```

Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.lean
```diff
@@ -0,0 +1,147 @@
...
+
+lemma Surjective.comp_iff [Surjective f] : Surjective (f ≫ g) ↔ Surjective g :=
+  ⟨fun _ ↦ of_comp f g, fun _ ↦ inferInstance⟩
+
+instance : MorphismProperty.RespectsIso @Surjective :=
+  surjective_eq_topologically ▸ topologically_respectsIso _ (fun e ↦ e.surjective)
+    (fun _ _ hf hg ↦ hg.comp hf)
+
+instance surjective_isLocalAtTarget : IsLocalAtTarget @Surjective := by
+  have := AlgebraicGeometry.instRespectsIsoSchemeSurjective
```
chrisflav, line 78:
```suggestion
  have : MorphismProperty.RespectsIso @Surjective := inferInstance
```
I don't know how stable it is to rely on the automatically generated instance name.

chrisflav: (alternatively: give the `RespectsIso` instance a name)

Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
erdOne: Hmmm I think `isomorphisms` is exactly this and already has some useful API around it.

chrisflav: But in terms of category theory which does not always play nicely with the type theoretic topology library.

Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.lean
erdOne: I personally think spelling it out is a good thing but mathlib (or at least I) uses `.1` and `.val` interchangeably so I end up using `f.1.base` most of the time. I don't think the vague `val` gives much information that is worth +2 chars but I'm not against a library-wide replacement if you feel like it.

chrisflav: At least `.val` indicates that it is a subtype where `.1` can really be anything. In any case, this is not important for this PR.

Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
erdOne: There is #12533 but it seems like a stale PR.

chrisflav: Hm that's sad. Okay, let's merge this one as is and refactor later (or just add the suggested formulation) when we have `IsHomeomorphism`.

chrisflav (approved):
Thanks! LGTM

Mathlib/AlgebraicGeometry/Morphisms/IsIso.lean
erdOne: I've asked on the relevant Zulip thread whether the author intend to continue and would probably take over if the author has decided to abandon it.
I'll add the suggestion once that PR is merged. I think this lemma is still useful though and I'll probably still keep this.

jcommelin (commented):
Thanks :tada:

bors merge

PR #14884, opened 2024-07-18
[Merged by Bors] - chore(Mathlib/Lean): remove use of autoImplicit

grunweg:
All eleven files in the Lean/ directory use autoImplicit, but only in a very minor way. Removing it seems easier than making this a permanent exception.

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #14885, opened 2024-07-18
chore(Data/Vector): reduce autoImplicit

grunweg:
After this PR, `autoImplicit` in this directory is limited to three declarations.

---

### DIFF
--- a/Mathlib/Data/Vector/Basic.lean
+++ b/Mathlib/Data/Vector/Basic.lean
@@ -19,8 +19,6 @@ import Mathlib.Control.Traversable.Basic
 This file introduces the infix notation `::ᵥ` for `Vector.cons`.
 -/
 
-set_option autoImplicit true
-
 universe u
 
 variable {n : ℕ}
@@ -133,11 +131,12 @@ theorem get_map {β : Type*} (v : Vector α n) (f : α → β) (i : Fin n) :
 #align vector.nth_map Mathlib.Vector.get_map
 
 @[simp]
-theorem map₂_nil (f : α → β → γ) : Vector.map₂ f nil nil = nil :=
+theorem map₂_nil {α β γ : Type*} (f : α → β → γ) : Vector.map₂ f nil nil = nil :=
   rfl
 
 @[simp]
-theorem map₂_cons (hd₁ : α) (tl₁ : Vector α n) (hd₂ : β) (tl₂ : Vector β n) (f : α → β → γ) :
+theorem map₂_cons {α β γ : Type*}
+    (hd₁ : α) (tl₁ : Vector α n) (hd₂ : β) (tl₂ : Vector β n) (f : α → β → γ) :
     Vector.map₂ f (hd₁ ::ᵥ tl₁) (hd₂ ::ᵥ tl₂) = f hd₁ hd₂ ::ᵥ (Vector.map₂ f tl₁ tl₂) :=
   rfl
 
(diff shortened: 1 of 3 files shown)


Mathlib/Data/Vector/MapLemmas.lean
```diff
@@ -56,10 +56,12 @@ theorem map_map (f₁ : β → γ) (f₂ : α → β) :
...
 
 section Binary
-variable (xs : Vector α n) (ys : Vector β n)
+
+variable {α β : Type} {n : ℕ} (xs : Vector α n) (ys : Vector β n)
+variable {σ₁ σ₂ : Type} {s₁ : σ₁} {s₂ : σ₂}
 
 @[simp]
-theorem mapAccumr₂_mapAccumr_left (f₁ : γ → β → σ₁ → σ₁ × ζ) (f₂ : α → σ₂ → σ₂ × γ) :
+theorem mapAccumr₂_mapAccumr_left {γ ζ : Type} (f₁ : γ → β → σ₁ → σ₁ × ζ) (f₂ : α → σ₂ → σ₂ × γ) :
```
fpvandoorn, line 64: Is the missing `*` intentional (also below and above)?
Please make these variables.

grunweg: If you look very closely, you'll see that all lemmas in this section alternatingly have `Type` and `Type*`: the way they are currently written, this is the best one can do. I could extract the Type* case as variables (highlighting the non-polymorphic case); would you prefer this?

grunweg:
awaiting-review

Mathlib/Data/Vector/MapLemmas.lean
riccardobrasca: Does something really break with `Type*` everywhere?

grunweg: I've tried this in [this branch](https://github.com/leanprover-community/mathlib4/compare/MR-vector-typestart): six new errors about "unsolved goals" appear. Be my guest to PR and fix them - I'd like to declare this out of scope of this PR.

grunweg: I have pushed a commit, however, which makes Type* the common case - to emphasize the `Type` is unusual.

riccardobrasca: `Type*` means `Type u` for a new universe variable `u`, where `Type` means `Type 1`: in particular `Type*` is strictly more general than `Type`.

Mathlib/Data/Vector/Basic.lean
```diff
@@ -814,16 +813,18 @@ theorem get_map₂ (v₁ : Vector α n) (v₂ : Vector β n) (f : α → β →
     · simp only [get_zero, head_cons]
     · simp only [get_cons_succ, ih]
 
+set_option autoImplicit true in
```
riccardobrasca, line 816:
```lean
@[simp]
theorem mapAccumr_cons {α : Type*} (xs : Vector α n) :
    ∀ {α₁ : Type} {β : Type*} {f : α → α₁ → α₁ × β} {x : α} {s : α₁},
    mapAccumr f (x ::ᵥ xs) s
    = let r := mapAccumr f xs s
      let q := f x r.1
      (q.1, q.2 ::ᵥ r.2) :=
  rfl
```
this seems to work. I guess the next one is similar.

grunweg:
#16154 solves this part in a much nicer way: let me close this PR in favour of it.

PR #14886, opened 2024-07-18
[Merged by Bors] - chore(CategoryTheory/Monad): rename `η'` to `η`

yuma-mizuno:
There are no differences between `η'` and `η` (unit for the monad `T`) anymore as we have a syntactic tautology `T.toFunctor = T` in Lean 4.

---


joelriou:
Thanks!

bors merge

PR #14887, opened 2024-07-19
[Merged by Bors] - chore: remove `#align` and `#align_import` statements

kim-em:
Per discussion on the maintainers list / meetings, we are now ready to remove `#align`, and it's time to do it!

This PR is generated by running the following script in the root of the repository:

```
find . -name "*.lean" -exec sed -i '' '/^set_option linter.uppercaseLean3 false/d' {} +
find . -name "*.lean" -exec sed -i '' '/^set_option align.precheck false/d' {} +
find . -name "*.lean" -exec sed -i '' -e '/^$/N;/^\n#align/d' -e '/^#align/d' {} +
find . -name "*.lean" -exec sed -i '' -e '/^$/N;/^\n#noalign/d' -e '/^#noalign/d' {} +
rm Mathlib/Init/Align.lean
find . -name "*.lean" -exec sed -i '' '/^import Mathlib.Init.Align/d' {} +
find . -name "*.lean" -exec sed -i '' -e '1{/^$/d;}' {} +
```

---

This will rot instantly, but can of course be regenerated by running the script again. I suggest that we try to only include scriptable fixes to this PR, and then if there is a certain amount of "by hand" cleanup then we do that in a followup PR. There will be some stray comments in `Mathlib/Init/` that these scripts don't delete -- I propose that we just ignore these for now, as we are rapidly removing all of these files so any such comments will be dealt with soon.

I've also verified that we can empty `Mathlib/Mathport/Rename.lean` except for the `import Lean` (and of course later we can remove that too), but that is not part of this PR.

There is a CI failure from mergify (can we just turn this off??) because the PR touches more than 1000 files. We can just squash merge when this is ready.


adomani:
LGTM!

I checked the `git diff` against master and I agree that all the changes that I detected are
* the removals of
  * empty lines,
  * the relevant `set_option`s and the `#(no)align`s;
* the file `Mathlib/Init/Align.lean`.

GitHub also confirms that nothing was *added*!

adomani:
Btw, quite a few comments get erased with the lines containing `set_option`, `#align`,...
I looked at those comments not exhaustively, but they all looked like porting notes or explanations for why the "porting" commands were used or what they were doing.

It seems ok to remove them.

jcommelin:
Do you want to add
```
find . -name "*.lean" -exec sed -i '' '/^import Mathlib.Mathport.Rename/d' {} +
```
to the script? Or shall we clean those up later?

semorrison:
> Do you want to add
> 
> ```
> find . -name "*.lean" -exec sed -i '' '/^import Mathlib.Mathport.Rename/d' {} +
> ```
> 
> to the script? Or shall we clean those up later?

Let's do that separately. It's easy to do later, and we need to minimize complexity here.

mattrobball:
bors merge

PR #14890, opened 2024-07-19
[Merged by Bors] - feat(CategoryTheory): pushout/pullback squares are stable by isomorphisms

joelriou:
The properties `IsPushout` and `IsPullback` are stable by isomorphisms.

---


erdOne:
Would it be useful as an iff?

joelriou:
> Would it be useful as an iff?
If we had a category of commutative squares, an `iff` lemma would be convenient, but in the unbundled setting, I do not think it would be useful.

erdOne:
Okay sure.
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14895, opened 2024-07-19
[Merged by Bors] - feat(RingTheory/Powerseries): generalize `coeff_one_pow`

rwst:
Removes the condition of `coeff_one_pow` that the constant coefficient must be 1.

---


scripts/noshake.json
```diff
@@ -302,12 +305,11 @@
   "Mathlib.Lean.Expr.ExtraRecognizers": ["Mathlib.Data.Set.Defs"],
   "Mathlib.Lean.Expr.Basic": ["Batteries.Logic"],
   "Mathlib.Init.Logic": ["Batteries.Logic"],
-  "Mathlib.Logic.Function.Defs": ["Mathlib.Tactic.AdaptationNote"],
```
riccardobrasca, line 305: Are these changes wanted?

rwst: These were made by `lake exe shake --fix`.

riccardobrasca: Meaning that `shake` complains without them?

riccardobrasca:
Can you add back the previous theorem, as a special case of the new result?

Mathlib/RingTheory/PowerSeries/Basic.lean
```diff
@@ -690,18 +692,35 @@ lemma coeff_one_mul (φ ψ : R⟦X⟧) : coeff R 1 (φ * ψ) =
     mul_comm, add_comm]
   norm_num
 
-/-- First coefficient of the `n`-th power of a power series with constant coefficient 1. -/
-lemma coeff_one_pow (n : ℕ) (φ : R⟦X⟧) (hC : constantCoeff R φ = 1) :
-    coeff R 1 (φ ^ n) = n * coeff R 1 φ := by
+/-- First coefficient of the `n`-th power of a power series. -/
+lemma coeff_one_pow (n : ℕ) (hn : 1 ≤ n) (φ : R⟦X⟧) :
```
AntoineChambert-Loir, line 696: It seems that theorem holds if n = 0 (for ridiculous reasons) : both sides are zero.

riccardobrasca: Even better then!

AntoineChambert-Loir:
Is it really needed, or cannot one just apply this result and go one with rewriting the constantCoeff and using `one_pow`?

riccardobrasca:
> Is it really needed, or cannot one just apply this result and go one with rewriting the constantCoeff and using `one_pow`?

Hmm, you're right, it's not that natural and we can live without it.

rwst:
Noted and closed, thanks.

riccardobrasca:
I think Antoine was answering my comment, saying there is no need to add the current result back. Your generalization looks totally fine.

scripts/noshake.json
rwst: OK, it builds without that line.

Mathlib/RingTheory/PowerSeries/Basic.lean
```diff
@@ -581,18 +583,35 @@ lemma coeff_one_mul (φ ψ : R⟦X⟧) : coeff R 1 (φ * ψ) =
...
+    coeff R 1 (φ ^ n) = n * coeff R 1 φ * (constantCoeff R φ) ^ (n - 1) := by
```
riccardobrasca, line 588:
```suggestion
/-- First coefficient of the `n`-th power of a power series. -/
lemma coeff_one_pow (n : ℕ) (φ : R⟦X⟧) :
    coeff R 1 (φ ^ n) = n * coeff R 1 φ * (constantCoeff R φ) ^ (n - 1) := by
  rcases Nat.eq_zero_or_pos n with (rfl | hn)
  · simp
```

riccardobrasca:
You also have a conflict... otherwise LGTM, thanks!

bors d+

rwst:
That conflict is not really there. I merged master without change, and CI builds Mathlib without problems.

riccardobrasca:
> That conflict is not really there. I merged master without change, and CI builds Mathlib without problems.

Did you pull before merging? It says it is in ` Mathlib/RingTheory/PowerSeries/Basic.lean `.

rwst:
> > That conflict is not really there. I merged master without change, and CI builds Mathlib without problems.
> 
> Did you pull before merging? It says it is in `Mathlib/RingTheory/PowerSeries/Basic.lean`.

I have edited it in browser.

PR #14896, opened 2024-07-19
[Merged by Bors] - chore: remove more easy autoImplicit

grunweg:
---

### DIFF
--- a/Mathlib/Data/List/EditDistance/Estimator.lean
+++ b/Mathlib/Data/List/EditDistance/Estimator.lean
@@ -23,10 +23,8 @@ This is then used in the implementation of `rewrite_search`
 to avoid needing the entire edit distance calculation in unlikely search paths.
 -/
 
-set_option autoImplicit true
-
 variable {α β δ : Type} [CanonicallyLinearOrderedAddCommMonoid δ]
-    (C : Levenshtein.Cost α β δ) (xs : List α) (ys : List β)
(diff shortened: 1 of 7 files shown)


Mathlib/Data/List/EditDistance/Estimator.lean
```diff
@@ -23,10 +23,8 @@ This is then used in the implementation of `rewrite_search`
 to avoid needing the entire edit distance calculation in unlikely search paths.
 -/
 
-set_option autoImplicit true
-
 variable {α β δ : Type} [CanonicallyLinearOrderedAddCommMonoid δ]
```
fpvandoorn, line 26: I know you didn't edit this, but this missing `*` is intentional?

fpvandoorn (commented):
These changes make it clear to me how "evil" `autoImplicit` is. with it, it's hard to even see that some lemmas are only for `Type`, not `Type*`

fpvandoorn:
bors d+

Mathlib/Data/List/EditDistance/Estimator.lean
grunweg: Good question! I checked; \alpha can be generalised, but making \beta or \delta take Type* errors all over the place. I didn't think much if that can be fixed, but will leave that to a follow-up.

grunweg:
Thanks for the quick review!
bors r+

PR #14901, opened 2024-07-19
[Merged by Bors] - chore (test/Lint): remove `#align` test

mattrobball:
This breaks testing for #14887 but can't be automated with `sed` so we remove it manually.

---


riccardobrasca:
bors d+

PR #14908, opened 2024-07-19
[Merged by Bors] - feat: the Galois connection induced by an arbitrary relation

AnthonyBordg:
Define a Galois connection from an arbitrary relation and prove this induces inverse bijections between two sets of fixed points.    

---


Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Rel
+
+/-! ### Pairs of adjoint maps defined by relations -/
+
+open OrderDual
+
+/-- `l` maps any set `J` of elements of type `α` to the set `{b : β | ∀ a ∈ J, R a b}` of
+elements `b` of type `β` such that `R a b` for every element `a` of `J`. -/
+def l (J : Set α) : (Set β)ᵒᵈ := toDual {b : β | ∀ a ∈ J, R a b}
```
adamtopaz, line 48: Could you please give this (and `u` below) a more descriptive name?

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/-- `u` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ ofDual I, R a b}`
+of elements `a` of type `α` such that `R a b` for every element `b` of `I`. -/
+def u (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ b ∈ ofDual I, R a b}
+
+/-- The pair of functions `l` and `u` forms a Galois connection. -/
+theorem to_galoisConnection : GaloisConnection (l R) (u R) := by
```
adamtopaz, line 55: `to_galoisConnection` doesn't follow our naming conventions for mathlib.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/-- `u` maps every element `I` to `lFixedPoints`. -/
+@[simp]
+theorem is_lFixedPoint (I : (Set β)ᵒᵈ) : u R I ∈ lFixedPoints R := by
+    unfold lFixedPoints; simp; apply le_antisymm
+    · apply (to_galoisConnection R).monotone_u; exact (to_galoisConnection R).l_u_le I
+    · exact (to_galoisConnection R).le_u_l (u R I)
+
+/-- The maps `l` and `u` induce inverse bijections between the sets of fixed points. -/
+def instEquivFixedPoints : lFixedPoints R ≃ uFixedPoints R :=
```
adamtopaz, line 89: I'm not sure why you called this `instEquiv...` as this is not an instance.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/-- `uFixedPoints` is the set of elements `I : (Set β)ᵒᵈ` satisfying `l (u I) = I`. -/
+def uFixedPoints := {I : (Set β)ᵒᵈ | l R (u R I) = I}
+
+open GaloisConnection
+
+/-- `l` maps every element `J` to `uFixedPoints`. -/
+@[simp]
+theorem is_uFixedPoint (J : Set α) : l R J ∈ uFixedPoints R := by
+    unfold uFixedPoints; simp; apply le_antisymm
```
adamtopaz, line 77: We avoid nonterminal `simp`s in tactic blocks.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,107 @@
...
+/-- `lFixedPoints` is the set of elements `J : Set α` satisfying `u (l J) = J`. -/
+def lFixedPoints := {J : Set α | u R (l R J) = J}
+
+/-- `uFixedPoints` is the set of elements `I : (Set β)ᵒᵈ` satisfying `l (u I) = I`. -/
+def uFixedPoints := {I : (Set β)ᵒᵈ | l R (u R I) = I}
+
+open GaloisConnection
+
+/-- `l` maps every element `J` to `uFixedPoints`. -/
+@[simp]
```
adamtopaz, line 75: It's not clear to me why you would want to tag this with `simp`. Similarly for the lemma below.

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: @adamtopaz, I first used `lAdjMap` (resp.`rAdjMap`) for the left (resp. right) map of the adjunction, then realised it would be more coherent wrt to the file `GaloisConnection.lean` to use `l` and `u`.

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Relic of a former error. Would `equivFixedPoints` be suitable?

adamtopaz: Yes, I think that's a reasonable name. Thanks!

Mathlib/Order/ReltoGaloisConnection.lean
adamtopaz: I think `l` should be named something like `rightDual` and `u` something like `leftDual`. Also, I think it's better to introduce these definitions without the order duals, develop an API for the definitions directly, then package that API in the Galois connection at the very end. So, for example, I envision that `rightDual` would be defined as 
```lean
def rightDual (J : Set α) : Set β := {b : β | ∀ a ∈ J, R a b}
```
and analogously for `leftDual`. 

Concerning the names, @YaelDillies would probably have a better sense of how to name these.

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: `OrderIso.to_galoisConnection` is actually misnamed as well. It should be `OrderIso.galoisConnection` or `OrderIso.gc`

AnthonyBordg: Ok, so it should simply be `Rel.galoisConnection` or without the prefix `galoisConnection`? Just for my personal enlightenment, which naming convention(s) `to_galoisConnection` doesn't follow?

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: > I think `l` should be named something like `rightDual` and `u` something like `leftDual`. 

I assume you mean the other way around, i.e. `l` should be named `leftDual`.

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: `to` is only used for forgetful inheritance of structures

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Not sure I understand what API you envision, but I renamed `l` (resp. `u`) into `leftDual` (resp. `rightDual`).

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+namespace Rel
+
+/-! ### Pairs of adjoint maps defined by relations -/
+
+open OrderDual
+
+/-- `leftDual` maps any set `J` of elements of type `α` to the set `{b : β | ∀ a ∈ J, R a b}` of
+elements `b` of type `β` such that `R a b` for every element `a` of `J`. -/
+def leftDual (J : Set α) : (Set β)ᵒᵈ := toDual {b : β | ∀ a ∈ J, R a b}
```
YaelDillies, line 49:
```suggestion
def leftDual (J : Set α) : (Set β)ᵒᵈ := toDual {b : β | ∀ ⦃a⦄, a ∈ J → R a b}
```
so that you don't need to provide both `a` and `ha : a ∈ J` when trying to deduce `R a b` from `b ∈ leftDual J`.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+/-- `rightDual` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ ofDual I, R a b}`
+of elements `a` of type `α` such that `R a b` for every element `b` of `I`. -/
+def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ b ∈ ofDual I, R a b}
```
YaelDillies, line 53: Same here

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+/-- The pair of functions `leftDual` and `rightDual` forms a Galois connection. -/
+theorem galoisConnection : GaloisConnection (leftDual R) (rightDual R) := by
+    intros J I; apply Iff.trans (b := ∀ b ∈ ofDual I, ∀ a ∈ J, R a b)
+    · constructor
+      · intro h; apply h
+      · intro h; apply h
```
YaelDillies, line 60:
```suggestion
    · constructor <;> intro h <;> apply h
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+    · constructor
+      · intro h; unfold rightDual; intro a ha b hb; apply h; simpa; exact ha
+      · intro h; intro b hb a ha; apply h; simpa; exact hb
```
YaelDillies, line 63: This is not in line with the style guide. Please put at most one tactic per line unless one of the following applies:
* It's a one-line proof
* The tactics are chained with `<;>`

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+def lFixedPoints := {J : Set α | rightDual R (leftDual R J) = J}
+
+/-- `rFixedPoints` is the set of elements `I : (Set β)ᵒᵈ` satisfying `leftDual (rightDual I) = I`.
+-/
+def rFixedPoints := {I : (Set β)ᵒᵈ | leftDual R (rightDual R I) = I}
+
+open GaloisConnection
+
+/-- `leftDual` maps every element `J` to `rFixedPoints`. -/
+theorem is_rFixedPoint (J : Set α) : leftDual R J ∈ rFixedPoints R := by
```
YaelDillies, line 77:
```suggestion
theorem leftDual_mem_rFixedPoint (J : Set α) : leftDual R J ∈ rFixedPoints R := by
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+    unfold rFixedPoints; apply le_antisymm
+    · exact (galoisConnection R).l_u_le (leftDual R J)
+    · apply (galoisConnection R).monotone_l; exact (galoisConnection R).le_u_l J
+
+/-- `rightDual` maps every element `I` to `lFixedPoints`. -/
+theorem is_lFixedPoint (I : (Set β)ᵒᵈ) : rightDual R I ∈ lFixedPoints R := by
```
YaelDillies, line 83:
```suggestion
theorem rightDual_mem_lFixedPoint (I : (Set β)ᵒᵈ) : rightDual R I ∈ lFixedPoints R := by
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+/-- The maps `leftDual` and `rightDual` induce inverse bijections between the sets of fixed points.
+-/
+def equivFixedPoints : lFixedPoints R ≃ rFixedPoints R :=
+    { toFun := fun ⟨J, _⟩ => ⟨leftDual R J, is_rFixedPoint R J⟩
+      invFun := fun ⟨I, _⟩ => ⟨rightDual R I, is_lFixedPoint R I⟩
+      left_inv := by
+          intro J; cases' J with J hJ; simp; rw[hJ]
+      right_inv := by
+          intro I; cases' I with I hI; simp; rw[hI] }
```
YaelDillies, line 96:
```suggestion
def equivFixedPoints : lFixedPoints R ≃ rFixedPoints R where
  toFun := fun ⟨J, _⟩ => ⟨leftDual R J, is_rFixedPoint R J⟩
  invFun := fun ⟨I, _⟩ => ⟨rightDual R I, is_lFixedPoint R I⟩
  left_inv J := by cases' J with J hJ; simp; rw [hJ]
  right_inv I := by cases' I with I hI; simp; rw [hI]
```
Always a space between `rw` and `[`

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+theorem le_imp_u_l_le {J J' : Set α} (h : J' ∈ lFixedPoints R) :
+    J ≤ J' → rightDual R (leftDual R J) ≤ J' := by
+    intro h₁; rw[← h]; apply (galoisConnection R).monotone_u;
+    apply (galoisConnection R).monotone_l; exact h₁
```
YaelDillies, line 101:
```suggestion
theorem le_imp_u_l_le {J J' : Set α} (h : J' ∈ lFixedPoints R) (h₁ : J ≤ J') :
    rightDual R (leftDual R J) ≤ J' := by
  rw [← h]
  apply (galoisConnection R).monotone_u
  apply (galoisConnection R).monotone_l
  exact h₁
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,110 @@
...
+
+theorem ge_imp_ge_l_u {I I' : (Set β)ᵒᵈ} (h : I' ∈ rFixedPoints R) :
+    I' ≥ I → I' ≥ leftDual R (rightDual R I) := by
+    intro h₁; rw [← h]; apply (galoisConnection R).monotone_l;
+    apply (galoisConnection R).monotone_u; exact h₁
```
YaelDillies, line 106: Same here

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Line 62 and line 63 shouldn't be considered one-line proofs, although they close a goal?

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: > I think it's better to introduce these definitions without the order duals, develop an API for the definitions directly, then package that API in the Galois connection at the very end.

I second this.

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: No, they are part of a proof that's longer than one line. They are one-line *sub*proofs.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -55,12 +55,19 @@ def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ ⦃b⦄, b ∈ ofDu
...
+    · constructor <;> intro h <;> apply h
     · constructor
-      · intro h; apply h
-      · intro h; apply h
-    · constructor
-      · intro h; unfold rightDual; intro a ha b hb; apply h; simpa; exact ha
-      · intro h; intro b hb a ha; apply h; simpa; exact hb
+      · intro h
+        unfold rightDual
+        intro a ha b hb
```
YaelDillies, line 62:
```suggestion
      · intro h a ha b hb
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -55,12 +55,19 @@ def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ ⦃b⦄, b ∈ ofDu
...
+        apply h
+        simpa
+        exact ha
+      · intro h
+        intro b hb a ha
```
YaelDillies, line 67:
```suggestion
      · intro h b hb a ha
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -55,12 +55,19 @@ def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ ⦃b⦄, b ∈ ofDu
...
+        apply h
+        simpa
+        exact hb
```
YaelDillies, line 70:
```suggestion
        exact h (by simpa) hb
```
or something like that?

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -55,12 +55,19 @@ def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ ⦃b⦄, b ∈ ofDu
...
-      · intro h; apply h
-    · constructor
-      · intro h; unfold rightDual; intro a ha b hb; apply h; simpa; exact ha
-      · intro h; intro b hb a ha; apply h; simpa; exact hb
+      · intro h
+        unfold rightDual
+        intro a ha b hb
+        apply h
+        simpa
+        exact ha
```
YaelDillies, line 65: Same here

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -95,26 +95,26 @@ theorem rightDual_mem_lFixedPoint (I : (Set β)ᵒᵈ) : rightDual R I ∈ lFixe
...
 -/
 def equivFixedPoints : lFixedPoints R ≃ rFixedPoints R where
-    toFun := fun ⟨J, _⟩ => ⟨leftDual R J, leftDual_mem_rFixedPoint R J⟩
-    invFun := fun ⟨I, _⟩ => ⟨rightDual R I, rightDual_mem_lFixedPoint R I⟩
-    left_inv := by intro J; cases' J with J hJ; simp; rw [hJ]
-    right_inv := by intro I; cases' I with I hI; simp; rw [hI]
+  toFun := fun ⟨J, _⟩ => ⟨leftDual R J, leftDual_mem_rFixedPoint R J⟩
+  invFun := fun ⟨I, _⟩ => ⟨rightDual R I, rightDual_mem_lFixedPoint R I⟩
+  left_inv := by intro J; cases' J with J hJ; simp; rw [hJ]
+  right_inv := by intro I; cases' I with I hI; simp; rw [hI]
```
YaelDillies, line 101: `where` notation lets introduce variables left to the `:=`
```suggestion
  left_inv J := by cases' J with J hJ; simp; rw [hJ]
  right_inv I := by cases' I with I hI; simp; rw [hI]
```

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: @adamtopaz, Would be happy to implement this, but I would need a precise specification of what you want here.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,114 @@
...
+
+open OrderDual
+
+/-- `leftDual` maps any set `J` of elements of type `α` to the set `{b : β | ∀ a ∈ J, R a b}` of
+elements `b` of type `β` such that `R a b` for every element `a` of `J`. -/
+def leftDual (J : Set α) : (Set β)ᵒᵈ := toDual {b : β | ∀ ⦃a⦄, a ∈ J → R a b}
+
+/-- `rightDual` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ ofDual I, R a b}`
+of elements `a` of type `α` such that `R a b` for every element `b` of `I`. -/
+def rightDual (I : (Set β)ᵒᵈ) : Set α := {a : α | ∀ ⦃b⦄, b ∈ ofDual I → R a b}
```
YaelDillies, line 53: @AnthonyBordg, here's what Adam is suggesting, concretely
```suggestion
def leftDual (J : Set α) : Set β := {b : β | ∀ ⦃a⦄, a ∈ J → R a b}

/-- `rightDual` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ ofDual I, R a b}`
of elements `a` of type `α` such that `R a b` for every element `b` of `I`. -/
def rightDual (I : Set β) : Set α := {a : α | ∀ ⦃b⦄, b ∈ I → R a b}
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,114 @@
...
+
+/-- The pair of functions `leftDual` and `rightDual` forms a Galois connection. -/
+theorem galoisConnection : GaloisConnection (leftDual R) (rightDual R) := by
```
YaelDillies, line 56: ... and
```suggestion
theorem galoisConnection : GaloisConnection (toDual ∘ leftDual R) (rightDual R ∘ ofDual) := by
```

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: See https://github.com/leanprover-community/mathlib4/pull/14908#discussion_r1685417366

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: @YaelDillies @adamtopaz 
I don't see what you really gain with these changes, however please note that the definition of `rFixedPoints` will be far more cumbersome, in particular you will break the symmetry in the expressions of `lFixedPoints` and `rFixedPoints`. Moreover, you won't be able to remove the mention of the dual order `(Set β)ᵒᵈ` in `rFixedPoints`. 
Is it really what you want?

YaelDillies: Why can't it simply be
```
def rFixedPoints := {I : Set β | R.leftDual (R.rightDual I) = I}
```
?

AnthonyBordg: In this case, how will you prove `leftDual_mem_rFixedPoint`, i.e. that an image is a fixed-point?

YaelDillies: Same proof. Why would it change? You're just adding a bunch of "identity" isomorphisms here and there, and they cancel out

AnthonyBordg: The adjoint maps now are (`toDual \circ leftDual`, `rightDual \circ ofDual`), not (`leftDual`, `rightDual`) anymore. You can still express the property of `J` being a left fixed point with `R.rightDual (R.leftDual J) = J`, since `(rightDual \circ ofDual) \circ (toDual \circ leftDual) = rightDual \circ leftDual`, i.e. the additional isos cancel out, but for a right fixed point they don't cancel out, the property of being a right fixed point have to use `(toDual \circ leftDual) \circ (rightDual \circ ofDual)` \neq `\leftDual \circ rightDual` (these last two maps are not even of the same type, i.e. automorphism of `(Set β)ᵒᵈ` vs of `Set β`).
Am I missing something?

YaelDillies: Sorry, not at a computer right now but morally if your statements are not about orders, then they shouldn't need anything related to `OrderDual`. `rFixedPoints_mem_rDual` isn't about order so you shouldn't need `OrderDual`, even though the intermediate lemma `galoisConnection` needs it.

AnthonyBordg: Thank you for your reply and sorry to bother on a Sunday, Yael :)
>if your statements are not about orders

I believe there are, no? $I$ being a fixed point under $F$ means $I \leq F (I)$ and $F(I)\leq I$.
Moreover, if these statements weren't about orders, you would prevent yourself from using the API for Galois connections ... so, there would be no point in proving that the induced bijections between fixed points come from a Galois connection. 
Since the change was suggested by Adam, maybe he can help?

YaelDillies: I'm not saying it *can't* be interpreted as an order theoretic statement, but rather that it *can* be interpreted without order theory. If all else fails, I'll be back at a computer first thing tomorrow morning ;)

YaelDillies: Done. Does our suggestion make more sense now?

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -67,25 +67,24 @@ theorem galoisConnection : GaloisConnection R.leftDual R.rightDual := by
...
 theorem leftDual_mem_rFixedPoint (J : Set α) : R.leftDual J ∈ R.rFixedPoints := by
   unfold rFixedPoints
   apply le_antisymm
-  · exact R.galoisConnection.l_u_le (R.leftDual J)
-  · apply R.galoisConnection.monotone_l; exact R.galoisConnection.le_u_l J
+  · apply R.gc_leftDual_rightDual.monotone_l; exact R.gc_leftDual_rightDual.le_u_l J
+  · exact R.gc_leftDual_rightDual.l_u_le (R.leftDual J)
 
 /-- `rightDual` maps every element `I` to `lFixedPoints`. -/
 theorem rightDual_mem_lFixedPoint (I : (Set β)ᵒᵈ) : R.rightDual I ∈ R.lFixedPoints := by
```
AnthonyBordg, line 83: Should this mention of the dual order `(Set β)ᵒᵈ` also been removed?

YaelDillies: Yep, well spotted!

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Thank you, Yael. I left a comment below https://github.com/leanprover-community/mathlib4/pull/14908#discussion_r1686241361

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Deleted. I'm happy with the changes. Thanks a lot Yael for your time and effort!

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,115 @@
...
+
+/-- The pair of functions `leftDual` and `rightDual` forms a Galois connection. -/
+theorem gc_leftDual_rightDual : GaloisConnection (toDual ∘ R.leftDual) (R.rightDual ∘ ofDual) := by
+  intros J I; apply Iff.trans (b := ∀ b ∈ ofDual I, ∀ a ∈ J, R a b)
+  · constructor <;> intro h <;> apply h
+  · constructor
+    · intro h a ha b hb
+      exact h (by simpa) hb a ha
+    · intro h b hb a ha
+      exact h (by simpa) hb
```
YaelDillies, line 63:
```suggestion
  intros J I
  change ∀ b ∈ ofDual I, ∀ a ∈ J, R a b)
  constructor
  · intro h a ha b hb
    exact h (by simpa) hb a ha
  · intro h b hb a ha
    exact h (by simpa) hb
```
or even just
```suggestion
  fun J I ↦ ⟨fun h a ha b hb ↦ h (by simpa) hb a ha, fun h b hb a hah b hb a ha ↦ h (by simpa) hb⟩
```

Mathlib/Order/ReltoGaloisConnection.lean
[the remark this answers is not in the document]
AnthonyBordg: Sure. Should it be `ReltogaloisConnection`? Are the conventions for file names the same than the general [naming conventions](https://leanprover-community.github.io/contribute/naming.html)?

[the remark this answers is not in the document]
AnthonyBordg: To rename the file not only locally on my computer, but also in my GitHub branch, what should I do? Run `lake exe mk_all`?

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,115 @@
...
+
+/-- The maps `leftDual` and `rightDual` induce inverse bijections between the sets of fixed points.
+-/
+def equivFixedPoints : R.lFixedPoints ≃ R.rFixedPoints where
+  toFun := fun ⟨J, _⟩ => ⟨R.leftDual J, R.leftDual_mem_rFixedPoint J⟩
+  invFun := fun ⟨I, _⟩ => ⟨R.rightDual I, R.rightDual_mem_lFixedPoint I⟩
+  left_inv J := by cases' J with J hJ; simp; rw [hJ]
+  right_inv I := by cases' I with I hI; simp; rw [hI]
+
+theorem le_imp_u_l_le {J J' : Set α} (h : J' ∈ R.lFixedPoints) :
```
YaelDillies, line 91: The naming convention is `conclusion_of_premise`, not `premise_imp_conclusion`

Mathlib/Order/ReltoGaloisConnection.lean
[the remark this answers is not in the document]
YaelDillies: This has nothing to do with `lake exe mk_all` (although you should also run that). To git, renaming a file is the same as removing all the content of a file and adding a bunch of content to another file. These are things that git tracks, so you should create a commit as usual.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-! ### Pairs of adjoint maps defined by relations -/
+
+open OrderDual
+
+/-- `leftDual` maps any set `J` of elements of type `α` to the set `{b : β | ∀ a ∈ J, R a b}` of
+elements `b` of type `β` such that `R a b` for every element `a` of `J`. -/
+def leftDual (J : Set α) : Set β := {b : β | ∀ ⦃a⦄, a ∈ J → R a b}
+
+/-- `rightDual` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ ofDual I, R a b}`
```
adamtopaz, line 51:
```suggestion
/-- `rightDual` maps any set `I` of elements of type `β` to the set `{a : α | ∀ b ∈ I, R a b}`
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+of elements `a` of type `α` such that `R a b` for every element `b` of `I`. -/
+def rightDual (I : Set β) : Set α := {a : α | ∀ ⦃b⦄, b ∈ I → R a b}
+
+/-- The pair of functions `leftDual` and `rightDual` forms a Galois connection. -/
+theorem gc_leftDual_rightDual : GaloisConnection (toDual ∘ R.leftDual) (R.rightDual ∘ ofDual) :=
+  fun J I ↦ ⟨fun h a ha b hb ↦ h (by simpa) ha, fun h b hb a ha ↦ h (by simpa) hb⟩
+
+/-! ### Induced equivalences and generation processes -/
```
adamtopaz, line 59: It's not clear to me what you mean by "generation process". I think a more descriptive title would be something like the following:
```suggestion
/-! ### Induced equivalences between fixed points. -/
```

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- `lFixedPoints` is the set of elements `J : Set α` satisfying `rightDual (leftDual J) = J`. -/
+def lFixedPoints := {J : Set α | R.rightDual (R.leftDual J) = J}
```
adamtopaz, line 62: Can you please rename this `leftFixedPoints` (and similarly for the right fixed points)?

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+def equivFixedPoints : R.lFixedPoints ≃ R.rFixedPoints where
+  toFun := fun ⟨J, _⟩ => ⟨R.leftDual J, R.leftDual_mem_rFixedPoint J⟩
+  invFun := fun ⟨I, _⟩ => ⟨R.rightDual I, R.rightDual_mem_lFixedPoint I⟩
+  left_inv J := by cases' J with J hJ; simp; rw [hJ]
+  right_inv I := by cases' I with I hI; simp; rw [hI]
+
+theorem u_l_le_of_le {J J' : Set α} (h : J' ∈ R.lFixedPoints) :
+    J ≤ J' → R.rightDual (R.leftDual J) ≤ J' := by
+  intro h₁
+  rw[← h]
```
adamtopaz, line 94:
```suggestion
  rw [← h]
```

adamtopaz (commented):
Thanks for making those changes! This is getting quite close to being ready to merge!

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- The maps `leftDual` and `rightDual` induce inverse bijections between the sets of fixed points.
+-/
+def equivFixedPoints : R.lFixedPoints ≃ R.rFixedPoints where
+  toFun := fun ⟨J, _⟩ => ⟨R.leftDual J, R.leftDual_mem_rFixedPoint J⟩
+  invFun := fun ⟨I, _⟩ => ⟨R.rightDual I, R.rightDual_mem_lFixedPoint I⟩
+  left_inv J := by cases' J with J hJ; simp; rw [hJ]
+  right_inv I := by cases' I with I hI; simp; rw [hI]
+
+theorem u_l_le_of_le {J J' : Set α} (h : J' ∈ R.lFixedPoints) :
```
adamtopaz, line 91: Since the names of `u` and `l` have been changed, the names of this and the following lemma should be changed accordingly as well.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+def rFixedPoints := {I : Set β | R.leftDual (R.rightDual I) = I}
+
+open GaloisConnection
+
+/-- `leftDual` maps every element `J` to `rFixedPoints`. -/
+theorem leftDual_mem_rFixedPoint (J : Set α) : R.leftDual J ∈ R.rFixedPoints := by
+  unfold rFixedPoints
+  apply le_antisymm
+  · apply R.gc_leftDual_rightDual.monotone_l; exact R.gc_leftDual_rightDual.le_u_l J
+  · exact R.gc_leftDual_rightDual.l_u_le (R.leftDual J)
```
pitmonticone, line 74: This can be turned into a proof term as follows:

```suggestion
theorem leftDual_mem_rFixedPoint (J : Set α) : R.leftDual J ∈ R.rFixedPoints :=
  le_antisymm (R.gc_leftDual_rightDual.monotone_l (R.gc_leftDual_rightDual.le_u_l J))
    (R.gc_leftDual_rightDual.l_u_le (R.leftDual J))
```

I have tested the code so you can safely commit this suggestion if you like.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- `rightDual` maps every element `I` to `lFixedPoints`. -/
+theorem rightDual_mem_lFixedPoint (I : Set β) : R.rightDual I ∈ R.lFixedPoints := by
+  unfold lFixedPoints
+  apply le_antisymm
+  · apply R.gc_leftDual_rightDual.monotone_u; exact R.gc_leftDual_rightDual.l_u_le I
+  · exact R.gc_leftDual_rightDual.le_u_l (R.rightDual I)
```
pitmonticone, line 81: This can be turned into a proof term as follows:

```suggestion
theorem rightDual_mem_lFixedPoint (I : Set β) : R.rightDual I ∈ R.lFixedPoints :=
  le_antisymm (R.gc_leftDual_rightDual.monotone_u (R.gc_leftDual_rightDual.l_u_le I)) (R.gc_leftDual_rightDual.le_u_l (R.rightDual I))
```

I have tested the code so you can safely commit this suggestion if you like.

Mathlib/Order/ReltoGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+  apply R.gc_leftDual_rightDual.monotone_l
+  exact h₁
+
+theorem ge_l_u_of_ge {I I' : Set β} (h : I' ∈ R.rFixedPoints) :
+    I' ≥ I → I' ≥ R.leftDual (R.rightDual I) := by
+  intro h₁
+  rw [← h]
+  apply R.gc_leftDual_rightDual.monotone_l
+  apply R.gc_leftDual_rightDual.monotone_u
+  exact h₁
```
pitmonticone, line 105: This can be turned into a proof term as follows:

```suggestion
theorem ge_l_u_of_ge {I I' : Set β} (h : I' ∈ R.rFixedPoints) :
    I' ≥ I → I' ≥ R.leftDual (R.rightDual I) :=
  fun h₁ ↦ h ▸ R.gc_leftDual_rightDual.monotone_l (R.gc_leftDual_rightDual.monotone_u h₁)
```

I have tested the code so you can safely commit this suggestion if you like.

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Are proof terms or tactics preferred in mathlib4, wrt legibility, maintenance etc?

Mathlib/Order/ReltoGaloisConnection.lean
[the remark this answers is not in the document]
AnthonyBordg: On training abroad the whole week, will try to find time this evening.

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: Really depends on the length. Here I think a tactic proof is justified

pitmonticone: As far as I know, there is no general convention in mathlib4 regarding the preference between proof terms and tactics. 

In my personal experience as an author (~ 7 months) and as a reviewer (~ 1 month), for relatively simple and short proofs such as these, fewer lines of code (and possibly faster execution) are typically considered better. 

Given a 5-10 lines simple proof (e.g. an intro and a sequence of apply), I think we can fairly say that a 1-2 lines proof term is better.

I don’t believe there is any significant loss in readability and maintainability in this case. 

Please @YaelDillies, correct me if I'm wrong.

pitmonticone: Ops, sorry @YaelDillies. For some reason your comment didn't show up on the mobile app...

YaelDillies: I think it's just not a big deal either way

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+def rightFixedPoints := {I : Set β | R.leftDual (R.rightDual I) = I}
+
+open GaloisConnection
+
+/-- `leftDual` maps every element `J` to `rightFixedPoints`. -/
+theorem leftDual_mem_rightFixedPoint (J : Set α) : R.leftDual J ∈ R.rightFixedPoints := by
+  unfold rightFixedPoints
+  apply le_antisymm
+  · apply R.gc_leftDual_rightDual.monotone_l; exact R.gc_leftDual_rightDual.le_u_l J
+  · exact R.gc_leftDual_rightDual.l_u_le (R.leftDual J)
```
pitmonticone, line 74: Updated:

```suggestion
theorem leftDual_mem_rightFixedPoint (J : Set α) : R.leftDual J ∈ R.rightFixedPoints :=
  le_antisymm (R.gc_leftDual_rightDual.monotone_l (R.gc_leftDual_rightDual.le_u_l J))
    (R.gc_leftDual_rightDual.l_u_le (R.leftDual J))
```

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- `rightDual` maps every element `I` to `leftFixedPoints`. -/
+theorem rightDual_mem_leftFixedPoint (I : Set β) : R.rightDual I ∈ R.leftFixedPoints := by
+  unfold leftFixedPoints
+  apply le_antisymm
+  · apply R.gc_leftDual_rightDual.monotone_u; exact R.gc_leftDual_rightDual.l_u_le I
+  · exact R.gc_leftDual_rightDual.le_u_l (R.rightDual I)
```
pitmonticone, line 81: Updated: 

```suggestion
theorem rightDual_mem_leftFixedPoint (I : Set β) : R.rightDual I ∈ R.leftFixedPoints :=
  le_antisymm (R.gc_leftDual_rightDual.monotone_u (R.gc_leftDual_rightDual.l_u_le I))
    (R.gc_leftDual_rightDual.le_u_l (R.rightDual I))
```

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+  left_inv J := by cases' J with J hJ; simp; rw [hJ]
+  right_inv I := by cases' I with I hI; simp; rw [hI]
+
+theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) :
+    J ≤ J' → R.rightDual (R.leftDual J) ≤ J' := by
+  intro h₁
+  rw [← h]
+  apply R.gc_leftDual_rightDual.monotone_u
+  apply R.gc_leftDual_rightDual.monotone_l
+  exact h₁
```
pitmonticone, line 97: Updated: 

```suggestion
theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) :
    J ≤ J' → R.rightDual (R.leftDual J) ≤ J' :=
  fun h₁ ↦ h ▸ R.gc_leftDual_rightDual.monotone_u (R.gc_leftDual_rightDual.monotone_l h₁)
```

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+theorem ge_leftDual_rightDual_of_ge {I I' : Set β} (h : I' ∈ R.rightFixedPoints) :
+    I' ≥ I → I' ≥ R.leftDual (R.rightDual I) := by
+  intro h₁
+  rw [← h]
+  apply R.gc_leftDual_rightDual.monotone_l
+  apply R.gc_leftDual_rightDual.monotone_u
+  exact h₁
```
pitmonticone, line 105: Updated:

```suggestion
theorem ge_leftDual_rightDual_of_ge {I I' : Set β} (h : I' ∈ R.rightFixedPoints) :
    I' ≥ I → I' ≥ R.leftDual (R.rightDual I) :=
  fun h₁ ↦ h ▸ R.gc_leftDual_rightDual.monotone_l (R.gc_leftDual_rightDual.monotone_u h₁)
```

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+a pair `leftDual` and `rightDual` of adjoint order-preserving maps between the corresponding posets
+`Set α` and `(Set β)ᵒᵈ`.
+We define `lFixedPoints` (resp. `rFixedPoints`) as the set of fixed points `J` (resp. `I`) of
+`Set α` (resp. `(Set β)ᵒᵈ`) such that `rightDual (leftDual J) = J` (resp.
+`leftDual (rightDual I) = I`).
+
+## Main Results
+
+⋆ `galoisConnection`: we prove that the maps `leftDual` and `rightDual` form a Galois connection.
+⋆ `equivFixedPoints`: we prove that the said maps induce inverse bijections between the sets
```
YaelDillies, line 22: And this one too

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+of fixed points.
```
YaelDillies, line 23:
```suggestion
  of fixed points.
```

Mathlib/Order/RelToGaloisConnection.lean
```diff
@@ -0,0 +1,109 @@
...
+
+## References
+
+⋆ Engendrement de topologies, démontrabilité et opérations sur les sous-topos, Olivia Caramello and
+Laurent Lafforgue (in preparation)
```
YaelDillies, line 28:
```suggestion
⋆ Engendrement de topologies, démontrabilité et opérations sur les sous-topos, Olivia Caramello and
  Laurent Lafforgue (in preparation)
```
or
```suggestion
Engendrement de topologies, démontrabilité et opérations sur les sous-topos, Olivia Caramello and
Laurent Lafforgue (in preparation)
```

Mathlib/Order/RelToGaloisConnection.lean
YaelDillies, line 1: After some thought, `Order.Rel.GaloisConnection` seems to be a better name for this file

Mathlib/Order/ReltoGaloisConnection.lean
adamtopaz: The only comment I have about these tactic blocks is that the `unfold` step is (most likely) unnecessary, and that's one tactic that we tend to avoid in mathlib proofs. @AnthonyBordg, could you try to get rid of these `unfold`s in your proofs?

Mathlib/Order/RelToGaloisConnection.lean
AnthonyBordg: Some help with this error `uncaught exception: no such file or directory (error code: 2)
  file: Mathlib/Order/«Order/Rel/GaloisConnection».lean`?

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: Done. Howewer, I have an error after the renaming the file `Order.Rel.GaloisConnection` as suggested by Yael.

Mathlib/Order/RelToGaloisConnection.lean
adamtopaz: The intention was to rename it to `GaloisConnection.lean` in the `Mathlib/Order/Rel/` directory (so that it could be imported with `import Mathlib.Order.Rel.GaloisConnection`).

Mathlib/Order/ReltoGaloisConnection.lean
YaelDillies: Seems like you forgot to run `lake exe mk_all`

Mathlib/Order/RelToGaloisConnection.lean
AnthonyBordg: I do not see any `Mathlib/Order/Rel` directory, am I supposed to create one?

YaelDillies: Yes please!

Mathlib/Order/ReltoGaloisConnection.lean
AnthonyBordg: I did run `lake exe mk_all`.

Mathlib/Order/Rel/GaloisConnection.lean
```diff
@@ -0,0 +1,108 @@
...
+theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) :
+    J ≤ J' → R.rightDual (R.leftDual J) ≤ J' := by
+  intro h₁
+  rw [← h]
+  apply R.gc_leftDual_rightDual.monotone_u
+  apply R.gc_leftDual_rightDual.monotone_l
+  exact h₁
+
+theorem ge_leftDual_rightDual_of_ge {I I' : Set β} (h : I' ∈ R.rightFixedPoints) :
+    I' ≥ I → I' ≥ R.leftDual (R.rightDual I) := by
```
YaelDillies, line 99: Please never `≥` but always `≤`. We don't want people having to guess two names and two statements
```suggestion
theorem leftDual_rightDual_le_of_le {I I' : Set β} (h : I' ∈ R.rightFixedPoints) :
    I ≤ I' → R.leftDual (R.rightDual I) ≤ I' := by
```
also you might as well pull out the assumption to before the colon
```suggestion
theorem leftDual_rightDual_le_of_le {I I' : Set β} (h : I' ∈ R.rightFixedPoints) (h₁ : I ≤ I') :
    R.leftDual (R.rightDual I) ≤ I' := by
```

AnthonyBordg: > Please never `≥` but always `≤`. We don't want people having to guess two names and two statements

Make sense!

Mathlib/Order/Rel/GaloisConnection.lean
```diff
@@ -87,17 +87,15 @@ def equivFixedPoints : R.leftFixedPoints ≃ R.rightFixedPoints where
   left_inv J := by cases' J with J hJ; simp; rw [hJ]
   right_inv I := by cases' I with I hI; simp; rw [hI]
 
-theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) :
-    J ≤ J' → R.rightDual (R.leftDual J) ≤ J' := by
-  intro h₁
+theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) (h₁ : J ≤ J'):
```
YaelDillies, line 90:
```suggestion
theorem rightDual_leftDual_le_of_le {J J' : Set α} (h : J' ∈ R.leftFixedPoints) (h₁ : J ≤ J') :
```

Mathlib/Order/Rel/GaloisConnection.lean
```diff
@@ -0,0 +1,106 @@
...
+
+⋆ Engendrement de topologies, démontrabilité et opérations sur les sous-topos, Olivia Caramello and
+  Laurent Lafforgue (in preparation)
+
+## Tags
+
+relation, Galois connection, induced bijection, fixed points
+-/
+
+section
```
YaelDillies, line 36: This section is kind of useless. Miht as well remove it

Mathlib/Order/Rel/GaloisConnection.lean
```diff
@@ -0,0 +1,106 @@
...
+
+universe u v
+
+variable {α : Type u} {β : Type v} (R : Rel α β)
```
YaelDillies, line 40: Unless necessary, we usually just do 
```suggestion
variable {α β : Type*} (R : Rel α β)
```

Mathlib/Order/Rel/GaloisConnection.lean
```diff
@@ -0,0 +1,102 @@
...
+⋆ Engendrement de topologies, démontrabilité et opérations sur les sous-topos, Olivia Caramello and
+  Laurent Lafforgue (in preparation)
+
+## Tags
+
+relation, Galois connection, induced bijection, fixed points
+-/
+
+universe u v
+
```
YaelDillies, line 37: Don't need these anymore
```suggestion
```

YaelDillies (approved):
Thanks!

maintainer merge

adamtopaz:
Thanks!
bors r+

PR #14910, opened 2024-07-19
chore: replace slow `convert` in `RingTheory/PowerSeries/Basic`

rwst:
Reduce compile time of `RingTheory/PowerSeries/Basic.lean` by 7-10s by replacing
a `convert` with a specific `erw`.

---


riccardobrasca:
!bench

riccardobrasca:
This is subsumed by #14913 right?

j-loreaux:
> This is subsumed by #14913 right?

Seems so, and that one is a much better change (i.e., not relying on defeq abuse within `erw`). I'm closing this. @rwst can reopen if he disagrees.

PR #14912, opened 2024-07-19
[Merged by Bors] - chore: remove `import Mathlib.Mathport.Rename`

riccardobrasca:
We remove all `import Mathlib.Mathport.Rename`.

This is done using

```bash
find . -name "*.lean" -exec sed -i '/^import Mathlib.Mathport.Rename/d' {} +
```

plus two fixing by hand of files that were missing various imports.

---

### DIFF
--- a/Mathlib/RepresentationTheory/Rep.lean
+++ b/Mathlib/RepresentationTheory/Rep.lean
@@ -424,7 +424,6 @@ def MonoidalClosed.linearHomEquiv : (A ⊗ B ⟶ C) ≃ₗ[k] B ⟶ A ⟶[Rep k 
   { (ihom.adjunction A).homEquiv _ _ with
     map_add' := fun _ _ => rfl
     map_smul' := fun _ _ => rfl }
-  set_option linter.uppercaseLean3 false in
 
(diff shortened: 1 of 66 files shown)


grunweg (commented):
Thanks; looks good!
maintainer merge

semorrison:
bors d+

Mathlib/RepresentationTheory/Rep.lean
```diff
@@ -424,7 +424,6 @@ def MonoidalClosed.linearHomEquiv : (A ⊗ B ⟶ C) ≃ₗ[k] B ⟶ A ⟶[Rep k
   { (ihom.adjunction A).homEquiv _ _ with
     map_add' := fun _ _ => rfl
     map_smul' := fun _ _ => rfl }
-  set_option linter.uppercaseLean3 false in
```
grunweg, line 427: FWIW, #14914 also includes this: you certainly need not search for all those comments. (Probably, this won't cause a conflict either - well, at least I hope so.)

riccardobrasca:
bors merge

PR #14922, opened 2024-07-19
[Merged by Bors] - feat (Algebra/Module/LinearMap/Defs): the identity map of the zero module is the zero map

matematiflo:
Add a lemma about the identity map of the zero module.

This contribution was created as part of the AIM workshop *Formalizing algebraic geometry* in June 2024.

---

### DIFF
--- a/Mathlib/Algebra/Module/LinearMap/Defs.lean
+++ b/Mathlib/Algebra/Module/LinearMap/Defs.lean
@@ -991,6 +991,10 @@ def toAddMonoidHom' : (M →ₛₗ[σ₁₂] M₂) →+ M →+ M₂ where
   map_add' := by intros; ext; rfl
 #align linear_map.to_add_monoid_hom' LinearMap.toAddMonoidHom'
 
+/-- If `M` is the zero module, then  the identity map of `M` is the zero map. -/
+theorem identityMapOfZeroModuleIsZero [Subsingleton M] : id (R := R₁) (M := M) = 0 :=
+  Subsingleton.eq_zero id
+
 end Arithmetic
 
 section Actions
--- a/Mathlib/Algebra/Module/LinearMap/Defs.lean
+++ b/Mathlib/Algebra/Module/LinearMap/Defs.lean
@@ -992,6 +992,7 @@ def toAddMonoidHom' : (M →ₛₗ[σ₁₂] M₂) →+ M →+ M₂ where
 #align linear_map.to_add_monoid_hom' LinearMap.toAddMonoidHom'
 
 /-- If `M` is the zero module, then  the identity map of `M` is the zero map. -/
+@[simp]
 theorem identityMapOfZeroModuleIsZero [Subsingleton M] : id (R := R₁) (M := M) = 0 :=
   Subsingleton.eq_zero id
 


Mathlib/Algebra/Module/LinearMap/Defs.lean
```diff
@@ -991,6 +991,10 @@ def toAddMonoidHom' : (M →ₛₗ[σ₁₂] M₂) →+ M →+ M₂ where
   map_add' := by intros; ext; rfl
 #align linear_map.to_add_monoid_hom' LinearMap.toAddMonoidHom'
 
+/-- If `M` is the zero module, then  the identity map of `M` is the zero map. -/
+theorem identityMapOfZeroModuleIsZero [Subsingleton M] : id (R := R₁) (M := M) = 0 :=
```
riccardobrasca, line 995:
```suggestion
@[simp]
theorem identityMapOfZeroModuleIsZero [Subsingleton M] : id (R := R₁) (M := M) = 0 :=
```

riccardobrasca (approved):
Thanks!

bors d+

riccardobrasca:
bors r-

What do you think about adding the `simp` tag?

Mathlib/Algebra/Module/LinearMap/Defs.lean
matematiflo: Done! Thanks 👍🏻

riccardobrasca:
@matematiflo don't worry if you are confused about `git`, but:
* the `simp` isn't added, I don't know whether you forgot to push or you messed up something in trying to apply my suggestion.
* in general, even if you get a delegation, please wait for CI (in practice wait for all the checks to be green) before merging.

But again, don't worry, the first time the process is a little confusing!

riccardobrasca:
Note that since I `bors r-`, one check will stay red (this is a bug in github in my opinion). You can safely ignore that. But all the other checks should be green.

matematiflo (commented):
Add `@[simp]` attribute. Will try `bors r+` again next.

riccardobrasca:
> bors r+

immagine
You should wait to the yellow thing here to become green. :D

riccardobrasca:
The rationale is the following: when you merge it, a bot will start working, but not only on your PR, it creates batches of several PRs at the same time. If your PR has a problem the whole batch will fail, and everybody will loose some time (nothing serious!). This is why we ask author to "wait for CI": this means that another bot has checked that your PR is OK.

bors r-

riccardobrasca:
Anyway this means that our instructions are not clear, it's surely not your fault.

riccardobrasca:
For example I suspect that a linter will complain saying that `simp can prove this, so it cannot be simp`. Of course this means that my suggestion wasn't accurate, but I am just lazy to test it by myself.

matematiflo:
OK. What threw me off is the message from 1h ago saying "✌️ matematiflo can now approve this pull request. To approve and merge a pull request, simply reply with bors r+. More detailed instructions are available [here](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)." I simply tried to follow that instruction.

riccardobrasca:
Yeah, that message is not accurate. I don't know whether it depends on us or on the bot. Let me stress that this is not a problem at all, even if you try to merge something that does not work at end, the worst can happen is that a computer waist some time, you cannot break mathlib.

matematiflo:
I checked again and `simp` cannot solve this, so we're safe on that front. As for the rest of the merging process, I am available if further actions are required on my end.

riccardobrasca:
bors merge

PR #14925, opened 2024-07-19
[Merged by Bors] - feat: results on the spectrum of elements in closed subalgebras of Banach algebras

j-loreaux:
The main result is well-known. Let `S` be a closed subalgebra of a Banach algebra `A`. If `x : S`, then the spectrum of `x` (in `S`) is the spectrum of `↑x : A` (in `A`) along with whichever, necessarily bounded, components of the complement (of the `A` spectrum) it happens to intersect.

Along the way we prove a few other useful results. For example, the boundary of the spectrum of `x` (in `S`) is a subset of the boundary of the spectrum of `↑x : A` (in `A`).

---

- [x] depends on: #14926 
- [x] depends on: #14927


Mathlib/Analysis/Normed/Algebra/Spectrum.lean
```diff
@@ -568,6 +568,114 @@ end CharacterSpace
...
+variable (S : SA) [hS : IsClosed (S : Set A)] (x : S)
+
+open Topology Filter
+
+open SubalgebraClass in
+/-- Let `S` be a closed subalgebra of a Banach algebra `A`. If `a : S` is invertible in `A`,
+and for all `x : S` sufficiently close to `a` within some filter `l`, `x` is invertible in `S`,
+then `a` is invertible in `S` as well. -/
+lemma _root_.Subalgebra.isUnit_of_isUnit_val_of_eventually {l : Filter S} {a : S}
+    (ha : IsUnit (a : A)) (hla : l ≤ 𝓝 a) (hl : ∀ᶠ x in l, IsUnit x) (hl' : l.NeBot) :
```
dupuisf, line 586:
```suggestion
lemma Subalgebra.isUnit_of_isUnit_val_of_eventually {l : Filter S} {a : S}
    (ha : IsUnit (a : A)) (hla : l ≤ 𝓝 a) (hl : ∀ᶠ x in l, IsUnit x) (hl' : l.NeBot) :
```
We're already in the root namespace (but you probably want to put all of this in a `namespace Subalgebra` block).

j-loreaux: This is actually there on purpose to avoid having the namespace open to avoid collisions.

dupuisf: Oh I see.

Mathlib/Analysis/Normed/Algebra/Spectrum.lean
```diff
@@ -568,6 +568,114 @@ end CharacterSpace
...
+  suffices Ring.inverse (val S x) = (val S ↑hx.unit⁻¹) from this ▸ Subtype.property _
+  rw [← (hx.map (val S)).unit_spec, Ring.inverse_unit (hx.map (val S)).unit, val]
+  apply Units.mul_eq_one_iff_inv_eq.mp
+  simpa [-IsUnit.mul_val_inv] using congr(($hx.mul_val_inv : A))
+
+open Set
+
+/-- If `S : Subalgebra 𝕜 A` is a closed subalgebra of a Banach algebra `A`, then for any
+`x : S`, the boundary of the spectrum of `x` relative to `S` is a subset of the spectrum of
+`↑x : A` relative to `A`. -/
```
dupuisf, line 607:
```suggestion
/-- If `S` is a closed subalgebra of a Banach algebra `A`, then for any
`x : S`, the boundary of the spectrum of `x` relative to `S` is a subset of the spectrum of
`↑x : A` relative to `A`. -/
```
Since it's not literally of that type.

dupuisf (commented):
Looks good!

bors d+

PR #14927, opened 2024-07-19
[Merged by Bors] - feat: if one set is clopen in another, then it is the union of connected components in the larger set

j-loreaux:
---

- [x] depends on: #15144

I'm happy to put `isClopen_preimage_val` somewhere else, but I'm simply not sure where it should go.


Mathlib/Topology/Connected/Basic.lean
```diff
@@ -3,7 +3,7 @@ Copyright (c) 2017 Johannes Hölzl. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Johannes Hölzl, Mario Carneiro, Yury Kudryashov
 -/
-import Mathlib.Data.Set.Image
+import Mathlib.Data.Set.Subset
```
semorrison, line 6: Adding an import that is only used after line 1000! :-) could this file be split?

j-loreaux: In fairness, it's a single file import (including transitive)

j-loreaux: #15144

Mathlib/Topology/Connected/Clopen.lean
```diff
@@ -121,6 +122,16 @@ theorem IsClopen.eq_univ [PreconnectedSpace α] {s : Set α} (h' : IsClopen s) (
     s = univ :=
   (isClopen_iff.mp h').resolve_left h.ne_empty
 
+open Set.Notation in
+lemma isClopen_preimage_val {X : Type*} [TopologicalSpace X] {u v : Set X}
+    (hu : IsOpen u) (huv : frontier u ∩ v = ∅) : IsClopen (v ↓∩ u) := by
```
Ruben-VandeVelde, line 127: Is `frontier u ∩ v = ∅` the best way to state this? Do we not have a disjointness predicate?

j-loreaux: Thanks, good point, I'll make that change.

Ruben-VandeVelde (approved):
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #14928, opened 2024-07-19
Feat(RingTheory/Flat): add FaithfullyFlat file with def and basic results

judithludwig:
Add definition of faithfully flat ring maps and show basic properties.
Co-authored-by: Florent Schaffhauser (github: matematiflo).

This contribution was created as part of the AIM workshop "Formalizing algebraic geometry" in
June 2024.

---


Mathlib/RingTheory/Flat/FaithfullyFlat.lean
```diff
@@ -0,0 +1,248 @@
...
+
+universe u v
+
+namespace Module
+
+variable (R : Type u) (M : Type v) [CommRing R] [AddCommGroup M] [Module R M]
+
+@[mk_iff] class FaithfullyFlat : Prop where
+  flat : Module.Flat R M := by infer_instance
+  zero_if_lTensor_zero :  ∀ ⦃N N': Type v⦄ [AddCommGroup N][Module R N][AddCommGroup N']
```
erdOne, line 40: The quantification over all `Type v` is not ideal. I'm not even sure if this implies faithfully flat for all `Type w`. Maybe `𝔪 • (⊤ : Submodule R M) ≠ ⊤` for all maximal ideal `𝔪` as the official definition?

matematiflo: We've rewritten the file using just one universe level (the previous choice was due to the analogy with the case of [flat modules](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/RingTheory/Flat/Basic.lean)). We are working towards the proof of the equivalence between various definitions, but we'd like to keep this one as the main one for now, since we have already proven some properties using that characterisation of faithfully flat modules (WIP).

erdOne: For flat modules, the official definition is using ideals for this precise reason of avoiding universe issues. If we are sticking with this definition for now I prefer the hetero-universal one with the universe in the definition the same as the universe of the ring, which is at least the right definition. But please add the condition with general universe as a todo.

PR #14934, opened 2024-07-20
[Merged by Bors] - chore(*): use ⊕ notation for `Sum`

urkud:
---


riccardobrasca:
Very nice! Let's merge before it rots. Thanks!

bors merge

PR #14941, opened 2024-07-20
[Merged by Bors] - feat (Algebra/Ring): add `isSemireal` class for commutative rings

matematiflo:
A semireal ring is a non-trivial commutative ring with unit `R` in which `-1 : R` is not a sum of squares. We add this notion to Mathlib and prove that linearly ordered fields are semireal rings.

We also introduce the type of sums of squares in a type `R` endowed with an `Add R`, `Zero R` and `Mul R`  instances.

---


Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,116 @@
+/-
+Copyright (c) 2024 Matematiflo. All rights reserved.
```
riccardobrasca, line 2:
```suggestion
Copyright (c) 2024 Florent Schaffhauser. All rights reserved.
```
Note sure if the linter is so smart to realize you didn't use your real name.

matematiflo: Ok, I'll change this and remove the unnecessary import that was detected during CI checks.

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,115 @@
+/-
+Copyright (c) 2024 Florent Schaffhauser. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Florent Schaffhauser.
+-/
+
+import Mathlib.Algebra.Ring.Defs
```
riccardobrasca, line 7:
```suggestion
Authors: Florent Schaffhauser
-/
import Mathlib.Algebra.Ring.Defs
```
TBH I am not sure what is the problem here. At worst just copy/paste from another file and change your name (The VS Code extension can also produce it for you IIRC)

riccardobrasca: Ah, it's the dot at the end of the line

matematiflo: OK, I'll try again

matematiflo: The formatting seems correct now :+1:

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+- The type `SumSqIn R`: a subtype of `R` whose terms are the sums of squares in `R`.
+
+## Auxiliary theorems
+
+- `SumSq_nonneg`: sums of squares are non-negative.
+- `isSumSq_sum`: if `S₁` and `S₂` are sums of squares in a semiring `R`, then so is `S₁ + S₂`.
+
+## References
+
+- *An introduction to real algebra*, by T.Y. Lam. Rocky Mountain J. Math. 14(4): 767-814 (1984). [DOI: 10.1216/RMJ-1984-14-4-767](https://projecteuclid.org/journals/rocky-mountain-journal-of-mathematics/volume-14/issue-4/An-introduction-to-real-algebra/10.1216/RMJ-1984-14-4-767.full).
```
riccardobrasca, line 36: You can add this to the bibtex file (you have to put the item in alphabetical order by hand I think), and cite it as [foo]

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+-/
+
+/--
+In a semiring `R`, the property of being a sum of squares is an inductive predicate: `0 : R` is a
+sum of squares and if `S` is a sum of squares, then for all `a : R`, `a ^ 2 + S` is a sum of squares
+in `R`.
+-/
+
+inductive isSumSq.{u} {R : Type u} [Semiring R] : R → Prop
```
riccardobrasca, line 45:
```suggestion
@[mk_iff]
inductive isSumSq.{u} {R : Type u} [Semiring R] : R → Prop
```
This automatically produce the relevant `iff` lemma

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  | zero                           : isSumSq (0 : R)
+  | add (x S : R) (pS : isSumSq S) : isSumSq (x ^ 2 + S)
+
+/--
+A semireal ring is a non-trivial commutative ring (with unit) in which `-1` is *not* a sum of
+squares. Note that `-1` does not make sense in a semiring.
+-/
+
+@[class, mk_iff]
+structure isSemireal.{u} (R : Type u) [CommRing R] : Prop where
```
riccardobrasca, line 55:
```suggestion
structure isSemireal (R : Type*) [CommRing R] : Prop where
```
Should be exactly the same

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  non_trivial        : (0 : R) ≠ 1
+  neg_one_not_SumSq  : ¬isSumSq (-1 : R)
+
+/--
+Let `R` be a linearly ordered semiring `R` in which the property `a ≤ b → ∃ c, a + c = b` holds
+(e.g. `R = ℕ`). If `S : R` is a sum of squares in `R`, then `0 ≤ S`. This is used to show that
+linearly ordered fields are semireal.
+-/
+
+theorem SumSq_nonneg  {R : Type _} [LinearOrderedSemiring R] [ExistsAddOfLE R] {S : R} :
```
riccardobrasca, line 65:
```suggestion
theorem SumSq_nonneg  {R : Type*} [LinearOrderedSemiring R] [ExistsAddOfLE R] {S : R} :
```
Similarly elsewhere.

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+    isSumSq S → 0 ≤ S := by
+  intro pS
+  induction pS with
```
riccardobrasca, line 68:
```suggestion
theorem SumSq_nonneg  {R : Type _} [LinearOrderedSemiring R] [ExistsAddOfLE R] {S : R}
  (pS : isSumSq S) : 0 ≤ S := by
  induction pS with
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  | zero          => simp only [le_refl]
+  | add x S _ ih  =>
+    apply add_nonneg ?_ ih
+    simp only [sq_nonneg]
+
+instance {R : Type _} [LinearOrderedField R] : isSemireal R where
```
riccardobrasca, line 74:
```suggestion
instance {R : Type*} [LinearOrderedField R] : isSemireal R where
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  /--Terms of type `SumSqIn R` are dependent pairs `⟨x, hx⟩` where `x : R` and `hx` is a proof of
+  the proposition `isSumSq x`-/ --#
+  val  : R
+  ppty : isSumSq val
+
+/--
+If `S1` and `S2` are sums of squares in a semiring `R`, then `S1 + S2` is a sum of squares in `R`.
+-/
+
+theorem isSumSq_sum {R : Type} [Semiring R] {S1 S2 : R} (h1 : isSumSq S1) (h2 : isSumSq S2) :
```
riccardobrasca, line 99:
```suggestion
theorem isSumSq_sum {R : Type*} [Semiring R] {S1 S2 : R} (h1 : isSumSq S1) (h2 : isSumSq S2) :
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+    apply not_le.mpr
+    exact neg_one_lt_zero
+
+/--
+The type of sums of squares in a semiring `R` is the subtype of `R` defined by the predicate
+`isSumSq : R → Prop`. It  can be defined as a structure whose terms are dependent pairs `⟨x, hx⟩`
+where `x : R` and `hx` is a proof of  the proposition `isSumSq x`.
+-/
+
+structure SumSqIn.{u} (R : Type u) [Semiring R] where
```
riccardobrasca, line 89:
```suggestion
structure SumSqIn (R : Type*) [Semiring R] where
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  | zero => simp only [zero_add]; exact h2
+  | add a S hS ih =>  rw [add_assoc]; exact isSumSq.add a (S + S2) ih
+
+/--
+Given terms `(S₁ S₂ : SumSqIn R)` where `Sᵢ = ⟨xᵢ, hxᵢ⟩`, we define `S1 + S2` to be the dependent
+pair `⟨x₁ + x₂, h⟩`, where `h` is a proof that `x₁ + x₂` is a sum of squares in `R`. The term `h` is
+obtained by applying the function `isSumSq_sum`.
+-/
+
+def add_SumSq {R : Type} [Semiring R] : SumSqIn R → SumSqIn R → SumSqIn R :=
```
riccardobrasca, line 111:
```suggestion
def add_SumSq {R : Type*} [Semiring R] : SumSqIn R → SumSqIn R → SumSqIn R :=
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+  fun (S1 S2 : SumSqIn R) => ⟨S1.val + S2.val, isSumSq_sum S1.ppty S2.ppty⟩
+
+instance {R : Type} [Semiring R] : Add (SumSqIn R) := ⟨add_SumSq⟩
```
riccardobrasca, line 114:
```suggestion
instance {R : Type*} [Semiring R] : Add (SumSqIn R) := ⟨add_SumSq⟩
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
riccardobrasca, line 65: I would call this `isSumSq.nonneg` to activate dot notation.

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,114 @@
...
+
+/--
+The type of sums of squares in a semiring `R` is the subtype of `R` defined by the predicate
+`isSumSq : R → Prop`. It  can be defined as a structure whose terms are dependent pairs `⟨x, hx⟩`
+where `x : R` and `hx` is a proof of  the proposition `isSumSq x`.
+-/
+
+structure SumSqIn.{u} (R : Type u) [Semiring R] where
+  /--Terms of type `SumSqIn R` are dependent pairs `⟨x, hx⟩` where `x : R` and `hx` is a proof of
+  the proposition `isSumSq x`-/ --#
```
riccardobrasca, line 91:
```suggestion
  the proposition `isSumSq x`-/
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
riccardobrasca: Also here, `isSumSq.sum` seems a better name (still for dot notation).

riccardobrasca (changes requested):
You have a lot of `(R : Type*) [Semiring R]`: you can probably declare it at the beginning of the section in a `variable` line. If you want to change impliciteness in only one declaration you can use `variable (R) in` just before the declaration.

It think it makes sense to create a file `isSumsq` in some other folder, and all the results that are related to being a sum of a square should go into the `isSumsq` namespace. Indeed, this seem rather a PR about being sum of a square rather then semireal (I understand it contains prerequisites of course).

Mathlib/Algebra/Ring/Semireal/Defs.lean
matematiflo: Well, my impression is that dot notation does not really help here, because the function takes `h1` and `h2` as inputs, not `S1` and `S2`, and `h1.sum h2` does not really reflect that `S1 + S2` is a sum of squares. If you prefer, I can make `S1` and `S2` explicit and then we'll have `S1.sum S2 h1 h2` as proof of `isSumSq(S1 + S2)`.

riccardobrasca: Ah, first of all it should be called `add`, not `sum` (`sum` is for big sums, with Sigma). If you call it `isSumSq.add` I think we can write `h1.h2` to mean the addition. Anyway we tend to follow this even if "dot notation does not fire", since it is very practical to have everything in the same namespace.

riccardobrasca (commented):
Indeed it is probably better to define the subblah of element that are sum of squares, so all the API becomes automatically available. It is at least a `AddSubmonoid`.

matematiflo:
Right, the issue is that `isSumSq.add` is already taken... do you want me to rename it in the definition of the predicate `isSumSq`?

riccardobrasca:
> Right, the issue is that `isSumSq.add` is already taken... do you want me to rename it in the definition of the predicate `isSumSq`?

Yes, that field should be called `sq_add` or something. Anyway I really think is better to define the submonoid of elements that are sum of squares. In practice it is the same you are doing, but in this way you got theorems like `isSumsq.sum` (with a big Sigma) for free, or the fact that the property is preserved by ring hom.

matematiflo:
> > Right, the issue is that `isSumSq.add` is already taken... do you want me to rename it in the definition of the predicate `isSumSq`?
> 
> Yes, that field should be called `sq_add` or something. Anyway I really think is better to define the submonoid of elements that are sum of squares. In practice it is the same you are doing, but in this way you got theorems like `isSumsq.sum` (with a big Sigma) for free, or the fact that the property is preserved by ring hom.

OK, `sq_add` it is! At this stage I don't need the submonoid of elements that are sums of squares so I would like to first separate the stuff about semireal rings from the stuff about sums of squares (two different files, as you suggested), and then I can possibly refactor the code using submonoids when I need it.

I would really prefer to focus on semireal rings for now, and not spend too much time on sums of squares.

Mathlib/Algebra/Ring/Semireal/Defs.lean
matematiflo: I added it to the references.bib file, but I am not sure how to make the citation work, I'll try!

riccardobrasca:
> > > Right, the issue is that `isSumSq.add` is already taken... do you want me to rename it in the definition of the predicate `isSumSq`?
> > 
> > 
> > Yes, that field should be called `sq_add` or something. Anyway I really think is better to define the submonoid of elements that are sum of squares. In practice it is the same you are doing, but in this way you got theorems like `isSumsq.sum` (with a big Sigma) for free, or the fact that the property is preserved by ring hom.
> 
> OK, `sq_add` it is! At this stage I don't need the submonoid of elements that are sums of squares so I would like to first separate the stuff about semireal rings from the stuff about sums of squares (two different files, as you suggested), and then I can possibly refactor the code using submonoids when I need it.
> 
> I would really prefer to focus on semireal rings for now, and not spend too much time on sums of squares.

Note that it is enough to write
```lean
inductive isSumSq {R : Type*} [Mul R] [Add R] [Zero R] : R → Prop
  | zero                           : isSumSq 0
  | add (x S : R) (pS : isSumSq S) : isSumSq (x * x + S)

def AddSubmonoid.isSumSq (R : Type*) [Ring R] : AddSubmonoid R where
  carrier := {r : R | _root_.isSumSq r}
  zero_mem' := isSumSq.zero
  add_mem' := sorry
```
and instead of `isSumSq r` one writes `hr : r ∈ isSumSq R` (after `open AddSubmonoid` or some other random name). `induction hr` should work as expected. I think this allows to concentrate on something else, as you automatically get a lot of results that you may need.

matematiflo:
> > > > Right, the issue is that `isSumSq.add` is already taken... do you want me to rename it in the definition of the predicate `isSumSq`?
> > > 
> > > 
> > > Yes, that field should be called `sq_add` or something. Anyway I really think is better to define the submonoid of elements that are sum of squares. In practice it is the same you are doing, but in this way you got theorems like `isSumsq.sum` (with a big Sigma) for free, or the fact that the property is preserved by ring hom.
> > 
> > 
> > OK, `sq_add` it is! At this stage I don't need the submonoid of elements that are sums of squares so I would like to first separate the stuff about semireal rings from the stuff about sums of squares (two different files, as you suggested), and then I can possibly refactor the code using submonoids when I need it.
> > I would really prefer to focus on semireal rings for now, and not spend too much time on sums of squares.
> 
> Note that it is enough to write
> 
> ```lean
> inductive isSumSq {R : Type*} [Mul R] [Add R] [Zero R] : R → Prop
>   | zero                           : isSumSq 0
>   | add (x S : R) (pS : isSumSq S) : isSumSq (x * x + S)
> 
> def AddSubmonoid.isSumSq (R : Type*) [Ring R] : AddSubmonoid R where
>   carrier := {r : R | _root_.isSumSq r}
>   zero_mem' := isSumSq.zero
>   add_mem' := sorry
> ```
> 
> and instead of `isSumSq r` one writes `hr : r ∈ isSumSq R` (after `open AddSubmonoid` or some other random name). `induction hr` should work as expected. I think this allows to concentrate on something else, as you automatically get a lot of results that you may need.

Yes, I understand :-) And the structure of submonoid will appear naturally at some point. But I really meant to take the more elementary approach. I hope that's ok for now.

I like the refactor of the predicate `isSumSq` with weaker assumptions on `R`, though, and I will incorporate it next!

Mathlib/Algebra/Ring/Semireal/SumsOfSquares.lean
```diff
@@ -58,7 +57,7 @@ theorem isSumSq.nonneg  {R : Type*} [LinearOrderedSemiring R] [ExistsAddOfLE R]
   | zero          => simp only [le_refl]
   | sq_add x S _ ih  =>
     apply add_nonneg ?_ ih
-    simp only [sq_nonneg]
+    simp only [Eq.symm (pow_two x), sq_nonneg]
```
riccardobrasca, line 60:
```suggestion
    simp only [← pow_two x, sq_nonneg]
```

matematiflo:
> Note that it is enough to write
> 
> ```lean
> inductive isSumSq {R : Type*} [Mul R] [Add R] [Zero R] : R → Prop
>   | zero                           : isSumSq 0
>   | add (x S : R) (pS : isSumSq S) : isSumSq (x * x + S)
> 
> def AddSubmonoid.isSumSq (R : Type*) [Ring R] : AddSubmonoid R where
>   carrier := {r : R | _root_.isSumSq r}
>   zero_mem' := isSumSq.zero
>   add_mem' := sorry
> ```
> 
> and instead of `isSumSq r` one writes `hr : r ∈ isSumSq R` (after `open AddSubmonoid` or some other random name). `induction hr` should work as expected. I think this allows to concentrate on something else, as you automatically get a lot of results that you may need.

Yes. Although, in that respect, it might be even better to consider directly the submonoid generated by the squares.

Mathlib/Algebra/Ring/Semireal/SumsOfSquares.lean
```diff
@@ -0,0 +1,93 @@
...
+    apply add_nonneg ?_ ih
+    simp only [sq_nonneg]
+
+/--
+The type of sums of squares in a semiring `R` is the subtype of `R` defined by the predicate
+`isSumSq : R → Prop`. It  can be defined as a structure whose terms are dependent pairs `⟨x, hx⟩`
+where `x : R` and `hx` is a proof of  the proposition `isSumSq x`.
+-/
+
+structure SumSqIn (R : Type*) [Semiring R] where
```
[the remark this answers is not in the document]
matematiflo: Yes. Although, in that respect, it might be even better to consider directly the submonoid generated by the squares.

matematiflo:
I have moved the `SumsOfSquares` file into the main `Ring` folder. 

I have left a `TODO` in that file, about a possible generalisation from semirings to types `R` with `AddZeroClass R` and `AddSemigroup R` instances (for some reason, having just these two does not work, there must be a distributivity property, or some other compatibility condition between `Add R` and `Mul R` used somewhere in the file and I am not sure what is the minimal framework for this; certainly in semirings it works).

The `Semireal.Defs` file should be good for now. Of course it will be expanded later. So it is basically down to the `SumsOfSquares` file. Do you want this PR to include the fact that it carries an `AddSubmonoid` structure? I should be able to include that feature soon, but I want to make sure I understand that this is the requested change. Thanks!

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,97 @@
...
+  /-- `val` is a term in `R` and `ppty` is a proof of the proposition `isSumSq val` -/
+  val  : R
+  ppty : isSumSq val
+
+/--
+If `S1` and `S2` are sums of squares in a semiring `R`, then `S1 + S2` is a sum of squares in `R`.
+-/
+
+/-
+*TODO*: `isSumSq.add` could be rewritten with weaker assumptions on `R` (we only need to
```
matematiflo, line 65: There should be a way to generalise `SumSq.add` from semirings to types `R` with `AddZeroClass R` and `AddSemigroup R` instances (for some reason, having just these two does not work, there must be a distributivity property, or some other compatibility condition between `Add R` and `Mul R` used somewhere in the file and I am not sure what is the minimal framework for this; certainly in semirings it works).

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,97 @@
...
+  | zero                              : isSumSq 0
+  | sq_add (x S : R) (pS : isSumSq S) : isSumSq (x * x + S)
+
+/--
+The type of sums of squares in a semiring `R` is the subtype of `R` defined by the predicate
+`isSumSq : R → Prop`. It  can be defined as a structure whose terms are dependent pairs `⟨x, hx⟩`
+where `x : R` and `hx` is a proof of  the proposition `isSumSq x`.
+-/
+
+structure SumSqIn (R : Type*) [Add R] [Mul R] [Zero R] where
```
riccardobrasca, line 55: I am sorry, but this definition is plainly wrong. By this I mean that what you are redefining essentially a subtype, but ignoring the library, and this is a recipe for suffering later. If you want to work only with `isSumSq` that is OK, but if you want something like `SumSqIn` (mathematically the subset of elements that are sum of squares) then `AddSubmoniid` is absolutely the way to go (or some other `...Sub` class if you want to think about minimal assumptions to have various properties, but this is not very important). I perfectly understand that you don't want to focus in this, but I think this is the way to spend *less* time with trivial properties of sum of squares.

Mathlib/Algebra/Ring/SumsOfSquares.lean
riccardobrasca: Let me see.

Mathlib/Algebra/Ring/SumsOfSquares.lean
matematiflo: OK, I'll change it then ☺️ But I don't think it is necessary to call it wrong. It was just a design choice.

Mathlib/Algebra/Ring/SumsOfSquares.lean
riccardobrasca:
```lean
theorem isSumSq.add {R : Type*} [Mul R] [AddMonoid R] {S1 S2 : R} (h1 : isSumSq S1) (h2 : isSumSq S2) :
    isSumSq (S1 + S2) := by
  induction h1 with
  | zero             => rwa [zero_add S2]
  | sq_add a S hS ih =>
    rw [add_assoc]
    exact sq_add _ _ ih
```

riccardobrasca: `AddZeroClass` plus `AddSemigroup` is exactly `AddMonoid` (not by definition, but it seems trivially true to me)

matematiflo: Yes! If I use `[AddMonoid R] [Mul R]`, it works 🙏🏻 

**EDIT:** Oh, I didn't see your code above 😓 Thanks!

Mathlib/Algebra/Ring/SumsOfSquares.lean
riccardobrasca: Yeah, sorry for that. I mean that I think it will make things more complicated later. Of course it is not mathematically wrong.

matematiflo: Right, and you have a point, too, when you say that it is more consistent with Mathlib this way, which is of course very important since this is a PR to Mathlib 😅 

I will retake your suggestion


```lean
def AddSubmonoid.isSumSq (R : Type*) [Ring R] : AddSubmonoid R where
  carrier := {r : R | _root_.isSumSq r}
  zero_mem' := isSumSq.zero
  add_mem' := sorry
```

and start from there.

riccardobrasca: Using what you already have that `sorry` is done. Then you can start working on semireal field, and adding to this file only the results you need.

matematiflo: Please ignore the question if you don't have time: what does `_root_` mean in the definition of the `carrier`?

riccardobrasca: In general if you have two declaration `bar` and `foo.bar` and you have `open foo` then sometimes Lean is confused if you write `bar` (as it doesn't understand if you want `bar` or `foo.bar`). Writing  `_root_.bar` means "I want the declaration called exactly `bar`, ignoring what I've opened.

Here what's happening is a little subtle: since we are defining something called `AddSubmonoid.isSumSq` Lean automatically opens the namespace `AddSubmonoid` inside the definition (this can be confusing sometimes, but it is usually useful). So now `isSumSq` can be `_root_.isSumSq` or `AddSubmonoid.isSumSq` and Lean is confused (not that we do inductive definitions all the time, so we really want to use a definition inside itself). If you call the submonoid `AddSubmonoid.isSumSqIn` or whatever the `_root_.` shouldn't be needed.

matematiflo: Ok (though I did not fully get the very last sentence). And we absolutely do want to add the `AddSubmonoid` prefix, right? You do not want to see the following, for instance:

```lean
def SumSqIn (R : Type*) [AddMonoid R] [Mul R] : AddSubmonoid R where
  carrier   := {S : R | isSumSq S}
  zero_mem' := isSumSq.zero
  add_mem'  := isSumSq.add
```

riccardobrasca: The name is much less important, I am not against this.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,77 @@
...
+-/
+
+/--
+In a type `R` with an addition, a zero element and a multiplication, the property of being a sum of
+squares is defined by an inductive predicate: `0 : R` is a sum of squares and if `S` is a sum of
+squares, then for all `a : R`, `a * a + S` is a sum of squares in `R`.
+-/
+
+@[mk_iff]
+inductive isSumSq {R : Type*} [Add R] [Zero R] [Mul R] : R → Prop
```
riccardobrasca, line 43:
```suggestion
-/
@[mk_iff]
inductive isSumSq {R : Type*} [Add R] [Zero R] [Mul R] (r : R) : Prop
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,77 @@
...
+  | zero                              : isSumSq 0
+  | sq_add (a S : R) (pS : isSumSq S) : isSumSq (a * a + S)
+
+/--
+If `S1` and `S2` are sums of squares in a semiring `R`, then `S1 + S2` is a sum of squares in `R`.
+-/
+
+theorem isSumSq.add {R : Type*} [AddMonoid R] [Mul R] {S1 S2 : R} (p1 : isSumSq S1)
```
riccardobrasca, line 51: Can you declare `variable {R : Type*}` at the beginning of the file?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,77 @@
...
+  induction p1 with
+  | zero             => rewrite [zero_add]; exact p2
+  | sq_add a S pS ih => rewrite [add_assoc]; exact isSumSq.sq_add a (S + S2) ih
+
+/--
+In an additive monoid with multiplication `R`, the type `SumSqIn R` is the submonoid of sums of
+squares in `R`.
+-/
+
+def SumSqIn (R : Type*) [AddMonoid R] [Mul R] : AddSubmonoid R where
```
riccardobrasca, line 62: Can you please add the fact that `SumSqIn` is the submonoid generated by the squares? It should be very easy and I think this is all we want to know about it.

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,46 @@
...
+A semireal ring is a non-trivial commutative ring (with unit) in which `-1` is *not* a sum of
+squares. Note that `-1` does not make sense in a semiring.
+-/
+
+@[class, mk_iff]
+structure isSemireal (R : Type*) [CommRing R] : Prop where
+  non_trivial        : (0 : R) ≠ 1
+  neg_one_not_SumSq  : ¬isSumSq (-1 : R)
+
+instance {R : Type*} [LinearOrderedField R] : isSemireal R where
```
riccardobrasca, line 40:
```suggestion
instance [LinearOrderedField R] : isSemireal R where
```
Adding `variable (R : Type*)` at the beginning of the file.

Mathlib/Algebra/Ring/SumsOfSquares.lean
matematiflo: Hmm, this particular one won't work, I think.

riccardobrasca: Ah sorry, you're right.

Mathlib/Algebra/Ring/SumsOfSquares.lean
matematiflo: Sure, I'll do that asap and resubmit!

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -27,17 +27,21 @@ semireal.
 [lam_1984](https://doi.org/10.1216/RMJ-1984-14-4-767)
 -/
 
+variable {R : Type*}
+
+variable (R) in
```
riccardobrasca, line 32: You can just have `R` explicit from the beginning (in both declarations).

matematiflo: OK! And then how do I make it implicit in certain declarations?

matematiflo: In the `SumsOfSquares` file, it makes more sense to me to start with `R` implicit. Is that a problem?

riccardobrasca: Variables in different files are completely independent, so you can do whatever you want. I don't understand why you want `R` implicit in the instance.

matematiflo: My bad, I was just testing out how to make implicit a variable that had been globally declared as explicit and I forgot to remove it 😖

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,128 @@
...
+  rw [py, ← AddMonoid.add_zero (y * y)]
+  apply isSumSq.sq_add
+  exact isSumSq.zero
+
+/--
+In an additive monoid with multiplication `R`, the submonoid generated by the squares is the set of
+sums of squares.
+-/
+
+theorem SquaresAddClosure [AddMonoid R] [Mul R] : AddSubmonoid.closure (Squares R) = SumSqIn R := by
```
riccardobrasca, line 101:
```suggestion
open AddSubmonoid in
/--
In an additive monoid with multiplication `R`, the submonoid generated by the squares is the set of
sums of squares.
-/
theorem SquaresAddClosure [AddMonoid R] : closure IsSquare = SumSqIn R := by
```
You see that Lean is perfectly happy with `IsSquare` here. This is identical to previous spelling, but we don't introduce a new definition that then needs to be unfolded.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,128 @@
...
+    apply px (SumSqIn R)
+    exact SquaresInSumSq
+  · intro px
+    apply AddSubmonoid.mem_closure.mpr
+    intro M pM
+    induction' px with a S _ pS
+    · exact AddSubmonoid.zero_mem M
+    · refine AddSubmonoid.add_mem M ?h.mpr.sq_add.a pS
+      apply pM
+      use a
```
riccardobrasca, line 116:
```suggestion
  refine le_antisymm (closure_le.2 (fun x hx ↦ SquaresInSumSq hx)) (fun x hx ↦ ?_)
  induction hx with
  | zero             => exact zero_mem _
  | sq_add a S _ ih  => exact AddSubmonoid.add_mem _ (subset_closure ⟨a, rfl⟩) ih
```
This proof is shorter, but I also think it closer to what we do mathematically: `SumSqIn` is a submnoid, so by the universal property of the closure to prove that `closure IsSquare` is included in `SumSqIn` is it enough to prove that `IsSquare` is in `SumSqIn`, and you have already proved it. The proof of the other inclusion is basically the same as yours, but we try to avoid using `induction'`, that sooner or later will disappear.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,128 @@
...
+
+/--
+Let `R` be a linearly ordered semiring in which the property `a ≤ b → ∃ c, a + c = b` holds
+(e.g. `R = ℕ`). If `S : R` is a sum of squares in `R`, then `0 ≤ S`. This is used in
+`Mathlib.Algebra.Ring.Semireal.Defs` to show that linearly ordered fields are semireal.
+-/
+
+theorem isSumSq.nonneg [LinearOrderedSemiring R] [ExistsAddOfLE R] {S : R}
+    (pS : isSumSq S) : 0 ≤ S := by
```
riccardobrasca, line 125:
```suggestion
/--
Let `R` be a linearly ordered semiring in which the property `a ≤ b → ∃ c, a + c = b` holds
(e.g. `R = ℕ`). If `S : R` is a sum of squares in `R`, then `0 ≤ S`. This is used in
`Mathlib.Algebra.Ring.Semireal.Defs` to show that linearly ordered fields are semireal.
-/
theorem isSumSq.nonneg (R : Type*) [LinearOrderedSemiring R] [ExistsAddOfLE R] {S : R}
    (pS : isSumSq S) : 0 ≤ S := by
```

riccardobrasca (approved):
A bunch of mostly stylistic comments, but after those it is ready to go, thanks!

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -0,0 +1,128 @@
...
+
+/--
+If `S1` and `S2` are sums of squares in a semiring `R`, then `S1 + S2` is a sum of squares in `R`.
+-/
+
+theorem isSumSq.add [AddMonoid R] [Mul R] {S1 S2 : R} (p1 : isSumSq S1)
+    (p2 : isSumSq S2) : isSumSq (S1 + S2) := by
+  induction p1 with
+  | zero             => rewrite [zero_add]; exact p2
+  | sq_add a S pS ih => rewrite [add_assoc]; exact isSumSq.sq_add a (S + S2) ih
```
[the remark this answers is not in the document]
matematiflo: Compared to `rewrite`, doesn't `rw` also try to close by `rfl`? Anyway, I changed it!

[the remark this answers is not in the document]
riccardobrasca: Ah yes, it's possible.

Mathlib/Algebra/Ring/SumsOfSquares.lean
matematiflo: This one does not quite work for me, I think I may be missing something regarding where you want me to use this (I assume in `SquaresAddClosure`).

matematiflo: No, but I added the prefix. I'll try again.

matematiflo: OK, got it!

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -0,0 +1,44 @@
...
+
+variable (R : Type*)
+
+/--
+A semireal ring is a non-trivial commutative ring (with unit) in which `-1` is *not* a sum of
+squares. Note that `-1` does not make sense in a semiring. Below we define the class `isSemiReal R`
+for all additive monoid `R` equipped with a multiplication, a multiplicative unit and a negation.
+-/
+@[mk_iff]
+class isSemireal [AddMonoid R] [Mul R] [One R] [Neg R] : Prop where
```
riccardobrasca, line 38:
```suggestion
class isSemireal [SubNegMonoid R] [Mul R] [One R] : Prop where
```
Crazily enough we already have a typeclass for a monoid with a random negation. But we don't have the variant `WithOne` I think (anyway who cares).

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/Algebra/Ring/Semireal/Defs.lean
matematiflo: Thanks for your help and assistance!

PR #14863, opened 2024-07-18
[Merged by Bors] - feat: Scaling of affine bases

YaelDillies:
Given a group `G` acting on the ring `k`, provide the `G`-action over the `k`-affine bases of a `k`-module `V`.

---

This could be generalised using some `IsScalarTower` conditions but I'm a bit lazy to sort out exactly what those are and what API we are missing.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14945, opened 2024-07-20
[Merged by Bors] - feat: add `SMul` instances to `Basis`

eric-wieser:
[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Renaming.20.60Basis.2Esmul.60/near/452825113) for the naming clash.

---

- [x] depends on: #15036


Mathlib/LinearAlgebra/Basis.lean
```diff
@@ -316,6 +316,38 @@ theorem coe_map : (b.map f : ι → M') = f ∘ b :=
 
 end Map
 
+section SMul
+variable {G G'}
+variable [Group G] [Group G']
+variable [DistribMulAction G M] [DistribMulAction G' M]
+variable [SMulCommClass G R M] [SMulCommClass G' R M]
```
YaelDillies, line 323: Just for my own understanding, what's the rule/is there a rule preferring `[SMulCommClass G R M]` over `[SMulCommClass R G M]`?

Mathlib/LinearAlgebra/Basis.lean
```diff
@@ -316,6 +316,38 @@ theorem coe_map : (b.map f : ι → M') = f ∘ b :=
...
+
+See also `Basis.unitsSMul` and `Basis.groupSMul`, for the cases when a different action is applied
+to each basis element. -/
+instance : SMul G (Basis ι R M) where
+  smul g b := b.map <| DistribMulAction.toLinearEquiv _ _ g
+
+@[simp]
+theorem smul_apply (g : G) (b : Basis ι R M) (i : ι) : (g • b) i = g • b i := rfl
+
+theorem coe_smul (g : G) (b : Basis ι R M) : ⇑(g • b) = g • ⇑b := rfl
```
YaelDillies, line 335:
```suggestion
@[norm_cast] theorem coe_smul (g : G) (b : Basis ι R M) : ⇑(g • b) = g • ⇑b := rfl
```

Mathlib/LinearAlgebra/Basis.lean
eric-wieser: It's arbitrary, the rule is "pick the one that Lean complained was missing".

Mathlib/LinearAlgebra/Basis.lean
```diff
@@ -316,6 +316,38 @@ theorem coe_map : (b.map f : ι → M') = f ∘ b :=
...
+@[simp]
+theorem smul_eq_map (g : M ≃ₗ[R] M) (b : Basis ι R M) : g • b = b.map g := rfl
+@[simp] theorem repr_smul_apply (g : G) (b : Basis ι R M) :
+    (g • b).repr = (DistribMulAction.toLinearEquiv _ _ g).symm.trans b.repr := rfl
+
+instance : MulAction G (Basis ι R M) :=
+  Function.Injective.mulAction _ DFunLike.coe_injective coe_smul
+
+instance [SMulCommClass G G' M] : SMulCommClass G G' (Basis ι R M) where
+  smul_comm _g _g' _b := DFunLike.ext _ _ fun _ => smul_comm _ _ _
```
YaelDillies, line 347: I think
```
instance [IsScalarTower G G' M] : IsScalarTower G G' (Basis ι R M)
```
also holds

Mathlib/LinearAlgebra/Basis.lean
YaelDillies: 😆 okay but you are (almost) always able to replace `smul_comm` by `← smul_comm` in which case it asks for the other one

Mathlib/LinearAlgebra/Basis.lean
eric-wieser: I skipped it because I can't think of a case where it actually applies, but since you asked for it I've added it.

Mathlib/LinearAlgebra/Basis.lean
eric-wieser: With the extra rule "avoid picking both, and use `.symm` internally to construct the other one".

eric-wieser: I guess aesthetically, I'd prefer to see `SMulCommClass G R M` when attaching a `G` action to something that has `R` and `M` involved in its type.

YaelDillies: Yeah, that's my reasoning as well in #14863

YaelDillies (approved):
This looks good, except that as mentioned in PM it doesn't work for the action of `Rˣ` on `R` when `R` is noncommutative. I think that's fine to not include that for now, but maybe leave a TODO.

maintainer merge

riccardobrasca:
There is a build error, otherwise LGTM, thanks!

bors d+

riccardobrasca:
@eric-wieser can you please fix the error? Thanks!

YaelDillies:
Per Zulip, it will require some renaming

eric-wieser:
bors merge

(@YaelDillies told me the new lemmas look fine)

PR #14946, opened 2024-07-20
[Merged by Bors] - feat(AlgebraicGeometry): Morphism properties local at the source.

erdOne:
---


semorrison:
bors merge

PR #14951, opened 2024-07-20
[Merged by Bors] - feat(Algebra/Homology): relation between pullback/pushout squares and kernels/cokernels in a preadditive category

joelriou:
In the particular case of abelian categories, one of the results in this PR is that a commutative square is a pullback square iff the corresponding sequence `0 ⟶ X₁ ⟶ X₂ ⊞ X₃ ⟶ X₄` is exact.

---


semorrison:
bors merge

PR #14958, opened 2024-07-20
[Merged by Bors] - chore(Data/Matroid/Closure): fix typo

apnelson1:
We add a single newline character so that a two-item list in the docstring for matroid closure appears with the items on separate lines in docgen.

---


semorrison:
bors merge

PR #14960, opened 2024-07-20
[Merged by Bors] - chore: split off `Defs` and `Induction` from `Nat/Factorization/Basic`

rwst:
`Nat/Factorization/Basic` has 900 lines and 5 imports. This PR reduces size and number of imported files in several downstream files by importing the minimal newly created `Nat/Factorization/Defs` or `Nat/Factorization/Induction`

---

### DIFF
--- a/Mathlib/Data/Nat/Factorization/Defs.lean
+++ b/Mathlib/Data/Nat/Factorization/Defs.lean
@@ -4,10 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Stuart Presnell
 -/
 import Mathlib.Data.Finsupp.Multiset
-import Mathlib.Data.Nat.GCD.BigOperators
-import Mathlib.Data.Nat.PrimeFin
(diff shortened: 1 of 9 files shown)


rwst:
!bench

Mathlib/Data/Nat/Factorization/Defs.lean
```diff
@@ -0,0 +1,288 @@
+/-
+Copyright (c) 2021 Stuart Presnell. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Stuart Presnell
+-/
+import Mathlib.Data.Finsupp.Multiset
+import Mathlib.NumberTheory.Padics.PadicVal
```
semorrison, line 7: It would be nice if we could further avoid this for the Defs file, but this is already a great improvement so

bors d+

rwst: I'm not sure if there is a file that would benefit from it. I'll have a look for a possible second iteration.

PR #14961, opened 2024-07-20
[Merged by Bors] - feat(Data/Set/Sigma): univ sigma preimage eq self

apnelson1:
This PR adds a single simp lemma stating that for  `α : ι → Type*` and `(s : Set (Σ i, α i))`, we have `univ.sigma (fun i ↦ Sigma.mk i ⁻¹' s) = s`. 

---

### DIFF
--- a/Mathlib/Data/Set/Sigma.lean
+++ b/Mathlib/Data/Set/Sigma.lean
@@ -44,7 +44,7 @@ theorem image_sigmaMk_preimage_sigmaMap {β : ι' → Type*} {f : ι → ι'} (h
   rcases hxy with ⟨rfl, hxy⟩; rw [heq_iff_eq] at hxy; subst y
   exact ⟨x, hys, rfl⟩
 
-@[simp] theorem iUnion_preimage_image_sigmaMk_eq  {f : (i : ι) → Set (α i)} :
+@[simp] theorem iUnion_preimage_image_sigmaMk_eq {f : (i : ι) → Set (α i)} :
     ⋃ i, Sigma.mk j ⁻¹' (Sigma.mk i '' (f i)) = f j := by
   aesop
 
--- a/Mathlib/Data/Set/Sigma.lean
+++ b/Mathlib/Data/Set/Sigma.lean
@@ -44,10 +44,6 @@ theorem image_sigmaMk_preimage_sigmaMap {β : ι' → Type*} {f : ι → ι'} (h
   rcases hxy with ⟨rfl, hxy⟩; rw [heq_iff_eq] at hxy; subst y
   exact ⟨x, hys, rfl⟩
 
-@[simp] theorem iUnion_preimage_image_sigmaMk_eq {f : (i : ι) → Set (α i)} :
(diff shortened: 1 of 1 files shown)


urkud:
Why do you need this lemma? Can you use `Set.sigma Set.univ f` instead of the union? BTW, could you please add a lemma saying that `Set.sigma` is equal to the indexed union of images?

apnelson1:
> Why do you need this lemma? Can you use `Set.sigma Set.univ f` instead of the union? BTW, could you please add a lemma saying that `Set.sigma` is equal to the indexed union of images?

ah, I hadn't realized the uses of `Set.sigma` for this application. I'll close the PR - thanks!

apnelson1:
> > Why do you need this lemma? Can you use `Set.sigma Set.univ f` instead of the union? BTW, could you please add a lemma saying that `Set.sigma` is equal to the indexed union of images?
> 
> ah, I hadn't realized the uses of `Set.sigma` for this application. I'll close the PR - thanks!

Scratch that - I've reopened! It turns out that the equivalent lemma with better spelling was also missing.

Mathlib/Data/Set/Sigma.lean
```diff
@@ -78,6 +78,10 @@ theorem univ_sigma_univ : (@univ ι).sigma (fun _ ↦ @univ (α i)) = univ := ex
 theorem sigma_univ : s.sigma (fun _ ↦ univ : ∀ i, Set (α i)) = Sigma.fst ⁻¹' s :=
   ext fun _ ↦ and_true_iff _
 
+@[simp] theorem univ_sigma_preimage (s : Set (Σ i, α i)) :
+    (@univ ι).sigma (fun i ↦ Sigma.mk i ⁻¹' s) = s :=
```
eric-wieser, line 82: a little more readable as
```suggestion
    (univ : Set ι).sigma (fun i ↦ Sigma.mk i ⁻¹' s) = s :=
```

Mathlib/Data/Set/Sigma.lean
```diff
@@ -78,6 +78,10 @@ theorem univ_sigma_univ : (@univ ι).sigma (fun _ ↦ @univ (α i)) = univ := ex
 theorem sigma_univ : s.sigma (fun _ ↦ univ : ∀ i, Set (α i)) = Sigma.fst ⁻¹' s :=
   ext fun _ ↦ and_true_iff _
 
+@[simp] theorem univ_sigma_preimage (s : Set (Σ i, α i)) :
```
eric-wieser, line 81:
```suggestion
@[simp] theorem univ_sigma_preimage_mk (s : Set (Σ i, α i)) :
```

eric-wieser (approved):
bors d+

Thanks!

YaelDillies:
@apnelson1 did you trigger the merge? What happened here?

apnelson1:
I replied bors r+ before noticing the actual comments. I'll quickly PR to address them.

YaelDillies:
Ah okay. Over on the reviewers chat we were a bit concerned that the PR had merged itself!

apnelson1:
#15057

PR #14962, opened 2024-07-20
[Merged by Bors] - chore: remove empty marker files

grunweg:
A few files were only created to mark the corresponding mathlib3 files as ported. They can safely be removed now.

---


j-loreaux:
Agreed, thanks!

bors merge

PR #14966, opened 2024-07-21
[Merged by Bors] - feat(Algebra/Polynomial/Module/Basic): add lemmas about `equivPolynomial` and `equivPolynomialSelf`

Command-Master:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14965, opened 2024-07-21
[Merged by Bors] - feat(RingTheory/Derivation/Basic): add `LinearEquiv.compDer` equivalents for `LinearMap.compDer` lemmas

Command-Master:
Add `LinearEquiv.compDer` equivalents for `LinearMap.compDer` lemmas.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14967, opened 2024-07-21
[Merged by Bors] - feat(RingTheory/Derivation/MapCoeffs): define `mapCoeffs` on a differential ring

Command-Master:
---

- [x] depends on: #14966
- [x] depends on: #14965 
- [x] depends on: #14699 
- [x] depends on: #14702 


Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -99,3 +101,87 @@ theorem apply_eval_eq (x : A) (p : A[X]) :
...
+variable [Algebra.IsSeparable A R]
+
+lemma algHom_deriv' (f : R →ₐ[A] R') (hf : Function.Injective f) (x : R) :
+    f (x′) = (f x)′ := algHom_deriv f hf x (Algebra.IsSeparable.isSeparable' x)
+
+omit [Nontrivial R] in
+lemma algEquiv_deriv' (f : R ≃ₐ[A] R') (x : R) :
+    f (x′) = (f x)′ :=
+  haveI := f.nontrivial
+  algHom_deriv' f.toAlgHom f.injective x
```
Command-Master, line 185: I'm uncertain about having these results here, perhaps they belong in `FieldTheory`

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -99,3 +101,117 @@ theorem apply_eval_eq (x : A) (p : A[X]) :
...
+variable {A : Type*} [CommRing A] [Differential A]
+
+/--
+A specialization of `Derivation.mapCoeffs` for the case of a differential ring.
+-/
+def mapCoeffs : Derivation ℤ A[X] A[X] :=
+  PolynomialModule.equivPolynomialSelf.compDer Differential.deriv.mapCoeffs
+
+@[simp]
+lemma mapCoeffs_apply (p : A[X]) (i) :
```
jcommelin, line 116:
```suggestion
lemma coeff_mapCoeffs (p : A[X]) (i) :
```

Mathlib/RingTheory/Derivation/MapCoeffs.lean
jcommelin: I think it's probably fine here.

Mathlib/RingTheory/Derivation/MapCoeffs.lean
```diff
@@ -99,3 +101,117 @@ theorem apply_eval_eq (x : A) (p : A[X]) :
...
+
+@[simp]
+lemma implicitDifferential_C (v : A[X]) (b : A) :
+    letI := implicitDifferential v
+    (C b)′ = C b′ := implicitDeriv_C v b
+
+@[simp]
+lemma implicitDifferential_X (v : A[X]) :
+    letI := implicitDifferential v
+    X′ = v := implicitDeriv_X v
```
jcommelin, line 177: I suggest leaving this block out of this PR, until you need them for something. The `letI` in the statements suggests to me that these will be somewhat tricky to use, and we might converge onto a more usable design late.

Command-Master: Alright, I'll add this in a later PR

jcommelin (commented):
Thanks :tada:

bors merge

PR #14970, opened 2024-07-21
[Merged by Bors] - refactor(Tactic/ToAdditive): avoid FindM, got removed in lean4

nomeata:
this code was using an internal data type from Lean4 that was removed in https://github.com/leanprover/lean4/pull/4795 so let's inline the three lines that we need here.

---

### DIFF
--- a/Mathlib/Tactic/ToAdditive/Frontend.lean
+++ b/Mathlib/Tactic/ToAdditive/Frontend.lean
@@ -474,9 +474,8 @@ structure Config : Type where
   existing : Option Bool := none
   deriving Repr
 
-open Lean.Expr.FindImpl in
 /-- Implementation function for `additiveTest`.
-  We cache previous applications of the function, using the same method that `Expr.find?` uses,
+  We cache previous applications of the function, using an expression cache using ptr equality
   to avoid visiting the same subexpression many times. Note that we only need to cache the
   expressions without taking the value of `inApp` into account, since `inApp` only matters when
   the expression is a constant. However, for this reason we have to make sure that we never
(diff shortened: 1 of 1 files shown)


Mathlib/Tactic/ToAdditive/Frontend.lean
```diff
@@ -486,13 +485,15 @@ open Lean.Expr.FindImpl in
   and we're not remembering the cache between these calls. -/
 unsafe def additiveTestUnsafe (findTranslation? : Name → Option Name)
   (ignore : Name → Option (List ℕ)) (e : Expr) : Option Name :=
-  let rec visit (e : Expr) (inApp := false) : OptionT FindM Name := do
+  let rec visit (e : Expr) (inApp := false) : OptionT (StateM (PtrSet Expr)) Name := do
```
grunweg, line 488: I'm just have an extremely local understanding of the code, so take this with a grain of salt: this was `StateT` in Lean core; was this change intentional?

grunweg (commented):
One question; the rest *looks* good to me. (But note that I'm just locally verifying things; I have no global understanding whatsoever of this code.)

grunweg:
Also, let me add: thanks for taking care of mathlib this way!

Mathlib/Tactic/ToAdditive/Frontend.lean
nomeata: Thanks for asking. It was `StateT (PtrSet Expr) Id`, which is the same as `StateM (PtrSet Expr)`

grunweg: :bulb: Makes sense, thanks!

grunweg (commented):
According to my (necessarily shallow) review, this looks good.
maintainer merge

semorrison:
bors merge

PR #14972, opened 2024-07-21
[Merged by Bors] - feat: the product of boundaryless manifolds is boundaryless

grunweg:
To facilitate this change, performs some preliminary clean-up to this file.
- Rename `boundary_eq_complement_interior` to `compl_interior` and flip its direction. (Happy to add a deprecation warning if desired.)
- Add the analogous lemma `ModelWithCorners.compl_boundary`
- Add  `Boundaryless.iff_boundary_eq_empty`: `M` is boundaryless iff its boundary is empty
- Trivially generalise the lemma about the boundary of product manifolds to `BoundarylessManifold` (not just the model being boundaryless).
- Move these lemmas into the `ModelWithCorners` namespace, where they arguably belong.

Extracted from my in-progress bordism theory branch.

---

I'm sorry that this PR is becoming such a mess: I'm happy to split it into more pieces. I'm not doing so automatically as reviewing several stacked PRs also imposes some overhead.


Mathlib/Geometry/Manifold/InteriorBoundary.lean
```diff
@@ -21,10 +21,12 @@ Define the interior and boundary of a manifold.
 - `ModelWithCorners.univ_eq_interior_union_boundary`: `M` is the union of its interior and boundary
 - `ModelWithCorners.interior_boundary_disjoint`: interior and boundary of `M` are disjoint
 - `BoundarylessManifold.isInteriorPoint`: if `M` is boundaryless, every point is an interior point
-
+- `ModelWithCorners.Boundaryless.{boundary_eq_empty, of_boundary_eq_empty}`:
```
semorrison, line 24: I'd prefer not to use `{` like this. Part of my brain immediately said "that is a really weird name for a universe"!

grunweg: Sure; fixed. (My mental model is shell expansion, akin to `git add `Geometry/Manifold/{Foo,Bar}.lean`.)

urkud: `doc-gen` linkifier doesn't work that way.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
```diff
@@ -123,36 +125,42 @@ instance : BoundarylessManifold I M where
...
 /-- Boundaryless manifolds have empty boundary. -/
-lemma Boundaryless.boundary_eq_empty : I.boundary M = ∅ := by
+lemma Boundaryless.boundary_eq_empty [BoundarylessManifold I M] : I.boundary M = ∅ := by
   rw [I.boundary_eq_complement_interior, I.interior_eq_univ, compl_empty_iff]
 
 instance [BoundarylessManifold I M] : IsEmpty (I.boundary M) :=
   isEmpty_coe_sort.mpr (Boundaryless.boundary_eq_empty I)
 
+/-- Manifolds with empty boundary are boundaryless. -/
+lemma Boundaryless.of_boundary_eq_empty (h : I.boundary M = ∅) : BoundarylessManifold I M where
```
urkud, line 154: This should be an `Iff` lemma.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
```diff
@@ -123,36 +125,42 @@ instance : BoundarylessManifold I M where
...
+  isInteriorPoint' x := by
+    show x ∈ I.interior M
+    rw [boundary_eq_complement_interior, compl_empty_iff] at h
```
urkud, line 157: I think that `boundary_eq_complement_interior` should be called `ModelWithCorners.compl_interior`, with swapped LHS and RHS.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
```diff
@@ -162,43 +170,53 @@ variable {I}
...
-  rw [ModelWithCorners.boundary_prod, ModelWithCorners.Boundaryless.boundary_eq_empty J]
+  rw [boundary_prod, Boundaryless.boundary_eq_empty J]
   have : Set.prod (univ : Set M) (∅ : Set N) = ∅ := Set.prod_empty
   rw [this, empty_union]
 
+/-- The product of two boundaryless manifolds is boundaryless. -/
+instance BoundarylessManifold.prod [BoundarylessManifold I M] [BoundarylessManifold J N] :
+    BoundarylessManifold (I.prod J) (M × N) := by
+  apply Boundaryless.of_boundary_eq_empty
+  rw [boundary_of_boundaryless_left, Boundaryless.boundary_eq_empty]
```
urkud, line 216: Does `simp [boundary_prod, Boundaryless.boundary_eq_empty]` work?

grunweg: It works to replace lines 2+ by the following
```
  simp [boundary_prod, Boundaryless.boundary_eq_empty]
  constructor
  · exact Set.prod_empty
  · exact Set.empty_prod
```
However, `simp` by itself does not apply these lemmas (even though they are marked simp). Supplying them explicitly doesn't work either. If you know a trick to get simp to use them, I'm all ears.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
grunweg: Done

Mathlib/Geometry/Manifold/InteriorBoundary.lean
grunweg: Done: I left the other direction, since applying it requires specifying I.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
urkud: Probably, this is the `Prod` vs `ModelProd` issue.

Mathlib/Geometry/Manifold/InteriorBoundary.lean
```diff
@@ -162,43 +179,53 @@ variable {I}
...
-  rw [ModelWithCorners.boundary_prod, ModelWithCorners.Boundaryless.boundary_eq_empty J]
+  rw [boundary_prod, Boundaryless.boundary_eq_empty J]
   have : Set.prod (univ : Set M) (∅ : Set N) = ∅ := Set.prod_empty
   rw [this, empty_union]
 
+/-- The product of two boundaryless manifolds is boundaryless. -/
+instance BoundarylessManifold.prod [BoundarylessManifold I M] [BoundarylessManifold J N] :
+    BoundarylessManifold (I.prod J) (M × N) := by
+  apply Boundaryless.of_boundary_eq_empty
+  simp [boundary_prod, Boundaryless.boundary_eq_empty]
```
urkud, line 225: Please avoid a non-terminal `simp`. Either restore the old proof or use `simp only`. Otherwise LGTM. Thanks!
bors d+

Mathlib/Geometry/Manifold/InteriorBoundary.lean
grunweg: Makes sense. I've left a comment.

grunweg:
Thanks for the review!
bors merge

PR #14974, opened 2024-07-21
[Merged by Bors] - feat: the empty set is a smooth manifold

grunweg:
- Prove that an empty type is a charted space
- An empty type is a smooth manifold over any pair (E, H)
- The empty manifold is boundaryless

Split out from my branch on bordism theory.

---


semorrison:
bors merge

PR #14975, opened 2024-07-21
feat: define n-manifolds and closed manifolds

grunweg:
Define structures `NManifold` and `ClosedManifold` for n-dimensional resp. closed smooth manifolds (i.e., compact without boundary). These are useful in their own right, and needed to state the basic definitions in bordism theory.

To show these definitions are workable, we state some standard examples:
- Euclidean space is an n-manifold
- the n-sphere is a closed n-manifold
- the two-torus is a closed 2-manifold

Defining n-manifolds posed a design question, whether to have the n-dimensionality as a typeclass assumption or a proof-valued field. The former could avoid unnecessary boilerplate, say when the dimension is ambient and always fixed. However, encoding the dimension in a typeclass is inflexible for dimension-changing constructions. Three examples:
- The product of an n- and an m-manifold is an n+m-manifold: I would prefer that a 2+3 and a 4+1-manifold are the same as a 5-manifold; the typeclass system and addition don't mix well.
- Bordism theory requires such equalities: a cobordism between two 4-manifolds is a certain 5-manifold; the general definition would state this as a 4+1-manifold, which needs to be unified with 5.
- Surgery between manifolds involves k-handles, the product of a k- and an n-k-manifold (for all k): this uses quite extensively that n=k+(n-k) for all k.

For these reasons, I have decided for the latter option.

I am open to also defining a class IsNManifold and adding an instance SmoothNManifold -> IsNManifold.

---

- [ ] depends on: #14972
- [ ] depends on: #14974

Bikeshedding welcome: what is the best order of arguments for n-manifolds and closed n-manifolds?
Right now, I've chosen `NManifold n M I`, but could also be convinced that `NManifold M I n` or some other permutation is better.

### DIFF
--- a/Mathlib/Geometry/Manifold/ClosedNManifold.lean
+++ b/Mathlib/Geometry/Manifold/ClosedNManifold.lean
@@ -1,0 +1,119 @@
+/-
+Copyright (c) 2024 Michael Rothgang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Michael Rothgang
+-/
+import Mathlib.Geometry.Manifold.Instances.Real
+import Mathlib.Geometry.Manifold.Instances.Sphere
+import Mathlib.Geometry.Manifold.InteriorBoundary
+
+/-!
+# Closed `n`-dimensional manifolds
+
+We define closed `n`-dimensional manifolds and show a few basic properties.
+
+## Main definitions
+TODO add parameters!
+* `ClosedManifold M I`: a topological manifold `M` is closed if it is compact and boundaryless
+* `NManifold n M I`: an n-manifold is a smooth `n`-dimensional manifold `M`.
+* `ClosedNManifold n M I`: a closed n-manifold `M` is both closed and an `n`-manifold
+
+## Main results
+* `ClosedManifold.prod`: the product of closed manifolds is closed
+* `NManifold.prod`: the product of an `n`-manifold and an `m`-manifold is an `n+m`-manifold
+* `ClosedNManifold.prod`: the product of a closed `n`-manifold and a closed `m`-manifold is a closed
+`n+m`-manifold
+* We prove a few basic examples
+  - the empty set is closed and an `n`-manifold, for every `n` (TODO insert!)
+  - Euclidean n-space is an n-manifold
+  - the standard n-sphere is a closed n-manifold
+  - the standard two-torus `S¹ × S¹` is a closed 2-manifold
+
+TODO: wait for the product results to be merged! investigate why the spheres fail!
+
+-/
+
+open scoped Manifold
+open Metric (sphere)
+open FiniteDimensional Set
+
+noncomputable section
+
+variable (n : ℕ) {𝕜 : Type*} [NontriviallyNormedField 𝕜]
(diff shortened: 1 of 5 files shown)


Mathlib/Geometry/Manifold/ClosedNManifold.lean
```diff
@@ -0,0 +1,125 @@
...
+noncomputable section
+
+variable (n : ℕ) {𝕜 : Type*} [NontriviallyNormedField 𝕜]
+  -- declare a smooth manifold `M` over the pair `(E, H)`.
+  {E : Type*} [NormedAddCommGroup E] [NormedSpace 𝕜 E] {H : Type*} [TopologicalSpace H]
+  (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
+  (I : ModelWithCorners 𝕜 E H) [SmoothManifoldWithCorners I M]
+
+/-- A topological manifold is called **closed** iff it is compact without boundary. -/
+structure ClosedManifold [CompactSpace M] [BoundarylessManifold I M]
```
sgouezel, line 47: I don't see what the use case for this class would be, instead of just assuming `[CompactSpace M] [BoundaryLessManifold I M]`. Can you explain a little bit more why you need it?

Mathlib/Geometry/Manifold/ClosedNManifold.lean
```diff
@@ -0,0 +1,125 @@
...
+variable {E' : Type*} [NormedAddCommGroup E'] [NormedSpace 𝕜 E']
+  {H' : Type*} [TopologicalSpace H'] (N : Type*) [TopologicalSpace N] [ChartedSpace H' N]
+  (J : ModelWithCorners 𝕜 E' H') [SmoothManifoldWithCorners J N]
+
+instance ClosedManifold.prod [CompactSpace M] [BoundarylessManifold I M]
+    [CompactSpace N] [BoundarylessManifold J N] :
+  ClosedManifold (M × N) (I.prod J) where
+
+/-- An **n-manifold** is a smooth `n`-dimensional manifold. -/
+structure NManifold [NormedAddCommGroup E]  [NormedSpace 𝕜 E] [FiniteDimensional 𝕜 E]
```
sgouezel, line 58: Again, I don't really see this as a good design. Instead, you should probably use an assumption `[Fact (finrank 𝕜 E = n)]` to say that the manifold is `n`-dimensional.

Mathlib/Geometry/Manifold/ClosedNManifold.lean
alreadydone: Yes, maybe you mean something like
```
class ClosedManifold extends CompactSpace M, BoundarylessManifold I M
```
?
Notice that in [#13887](https://github.com/leanprover-community/mathlib4/pull/13887/files) I just use `ChartedSpace ℝⁿ M` to say M is a topological n-manifold (since ℝⁿ is a boundaryless model, M must be boundaryless), and `SmoothManifoldWithCorners (𝓡 n) M` to say it's moreover smooth.

Also, in my opinion if you want to bake in smoothness then the names should contain `Smooth`.

grunweg:
Thanks for the comments! I looked again and realised I don't need this class - which means my code can be simplified. That's also fine - let me close this PR instead.

PR #14976, opened 2024-07-21
feat(Combinatorics/SimpleGraph/Subgraph): Definition of symmDiff and simp lemmas

pimotte:
Introduces the symmetrics difference on subgraphs. 
In preparation for Tutte's theorem. 

---


YaelDillies (commented):
I'm confused, isn't the symmetric difference already accessible through the `∆` (`\inc`) notation?

pimotte:
My bad, I totally overlooked that 🙈

PR #14982, opened 2024-07-21
[Merged by Bors] - fix: reformulate the message when there are too many files to display

adomani:
The new message should make it clear that the import changes are not displayed because they are too many, not that the PR should be rejected because of this!

Reported in #14471.

---


j-loreaux:
Thanks!

bors merge

PR #14985, opened 2024-07-21
[Merged by Bors] - feat(Pullback/CommSq): add more pasting lemmas for `IsPullback`

callesonne:
This PR adds two (variants) of pasting lemmas in the `IsPullback` framework. These are variants where the second square has a morphism that is induced from the universal property of the first.

This work was inspired by the AIM workshop "Formalizing algebraic geometry" in June 2024.

---

These lemmas are used in PR #14208

### DIFF
--- a/Mathlib/CategoryTheory/Limits/Shapes/Pullback/CommSq.lean
+++ b/Mathlib/CategoryTheory/Limits/Shapes/Pullback/CommSq.lean
@@ -274,22 +274,22 @@ noncomputable def isoIsPullback (h : IsPullback fst snd f g) (h' : IsPullback fs
     P ≅ P' :=
   IsLimit.conePointUniqueUpToIso h.isLimit h'.isLimit
 
-@[simp]
+@[reassoc (attr := simp)]
 theorem isoIsPullback_hom_fst (h : IsPullback fst snd f g) (h' : IsPullback fst' snd' f g) :
     (h.isoIsPullback h').hom ≫ fst' = fst :=
   IsLimit.conePointUniqueUpToIso_hom_comp h.isLimit h'.isLimit WalkingCospan.left
 
-@[simp]
+@[reassoc (attr := simp)]
 theorem isoIsPullback_hom_snd (h : IsPullback fst snd f g) (h' : IsPullback fst' snd' f g) :
     (h.isoIsPullback h').hom ≫ snd' = snd :=
   IsLimit.conePointUniqueUpToIso_hom_comp h.isLimit h'.isLimit WalkingCospan.right
 
-@[simp]
+@[reassoc (attr := simp)]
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Limits/Shapes/Pullback/CommSq.lean
```diff
@@ -201,6 +201,21 @@ theorem cone_snd (h : IsPullback fst snd f g) : h.cone.snd = snd :=
...
+@[reassoc (attr := simp)]
+lemma lift_fst (hP : IsPullback fst snd f g) {W : C} (h : W ⟶ X) (k : W ⟶ Y)
+    (w : h ≫ f = k ≫ g) : hP.lift h k w ≫ fst = h :=
+  PullbackCone.IsLimit.lift_fst hP.isLimit h k w
+
+@[reassoc (attr := simp)]
+lemma lift_snd (hP : IsPullback fst snd f g) {W : C} (h : W ⟶ X) (k : W ⟶ Y)
+    (w : h ≫ f = k ≫ g) : hP.lift h k w ≫ snd = k :=
+  PullbackCone.IsLimit.lift_snd hP.isLimit h k w
+
```
joelriou, line 218: Here, you may also add a `hom_ext` lemma: a morphism to `P` is determined by the two compositions. (Same for pushouts.)

Mathlib/CategoryTheory/Limits/Shapes/Pullback/CommSq.lean
```diff
@@ -249,6 +264,38 @@ theorem of_hasBinaryProduct [HasBinaryProduct X Y] [HasZeroObject C] [HasZeroMor
...
+variable {P': C} {fst' : P' ⟶ X} {snd' : P' ⟶ Y}
+
+/-- Any object at the top left of a pullback square is isomorphic to the object at the top left
+of any other pullback square with the same cospan. -/
+noncomputable def isoIsPullback (h : IsPullback fst snd f g) (h' : IsPullback fst' snd' f g) :
+    P ≅ P' :=
+  IsLimit.conePointUniqueUpToIso h.isLimit h'.isLimit
+
+@[simp]
```
joelriou, line 277: This also applies to other lemmas (basically anytime the LHS is a composition):
```suggestion
@[reassoc (attr := simp)]
```

callesonne: Thanks, forgot about that!

Mathlib/CategoryTheory/Limits/Shapes/Pullback/CommSq.lean
callesonne: Fixed :)

joelriou:
Thanks!

bors merge

PR #14986, opened 2024-07-21
[Merged by Bors] - feat(LinearAlgebra/QuadraticForm/Basic): weaken invertibility hypothesis on 2

MichaelStollBayreuth:
The aim of this is to eventually enable base-change of quadratic maps from a (not necessarily (torsion-)free) abelian group to an abelian group on which 2 is invertible (e.g., the real numbers). The current set-up requires 2 to be invertible in the base ring (which would be the integers in the situation above). This PR weakens the assumption for the definition of `QuadraticMap.associatedHom` and `QuadraticMap.associated` from `[Invertible (2 : R)]` to `[Invertible (2 : Module.End R N)]`, where `N` is the target module of he quadratic map.

See [this discussion](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Quadratic.20forms.20with.20values.20in.20a.20larger.20ring/near/450582700) on Zulip.

On the way, we fix some mistakes in docstrings.

---


MichaelStollBayreuth:
!bench

MichaelStollBayreuth:
<b>General information:</b>
```text
build:                                                 +11.164 * 10⁹ (+0.00917%)
```


<b>4 files got slower by at least 10⁹ instructions:</b>
```text
Mathlib.LinearAlgebra.QuadraticForm.Basic:             +9.8542 * 10⁹ (+5.14%)
Mathlib.LinearAlgebra.QuadraticForm.TensorProduct:     +2.1026 * 10⁹ (+1.88%)
Mathlib.LinearAlgebra.QuadraticForm.TensorProduct.Isometries:
                                                       +1.1630 * 10⁹ (+1.30%)
Mathlib.LinearAlgebra.QuadraticForm.IsometryEquiv:     +1.1469 * 10⁹ (+6.65%)
```


<b>2 files got faster by at least 10⁹ instructions:</b>
```text
Mathlib.LinearAlgebra.QuadraticForm.Prod:              -2.0036 * 10⁹ (-4.77%)
Mathlib.LinearAlgebra.QuadraticForm.Dual:              -1.8970 * 10⁹ (-6.24%)
```


urkud:
`Invertible` is a data carrying typeclass, so

- please don't use tactic and `rw` to construct the data part, use it only for the proof;
- I understand that you're going to add another instance for `Invertible (2 : Module.End R M)` that works if `R = Int` and `M = Real`. How are you going to avoid non-defeq diamonds in case `R = M = Real`?

MichaelStollBayreuth:
I changed the definition of the `Invertible` instance.
Regarding the second point above, I'm not sure. For my application, I could just introduce the concrete instance where `R = ℤ` and `M = ℝ`, but of course, a more general setting would be desirable. Do you have some proposal how one could achieve what I want without running into that problem?
(And sorry for the slow reaction. I was on vacation, at a conference and otherwise busy...)

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +824,104 @@ open LinearMap (BilinMap)
...
-variable [Module S N] [IsScalarTower S R N]
-variable [Invertible (2 : R)] {B₁ : BilinMap R M R}
+variable [CommRing R] [AddCommGroup M] [Module R M]
+
+/-- If `2` is invertible in `R`, then it is also invertible in `End_R M`. -/
+instance [Invertible (2 : R)] : Invertible (2 : Module.End R M) :=
+  map_ofNat (algebraMap R (Module.End R M)) 2 ▸ Invertible.map (algebraMap R (Module.End R M)) 2
+
+/-- If `2` is invertible in `R`, then applying the inverse of `2` in `End_R M` to an element
+of `M` is the same as multplying by the inverse of `2` in `R`. -/
```
loefflerd, line 834: typo

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +824,104 @@ open LinearMap (BilinMap)
 
 section AssociatedHom
 
-variable [CommRing R] [AddCommGroup M] [Module R M] [AddCommGroup N] [Module R N]
-variable (S) [CommSemiring S] [Algebra S R]
-variable [Module S N] [IsScalarTower S R N]
-variable [Invertible (2 : R)] {B₁ : BilinMap R M R}
+variable [CommRing R] [AddCommGroup M] [Module R M]
+
+/-- If `2` is invertible in `R`, then it is also invertible in `End_R M`. -/
```
loefflerd, line 829: Let's not mix LaTeX and unicode notations. `End R M` is the mathlib notation.

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +824,104 @@ open LinearMap (BilinMap)
...
 /-- `associatedHom` is the map that sends a quadratic map on a module `M` over `R` to its
 associated symmetric bilinear map.  As provided here, this has the structure of an `S`-linear map
-where `S` is a commutative subring of `R`.
+where `S` is a commutative ring and `R` is an `S`-algebra.
 
 Over a commutative ring, use `QuadraticMap.associated`, which gives an `R`-linear map.  Over a
 general ring with no nontrivial distinguished commutative subring, use `QuadraticMap.associated'`,
 which gives an additive homomorphism (or more precisely a `ℤ`-linear map.) -/
 def associatedHom : QuadraticMap R M N →ₗ[S] (BilinMap R M N) :=
   -- TODO: this `center` stuff is vertigial from an incorrect non-commutative version, but we leave
```
loefflerd, line 855: vestigial (nothing to do with vertices!)

loefflerd (commented):
This looks very reasonable to me. I'm no expert on this corner of the library though; I left some comments, but if they are based on misunderstandings, then don't hesitate to tell me I'm wrong.

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
jcommelin:
```suggestion
of `M` is the same as multiplying by the inverse of `2` in `R`. -/
```

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +824,104 @@ open LinearMap (BilinMap)
...
   -- it behind to make a future refactor to a *correct* non-commutative version easier in future.
-  (⟨⅟2, Set.invOf_mem_center (Set.ofNat_mem_center _ _)⟩ : Submonoid.center R) •
-    { toFun := polarBilin
-      map_add' := fun _x _y => LinearMap.ext₂ <| polar_add _ _
-      map_smul' := fun _c _x => LinearMap.ext₂ <| polar_smul _ _ }
+  -- This was the code prior to weakining the invertibility hypothesis on 2:
+  --   (⟨⅟2, Set.invOf_mem_center (Set.ofNat_mem_center _ _)⟩ : Submonoid.center R) •
+  --    { toFun := polarBilin
+  --      map_add' := fun _x _y => LinearMap.ext₂ <| polar_add _ _
+  --      map_smul' := fun _c _x => LinearMap.ext₂ <| polar_smul _ _ }
```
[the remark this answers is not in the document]
MichaelStollBayreuth: This is what the "TODO" is referring to, so I thought it is  better left.

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -655,10 +656,13 @@ end QuadraticMap
...
-basically identical to that of symmetric bilinear maps. The map from quadratic
-maps to bilinear maps giving this identification is called the `QuadraticMap.associated`
-quadratic map.
+If multiplication by 2 is invertible on the target module `N` of
+`QuadraticMap R M N`, then there is a linear bijection `QuadraticMap.associated`
+between quadratic maps `Q` over `R` from `M` to `N` and symmetric bilinear maps
+`B : M →ₗ[R] M →ₗ[R] → N` such that `BilinMap.toQuadraticMap B = Q`
+(see `QuadraticMap.associated_rightInverse`). The associated bilinear map is half
+`Q.polarBilin` (see `QuadraticMap.two_nsmul_associated`); this is where the invertibility condition
+comes from. We spell the condition as `[Invertible (2 : Module.End R N)]`.
```
eric-wieser, line 665: Can you add a comment here explaining why it is interesting to talk about cases where `2 : R` is not invertible but `(2 : Module.End R N)` is?

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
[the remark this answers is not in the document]
eric-wieser: I think we can drop it now, the construction with `center` is only really of interest now for going from `Invertible (2 : R)` to `Invertible (2 : End R M)`

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +824,105 @@ open LinearMap (BilinMap)
...
 
-variable [CommRing R] [AddCommGroup M] [Module R M] [AddCommGroup N] [Module R N]
-variable (S) [CommSemiring S] [Algebra S R]
-variable [Module S N] [IsScalarTower S R N]
-variable [Invertible (2 : R)] {B₁ : BilinMap R M R}
+variable [CommRing R] [AddCommGroup M] [Module R M]
+
+/-- If `2` is invertible in `R`, then it is also invertible in `End R M`. -/
+instance [Invertible (2 : R)] : Invertible (2 : Module.End R M) :=
+  map_ofNat (algebraMap R (Module.End R M)) 2 ▸ Invertible.map (algebraMap R (Module.End R M)) 2
```
eric-wieser, line 831: It's not usually a good idea to construct data with `▸`; can you use `Invertible.copy` or similar instead?

eric-wieser:
> The aim of this is to eventually enable base-change of quadratic maps from a (not necessarily (torsion-)free) abelian group to an abelian group on which 2 is invertible (e.g., the real numbers). The current set-up requires 2 to be invertible in the base ring (which would be the integers in the situation above).

Do we want to deal with an even more general case such as a base-change from `Nat` to `Int`, in which case really the concept we want is "`Q` is halveable", not "all elements in the destination module are halveable.

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
MichaelStollBayreuth: OK; seems to work.

MichaelStollBayreuth:
I guess the weakest possible condition would be "all values of `Q` have a unique half". This looks like it would be fairly clumsy to implement. Maybe let's wait until somebody really needs more generality?

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +827,106 @@ open LinearMap (BilinMap)
...
+  (Invertible.map (algebraMap R (Module.End R M)) 2).copy 2 <|
+    (map_ofNat (algebraMap R (Module.End R M)) 2).symm
+
+/-- If `2` is invertible in `R`, then applying the inverse of `2` in `End R M` to an element
+of `M` is the same as multiplying by the inverse of `2` in `R`. -/
+@[simp]
+lemma half_moduleEnd_apply_eq_half_smul [Invertible (2 : R)] (x : M) :
+    ⅟ (2 : Module.End R M) x = ⅟ (2 : R) • x := by
+  rw [← LinearMap.smul_def, invOf_smul_eq_iff, two_smul, ← add_smul, invOf_two_add_invOf_two,
+    one_smul]
```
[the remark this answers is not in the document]
MichaelStollBayreuth: Is this a benefit of the "better" instance?

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -820,65 +827,105 @@ open LinearMap (BilinMap)
...
-variable [CommRing R] [AddCommGroup M] [Module R M] [AddCommGroup N] [Module R N]
-variable (S) [CommSemiring S] [Algebra S R]
-variable [Module S N] [IsScalarTower S R N]
-variable [Invertible (2 : R)] {B₁ : BilinMap R M R}
+variable [CommRing R] [AddCommGroup M] [Module R M]
+
+/-- If `2` is invertible in `R`, then it is also invertible in `End R M`. -/
+instance [inst : Invertible (2 : R)] : Invertible (2 : Module.End R M) :=
+  (Invertible.map (algebraMap R (Module.End R M)) 2).copy 2 <|
+    (map_ofNat (algebraMap R (Module.End R M)) 2).symm
```
eric-wieser, line 835: This holds more generally as
```lean
variable [Semiring R] [AddCommMonoid M] [Module R M]

instance : SMulCommClass R (↥(Submonoid.center R)) M where
  smul_comm r r' m := by
    simp_rw [Submonoid.smul_def, smul_smul, (Set.mem_center_iff.1 r'.prop).1]

/-- If `2` is invertible in `R`, then it is also invertible in `End R M`. -/
instance [Invertible (2 : R)] : Invertible (2 : Module.End R M) where
  invOf :=  (⟨⅟2, Set.invOf_mem_center (Set.ofNat_mem_center _ _)⟩ : Submonoid.center R) • (1 : Module.End R M)
  invOf_mul_self := by
    ext m
    dsimp [Submonoid.smul_def]
    rw [← ofNat_smul_eq_nsmul R, invOf_smul_smul (2 : R) m]
  mul_invOf_self := by
    ext m
    dsimp [Submonoid.smul_def]
    rw [← ofNat_smul_eq_nsmul R, smul_invOf_smul (2 : R) m]
```
(which also preserves the code I suggested you delete elsewhere)

jcommelin (commented):
This has been open for way too long. If at some point we want to replace `[Invertible x]` by `[IsUnit x]` or `[Fact (IsUnit x)]` we can do such a refactor then. For now, this is an improvement over the status quo.

bors merge

PR #14987, opened 2024-07-21
[Merged by Bors] - feat(FieldTheory/Relrank): Relative rank of subfields and intermediate fields

acmepjz:
We define `Subfield.relrank A B` and `IntermediateField.relrank A B`, which is `[B : A ⊓ B]`, in particular, when `A ≤ B` it is `[B : A]`, the degree of the field extension `B / A`. This is similar to `Subgroup.relindex`.

---

- [x] depends on: #15148
- [x] depends on: #15193
- [x] depends on: #15238

discussion: https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2314987.20Relative.20rank.20of.20intermediate.20fields/near/453033687


tb65536:
Is it certain that we want this to be cardinal valued rather than nat valued?

acmepjz:
> Is it certain that we want this to be cardinal-valued rather than nat-valued? Personally, I feel like the convenience of nat-valued would outweigh the benefit of distinguishing countable vs uncountable degree extensions (my impression is that for infinite degree extensions, you usually care more about transcendence degree than the usual degree as a cardinal).

Personally I'd like to make `relrank` `Cardinal`-valued, following the naming of `rank` and `finrank`. We can also define `relfinrank` which is `Nat`-valued.

jcommelin:
I think I agree with @tb65536 that Nat-valued is much more useful. I also agree with @acmepjz that this creates a bit of an inconsistency naming-wise. But maybe we should fix that be renaming `rank` to `cardinal_rank` (`crank` :rofl:) or something like that, so that `finrank` can be moved to `rank`.
Since `relrank` for subgroups is already Nat-valued, I think it doesn't hurt to do the same here.

Mathlib/FieldTheory/Relrank.lean
```diff
@@ -0,0 +1,532 @@
...
+variable {A B} in
+theorem relfinrank_mul_finrank_top (h : A ≤ B) : relfinrank A B * finrank B E = finrank A E := by
+  simpa using congr(toNat $(relrank_mul_rank_top h))
+
+@[simp]
+theorem relrank_top_left : relrank ⊤ A = 1 := relrank_eq_one_of_le le_top
+
+@[simp]
+theorem relfinrank_top_left : relfinrank ⊤ A = 1 := by
+  simp [relfinrank_eq_toNat_relrank]
```
tb65536, line 114: This can use `relfinrank_eq_one_of_le`

acmepjz: Thanks!

Mathlib/FieldTheory/Relrank.lean
```diff
@@ -0,0 +1,530 @@
...
+This file contains basics about the relative rank of subfields and intermediate fields.
+
+## Main definitions
+
+- `Subfield.relrank A B`, `IntermediateField.relrank A B`:
+  defined to be `[B : A ⊓ B]` as a `Cardinal`.
+  In particular, when `A ≤ B` it is `[B : A]`, the degree of the field extension `B / A`.
+  This is similar to `Subgroup.relindex` but it is `Cardinal` valued.
+
+- `Subfield.relfinrank A B`, `IntermediateField.relfinrank A B`:
```
alreadydone, line 21: Should these be `relRank` and `relFinrank`? I think the group theory one should be `relIndex` too. And then the file name would be RelRank ...

acmepjz: I have no preferences on this. Maybe you can open a poll on zulip...

alreadydone: Let's not worry about this here.

alreadydone (approved):
Just a few minor questions

Mathlib/FieldTheory/Relrank.lean
```diff
@@ -0,0 +1,530 @@
...
+theorem rank_comap {L : Type v} [Field L] (f : L →+* E) :
+    Module.rank (A.comap f) L = relrank A f.fieldRange := by
+  simpa only [lift_id] using A.lift_rank_comap f
+
+theorem finrank_comap (f : L →+* E) : finrank (A.comap f) L = relfinrank A f.fieldRange := by
+  simpa using congr(toNat $(lift_rank_comap A f))
+
+theorem lift_relrank_map_map (f : E →+* L) :
+    lift.{v} (relrank (A.map f) (B.map f)) = lift.{w} (relrank A B) := by
+  rw [← lift_relrank_comap, comap_map]
```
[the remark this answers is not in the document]
acmepjz: I didn't think of it. Actually initially the APIs are copied from that of `Subgroup.relindex` which only has `comap` one. The `map` one is added later, as in the field case, all maps are injective, so there are some more results.

Mathlib/FieldTheory/Relrank.lean
```diff
@@ -0,0 +1,530 @@
...
+    · rintro ⟨h1, a, h2, rfl⟩
+      exact ⟨a, ⟨mem_comap.2 h1, h2⟩, rfl⟩
+  let j' : ↥(A.comap f ⊓ B) ≃+* j.fieldRange :=
+    RingEquiv.ofLeftInverse (Function.leftInverse_invFun j.injective)
+  let f' : B ≃+* (f.comp B.subtype).fieldRange :=
+    RingEquiv.ofLeftInverse (Function.leftInverse_invFun (f.comp B.subtype).injective)
+  have hf : (f.comp B.subtype).fieldRange = B.map f := by
+    rw [← RingHom.map_fieldRange, fieldRange_subtype]
+  exact Algebra.lift_rank_eq_of_equiv_equiv (j'.trans (.subfieldCongr hj))
+    (f'.trans (.subfieldCongr hf)) rfl
```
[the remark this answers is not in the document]
acmepjz: Good observation!

Mathlib/FieldTheory/Relrank.lean
```diff
@@ -0,0 +1,527 @@
...
+theorem relrank_top_right : relrank A ⊤ = Module.rank A E := by
+  rw [relrank_eq_rank_of_le (show A ≤ ⊤ from le_top), extendScalars_top,
+    IntermediateField.topEquiv.toLinearEquiv.rank_eq]
+
+@[simp]
+theorem relfinrank_top_right : relfinrank A ⊤ = finrank A E := by
+  simp [relfinrank_eq_toNat_relrank, finrank]
+
+theorem lift_relrank_comap (f : L →+* E) (B : Subfield L) :
+    lift.{v} (relrank (A.comap f) B) = lift.{w} (relrank A (B.map f)) := by
```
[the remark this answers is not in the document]
acmepjz: Thanks for the alternative proof! I'll check them.

> It doesn't seem that you're following the `relindex` API exactly?

That probably because some of the proof for `relindex` does not work in this case. So I changed the order of them.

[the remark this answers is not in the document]
acmepjz: Done.

alreadydone:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #14988, opened 2024-07-21
[Merged by Bors] - refactor(LinearAlgebra/BilinearForm/TensorProduct): Tensor products of bilinear maps

mans0954:
Extend the results in `Mathlib/LinearAlgebra/BilinearForm/TensorProduct.lean` from Bilinear Forms to Bilinear Maps

Similarly for `LinearAlgebra/QuadraticForm/TensorProduct`.

I had originally intended to extend much more, but was blocked by not being able to extend `LinearMap.IsSymm` from forms to maps.

The existing `BilinMap` names were actually specific to bilinear forms; these have been renamed to make room for the generalizations.

Moves:
- `LinearMap.BilinMap.tensorDistrib` → `LinearMap.BilinForm.tensorDistrib`
- `LinearMap.BilinMap.tensorDistrib_tmul` → `LinearMap.BilinForm.tensorDistrib_tmul`
- `LinearMap.BilinMap.tmul` → `LinearMap.BilinForm.tmul`
- `LinearMap.BilinMap.baseChange` → `LinearMap.BilinForm.baseChange`
- `LinearMap.BilinMap.baseChange_tmul` → `LinearMap.BilinForm.baseChange_tmul`

---

- [x] depends on: #17454


mans0954:
CC: @eric-wieser

Mathlib/LinearAlgebra/BilinearForm/Hom.lean
```diff
@@ -305,6 +306,25 @@ theorem comp_congr (e : M' ≃ₗ[R] M'') (B : BilinForm R M) (l r : M' →ₗ[R
       B.comp (l.comp (e.symm : M'' →ₗ[R] M')) (r.comp (e.symm : M'' →ₗ[R] M')) :=
   rfl
 
+variable {N₁ N₂ : Type*}
+variable [AddCommMonoid N₁] [AddCommMonoid N₂] [Module R N₁] [Module R N₂]
+
+/-- When `N₁` and `N₂` are equivalent, bilinear maps on `M` into `N₁` are equivalent to bilinear
+maps into `N₂`. -/
+@[simps]
+def _root_.LinearMap.BilinMap.congr₂ (e : N₁ ≃ₗ[R] N₂) : BilinMap R M N₁ ≃ₗ[R] BilinMap R M N₂ where
```
eric-wieser, line 315: This looks quite similar to [LinearEquiv.congrRight](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Module/Equiv/Basic.html#LinearEquiv.congrRight); I think we can use that to implement this?

mans0954: Thanks. Done. In a similar vein, we can also implement `LinearMap.BilinForm.congr` as:
```
def congr (e : M ≃ₗ[R] M') : BilinForm R M ≃ₗ[R] BilinForm R M' :=
  LinearEquiv.congrRight (LinearEquiv.congrLeft _ _ e) ≪≫ₗ LinearEquiv.congrLeft _ _ e
```
I've pushed a separate commit with this, but perhaps there are reasons why it was previously done directly? (Performance?)

mans0954:
@eric-wieser did you have any further thoughts on this please?

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -579,6 +579,23 @@ def _root_.LinearMap.compQuadraticMap' [CommSemiring S] [Algebra S R] [Module S
     (f : N →ₗ[S] P) (Q : QuadraticMap R M N) : QuadraticMap S M P :=
   _root_.LinearMap.compQuadraticMap f Q.restrictScalars
 
+
+/-- When `N₁` and `N₂` are equivalent, bilinear maps on `M` into `N₁` are equivalent to bilinear
```
erdOne, line 583:
```suggestion
/-- When `N₁` and `N₂` are equivalent, quadratic maps on `M` into `N₁` are equivalent to quadratic
```

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
```diff
@@ -52,14 +85,15 @@ noncomputable def tensorDistrib :
       (QuadraticMap.associated : QuadraticForm A M₁ →ₗ[A] BilinForm A M₁)
       (QuadraticMap.associated : QuadraticForm R M₂ →ₗ[R] BilinForm R M₂)
   toQ ∘ₗ tmulB ∘ₗ toB
+--
```
erdOne, line 88: What's this?

erdOne (commented):
Looks good modulo minor comments

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
mans0954: Must be an artifact. Removed now, thanks.

Mathlib/LinearAlgebra/QuadraticForm/Basic.lean
```diff
@@ -579,6 +579,23 @@ def _root_.LinearMap.compQuadraticMap' [CommSemiring S] [Algebra S R] [Module S
...
+  invFun Q := e.symm.compQuadraticMap Q
+  left_inv _ := ext fun x => by
+    simp only [LinearMap.compQuadraticMap_apply, LinearEquiv.coe_coe, LinearEquiv.symm_apply_apply]
+  right_inv _ := ext fun x => by
+    simp only [LinearMap.compQuadraticMap_apply, LinearEquiv.coe_coe, LinearEquiv.apply_symm_apply]
+  map_add' _ _ := ext fun x => by
+    simp only [LinearMap.compQuadraticMap_apply, add_apply, map_add, LinearEquiv.coe_coe]
+  map_smul' _ _ := ext fun x => by
+    simp only [LinearMap.compQuadraticMap_apply, smul_apply, LinearMapClass.map_smul,
+      LinearEquiv.coe_coe, RingHom.id_apply]
```
eric-wieser, line 597: Could you split this and the corresponding `BilinForm.congr₂` to their own PR?

It would be good to have the
```

@[simp]
theorem congr_refl : congr (.refl R N) = .refl R (QuadraticMap R M N) := rfl

theorem congr_symm (e : N ≃ₗ[R] P) : congr e.symm = (congr e (M := M)).symm := rfl
```
lemmas for this definition as well, like we do for `BilinForm`.

mans0954: #17454

Mathlib/LinearAlgebra/BilinearForm/TensorProduct.lean
```diff
@@ -34,23 +36,49 @@ open LinearMap (BilinMap BilinForm)
...
+variable [SMulCommClass R A N₁] [IsScalarTower R A N₁]
+variable [Module R M₂] [Module R N₂]
+
+variable (R A) in
+/-- The tensor product of two bilinear maps injects into bilinear maps on tensor products.
+
+Note this is heterobasic; the bilinear map on the left can take values in a module over a
+(commutative) algebra over the ring of the module in which the right bilinear map is valued. -/
+def tensorDistrib' :
+    ((BilinMap A M₁ N₁) ⊗[R] (BilinMap R M₂ N₂)) →ₗ[A] ((BilinMap A (M₁ ⊗[R] M₂) (N₁ ⊗[R] N₂))) :=
```
eric-wieser, line 51:
```suggestion
    (BilinMap A M₁ N₁ ⊗[R] BilinMap R M₂ N₂) →ₗ[A] BilinMap A (M₁ ⊗[R] M₂) (N₁ ⊗[R] N₂) :=
```

Mathlib/LinearAlgebra/BilinearForm/TensorProduct.lean
```diff
@@ -34,23 +36,49 @@ open LinearMap (BilinMap BilinForm)
...
-variable [Module R M₂]
+variable [SMulCommClass R A N₁] [IsScalarTower R A N₁]
+variable [Module R M₂] [Module R N₂]
+
+variable (R A) in
+/-- The tensor product of two bilinear maps injects into bilinear maps on tensor products.
+
+Note this is heterobasic; the bilinear map on the left can take values in a module over a
+(commutative) algebra over the ring of the module in which the right bilinear map is valued. -/
+def tensorDistrib' :
```
eric-wieser, line 50: Instead of priming this one, let's rename the other one to `BilinForm.tensorDistrib`

mans0954: Done.

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
```diff
@@ -16,38 +16,76 @@ import Mathlib.LinearAlgebra.QuadraticForm.Basic
...
+
 variable (R A) in
 /-- The tensor product of two quadratic forms injects into quadratic forms on tensor products.
 
 Note this is heterobasic; the quadratic form on the left can take values in a larger ring than
 the one on the right. -/
 -- `noncomputable` is a performance workaround for mathlib4#7103
+-- Should be
+-- `(AlgebraTensorModule.rid R A A).congrQuadraticMap.toLinearMap ∘ₗ QuadraticMap.tensorDistrib R A`
+-- but then `associated_tmul` breaks
```
eric-wieser, line 81: Is this still true? Does "should" here just mean "could be golfed to", or also "has a better defeq as"?

Mathlib/LinearAlgebra/BilinearForm/TensorProduct.lean
```diff
@@ -20,37 +21,68 @@ import Mathlib.LinearAlgebra.TensorProduct.Tower
...
+
+@[simp]
+theorem tensorDistrib_tmul (B₁ : BilinMap A M₁ N₁) (B₂ : BilinMap R M₂ N₂) (m₁ : M₁) (m₂ : M₂)
+    (m₁' : M₁) (m₂' : M₂) :
+    tensorDistrib R A (B₁ ⊗ₜ B₂) (m₁ ⊗ₜ m₂) (m₁' ⊗ₜ m₂')
+      = B₁ m₁ m₁' ⊗ₜ B₂ m₂ m₂' :=
+  rfl
+
+/-- The tensor product of two bilinear forms, a shorthand for dot notation. -/
+protected abbrev tmul (B₁ : BilinMap A M₁ N₁) (B₂ : BilinMap  R M₂ N₂) :
```
eric-wieser, line 69:
```suggestion
protected abbrev tmul (B₁ : BilinMap A M₁ N₁) (B₂ : BilinMap R M₂ N₂) :
```

eric-wieser (commented):
Can you fill out the "moves" section of the PR template to indicate that `BilinMap.tmul` and friends have moved to `BilinForm.tmul`?

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
mans0954: Yes. I meant "should" mathematically as it's then defined as a special case of the `QuadraticMap` version.

mans0954: @eric-wieser I've pushed a solution, but it's not very pretty as it involves keeping the old definition of `QuadraticForm.tensorDistrib` around as well. Perhaps I should restore the definition I had before and then just add a result to show that it is equal to `(AlgebraTensorModule.rid R A A).congrQuadraticMap.toLinearMap ∘ₗ QuadraticMap.tensorDistrib R A`?

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
```diff
@@ -68,17 +95,45 @@ protected noncomputable abbrev tmul (Q₁ : QuadraticForm A M₁) (Q₂ : Quadra
...
+  let toB := AlgebraTensorModule.map
+      (QuadraticMap.associated : QuadraticForm A M₁ →ₗ[A] BilinForm A M₁)
+      (QuadraticMap.associated : QuadraticForm R M₂ →ₗ[R] BilinForm R M₂)
+  toQ ∘ₗ tmulB ∘ₗ toB
+
+/-- The tensor product of two quadratic forms, a shorthand for dot notation. -/
+-- `noncomputable` is a performance workaround for mathlib4#7103
+protected noncomputable abbrev tmul' (Q₁ : QuadraticForm A M₁) (Q₂ : QuadraticForm R M₂) :
+    QuadraticForm A (M₁ ⊗[R] M₂) :=
+  tensorDistrib' R A (Q₁ ⊗ₜ[R] Q₂)
```
eric-wieser, line 120: Do these exist just for `helper`? I would hope they can be avoided by repairing the proof below.

mans0954: I've got rid of these definitions and `helper` - essentially by hiding them inside the proof of `associated_tmul`. If that's not good enough, I can have another go tomorrow.

eric-wieser: I suspect the right answer here is to prove the `QuadraticMap` version first.

eric-wieser: Sorry, I now realize that is a silly suggestion. I think what you have is ok with a comment explaining the unusual proof.

mans0954: Done, thanks.

Mathlib/LinearAlgebra/QuadraticForm/TensorProduct.lean
```diff
@@ -95,6 +95,13 @@ protected noncomputable abbrev tmul (Q₁ : QuadraticForm A M₁) (Q₂ : Quadra
     QuadraticForm A (M₁ ⊗[R] M₂) :=
   tensorDistrib R A (Q₁ ⊗ₜ[R] Q₂)
 
+/- Previously `QuadraticForm.tensorDistrib` was defined in a similar way to
+`QuadraticMap.tensorDistrib`. We now obtain the definition of `QuadraticForm.tensorDistrib`
+from `QuadraticMap.tensorDistrib` using `A ⊗[R] R ≃ₗ[A] A`. Hypothesis `e1` below shows that the
+new definition is equal to the old, and allows us to reuse the old proof.
+
+TODO: Define `IsSymm` for bilinear maps and generalise this result to Quadratic Maps.
+-/
```
eric-wieser, line 104: Can you put this inside the proof just before `have e1`? After all, this is a comment about the proof, not the theorem

mans0954: Done, thanks.

mans0954: For a proper solution we probably want something like:
```
variable [AddCommMonoid M] [CommSemiring R] [Module R M] [StarRing R] [Star M]

/-- The proposition that a sesquilinear form is conjugate symmetric -/
def LinearMap.IsSymm (B : M →ₛₗ[starRingEnd R] M →ₗ[R] R) : Prop :=
  ∀ x y, star (B x y) = B y x
```

But preferably with `[NonUnitalSemiring R]` (or even `[NonUnitalNonAssocSemiringSemiring R]`) instead of `[CommSemiring R]` in order that this would cover Hilbert C*-modules.

mans0954: I've come up with this: #18406

eric-wieser:
bors merge

Thanks!

PR #14990, opened 2024-07-21
[Merged by Bors] - feat: definition of linear topologies

AntoineChambert-Loir:
A topology on a module is linear if it is invariant by translation and if there 
is a basis of neighborhoods consisting of submodules.

We are most interested in the case of rings: a topology on a ring is linear 
if it is linear for both the left- and right-module structures on R over itself.
This is equivalent to being invariant by translation and admitting a 
basis of neighborhoods consisting of two-sided ideals.

This will be used in a subsequent PR to evaluate multivariate power series.
We will also show that the natural topology on `MvPowerSeries S R` is a linear topology when `S` has a linear topology (e.g the discrete topology).

Co-authored-by: @mariainesdff and @ADedecker 

---


ADedecker:
(Note: we talked about this PR IRL, so I'm marking it as awaiting-author)

AntoineChambert-Loir:
Well, I believe I am not ready for the moment to implement this big change, and I would like to go forward with the stuff that is already done and comes after this. So I propose that we leave this interesting refactoring to later work.

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,110 @@
...
+-/
+
+import Mathlib.RingTheory.TwoSidedIdeal.Operations
+import Mathlib.Topology.Algebra.Ring.Basic
+import Mathlib.Topology.Algebra.OpenSubgroup
+
+/-! # Linear topologies on rings
+
+Following Bourbaki, *Algebra II*, chapter 4, §2, n° 3, a topology on a ring `R` is *linear* if
+it is invariant by translation and there admits a basis of neighborhoods of 0 consisting of
```
ADedecker, line 14:
```suggestion
it is invariant by translation and admits a basis of neighborhoods of 0 consisting of
```

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,110 @@
...
+two-sided ideals.
+
+- `IsLinearTopology.tendsto_zero_mul`: for `f, g : ι → R` such that `f i` converges to `0`,
```
ADedecker, line 17: This doesn't match the lemma's new name, could you fix it?

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,113 @@
...
+`f i * g i` converges to `0`.
+
+-/
+
+open scoped Topology
+open Filter
+
+section Ring
+
+variable (α : Type*) [Ring α]
```
jcommelin, line 30: I would prefer `R` over `alpha`.

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,113 @@
...
+-/
+
+import Mathlib.RingTheory.TwoSidedIdeal.Operations
+import Mathlib.Topology.Algebra.Ring.Basic
+import Mathlib.Topology.Algebra.OpenSubgroup
+
+/-! # Linear topologies on rings
+
+Following Bourbaki, *Algebra II*, chapter 4, §2, n° 3, a topology on a ring `R` is *linear* if
+it is invariant by translation and admits a basis of neighborhoods of 0 consisting of
```
jcommelin, line 14: Does this imply that it is a topological ring: i.e. multiplication as bilinear map is continuous?

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,113 @@
...
+
+open scoped Topology
+open Filter
+
+section Ring
+
+variable (α : Type*) [Ring α]
+
+/-- A topology on a ring is linear if its topology is defined by a family of ideals. -/
+class IsLinearTopology [TopologicalSpace α] [TopologicalRing α] where
```
jcommelin, line 33: Similar to the comment above: is the translation-invariance mentioned above the same as `[TopologicalRing]` here?

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,113 @@
...
+    IsLinearTopology α where
+  hasBasis_twoSidedIdeal := h.to_hasBasis (fun i hi ↦ ⟨s i, h.mem_of_mem hi, subset_rfl⟩)
+    (fun _ ↦ h.mem_iff.mp)
+
+theorem isLinearTopology_iff_hasBasis_twoSidedIdeal :
+    IsLinearTopology α ↔ (𝓝 0).HasBasis
+      (fun I : TwoSidedIdeal α ↦ (I : Set α) ∈ 𝓝 0) (fun I : TwoSidedIdeal α ↦ (I : Set α)) :=
+  ⟨fun _ ↦ IsLinearTopology.hasBasis_twoSidedIdeal, fun h ↦ .mk_of_twoSidedIdeal h⟩
+
+theorem isLinearTopology_iff_hasBasis_open_twoSidedIdeal :
```
jcommelin, line 69: Apart from these two, all other lemmas begin their name with `IsLinearTopology.`. So I would put everything in a namespace, and prepend `_root_.` to these two names.

Mathlib/Topology/Algebra/LinearTopology.lean
ADedecker: Yes, replacing `[TopologicalRing α]` by `[ContinuousConstSMul α α]` would give (mathematically) the same definition. However:
- I've essentially proved that in a PR depending on #18437, so I don't think there's much sense redoing the work here instead of waiting for it to arrive.
- Even when we have these tools, I'm not sure we want the typeclass system to spend too much time synthesizing `TopologicalRing` instances. I would prefer that, in concrete cases, we just prove `TopologicalRing` first (using the special criterion), and then `LinearTopology`.

jcommelin: Thanks for clarifying!

In that case I completely agree with the design choice to assume `[TopologicalRing]`. But please point this out. E.g. in a `## Todo` that points to your wip.

Mathlib/Topology/Algebra/LinearTopology.lean
AntoineChambert-Loir: We can change that.

Mathlib/Topology/Algebra/LinearTopology.lean
AntoineChambert-Loir: Yes, but Bourbaki does it the other way round, I believe. In mathlib, we have to start with imposing that we have a topological ring. This is a bit sad, because for `IsLinearTopology.mk_of_twoSidedIdeal`, we could (mathematically) dispose of the hypothesis that we start from a topological ring.

AntoineChambert-Loir: Maybe we could bundle the topological ring hypothesis in the class `IsLinearTopology`.

AntoineChambert-Loir: @ADedecker , what do you think?

Mathlib/Topology/Algebra/LinearTopology.lean
AntoineChambert-Loir: OK.

Mathlib/Topology/Algebra/LinearTopology.lean
AntoineChambert-Loir: I have made a small modification: `LinearTopology` doesn't assume `TopologicalRing` anymore. The assumption is only added for those lemmas that require it. I have also added a note.

Mathlib/Topology/Algebra/LinearTopology.lean
kbuzzard: Yes but it is not changed? There seem to be several conversations here which were resolved but no action was taken?

AntoineChambert-Loir: I had made the push in the wrong branch…

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,148 @@
...
+ apply HasBasis.mk
+ intro t
+ simp only [mem_nhds_discrete, SetLike.mem_coe, TwoSidedIdeal.zero_mem, true_and]
+ constructor
+ · intro ht
+   use ⊥
+   change {0} ⊆ t
+   simp only [Set.singleton_subset_iff, ht]
+ · rintro ⟨J, hJt⟩
+   exact hJt J.zero_mem
```
jcommelin, line 105:
```suggestion
  rw [isLinearTopology_iff_hasBasis_twoSidedIdeal]
  apply HasBasis.mk
  intro t
  simp only [mem_nhds_discrete, SetLike.mem_coe, TwoSidedIdeal.zero_mem, true_and]
  constructor
  · intro ht
    use ⊥
    change {0} ⊆ t
    simp only [Set.singleton_subset_iff, ht]
  · rintro ⟨J, hJt⟩
    exact hJt J.zero_mem
```

ADedecker, line 105:
```suggestion
instance [DiscreteTopology R] : IsLinearTopology R :=
  have : HasBasis (𝓝 0 : Filter R) (fun _ ↦ True) (fun (_ : Unit) ↦ (⊥ : TwoSidedIdeal R)) := by
    rw [nhds_discrete]
    exact hasBasis_pure _
  .mk_of_twoSidedIdeal this
```
This (ab)uses the fact that `⊥ : TwoSidedIdeal R` coerces exactly to `{0}`.

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,148 @@
...
+
+theorem _root_.isLinearTopology_iff_hasBasis_open_twoSidedIdeal [TopologicalRing R] :
+    IsLinearTopology R ↔ (𝓝 0).HasBasis
+      (fun I : TwoSidedIdeal R ↦ IsOpen (I : Set R)) (fun I : TwoSidedIdeal R ↦ (I : Set R)) :=
+  ⟨fun _ ↦ hasBasis_open_twoSidedIdeal, fun h ↦ .mk_of_twoSidedIdeal h⟩
+
+instance [IsLinearTopology R] : Nonempty { J : TwoSidedIdeal R | (J : Set R) ∈ 𝓝 0} := by
+  obtain ⟨J, hJ, _⟩ :=
+    ((hasBasis_twoSidedIdeal (R := R)).mem_iff' Set.univ).mp (Filter.univ_mem)
+  exact ⟨J, hJ⟩
```
ADedecker, line 92: This does not require `IsLinearTopology`: the whole ring is always a two-sided ideal belonging to any filter. Do you really need this for some reason? If that's in a follow up PR, could you remove it from this one and we'll talk about it later?

AntoineChambert-Loir: For whatever lemma in #15007, Lean wanted that instance. I can remove it for the moment.

ADedecker (approved):
I've just had a last look, to me this is ready to go.

Mathlib/Topology/Algebra/LinearTopology.lean
```diff
@@ -0,0 +1,133 @@
...
+
+open scoped Topology
+open Filter
+
+namespace IsLinearTopology
+
+section Ring
+
+/-- A topology on a ring is linear if its topology is defined by a family of ideals. -/
+class _root_.IsLinearTopology (R : Type*) [Ring R] [TopologicalSpace R] where
```
urkud, line 45: I know nothing about this, but [Wikipedia](https://en.wikipedia.org/wiki/Linear_topology) defines linear topology in a different context, and it doesn't agree with this definition (it uses left submodules, not two sided ideals). I do not suggest that you change the definition (see start of this comment), but I think that existence of different definitions and some relations between them should be discussed in the module docstring.

ADedecker:
@urkud Thanks for your comment. Indeed what we were defining is a special case about linearly topologized modules. I adapted the file to talk about modules.

The only subtlety is that, to check this does indeed give the same thing for non-commutative rings, one needs to start from a basis of left ideals and one of right ideals to build one made of two-sided ideals, which needs a slight trick. I can move this part to a later PR if needed, but I think it also fits here.

jcommelin:
Thanks :tada:

bors merge

PR #14999, opened 2024-07-22
[Merged by Bors] - feat: Convex functions are continuous

YaelDillies:
Prove that convex functions from an open set in a finite dimensional real normed space are locally Lipschitz on that set.

---

- [x] depends on: #14660
- [x] depends on: #16111
- [x] depends on: #16115
- [x] depends on: #16117
- [x] depends on: #16118

### DIFF
--- a/Mathlib/Analysis/Convex/Continuous.lean
+++ b/Mathlib/Analysis/Convex/Continuous.lean
@@ -207,6 +207,7 @@ protected lemma ConvexOn.locallyLipschitz (hf : ConvexOn ℝ univ f) : LocallyLi
 protected lemma ConcaveOn.locallyLipschitz (hf : ConcaveOn ℝ univ f) : LocallyLipschitz f := by
   simpa using hf.locallyLipschitzOn_interior
 
+-- Commented out since `intrinsicInterior` is not imported (but should be once these are proved)
 -- proof_wanted ConvexOn.locallyLipschitzOn_intrinsicInterior (hf : ConvexOn ℝ C f) :
 --     ContinuousOn f (intrinsicInterior ℝ C)
 
--- a/Mathlib/Analysis/Convex/Continuous.lean
+++ b/Mathlib/Analysis/Convex/Continuous.lean
@@ -16,14 +16,11 @@ open FiniteDimensional Metric Set List Bornology
 open scoped Topology
 
 variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E]
-  {C : Set E} {f : E → ℝ} {x₀ : E} {ε r : ℝ}
+  {C : Set E} {f : E → ℝ} {x₀ : E} {ε r r' M : ℝ}
 
-lemma ConvexOn.exists_lipschitzOnWith_of_isBounded (hf : ConvexOn ℝ (ball x₀ r) f) (hε : 0 < ε)
-    (hf' : IsBounded (f '' ball x₀ r)) : ∃ K, LipschitzOnWith K f (ball x₀ (r - ε)) := by
-  rw [isBounded_iff_subset_ball 0] at hf'
-  simp only [Set.subset_def, mem_image, mem_ball, dist_zero_right, Real.norm_eq_abs,
-    forall_exists_index, and_imp, forall_apply_eq_imp_iff₂] at hf'
-  obtain ⟨M, hM⟩ := hf'
+lemma ConvexOn.lipschitzOnWith_of_abs_le (hf : ConvexOn ℝ (ball x₀ r) f) (hε : 0 < ε)
+    (hM : ∀ a, dist a x₀ < r → |f a| ≤ M) :
+    LipschitzOnWith (2 * M / ε).toNNReal f (ball x₀ (r - ε)) := by
   set K := 2 * M / ε with hK
   have oneside {x y : E} (hx : x ∈ ball x₀ (r - ε)) (hy : y ∈ ball x₀ (r - ε)) :
       f x - f y ≤ K * ‖x - y‖ := by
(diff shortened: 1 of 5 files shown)


Mathlib/Analysis/Normed/Group/Basic.lean
```diff
@@ -892,8 +878,90 @@ theorem SeminormedGroup.uniformCauchySeqOn_iff_tendstoUniformlyOn_one {f : ι 
     uniformCauchySeqOn_iff_uniformCauchySeqOnFilter,
     SeminormedGroup.uniformCauchySeqOnFilter_iff_tendstoUniformlyOnFilter_one]
 
+variable [T0Space E]
```
eric-wieser, line 881: What's the motivation for swapping this for `NormedGroup`? The two spellings are equivalent, right?

eric-wieser: (phrased another way: this refactor looks out of scope for this PR, and if you're serious about wanting to avoid NormedGroup entirely in favor of the less-bundled working, probably deserves a Zulip thread)

YaelDillies: Yes, true. They are equivalent. Why do we even have lemmas stated with `SeminormedGroup` + `T0Space`?

docs/undergrad.yaml
```diff
@@ -356,7 +356,8 @@ Single Variable Real Analysis:
     Weierstrass trigonometric approximation theorem: 'span_fourier_closure_eq_top'
   Convexity:
     convex functions of a real variable: 'ConvexOn'
-    continuity and differentiability of convex functions: 'https://en.wikipedia.org/wiki/Convex_function#Functions_of_one_variable'
+    continuity of convex functions: 'ConvexOn.continuousOn'
+    differentiability of convex functions: 'https://en.wikipedia.org/wiki/Convex_function#Functions_of_one_variable'
```
YaelDillies, line 360: @PatrickMassot, just checking: Is it fine if I split this item in two?

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,208 @@
...
+--     ContinuousOn f (intrinsicInterior ℝ C)
+
+-- proof_wanted ConcaveOn.locallyLipschitzOn_intrinsicInterior (hf : ConcaveOn ℝ C f) :
+--     ContinuousOn f (intrinsicInterior ℝ C)
+
+-- proof_wanted ConvexOn.continuousOn_intrinsicInterior (hf : ConvexOn ℝ C f) :
+--     ContinuousOn f (intrinsicInterior ℝ C)
+
+-- proof_wanted ConcaveOn.continuousOn_intrinsicInterior (hf : ConcaveOn ℝ C f) :
+--     ContinuousOn f (intrinsicInterior ℝ C)
```
YaelDillies, line 208: Just noting that those `proof_wanted` are linted against by a combination of the unused arguments linter and the new variable inclusion mechanism linter!

cc @Kha

docs/undergrad.yaml
jcommelin: my impression is that we really want a bijection between undergrad.yaml and the entries in the corresponding PDF sanctioned by the French government (or whatever).

So it would be find to include this in one of the other yaml's, but it will have to wait a bit before this one can be ticked off.

@PatrickMassot feel free to correct me if I'm wrong.

Mathlib/Analysis/Convex/Continuous.lean
Kha: That command should probably just disable those linters locally

YaelDillies: Opened https://github.com/leanprover-community/batteries/issues/938

docs/undergrad.yaml
eric-wieser: We've performened splits in the past - see "power series expansion of elementary functions" below, which gained sub-items.

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,213 @@
...
+    have hwyz : δ • w + (1 - δ) • y = z := by simp [w, hδ₀.ne', sub_smul]
+    have hw : dist w x₀ < ε := by
+      simpa [w, ← hx₀', dist_smul₀, abs_of_nonneg, hδ₀.le, inv_mul_lt_iff', hδ₀]
+    calc
+      f z ≤ max (f w) (f y) :=
+        hf.le_max_of_mem_segment (hr hw).2 hy ⟨_, _, hδ₀.le, sub_nonneg.2 hδ₁.le, by simp, hwyz⟩
+      _ ≤ max r (f y) := by gcongr; exact (hr hw).1
+  tfae_have 5 → 6
+  · rintro h x₀ hx₀
+    obtain ⟨r, hr⟩ := h hx₀
```
eric-wieser, line 113: I think this might make a good auxiliary lemma, after you've matched the quantifiers on this line?

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,210 @@
...
+  simp_rw [dist_eq_norm_sub, Real.norm_eq_abs, abs_sub_le_iff]
+  exact ⟨oneside hx hy, norm_sub_rev x _ ▸ oneside hy hx⟩
+
+lemma ConcaveOn.exists_lipschitzOnWith_of_isBounded (hf : ConcaveOn ℝ (ball x₀ r) f) (hε : 0 < ε)
+    (hf' : IsBounded (f '' ball x₀ r)) : ∃ K, LipschitzOnWith K f (ball x₀ (r - ε)) := by
+  replace hf' : IsBounded ((-f) '' ball x₀ r) := by convert hf'.neg; ext; simp [neg_eq_iff_eq_neg]
+  simpa using hf.neg.exists_lipschitzOnWith_of_isBounded hε hf'
+
+lemma ConvexOn.continuousOn_tfae (hC : IsOpen C) (hC' : C.Nonempty) (hf : ConvexOn ℝ C f) :
+    TFAE [LocallyLipschitzOn C f, ContinuousOn f C, ∃ x₀ ∈ C, ContinuousAt f x₀,
```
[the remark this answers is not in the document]
b-mehta: Make sure to do the same for ConcaveOn!

docs/undergrad.yaml
YaelDillies: I have done so

Mathlib/Analysis/Convex/Continuous.lean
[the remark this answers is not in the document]
YaelDillies: Whoops, did so, but not so well

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,220 @@
...
+-/
+
+open FiniteDimensional Metric Set List Bornology
+open scoped Topology
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E]
+  {C : Set E} {f : E → ℝ} {x₀ : E} {ε r : ℝ}
+
+lemma ConvexOn.exists_lipschitzOnWith_of_isBounded (hf : ConvexOn ℝ (ball x₀ r) f) (hε : 0 < ε)
+    (hf' : IsBounded (f '' ball x₀ r)) : ∃ K, LipschitzOnWith K f (ball x₀ (r - ε)) := by
```
ocfnash, line 22: How about:
```suggestion
lemma ConvexOn.exists_lipschitzOnWith_of_isBounded (hf : ConvexOn ℝ (ball x₀ r) f)
    (hf' : IsBounded (f '' ball x₀ r)) {r' : ℝ} (h : r' < r) :
    ∃ K, LipschitzOnWith K f (ball x₀ r') := by
```

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,220 @@
...
+        rw [sub_eq_add_neg (f _), two_mul]
+        gcongr
+        · exact (le_abs_self _).trans <| (hM _ hz).le
+        · exact (neg_le_abs _).trans <| (hM _ hx').le
+  refine ⟨K.toNNReal, .of_dist_le' fun x hx y hy ↦ ?_⟩
+  simp_rw [dist_eq_norm_sub, Real.norm_eq_abs, abs_sub_le_iff]
+  exact ⟨oneside hx hy, norm_sub_rev x _ ▸ oneside hy hx⟩
+
+lemma ConcaveOn.exists_lipschitzOnWith_of_isBounded (hf : ConcaveOn ℝ (ball x₀ r) f) (hε : 0 < ε)
+    (hf' : IsBounded (f '' ball x₀ r)) : ∃ K, LipschitzOnWith K f (ball x₀ (r - ε)) := by
```
ocfnash, line 69: As above, I think preferable to use `r' < r` rather than `r - ε`.

Mathlib/Analysis/Convex/Continuous.lean
```diff
@@ -0,0 +1,220 @@
...
+  replace hf' : IsBounded ((-f) '' ball x₀ r) := by convert hf'.neg; ext; simp [neg_eq_iff_eq_neg]
+  simpa using hf.neg.exists_lipschitzOnWith_of_isBounded hε hf'
+
+lemma ConvexOn.isBoundedUnder_abs (hf : ConvexOn ℝ C f) {x₀ : E} (hC : C ∈ 𝓝 x₀) :
```
ocfnash, line 73: Just a remark that this lemma seems to make a case for a `ConvexAt` predicate.

ocfnash (approved):
Thanks!

bors d+

Mathlib/Analysis/Convex/Continuous.lean
YaelDillies: I tend to agree 😁 but I have no other use case in mind for now

Mathlib/Analysis/Convex/Continuous.lean
YaelDillies: I am not convinced. As you see in the calculations, `ε` is the important quantity, and one could write a version of the lemma giving the Lipschitz constant in terms of the bound and `ε`.

YaelDillies:
bors merge

PR #15001, opened 2024-07-22
[Merged by Bors] - refactor(AlgebraicGeometry): Introduce `Scheme.Opens`.

erdOne:
Introduced `Scheme.Opens` as the type of open sets of a scheme.  Provided `instance : CoeOut X.Opens Scheme` to replace the notation `X ∣_ᵤ U`, and renamed `Scheme.ιOpens` to `Scheme.Opens.ι` to enable dot notation.

---

Main changes are in `Restrict.lean`.


erdOne:
!bench

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
-abbrev Scheme.ιOpens {X : Scheme.{u}} (U : Opens X) : X ∣_ᵤ U ⟶ X := X.ofRestrict _
+@[simps! val_base_apply]
+def ι : ↑U ⟶ X := X.ofRestrict _
+
+instance : IsOpenImmersion U.ι := inferInstanceAs (IsOpenImmersion (X.ofRestrict _))
+
+lemma toScheme_carrier : (U : Type u) = (U : Set X) := rfl
+lemma toScheme_presheaf_obj (V) : Γ(U, V) = Γ(X, U.ι ''ᵁ V) := rfl
+@[simp] lemma toScheme_presheaf_map {V W} (i : V ⟶ W) :
+    U.toScheme.presheaf.map i = X.presheaf.map (U.ι.opensFunctor.map i.unop).op := rfl
```
chrisflav, line 57:
```suggestion
lemma toScheme_carrier : (U : Type u) = (U : Set X) := rfl

lemma toScheme_presheaf_obj (V) : Γ(U, V) = Γ(X, U.ι ''ᵁ V) := rfl

@[simp]
lemma toScheme_presheaf_map {V W} (i : V ⟶ W) :
    U.toScheme.presheaf.map i = X.presheaf.map (U.ι.opensFunctor.map i.unop).op := rfl
```
This seems more consistent with what follows below.

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
+      U.toScheme.presheaf.germ x ≫ (U.stalkIso x.1).hom =
+        X.presheaf.germ ⟨x.1.1, show x.1.1 ∈ U.ι ''ᵁ V from ⟨x.1, x.2, rfl⟩⟩ :=
+    PresheafedSpace.restrictStalkIso_hom_eq_germ _ _ _ _ _
+
+end Scheme.Opens
+
+/-- If `U` is a family of open sets that covers `X`, then `X.restrict U` forms an `X.open_cover`. -/
+@[simps! J obj map]
+def Scheme.openCoverOfSuprEqTop {s : Type*} (X : Scheme.{u}) (U : s → X.Opens)
+    (hU : ⨆ i, U i = ⊤) : X.OpenCover where
```
chrisflav, line 130:
```suggestion
def Scheme.openCoverOfiSupEqTop {s : Type*} (X : Scheme.{u}) (U : s → X.Opens)
    (hU : ⨆ i, U i = ⊤) : X.OpenCover where
```

chrisflav (commented):
I'll review more tomorrow.

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
 
 attribute [-simp] eqToHom_op in
 /-- The global sections of the restriction is isomorphic to the sections on the open set. -/
 @[simps!]
-def Scheme.resTop (X : Scheme.{u}) (U : Opens X) : Γ(X ∣_ᵤ U, ⊤) ≅ Γ(X, U) :=
-  X.presheaf.mapIso (eqToIso U.openEmbedding_obj_top.symm).op
+def topIso : Γ(U, ⊤) ≅ Γ(X, U) :=
+  X.presheaf.mapIso (eqToIso U.ι_image_top.symm).op
+
+/-- The stalks of an open subscheme is isomorphic to the stalk of the original scheme. -/
```
chrisflav, line 113:
```suggestion
/-- The stalks of an open subscheme are isomorphic to the stalks of the original scheme. -/
```

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
+def opensRestrict :
+    Scheme.Opens U ≃ { V : X.Opens // V ≤ U } :=
+  (IsOpenImmersion.opensEquiv (U.ι)).trans (Equiv.subtypeEquivProp (by simp))
 
-instance ΓRestrictAlgebra {X : Scheme.{u}} {Y : TopCat.{u}} {f : Y ⟶ X} (hf : OpenEmbedding f) :
-    Algebra (Scheme.Γ.obj (op X)) (Scheme.Γ.obj (op <| X.restrict hf)) :=
-  (Scheme.Γ.map (X.ofRestrict hf).op).toAlgebra
+instance ΓRestrictAlgebra {X : Scheme.{u}} (U : X.Opens) :
+    Algebra (Scheme.Γ.obj (op X)) (Scheme.Γ.obj (op U)) :=
+  (Scheme.Γ.map U.ι.op).toAlgebra
```
chrisflav, line 150:
```suggestion
instance ΓRestrictAlgebra {X : Scheme.{u}} (U : X.Opens) :
    Algebra Γ(X, ⊤) Γ(U, ⊤) :=
  (Scheme.Γ.map U.ι.op).toAlgebra
```

Mathlib/AlgebraicGeometry/Properties.lean
```diff
@@ -203,6 +202,10 @@ instance (priority := 900) isReduced_of_isIntegral [IsIntegral X] : IsReduced X
   · haveI : Nonempty U := by simpa
     infer_instance
 
+instance Scheme.component_nontrivial (X : Scheme.{u}) (U : Opens X) [Nonempty U] :
+    Nontrivial Γ(X, U) :=
```
chrisflav, line 206:
```suggestion
instance Scheme.component_nontrivial (X : Scheme.{u}) (U : X.Opens) [Nonempty U] :
    Nontrivial Γ(X, U) :=
```

chrisflav (commented):
It looks like it does not make anything worse and some things definitely look better. The old `X ∣_ᵤ U` looked a bit funky in some places. Also that we can use dot notation on the open set is nice and opens up a potential refactor of `IsAffineOpen`.

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
-    (ιOpens U).app U = X.presheaf.map (eqToHom (by simp)).op := rfl
+lemma ι_app_self : U.ι.app U = X.presheaf.map (eqToHom (X := U.ι ''ᵁ _) (by simp)).op := rfl
 
-lemma Scheme.eq_restrict_presheaf_map_eqToHom {X : Scheme.{u}} (U : Opens X) {V W : Opens U}
-    (e : Scheme.ιOpens U ''ᵁ V = ιOpens U ''ᵁ W) :
-  X.presheaf.map (eqToHom e).op =
-    (X ∣_ᵤ U).presheaf.map (eqToHom <| U.openEmbedding.functor_obj_injective e).op := rfl
+lemma eq_presheaf_map_eqToHom {V W : Opens U} (e : U.ι ''ᵁ V = U.ι ''ᵁ W) :
+    X.presheaf.map (eqToHom e).op =
+      U.toScheme.presheaf.map (eqToHom <| U.openEmbedding.functor_obj_injective e).op := rfl
```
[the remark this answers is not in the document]
erdOne: I don't think so.

Mathlib/AlgebraicGeometry/Restrict.lean
erdOne: This is copy pasted from elsewhere. I'll make this change in a follow-up PR.

Mathlib/AlgebraicGeometry/Restrict.lean
erdOne: This is copy pasted from elsewhere. I'll make this change in a follow-up PR.

Mathlib/AlgebraicGeometry/Restrict.lean
```diff
@@ -34,55 +34,123 @@ variable {C : Type u₁} [Category.{v} C]
...
+    Quiver.Hom.unop_op, Hom.opensFunctor_map_homOfLE, Opens.coe_inclusion, ← Functor.map_comp]
+  rfl
+
+@[simp]
+lemma ι_appIso (V) : U.ι.appIso V = Iso.refl _ :=
+  X.ofRestrict_appIso _ _
+
+@[simp]
+lemma opensRange_ι : U.ι.opensRange = U :=
+  Opens.ext Subtype.range_val
```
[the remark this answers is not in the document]
erdOne: I think the naming convention calls for `opensRange_ι` and `appIso_ι` but defs that are morally "components" of the main thing is put in the end. Not 100% sure though.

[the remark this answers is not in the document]
chrisflav: Okay.

chrisflav (approved):
Thanks! LGTM.

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -365,11 +364,12 @@ theorem SpecΓIdentity_hom_app_fromSpec :
...
-    Scheme.ofRestrict_app_self]
-  simp only [Category.assoc]
-  dsimp only [asIso_inv, Functor.op_obj, unop_op]
-  rw [← Functor.map_comp_assoc, ← op_comp, eqToHom_trans, Scheme.eq_restrict_presheaf_map_eqToHom,
+  rw [Scheme.comp_app, Scheme.comp_app, ΓSpecIso_obj_hom, Scheme.Opens.ι_app_self]
+  simp only [asIso_inv, Scheme.comp_coeBase, Opens.map_comp_obj, Scheme.Opens.topIso_inv,
+    Opens.map_top, Functor.id_obj, Functor.comp_obj,
+    Functor.rightOp_obj, Scheme.Γ_obj, unop_op, Scheme.Spec_obj, Scheme.Opens.topIso_hom,
+    Category.assoc]
+  rw [← Functor.map_comp_assoc, ← op_comp, eqToHom_trans, Scheme.Opens.eq_presheaf_map_eqToHom,
```
erdOne, line 372: The slowdown on this file comes from this proof and it will be improved in a later PR where the unit of the gamma-spec adjunction is specialized and be provided dedicated API.

mattrobball: This whole proof can be solved by a single `simp only [many things are here]` and the `have` can be deleted.

erdOne: Do you mean introducing intermediate lemmas? I'm not sure how to make this work.

erdOne: Okay I managed to condense it into two blobs of simp and removed the nolint and the maxheartbeat.

mattrobball:
You might be able to improve your performance by removing the attribute making `Quiver.Hom` reducible. It doesn't seem to do anything except make Lean work harder now

Mathlib/AlgebraicGeometry/AffineScheme.lean
erdOne: By the way does `rw` unfolds things that `simp` does not? I though they look through the same transparency level (both can see through reducible defs but not semireducible ones) but am I wrong?

mattrobball: I *think* `rw` can see through `semi-reducible` terms. The discrimination tree for `simp` is pretty good at filtering things that are only defeq at default transparency.

mattrobball: Hmm. I think it is *supposed* to operate at `reducible` but I can't tell if it is doing anything atm

erdOne:
!bench

Mathlib/AlgebraicGeometry/AffineScheme.lean
mattrobball: The actual substitution via `abstract` is done at `reducible` transparency but there is at least one other `isDefEq` call that is done at `default`. So we start unfolding these more than is desired here

erdOne: Is that intended? If this happens to `rw` everywhere, it sounds like a performance hazard to me.

mattrobball: With the huge caveat that I just looked at the file for a couple of minutes, no.

mattrobball:
Ready to go?

Mathlib/AlgebraicGeometry/AffineScheme.lean
mattrobball: I'll check on this at office hours tomorrow at the latest

mattrobball:
bors merge

PR #15005, opened 2024-07-22
[Merged by Bors] - refactor: Make sure concrete Heyting algebras define their `himp` explicitly

YaelDillies:
There are a few structures in Mathlib which mathematically are Heyting algebras but do not actually define their Heyting implication, or define it implicitly as `x ⇨ y := y ⊔ xᶜ` when a nicer formula is available. This PR makes sure this is not the case.

---


Mathlib/Combinatorics/SimpleGraph/Finsubgraph.lean
```diff
@@ -59,21 +61,45 @@ instance : Sup G.Finsubgraph :=
...
-instance [Finite V] : Top G.Finsubgraph :=
-  ⟨⟨⊤, finite_univ⟩⟩
+section Finite
+variable [Finite V]
+
+instance instTop : Top G.Finsubgraph where top := ⟨⊤, finite_univ⟩
+instance instSupSet : SupSet G.Finsubgraph where sSup s := ⟨⨆ G ∈ s, ↑G, Set.toFinite _⟩
+instance instInfSet : InfSet G.Finsubgraph where sInf s := ⟨⨅ G ∈ s, ↑G, Set.toFinite _⟩
+
+@[simp, norm_cast] lemma coe_top : (⊤ : G.Finsubgraph) = (⊤ : G.Subgraph) := rfl
```
eric-wieser, line 82: I'd be tempted to write the `\u`s explicitly here, just to make the naming easy to justify.

Mathlib/Combinatorics/SimpleGraph/Subgraph.lean
```diff
@@ -879,10 +882,6 @@ taking the portion of `G` that intersects `G'`. -/
 protected abbrev restrict {G' : G.Subgraph} : G.Subgraph → G'.coe.Subgraph :=
   Subgraph.comap G'.hom
 
-@[simp]
-lemma verts_coeSubgraph {G' : Subgraph G} (G'' : Subgraph G'.coe) :
```
eric-wieser, line 883: Where did this go?

YaelDillies: The prophecy...

jcommelin:
Thanks :tada:

bors merge

PR #15009, opened 2024-07-22
[Merged by Bors] - feat(Analysis/Complex/Hadamard): generalize Hadamard's three lines theorem

madeve-unipi:
Make Hadamard's three lines theorem work on any vertical strip of the complex plane (i.e., the set of complex numbers of real part between `l` and `u`, provided l is less than u). Previously, the theorem would only work for `l=0` and `u=1`.

Moves: 
- norm_le_interp_of_mem_verticalClosedStrip -> norm_le_interp_of_mem_verticalClosedStrip₀₁
- norm_le_interp_of_mem_verticalClosedStrip' -> norm_le_interp_of_mem_verticalClosedStrip₀₁'

---


dupuisf:
This looks quite good. Can you merge master to get rid of the merge conflict?

madeve-unipi:
@dupuisf I merged master but I'm not sure what went wrong with building mathlib

dupuisf:
If you click on the red X next to the failed commit, you can see which step failed. Here it's the "build" step, and if you 
click on "Details", you can see what went wrong. In this case, the issue is that there are several unused arguments.

dupuisf:
> I wonder why this wasn't an issue in the original commit, but it's fixed now. Thank you!

The latest version of Lean includes a change in the variable inclusion mechanism, leading to a lot of breakage like this.

dupuisf (commented):
This looks quite good. I have a first round of comments, mostly about coding style and removing lemmas that are a bit out of place in this file. Also, I have one request: could you add back versions of the theorem with the strip from 0 to 1? It's a bit easier to use since there's no division by `u-l` in the exponents and so on.

madeve-unipi:
Thank you for your comments. I should have fixed everything that was brought up.

> Also, I have one request: could you add back versions of the theorem with the strip from 0 to 1? It's a bit easier to use since there's no division by `u-l` in the exponents and so on.

I am not sure I understand what you mean here. The old versions should still be usable under [theoremname]\_{01}, see lines 371 and 385 (I'm sorry but I don't know how to paste them here like you did in the review).

dupuisf:
Another quick note: to make things easier for reviewers, can you click on "resolve conversation" whenever a comment has been addressed?

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
         simp only [mem_preimage, one_re, mem_singleton_iff, comp_apply,
           and_self]
 
+
+/--An auxiliary function to prove the general statement of Hadamard's three lines theorem.-/
+def scale (f: ℂ → E) (l u : ℝ) : ℂ → E := fun z ↦ f (l + z • (u-l))
```
dupuisf, line 431:
```suggestion
/-- An auxiliary function to prove the general statement of Hadamard's three lines theorem.-/
def scale (f: ℂ → E) (l u : ℝ) : ℂ → E := fun z ↦ f (l + z • (u-l))
```
and likewise below

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+
+/--The transformation on ℂ that is used for `scale` maps the strip ``re ⁻¹' (l,u)``
+  to the strip ``re ⁻¹' (0,1)``-/
+lemma scale_mapsto_dom {l u : ℝ} (hul: l<u) {z: ℂ} (hz: z ∈ verticalStrip 0 1) :
+    l + z * (u-l)  ∈ verticalStrip l u := by
```
dupuisf, line 436: I think this should either be named something like `scale_mem_verticalStrip_of_mem_verticalStrip`, or else actually use `Set.MapsTo`.

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+/--A bound to the norm of `f` on the line `z.re=u` induces a bound to the norm of `scale f l u z`
+  on the line `z.re=1`. -/
+lemma scale_bound_right {f: ℂ → E} {l u b : ℝ} (hb : ∀ z ∈ re ⁻¹' {u}, ‖f z‖ ≤ b) :
+    ∀ z ∈ re ⁻¹' {1}, ‖scale f l u z‖ ≤ b := by
+  simp only [scale, mem_preimage, mem_singleton_iff, smul_eq_mul]
+  intro z hz
+  exact hb (↑l + z * (↑u - ↑l)) (by simp[hz])
+
+/--A technical lemma needed in the proof-/
+lemma fun_arg_eq {l u: ℝ} (hul: l < u) (z: ℂ) :
```
dupuisf, line 534:
```suggestion
/--A technical lemma needed in the proof-/
private lemma fun_arg_eq {l u: ℝ} (hul: l < u) (z: ℂ) :
```
Let's make this a private lemma, something like this should definitely not appear in the docs.

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+  · apply DiffContOnCl.const_add
+    apply DiffContOnCl.smul_const
+    exact Differentiable.diffContOnCl differentiable_id'
+  · rw [MapsTo]
+    intro z hz
+    exact scale_mapsto_dom hul hz
+
+
+/-- The norm of the function `scale f l u` is bounded above on the closed strip `re⁻¹' [0, 1]`-/
+lemma scale_bddAbove {f: ℂ → E} {l u : ℝ} (hul: l< u)
```
dupuisf, line 500:
```suggestion
/-- The norm of the function `scale f l u` is bounded above on the closed strip `re⁻¹' [0, 1]`-/
lemma scale_bddAbove {f: ℂ → E} {l u : ℝ} (hul: l < u)
```
Can you standardize the spacing in the file? Normally things like `u-l` should really be written as `u - l` and so on.

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+    · exact hz.1
+  · rw [← sub_le_sub_iff_right (l / (u-l)), add_sub_assoc, sub_self, add_zero, div_sub_div_same,
+      div_le_one (by simp[hul]), sub_le_sub_iff_right l]
+    exact hz.2
+
+
+/-- The function `scale f l u` is `diffContOnCl`. -/
+lemma scale_diffContOnCl [NormedSpace ℂ E] {f: ℂ → E} {l u : ℝ} (hul: l < u)
+    (hd : DiffContOnCl ℂ f (verticalStrip l u)) : DiffContOnCl ℂ (scale f l u)
+    (verticalStrip 0 1) := by
```
dupuisf, line 488:
```suggestion
lemma scale_diffContOnCl [NormedSpace ℂ E] {f: ℂ → E} {l u : ℝ} (hul: l < u)
    (hd : DiffContOnCl ℂ f (verticalStrip l u)) : 
    DiffContOnCl ℂ (scale f l u) (verticalStrip 0 1) := by
```
It's much easier to read this way.

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,229 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+`z.re = u`, `‖f z‖ ≤ b` for some `b ∈ ℝ` then for all `z` in the closed strip
+`re ⁻¹' [l,u]` the inequality `‖f(z)‖ ≤ a ^ (1 - (z.re-l)/(u-l)) * b ^ ((z.re-l)/(u-l))` holds. -/
+lemma norm_le_interp_of_mem_verticalClosedStrip' [NormedSpace ℂ E] {f : ℂ → E} {z : ℂ} {a b l u : ℝ}
+    (hul: l < u) (hz : z ∈ verticalClosedStrip l u) (hd : DiffContOnCl ℂ f (verticalStrip l u))
+    (hB : BddAbove ((norm ∘ f) '' (verticalClosedStrip l u)))
+    (ha : ∀ z ∈ re ⁻¹' {l}, ‖f z‖ ≤ a) (hb : ∀ z ∈ re ⁻¹' {u}, ‖f z‖ ≤ b) :
+    ‖f z‖ ≤ a ^ (1 - (z.re-l)/(u-l)) * b ^ ((z.re-l)/(u-l)) := by
+
+  have hgoal := norm_le_interp_of_mem_verticalClosedStrip₀₁' (scale f l u)
+    (scale_mem_strip hul hz) (scale_diffContOnCl hul hd) (scale_bddAbove hul hB)
```
dupuisf, line 647:
```suggestion
  have hgoal := norm_le_interp_of_mem_verticalClosedStrip₀₁' (scale f l u)
    (scale_mem_strip hul hz) (scale_diffContOnCl hul hd) (scale_bddAbove hul hB)
```

Mathlib/Analysis/Complex/Hadamard.lean
madeve-unipi: I've changed the name (and the following ones) as suggested.
While doing so, I remembered that technically this lemma refers to the map `\varphi: \mathbb{C} \to \mathbb{C}` given by `z \mapsto l + z * (u - l)`, whereas scale maps a function f to a function `(scale f) z = f (\varphi z)`. I am not sure whether this is fine or I should define this `\varphi` separately (possibly with a nicer name) and make the lemma about it.

dupuisf: Yeah I think that's fine.

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -10,16 +10,17 @@ import Mathlib.Analysis.Complex.PhragmenLindelof
...
-In this file we present a proof of a special case of Hadamard's three-lines Theorem.
+In this file we present a proof of Hadamard's three-lines Theorem.
 
 ## Main result
 
 - `norm_le_interp_of_mem_verticalClosedStrip` :
-Hadamard three-line theorem on `re ⁻¹' [0,1]`: If `f` is a bounded function, continuous on
-`re ⁻¹' [0,1]` and differentiable on `re ⁻¹' (0,1)`, then for
+Hadamard three-line theorem: If `f` is a bounded function, continuous on
+`re ⁻¹' [l,u]` and differentiable on `re ⁻¹' (l,u)`, then for
```
YaelDillies, line 19:
```suggestion
`re ⁻¹' [l, u]` and differentiable on `re ⁻¹' (l, u)`, then for
```

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -393,12 +397,13 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
       rcases h with ⟨h0, h1⟩
       rw [ne_comm] at h0 h1
       simp_rw [interpStrip_eq_of_pos f _ (lt_of_le_of_ne (sSupNormIm_nonneg f 0) h0)
-        (lt_of_le_of_ne (sSupNormIm_nonneg f 1) h1)]
+      (lt_of_le_of_ne (sSupNormIm_nonneg f 1) h1)]
```
YaelDillies, line 400: That's against the style guide. Please revert
```suggestion
        (lt_of_le_of_ne (sSupNormIm_nonneg f 1) h1)]
```

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -393,12 +397,13 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
       simp only [norm_eq_abs, map_mul]
       rw [abs_cpow_eq_rpow_re_of_pos ((Ne.le_iff_lt h0).mp (sSupNormIm_nonneg f _)) _]
       rw [abs_cpow_eq_rpow_re_of_pos ((Ne.le_iff_lt h1).mp (sSupNormIm_nonneg f _)) _]
       simp only [sub_re, one_re, le_refl]
-  apply (norm_le_interpStrip_of_mem_verticalClosedStrip f hz hd hB).trans (this.trans _)
+
```
YaelDillies, line 405: Why a blank line? If you feel like this is a nice spot to semantically split the proof, then you should add a comment, not a blank line
```suggestion
```

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,232 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
...
+  simp only [verticalClosedStrip, mem_preimage, add_re, ofReal_re, mul_re, sub_re, sub_im,
+    ofReal_im, sub_self, mul_zero, sub_zero, mem_Icc, le_add_iff_nonneg_right]
+  simp only [verticalClosedStrip, mem_preimage, mem_Icc] at hz
+  obtain ⟨hz₁, hz₂⟩ := hz
+  simp only [sub_pos, hul, mul_nonneg_iff_of_pos_right, hz₁, true_and]
+  rw [add_comm, ← sub_le_sub_iff_right l, add_sub_assoc, sub_self, add_zero]
+  nth_rewrite 2 [← one_mul (u - l)]
+  gcongr
+  simp only [sub_nonneg]
+  exact le_of_lt hul
```
YaelDillies, line 462:
```suggestion
  have := sub_nonneg.2 hul.le
  gcongr
```

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +426,232 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
         simp only [mem_preimage, one_re, mem_singleton_iff, comp_apply,
           and_self]
 
+
```
YaelDillies, line 429: We only put one space between declarations
```suggestion
```

madeve-unipi:
@YaelDillies Thank you very much! I fixed the things you pointed out.

YaelDillies (commented):
I've moved some lemmas around so that you don't need to remove the `variable [NormedSpace ℂ E]` anymore.

Some general remark: Why don't generalise the existing lemmas from the 0-1 strip to a general strip rather than reducing the general case to the 0-1 strip by a transformation?
To keep Frédéric happy, you can still keep the 0-1 special case, but prove it in terms of the general case.

madeve-unipi:
> Some general remark: Why don't generalise the existing lemmas from the 0-1 strip to a general strip rather than reducing the general case to the 0-1 strip by a transformation?

This is essentially because I had proved the result from scratch for a university course (https://florisvandoorn.com/BonnAnalysis/) and then we realized the 0-1 strip case was already here. My proof was the one in Stein-Shakarchi's Functional Analysis book. It was very convenient to have 0 and 1 and then I obtained the rest via a transformation, so I just adapted that part to commit here. I haven't really studied the details of how the current mathlib proof actually works. I can try to go through it if it works better, but I don't think I will have enough time for that in the next couple of weeks.

YaelDillies:
Okay, fair enough. I think mathlib's proof could be easy to adapt by applying the transformation earlier (although I'm not sure). If you don't have time for now, this can be left as a TODO.

Please address my comments above, however.

madeve-unipi:
I am sorry I completely forgot to do this. I removed the unnecessary brackets around `verticalClosedStrip 0 1` and `sSupNormIm` throughout the document. It is not entirely clear to me whether or not more brackets can be removed. For example, why is `f '' VerticalClosedStrip 0 1 ` perfectly fine, but `f  ' ' (re ⁻¹' {x})` requires bracketing?

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -296,10 +301,126 @@ lemma diffContOnCl_interpStrip :
...
+    then 0
+    else sSupNormIm f l ^ (1 - ((z - l) / (u - l))) * sSupNormIm f u ^ ((z - l) / (u - l))
+
+/-- An auxiliary function to prove the general statement of Hadamard's three lines theorem. -/
+def scale (f : ℂ → E) (l u : ℝ) : ℂ → E := fun z ↦ f (l + z • (u - l))
+
+/-- The transformation on ℂ that is used for `scale` maps the closed strip ``re ⁻¹' [l, u]``
+  to the closed strip ``re ⁻¹' [0, 1]``. -/
+lemma scale_id_mem_verticalClosedStrip_of_mem_verticalClosedStrip {l u : ℝ} (hul : l < u) {z : ℂ}
+    (hz : z ∈ verticalClosedStrip 0 1) : l + z * (u - l)  ∈ verticalClosedStrip l u := by
```
YaelDillies, line 316:
```suggestion
    (hz : z ∈ verticalClosedStrip 0 1) : l + z * (u - l) ∈ verticalClosedStrip l u := by
```

Mathlib/Analysis/Complex/Hadamard.lean
```diff
@@ -421,5 +542,107 @@ lemma norm_le_interp_of_mem_verticalClosedStrip' (f : ℂ → E) {z : ℂ} {a b
         simp only [mem_preimage, one_re, mem_singleton_iff, comp_apply,
           and_self]
 
+/-- The transformation on ℂ that is used for `scale` maps the strip ``re ⁻¹' (l, u)``
+  to the strip ``re ⁻¹' (0, 1)``. -/
+lemma scale_id_mem_verticalStrip_of_mem_verticalStrip {l u : ℝ} (hul : l < u) {z : ℂ}
+    (hz : z ∈ verticalStrip 0 1) : l + z * (u - l)  ∈ verticalStrip l u := by
```
YaelDillies, line 548:
```suggestion
    (hz : z ∈ verticalStrip 0 1) : l + z * (u - l) ∈ verticalStrip l u := by
```

YaelDillies (approved):
Thanks! Maybe merge master once to make sure everything still works.

maintainer delegate

jcommelin (commented):
bors d+

madeve-unipi:
I merged master and committed your spacing fix, everything looks fine to me

madeve-unipi:
> Because precedences are wrong. Feel free to open an issue for that on this repo!

Do you mean that `f '' (re ⁻¹' {x})` should not require brackets and  this is a Lean issue I should open?
Does this look right?

Title: Notation precedence issue with `' '`
Description: The expression `f '' (re ⁻¹' {x})` requires brackets when it should not. See for example Mathlib/Analysis/Complex/Hadamard.lean

madeve-unipi:
bors merge

YaelDillies:
> Do you mean that `f '' (re ⁻¹' {x})` should not require brackets and this is a Lean issue I should open?
> Does this look right?

It's not a Lean issue, it's a Mathlib issue

PR #15010, opened 2024-07-22
[Merged by Bors] - feat: add the `cdot` linter

adomani:
The `cdot` linter is a syntax linter that flags uses of the "cdot" `·` that are achieved by typing a character different from `·`.

For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter.

---

### DIFF
--- a/Mathlib/Tactic/Linter/Lint.lean
+++ b/Mathlib/Tactic/Linter/Lint.lean
@@ -158,4 +158,66 @@ initialize addLinter missingEndLinter
 
 end MissingEnd
 
+/-!
+# The `cdot` linter
+
+The `cdot` linter is a syntax-linter that flags uses of the "cdot" `·` that are achieved
+by typing a character different form `·`.
+For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter.
+-/
+
+/--
+The `cdot` linter flags uses of the "cdot" `·` that are achieved by typing a character
+different form `·`.
+For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter. -/
+register_option linter.cdot : Bool := {
+  defValue := true
+  descr := "enable the `cdot` linter"
+}
+
+/-- `isCDot? stx` checks whether `stx` is a `Syntax` node corresponding to a `cdot` typed with
+the character `·`. -/
+def isCDot? : Syntax → Bool
+  | .node _ ``cdotTk #[.node _ `patternIgnore #[.node _ _ #[.atom _ v]]] => v == "·"
+  | .node _ ``Lean.Parser.Term.cdot #[.atom _ v] => v == "·"
+  | _ => false
+
+/--
+`findDot stx` extracts from `stx` the syntax nodes of `kind` `Lean.Parser.Term.cdot` or `cdotTk`. -/
+partial
+def findDot : Syntax → Array Syntax
+  | stx@(.node _ k args) =>
+    let dargs := (args.map findDot).flatten
+    match k with
+      | ``Lean.Parser.Term.cdot => dargs.push stx
+      | ``cdotTk => dargs.push stx
+      | _ =>  dargs
+  |_ => #[]
+
+/-- `unwanted.cDot stx` returns an array of syntax atoms corresponding to `cdot`s that are not
+written with the character `·`.
+This is precisely what the `cdot` linter flags.
+-/
+def unwanted.cDot (stx : Syntax) : Array Syntax :=
+  (findDot stx).filter (!isCDot? ·)
+
+namespace CDotLinter
+
+/-- Gets the value of the `linter.generic` option. -/
+def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.cdot o
+
+@[inherit_doc linter.cdot]
+def cdotLinter : Linter where run := withSetOptionIn fun stx => do
+    unless getLinterHash (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    for s in unwanted.cDot stx do
+      Linter.logLint linter.cdot s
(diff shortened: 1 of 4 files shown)


Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -158,4 +158,66 @@ initialize addLinter missingEndLinter
 
 end MissingEnd
 
+/-!
+# The `cdot` linter
+
+The `cdot` linter is a syntax-linter that flags uses of the "cdot" `·` that are achieved
+by typing a character different form `·`.
```
grunweg, line 165:
```suggestion
by typing a character different from `·`.

File: Mathlib/Tactic/Linter/Lint.lean [hunk 2]
```diff
@@ -158,4 +158,66 @@ initialize addLinter missingEndLinter
...
+      | ``Lean.Parser.Term.cdot => dargs.push stx
+      | ``cdotTk => dargs.push stx
+      | _ =>  dargs
+  |_ => #[]
+
+/-- `unwanted.cDot stx` returns an array of syntax atoms corresponding to `cdot`s that are not
+written with the character `·`.
+This is precisely what the `cdot` linter flags.
+-/
+def unwanted.cDot (stx : Syntax) : Array Syntax :=
```
grunweg, line 201: Given this is the first linter for unwanted syntax, what do you think about naming this "unwanted_cdot" instead?

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -158,4 +158,66 @@ initialize addLinter missingEndLinter
...
+the character `·`. -/
+def isCDot? : Syntax → Bool
+  | .node _ ``cdotTk #[.node _ `patternIgnore #[.node _ _ #[.atom _ v]]] => v == "·"
+  | .node _ ``Lean.Parser.Term.cdot #[.atom _ v] => v == "·"
+  | _ => false
+
+/--
+`findDot stx` extracts from `stx` the syntax nodes of `kind` `Lean.Parser.Term.cdot` or `cdotTk`. -/
+partial
+def findDot : Syntax → Array Syntax
```
grunweg, line 188: Can you call this something with "cdot", please?
(I'm unsure how to capitalise names with cdot in it. I may give that bikeshed more thought after lunch.)

test/Lint.lean
```diff
@@ -57,3 +57,23 @@ set_option linter.dupNamespace true in
...
+
+set_option linter.cdot false in
+/--
+warning: Please, use '·' (typed as `\·`) instead of '.' as 'cdot'.
+note: this linter can be disabled with `set_option linter.cdot false`
+---
+warning: Please, use '·' (typed as `\·`) instead of '.' as 'cdot'.
+note: this linter can be disabled with `set_option linter.cdot false`
+---
+warning: Please, use '·' (typed as `\·`) instead of '.' as 'cdot'.
```
grunweg, line 69: Would you like to test a second kind of wrong dot as well, e.g. change of the wrong dots? (I was wondering if the wrong dot was hardcoded - the answer is "clearly not", but a test could be nice.)

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -158,4 +158,66 @@ initialize addLinter missingEndLinter
 
 end MissingEnd
 
+/-!
```
grunweg, line 161: This is starting a bikeshed, but: intuitively, I would rather add this to `Linters/Style`. (This linter is purely about coding style, and not at all relevant to correctness.) What do you think?

I agree this starts a broader conversation how to structure the `Linters` directory: would we rather have separate files for individual linters, or are we fine grouping linters if each is reasonably short? (In any case, having an overview of all linters in some doc-string would be nice --- but that's definitely out of scope for this PR.)

grunweg (commented):
Looks great in general! Here's a first round of small comments: I'll eat now, and will return to this afterwards.

Mathlib/Tactic/Linter/Lint.lean
adomani: Renamed it to `unwanted_cdot`.

test/Lint.lean
adomani: I added another test, with `(. + ·)`, swapping the positions of `·` and `.`.

Mathlib/Tactic/Linter/Lint.lean
adomani: Renamed to `findCDot`.  My instinct would be to use the same capitalization for `C` and `D`, like `ZMod`, `EReal`, `PUnit`,...

Mathlib/Tactic/Linter/Lint.lean
adomani: I agree that this is more "stylistic" than "correctness", but the other linters also do not really check correctness.

Anyway, I am happy if this linter goes somewhere else, but, I also agree that this is a discussion that should happen elsewhere!

Mathlib/Tactic/Linter/Lint.lean
grunweg: Works for me!

Mathlib/Tactic/Linter/Lint.lean
grunweg: So I'll start a zulip discussion/propose a re-organisation in a PR? Sounds good to me.

test/Lint.lean
grunweg: Thanks! I had meant something else: are there other characters which are treated like a cdot, but are also bad.
Reading `Init/NotationExtra.lean` in core, the current answer is no - so this is fine as-is.

grunweg (commented):
Thanks for writing this! I just pushed a commit with tiny style-only tweaks directly.
Looks good to me now.
maintainer merge

PR #15008, opened 2024-07-22
[Merged by Bors] - chore: tidy various files

Ruben-VandeVelde:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #13791, opened 2024-06-13
refactor: Primrec and Partrec

digama0:
General cleanup of the `Primrec` and `Partrec` files, to better adjust to lean 4 things. The main user-visible change is that `Primrec₂` is no longer a `def` but an `abbrev`, because it was causing inference issues in lean 4. I also removed all the nonterminal `simp`s in `PartrecCode.lean`.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15011, opened 2024-07-22
[Merged by Bors] - style: remove more trailing semicolons

grunweg:
---

- [ ] depends on: #15008
- [ ] depends on: #15015


Mathlib/Analysis/SpecialFunctions/Integrals.lean
```diff
@@ -255,7 +255,7 @@ theorem intervalIntegrable_inv_one_add_sq :
 /-! ### Integrals of the form `c * ∫ x in a..b, f (c * x + d)` -/
 
 
--- Porting note (#10618): was @[simp];
+-- Porting note (#10618): was @[simp]
 -- simpNF says LHS does not simplify when applying lemma on itself
```
j-loreaux, line 259: Did you check that this is still applicable?

grunweg: I had not. I just pushed a commit adding back @[simp]: at least #lint passes; let's see if CI has anything else to say.

grunweg: (If successful, that change will be a separate PR.)

grunweg: And linting passed :tada: preparing a PR now

j-loreaux: @grunweg did you know you can run `#lint` right after the decl? It's not a global test, but it would probably have told you whether this particular issue was still a problem or not.

grunweg: Yes, I do, thanks for asking!
#15015 re-simps this lemmas and a few more

Mathlib/Computability/Partrec.lean
```diff
@@ -546,8 +546,8 @@ theorem bind_decode_iff {f : α → β → Option σ} :
             snd).bind
         (Computable.comp hf fst).to₂.partrec₂)
       fun n => by
-        simp; cases decode (α := α) n.unpair.1 <;> simp;
-          cases decode (α := β) n.unpair.2 <;> simp,
+        simp; cases decode (α := α) n.unpair.1 <;> simp
```
eric-wieser, line 549: could clear up another semicolon here at the same time:
```suggestion
        simp
        cases decode (α := α) n.unpair.1 <;> simp
```

eric-wieser (approved):
bors d+

Mathlib/Computability/Partrec.lean
eric-wieser: (which is to say; I think this is deliberately a long wrapped line of semicolons, and the fix is to remove all the semicolons, not just the one at EOL)

grunweg: Good point, fixed!

grunweg:
Thank you for the quick review.
bors merge

grunweg:
Now the style linter recognises a simp as non-terminal: I just squeezed it; #13791 will refactor that file anyway.

j-loreaux:
bors merge

PR #15012, opened 2024-07-22
[Merged by Bors] - feat: (Analysis.InnerProductSpace) Decomposition of Finite-Dimensional Inner Product Space as Direct Sum of Simultaneous Eigenspaces of Commuting Symmetric Operators

JonBannon:
Introduced `Mathlib.Analysis.InnerProductSpace.Simultaneous.lean` as a receptacle for results concerning simultaneous eigenspace decomposition results for tuples of commuting symmetric operators on (finite-dimensional) Hilbert space. 

Includes decomposition of finite-dimensional Hilbert space as internal direct sum of simultaneous eigenspaces of a commuting pair of symmetric operators.

Co-Authored by: Jack Cheverton <jt18chev@siena.edu>

Co-Authored by: Samyak Dhar Tuladhar <sd10tula@siena.edu>

---


Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+-/
+
+import Mathlib.Analysis.InnerProductSpace.Spectrum
+import Mathlib.Analysis.InnerProductSpace.Projection
+import Mathlib.Order.CompleteLattice
+
+/-! # Simultaneous eigenspaces of a commuting pair of symmetric operators
+
+This file collects various decomposition results for simultaneous eigenspaces of a commuting pair
+of symmetric operators on a finite-dimensional Hilbert space.
```
j-loreaux, line 14: I know they're the same, but I'm suggesting this change because we don't actually assume `[CompleteSpace E]` here.
```suggestion
of symmetric operators on a finite-dimensional inner product space.
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+
+# Main Result
+
+* `DirectSum.IsInternal_of_simultaneous_eigenspaces_of_commuting_symmetric_pair` establishes that
```
j-loreaux, line 18: I think this name is too long otherwise. Note also, we're beginning to use dot notation even if there are no arguments with that type as input, because we can still use it with the *expected type* dot notation.
```suggestion
* `DirectSum.IsInternal.eigenspaces_of_commuting_symmetric_pair` establishes that
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+   if `{A B : E →ₗ[𝕜] E}`, then `IsSymmetric A`, `IsSymmetric B` and `A ∘ₗ B = B ∘ₗ A` imply that
+   `E` decomposes as an internal direct sum of the pairwise orthogonal spaces
+   `eigenspace B μ ⊓ eigenspace A ν`
+
+## TODO
+
+Develop `Diagonalization` class that selects a basis of (simultaneous) eigenvectors in a
+principled way.
+
```
j-loreaux, line 27: This is not detailed enough to explain what you mean. It sounds like you just want to use `Classical.choice` to pick a basis, but I think that's not what you intend.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+## Tags
+
+self-adjoint operator, simultaneous eigenspaces, simultaneous diagonalization
+
+-/
+
+variable {𝕜 : Type*} [RCLike 𝕜]
+variable {E : Type*} [NormedAddCommGroup E] [InnerProductSpace 𝕜 E] [FiniteDimensional 𝕜 E]
+
+local notation "⟪" x ", " y "⟫" => @inner 𝕜 E _ x y
```
j-loreaux, line 37: You should just be able to use the notation `⟪x, y⟫_𝕜` throughout without needing to introduce local notation. If you copied this from other files, it's probably just a vestige from before we had the notation I mentioned working reliably for `RCLike 𝕜`.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+open Module.End
+
+namespace LinearMap
+
+namespace IsSymmetric
+
+section Pair
+
+variable {α : 𝕜} {A B : E →ₗ[𝕜] E} (hA : A.IsSymmetric) (hB : B.IsSymmetric)
+    (hAB : A ∘ₗ B = B ∘ₗ A)
```
j-loreaux, line 48: This should be defeq, but I'm not sure it's actually better, so I'm not necessarily advocating for it yet. Mainly I'm putting this here as a reminder to myself that it's something to consider.
```suggestion
    (hAB : Commute A B)
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+
+theorem eigenspace_invariant (α : 𝕜) : ∀ v ∈ (eigenspace A α), (B v ∈ eigenspace A α) := by
```
j-loreaux, line 50: The name makes it seem like the eigenspace for `A` is invariant for `A`, which is obviously true, but it's not the statement here. The name needs to convey that it's for the other operator. Maybe `eigenspace_invariant_of_commute` (especially with the `Commute A B` suggestion above).

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+  intro v hv
+  rw [eigenspace, mem_ker, sub_apply, Module.algebraMap_end_apply, ← comp_apply A B v, hAB,
+  comp_apply B A v, ← map_smul, ← map_sub, hv, map_zero] at *
+
+theorem invariant_subspace_inf_eigenspace_eq_restrict {F : Submodule 𝕜 E} (S : E →ₗ[𝕜] E)
+    (μ : 𝕜) (hInv : ∀ v ∈ F, S v ∈ F) : (eigenspace S μ) ⊓ F =
+    Submodule.map (Submodule.subtype F)
+    (eigenspace (S.restrict (hInv)) μ) := by
```
j-loreaux, line 58: Especially since the statement of this theorem is hard to parse, a docstring which explains the statement in human-friendly terms would be nice.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+      simp only [← hy.2, Submodule.coeSubtype, SetLike.coe_mem]
+
+theorem invariant_subspace_inf_eigenspace_eq_restrict' : (fun (γ : 𝕜) ↦
+    Submodule.map (Submodule.subtype (eigenspace A α)) (eigenspace (B.restrict
+    (eigenspace_invariant hAB α)) γ)) = (fun (γ : 𝕜) ↦ (eigenspace B γ ⊓ eigenspace A α)) := by
+  funext γ
+  exact Eq.symm (invariant_subspace_inf_eigenspace_eq_restrict B γ (eigenspace_invariant hAB α))
+
+theorem iSup_restrict_eq_top: (⨆ γ , (eigenspace (LinearMap.restrict B
+    (eigenspace_invariant hAB α)) γ)) = ⊤ := by
```
j-loreaux, line 84: docstrings are nice. :smile:

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+    rw [← Submodule.orthogonal_eq_bot_iff]
+    exact orthogonalComplement_iSup_eigenspaces_eq_bot (LinearMap.IsSymmetric.restrict_invariant hB
+    (eigenspace_invariant hAB α))
+
+theorem iSup_simultaneous_eigenspaces_eq_top :
+    (⨆ (α : 𝕜), (⨆ (γ : 𝕜), (eigenspace B γ ⊓ eigenspace A α))) = ⊤ := by
```
j-loreaux, line 90: This is half of your main result, it should definitely get a docstring. Also, I understand the temptation to use the word `simultaneous`, but I would use `inf` because it matches the statement better.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,123 @@
...
+  have : (fun (α : 𝕜) ↦  eigenspace A α)  = fun (α : 𝕜) ↦
+      (⨆ (γ : 𝕜), (eigenspace B γ ⊓ eigenspace A α)) := by
+    funext; rw [← invariant_subspace_inf_eigenspace_eq_restrict' hAB,
+       ← Submodule.map_iSup, iSup_restrict_eq_top hB hAB, Submodule.map_top,
+       Submodule.range_subtype]
+  rw [← Submodule.orthogonal_eq_bot_iff.mp (hA.orthogonalComplement_iSup_eigenspaces_eq_bot), this]
+
+theorem orthogonality_of_simultaneous_eigenspaces_of_pairwise_commuting_symmetric :
+    OrthogonalFamily 𝕜 (fun (i : 𝕜 × 𝕜) => (eigenspace B i.1 ⊓ eigenspace A i.2 : Submodule 𝕜 E))
+    (fun i => (eigenspace B i.1 ⊓ eigenspace A i.2).subtypeₗᵢ) := by
```
j-loreaux, line 100: This is the other half of your main result, it should definitely get a docstring. Same comment about `simultaneous` versus `inf`, and also I don't think you mean `pairwise`.

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
```diff
@@ -250,7 +250,6 @@ end Version2
 end IsSymmetric
 
 end LinearMap
-
 section Nonneg
```
j-loreaux, line 253: Let's not introduce random changes to the diff.

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
```diff
@@ -280,4 +279,5 @@ theorem eigenvalue_pos_of_pos {μ : ℝ} {T : E →ₗ[𝕜] E} (hμ : HasEigenv
     exact mod_cast this
   exact (mul_pos_iff_of_pos_right hpos).mp (this ▸ hnn v)
 
+
 end Nonneg
```
j-loreaux, line 283: ditto

j-loreaux (commented):
This is a first pass.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: When I omitted it before, something broke. Let me check again...

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Maybe I remove this, as I don't really know what I intend, either. I think you have an idea of what this is, though...so I will go and look at your comments and maybe update this accordingly. I don't want to forget what you suggested, and so wanted a reminder here...

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Point taken!

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Oh.. I DEFINITELY don't mean pairwise...

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
JonBannon: Can I get rid of this?

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
JonBannon: How can I remove these?

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Here's an attempt.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: I realize that I'm shying away from this because it's hard for some of the technical results. All the more reason to do it!!

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
j-loreaux: Note the `_𝕜` in my notation.

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
j-loreaux: Put the blank line back in.

Mathlib/Analysis/InnerProductSpace/Spectrum.lean
j-loreaux: remove the blank line

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Yes...somehow this doesn't work. I was wondering why that was when we originally wrote this, because checking...these are, I think, definitionally equal.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Need to sort the commute issue...

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: I stole something from a communication we had earlier and put it in place of this comment as a reminder. If it is still unclear, I can remove the to do altogether.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Everything works now, so I can resolve this.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+  rw [eigenspace, mem_ker, sub_apply, Module.algebraMap_end_apply, ← comp_apply A B v, hAB,
+  comp_apply B A v, ← map_smul, ← map_sub, hv, map_zero] at *
+
+/--The inf of an eigenspace of an operator with another invariant subspace
+agrees with the corresponding eigenspace of the restriction of that operator to the
+invariant subspace-/
+theorem invariant_subspace_inf_eigenspace_eq_restrict {F : Submodule 𝕜 E} (S : E →ₗ[𝕜] E)
+    (μ : 𝕜) (hInv : ∀ v ∈ F, S v ∈ F) : (eigenspace S μ) ⊓ F =
+    Submodule.map (Submodule.subtype F)
+    (eigenspace (S.restrict (hInv)) μ) := by
```
j-loreaux, line 61: You can get this for free from [Submodule.inf_genEigenspace](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/Eigenspace/Basic.html#Submodule.inf_genEigenspace) if you first prove that, for symmetric operators, (a) the restriction to an invariant subspace is symmetric (which we may already have), and (b) the generalized eigenspace is the eigenspace.

We want both of those, so you might as well find / add them. That way we don't duplicate the proof.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+      simpa [mem_eigenspace_iff]
+    · simp only [Submodule.coeSubtype] at h
+      obtain ⟨_, hy⟩ := h
+      simp only [← hy.2, Submodule.coeSubtype, SetLike.coe_mem]
+
+/--Auxiliary: function version of `invariant_subspace_inf_eigenspace_eq_restrict` -/
+theorem invariant_subspace_inf_eigenspace_eq_restrict' : (fun (γ : 𝕜) ↦
+    Submodule.map (Submodule.subtype (eigenspace A α)) (eigenspace (B.restrict
+    (eigenspace_invariant_of_commute hAB α)) γ))
+    = (fun (γ : 𝕜) ↦ (eigenspace B γ ⊓ eigenspace A α)) := by
```
j-loreaux, line 84: Why don't you just have `(γ : 𝕜)` as an argument, instead of viewing this as a function equality?

Also, I would call this `eigenspace_restrict_eigenspace_eq_inf`, or something like that.
```suggestion
theorem invariant_subspace_inf_eigenspace_eq_restrict' : (fun (γ : 𝕜) ↦
    Submodule.map (Submodule.subtype (eigenspace A α)) (eigenspace (B.restrict
    (eigenspace_invariant_of_commute hAB α)) γ))
    = (fun (γ : 𝕜) ↦ (eigenspace B γ ⊓ eigenspace A α)) := by
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+  funext γ
+  exact Eq.symm (invariant_subspace_inf_eigenspace_eq_restrict B γ
+    (eigenspace_invariant_of_commute hAB α))
+
+/-- If A and B are commuting symmetric operators on a finite dimensional inner product space
+then the eigenspaces of the restriction of B to any eigenspace of A exhaust that eigenspace.-/
+theorem iSup_inf_eq_top: (⨆ γ , (eigenspace (LinearMap.restrict B
+    (eigenspace_invariant_of_commute hAB α)) γ)) = ⊤ := by
```
j-loreaux, line 92: In my mind, the more natural statement here would be `⨆ γ, eigenspace B γ ⊓ eigenspace A α = eigenspace A α`. Is there are particular reason that statemnet is bad for your purposes?

In either case, the naming is a problem because `eigenspace` is not mentioned.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+    rw [← Submodule.orthogonal_eq_bot_iff]
+    exact orthogonalComplement_iSup_eigenspaces_eq_bot (LinearMap.IsSymmetric.restrict_invariant hB
+    (eigenspace_invariant_of_commute hAB α))
+
+/--If A and B are commuting symmetric operators acting on a finite dimensional inner product space,
+then the simultaneous eigenspaces of A and B exhaust the space. -/
+theorem iSup_simultaneous_eigenspaces_eq_top :
+    (⨆ (α : 𝕜), (⨆ (γ : 𝕜), (eigenspace B γ ⊓ eigenspace A α))) = ⊤ := by
```
j-loreaux, line 100: possibly with `_eq_top` appended.
```suggestion
theorem iSup_iSup_eigenspace_inf_eigenspace :
    (⨆ (α : 𝕜), (⨆ (γ : 𝕜), (eigenspace B γ ⊓ eigenspace A α))) = ⊤ := by
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+  have : (fun (α : 𝕜) ↦  eigenspace A α)  = fun (α : 𝕜) ↦
+      (⨆ (γ : 𝕜), (eigenspace B γ ⊓ eigenspace A α)) := by
+    funext γ
+    rw [← invariant_subspace_inf_eigenspace_eq_restrict' hAB, ← Submodule.map_iSup,
+      iSup_inf_eq_top hB hAB, Submodule.map_top, Submodule.range_subtype]
+  rw [← Submodule.orthogonal_eq_bot_iff.mp (hA.orthogonalComplement_iSup_eigenspaces_eq_bot), this]
+
+/--The simultaneous eigenspaces of a pair of commuting symmetric operators form an
+`OrthogonalFamily`.-/
+theorem orthogonality_of_simultaneous_eigenspaces_of_commuting_symmetric_pair :
```
j-loreaux, line 110: This matches the statement better.
```suggestion
theorem orthogonalFamily_eigenspace_inf_eigenspace :
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,135 @@
...
+  have H:=  (Iff.not (Iff.symm Prod.ext_iff)).mpr hij
+  push_neg at H
+  by_cases C: i.1 = j.1
+  <;> intro w ⟨hw1, hw2⟩
+  · exact orthogonalFamily_iff_pairwise.mp hA.orthogonalFamily_eigenspaces (H C) hv2 w hw2
+  · exact orthogonalFamily_iff_pairwise.mp hB.orthogonalFamily_eigenspaces C hv1 w hw1
+
+/-- Given a commuting pair of symmetric linear operators on a finite dimensional inner product space
+, the space decomposes as an internal direct sum of simultaneous eigenspaces of these operators. -/
+theorem DirectSum.IsInternal_of_simultaneous_eigenspaces_of_commuting_symmetric_pair:
```
j-loreaux, line 124: Or possibly: `DirectSum.IsInternal.simultaneous_eigenspace_of_commute`
```suggestion
theorem DirectSum.IsInternal.eigenspace_inf_eigenspace :
```

j-loreaux (commented):
After you address these, I'll have another look where I really dig into the proofs myself.

j-loreaux:
Note: there is also [Module.End.iSup_genEigenspace_restrict_eq_top](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/Eigenspace/Triangularizable.html#Module.End.iSup_genEigenspace_restrict_eq_top), so really you want the result I mentioned about generalized eigenspaces of symmetric operators.

JonBannon:
Great! Thanks for these. Will work on them.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Fantastic! This is good to know.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: For some reason, yes. When we originally tried this (we used the natural statement you mentioned) there were some issues in resolving the `\top` to that eigenspace. Things worked with the \top, though. Maybe this is another situation about transparency?

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+and a proof obligation that the basis vectors are eigenvectors.
+
+## Tags
+
+self-adjoint operator, simultaneous eigenspaces, simultaneous diagonalization
+
+-/
+
+variable {𝕜 : Type*} [RCLike 𝕜]
+variable {E : Type*} [NormedAddCommGroup E] [InnerProductSpace 𝕜 E] [FiniteDimensional 𝕜 E]
```
j-loreaux, line 36: Normally we put all the type variables first
```suggestion
variable {𝕜 E : Type*} [RCLike 𝕜]
variable [NormedAddCommGroup E] [InnerProductSpace 𝕜 E] [FiniteDimensional 𝕜 E]
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+
+variable {α : 𝕜} {A B : E →ₗ[𝕜] E} (hA : A.IsSymmetric) (hB : B.IsSymmetric)
+    (hAB : A ∘ₗ B = B ∘ₗ A)
+
+/--If a pair of operators commute, then the eigenspaces of one are invariant under the other.-/
+theorem eigenspace_invariant_of_commute (α : 𝕜) :
+    ∀ v ∈ (eigenspace A α), (B v ∈ eigenspace A α) := by
+  intro v hv
+  rw [eigenspace, mem_ker, sub_apply, Module.algebraMap_end_apply, ← comp_apply A B v, hAB,
+  comp_apply B A v, ← map_smul, ← map_sub, hv, map_zero] at *
```
j-loreaux, line 54: style
```suggestion
    comp_apply B A v, ← map_smul, ← map_sub, hv, map_zero] at *
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+/--The inf of an eigenspace of an operator with another invariant subspace
+agrees with the corresponding eigenspace of the restriction of that operator to the
+invariant subspace-/
+theorem eigenspace_restrict_eigenspace_eq_inf : (fun (γ : 𝕜) ↦
+    Submodule.map (Submodule.subtype (eigenspace A α)) (eigenspace (B.restrict
+    (eigenspace_invariant_of_commute hAB α)) γ))
+    = (fun (γ : 𝕜) ↦ (eigenspace A α ⊓ eigenspace B γ)) := by
+  funext γ
+  exact Eq.symm (Submodule.inf_genEigenspace B (eigenspace A α)
+    (eigenspace_invariant_of_commute hAB α) (μ := γ) (k := 1))
```
j-loreaux, line 65: I don't think you want the function version. Just pass `γ` as a parameter, like below. Note that I have also reversed the direction of the equality, and we can simplify the name that way.
```
open Submodule in
/-- The intersection of eigenspaces of commuting selfadjoint operators is equal to the eigenspace of
one operator restricted to the eigenspace of the other, which is an invariant subspace because the
operators commute. -/
theorem eigenspace_inf_eigenspace (α γ : 𝕜) :
    eigenspace A α ⊓ eigenspace B γ = map (Submodule.subtype (eigenspace A α))
      (eigenspace (B.restrict (eigenspace_invariant_of_commute hAB α)) γ) :=
  (eigenspace A α).inf_genEigenspace _ _ (k := 1)
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+
+/-- If A and B are commuting symmetric operators on a finite dimensional inner product space
+then the eigenspaces of the restriction of B to any eigenspace of A exhaust that eigenspace.-/
+theorem iSup_inf_eq_top: (⨆ γ , (eigenspace (LinearMap.restrict B
+    (eigenspace_invariant_of_commute hAB α)) γ)) = ⊤ := by
+    rw [← Submodule.orthogonal_eq_bot_iff]
+    exact orthogonalComplement_iSup_eigenspaces_eq_bot (LinearMap.IsSymmetric.restrict_invariant hB
+    (eigenspace_invariant_of_commute hAB α))
```
j-loreaux, line 73: As I mentioned before, this name doesn't work because it doesn't mention `eigenspace` anywhere. But additionally, it's nicer if you *don't* work in the subtype, but that means you need to do slightly more work in the proof.
```suggestion
then the supremum of intersection of eigenspaces is precisely the non-indexed eigenspace. -/
theorem iSup_eigenspace_inf_eigenspace :
    (⨆ γ, eigenspace A α ⊓ eigenspace B γ) = eigenspace A α := by
  conv_rhs => rw [← (eigenspace A α).map_subtype_top]
  simp only [eigenspace_inf_eigenspace hAB, ← Submodule.map_iSup]
  congr 1
  rw [← Submodule.orthogonal_eq_bot_iff]
  exact orthogonalComplement_iSup_eigenspaces_eq_bot <|
    hB.restrict_invariant <| eigenspace_invariant_of_commute hAB α
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+/--If A and B are commuting symmetric operators acting on a finite dimensional inner product space,
+then the simultaneous eigenspaces of A and B exhaust the space. -/
+theorem iSup_iSup_eigenspace_inf_eigenspace_eq_top :
+    (⨆ (α : 𝕜), (⨆ (γ : 𝕜), (eigenspace A α ⊓ eigenspace B γ))) = ⊤ := by
+  have : (fun (α : 𝕜) ↦  eigenspace A α)  = fun (α : 𝕜) ↦
+      (⨆ (γ : 𝕜), (eigenspace A α ⊓ eigenspace B γ)) := by
+    funext γ
+    rw [← eigenspace_restrict_eigenspace_eq_inf hAB, ← Submodule.map_iSup,
+      iSup_inf_eq_top hB hAB, Submodule.map_top, Submodule.range_subtype]
+  rw [← Submodule.orthogonal_eq_bot_iff.mp (hA.orthogonalComplement_iSup_eigenspaces_eq_bot), this]
```
j-loreaux, line 84: see, you really did want to pass `γ` as a parameter in the earlier version. Does this make sense?
```suggestion
/-- If A and B are commuting symmetric operators acting on a finite dimensional inner product space,
then the simultaneous eigenspaces of A and B exhaust the space. -/
theorem iSup_iSup_eigenspace_inf_eigenspace_eq_top :
    (⨆ α, ⨆ γ, eigenspace A α ⊓ eigenspace B γ) = ⊤ := by
  simpa [iSup_eigenspace_inf_eigenspace hB hAB] using
    Submodule.orthogonal_eq_bot_iff.mp <| hA.orthogonalComplement_iSup_eigenspaces_eq_bot
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+    OrthogonalFamily 𝕜 (fun (i : 𝕜 × 𝕜) => (eigenspace A i.2 ⊓ eigenspace B i.1 : Submodule 𝕜 E))
+    (fun i => (eigenspace A i.2 ⊓ eigenspace B i.1).subtypeₗᵢ) := by
+  refine orthogonalFamily_iff_pairwise.mpr ?_
+  intro i j hij v ⟨hv1 , hv2⟩
+  have H:=  (Iff.not (Iff.symm Prod.ext_iff)).mpr hij
+  push_neg at H
+  by_cases C: i.1 = j.1
+  <;> intro w ⟨hw1, hw2⟩
+  · exact orthogonalFamily_iff_pairwise.mp hA.orthogonalFamily_eigenspaces (H C) hv1 w hw1
+  · exact orthogonalFamily_iff_pairwise.mp hB.orthogonalFamily_eigenspaces C hv2 w hw2
```
j-loreaux, line 98: cleanup. I think this is slightly nicer because it's more symmetric.
```suggestion
    OrthogonalFamily 𝕜 (fun i : 𝕜 × 𝕜 ↦ (eigenspace A i.2 ⊓ eigenspace B i.1 : Submodule 𝕜 E))
    (fun i => (eigenspace A i.2 ⊓ eigenspace B i.1).subtypeₗᵢ) :=
  OrthogonalFamily.of_pairwise fun i j hij v ⟨hv1 , hv2⟩ ↦ by
    obtain (h₁ | h₂) : i.1 ≠ j.1 ∨ i.2 ≠ j.2 := by rwa [Ne.eq_def, Prod.ext_iff, not_and_or] at hij
    all_goals intro w ⟨hw1, hw2⟩
    · exact hB.orthogonalFamily_eigenspaces.pairwise h₁ hv2 w hw2
    · exact hA.orthogonalFamily_eigenspaces.pairwise h₂ hv1 w hw1
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,113 @@
...
+
+/-- Given a commuting pair of symmetric linear operators on a finite dimensional inner product space
+, the space decomposes as an internal direct sum of simultaneous eigenspaces of these operators. -/
+theorem DirectSum.IsInternal_of_simultaneous_eigenspaces_of_commuting_symmetric_pair:
+    DirectSum.IsInternal (fun (i : 𝕜 × 𝕜) ↦ (eigenspace A i.2 ⊓ eigenspace B i.1)):= by
+  apply (OrthogonalFamily.isInternal_iff
+    (orthogonalFamily_eigenspace_inf_eigenspace hA hB)).mpr
```
j-loreaux, line 105: hanging `,` in the docstring, and use of dot notation in the proof. Also, note that the previous name was overly namespaced, and didn't give you anything you would want (like dot notation for `DirectSum.IsInternal`. I think this name is better.
```suggestion
/-- Given a commuting pair of symmetric linear operators on a finite dimensional inner product
space, the space decomposes as an internal direct sum of simultaneous eigenspaces of these
operators. -/
theorem directSum_isInteral_of_commute :
    DirectSum.IsInternal (fun (i : 𝕜 × 𝕜) ↦ (eigenspace A i.2 ⊓ eigenspace B i.1)):= by
  apply (orthogonalFamily_eigenspace_inf_eigenspace hA hB).isInternal_iff.mpr
```

j-loreaux (commented):
After these changes I think it will be ready for the maintainer queue.

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: I wonder if I can shorten this further using the generalized eigenspace result. I tried a little but wasn't successful. One silly question: why did you specify the depth for `congr` here? I wouldn't have thought to do this.

j-loreaux: I don't see an obvious way to use generalized eigenspaces further here.

When `congr` or `convert` don't do what I expect / want (or produce a million goals), it's generally because they have gone too deep. So, if it doesn't do exactly what I think it should, I immediately go for limiting depth. Sometimes this makes me realize I was being stupid, and I *shouldn't* have expected them to work. Other times, limiting the depth is exactly what's needed.

JonBannon:
👍

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,108 @@
...
+import Mathlib.LinearAlgebra.Eigenspace.Basic
+
+/-! # Simultaneous eigenspaces of a commuting pair of symmetric operators
+
+This file collects various decomposition results for simultaneous eigenspaces of a commuting pair
+of symmetric operators on a finite-dimensional inner product space.
+
+# Main Result
+
+* `DirectSum.IsInternal.eigenspaces_of_commuting_symmetric_pair` establishes that
```
j-loreaux, line 19: This needs to be updated to reflect the new name.
```suggestion
* `LinearMap.IsSymmetric.directSum_isInternal_of_commute` establishes that
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
```diff
@@ -0,0 +1,108 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jon Bannon, Jack Cheverton, Samyak Dhar Tuladhar
+-/
+
+import Mathlib.Analysis.InnerProductSpace.Spectrum
+import Mathlib.Analysis.InnerProductSpace.Projection
+import Mathlib.Order.CompleteLattice
+import Mathlib.LinearAlgebra.Eigenspace.Basic
+
+/-! # Simultaneous eigenspaces of a commuting pair of symmetric operators
```
j-loreaux, line 12: "Simultaneous", while I've heard the term used before, I think only makes sense in the context of *diagonalization*, and, strictly speaking, there is no diagonalization here. (in the sense that we don't get a collection of eigenvalues and eigenvectors.). Also, don't focus on pairs here, since you want to add the version for finite collections of commuting operators shortly.
```suggestion
/-! # Joint eigenspaces of commuting symmetric operators
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
j-loreaux, line 1: I think we should change the name to `JointEigenspace` for the reasons outlined above.

Mathlib.lean
```diff
@@ -1081,6 +1081,7 @@ import Mathlib.Analysis.InnerProductSpace.Positive
 import Mathlib.Analysis.InnerProductSpace.ProdL2
 import Mathlib.Analysis.InnerProductSpace.Projection
 import Mathlib.Analysis.InnerProductSpace.Rayleigh
+import Mathlib.Analysis.InnerProductSpace.Simultaneous
```
j-loreaux, line 1084:
```suggestion
import Mathlib.Analysis.InnerProductSpace.JointEigenspace
```

Mathlib/Analysis/InnerProductSpace/Simultaneous.lean
JonBannon: Oops!!

j-loreaux:
I've had a lot of input in this.

maintainer merge

Mathlib/Analysis/InnerProductSpace/JointEigenspace.lean
```diff
@@ -0,0 +1,108 @@
...
+namespace IsSymmetric
+
+section Pair
+
+variable {α : 𝕜} {A B : E →ₗ[𝕜] E} (hA : A.IsSymmetric) (hB : B.IsSymmetric)
+    (hAB : A ∘ₗ B = B ∘ₗ A)
+
+/--If a pair of operators commute, then the eigenspaces of one are invariant under the other.-/
+theorem eigenspace_invariant_of_commute (α : 𝕜) :
+    ∀ v ∈ (eigenspace A α), (B v ∈ eigenspace A α) := by
```
dupuisf, line 51:
```suggestion
theorem eigenspace_invariant_of_commute (α : 𝕜) :
    ∀ v ∈ eigenspace A α, B v ∈ eigenspace A α := by
```

dupuisf: Also, having explicit arguments in a variable command somewhere above (i.e. `hA`, `hB` and `hAB` in this case) is generally discouraged, because then we can't tell how to actually use the lemma just by looking at the lemma statement.

dupuisf (commented):
Can you merge master into this? The variable inclusion mechanism has changed in the latest version of Lean and might lead to breakage here, so it's easier to do this before trying to merge.

j-loreaux:
Good catch, thanks

dupuisf:
Great; thanks!

bors r+

PR #15017, opened 2024-07-22
[Merged by Bors] - feat(lint-style): rewrite focusing dot check in Lean

grunweg:
as a syntax linter, extending `adomani`'s `cdot` syntax linter.

---


adomani:
Thanks for this!

How about extending the `cdot` linter, though?  Here is a sample of how to do that:
```lean
import Mathlib.Tactic.Linter

open Lean in

elab "line_breaks? " cmd:command : command => do
  Elab.Command.elabCommand cmd
  for cdot in Mathlib.Linter.findCDot cmd do
    match cdot.find? (·.isOfKind `token.«· ») with
      | some (.node _ _ #[.atom (.original _ _ afterCDot _) _]) =>
        if afterCDot.contains '\n' && afterCDot.trimLeft.isEmpty then
          logWarningAt cdot m!"Found an isolated `·`!"
      | _ => return

line_breaks?
--inspect
example : True ∧ True := by
  constructor
--v this `·` is flagged
  ·
    exact .intro
  · -- if I write a long comment here
    -- there is no warning
    exact .intro
```

grunweg:
@adomani Good idea, that is even nicer. I just adapted the linter accordingly; it should be ready for review again.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -159,12 +159,14 @@ end MissingEnd
 The `cdot` linter is a syntax-linter that flags uses of the "cdot" `·` that are achieved
 by typing a character different from `·`.
 For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter.
+It also flags "isolated cdots", i.e. when the `⬝` is on its own line.
```
joneugster, line 162:
```suggestion
It also flags "isolated cdots", i.e. when the `·` is on its own line.
```
I believe you used `⬝` (`\cdot`) instead of  `·` (`\.` or `\centerdot`). Maybe the `"cdot"` above is confusing and should be adjusted...

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -159,12 +159,14 @@ end MissingEnd
...
 -/
 
 /--
 The `cdot` linter flags uses of the "cdot" `·` that are achieved by typing a character
 different from `·`.
-For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter. -/
+For instance, a "plain" dot `.` is allowed syntax, but is flagged by the linter.
+It also flags "isolated cdots", i.e. when the `⬝` is on its own line. -/
```
joneugster, line 169:
```suggestion
It also flags "isolated cdots", i.e. when the `·` is on its own line. -/
```

joneugster (commented):
I think this is ready to be merged, but I think it make sense to add a dependency on #15845, merge that first, and then make sure this linter is imported in `Mathlib.Init`

test/Lint.lean
```diff
@@ -210,24 +276,6 @@ example : ℕ → ℕ := set_option linter.style.lambdaSyntax true in λ _ ↦ 0
 
 end lambdaSyntaxLinter
 
-set_option linter.globalAttributeIn false in
```
grunweg, line 213: This fixes a merge error: the same test is already present on line 158.

test/Lint.lean
```diff
@@ -89,6 +91,70 @@ note: this linter can be disabled with `set_option linter.cdot false`
...
+
+#guard_msgs in
+set_option linter.cdot true in
+example : True := by
+  have : Nat := by
+    -- This is how code should look: no error.
+    · -- comment
+      exact 37
+  trivial
+
```
adomani, line 155:
```suggestion
```
Is this just a git merge gone wrong and it is simply a repetition of the tests above it?

grunweg: That was a merge error - thanks for spotting!

adomani (approved):
Thanks!

As I was involved in writing part of the code, I'll just approve of these changes, but it looks good!

test/Lint.lean
joneugster: Indeed, although its [line 127 on the resulting file](https://github.com/leanprover-community/mathlib4/blob/3c4d23a560483d82fe8ad39580429fcfb3bc5936/test/Lint.lean#L127-L143).

joneugster:
Thanks!

maintainer merge

Vierkantor (approved):
bors merge

jcommelin:
bors merge

PR #15024, opened 2024-07-22
chore(Algebra.Basic): scope and rename universal `Nat`-algebra and `Int`-algebra instances

mattrobball:
We rename `algebraNat` to `Nat.instAlgebraOfSemiring` and `algebraInt` to `Int.instAlgebraOfRing` to be consistent with existing naming conventions. We also scope them to avoid typeclass search tripping over them unnecessarily.

---

Moves:
- algebraNat -> Nat.instAlgebraOfSemiring
- algebraInt -> Int.instAlgebraOfRing

### DIFF
--- a/Archive/Wiedijk100Theorems/AbelRuffini.lean
+++ b/Archive/Wiedijk100Theorems/AbelRuffini.lean
@@ -31,7 +31,7 @@ namespace AbelRuffini
 
 open Function Polynomial Polynomial.Gal Ideal
 
-open scoped Polynomial
+open scoped Int Polynomial
 
 attribute [local instance] splits_ℚ_ℂ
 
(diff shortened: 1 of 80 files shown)


mattrobball:
!bench

eric-wieser:
I don't think these should be scoped. I also thing `Semiring.toNatAlgebra` might be a better name.

mattrobball:
We've already seen situations where  a search for `instClassExtendsSMul'` goes through intermediate chaining instances. With `Nat` or `Int` in scope, they get caught up. I think we need some mechanism to limit their scopes. 

I have a separate experiment where I used imports to do this. The advantage of this is specificity. 

Do we have much in the `Semiring/Ring` namespace?

eric-wieser:
`AddCommMonoid.natModule` is precedent for putting things in the `Semiring` namespace here. I'd argue that we should have a `to` prefix rather than `inst`, since this could be implemented as inheritance.

PR #15027, opened 2024-07-22
[Merged by Bors] - chore(lint-style.sh): remove duplicate check for case clashes

grunweg:
The check_case action (run earlier in the same workflow step) runs the same check across the entire repository.

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #15033, opened 2024-07-22
[Merged by Bors] - feat(AlgebraicGeometry): factorization of morphisms with affine target

chrisflav:
---


Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -677,6 +685,55 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
 
 end ZeroLocus
 
+section Factorization
+
+/-- `f : X ⟶ Spec A` factors via some `Spec B` as `f = g ≫ h` such that `Γ(g)` is
+injective and `Γ(h)` is surjective. -/
+lemma exists_factorization_of_spec {X : Scheme.{u}} {A : CommRingCat} (f : X ⟶ Spec A) :
```
erdOne, line 692: I think this deserves definitions instead of exists statements. (`A` is the scheme theoretic image, and we might want to glue them later)

chrisflav: I finally got to this and turned the exists statements into definitions.

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -712,6 +720,103 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
...
+section Factorization
+
+variable {X : Scheme.{u}} {A : CommRingCat}
+
+/-- If `X ⟶ Spec A` is a morphism of schemes, then `Spec` of `specTargetImage f` is the
+scheme-theoretic image of `f` and `f` factors as
+`specTargetImageFactorization f ≫ Spec.map (specTargetImageRingHom f)`
+(see `specTargetImageFactorization_comp`). -/
+def specTargetImage (f : X ⟶ Spec A) : CommRingCat :=
+  CommRingCat.of (A ⧸ (RingHom.ker <| (((ΓSpec.adjunction).homEquiv X (op A)).symm f).unop))
```
erdOne, line 732: I think the ideal could get its definition as well as it appears multiple times below.

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -712,6 +720,103 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
...
+  Spec.map (specTargetImageRingHom (f ≫ Y.isoSpec.hom)) ≫ Y.isoSpec.inv
+
+lemma affineTargetImageInclusion_app_surjective :
+    Function.Surjective <| (affineTargetImageInclusion f).app ⊤ := by
+  simp only [Scheme.comp_coeBase, Opens.map_comp_obj, Opens.map_top, Scheme.comp_app,
+    CommRingCat.coe_comp, affineTargetImageInclusion]
+  apply Function.Surjective.comp
+  · haveI : (toMorphismProperty (fun f ↦ Function.Surjective f)).RespectsIso := by
+      rw [← toMorphismProperty_respectsIso_iff]
+      exact surjective_respectsIso
```
erdOne, line 796: We have [`AlgebraicGeometry.instRespectsIsoSchemeSurjective`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/AlgebraicGeometry/Morphisms/UnderlyingMap.html#AlgebraicGeometry.instRespectsIsoSchemeSurjective)

Mathlib/AlgebraicGeometry/GammaSpecAdjunction.lean
```diff
@@ -467,6 +467,27 @@ theorem toOpen_unit_app_val_c_app' {X : Scheme.{u}} (U : Opens (PrimeSpectrum Γ
 
 end ΓSpec
 
+lemma toSpecΓ_ΓSpec_adjunction_homEquiv {X : Scheme.{u}} {B : CommRingCat} (φ : B ⟶ Γ(X, ⊤)) :
+    toSpecΓ B ≫ ((ΓSpec.adjunction.homEquiv X (op B)) φ.op).app ⊤ = φ := by
+  rw [AlgebraicGeometry.ΓSpec.adjunction_homEquiv]
```
erdOne, line 472: You should use `(Scheme.ΓSpecIso B).inv` instead of `toSpecΓ` (they are defeq) because the latter is in terms of locally ringed spaces. Maybe I should add a warning. And the proof becomes
```
example {X : Scheme.{u}} {B : CommRingCat} (φ : B ⟶ Γ(X, ⊤)) :
    (Scheme.ΓSpecIso B).inv ≫ ((ΓSpec.adjunction.homEquiv X (op B)) φ.op).app ⊤ = φ := by
  simp [Adjunction.homEquiv_apply]
  rw [ΓSpec.adjunction_unit_app_app_top]
  simp
```
Either you should squeeze it or maybe investigate why the simp lemma does not trigger.
Anyways in general you shouldn't unfold `ΓSpec.adjunction` at all and use the simp lemmas provided. It should be always possible as the morphism are uniquely identified by the adjunction properties.

Mathlib/AlgebraicGeometry/AffineScheme.lean
chrisflav: This does not seem to apply here, since here we are talking about the underlying ring map being surjective, not the map of topological spaces.

Mathlib/AlgebraicGeometry/GammaSpecAdjunction.lean
chrisflav: Ah I remember, you have told me before. We should just make all of the forbidden ones private.

erdOne:
LGTM (sorry forgot about it)
maintainer merge

joelriou:
Thanks!

bors merge

PR #15040, opened 2024-07-22
[Merged by Bors] - chore: update Batteries and deprecate LazyList

kim-em:
`LazyList` was deprecated in Batteries (with an indication to use `MLList` instead). The uses in Mathlib were trivial, as they were immediately forced to a `List` anyway. This PR deprecates the code about `LazyList`, and replaces the lazy-but-not-really code with strict versions.

### DIFF
--- a/Mathlib/Data/LazyList/Basic.lean
+++ b/Mathlib/Data/LazyList/Basic.lean
@@ -66,7 +66,8 @@ instance : LawfulTraversable LazyList := by
         Function.comp, Thunk.pure, ofList]
     · apply ih
 
-@[deprecated (since := "2024-07-22"), simp] theorem bind_singleton {α} (x : LazyList α) : x.bind singleton = x := by
+@[deprecated (since := "2024-07-22"), simp]
+theorem bind_singleton {α} (x : LazyList α) : x.bind singleton = x := by
   induction x using LazyList.rec (motive_2 := fun xs => xs.get.bind singleton = xs.get) with
   | nil => simp [LazyList.bind]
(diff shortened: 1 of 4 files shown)


Mathlib/Testing/SlimCheck/Functions.lean
```diff
@@ -393,12 +393,12 @@ def Perm.slice [DecidableEq α] (n m : ℕ) :
 /-- A lazy list, in decreasing order, of sizes that should be
```
Vierkantor, line 393:
```suggestion
/-- A list, in decreasing order, of sizes that should be
```

Mathlib/Data/LazyList/Basic.lean
```diff
@@ -11,11 +11,10 @@ import Mathlib.Lean.Thunk
 /-!
 ## Definitions on lazy lists
 
-This file contains various definitions and proofs on lazy lists.
-
-TODO: This file will soon be deprecated.
+This file is entirely deprecated, and contains various definitions and proofs on lazy lists.
```
Vierkantor, line 14: I was wondering if we should add a pointer to `MLList` here, but it's probably fine since `LazyList` already says so in the deprecation warning.

Vierkantor (approved):
Unfortunately GitHub does not let me comment there, but the import of `Batteries.Data.LazyList` in `Mathlib/Testing/SlimCheck/Functions.lean` can be deleted (the shake linter also caught this, thankfully).

LGTM otherwise!

bors d+

semorrison:
bors merge

PR #15041, opened 2024-07-22
[Merged by Bors] - feat(Group/Measure): drop a `MeasurableMul` assumption

urkud:
In fact, for `μ ≠ 0` the definition still implies that the multiplication is (a.e.) measurable,
but Lean doesn't have to search for this instance now.

---


RemyDegenne:
Thanks!
bors r+

PR #15044, opened 2024-07-23
[Merged by Bors] - feat(CategoryTheory): preservation of pullback/pushout squares

joelriou:
The main result in this PR is that a commutative square `sq : Square C` is a pullback square iff it is mapped to a pullback square in the category of types when we apply the functor `coyoneda.obj X` for any `X : Cᵒᵖ`. In order to do this, it is shown that the proprety `Square.IsPullback` is preserved when a functor preserving this pullback is applied. A similar result is obtained for pushout squares, which required developing the "opposites" API for squares. 

---

- [x] depends on: #15157


Mathlib/CategoryTheory/Limits/Preserves/Shapes/Square.lean
```diff
@@ -0,0 +1,99 @@
...
+  Square.IsPullback.mk _ (isLimitPullbackConeMapOfIsLimit F sq.fac h.isLimit)
+
+lemma IsPushout.map (h : sq.IsPushout) (F : C ⥤ D) [PreservesColimit (span sq.f₁₂ sq.f₁₃) F] :
+    (sq.map F).IsPushout :=
+  Square.IsPushout.mk _ (isColimitPushoutCoconeMapOfIsColimit F sq.fac h.isColimit)
+
+variable (sq)
+
+lemma isPullback_iff_map_coyoneda_isPullback :
+    sq.IsPullback ↔ ∀ (X : Cᵒᵖ), (sq.map (coyoneda.obj X)).IsPullback := by
```
erdOne, line 50: Doesn't this follow from the fact that `yoneda` preserves & reflects limits and that limits in functor categories can be checked pointwise? Is it possible to specialize these results to `IsPullback` and get this lemma?

joelriou: I have developped the API in #15157.

erdOne:
Thanks!
maintainer merge

mattrobball:
bors merge

PR #15047, opened 2024-07-23
[Merged by Bors] - feat(NNRat): More basic lemmas

YaelDillies:
From LeanAPAP

---


Mathlib/Data/NNRat/Defs.lean
```diff
@@ -56,7 +56,8 @@ theorem ext : (p : ℚ) = (q : ℚ) → p = q :=
 protected theorem coe_injective : Injective ((↑) : ℚ≥0 → ℚ) :=
   Subtype.coe_injective
 
-@[simp, norm_cast]
+-- We want to use this lemma earlier than the lemma simp can prove it with
```
b-mehta, line 59: Let's name the lemma in question here: which is the specific one that you want to use it earlier than?

b-mehta (approved):
The changes here appear sensible and I agree they are useful in practice. I'd like someone else to double-check the instance changes are usable.

maintainer merge

YaelDillies:
Probably @eric-wieser wants to have a look

mattrobball:
I am not a fan of the import changes.

Mathlib/Data/Rat/Cast/Lemmas.lean
```diff
@@ -3,8 +3,10 @@ Copyright (c) 2019 Johannes Hölzl. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Johannes Hölzl, Mario Carneiro
 -/
-import Mathlib.Data.Rat.Cast.Defs
 import Mathlib.Algebra.Field.Basic
+import Mathlib.Data.NNRat.Lemmas
```
mattrobball, line 7: This seems like a big change in the import structure.

Mathlib/Data/Rat/Cast/Defs.lean
```diff
@@ -8,6 +8,7 @@ import Mathlib.Algebra.Group.Commute.Basic
 import Mathlib.Algebra.GroupWithZero.Units.Lemmas
 import Mathlib.Algebra.Order.Field.Rat
 import Mathlib.Data.Int.Cast.Lemmas
+import Mathlib.Data.NNRat.Lemmas
```
mattrobball, line 11: Given the current design, I don't really want `NNRat` to go first before `Rat`.

Mathlib/Algebra/Algebra/Rat.lean
```diff
@@ -25,25 +26,89 @@ theorem map_rat_algebraMap [Semiring R] [Semiring S] [Algebra ℚ R] [Algebra 
...
+    Nat.cast_smul_eq_nsmul]
+
+end Semiring
+
+section DivisionSemiring
+variable [DivisionSemiring S] [CharZero S]
+
+instance instAlgebra : Algebra ℚ≥0 R where
+  smul_def' := smul_def
+  toRingHom := castHom _
```
eric-wieser, line 51: I think you need to inline `castHom` here, otherwise you get defeq-but-not-reducible-defeq diamonds. (cc @mattrobball)

Mathlib/Algebra/Algebra/Rat.lean
```diff
@@ -25,25 +26,89 @@ theorem map_rat_algebraMap [Semiring R] [Semiring S] [Algebra ℚ R] [Algebra 
...
+  commutes' := cast_commute
+
+instance instLinearMapClass [FunLike F R S] [RingHomClass F R S] :
```
eric-wieser, line 54: Arguably, `RingHomClass.toNNRatLinearMapClass` would be a more accurate name.

Mathlib/Algebra/Algebra/Rat.lean
mattrobball: I am guessing that `castHom` semi-reducible?

Mathlib/Algebra/Algebra/Rat.lean
YaelDillies: Yes, true

Mathlib/Algebra/Algebra/Rat.lean
YaelDillies: Oh, because `Algebra` also implements a coercion... I will make `castHom` reducible, since that's how we use it in practice anyway

Mathlib/Data/Rat/Cast/Defs.lean
YaelDillies: Sorry, I don't understand this complaint. `NNRat` should always be earlier than `Rat` since it's used in `Semifield` while `Rat` is used in `Field` only.

Oh I guess you mean "the current design of `NNRat` as a subtype of `Rat`" in which case I shall reply "That design needs to be changed for independent reasons anyway, and I am willing to make it happen soon".

Mathlib/Data/NNRat/Defs.lean
```diff
@@ -56,7 +56,8 @@ theorem ext : (p : ℚ) = (q : ℚ) → p = q :=
 protected theorem coe_injective : Injective ((↑) : ℚ≥0 → ℚ) :=
   Subtype.coe_injective
 
-@[simp, norm_cast]
+-- We want to use this lemma earlier than the lemma simp can prove it with, namely `NNRat.cast_inj`
+@[simp, norm_cast, nolint simpNF]
```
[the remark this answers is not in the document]
YaelDillies: Uh, maybe you're right, but I have been using this `nolint simpNF` pattern for a while so there will be a lot of occurrences to be fixed if you truly believe this is a better solution.

Mathlib/Algebra/Algebra/Rat.lean
```diff
@@ -25,25 +26,89 @@ theorem map_rat_algebraMap [Semiring R] [Semiring S] [Algebra ℚ R] [Algebra 
...
+  smul_comm q a b := by simp [smul_def, mul_smul_comm]
 
-/-- The rational numbers are an algebra over the non-negative rationals. -/
-instance : Algebra NNRat ℚ :=
-  NNRat.coeHom.toAlgebra
+instance instSMulCommClass' [SMulCommClass S R S] : SMulCommClass R ℚ S :=
+  have := SMulCommClass.symm S R S; SMulCommClass.symm _ _ _
 
-/-- The two `Algebra ℚ ℚ` instances should coincide. -/
-example : algebraRat = Algebra.id ℚ :=
```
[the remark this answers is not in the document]
YaelDillies: I created a new file under `test` and wrote
```
import Mathlib.Algebra.Algebra.Rat

/-- The two `Algebra ℚ ℚ` instances should coincide. -/
example : algebraRat = Algebra.id ℚ := rfl
```
and it failed 😨 

Turns out `algebraRat` was a bound variable as we haven't turned off auto-implicit in the `test` folder 😅

Mathlib/Data/Rat/Cast/Lemmas.lean
YaelDillies: I know why this is necessary and I will say two things:
1. It's avoidable in a way that doesn't involve unnecessarily splitting files
2. I am working on this way, but it's definitely a later PR and I don't really want to get into the weeds of that now

Mathlib/Data/Rat/Cast/Defs.lean
mattrobball: Subtype of `Rat` is fine in and of itself. But `NNRat.Defs` imports way too much bundled ordered algebra classes to do what it needs to do. I'd prefer that gets fixed before we add more on top of it.

YaelDillies: `Data.NNRat.Defs` is already imported in `Data.Rat.Cast.Defs` on master, so I do not understand your comment

mattrobball: Sorry, I kinda lost this. I already have a PR disconnecting the basics of `NNRat.Defs` #15072 from bundled ordered algebra that this might conflict with. Can we wait until that gets sorted out?

Mathlib/Data/NNRat/Defs.lean
```diff
@@ -30,6 +30,16 @@ Whenever you state a lemma about the coercion `ℚ≥0 → ℚ`, check that Lean
...
 -/
 
+library_note "specialised high prio simp lemma" /--
+It sometimes happens that a simp lemma declared early in the library can be proved by `simp` using
+later, more general simp lemmas. In that case, the following reasons might be arguments for the
+early lemma to be tagged `@[simp high]` (rather than `@[simp, nolint simpNF]` or unsimped):
+1. There is a significant portion of the library which needs the early lemma to be available via
+  `simp` and which doesn't have access to the more general lemmas.
+2. The more general lemmas have more complicated typeclass assumptions, causing rewrites with them
+  to be slower.
```
eric-wieser, line 40:
```suggestion
1. There is a significant portion of the library which needs the early lemma to be available via
   `simp` and which doesn't have access to the more general lemmas.
2. The more general lemmas have more complicated typeclass assumptions, causing rewrites with them
   to be slower.
```

Mathlib/Data/NNRat/Defs.lean
```diff
@@ -30,6 +30,16 @@ Whenever you state a lemma about the coercion `ℚ≥0 → ℚ`, check that Lean
 `Subtype.val`. Else your lemma will never apply.
 -/
 
+library_note "specialised high prio simp lemma" /--
+It sometimes happens that a simp lemma declared early in the library can be proved by `simp` using
+later, more general simp lemmas. In that case, the following reasons might be arguments for the
+early lemma to be tagged `@[simp high]` (rather than `@[simp, nolint simpNF]` or unsimped):
```
eric-wieser, line 36:
```suggestion
early lemma to be tagged `@[simp high]` (rather than `@[simp, nolint simpNF]` or un`@[simp]`ed):
```

Mathlib/Data/NNRat/Defs.lean
```diff
@@ -30,6 +30,16 @@ Whenever you state a lemma about the coercion `ℚ≥0 → ℚ`, check that Lean
 `Subtype.val`. Else your lemma will never apply.
 -/
 
+library_note "specialised high prio simp lemma" /--
```
eric-wieser, line 33: I think we usually avoid abbreviations here?
```suggestion
library_note "specialised high priority simp lemma" /--
```
and at the references.

Mathlib/Data/NNRat/Defs.lean
```diff
@@ -30,6 +30,16 @@ Whenever you state a lemma about the coercion `ℚ≥0 → ℚ`, check that Lean
...
+It sometimes happens that a simp lemma declared early in the library can be proved by `simp` using
```
eric-wieser, line 34:
```suggestion
It sometimes happens that a simp lemma declared early in the library can be proved by `simp` using
```

Mathlib/Data/Rat/Cast/CharZero.lean
```diff
@@ -34,64 +28,90 @@ theorem cast_inj [CharZero α] : ∀ {m n : ℚ}, (m : α) = n ↔ m = n
...
 
-variable (α)
-variable [CharZero α]
+@[simp, norm_cast] lemma cast_mul (p q : ℚ) : ↑(p * q) = (p * q : α) :=
+  cast_mul_of_ne_zero (Nat.cast_ne_zero.2 p.pos.ne') (Nat.cast_ne_zero.2 q.pos.ne')
 
+variable (α) in
 /-- Coercion `ℚ → α` as a `RingHom`. -/
-def castHom : ℚ →+* α where
+@[reducible] def castHom : ℚ →+* α where
```
eric-wieser, line 47: On second thoughts, let's leave this can of worms closed for now:
```suggestion
def castHom : ℚ →+* α where
```
We can fix it in a later PR.

Mathlib/Data/Rat/Cast/CharZero.lean
```diff
@@ -34,64 +28,90 @@ theorem cast_inj [CharZero α] : ∀ {m n : ℚ}, (m : α) = n ↔ m = n
...
+
+@[simp, norm_cast] lemma cast_add (p q : ℚ≥0) : ↑(p + q) = (p + q : α) :=
+  cast_add_of_ne_zero (Nat.cast_ne_zero.2 p.den_pos.ne') (Nat.cast_ne_zero.2 q.den_pos.ne')
+
+@[simp, norm_cast] lemma cast_mul (p q) : (p * q : ℚ≥0) = (p * q : α) :=
+  cast_mul_of_ne_zero (Nat.cast_ne_zero.2 p.den_pos.ne') (Nat.cast_ne_zero.2 q.den_pos.ne')
+
+variable (α) in
+/-- Coercion `ℚ≥0 → α` as a `RingHom`. -/
+@[reducible] def castHom : ℚ≥0 →+* α where
```
eric-wieser, line 95: idem
```suggestion
def castHom : ℚ≥0 →+* α where
```

Mathlib/Data/Rat/Cast/Defs.lean
```diff
@@ -222,27 +223,56 @@ theorem map_ratCast [DivisionRing α] [DivisionRing β] [RingHomClass F α β] (
...
+
+/-- If monoid with zero homs `f` and `g` from `ℚ` agree on the integers then they are equal. -/
 theorem ext_rat' (h : ∀ m : ℤ, f m = g m) : f = g :=
   (DFunLike.ext f g) fun r => by
     rw [← r.num_div_den, div_eq_mul_inv, map_mul, map_mul, h, ← Int.cast_natCast,
       eq_on_inv₀ f g]
     apply h
 
-/-- If `f` and `g` agree on the integers then they are equal `φ`.
+/-- If monoid with zero homs `f` and `g` from `ℚ≥0` agree on the integers then they are equal.
```
eric-wieser, line 266:
```suggestion
/-- If monoid with zero homs `f` and `g` from `ℚ` agree on the integers then they are equal.
```

Mathlib/Data/Rat/Cast/Defs.lean
```diff
@@ -222,27 +223,56 @@ theorem map_ratCast [DivisionRing α] [DivisionRing β] [RingHomClass F α β] (
...
 
 See note [partially-applied ext lemmas] for why `comp` is used here. -/
 @[ext]
 theorem ext_rat {f g : ℚ →*₀ M₀}
     (h : f.comp (Int.castRingHom ℚ : ℤ →*₀ ℚ) = g.comp (Int.castRingHom ℚ)) : f = g :=
   ext_rat' <| DFunLike.congr_fun h
 
-/-- Positive integer values of a morphism `φ` and its value on `-1` completely determine `φ`. -/
+/-- If monoid with zero homs `f` and `g` from `ℚ≥0` agree on the positive naturals and `-1` then
```
eric-wieser, line 274:
```suggestion
/-- If monoid with zero homs `f` and `g` from `ℚ` agree on the positive naturals and `-1` then
```

Mathlib/Data/Rat/Cast/Lemmas.lean
```diff
@@ -20,6 +21,11 @@ namespace Rat
 
 variable {α : Type*} [DivisionRing α]
 
+@[simp, norm_cast]
+lemma cast_pow (p : ℚ) (n : ℕ) : ↑(p ^ n) = (p ^ n : α) := by
```
eric-wieser, line 25: Do we have the `zpow` version somewhere else?

test/algebra_rat.lean
eric-wieser, line 1: Can you put this in the instance_diamonds folder?

eric-wieser (approved):
My preference would be to put this in before #15072; @YaelDillies, are you happy to help resolve conflicts in the other PR if they go in that order?

Mathlib/Data/NNRat/Defs.lean
YaelDillies: Uh, that doesn't respect the "indent by two spaces" rule. We don't right-flush anything else, eg in `calc` syntax

Mathlib/Data/Rat/Cast/Lemmas.lean
YaelDillies: Yes, `NNRat.cast_zpow` in `Data.Rat.Cast.CharZero`

YaelDillies:
> My preference would be to put this in before #15072; @YaelDillies, are you happy to help resolve conflicts in the other PR if they go in that order?

That's my preference too, especially since there is so much of APAP waiting on this PR to be merged. I am of course happy to resolve conflicts in the other PR

Mathlib/Data/NNRat/Defs.lean
eric-wieser: Ultimately what matters is whether our markdown formatter allows this.

... which I guess also doesn't matter because `library_note`s don't yet render anywhere:.

Mathlib/Data/Rat/Cast/CharZero.lean
```diff
@@ -66,32 +51,67 @@ def castHom : ℚ →+* α where
...
+  rw [NNRat.cast_def, NNRat.cast_def, Commute.div_eq_div_iff] at hpq
+  rw [← p.num_div_den, ← q.num_div_den, div_eq_div_iff]
+  norm_cast at hpq ⊢
+  any_goals norm_cast
+  any_goals apply den_ne_zero
+  exact Nat.cast_commute ..
+
+@[simp, norm_cast] lemma cast_inj : (p : α) = q ↔ p = q := cast_injective.eq_iff
+
+@[simp, norm_cast] lemma cast_eq_zero : (q : α) = 0 ↔ q = 0 := by rw [← cast_zero, cast_inj]
```
eric-wieser, line 84: I think this might be `eq_iff'`?

eric-wieser (approved):
bors d+

Mathlib/Data/Rat/Cast/CharZero.lean
YaelDillies: Doesn't feel like a huge golf to me, but okay

YaelDillies:
bors merge

mattrobball:
bors r-

Sorry. The discussion wasn't finished

mattrobball:
@YaelDillies thanks for the offer of help. I appreciate anything you can provide. While I dislike reforming (what I view as) bad design, I understand you have things to move forward. 

bors merge

PR #15050, opened 2024-07-23
[Merged by Bors] - feat(Algebra/GeomSum): add two lemmas

rwst:
Add missing lemma `x ^ m - 1 ∣ x ^ (m * n) - 1` for `x` both in `Nat` and `Ring`.

Co-authored-by: grunweg <rothgami@math.hu-berlin.de>

---

### DIFF
--- a/Mathlib/Algebra/GeomSum.lean
+++ b/Mathlib/Algebra/GeomSum.lean
@@ -191,6 +191,11 @@ theorem nat_sub_dvd_pow_sub_pow (x y n : ℕ) : x - y ∣ x ^ n - y ^ n := by
   · have : x ^ n ≤ y ^ n := Nat.pow_le_pow_left h.le _
     exact (Nat.sub_eq_zero_of_le this).symm ▸ dvd_zero (x - y)
 
+lemma nat_pow_sub_dvd_pow_mul_sub (x m n : ℕ) : x ^ m - 1 ∣ x ^ (m * n) - 1 := by
+  nth_rw 2 [← Nat.one_pow n]
+  rw [Nat.pow_mul x m n]
+  apply nat_sub_dvd_pow_sub_pow (x ^ m) 1
+
 theorem Odd.add_dvd_pow_add_pow [CommRing α] (x y : α) {n : ℕ} (h : Odd n) :
     x + y ∣ x ^ n + y ^ n := by
   have h₁ := geom_sum₂_mul x (-y) n
--- a/Mathlib/Algebra/GeomSum.lean
+++ b/Mathlib/Algebra/GeomSum.lean
@@ -5,6 +5,7 @@ Authors: Neil Strickland
 -/
 import Mathlib.Algebra.BigOperators.Intervals
 import Mathlib.Algebra.BigOperators.Ring
+import Mathlib.Algebra.Group.NatPowAssoc
 import Mathlib.Algebra.Order.BigOperators.Ring.Finset
 import Mathlib.Algebra.Order.Field.Basic
 import Mathlib.Algebra.Order.Ring.Abs
@@ -174,6 +175,13 @@ theorem Commute.sub_dvd_pow_sub_pow [Ring α] {x y : α} (h : Commute x y) (n : 
 theorem sub_dvd_pow_sub_pow [CommRing α] (x y : α) (n : ℕ) : x - y ∣ x ^ n - y ^ n :=
   (Commute.all x y).sub_dvd_pow_sub_pow n
 
+theorem nat_sub_dvd_pow_sub_pow (x y n : ℕ) : x - y ∣ x ^ n - y ^ n := by
+  rcases le_or_lt y x with h | h
+  · have : y ^ n ≤ x ^ n := Nat.pow_le_pow_left h _
+    exact mod_cast sub_dvd_pow_sub_pow (x : ℤ) (↑y) n
+  · have : x ^ n ≤ y ^ n := Nat.pow_le_pow_left h.le _
+    exact (Nat.sub_eq_zero_of_le this).symm ▸ dvd_zero (x - y)
+
 theorem one_sub_dvd_one_sub_pow [Ring α] (x : α) (n : ℕ) :
     1 - x ∣ 1 - x ^ n := by
   conv_rhs => rw [← one_pow n]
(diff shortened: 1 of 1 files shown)


grunweg (commented):
Thanks for your contribution! I'm wondering: can `x` be generalised from the natural numbers to, say, an `AddMonoid` (or perhaps `AddCommMonoid`)?

awaiting-author

Mathlib/Algebra/GeomSum.lean
```diff
@@ -191,6 +191,11 @@ theorem nat_sub_dvd_pow_sub_pow (x y n : ℕ) : x - y ∣ x ^ n - y ^ n := by
   · have : x ^ n ≤ y ^ n := Nat.pow_le_pow_left h.le _
     exact (Nat.sub_eq_zero_of_le this).symm ▸ dvd_zero (x - y)
 
+lemma nat_pow_sub_dvd_pow_mul_sub (x m n : ℕ) : x ^ m - 1 ∣ x ^ (m * n) - 1 := by
```
j-loreaux, line 194: I'm confused why we're writing this at all. It follows, in more generality (i.e., `Ring` instead of `Nat`), immediately from `sub_one_dvd_pow_sub_one` by using `x := x ^ m` (and `pow_mul`)

j-loreaux (commented):
Why this lemma in particular? None of the others are generalized in this way.

Mathlib/Algebra/GeomSum.lean
rwst: Why not remove all the simple lemmas, and let not-quite-top mathematicians (and AI) struggle and waste their time to derive them when they need them, inventing the wheel over and over again? 

I'm done with "oh and if you come upon a lemma that's missing just submit a PR".

Sorry for the rant.

rwst:
> Thanks for your contribution! I'm wondering: can `x` be generalised from the natural numbers to, say, an `AddMonoid` (or perhaps `AddCommMonoid`)?
> 
> awaiting-author

I'll put it on my list (also for `nat_sub_dvd_pow_sub_pow`).

grunweg:
Let me label it accordingly.

rwst:
Can there be exponentiation in an additive monoid? When trying this, I get:
```
[AddCommMonoid α] (x : α) (m n : ℕ) : x ^ m...

failed to synthesize
  HPow α ℕ ?m.62821
```

rwst:
Also using a mere `CommMonoid` will complain about subtraction. I mean, this is certainly not the first time the question came up of generalizing a `Nat` lemma that uses exponentiation (in the usual sense) and subtraction (when negative results do not occur). What structure would be the next parent of `Nat`? As I find, it cannot be `CommMonoid`.

grunweg:
Good point. Defining powers (with natural exponent) in a monoid works the same as for natural numbers: you inductively define $a^{n+1}=a^n * n$, and prove that this has the properties you expect. You're right about subtraction, though...

So, perhaps the best result to state is actually for rings. Given natural numbers, can you use the `zify` to convert this to a result about integers, and then apply the integer result? That should remove the need for your lemma in applications.

rwst:
Intuitively, the `push_cast` needs to be told that `x^m-1` is nonnegative, so the cast can move to the `x`.

Mathlib/Algebra/GeomSum.lean
j-loreaux: Sorry, point taken. But still, the generality you want is `Ring`. The fact that it happens to hold for `Nat` is essentially a coincidence. (Because `Nat` subtraction is awful, and it just so happens that either, (a) it's zero, and zero divide anything, it (b) it's nonzero, in which case subtracting one is still well-behaved.)

That said, if you need this lemma for `Nat`, it's fine to add that special case.

rwst:
To recap, the new goal of this PR is to have both `Nat` and `Ring`/`CommRing` versions of

1. `x - y ∣ x ^ n - y ^ n`
2. `x ^ m - 1 ∣ x ^ (m * n) - 1`

There is, however, no need to base the `Nat` proof on the resp. ring version if the `Nat` proof is more simple. This applies to both cases because I find the `zify`ed version of (2) is more complicated than what I originally proposed here, because of multiple cases to be handled (`x` zero/nonzero, `m` zero/nonzero, `m*n` zero/nonzero). This is necessary since the solution was to use `zify [lemma1, lemma2]`:

```
lemma nat_pow_sub_dvd_pow_mul_sub (x m n : ℕ) : x ^ m - 1 ∣ x ^ (m * n) - 1 := by
  have : x = 0 ∨ 0 < x := Nat.eq_zero_or_pos x
  rcases this with h | h
  rw [h, zero_pow_eq]
  have hpos₁ : 1 ≤ x ^ m := by    
    sorry
  have hpos₂ : 1 ≤ x ^ (m * n) := by
    sorry
  zify [hpos₁, hpos₂]
  let a := foo x m n
  push_cast at a
  convert a
```

So, I'll keep the originally proposed `Nat` version of (2), and add the ring version of (2) by @grunweg. The `Nat` version of (1) is `nat_sub_dvd_pow_sub_pow` already in Mathlib, and the ring version of (1) is `sub_dvd_pow_sub_pow` already in Mathlib. What do you think?

Mathlib/Algebra/GeomSum.lean
rwst: See the latest comment in the PR main comment thread.

Mathlib/Algebra/GeomSum.lean
```diff
@@ -184,12 +192,15 @@ theorem sub_one_dvd_pow_sub_one [Ring α] (x : α) (n : ℕ) :
...
-    exact (Nat.sub_eq_zero_of_le this).symm ▸ dvd_zero (x - y)
+lemma pow_one_sub_dvd_pow_mul_sub_one [Ring α] (x : α) (m n : ℕ) :
+    ((x ^ m) - 1 : α) ∣ (x ^ (m * n) - 1) := by
+  rw [npow_mul]
+  exact sub_one_dvd_pow_sub_one (x := x ^ m) (n := n)
+
+lemma nat_pow_one_sub_dvd_pow_mul_sub_one (x m n : ℕ) : x ^ m - 1 ∣ x ^ (m * n) - 1 := by
+  nth_rw 2 [← Nat.one_pow n]
+  rw [Nat.pow_mul x m n]
+  apply nat_sub_dvd_pow_sub_pow (x ^ m) 1
```
j-loreaux, line 203: First, I'll note that what I said before still applies, and moreover, you can turn this idea into a proof (using `zify` as Michael suggested), which is approximately the same complexity (in terms of a proof script) as what is written here.

> But still, the generality you want is Ring. The fact that it happens to hold for Nat is essentially a coincidence. (Because Nat subtraction is awful, and it just so happens that either, (a) it's zero, and zero divides anything, it (b) it's nonzero, in which case subtracting one is still well-behaved.)

```suggestion
  nth_rw 2 [← Nat.one_pow n]
  rw [Nat.pow_mul x m n]
  apply nat_sub_dvd_pow_sub_pow (x ^ m) 1
```

Mathlib/Algebra/GeomSum.lean
```diff
@@ -174,6 +175,13 @@ theorem Commute.sub_dvd_pow_sub_pow [Ring α] {x y : α} (h : Commute x y) (n :
 theorem sub_dvd_pow_sub_pow [CommRing α] (x y : α) (n : ℕ) : x - y ∣ x ^ n - y ^ n :=
   (Commute.all x y).sub_dvd_pow_sub_pow n
 
+theorem nat_sub_dvd_pow_sub_pow (x y n : ℕ) : x - y ∣ x ^ n - y ^ n := by
+  rcases le_or_lt y x with h | h
+  · have : y ^ n ≤ x ^ n := Nat.pow_le_pow_left h _
+    exact mod_cast sub_dvd_pow_sub_pow (x : ℤ) (↑y) n
```
j-loreaux, line 181: Optional of course, but this just shows how to use `zify` to do this:
```suggestion
  · zify [h, Nat.pow_le_pow_left]
    exact sub_dvd_pow_sub_pow _ _ _
```

j-loreaux:
I think the lemmas as stated are fine now. I don't really care which way you choose to prove them, but note that the `zify` proofs can be made to work easily, as I show above.

bors d+

PR #14539, opened 2024-07-08
[Merged by Bors] - chore(scripts): require all scripts to be documented in the README

grunweg:
This is valuable so
- contributors can discover existing useful tools,
- existing knowledge is documented and not lost
- it (hopefully) inspires to re-consider if all each script is useful/why we have it. Having a [bounded number of entry-points](https://matklad.github.io/2023/12/31/O(1)-build-file.html) to CI is useful to have.

This is implemented as a new check, piggy-backing the `lint_style` executable.

To make CI pass, this PR adds a README file to `scripts`, containing stub descriptions for the current files. Future PRs should fill this out.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15051, opened 2024-07-23
[Merged by Bors] - feat: remove `lint-style.sh`

grunweg:
In detail, this PR
- moves the executable bit check into the .yml files, instead of the shell script
- adds a `--fix` flag to `lint-style`, for fixing style errors (when possible)
- moves calling `lint-style.py` into `lake exe lint-style`, and
- switches the CI workflow to call `lake exe lint-style` instead, and removes `lint-style.sh`

Ideally, `lint-style.py` will be completely rewritten in Lean soon, so this calling of an external process is only temporary.

Helps with #14539.

----------


semorrison:
bors merge

grunweg:
Thank you for the fast review!

PR #15056, opened 2024-07-23
[Merged by Bors] - refactor(CategoryTheory/Sites): Allow non-fully-faithful dense subsites.

erdOne:
---

- [x] depends on: #15062 
- [x] depends on: #15063 


Mathlib/CategoryTheory/Sites/Equivalence.lean
```diff
@@ -102,16 +60,28 @@ instance : IsCoverDense e.inverse J where
...
+  infer_instance
+
+theorem coverPreserving : CoverPreserving J (e.inverse.inducedTopology J) e.functor :=
+  IsDenseSubsite.coverPreserving _ _ _
+
 /-- A class saying that the equivalence `e` transports the Grothendieck topology `J` to `K`. -/
 class TransportsGrothendieckTopology : Prop where
   /-- `K` is equal to the induced topology. -/
-  eq_inducedTopology : K = (e.locallyCoverDense J).inducedTopology
+  eq_inducedTopology : K = e.inverse.inducedTopology J
```
erdOne, line 78: Can this be replaced with the new `IsDenseSubsite` typeclass? @dagurtomas

joelriou: I would think that for an equivalence of categories, the `TransportsGrothendieckTopology` condition should be equivalent to the `functorPushforward_mem_iff` condition from `IsDenseSubsite` (while the first three should be automatic). However, the main goal of the `TransportsGrothendieckTopology` API is to show that the `HasSheafify` assumption can be transported by an equivalence of categories, and for this we cannot use the main result `sheafEquiv` of this new `IsDenseSubsite` API because of the existence of limits assumption. Then, it could be interesting to write an `iff` lemma for `IsDenseSubsite` in the case of an equivalence, and maybe change the definition of `TransportsGrothendieckTopology`, but I do not think that this needs to be done in this PR.

erdOne: Yes the results below needs to stay. I'm just saying that the type class itself could be replaced, and should it?
Not in this PR anyways.

Mathlib/CategoryTheory/Sites/DenseSubsite.lean
```diff
@@ -482,51 +494,183 @@ end IsCoverDense
...
 
 instance (Y : Sheaf J A) : IsIso ((G.sheafAdjunctionCocontinuous A J K).counit.app Y) := by
-    haveI : IsIso ((sheafToPresheaf J A).map
-        ((G.sheafAdjunctionCocontinuous A J K).counit.app Y)) := by
-      dsimp
-      rw [sheafAdjunctionCocontinuous_counit_app_val]
-      infer_instance
-    apply ReflectsIsomorphisms.reflects (sheafToPresheaf J A)
+  apply (config := { allowSynthFailures := true })
+  ReflectsIsomorphisms.reflects (sheafToPresheaf J A)
```
joelriou, line 638:
```suggestion
    ReflectsIsomorphisms.reflects (sheafToPresheaf J A)
```

Mathlib/CategoryTheory/Sites/DenseSubsite.lean
```diff
@@ -482,51 +494,183 @@ end IsCoverDense
...
+  apply (config := { allowSynthFailures := true }) NatIso.isIso_of_isIso_app
```
joelriou, line 639: Here, I believe you could use `NatTrans.isIso_iff_isIso_app`.

joelriou:
Thanks!

bors d+

erdOne:
Thanks for the review! 
bors merge

Mathlib/CategoryTheory/Sites/Equivalence.lean
dagurtomas: I agree that this is probably equivalent, but I haven't worked out the details. 

#13486 has now been adapted and is ready for review.

PR #15060, opened 2024-07-23
[Merged by Bors] - feat(Data/Matroid/Map): add missing lemmas

apnelson1:
We add a few missing lemmas, mostly about `Matroid.Basis`, in `Matroid/Map`, and also fix a couple of typos in the same file. 

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #15057, opened 2024-07-23
[Merged by Bors] - fix(Data/Set/Sigma): address comments for #14961

apnelson1:
I mistakenly approved the merge for #14961 before addressing the comments. This changes the content of the PR slightly to address them. 

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #12380, opened 2024-04-23
feat: direct sum of matroids

madvorak:
Co-authored-by: Ivan Sergeev <i.i.sergeyev@gmail.com>
Discussion: https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2312380.20direct.20sum.20of.20matroids

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15061, opened 2024-07-23
[Merged by Bors] - feat(Data/Matroid/Sum): Sums of matroids

apnelson1:
This PR defines the sum of a pair or collection of matroids, and gives basic API. 

The material in `Matroid/Map` makes this both more general and shorter than #12380 , which got a bit snarled up in review. 

- [x] depends on: #15060
- [x] depends on: #15057

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #15062, opened 2024-07-23
[Merged by Bors] - feat(CategoryTheory/Site): Locally fully-faithful functors into sites.

erdOne:
---

### DIFF
--- a/Mathlib/CategoryTheory/Sites/LocallyFullyFaithful.lean
+++ b/Mathlib/CategoryTheory/Sites/LocallyFullyFaithful.lean
@@ -1,0 +1,122 @@
+/-
+Copyright (c) 2024 Andrew Yang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Andrew Yang
+-/
+import Mathlib.CategoryTheory.Sites.SheafOfTypes
+
+/-!
+# Locally fully faithful functors into sites
+
+## Main results
+
+- `CategoryTheory.Functor.IsLocallyFull`:
+  A functor `G : C ⥤ D` is locally full wrt a topology on `D` if for every `f : G.obj U ⟶ G.obj V`,
+  the set of `G.map fᵢ : G.obj Wᵢ ⟶ G.obj U` such that `G.map fᵢ ≫ f` is
+  in the image of `G` is a coverage of the topology on `D`.
+- `CategoryTheory.Functor.IsLocallyFaithful`:
+  A functor `G : C ⥤ D` is locally faithful wrt a topology on `D` if for every `f₁ f₂ : U ⟶ V` whose
+  image in `D` are equal, the set of `G.map gᵢ : G.obj Wᵢ ⟶ G.obj U` such that `gᵢ ≫ f₁ = gᵢ ≫ f₂`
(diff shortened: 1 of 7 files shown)


Mathlib/CategoryTheory/Sites/LocallyFullyFaithful.lean
```diff
@@ -0,0 +1,122 @@
...
+  the set of `G.map fᵢ : G.obj Wᵢ ⟶ G.obj U` such that `G.map fᵢ ≫ f` is
+  in the image of `G` is a coverage of the topology on `D`.
+- `CategoryTheory.Functor.IsLocallyFaithful`:
+  A functor `G : C ⥤ D` is locally faithful wrt a topology on `D` if for every `f₁ f₂ : U ⟶ V` whose
+  image in `D` are equal, the set of `G.map gᵢ : G.obj Wᵢ ⟶ G.obj U` such that `gᵢ ≫ f₁ = gᵢ ≫ f₂`
+  is a coverage of the topology on `D`.
+
+## References
+
+* https://arxiv.org/pdf/1906.08737
```
joelriou, line 24: Could you add this to `references.bib`.

Mathlib/CategoryTheory/Sites/LocallyFullyFaithful.lean
```diff
@@ -0,0 +1,122 @@
...
+universe w v u
+
+namespace CategoryTheory
+
+variable {C : Type*} [Category C] {D : Type*} [Category D] {E : Type*} [Category E] (G : C ⥤ D)
+variable (J : GrothendieckTopology C) (K : GrothendieckTopology D)
+variable {L : GrothendieckTopology E}
+
+/--
+For a functor `G : C ⥤ D`, and an morphism `f : G.obj U ⟶ G.obj V`,
```
joelriou, line 37:
```suggestion
For a functor `G : C ⥤ D`, and a morphism `f : G.obj U ⟶ G.obj V`,
```

Mathlib/CategoryTheory/Sites/LocallyFullyFaithful.lean
```diff
@@ -0,0 +1,122 @@
...
+`Sieve.hasLift G f` is the sieve of `U`
+consisting of those arrows whose composition with `f` has a lift in `G`.
+-/
+def Sieve.hasLift {U V : C} (f : G.obj U ⟶ G.obj V) : Sieve U where
```
joelriou, line 41: I would rather see these definitions in the `Functor` namespace. I may suggest names like `Functor.imageSieve` and `Functor.equalizerSieve` in order to have names that are parallel to what we have for the definition of locally injective/surjective morphisms of presheaves. (The notions should be kind of related, but are not because of universe heterogeneity.)

erdOne: I have renamed it to `Functor.imageSieve` and added a lemma connecting it to `Presheaf.imageSieve`. However `Sieve.equalizer` does not mention functors in its definition and I'm not sure if I should make the same change there.

joelriou:
Thanks!

bors merge

PR #15064, opened 2024-07-23
[Merged by Bors] - feat: the completion of a normed and completable topological field is a normed field

smmercuri:
This PR contains the result that the `UniformSpace.Completion` of a normed and completable topological field is also a normed field. 

This has been split from the larger PR #13577.

---


Mathlib/Analysis/Normed/Module/Completion.lean
```diff
@@ -100,6 +101,24 @@ instance [SeminormedCommRing A] [NormedAlgebra 𝕜 A] [UniformContinuousConstSM
...
+      rw [← Completion.coe_sub, norm_coe, Completion.dist_eq, dist_eq_norm]
+  norm_mul' x y := by
+    refine Completion.induction_on₂ x y ?_ ?_ <;> clear x y
+    · exact isClosed_eq (Continuous.comp continuous_norm continuous_mul)
+        (Continuous.comp _root_.continuous_mul
+          (Continuous.prod_map continuous_norm continuous_norm))
+    · intro x y
+      simp only [← coe_mul, norm_coe]
+      exact norm_mul x y
+
```
j-loreaux, line 121: Notes:
1. You don't need to include `toField` because that's already an instance
2. `fun_prop` can solve a lot of continuity-related goals here, as evidenced below.
3. a judicious `simp` call can clear those secondary goals more easily.
4. `clear` is not often something you actually need.
5. Since the `UniformSpace.Completion` namespace is open, you don't need to prefix `Completion` everywhere

```suggestion
  dist_eq x y := by
    refine induction_on₂ x y ?_ (by simp [← coe_sub, dist_eq_norm])
    exact isClosed_eq (uniformContinuous_extension₂ _).continuous (by fun_prop)
  norm_mul' x y := induction_on₂ x y (isClosed_eq (by fun_prop) (by fun_prop)) (by simp [← coe_mul])
```

j-loreaux (commented):
looks pretty good thanks! Just a few minor points to help clean it up. Sorry for the delay in review.

Mathlib/Analysis/Normed/Module/Completion.lean
smmercuri: Thanks! changes made

j-loreaux:
bors merge

PR #15068, opened 2024-07-23
[Merged by Bors] - feat: the pullback of a completable topological field is a completable topological field

smmercuri:
This PR contains results showing that the pullback of a completable topological field under a uniform inducing ring homomorphism is a completable topological field.

---

This has been split from the larger PR #13577.

### DIFF
--- a/Mathlib/Topology/Algebra/UniformField.lean
+++ b/Mathlib/Topology/Algebra/UniformField.lean
@@ -195,8 +195,6 @@ instance (priority := 100) completableTopField_of_complete (L : Type*) [Field L]
         map (fun x => x⁻¹) F ≤ map (fun x => x⁻¹) (𝓝 x) := map_mono hx
         _ ≤ 𝓝 x⁻¹ := continuousAt_inv₀ hx'
 
-section Pullback
-
 variable {α β : Type*} [Field β] [b : UniformSpace β] [CompletableTopField β]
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/Algebra/UniformField.lean
```diff
@@ -194,3 +194,33 @@ instance (priority := 100) completableTopField_of_complete (L : Type*) [Field L]
...
+  rw [← Filter.push_pull', ← map_zero f, ← hf.inducing.nhds_eq_comap, inf_F, Filter.map_bot]
+
+namespace UniformSpace
+
+instance comap_completableTopField (f : α →+* β)
+    [@T0Space α (UniformSpace.comap f b).toTopologicalSpace] :
+    @CompletableTopField _ _ (UniformSpace.comap f b) :=
+  letI := UniformSpace.comap f b; (uniformInducing_iff_uniformSpace.2 rfl).completableTopField
+
+end UniformSpace
```
dupuisf, line 224:
```suggestion
```
We generally don't have instances like this. Instead, we would just use `UniformInducing.completableTopField` on a case-by-case basis to construct instances as needed.

smmercuri: Done!

dupuisf:
Thanks!

bors r+

PR #15065, opened 2024-07-23
[Merged by Bors] - chore (Algebra.Order.Group.Abs): split file into unbundled and bundled material 

mattrobball:
Currently `Algebra.Order.Group.Abs` imports `OrderedCommMonoid` and other bundled algebra classes despite much of the file using unbundled ordered algebra typeclasses. We split into `Algebra.Order.Group.Unbundled.Abs` and `Algebra.Order.Group.Abs` to separate the bundled and unbundled parts of the file and minimize imports downstream.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15069, opened 2024-07-23
[Merged by Bors] - chore (Algebra.Order.Group.Int): split file into unbundled and bundled ordered algebra

mattrobball:
All but ~5 lines of this file did not require bundled ordered algebra classes. We split them off into `Algebra.Order.Group.Unbundled.Int` leaving the `LinearOrderedAddCommGroup` instance. 

---

- [x] depends on: #15065 


j-loreaux:
I'm delegating this in case we want to wait until after the new `variable` inclusion mechanism, as I imagine splits might make PR harder. But maybe not, I'll let you decide.

bors d+

mattrobball:
Since this is about a concrete type, `Int`, there are not many parameters to the statements.

mattrobball:
bors merge

PR #15070, opened 2024-07-23
[Merged by Bors] - refactor(AlgebraicGeometry): stalk maps

chrisflav:
Introduces `Scheme.Hom.stalkMap` and `LocallyRingedSpace.Hom.stalkMap` with the API copied from presheafed spaces. Removes `LocallyRingedSpace.stalk` and `LocallyRingedSpace.ΓToStalk`. Instead uses `Presheaf.stalk` and `Presheaf.germ` directly.

---

The intent is to get rid of `erw`s when working with stalks on schemes. Especially composition of stalk maps caused issues before.


erdOne:
!bench

Mathlib/Topology/Sheaves/Stalks.lean
```diff
@@ -91,16 +91,28 @@ theorem stalkFunctor_obj (ℱ : X.Presheaf C) (x : X) : (stalkFunctor C x).obj 
...
+/-- The germ of a global section of a presheaf at a point. -/
+def Γgerm (F : X.Presheaf C) (x : X) : F.obj (op ⊤) ⟶ stalk F x :=
+  F.germ ⟨x, show x ∈ ⊤ by trivial⟩
+
 theorem germ_res (F : X.Presheaf C) {U V : Opens X} (i : U ⟶ V) (x : U) :
     F.map i.op ≫ germ F x = germ F (i x : V) :=
   let i' : (⟨U, x.2⟩ : OpenNhds x.1) ⟶ ⟨V, (i x : V).2⟩ := i
   colimit.w ((OpenNhds.inclusion x.1).op ⋙ F) i'.op
 
+lemma Γgerm_res (F : X.Presheaf C) {U : Opens X} {i : U ⟶ ⊤} (x : U) :
```
erdOne, line 103:
```suggestion
@[reassoc]
lemma Γgerm_res (F : X.Presheaf C) {U : Opens X} {i : U ⟶ ⊤} (x : U) :
```

erdOne: I think calling it `map_germ_eq_ Γgerm` is more predictable.

Mathlib/Topology/Sheaves/Stalks.lean
```diff
@@ -115,6 +127,13 @@ theorem stalkFunctor_map_germ {F G : X.Presheaf C} (U : Opens X) (x : U) (f : F
     germ F x ≫ (stalkFunctor C x.1).map f = f.app (op U) ≫ germ G x :=
   colimit.ι_map (whiskerLeft (OpenNhds.inclusion x.1).op f) (op ⟨U, x.2⟩)
 
+@[reassoc, elementwise]
+lemma TopCat.Presheaf.germ_congr {C : Type*} [Category C] [HasColimits C] {X : TopCat}
+    (F : TopCat.Presheaf C X) {U : Opens X} (x x' : U) (h : x = x') :
+    F.germ x ≫ eqToHom (by rw [h]) = F.germ x' := by
+  subst h
+  simp
```
erdOne, line 135: I'm not sure we should have this. The intent is to only use `TopCat.Presheaf.stalkCongr`.

erdOne (commented):
Thanks for doing this!

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -543,6 +543,12 @@ theorem fromSpec_primeIdealOf (x : U) :
     Scheme.ofRestrict_val_base]
   rfl -- `rfl` was not needed before #13170
 
+section
+
+variable (y : PrimeSpectrum Γ(X, U)) (hy : hU.fromSpec.1.base y ∈ U)
+
+end
+
```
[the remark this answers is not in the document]
chrisflav: A forgotten leftover, I already deleted it.

Mathlib/Topology/Sheaves/Stalks.lean
chrisflav: The LHS can appear in proof terms though, even if using `TopCat.Presheaf.stalkCongr` as much as possible. So I think it is still useful, but I don't think I actually use it elsewhere in this PR. So if you insist, I remove it.

erdOne: Are there any examples? In theory we shouldn't have any `X.stalk a = X.stalk b` appearing in proof terms as they are equalities of types and are evil.

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -690,6 +690,20 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
...
 
+section Stalks
+
+/-- Variant of `AlgebraicGeometry.localRingHom_comp_stalkIso` for `Spec.map`. -/
+@[elementwise]
+lemma Scheme.localRingHom_comp_stalkIso {R S : CommRingCat} (f : R ⟶ S) (p : PrimeSpectrum S) :
+    (StructureSheaf.stalkIso R (PrimeSpectrum.comap f p)).hom ≫
+      CategoryStruct.comp (Y := CommRingCat.of (Localization.AtPrime p.asIdeal))
+        (Localization.localRingHom (PrimeSpectrum.comap f p).asIdeal p.asIdeal f rfl)
+        (StructureSheaf.stalkIso S p).inv =
```
erdOne, line 701: What's with the `CategoryStruct.comp` here? Maybe use `CommRingCat.ofHom` instead?

Mathlib/Topology/Sheaves/Stalks.lean
chrisflav: I had one when preparing this PR, but later found a different solution, I can't find the example now. Let me remove this and if it reappears, I give you the example.

Re: No `X.stalk a = X.stalk b`. That is not what appears here though, right? The lemma is about comparing the maps `F.germ x` and `F.germ x'`.

erdOne: The `(by rw [h])` part is `X.stalk a = X.stalk b` I think.

Mathlib/AlgebraicGeometry/Morphisms/ClosedImmersion.lean
```diff
@@ -86,7 +86,8 @@ theorem spec_of_surjective {R S : CommRingCat} (f : R ⟶ S) (h : Function.Surje
     IsClosedImmersion (Spec.map f) where
   base_closed := PrimeSpectrum.closedEmbedding_comap_of_surjective _ _ h
   surj_on_stalks x := by
-    erw [← localRingHom_comp_stalkIso, CommRingCat.coe_comp, CommRingCat.coe_comp]
+    rw [← Scheme.localRingHom_comp_stalkIso]
+    erw [CommRingCat.coe_comp, CommRingCat.coe_comp]
```
erdOne, line 90: Or maybe there are better formalizations of the LHS of `Scheme.localRingHom_comp_stalkIso` that gets rid of these?

Mathlib/AlgebraicGeometry/OpenImmersion.lean
```diff
@@ -334,19 +334,25 @@ theorem to_iso {X Y : Scheme.{u}} (f : X ⟶ Y) [h : IsOpenImmersion f] [Epi f.1
...
+instance stalk_iso {X Y : Scheme.{u}} (f : X ⟶ Y) [IsOpenImmersion f] (x : X) :
+    IsIso (f.stalkMap x) :=
+  inferInstanceAs <| IsIso (f.val.stalkMap x)
+
 theorem iff_stalk_iso {X Y : Scheme.{u}} (f : X ⟶ Y) :
-    IsOpenImmersion f ↔ OpenEmbedding f.1.base ∧ ∀ x, IsIso (PresheafedSpace.stalkMap f.1 x) :=
-  ⟨fun H => ⟨H.1, inferInstance⟩, fun ⟨h₁, h₂⟩ => @IsOpenImmersion.of_stalk_iso _ _ f h₁ h₂⟩
+    IsOpenImmersion f ↔ OpenEmbedding f.1.base ∧ ∀ x, IsIso (f.stalkMap x) :=
+  ⟨fun H => ⟨H.1, fun x ↦ inferInstanceAs <| IsIso (f.val.stalkMap x)⟩,
+   fun ⟨h₁, h₂⟩ => @IsOpenImmersion.of_stalk_iso _ _ f h₁ h₂⟩
```
erdOne, line 348:
```suggestion
    fun ⟨h₁, h₂⟩ => @IsOpenImmersion.of_stalk_iso _ _ f h₁ h₂⟩
```

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
```diff
@@ -74,37 +74,30 @@ structure Hom (X Y : LocallyRingedSpace.{u}) : Type _ where
...
--- TODO define `sheaf.stalk` so we can write `X.𝒪.stalk` here?
-/-- The stalk of a locally ringed space, just as a `CommRing`.
--/
-noncomputable def stalk (X : LocallyRingedSpace.{u}) (x : X) : CommRingCat :=
-  X.presheaf.stalk x
-
 -- Porting note (#10754): added this instance to help Lean realize stalks are local
 -- (so that `0 ≠ 1` works below)
-instance stalkLocal (x : X) : LocalRing <| X.stalk x := X.localRing x
+instance stalkLocal (x : X) : LocalRing <| X.presheaf.stalk x := X.localRing x
```
erdOne, line 87: This is already present on line 61 right? Not sure why we had it twice...

chrisflav: Removed it.

Mathlib/AlgebraicGeometry/Morphisms/ClosedImmersion.lean
chrisflav: I got rid of the `erw`s, but with a different technique that I think is more useful in general.

Mathlib/AlgebraicGeometry/AffineScheme.lean
chrisflav: Yes, that fixed it. (I copied this from somewhere)

chrisflav:
(btw: I am not sure why the bot does not remove the `merge-conflict` label)

mattrobball:
I think the increases are mostly due to the set of simp theorems about equivalences. Where are these needed?

chrisflav:
> I think the increases are mostly due to the set of simp theorems about equivalences. Where are these needed?

The general ones are used in `stalkMap_hom_inv` and co, which are not used in this PR, but which come up from time to time and are certainly annoying to prove on the fly.

Mathlib/AlgebraicGeometry/Scheme.lean
```diff
@@ -570,4 +575,100 @@ theorem Scheme.Spec_map_presheaf_map_eqToHom {X : Scheme} {U V : X.Opens} (h : U
...
+
+@[reassoc (attr := simp), elementwise (attr := simp)]
+lemma stalkMap_hom_inv {X Y : Scheme.{u}} (e : X ≅ Y) (y : Y) :
+    e.hom.stalkMap (e.inv.val.base y) ≫ e.inv.stalkMap y = eqToHom (by simp) :=
+  LocallyRingedSpace.stalkMap_hom_inv (Scheme.forgetToLocallyRingedSpace.mapIso e) y
+
+@[reassoc (attr := simp), elementwise (attr := simp)]
+lemma stalkMap_inv_hom {X Y : Scheme.{u}} (e : X ≅ Y) (x : X) :
+    e.inv.stalkMap (e.hom.val.base x) ≫ e.hom.stalkMap x = eqToHom (by simp) :=
+  LocallyRingedSpace.stalkMap_inv_hom (Scheme.forgetToLocallyRingedSpace.mapIso e) x
```
chrisflav, line 637: Should the RHS of these be stated in terms of `TopCat.Presheaf.stalkCongr` instead? @erdOne

Mathlib/AlgebraicGeometry/Scheme.lean
```diff
@@ -147,6 +147,11 @@ lemma appLE_congr (e : V ≤ f ⁻¹ᵁ U) (e₁ : U = U') (e₂ : V = V')
 noncomputable def homeomorph [IsIso f] : X ≃ₜ Y :=
   TopCat.homeoOfIso (asIso <| f.val.base)
 
+/-- A morphism of schemes `f : X ⟶ Y` induces a local ring homomorphis from `Y.presheaf.stalk (f x)`
+to `X.presheaf.stalk x` for any `x : X`. -/
+def stalkMap (x : X) : Y.presheaf.stalk (f.val.base x) ⟶ X.presheaf.stalk x :=
```
mattrobball, line 152: If you want access to the dot notation, can this just be an `abbrev`?

chrisflav: I thought the principle is to hide the implementation detail that schemes are implemented as presheafed spaces as much as possible, so we mostly mirror the API. Do you think performance is better if this is an `abbrev`?

Mathlib/AlgebraicGeometry/Scheme.lean
erdOne: Yes.

chrisflav: I did state things in terms of `stalkSpecializes` as that's what `stalkCongr.hom` simps to.

erdOne:
!bench

erdOne (approved):
LGTM thanks!

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -677,6 +677,29 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
 
 end ZeroLocus
 
+section Stalks
+
+/-- Variant of `AlgebraicGeometry.localRingHom_comp_stalkIso` for `Spec.map`. -/
+@[elementwise]
+lemma Scheme.localRingHom_comp_stalkIso {R S : CommRingCat} (f : R ⟶ S) (p : PrimeSpectrum S) :
```
mattrobball, line 684:
```suggestion
lemma Scheme.localRingHom_comp_stalkIso {R S : CommRingCat.{u}} (f : R ⟶ S) (p : PrimeSpectrum S) :
```

Mathlib/AlgebraicGeometry/AffineScheme.lean
```diff
@@ -677,6 +677,29 @@ lemma Scheme.eq_zeroLocus_of_isClosed_of_isAffine (X : Scheme.{u}) [IsAffine X]
...
+    (StructureSheaf.stalkIso R (PrimeSpectrum.comap f p)).hom ≫
+      (CommRingCat.ofHom <| Localization.localRingHom
+        (PrimeSpectrum.comap f p).asIdeal p.asIdeal f rfl) ≫
+      (StructureSheaf.stalkIso S p).inv = (Spec.map f).stalkMap p :=
+  AlgebraicGeometry.localRingHom_comp_stalkIso f p
+
+/-- Given a morphism of rings `f : R ⟶ S`, the stalk map of `Spec S ⟶ Spec R` at
+a prime of `S` is isomorphic to the localized ring homomorphism. -/
+def Scheme.arrowStalkMapSpecIso {R S : CommRingCat} (f : R ⟶ S) (p : PrimeSpectrum S) :
```
mattrobball, line 693:
```suggestion
def Scheme.arrowStalkMapSpecIso {R S : CommRingCat.{u}} (f : R ⟶ S) (p : PrimeSpectrum S) :
```

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
```diff
@@ -313,6 +302,121 @@ instance component_nontrivial (X : LocallyRingedSpace.{u}) (U : Opens X.carrier)
...
+@[simp]
+lemma stalkMap_id (X : LocallyRingedSpace.{u}) (x : X) :
+    (𝟙 X : X ⟶ X).stalkMap x = 𝟙 (X.presheaf.stalk x) :=
+  PresheafedSpace.stalkMap.id _ x
+
+lemma stalkMap_comp {X Y Z : LocallyRingedSpace.{u}} (f : X ⟶ Y) (g : Y ⟶ Z) (x : X) :
+    (f ≫ g : X ⟶ Z).stalkMap x = g.stalkMap (f.val.base x) ≫ f.stalkMap x :=
+  PresheafedSpace.stalkMap.comp f.val g.val x
+
+@[reassoc, elementwise]
```
mattrobball, line 340: New uses `elementwise` seem to be responsible for about half the regression.

Mathlib/AlgebraicGeometry/Scheme.lean
```diff
@@ -570,4 +575,110 @@ theorem Scheme.Spec_map_presheaf_map_eqToHom {X : Scheme} {U V : X.Opens} (h : U
...
+@[simp]
+lemma stalkMap_id (X : Scheme.{u}) (x : X) :
+    (𝟙 X : X ⟶ X).stalkMap x = 𝟙 (X.presheaf.stalk x) :=
+  PresheafedSpace.stalkMap.id _ x
+
+lemma stalkMap_comp {X Y Z : Scheme.{u}} (f : X ⟶ Y) (g : Y ⟶ Z) (x : X) :
+    (f ≫ g : X ⟶ Z).stalkMap x = g.stalkMap (f.val.base x) ≫ f.stalkMap x :=
+  PresheafedSpace.stalkMap.comp f.val g.val x
+
+@[reassoc, elementwise]
```
mattrobball, line 613: Similarly here there is a surprising penalty for this.

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
```diff
@@ -74,37 +74,26 @@ structure Hom (X Y : LocallyRingedSpace.{u}) : Type _ where
...
--- (so that `0 ≠ 1` works below)
-instance stalkLocal (x : X) : LocalRing <| X.stalk x := X.localRing x
-
 /-- A morphism of locally ringed spaces `f : X ⟶ Y` induces
 a local ring homomorphism from `Y.stalk (f x)` to `X.stalk x` for any `x : X`.
 -/
-noncomputable def stalkMap {X Y : LocallyRingedSpace.{u}} (f : X ⟶ Y) (x : X) :
-    Y.stalk (f.1.1 x) ⟶ X.stalk x :=
-  PresheafedSpace.stalkMap f.1 x
+noncomputable def Hom.stalkMap {X Y : LocallyRingedSpace.{u}} (f : Hom X Y) (x : X) :
```
mattrobball, line 88: I am not a big fan of the `Hom` hack here. I think this should be brought up with core but shouldn't hold the PR back.

erdOne: What do you mean by the `Hom` hack? Do you mean using `(f : Hom X Y)` instead of `(f : X ⟶ Y)` to make dot notation work?

mattrobball: Yes

Mathlib/AlgebraicGeometry/Scheme.lean
erdOne: Do you mean the existence of the simp lemmas slow things down or the generation of the simp lemmas are slow? I'll try to generate them by hand if it is the latter.

mattrobball: I think this PR is fine. But someone should look at `elementwise`. I suspect it is doing too much `simp` in its internals.

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
erdOne: I'm not that happy about it too. "`Quiver.Hom` is a semireducible def so it is unreasonable to tell the dot notation to see through it". This is the answer I got the last time I asked.

erdOne:
(For the record, Christian is on holiday and I'm taking over)

mattrobball:
bors merge

erdOne:
@mattrobball I'd like to try writing the `elementwise` lemmas by hand. They use `ConcreteCategory.instFunLike` instead of the right `funlike` instance so should probably be rewritten anyways.

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
```diff
@@ -313,6 +302,167 @@ instance component_nontrivial (X : LocallyRingedSpace.{u}) (U : Opens X.carrier)
...
+-- lemma stalkMap_germ' (U : Opens Y) (x : X) (hx : f.val.base x ∈ U) :
+--     Y.presheaf.germ ⟨f.val.base x, hx⟩ ≫ f.stalkMap x =
+--       f.val.c.app (op U) ≫ X.presheaf.germ (U := (Opens.map f.val.base).obj U) ⟨x, hx⟩ :=
+--   PresheafedSpace.stalkMap_germ' f.val U x hx
+
+-- @[simp]
+-- lemma stalkMap_germ'_apply (U : Opens Y) (x : X) (hx : f.val.base x ∈ U) (y) :
+--     f.stalkMap x (Y.presheaf.germ ⟨f.val.base x, hx⟩ y) =
+--       X.presheaf.germ (U := (Opens.map f.val.base).obj U) ⟨x, hx⟩ (f.val.c.app (op U) y) :=
+--   PresheafedSpace.stalkMap_germ_apply f.val U ⟨x, hx⟩ y
```
erdOne, line 423: For the record, this is the culprit of the slow down.
Still trying to figure out why. This isn't particularly slow on my machine.

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

erdOne:
!bench

Mathlib/Geometry/RingedSpace/LocallyRingedSpace.lean
erdOne: (or maybe the last lemma)

erdOne:
I have given up. I tried various hacks that seems to improve the runtime on my local machine but the slow down is still there. @mattrobball if you are interested please give it a look. If not I think let's just merge it.

mattrobball:
Can you summarize any observations? It is hard to track from the git log. Also, you can use `time -v lake env lean --profile XX/YY/ZZ.lean` on your machine to get statistics for your file. This is helpful so you don't have to wait for `!bench`

erdOne:
So by the bench [before](https://github.com/leanprover-community/mathlib4/pull/15070#issuecomment-2252192141) and [after](https://github.com/leanprover-community/mathlib4/pull/15070#issuecomment-2252332931) the commit [859763](https://github.com/leanprover-community/mathlib4/pull/15070/commits/859763d20c5463cca94121d53eb2440c9294d11b), one can see that `stalkMap_germ'_apply` is single-handedly responsible for the +25% slowdown. I've tried various methods (providing types to arguments, filling in implicit variables by hand) that speeds up the elaboration time of that lemma from 2s to 1s on my local machine, but it doesn't seem to affect anything.

And the command is useful. Thanks! I didn't know about the `--profile` flag. I guess I can experiment some more now.

erdOne:
It doesn't give me instruction count though. Can I somehow get that locally?

mattrobball:
Maybe. I don't know how and use CPU time as a proxy there. It is fairly stable on your own machine.

erdOne:
Yeah but I don't seem to see the +25% instructions (or +25% cpu time) that I'm supposed to see.

mattrobball:
Odd. I can see the different between this branch and master on my machine. I will poke slightly and then merge

mattrobball:
It seems nothing in this PR is the culprit for the increases. During unification, Lean unfolds **everything** to check that `x =?= ↑⟨x, hx⟩` for `(x : X) (hx : α.base x ∈ U)` multiple times. 

I am not sure whether we can optimize this outside of core. 

@erdOne do you still prefer the custom `apply` theorems over the `elementwise` ones?

mattrobball:
[Here is a gist](https://gist.github.com/mattrobball/8a377f332566c085136e98956172a647) of one example of the problem.

erdOne:
Yes they are needed regardless.
I suspect unbundling `TopCat.Presheaf.germ` so that there is no coercion in the signature might help. I might try that next.

mattrobball:
Any other changes anyone wants before it is merged?

erdOne:
I think it's good to go. Christian is on vacation.

mattrobball:
bors merge

PR #15074, opened 2024-07-23
[Merged by Bors] - feat: Real.exists_natCast_add_one_le_pow_of_one_le

pechersky:
Necessary for arguments in approaching a power from above

---

### DIFF
--- a/Mathlib/Data/Real/Archimedean.lean
+++ b/Mathlib/Data/Real/Archimedean.lean
@@ -338,7 +338,7 @@ theorem iInter_Iic_rat : ⋂ r : ℚ, Iic (r : ℝ) = ∅ := by
 /-- Exponentiation is eventually larger than linear growth. -/
 lemma exists_natCast_add_one_lt_pow_of_one_lt {a : ℝ} (ha : 1 < a) :
     ∃ m : ℕ, (m + 1 : ℝ) < a ^ m := by
-  obtain ⟨k, posk, hk⟩ : ∃ (k : ℕ) (_ : 0 < k), a > 1 / k + 1 := by
+  obtain ⟨k, posk, hk⟩ : ∃ k : ℕ, 0 < k ∧ 1 / k + 1 < a := by
     contrapose! ha
     refine le_of_forall_lt_rat_imp_le ?_
     intro q hq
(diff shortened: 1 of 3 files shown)


Mathlib/Data/Real/Archimedean.lean
```diff
@@ -335,4 +335,30 @@ theorem iUnion_Iic_rat : ⋃ r : ℚ, Iic (r : ℝ) = univ := by
 theorem iInter_Iic_rat : ⋂ r : ℚ, Iic (r : ℝ) = ∅ := by
   exact iInter_Iic_eq_empty_iff.mpr not_bddBelow_coe
 
+/-- Exponentiation is eventually larger than linear growth. -/
+lemma exists_natCast_add_one_lt_pow_of_one_lt {a : ℝ} (ha : 1 < a) :
+    ∃ m : ℕ, (m + 1 : ℝ) < a ^ m := by
+  obtain ⟨k, posk, hk⟩ : ∃ (k : ℕ) (_ : 0 < k), a > 1 / k + 1 := by
```
Ruben-VandeVelde, line 341:
```suggestion
  obtain ⟨k, posk, hk⟩ : ∃ k : ℕ, 0 < k ∧ 1 / k + 1 < a := by
```
(this does not require further changes)

Ruben-VandeVelde (approved):
lgtm otherwise

Ruben-VandeVelde:
maintainer merge

riccardobrasca (commented):
The fact that `exists_natCast_add_one_lt_pow_of_one_lt` is missing really means that we don't know the limit of `(m+1)/a^m`?

riccardobrasca:
Anyway LGTM, thanks!

bors d+

pechersky:
I looked for this limit, or anything like mul_le_pow which would be necessary for it, but couldn't find it. The closest thing I found was Bernoulli's inequality at `one_add_mul_self_lt_rpow_one_add`.

@riccardobrasca would you like me to add the tendsto statement? I can strengthen the exists to say that all n greater than m also satisfy this inequality. I think that statement would need a lemma that would look like a + b < a * b, which I couldn't find either.

riccardobrasca:
I don't have time anymore today, but we can just merge this and add the limit later if we need it. Thanks!

PR #15077, opened 2024-07-23
[Merged by Bors] - feat(Analysis/Normed/Field/Ultra): nonarchimedean iff norm of nats le one

pechersky:
---

- [x] depends on: #15074
- [x] depends on: #14768


loefflerd:
This PR needs more comments and explanation. 

You have a proof of ~50 lines without any explanation of what's going on – there's not even a docstring for the theorem as a whole, let alone documentation of individual steps. I heard someone (Patrick Massot I think?) suggest as a rule-of-thumb that you should always have a comment every 10 lines at least.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
+/-
+Copyright (c) 2024 Yakov Pechersky. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yakov Pechersky
+-/
+import Mathlib.Analysis.Normed.Field.Basic
+import Mathlib.Analysis.Normed.Group.Ultra
+
+/-!
+## Sufficient contidition to have an ultrametric norm on a field
```
loefflerd, line 10:
```suggestion
# Sufficient conditions to have an ultrametric norm on a field
```

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+facts about the existing norm.
+
+## Main results
+
+* `isUltrametricDist_of_forall_norm_natCast_le_one`: a norm in a field is ultrametric if the norm
+  of the image of a natural is less than or equal to one
+
+## Implementation details
+
+The proof relies on a bounded-from-above argument, so the imports need to include `rpow`
```
loefflerd, line 22: I didn't see any `rpow` in the proof...

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+
+## Tags
+
+ultrametric, nonarchimedean
+-/
+open Metric IsUltrametricDist NNReal
```
loefflerd, line 28: You don't need to open `IsUltrametricDist` since all your results are inside it.

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+
+lemma isUltrametricDist_of_forall_norm_add_one_le_max_norm_one
+    (h : ∀ x : R, ‖x + 1‖ ≤ max ‖x‖ 1) : IsUltrametricDist R := by
+  apply isUltrametricDist_of_forall_norm_add_le_max_norm
+  intro x y
+  rcases eq_or_ne y 0 with rfl|hy
+  · simp
+  rw [← div_le_div_right (c := ‖y‖) (by simpa using hy), ← norm_div, add_div, div_self hy,
+      ← max_div_div_right (norm_nonneg _), div_self (by simp [hy]), ← norm_div]
+  exact h _
```
loefflerd, line 44:
```suggestion
    (h : ∀ x : R, ‖x + 1‖ ≤ max ‖x‖ 1) : IsUltrametricDist R := by
  refine isUltrametricDist_of_forall_norm_add_le_max_norm (fun x y ↦ ?_)
  rcases eq_or_ne y 0 with rfl | hy
  · simpa only [add_zero] using le_max_left _ _
  · have p : 0 < ‖y‖ := norm_pos_iff.mpr hy
    simpa only [div_add_one hy, norm_div, div_le_iff₀ p, max_mul_of_nonneg _ _ p.le, one_mul,
      div_mul_cancel₀ _ p.ne'] using h (x / y)
```
(I rejigged the proof to have fewer backwards rewrites, it's completely cosmetic)

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+  apply isUltrametricDist_of_forall_norm_add_one_le_max_norm_one
+  intro x
+  cases le_or_lt ‖x‖ 1 with
+  | inl h => simpa [h] using this _ h
+  | inr h =>
+    suffices ‖x + 1‖ ≤ ‖x‖ by simp [this]
+    rw [← div_le_div_right (c := ‖x‖) (h.trans' zero_lt_one), div_self (h.trans' zero_lt_one).ne',
+        ← norm_div, add_div, div_self (by simpa using (h.trans' zero_lt_one)), add_comm]
+    apply this
+    simp [inv_le_one_iff₀, h.le]
```
loefflerd, line 61: The structure of this proof is a little strange since you only use the main hypothesis in the initial `have` block. Maybe worth splitting off a new lemma whose hypothesis is `∀ x : R, ‖x‖ ≤ 1 → ‖x + 1‖ ≤ 1`?

Golfs of the proof:
```suggestion
lemma isUltrametricDist_of_forall_norm_sub_one_of_norm_le_one
    (h : ∀ x : R, ‖x‖ ≤ 1 → ‖x - 1‖ ≤ 1) : IsUltrametricDist R := by
  have {x : R} (hx : ‖x‖ ≤ 1) : ‖x + 1‖ ≤ 1 := by
    simpa only [← neg_add', norm_neg] using h (-x) (norm_neg x ▸ hx)
  refine isUltrametricDist_of_forall_norm_add_one_le_max_norm_one fun x ↦ ?_
  rcases le_or_lt ‖x‖ 1 with h | h
  · exact (this h).trans (le_max_right _ _)
  · suffices ‖x + 1‖ ≤ ‖x‖ from this.trans (le_max_left _ _)
    have p : 0 < ‖x‖ := h.trans' zero_lt_one
    rw [← div_le_div_right p, div_self p.ne', ← norm_div, add_div, div_self (norm_pos_iff.mp p),
      add_comm]
    exact this (by simpa only [one_div, norm_inv] using inv_le_one_of_one_le₀ h.le)
```

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+
+/-- To prove that a normed field is nonarchimedean, it suffices to prove that the norm
```
loefflerd, line 63: I think the style of this proof is great now, this is a model for how long and complex proofs should be written. But I wonder if it would be improved by splitting it into two lemmas, with one of them being the `suffices` block in the current proof. (You can always make one half `private` if you don't think it's interesting.)

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,146 @@
...
+of the image of any natural is less than or equal to one. -/
+lemma isUltrametricDist_of_forall_norm_natCast_le_one
+    (h : ∀ n : ℕ, ‖(n : F)‖ ≤ 1) : IsUltrametricDist F := by
+  -- we first use our hypothesis about the norm of naturals to have that multiplication by
+  -- naturals keeps the norm small
+  replace h : ∀ (x : F) (n : ℕ), ‖n • x‖ ≤ ‖x‖ := by
```
loefflerd, line 69: You can always golf `have h : ∀ (x : foo), bar := by intro x; ...` to `have h (x : foo) : bar`, and the same with `replace`. (Doesn't work with `suffices` though, unfortunately.)

loefflerd (commented):
Here's some comments on the initial part, I didn't get into the details of the big proof yet

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,156 @@
...
+  -- naturals keeps the norm small
+  replace h (x : R) (n : ℕ) : ‖n • x‖ ≤ ‖x‖ := by
+    rw [nsmul_eq_mul, norm_mul]
+    rcases (norm_nonneg x).eq_or_lt with hx|hx
+    · simp [← hx]
+    · rw [mul_le_iff_le_one_left hx]
+      exact h _
+  -- we distribute the LHS using the binomial theorem  where all terms in the sum
+  -- will be of the form `k • x ^ i` for some binomial coefficient `k : ℕ`
+  -- and the terminal terms with powers `0` and `m` will have the binomial coefficient as `1`
```
loefflerd, line 121:
```suggestion
    rcases (norm_nonneg x).eq_or_lt with hx | hx
    · simp only [← hx, mul_zero, le_refl]
    · simpa only [mul_le_iff_le_one_left hx] using h _
  -- we expand the LHS using the binomial theorem, and apply the hypothesis to bound each term by
  -- a power of ‖x‖
```

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,156 @@
...
+    -- assuming commutativity over the whole ring
+    simp_rw [← norm_pow, (Commute.one_right x).add_pow', one_pow, mul_one,
+      -- and this means we have to now deconstruct the antidiagonal sum
+      Finset.Nat.sum_antidiagonal_eq_sum_range_succ (fun i _ ↦ m.choose i • x ^ i) m]
+    -- and prove the inequality term by term, where the binomial coefficient keeps the norm small
+    exact (norm_sum_le _ _).trans (Finset.sum_le_sum fun _ _ ↦ h _ _)
+  -- we now suffice to show that the sum of powers of norms is less than the
+  -- multiple of the larger of the two
+  refine key.trans ?_
+  -- and the number of terms in the sum is precisely `m + 1`
```
loefflerd, line 138:
```suggestion
  transitivity ∑ k ∈ Finset.range (m + 1), ‖x‖ ^ k
  · simpa only [← norm_pow, (Commute.one_right x).add_pow, one_pow, mul_one, nsmul_eq_mul,
      Nat.cast_comm] using (norm_sum_le _ _).trans (Finset.sum_le_sum fun _ _ ↦ h _ _)
  -- the nature of the norm means that one of `1` and `‖x‖ ^ m` is the largest of the two, so the 
  -- other terms in the binomial expansion are bounded by the max of these, and the number of terms
  -- in the sum is precisely `m + 1`
```

Mathlib/Analysis/Normed/Field/Ultra.lean
```diff
@@ -0,0 +1,156 @@
...
+  -- which we show by comparing the terms in the sum one by one
+  refine Finset.sum_le_sum fun i hi ↦ ?_
+  · rcases i with (_|i)
+    · simp
+    refine pow_le_one₀ (norm_nonneg _) ?_
+    rw [← one_pow m] at hx
+    refine le_of_pow_le_pow_left ?_ zero_le_one hx
+    rintro rfl
+    simp at hi
+  · refine pow_le_pow_right₀ ?_ (by simpa [Nat.lt_succ] using hi)
```
loefflerd, line 150:
```suggestion
  · rcases eq_or_ne m 0 with rfl | hm
    · simp only [pow_zero, le_refl,
        show i = 0 by simpa only [zero_add, Finset.range_one, Finset.mem_singleton] using hi]
    · rw [pow_le_one_iff_of_nonneg (norm_nonneg _) hm] at hx
      exact pow_le_one₀ (norm_nonneg _) hx
  · refine pow_le_pow_right₀ ?_ (by simpa only [Finset.mem_range, Nat.lt_succ] using hi)
```

loefflerd (commented):
This is a beautiful argument, I really enjoyed reading this proof! I've suggested some golfs, and some rephrasings of the explanatory comments (but the latter are rather subjective, I'm giving you my perspective on the proof as I understood it, so feel free to adapt / ignore the comments).

loefflerd:
LGTM! 

maintainer merge

kbuzzard:
Many thanks to all involved here.

bors merge

PR #15088, opened 2024-07-23
[Merged by Bors] - feat: add `[s|i]sup_toXXX` for intermediate fields and subalgebras

acmepjz:
- `IntermediateField.[s|i]sup_toSubfield` which states that the `sup`, `sSup` and `iSup` for `IntermediateField` is the same as for `Subfield`.
- `Algebra.[s|i]sup_toSubsemiring` which states that the `sup`, `sSup` and `iSup` for `Subalgebra` is the same as for `Subsemiring`.
- Also add a missing `Algebra.iInf_toSubsemiring`.

---

discussion: https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/IntermediateField.2Esup_toSubfield


acmepjz:
Probably `Subalgebra->Subring->Subsemiring` version should also be added...

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -695,6 +695,16 @@ theorem inf_toSubsemiring (S T : Subalgebra R A) :
     (S ⊓ T).toSubsemiring = S.toSubsemiring ⊓ T.toSubsemiring :=
   rfl
 
+@[simp]
+theorem sup_toSubsemiring (S T : Subalgebra R A) :
+    (S ⊔ T).toSubsemiring = S.toSubsemiring ⊔ T.toSubsemiring := by
+  rw [← S.toSubsemiring.closure_eq, ← T.toSubsemiring.closure_eq, ← Subsemiring.closure_union]
+  change Subsemiring.closure (Set.range (algebraMap R A) ∪ (S ∪ T)) = Subsemiring.closure (S ∪ T)
```
Ruben-VandeVelde, line 702: I don't think it's a great idea to pull this out of thin air. I would add
```
@[simp]
theorem adjoin_toSubsemiring (s : Set A) :
    (adjoin R s).toSubsemiring = Subsemiring.closure (Set.range (algebraMap R A) ∪ s) :=
  rfl
```
after `def adjoin (s : Set A) : Subalgebra R A := ...` and
```
theorem sup_def (S T : Subalgebra R A) : S ⊔ T = adjoin R (S ⊔ T : Set A) := rfl
```
somewhere after `instance : CompleteLattice (Subalgebra R A) where`, and then this can become
```suggestion
  simp_rw [sup_def, adjoin_toSubsemiring, Subalgebra.coe_toSubsemiring, Set.sup_eq_union]
```

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -695,6 +695,16 @@ theorem inf_toSubsemiring (S T : Subalgebra R A) :
...
+  congr 1
+  rw [Set.union_eq_right]
+  rintro _ ⟨x, rfl⟩
+  exact Or.inl (algebraMap_mem S x)
```
Ruben-VandeVelde, line 706:
```suggestion
  exact Set.mem_union_left _ (algebraMap_mem S x)
```

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -712,6 +722,24 @@ theorem sInf_toSubsemiring (S : Set (Subalgebra R A)) :
...
+@[simp]
+theorem sSup_toSubsemiring (S : Set (Subalgebra R A)) (hS : S.Nonempty) :
+    (sSup S).toSubsemiring = sSup (toSubsemiring '' S) := by
+  have h : toSubsemiring '' S = Subsemiring.closure '' (SetLike.coe '' S) := by
+    rw [Set.image_image]
+    congr! with x
+    exact x.toSubsemiring.closure_eq.symm
+  rw [h, sSup_image, ← Subsemiring.closure_sUnion]
+  change Subsemiring.closure (Set.range (algebraMap R A) ∪ ⋃₀ (SetLike.coe '' S)) =
+    Subsemiring.closure (⋃₀ (SetLike.coe '' S))
```
Ruben-VandeVelde, line 735: Similar issue here

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -725,6 +753,18 @@ theorem iInf_toSubmodule {ι : Sort*} (S : ι → Subalgebra R A) :
     toSubmodule (⨅ i, S i) = ⨅ i, toSubmodule (S i) :=
   SetLike.coe_injective <| by simp
 
+@[simp]
+theorem iInf_toSubsemiring {ι : Sort*} (S : ι → Subalgebra R A) :
+    (iInf S).toSubsemiring = ⨅ i, (S i).toSubsemiring := by
+  simp only [iInf, sInf_toSubsemiring, ← Set.range_comp]
+  rfl
```
Ruben-VandeVelde, line 760:
```suggestion
  simp only [iInf, sInf_toSubsemiring, ← Set.range_comp, Function.comp]
```

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -725,6 +753,18 @@ theorem iInf_toSubmodule {ι : Sort*} (S : ι → Subalgebra R A) :
...
+
+@[simp]
+theorem iSup_toSubsemiring {ι : Sort*} [Nonempty ι] (S : ι → Subalgebra R A) :
+    (iSup S).toSubsemiring = ⨆ i, (S i).toSubsemiring := by
+  simp only [iSup, Set.range_nonempty, sSup_toSubsemiring, ← Set.range_comp]
+  rfl
```
Ruben-VandeVelde, line 766:
```suggestion
  simp only [iSup, Set.range_nonempty, sSup_toSubsemiring, ← Set.range_comp, Function.comp]
```

Mathlib/FieldTheory/Adjoin.lean
```diff
@@ -140,6 +140,16 @@ theorem inf_toSubfield (S T : IntermediateField F E) :
     (S ⊓ T).toSubfield = S.toSubfield ⊓ T.toSubfield :=
```
Ruben-VandeVelde, line 140: Same comments in this file

Ruben-VandeVelde (changes requested):
Thanks!

acmepjz:
Sorry I'm on vacation at August; will look at it later.

acmepjz:
Done.

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
acmepjz: Unfortunately, that's not working after merge master.

Ruben-VandeVelde: Try with `Function.comp_def` instead, please

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -618,6 +618,10 @@ def adjoin (s : Set A) : Subalgebra R A :=
   { Subsemiring.closure (Set.range (algebraMap R A) ∪ s) with
     algebraMap_mem' := fun r => Subsemiring.subset_closure <| Or.inl ⟨r, rfl⟩ }
 
+@[simp]
+theorem adjoin_toSubsemiring (s : Set A) :
```
b-mehta, line 622: Might be nicer to get this by tagging `adjoin` with `simps`

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -637,6 +641,10 @@ instance : CompleteLattice (Subalgebra R A) where
   bot := (Algebra.ofId R A).range
   bot_le _S := fun _a ⟨_r, hr⟩ => hr ▸ algebraMap_mem _ _
 
+theorem sup_def (S T : Subalgebra R A) : S ⊔ T = adjoin R (S ∪ T : Set A) := rfl
+
+theorem sSup_def (S : Set (Subalgebra R A)) : sSup S = adjoin R (⋃₀ (SetLike.coe '' S)) := rfl
```
b-mehta, line 646: The rhs here looks like it should be stated with iSup

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
```diff
@@ -711,6 +729,22 @@ theorem sInf_toSubsemiring (S : Set (Subalgebra R A)) :
     (sInf S).toSubsemiring = sInf (Subalgebra.toSubsemiring '' S) :=
   SetLike.coe_injective <| by simp
 
+open Subalgebra in
+@[simp]
+theorem sSup_toSubsemiring (S : Set (Subalgebra R A)) (hS : S.Nonempty) :
+    (sSup S).toSubsemiring = sSup (toSubsemiring '' S) := by
```
b-mehta, line 735: Similarly `iSup` on the right here looks nicer

Mathlib/FieldTheory/Adjoin.lean
```diff
@@ -42,6 +42,10 @@ def adjoin : IntermediateField F E :=
   { Subfield.closure (Set.range (algebraMap F E) ∪ S) with
     algebraMap_mem' := fun x => Subfield.subset_closure (Or.inl (Set.mem_range_self x)) }
 
+@[simp]
+theorem adjoin_toSubfield :
```
b-mehta, line 46: as previously

acmepjz: In this case it is not working, since the definition of `IntermediateField` extends `Subalgebra` but not `Subfield`. The `@[simps]` is not working at all, and `@[simps!]` creates `IntermediateField.adjoin_carrier` which is unwanted, `@[simps! toSubfield]` is not working either.

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
acmepjz: Good idea!

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
acmepjz: I'm not sure about this. The existing result `sInf_toSubsemiring` uses `sInf` on RHS. For consistency, I think this should also use `sSup` instead of `iSup`.

Mathlib/Algebra/Algebra/Subalgebra/Basic.lean
acmepjz: I'm not sure about this. Since LHS is `s` version, I think the RHS should also be `s` version, i.e. `sUnion`. Could you give an equivalent statement in `iUnion`?

jcommelin (commented):
Thanks :tada:

bors merge

PR #15091, opened 2024-07-24
[Merged by Bors] - chore: address `elab_as_elim` workarounds

kmill:
There were a number of places that eliminator elaboration was suppressed using the `(motive := _)` trick to be used in conjuction with `apply`, but it's generally clearer to use `induction`, `refine`, or, sometimes, `rintro` (for `Quot`).

---

### DIFF
--- a/Mathlib/RingTheory/Flat/EquationalCriterion.lean
+++ b/Mathlib/RingTheory/Flat/EquationalCriterion.lean
@@ -263,7 +263,7 @@ theorem exists_factorization_of_comp_eq_zero_of_free [Flat R M] {K N : Type u}
   have (K' : Submodule R K) (hK' : K'.FG) : ∃ (κ : Type u) (_ : Fintype κ) (a : N →ₗ[R] (κ →₀ R))
       (y : (κ →₀ R) →ₗ[R] M), x = y ∘ₗ a ∧ K' ≤ LinearMap.ker (a ∘ₗ f) := by
     revert N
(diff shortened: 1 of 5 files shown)


Mathlib/RingTheory/Flat/EquationalCriterion.lean
```diff
@@ -263,7 +263,7 @@ theorem exists_factorization_of_comp_eq_zero_of_free [Flat R M] {K N : Type u}
   have (K' : Submodule R K) (hK' : K'.FG) : ∃ (κ : Type u) (_ : Fintype κ) (a : N →ₗ[R] (κ →₀ R))
       (y : (κ →₀ R) →ₗ[R] M), x = y ∘ₗ a ∧ K' ≤ LinearMap.ker (a ∘ₗ f) := by
     revert N
-    apply Submodule.fg_induction (P := _) (N := K') (hN := hK')
+    apply Submodule.fg_induction (N := K') (hN := hK')
```
kmill, line 266: This is still `apply`, but `Submodule.fg_induction` isn't marked `elab_as_elim` and the `(P := _)` is unnecessary regardless.

Making it be `elab_as_elim` seems like it might need to wait until the next Lean release.

grunweg (commented):
Thanks for doing this! I agree the new diff is cleaner. I have not verified `elab_as_elim details`.
maintainer merge

mattrobball:
bors merge

PR #15096, opened 2024-07-24
[Merged by Bors] - feat(Nat/log): Nat.log 2 = Nat.log2 and linear bound on Nat.log

Jun2M:
There is a `Nat.log2` function in Lean (Init/Data/Nat/Log2.lean), which corresponds to the floor log base 2 over Nat.
This PR proves this connection and uses it to show that `Nat.log b n ≤ n` as a simple upper bound on Nat.log.

---

### DIFF
--- a/Mathlib/Data/Nat/Log.lean
+++ b/Mathlib/Data/Nat/Log.lean
@@ -197,6 +197,29 @@ theorem add_pred_div_lt {b n : ℕ} (hb : 1 < b) (hn : 2 ≤ n) : (n + b - 1) / 
     succ_pred_eq_of_pos (by omega)]
   exact Nat.add_le_mul hn hb
 
+lemma log2_eq_log_two {n : ℕ} : Nat.log2 n = Nat.log 2 n := by
+  induction n using Nat.strongRec
+  next n ih =>
+    unfold Nat.log2 Nat.log
+    by_cases h : n = 0
+    · simp [h]
+    · congr
+      simp
+      exact ih (n/2) (Nat.div_lt_self (Nat.pos_of_ne_zero h) (by simp))
+
+lemma log_two_le_self {n : ℕ} : Nat.log 2 n ≤ n := by
+  rw [← Nat.log2_eq_log_two]
+  exact log2_le_self n
+
+lemma log_le_self {b n : ℕ} : b.log n ≤ n := by
+  induction b with
+  | zero => simp
+  | succ m ih =>
+    by_cases h : 1 < m
+    · exact le_trans (Nat.log_anti_left h (by omega)) ih
+    · simp [Nat.le_one_iff_eq_zero_or_eq_one] at h
+      obtain rfl | rfl := h <;> simp [Nat.log_two_le_self]
+
 /-! ### Ceil logarithm -/
 
 
--- a/Mathlib/Data/Nat/Log.lean
+++ b/Mathlib/Data/Nat/Log.lean
@@ -204,7 +204,7 @@ lemma log2_eq_log_two {n : ℕ} : Nat.log2 n = Nat.log 2 n := by
     by_cases h : n = 0
     · simp [h]
     · congr
-      simp
+      simp only [ge_iff_le, lt_succ_self, and_true]
       exact ih (n/2) (Nat.div_lt_self (Nat.pos_of_ne_zero h) (by simp))
 
 lemma log_two_le_self {n : ℕ} : Nat.log 2 n ≤ n := by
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Nat/Log.lean
```diff
@@ -197,6 +197,29 @@ theorem add_pred_div_lt {b n : ℕ} (hb : 1 < b) (hn : 2 ≤ n) : (n + b - 1) /
     succ_pred_eq_of_pos (by omega)]
   exact Nat.add_le_mul hn hb
 
+lemma log2_eq_log_two {n : ℕ} : Nat.log2 n = Nat.log 2 n := by
+  induction n using Nat.strongRec
+  next n ih =>
```
Ruben-VandeVelde, line 202: I suggest using ` induction ... with` here as well

Ruben-VandeVelde (approved):

Mathlib/Data/Nat/Log.lean
```diff
@@ -197,6 +197,29 @@ theorem add_pred_div_lt {b n : ℕ} (hb : 1 < b) (hn : 2 ≤ n) : (n + b - 1) /
...
+    unfold Nat.log2 Nat.log
+    by_cases h : n = 0
+    · simp [h]
+    · congr
+      simp only [ge_iff_le, lt_succ_self, and_true]
+      exact ih (n/2) (Nat.div_lt_self (Nat.pos_of_ne_zero h) (by simp))
+
+lemma log_two_le_self {n : ℕ} : Nat.log 2 n ≤ n := by
```
urkud, line 210: This works, either here or right after `pow_log_le_self`:
```lean
lemma log_lt_self (b : ℕ) {x : ℕ} (hx : x ≠ 0) : log b x < x :=
  match le_or_lt b 1 with
  | .inl h => log_of_left_le_one h x ▸ Nat.pos_iff_ne_zero.2 hx
  | .inr h => log_lt_of_lt_pow hx <| lt_pow_self h _

lemma log_le_self (b x : ℕ) : log b x ≤ x :=
  if hx : x = 0 then by simp [hx]
  else (log_lt_self b hx).le
```

Jun2M: suggestion accepted!

Mathlib/Data/Nat/Log.lean
Jun2M: I attempted to the format you suggested but for whatever name I give after |, it seems to throw error "invalid alternative name 'n' " or such. Would it be possible to get a more detailed help?

Mathlib/Data/Nat/Log.lean
```diff
@@ -197,6 +206,18 @@ theorem add_pred_div_lt {b n : ℕ} (hb : 1 < b) (hn : 2 ≤ n) : (n + b - 1) /
...
+lemma log2_eq_log_two {n : ℕ} : Nat.log2 n = Nat.log 2 n := by
+  induction n using Nat.strongRec with
+  | _ =>
+    rename_i n ih
+    unfold Nat.log2 Nat.log
+    by_cases h : n = 0
+    · simp [h]
+    · congr
+      simp only [ge_iff_le, lt_succ_self, and_true]
+      exact ih (n/2) (Nat.div_lt_self (Nat.pos_of_ne_zero h) (by simp))
```
tb65536, line 218: Here's a golf. And I wonder if the `suffices` is already in mathlib somewhere?
```suggestion
  suffices h : ∀ {m}, m ≤ Nat.log2 n ↔ m ≤ Nat.log 2 n from
    le_antisymm (h.mp le_rfl) (h.mpr le_rfl)
  rcases eq_or_ne n 0 with rfl | hn
  · simp only [log2_zero, le_zero_eq, log_zero_right, implies_true]
  simp only [Nat.le_log2 hn, ← Nat.pow_le_iff_le_log Nat.one_lt_two hn, implies_true]
```

Jun2M: Suggestion committed!

Mathlib/Data/Nat/Log.lean
```diff
@@ -197,6 +206,14 @@ theorem add_pred_div_lt {b n : ℕ} (hb : 1 < b) (hn : 2 ≤ n) : (n + b - 1) /
     succ_pred_eq_of_pos (by omega)]
   exact Nat.add_le_mul hn hb
 
+lemma log2_eq_log_two {n : ℕ} : Nat.log2 n = Nat.log 2 n := by
+  rcases eq_or_ne n 0 with rfl | hn
+  · rw [log2_zero, log_zero_right]
+  apply eq_of_forall_le_iff
+  intro m
+  rw [Nat.le_log2 hn, ← Nat.pow_le_iff_le_log Nat.one_lt_two hn]
+
```
urkud, line 215:
```suggestion
```

urkud:
Thanks! :tada:
bors merge

PR #15097, opened 2024-07-24
[Merged by Bors] - feat: the long line linter as a syntax linter

adomani:
Tired of getting your PRs rejected by CI because of a long line?  These days are over!

The new `longLine` linter will highlight long lines, so your PRs will always be green!

:smile: 

Besides adding the linter, the PR removes the code for the older `TextBased` linter
and adds a few `set_option linter.longLine false`s in test files and in a single reference in `Mathlib.Data.QPF.Multivariate.Basic`.

---

- [ ] depends on: #15104

- [ ] depends on: #15107 removing a long line that sneaked in after the first batch
- [ ] depends on: #15111 linting the tests


grunweg (commented):
Thanks for rewriting another text-based linter. I pushed a few minor tweaks directly: this looks good to me. The only question is if the limit should be 100 or 101 characters.

adomani:
I went ahead and shortened the lines exceeding 100 characters: #15104.

Mathlib/Data/QPF/Multivariate/Basic.lean
```diff
@@ -67,10 +67,13 @@ matched because they preserve the properties of QPF. The latter example,
    * Const
 
 each proves that some operations on functors preserves the QPF structure
+-/
 
+set_option linter.longLine false in
```
grunweg, line 72: I'm a bit torn about this change, but don't feel strongly about this.

grunweg (commented):
No further comments from me - feel free to maintainer merge on my behalf once the dependent PR has landed.

Mathlib/Data/QPF/Multivariate/Basic.lean
```diff
@@ -67,10 +67,13 @@ matched because they preserve the properties of QPF. The latter example,
...
+/-!
 ## Reference
 
-! * [Jeremy Avigad, Mario M. Carneiro and Simon Hudon, *Data Types as Quotients of Polynomial Functors*][avigad-carneiro-hudon2019]
+[Jeremy Avigad, Mario M. Carneiro and Simon Hudon, *Data Types as Quotients of Polynomial Functors*][avigad-carneiro-hudon2019]
```
adomani, line 76: Oh the irony.  The other linter now complains about this one!

grunweg: :see_no_evil:

adomani:
@grunweg, I was going to make the two linter coexist for a period, but the failure on the reference made me decide to remove the older one with the 101 character limit.  Let me know if I forgot to remove/update something.

grunweg:
There's a style exception in `style-exceptions.txt` which you can remove now; looks good otherwise!

Apart from that, he linter now rightfully complains about existing tests - I think that is for the better, but needs some fixing as well.

adomani:
Thanks for confirming and for the `style-exception.txt` tip!  I removed the exception, hopefully I've cleared up all the tests as well.

Mathlib/Data/QPF/Multivariate/Basic.lean
adomani: Is it the segmentation of the doc-module or actually using the `set_option`?

I haven't actually tried, but I think that keeping a single doc-module block and using `set_option` between the `import`s and the doc-module would trigger the "late documentation" linter...

grunweg: It is the segmentation - and probably you're right!

test/Lint.lean
```diff
@@ -86,3 +86,32 @@ note: this linter can be disabled with `set_option linter.cdot false`
...
+set_option linter.longLine true in
+/-!                                                                                                -/
+
+#guard_msgs in
+-- Lines with more than 100 characters containing URLs are allowed.
+set_option linter.longLine true in
+/-!  http                                                                                          -/
+
+set_option linter.longLine true
+-- check that the *argument* of `#guard_msgs` is *not* exempt from the linter
```
grunweg, line 105:
```suggestion
-- The *argument* of `#guard_msgs` is *not* exempt from the linter.

File: test/Lint.lean [hunk 4]
```diff
@@ -86,3 +86,32 @@ note: this linter can be disabled with `set_option linter.cdot false`
...
+/--
+warning: This line exceeds the 100 character limit, please shorten it!
+note: this linter can be disabled with `set_option linter.longLine false`
+-/
+#guard_msgs in                                                                            #guard true
+
+-- check that the doc-string of `#guard_msgs` is exempt from the linter
```
grunweg, line 112: (perhaps with additional line wrapping)
```suggestion
-- However, the *doc-string* of `#guard_msgs` is exempt from the linter: these are automatically generated, hence linting them is not helpful.

File: test/RewriteSearch/Basic.lean [hunk 5]
```diff
@@ -17,12 +17,15 @@ example (xs ys : List α) : (xs ++ ys).length = ys.length + xs.length := by
 -- This worked in previous versions, but for now doesn't.
 -- There are of course better tools for AC rewriting, but it would be nice if `rw_search`
 -- could do a little of it in the course of a longer rewrite.
+set_option linter.longLine false in
```
grunweg, line 20: You're unlinting this as it is a commented guard_msg - makes sense.

test/fun_prop.lean
```diff
@@ -202,6 +203,7 @@ A silly example that everything together works as expected
 example (f : ℝ → ℝ → (ℝ →L[ℝ] ℝ)) (hf : Continuous (fun (x,y) => f x y)) :
     Measurable fun x => (f (x / x) (x * x) 1 + x) := by fun_prop
 
+set_option linter.longLine false in
```
grunweg, line 206: Are you excepting this because of the #check?

test/polyrith.lean
```diff
@@ -5,6 +5,8 @@ Authors: Dhruv Bhatia, Robert Y. Lewis, Mario Carneiro
 -/
 import Mathlib.Tactic.Polyrith
 
+set_option linter.longLine false
```
grunweg, line 8: Why this one?

grunweg (commented):
A few small comments. This is really close to landing - thanks a lot for persevering and working your way through all the small details that came up!

grunweg:
(And I'm a bit surprised be the last commit: did you intend to put it in this PR?)

test/Lint.lean
adomani: I added this manually, since otherwise the long line linter complains. :smile:

test/fun_prop.lean
adomani: Yes, indeed.

Given that this file is intended as both a test and a tutorial on how to use `fun_prop`, I tried to keep the original formatting.

test/polyrith.lean
adomani: Most of this file is commented out and it is probably expected to be uncommented once more work on `polyrith` makes it valid again.  Silencing the linter once, instead of on each "offending" segment seemed less invasive.

adomani:
The last commit is the one about `import`s, right?  I think that the linter should look at the imports also, right?

Let me know if you think that `import`s should be treated differently!

test/polyrith.lean
grunweg: Optionally, you can write a comment (so future contributors modifying this file will know). Not something I will insist on, though.

grunweg:
Yes, that was the commit I meant. Let's leave this to a follow-up.
Thanks for persevering all the way, let's land this!
maintainer merge

semorrison:
bors merge

PR #15110, opened 2024-07-24
[Merged by Bors] - feat(Data/Set/Basic, Order/RelClasses): add @[simp] to subset_refl and subset_empty_iff

apnelson1:
As per the description. 

---


apnelson1:
!bench

bustercopley:
!bench

kbuzzard:
benchmarking costs time (and hence money); it's probably best to only benchmark once CI has declared that mathlib is compiling.

bustercopley:
> benchmarking costs time (and hence money); it's probably best to only benchmark once CI has declared that mathlib is compiling (because that's the only way to get meaningful results).

Huh. I assumed it waited for the build to succeed before doing anything at all. It would be convenient if it did.

bustercopley:
!bench

semorrison:
bors merge

PR #15111, opened 2024-07-24
[Merged by Bors] - chore: long lines in tests

adomani:
Hopefully, this is the last adaptation to the long line linter: it now reached the tests, so this should really be everything!

---

### DIFF
--- a/test/fun_prop.lean
+++ b/test/fun_prop.lean
@@ -46,7 +46,7 @@ attribute [fun_prop]
 
 /-!
 Measurability also behaves nicely with respect to taking products.
-Let's mark the product constructor
+Let's mark the product constructor.
 -/
 
 attribute [fun_prop]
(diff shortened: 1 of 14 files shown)


test/fun_prop.lean
```diff
@@ -44,7 +45,8 @@ attribute [fun_prop]
   measurable_pi_lambda
 
 /-!
-Measurability also behaves nicely with respect to taking products. Let's mark the product constructor
+Measurability also behaves nicely with respect to taking products.
+Let's mark the product constructor
```
grunweg, line 49:
```suggestion
Let's mark the product constructor.

File: test/fun_prop.lean [hunk 2]
```diff
@@ -59,7 +61,8 @@ or
 ```
 Measurable.fst : Measurable f → Measurable fun x => Prod.fst (f x)
 ```
-Tactic `fun_prop` can work with both versions and it should be sufficient to provide just one of them.
+Tactic `fun_prop` can work with both versions and it should be sufficient to provide just one of
```
grunweg, line 64: How about
"The tactic `fun_prop` can work with both versions; it should be sufficient to provide just one of them."?

grunweg (commented):
Looks good also, thanks! Two small nit-picks, which I consider optional.
maintainer merge

test/fun_prop.lean
adomani: Thanks: changed!

semorrison:
bors merge

PR #15113, opened 2024-07-24
[Merged by Bors] - feat: add `map_i[I]nf` for sub(semi)ring, subalgebra, subfield etc.

acmepjz:
... which require that the map in question is injective (this automatically holds for fields).

Added for:

- sub(semi)group, submonoid, and their additive counterpart
- sub(semi)ring, (star)subalgebra, and their non-unital counterpart
- subfield and intermediate fields.

Also add `(coe|mem)_iInf` for subfield and (non-unital)subsemiring.

---


j-loreaux:
Can you also add versions for `StarSubalgebra`, and (assuming it still holds) for all the `NonUnital` subobjects too?

acmepjz:
Probably yes, but I'm on vacation at August. Will look at it later.

acmepjz:
> Can you also add versions for `StarSubalgebra`, and (assuming it still holds) for all the `NonUnital` subobjects too?

Done.

Mathlib/RingTheory/NonUnitalSubring/Basic.lean
```diff
@@ -800,6 +800,14 @@ theorem map_iSup {ι : Sort*} (f : F) (s : ι → NonUnitalSubring R) :
     (iSup s).map f = ⨆ i, (s i).map f :=
   (gc_map_comap f).l_iSup
 
+theorem map_inf (s t : NonUnitalSubring R) (f : F) (hf : Function.Injective f) :
+    (s ⊓ t).map f = s.map f ⊓ t.map f := SetLike.coe_injective (Set.image_inter hf)
+
+theorem map_iInf {ι : Sort*} [Nonempty ι] (f : F) (hf : Function.Injective f)
+    (s : ι → NonUnitalSubring R) : (iInf s).map f = ⨅ i, (s i).map f := by
```
j-loreaux, line 807: let's not mix notations, and we generally prefer `map` beforehand, despite dot notation. Can you please change these issues here and throughout?
```suggestion
    (s : ι → NonUnitalSubring R) : map f (⨅ i, s i) = ⨅ i, map f (s i) := by
```

j-loreaux (commented):
Aside from style issues, this seems fine to me. Thanks! Please wait for CI before merging.

bors d+

Mathlib/RingTheory/NonUnitalSubring/Basic.lean
acmepjz: I'm not sure about this. Since the existing part of this file already uses dot notation, I think it's OK to follow these styles?

j-loreaux: Fair enough. It's not a big deal either way.

j-loreaux (commented):
Okay, whatever you did for consistency is fine.

PR #15116, opened 2024-07-24
[Merged by Bors] - refactor: Make `circle` a type `Circle`

YaelDillies:
See https://github.com/leanprover-community/mathlib4/wiki/Tradeoffs-of-concrete-types-defined-as-subobjects for context

---

### DIFF
--- a/Mathlib/Analysis/SpecialFunctions/Complex/CircleAddChar.lean
+++ b/Mathlib/Analysis/SpecialFunctions/Complex/CircleAddChar.lean
@@ -22,7 +22,7 @@ open Complex Function
 open scoped Real
 
 /-- The canonical map from the additive to the multiplicative circle, as an `AddChar`. -/
-noncomputable def AddCircle.toCircle_addChar {T : ℝ} : AddChar (AddCircle T) circle where
+noncomputable def AddCircle.toCircle_addChar {T : ℝ} : AddChar (AddCircle T) Circle where
   toFun := toCircle
   map_zero_eq_one' := toCircle_zero
   map_add_eq_mul' := toCircle_add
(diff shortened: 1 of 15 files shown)


j-loreaux:
!bench

j-loreaux:
Personally, I like this change.

YaelDillies:
Oh hey Jireh. I had a question for you specifically: Your lemmas like [`unitary.spectrum_subset_circle`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/NormedSpace/Star/Spectrum.html#unitary.spectrum_subset_circle) are the only reason I didn't deprecate `circle` in favor of `Circle`. Do you think we should deprecate `circle` and change the statements of these lemmas? or leave it as the PR currently is?

j-loreaux:
Thanks for considering that. It seems we still have to stabilize on a preferred spelling for the circle as a set in the plane. Somewhere else we use `Metric.sphere`.

I'm fine if we change the statements and deprecate. Either way, we'll need a way to refer to the set, and coercing the set to a type and then back to a set is definitely not the way!

YaelDillies:
Okay, what about I leave this PR as is, and if you want to deprecate `circle` then you can change the few theorems of yours that use it?

j-loreaux:
Okay, I'll get to it soon. If it's done before others comment on this PR, then maybe you can just deprecate here. But if this is done first, that's fine.

j-loreaux:
#15343

b-mehta:
I'm un-requesting myself for review here - I don't know this part of the library well enough to have an informed opinion.

Mathlib/Analysis/SpecialFunctions/Complex/CircleAddChar.lean
```diff
@@ -31,17 +31,26 @@ open AddCircle
...
+    Function.Periodic.lift_coe, mul_zero, Circle.exp_zero]
 
 lemma toCircle_intCast (j : ℤ) :
     toCircle (j : ZMod N) = exp (2 * π * I * j / N) := by
-  rw [toCircle, AddChar.compAddMonoidHom_apply, toCircle_addChar, AddChar.coe_mk,
-    AddCircle.toCircle, toAddCircle_intCast, Function.Periodic.lift_coe, expMapCircle_apply]
+  rw [toCircle, AddChar.coe_mk, AddCircle.toCircle, toAddCircle_intCast,
+    Function.Periodic.lift_coe, Circle.exp_apply]
   push_cast
   ring_nf
```
loefflerd, line 55: May I ask why you are making this change? It does not seem to be necessary for the rest of the PR; and the new definition of `ZMod.toCircle` is much longer than the old one.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15118, opened 2024-07-25
[Merged by Bors] - chore: deprecate `sizeOf_dropSlice_lt`

kim-em:
As far as I can tell this lemma is not used anywhere. If it were needed somewhere, it would be better to proceed via `Sublist` in any case.

As a consolation prize I've added the simpler lemma `length_dropSlice`.


grunweg (commented):
I agree that using `length` makes much more sense for Lists (but I'm not an expert on this part of the library). Let's have some more eyes on this:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15123, opened 2024-07-25
[Merged by Bors] - feat(RingTheory/Unramified): Classification of unramifield field extensions.

erdOne:
---


Ruben-VandeVelde (approved):
maintainer merge

Mathlib/Algebra/Polynomial/Taylor.lean
```diff
@@ -125,4 +125,11 @@ theorem sum_taylor_eq {R} [CommRing R] (f : R[X]) (r : R) :
   rw [← comp_eq_sum_left, sub_eq_add_neg, ← C_neg, ← taylor_apply, taylor_taylor, neg_add_self,
     taylor_zero]
 
+theorem eval_add_of_sq_eq_zero {A} [CommSemiring A] (p : Polynomial A) (x y : A) (hy : y ^ 2 = 0) :
```
riccardobrasca, line 128: `y` can be implicit here.

Mathlib/RingTheory/Unramified/Field.lean
```diff
@@ -0,0 +1,212 @@
...
+  rw [LocalRing.isField_iff_maximalIdeal_eq, eq_bot_iff]
+  intro x hx
+  obtain ⟨x, rfl⟩ := (bijective_of_isAlgClosed_of_localRing K A).surjective x
+  show _ = 0
+  rw [← (algebraMap K A).map_zero]
+  by_contra hx'
+  exact hx ((isUnit_iff_ne_zero.mpr
+    (fun e ↦ hx' ((algebraMap K A).congr_arg e))).map (algebraMap K A))
+
+theorem isReduced_of_field :
```
riccardobrasca, line 122: The name is strange, isn't it? There is no `field` assumption.

riccardobrasca (approved):
Thanks!

bors d+

Mathlib/RingTheory/Unramified/Field.lean
erdOne: The statement is "an unramified algebra over a field is reduced", so I gave it the name `FormallyUnramified.isReduced_of_field`. Do you have a better suggestion?

Mathlib/Algebra/Polynomial/Taylor.lean
erdOne: It needs to be explicit in my application. Do you prefer making `y` implicit here and using `(y := _)` at the application?

riccardobrasca: If you need explicit then no problem leaving as it is. Thanks!

Mathlib/RingTheory/Unramified/Field.lean
riccardobrasca: Well, not really. Let's merge this, we can always change the name later.

grunweg:
Coming here from PR triage: is there anything left to do on this PR? It seems this is just waiting for somebody to press the "merge" button.

urkud:
I merged `master` to see if CI is still happy. @riccardobrasca Could you please have another look?

riccardobrasca:
LGTM, but there is still a build error.

erdOne:
bors merge

PR #15127, opened 2024-07-25
[Merged by Bors] - refactor(Quot): use dependent arrows in congruence arguments for  `map` and `map₂` 

goens:
The `map`  and `map₂` functions for `Quot` and `Quotient` used a Lean 3 version of congruence arguments that avoids dependent arrows. This PR changes those arguments to have the more recognizable formulation of congruence.

---


Mathlib/Data/Quot.lean
```diff
@@ -755,3 +755,12 @@ lemma Equivalence.quot_mk_eq_iff {α : Type*} {r : α → α → Prop} (h : Equi
     | symm _ _ _ H => exact h.symm H
     | trans _ _ _ _ _ h₁₂ h₂₃ => exact h.trans h₁₂ h₂₃
   · exact Quot.sound
+
+lemma congr_liftFun {α β : Type} [HasEquiv α] [HasEquiv β] {f : α → β}
+    (h : ∀ x y, x ≈ y → f x ≈ f y) : ((· ≈ ·) ⇒ (· ≈ ·)) f f := by
+  intro x y hxy; exact h x y hxy
```
eric-wieser, line 761:
```suggestion
    (h : ∀ x y, x ≈ y → f x ≈ f y) : ((· ≈ ·) ⇒ (· ≈ ·)) f f := h
```
which is to say, the proof is just "you already gave me the proof"!

eric-wieser (commented):
My suspicion is that any proof that uses these lemmas would be shorter without using them.

urkud:
I have the same impression.

goens:
Maybe this is correct, maybe proofs would be shorter, but not necessarily more readable. The point of this is that congruence is very often not phrased in terms of `⇒` and thus this makes things more readable, in my opinion.

Unfortunately also the classical definition of congruence (used here) does not seem to be reduce to `⇒` so that just passing the proofs of congruence to `map` would work.

eric-wieser:
> Unfortunately also the classical definition of congruence (used here) does not seem to be reduce to `⇒` 

I would claim that classically, the order of universal quantifiers doesn't really exist, and that is just a concern than appears in formalization.

> so that just passing the proofs of congruence to `map` would not work without these lemmas.

All one has to do is replace `intro x₁ x₂ y₁ y₂ hx hy` in your proof with `intro x₁ x₂ hx y₁ y₂ hy`

eric-wieser:
Could you give an example of the type of code that you're claiming doesn't currently work? I suspect the fix for the example you give is as simple as adding an `@` somewhere.

eric-wieser:
Here's what I think is a direct refutation of your claim:
```lean
import Mathlib

variable
  {α β}
  [sα : Setoid α] [sβ : Setoid β]
  (f : α → β)
  (hf : ∀ a b, a ≈ b → f a ≈ f b) -- this does not use `⇒`, but it works fine anyway

def foo : Quotient sα → Quotient sβ := Quotient.map f hf
```

eric-wieser:
`(h : ∀ a b, a ≈ b → f a = f b)` is the argument type taken by `Quotient.lift` ,not `Quotient.map`.

goens:
Sorry, it took me a bit to get around back to writing this example properly. This works:
```
variable (n : ℕ)
instance instEq : Equivalence (Nat.ModEq n) := ⟨refl, Nat.ModEq.symm, Nat.ModEq.trans⟩
instance instS : Setoid ℕ := ⟨(Nat.ModEq n), instEq n⟩
theorem add_congr : ∀ (a b c d : ℕ), (h₁ : a ≡ b [MOD n]) → (h₂ : c ≡ d [MOD n]) → (a+c) ≡ b+d [MOD n] :=
  fun _ _ _ _ h₁ h₂ => h₁.add h₂

def setoid_add {n}: Quotient (instS n) → Quotient (instS n) → Quotient (instS n) :=
  Quotient.map₂  Nat.add (congr_liftFun₂ (add_congr n))
    (sa := instS n) (sb := instS n) (sc := instS n)
```

but if I change it to 

```
def setoid_add {n}: Quotient (instS n) → Quotient (instS n) → Quotient (instS n) :=
  Quotient.map₂  Nat.add (add_congr n)
    (sa := instS n) (sb := instS n) (sc := instS n)
 ```
    
I get the error:
```
application type mismatch
  Quotient.map₂ Nat.add ⋯
argument
  add_congr n
has type
  ∀ (a b c d : ℕ), a ≡ b [MOD n] → c ≡ d [MOD n] → a + c ≡ b + d [MOD n] : Prop
but is expected to have type
  ((fun x x_1 ↦ x ≈ x_1) ⇒ (fun x x_1 ↦ x ≈ x_1) ⇒ fun x x_1 ↦ x ≈ x_1) Nat.add Nat.add : Prop
```

I think the problem is on the `map₂` case Lean doesn't manage to see through the definition. An example more or less along these lines was my original motivation. I should have described this with the PR originally, sorry for that (and thanks for the reviews, I appreciate you taking the time to discuss this with me!)

**Comment by goens** [comment]
> `(h : ∀ a b, a ≈ b → f a = f b)` is the argument type taken by `Quotient.lift`, not `Quotient.map`.

right, but I don't see why map wouldn't take the same to be honest, that's the point. Of course we can use lift, but then what's the point of map. I think map is convenient, but it's not convenient if it doesn't accept a congruence proof but rather only this `⇒` generalization that doesn't unify with a congruence proof.

**Comment by eric-wieser** [comment]
In this case the spelling I'd advocate is:
```lean
import Mathlib
variable (n : ℕ)
instance instEq : Equivalence (Nat.ModEq n) := ⟨refl, Nat.ModEq.symm, Nat.ModEq.trans⟩
instance instS' : Setoid ℕ := ⟨(Nat.ModEq n), instEq n⟩
theorem add_congr : ∀ {a b c d : ℕ}, (h₁ : a ≡ b [MOD n]) → (h₂ : c ≡ d [MOD n]) → (a+c) ≡ b+d [MOD n] :=
  fun h₁ h₂ => h₁.add h₂

def setoid_add {n}: Quotient (instS' n) → Quotient (instS' n) → Quotient (instS' n) :=
  Quotient.map₂  Nat.add fun _ _ h₁ _ _ h₂ => add_congr _ h₁ h₂
```
where `fun` does the work of adjusting the argument order

goens:
Yeah, that works, but then you need to see under the definitions/implementation details to reorder those arguments. I think it's good design to have abstractions that are familiar to people. In my opinion, `map₂` (or at least some `map₂'`) should take these congruence proofs, not an inline lambda rearranging the arguments to make it fit some other generalization. I'm happy to adapt the PR to provide that `map₂ `, but if you think this abstraction is not useful or helpful, feel free to close this PR.

goens:
Thanks for that @YaelDillies, I very much agree with that take. I wasn't aware this was a remnant from Lean 3 and thought it was for some generalization, so I didn't think that refactor was up for discussion. 

I've repurposed this PR, removing the two not-so-useful lemmas and instead changing the congruence hypothesis. Sorry it took so long, sometimes I miss the notifications on github.

Mathlib/Data/Quot.lean
```diff
@@ -227,11 +227,11 @@ protected def hrecOn₂ (qa : Quotient sa) (qb : Quotient sb) (f : ∀ a b, φ 
 
 /-- Map a function `f : α → β` that sends equivalent elements to equivalent elements
 to a function `Quotient sa → Quotient sb`. Useful to define unary operations on quotients. -/
-protected def map (f : α → β) (h : ((· ≈ ·) ⇒ (· ≈ ·)) f f) : Quotient sa → Quotient sb :=
+protected def map (f : α → β) (h : ∀ a b : α, a ≈ b → f a ≈ f b) : Quotient sa → Quotient sb :=
```
YaelDillies, line 230:
```suggestion
protected def map (f : α → β) (h : ∀ ⦃a b : α⦄, a ≈ b → f a ≈ f b) : Quotient sa → Quotient sb :=
```
will probably be much nicer to use

Mathlib/Data/Quot.lean
```diff
@@ -240,12 +240,13 @@ variable {γ : Sort*} {sc : Setoid γ}
 /-- Map a function `f : α → β → γ` that sends equivalent elements to equivalent elements
 to a function `f : Quotient sa → Quotient sb → Quotient sc`.
 Useful to define binary operations on quotients. -/
-protected def map₂ (f : α → β → γ) (h : ((· ≈ ·) ⇒ (· ≈ ·) ⇒ (· ≈ ·)) f f) :
+protected def map₂ (f : α → β → γ) (h : ∀ a₁ a₂ b₁ b₂, a₁ ≈ a₂ → b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```
YaelDillies, line 243: and here too
```suggestion
protected def map₂ (f : α → β → γ) (h : ∀ ⦃a₁ a₂ b₁ b₂⦄, a₁ ≈ a₂ → b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```

YaelDillies (commented):
Thanks for taking this back up! Looks like the build is failing.

On the topic of github notifications, what I found works for me is to look at all notifications in my inbox, read and unread, and mark each of them done once I have treated it.

Mathlib/Data/Quot.lean
```diff
@@ -240,12 +240,13 @@ variable {γ : Sort*} {sc : Setoid γ}
 /-- Map a function `f : α → β → γ` that sends equivalent elements to equivalent elements
 to a function `f : Quotient sa → Quotient sb → Quotient sc`.
 Useful to define binary operations on quotients. -/
-protected def map₂ (f : α → β → γ) (h : ((· ≈ ·) ⇒ (· ≈ ·) ⇒ (· ≈ ·)) f f) :
+protected def map₂ (f : α → β → γ) (h : ∀ ⦃a₁ a₂ b₁ b₂⦄, a₁ ≈ a₂ → b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```
YaelDillies, line 243: Very empirical, but grouping semi-implicit arguments like this is usually better
```suggestion
protected def map₂ (f : α → β → γ) (h : ∀ ⦃a₁ a₂⦄, a₁ ≈ a₂ → ∀ ⦃b₁ b₂⦄, b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```

YaelDillies: This way, the `⦃b₁ b₂⦄` binders don't turn implicit when you only provide the `a₁ ≈ a₂` argument

Mathlib/Data/Quot.lean
```diff
@@ -240,12 +240,14 @@ variable {γ : Sort*} {sc : Setoid γ}
 /-- Map a function `f : α → β → γ` that sends equivalent elements to equivalent elements
 to a function `f : Quotient sa → Quotient sb → Quotient sc`.
 Useful to define binary operations on quotients. -/
-protected def map₂ (f : α → β → γ) (h : ((· ≈ ·) ⇒ (· ≈ ·) ⇒ (· ≈ ·)) f f) :
+protected def map₂ (f : α → β → γ)
+ (h : ∀ ⦃a₁ a₂⦄, a₁ ≈ a₂ → ∀ ⦃b₁ b₂⦄, b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```
YaelDillies, line 244:
```suggestion
   (h : ∀ ⦃a₁ a₂⦄, a₁ ≈ a₂ → ∀ ⦃b₁ b₂⦄, b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) :
```

Mathlib/Data/Quot.lean
```diff
@@ -240,12 +240,14 @@ variable {γ : Sort*} {sc : Setoid γ}
...
     Quotient sa → Quotient sb → Quotient sc :=
   Quotient.lift₂ (fun x y ↦ ⟦f x y⟧) fun _ _ _ _ h₁ h₂ ↦ Quot.sound <| h h₁ h₂
 
 @[simp]
-theorem map₂_mk (f : α → β → γ) (h : ((· ≈ ·) ⇒ (· ≈ ·) ⇒ (· ≈ ·)) f f) (x : α) (y : β) :
+theorem map₂_mk (f : α → β → γ)
+(h : ∀ a₁ a₂, a₁ ≈ a₂ → ∀ b₁ b₂, b₁ ≈ b₂ → f a₁ b₁ ≈ f a₂ b₂) (x : α) (y : β) :
```
YaelDillies, line 250: Do you know you can do
```suggestion
theorem map₂_mk (f : α → β → γ) (h) (x : α) (y : β) :
```
?

Mathlib/Data/Quot.lean
```diff
@@ -227,11 +227,11 @@ protected def hrecOn₂ (qa : Quotient sa) (qb : Quotient sb) (f : ∀ a b, φ 
 
 /-- Map a function `f : α → β` that sends equivalent elements to equivalent elements
 to a function `Quotient sa → Quotient sb`. Useful to define unary operations on quotients. -/
-protected def map (f : α → β) (h : ((· ≈ ·) ⇒ (· ≈ ·)) f f) : Quotient sa → Quotient sb :=
+protected def map (f : α → β) (h : ∀ ⦃a b : α⦄, a ≈ b → f a ≈ f b) : Quotient sa → Quotient sb :=
   Quot.map f h
 
 @[simp]
-theorem map_mk (f : α → β) (h : ((· ≈ ·) ⇒ (· ≈ ·)) f f) (x : α) :
+theorem map_mk (f : α → β) (h : ∀ a b, a ≈ b → f a ≈ f b) (x : α) :
```
YaelDillies, line 234:
```suggestion
theorem map_mk (f : α → β) (h) (x : α) :
```

Mathlib/Data/Quot.lean
```diff
@@ -79,8 +79,8 @@ protected def hrecOn₂ (qa : Quot ra) (qb : Quot rb) (f : ∀ a b, φ ⟦a⟧ 
 
 /-- Map a function `f : α → β` such that `ra x y` implies `rb (f x) (f y)`
 to a map `Quot ra → Quot rb`. -/
-protected def map (f : α → β) (h : (ra ⇒ rb) f f) : Quot ra → Quot rb :=
-  (Quot.lift fun x ↦ ⟦f x⟧) fun x y (h₁ : ra x y) ↦ Quot.sound <| h h₁
+protected def map (f : α → β) (h : ∀ a b : α, ra a b → rb (f a) (f b)) : Quot ra → Quot rb :=
+  Quot.lift (fun x => Quot.mk rb (f x)) <| fun x y hra ↦ Quot.sound <|  h x y hra
```
YaelDillies, line 83: `fun` has very low precedence precisely for this kind of situations
```suggestion
  Quot.lift (fun x => Quot.mk rb (f x)) fun x y hra ↦ Quot.sound <|  h x y hra
```

Mathlib/Data/Quot.lean
goens: I didn't know, no, thanks!

Mathlib/Data/Quot.lean
```diff
@@ -693,7 +694,7 @@ theorem hrecOn₂'_mk'' {φ : Quotient s₁ → Quotient s₂ → Sort*}
 
 /-- Map a function `f : α → β` that sends equivalent elements to equivalent elements
 to a function `Quotient sa → Quotient sb`. Useful to define unary operations on quotients. -/
-protected def map' (f : α → β) (h : (s₁.r ⇒ s₂.r) f f) : Quotient s₁ → Quotient s₂ :=
+protected def map' (f : α → β) (h : ∀ a b, a ≈ b → f a ≈ f b) : Quotient s₁ → Quotient s₂ :=
```
eric-wieser, line 697: This very deliberately uses `.r` and not `≈`; that's the purpose of the `'` in the name.

goens: gotcha @eric-wieser, makes sense that it does - I've changed it accordingly. I don't see how that's obvious from the `'` in the name to be frank.

Mathlib/Data/Quot.lean
```diff
@@ -79,8 +79,8 @@ protected def hrecOn₂ (qa : Quot ra) (qb : Quot rb) (f : ∀ a b, φ ⟦a⟧ 
 
 /-- Map a function `f : α → β` such that `ra x y` implies `rb (f x) (f y)`
 to a map `Quot ra → Quot rb`. -/
-protected def map (f : α → β) (h : (ra ⇒ rb) f f) : Quot ra → Quot rb :=
-  (Quot.lift fun x ↦ ⟦f x⟧) fun x y (h₁ : ra x y) ↦ Quot.sound <| h h₁
+protected def map (f : α → β) (h : ∀ a b : α, ra a b → rb (f a) (f b)) : Quot ra → Quot rb :=
```
YaelDillies, line 82:
```suggestion
protected def map (f : α → β) (h : ∀ ⦃a b : α⦄, ra a b → rb (f a) (f b)) : Quot ra → Quot rb :=
```

YaelDillies (approved):
Can you please update the PR title/description? Thanks!

maintainer delegate

goens:
@YaelDillies done, changed! Thanks a lot for the patience/shepherding to make this PR more useful (also learned a trick or two through the reviews)!

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/Data/Quot.lean
```diff
@@ -702,15 +704,7 @@ theorem map'_mk'' (f : α → β) (h) (x : α) :
...
-protected def map₂' (f : α → β → γ) (h : (s₁.r ⇒ s₂.r ⇒ s₃.r) f f) :
-    Quotient s₁ → Quotient s₂ → Quotient s₃ :=
-  Quotient.map₂ f h
-
-@[simp]
-theorem map₂'_mk'' (f : α → β → γ) (h) (x : α) :
-    (Quotient.mk'' x : Quotient s₁).map₂' f h =
-      (Quotient.map' (f x) (h (Setoid.refl x)) : Quotient s₂ → Quotient s₃) :=
-  rfl
+@[deprecated (since := "2024-12-01")] protected alias map₂' := Quotient.map₂
```
eric-wieser, line 707: Why did we deprecate this?

yuma-mizuno: I revived this #29359.

PR #15129, opened 2024-07-25
chore: remove or tweak outdated comments about 'when X is ported'

grunweg:
- [x] depends on: #15219
- [x] depends on: #15221

---

### DIFF
--- a/Mathlib/Tactic/ApplyFun.lean
+++ b/Mathlib/Tactic/ApplyFun.lean
@@ -13,7 +13,7 @@ import Mathlib.Order.Hom.Basic
 Apply a function to an equality or inequality in either a local hypothesis or the goal.
 
 ## Porting notes
-When the `mono` tactic has been ported we can attempt to automatically discharge `Monotone f` goals.
+We can attempt to automatically discharge `Monotone f` goals using `mono`.
 -/
 
 namespace Mathlib.Tactic
@@ -52,8 +52,7 @@ def applyFunHyp (f : Term) (using? : Option Term) (h : FVarId) (g : MVarId) :
           | none => do
             let f ← elabTermForApply f
             let ng ← mkFreshExprMVar (← mkAppM ``Function.Injective #[f])
-            -- TODO attempt to solve this goal using `mono` when it has been ported,
-            -- via `synthesizeUsing`.
+            -- TODO attempt to solve this goal using `mono` via `synthesizeUsing`.
             pure (ng, [ng.mvarId!])
         pure (← mkAppM' (← mkAppM ``Function.Injective.ne #[injective_f]) #[d.toExpr], newGoals)
       | _ => throwError
@@ -66,8 +65,7 @@ def applyFunHyp (f : Term) (using? : Option Term) (h : FVarId) (g : MVarId) :
         | none => do
           let f ← elabTermForApply f
           let ng ← mkFreshExprMVar (← mkAppM ``StrictMono #[f])
-          -- TODO attempt to solve this goal using `mono` when it has been ported,
-          -- via `synthesizeUsing`.
+          -- TODO attempt to solve this goal using `mono`, via `synthesizeUsing`.
           pure (ng, [ng.mvarId!])
       pure (← mkAppM' strict_monotone_f #[d.toExpr], newGoals)
     | (``LE.le, _) =>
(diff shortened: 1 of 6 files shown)


Mathlib/CategoryTheory/ConcreteCategory/Bundled.lean
```diff
@@ -60,8 +60,7 @@ instance search is able to "see" that a morphism R ⟶ S in Ring is really
 a (semi)ring homomorphism from R.α to S.α, and not merely from
 `(Bundled.map @Ring.toSemiring R).α` to `(Bundled.map @Ring.toSemiring S).α`.
 
-TODO: Once at least one use of this has been ported, check if this still needs to be reducible in
-Lean 4.
+TODO: check if this still needs to be reducible in Lean 4.
```
j-loreaux, line 63: It's not reducible, so apparently not.

Mathlib/Tactic/ApplyFun.lean
```diff
@@ -13,7 +13,7 @@ import Mathlib.Order.Hom.Basic
 Apply a function to an equality or inequality in either a local hypothesis or the goal.
 
 ## Porting notes
-When the `mono` tactic has been ported we can attempt to automatically discharge `Monotone f` goals.
+We can attempt to automatically discharge `Monotone f` goals using `mono`.
```
j-loreaux, line 16: have you tried this?

Mathlib/Tactic/ApplyFun.lean
```diff
@@ -52,8 +52,7 @@ def applyFunHyp (f : Term) (using? : Option Term) (h : FVarId) (g : MVarId) :
           | none => do
             let f ← elabTermForApply f
             let ng ← mkFreshExprMVar (← mkAppM ``Function.Injective #[f])
-            -- TODO attempt to solve this goal using `mono` when it has been ported,
-            -- via `synthesizeUsing`.
+            -- TODO attempt to solve this goal using `mono` via `synthesizeUsing`.
```
j-loreaux, line 55: diito

Mathlib/Tactic/ApplyFun.lean
```diff
@@ -66,8 +65,7 @@ def applyFunHyp (f : Term) (using? : Option Term) (h : FVarId) (g : MVarId) :
         | none => do
           let f ← elabTermForApply f
           let ng ← mkFreshExprMVar (← mkAppM ``StrictMono #[f])
-          -- TODO attempt to solve this goal using `mono` when it has been ported,
-          -- via `synthesizeUsing`.
+          -- TODO attempt to solve this goal using `mono`, via `synthesizeUsing`.
```
j-loreaux, line 68: ditto

Mathlib/Tactic/ApplyFun.lean
```diff
@@ -78,8 +76,7 @@ def applyFunHyp (f : Term) (using? : Option Term) (h : FVarId) (g : MVarId) :
         | none => do
           let f ← elabTermForApply f
           let ng ← mkFreshExprMVar (← mkAppM ``Monotone #[f])
-          -- TODO attempt to solve this goal using `mono` when it has been ported,
-          -- via `synthesizeUsing`.
+          -- TODO attempt to solve this goal using `mono`, via `synthesizeUsing`.
```
j-loreaux, line 79: ditto

j-loreaux (commented):
did you see if the `gcongr` golfing had been done?

I don't see the point of changing the porting note much if we're mainly just leaving the TODOs there. It doesn't really reduce the technical debt.

grunweg:
I have fixed `test/CategoryTheory/Elementwise` in a separate PR: #15219.

Mathlib/CategoryTheory/ConcreteCategory/Bundled.lean
grunweg: Removed that comment in #15221

grunweg:
I've split out and #15219 and #15221 as follow-ups by two of your comments.
I did check about gcongr; this golfing has been done, just the comments not updated. I don't see any easy further golfing opportunities. Of course, there might always be future opportunities, but I don't think these comments are useful for that purpose.

For the other changes: the reason I'm proposing them is that any comment saying "when the port is done" screams "I'm outdated to me". I would like to avoid a future reader wondering "how outdated is this", only to realise "no, it's still current"; I see no need for this.
If you disagree, I'm happy to remove them from this PR.

j-loreaux:
Currently it's one of the ways we track our technical debt. If we remove porting notes, then they won't be tracked. Right?

PR #15135, opened 2024-07-25
[Merged by Bors] - refactor(RingTheory/Unramified): Switch official definition to allow general universes

erdOne:
Under this new definition, `Algebra.FormallySmooth` is only defined for commutative rings.

---


Mathlib/RingTheory/Etale/Basic.lean
```diff
@@ -34,8 +34,8 @@ namespace Algebra
 
 section
 
-variable (R : Type u) [CommSemiring R]
-variable (A : Type u) [Semiring A] [Algebra R A]
+variable (R : Type u) [CommRing R]
```
jcommelin, line 37: Are you intentionally switching from semirings to rings? If so, please write a short reasoning in the PR commit message. I suppose this is because the equivalence with the old definition of formally unramified is only valid in the case of rings, right?

Mathlib/RingTheory/Unramified/Basic.lean
```diff
@@ -3,20 +3,19 @@ Copyright (c) 2022 Andrew Yang. All rights reserved.
...
 import Mathlib.RingTheory.QuotientNilpotent
-import Mathlib.RingTheory.TensorProduct.Basic
+import Mathlib.RingTheory.Kaehler.Basic
 
 /-!
 
 # Unramified morphisms
 
-An `R`-algebra `A` is formally unramified if for every `R`-algebra,
+An `R`-algebra `A` is formally unramified f `Ω[A⁄R]` is trivial.
```
jcommelin, line 15:
```suggestion
An `R`-algebra `A` is formally unramified if `Ω[A⁄R]` is trivial.
```

jcommelin (commented):
bors d+

Mathlib/RingTheory/Etale/Basic.lean
erdOne: Yes. The differential module only makes sense when they are commutative rings.
I have no idea if the original definition makes sense for noncommutative semirings or not anyways.
The PR message is updated.

erdOne:
bors merge

erdOne:
bors merge

PR #15136, opened 2024-07-25
chore: fix theorem naming referencing `Units.oneSub`

j-loreaux:
---

### DIFF
--- a/Mathlib/Analysis/NormedSpace/Units.lean
+++ b/Mathlib/Analysis/NormedSpace/Units.lean
@@ -98,7 +98,7 @@ open scoped Classical
 
 open Asymptotics Filter Metric Finset Ring
 
-theorem inverse_one_sub (t : R) (h : ‖t‖ < 1) : inverse (1 - t) = ↑(Units.oneSub t h)⁻¹ := by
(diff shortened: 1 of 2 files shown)


Ruben-VandeVelde:
It seems like there's no deprecated aliases added?

j-loreaux:
oops, sorry, will do.

Mathlib/Analysis/NormedSpace/Units.lean
```diff
@@ -112,19 +114,22 @@ theorem inverse_add (x : Rˣ) :
   rw [← x.val_add t ht, inverse_unit, Units.add, Units.copy_eq, mul_inv_rev, Units.val_mul,
     ← inverse_unit, Units.val_oneSub, sub_neg_eq_add]
 
-theorem inverse_one_sub_nth_order' (n : ℕ) {t : R} (ht : ‖t‖ < 1) :
+theorem inverse_oneSub_nth_order' (n : ℕ) {t : R} (ht : ‖t‖ < 1) :
```
fpvandoorn, line 117: Some of these lemmas don't actually contain `Units.oneSub`, so I'm not sure if camelCase is actually better. Are you sure you want to rename these?

In either case:
bors d+

j-loreaux: fair enough, I guess that's why they are named this way in the first place. I had originally thought it was just a consequence of porting issues. Closing this.

PR #15139, opened 2024-07-25
[Merged by Bors] - chore: make `#where` command commonly available again

kmill:
---


fpvandoorn:
bors d+

Maybe other Battery-commands should be checked as well?

kmill:
bors r+

(I agree with the idea to make sure other batteries commands are included, but I'll leave that for someone else! :-))

PR #15140, opened 2024-07-25
chore (Elementwise): change default for `simp` reduction to `false`

mattrobball:
Currently, `elementwise` tries to apply `simp` to both sides of an equality to maintain normal form. Doing `simp` on both sides is not without cost. We change this to an opt-in behavior, where the user can choose to apply `simp` on both sides of the equality using `elementwise reduce`.

---


mattrobball:
!bench

mattrobball:
Less benefit than #15070 seemed to indicate

PR #15141, opened 2024-07-25
[Merged by Bors] - feat(RingTheory/Unramified): Formally unramified product of rings

erdOne:
---


Mathlib/RingTheory/Unramified/Pi.lean
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Algebra.FormallyUnramified
+
+universe u v
+
+variable {R : Type max u v} {I : Type v} [Finite I] (f : I → Type max u v)
+variable [CommRing R] [∀ i, CommRing (f i)] [∀ i, Algebra R (f i)]
+
+theorem pi_iff :
+    FormallyUnramified R (∀ i, f i) ↔ ∀ i, FormallyUnramified R (f i) := by
```
jcommelin, line 27: Do you have a strong reason for calling the family of algebras `f`? I would much prefer `A`, as in the module docstring. I also have a mild preference for `Π` over `∀` when working with non-props.

jcommelin (commented):
bors d+

PR #15149, opened 2024-07-25
[Merged by Bors] - chore: robustifying for debug.byAsSorry (part 8)

kim-em:
See [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Running.20Mathlib.20under.20.60set_option.20debug.2EbyAsSorry.60).

### DIFF
--- a/Mathlib/Algebra/IsPrimePow.lean
+++ b/Mathlib/Algebra/IsPrimePow.lean
@@ -27,7 +27,7 @@ theorem isPrimePow_def : IsPrimePow n ↔ ∃ (p : R) (k : ℕ), Prime p ∧ 0 <
 natural `k` such that `n` can be written as `p^(k+1)`. -/
 theorem isPrimePow_iff_pow_succ : IsPrimePow n ↔ ∃ (p : R) (k : ℕ), Prime p ∧ p ^ (k + 1) = n :=
   (isPrimePow_def _).trans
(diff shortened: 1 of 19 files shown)


grunweg (commented):
Looks reasonable, thanks!
maintainer merge

riccardobrasca:
The Declarations diff bot says something changed, can you please check this? Anyway it doesn't seem serious, thanks!

bors d+

semorrison:
> The Declarations diff bot says something changed, can you please check this? Anyway it doesn't seem serious, thanks!
> 
> bors d+

@adomani, could the diff bot include a link to some explanation of what the pluses and minuses in
```
++-- symm
+-+- conj_eq
```
mean?

semorrison:
bors merge

adomani:
I can add a link with an explanation.  Here is what it is anyway!

When the script finds exactly one pair of a name added and removed, then it prints nothing.  When it finds anything else, then it reports how many times each name was added and removed.  So, what this diff means is that there are two "new" `symm`s and two "old" `symm`s disappeared.

I personally slightly prefer it to leave the output as is, since this is giving a little more information than reporting nothing in cases like this one, while not being too much information.  If you disagree, I can simply replace the output to add up the signs from the diff and only report non-zero sums.

PR #15152, opened 2024-07-26
[Merged by Bors] - chore (GCDMonoid.Nat): avoid bundled ordered algebra and move `Init.Data.Int.Order`

mattrobball:
We swap out some bundled ordered algebra API calls for native `Nat` and `Int` ones. We also add one. This allows us to avoid needing bundled ordered algebra in this file and pushes its import further downstream. Since it touches `Init.Data.Int.Order`, we rename it `Data.Int.Order.Basic`. 

---

- [x] depends on: #15069 


jcommelin (approved):
Thanks :tada: 

bors d+

mattrobball:
bors merge

PR #15157, opened 2024-07-26
[Merged by Bors] - feat(CategoryTheory/Limits): a pullback cone is limit iff it is so after the application of coyoneda.obj

joelriou:
---


erdOne:
Thanks a lot!
maintainer merge

PR #15019, opened 2024-07-22
[Merged by Bors] - feat(Mathlib.RingTheory.MvPowerSeries.Evaluation): evaluation of power series

AntoineChambert-Loir:
Define the evaluation of (Mv)PowerSeries in complete hausdorff linear algebras

Co-authored-by: María Inés de Frutos Fernández

---

- [x] depends on: #14990 
- [x] depends on: #14866 
- [x] depends on: #14865
- [x] depends on: #20958
- [x] depends on: #20971

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15007, opened 2024-07-22
[Merged by Bors] - feat(RingTheory.MvPowerSeries.LinearTopology): a linear topology on the ring of power series

AntoineChambert-Loir:
- `MvPowerSeries.basis`: the ideals of the ring of multivariate power series
all coefficients the exponent of which is smaller than some bound vanish.

- `MvPowerSeries.isLinearTopology` : if `R` has a linear topology,
then the product topology on `MvPowerSeries σ R` is a linear topology,
as witnessed by the ideals from `MvPowerSeries.basis` which form a basis of neighborhoods of `0`
consisting of two-sided ideals.

This applies in particular when `R` has the discrete topology.

Co-authored-by: @mariainesdff 

---

- [x] depends on: #14990 
- [x] depends on: #14866 
[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #14866, opened 2024-07-18
[Merged by Bors] - feat(RingTheory.MvPowerSeries.Topology): define the product topology on mv power series

AntoineChambert-Loir:
Let `R` be with `Semiring R` and `TopologicalSpace R`
In this file we define the topology on `MvPowerSeries σ R`
that corresponds to the simple convergence on its coefficients,
aka the product topology.
It is the coarsest topology for which all coefficients maps are continuous.

When `R` has `UniformSpace R`, we define the corresponding uniform structure.

When the type of coefficients has the discrete topology,
it corresponds to the topology defined by [bourbaki1981], chapter 4, §4, n°2

Coauthored with María Inés de Frutos Fernández @mariainesdff 

---

- [x] depends on: #14865 
[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15158, opened 2024-07-26
[Merged by Bors] - feat(MvPowerSeries/Substitution): substitution of power series inside power series

AntoineChambert-Loir:
Define substitution of power series other power series. This is an application of evaluation, where the ground ring is given the discrete topology. The condition for being defined is that the constant coefficient is nilpotent (and, for infinitely many variables, that the family tends to 0).

Co-authored-by: @mariainesdff 

---

- [x] depends on: #15007
- [x] depends on: #15019
- [x] depends on: #14990 
- [x] depends on: #14866 
- [x] depends on: #14865
- [x] depends on: #20958
- [x] depends on: #23067
- [x] depends on: #23511 


Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+-/
+
+section rescale
+
+/-- Scale multivariate power series -/
+noncomputable def rescale (a : σ → R) (f : MvPowerSeries σ R) :
+    MvPowerSeries σ R :=
+  subst (a • X) f
+
+theorem scale_eq_subst (a : σ → R) (f : MvPowerSeries σ R) :
```
AntoineChambert-Loir, line 515:
```suggestion
theorem rescale_eq_subst (a : σ → R) (f : MvPowerSeries σ R) :
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+    MvPowerSeries.monomial R e (r ^ (Finsupp.sum e fun _ n => n))
+      = (Finsupp.prod e fun s e => (r • MvPowerSeries.X s) ^ e) := by
+  rw [prod_smul_X_eq_smul_monomial_one, ← map_smul, smul_eq_mul, mul_one]
+  simp only [Finsupp.sum, Finsupp.prod, Finset.prod_pow_eq_pow_sum]
+
+section DiscreteUniformity
+
+/-- The discrete uniformity -/
+class DiscreteUniformity (α : Type*) [u : UniformSpace α] : Prop where
+  eq_principal_idRel : uniformity α = Filter.principal idRel
```
ADedecker, line 110: This deserves to be in a much more basic file (and, ideally, in another PR...)

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+  classical
+  simp_rw [tendsto_atTop_nhds] at ha
+  obtain ⟨s, hs⟩ := ha {a} rfl (isOpen_discrete _)
+  apply Set.Finite.subset s.finite_toSet
+  intro b
+  rw [Function.mem_support, not_imp_comm]
+  intro hb
+  let hs' := hs (insert b s) (s.subset_insert b)
+  simp only [Set.mem_singleton_iff] at hs hs'
+  simpa [Finset.sum_insert hb, hs, add_left_eq_self] using hs'
```
ADedecker, line 43:
```suggestion
@[to_additive]
theorem Multipliable.finite_mulSupport
    {α : Type*} [CommGroup α] [TopologicalSpace α] [DiscreteTopology α]
    {β : Type*} (f : β → α) (h : Multipliable f) : Set.Finite f.mulSupport := by
  haveI : IsTopologicalGroup α := ⟨⟩
  exact h.tendsto_cofinite_one (discreteTopology_iff_singleton_mem_nhds.mp ‹_› 1)
```

This should go right after [Summable.tendsto_cofinite_zero](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Algebra/InfiniteSum/Group.html#Summable.tendsto_cofinite_zero)

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+
+theorem IsNilpotent.finsum {α : Type*} [CommSemiring α] {β : Type*} {f : β → α}
+    (hf : ∀ b, IsNilpotent (f b)) :
+    IsNilpotent (finsum f) := by
+  classical
+  by_cases h : Set.Finite f.support
+  · rw [finsum_def, dif_pos h]
+    exact Commute.isNilpotent_sum (fun b _ ↦ hf b) (fun i j hi hj ↦ Commute.all _ _)
+  · simp only [finsum_def, dif_neg h, IsNilpotent.zero]
```
ADedecker, line 52:
```suggestion
  rw [finsum_def]
  split_ifs with h
  · exact Commute.isNilpotent_sum (fun b _ ↦ hf b) (fun i j hi hj ↦ Commute.all _ _)
  · exact IsNilpotent.zero
```

ADedecker: This should go right next to [isNilpotent_sum](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Nilpotent/Basic.html#isNilpotent_sum), and have a `Commute` version like [Commute.isNilpotent_sum](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Nilpotent/Basic.html#Commute.isNilpotent_sum).

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+
+section DiscreteUniformity
+
+/-- The discrete uniformity -/
+class DiscreteUniformity (α : Type*) [u : UniformSpace α] : Prop where
+  eq_principal_idRel : uniformity α = Filter.principal idRel
+
+/-- The bot uniformity is the discrete uniformity -/
+instance (α : Type*) : @DiscreteUniformity α ⊥ := by
+  apply @DiscreteUniformity.mk α ⊥ rfl
```
ADedecker, line 114: Anticipating the review of the auxiliary PR: this shouldn't be an instance, in the same way [discreteTopology_bot](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Order.html#discreteTopology_bot) isn't.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
ADedecker, line 110: I think the definition should rather be `u = ⊥`, like [DiscreteTopology](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Order.html#DiscreteTopology). You can then use [UniformSpace.uniformSpace_eq_bot](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/UniformSpace/Basic.html#UniformSpace.uniformSpace_eq_bot) to get your characterization.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+  rw [discreteTopology_iff_singleton_mem_nhds]
+  intro a
+  rw [UniformSpace.mem_nhds_iff]
+  simp only [Set.subset_singleton_iff, DiscreteUniformity.eq_principal_idRel]
+  simp only [Filter.mem_principal, idRel_subset]
+  use Set.diagonal α
+  simp only [Set.mem_diagonal_iff, implies_true, true_and]
+  intro x
+  simp only [UniformSpace.ball, Set.mem_preimage, Set.mem_diagonal_iff]
+  exact fun a => a.symm
```
ADedecker, line 128: Once the definition is in terms of the lattice of uniform structures, this should be a simple consequence of [UniformSpace.toTopologicalSpace_bot](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/UniformSpace/Basic.html#UniformSpace.toTopologicalSpace_bot).

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+@[to_additive "The discrete uniformity makes an additive group a `UniformAddGroup`"]
+instance {R : Type*} [Group R] [UniformSpace R] [DiscreteUniformity R] :
+    UniformGroup R where
+  uniformContinuous_div := fun s hs ↦ by
+    simp only [uniformity_prod, DiscreteUniformity.eq_principal_idRel, Filter.comap_principal,
+      Filter.inf_principal, Filter.map_principal, Filter.mem_principal, Set.image_subset_iff]
+    rintro ⟨⟨x, y⟩, z, t⟩
+    simp only [Set.mem_inter_iff, Set.mem_preimage, mem_idRel, and_imp]
+    rintro ⟨rfl⟩ ⟨rfl⟩
+    exact mem_uniformity_of_eq hs rfl
```
ADedecker, line 140: This should be split in two parts:
- a product of discrete uniform spaces is again discrete (I'd say you want to use [uniformity_prod_eq_comap_prod](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/UniformSpace/Basic.html#uniformity_prod_eq_comap_prod))
- any function from a discrete uniform space is uniformly continuous

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,1096 @@
...
+      exact (fun a ↦ Set.mem_image_of_mem (fun a => (a, a)) (Set.mem_univ a)))
+    obtain ⟨x, hx⟩ := f_NeBot.nonempty_of_mem ht
+    use x
+    intro s hs
+    apply f.sets_of_superset ht
+    intro y hy
+    convert mem_of_mem_nhds hs
+    apply symm
+    simpa only [d, Set.image_univ, Set.range_diag, Set.mem_diagonal_iff]
+      using ht' (Set.mk_mem_prod hx hy)
```
ADedecker, line 160: This part should be obtained by using (and updating) the code near [UniformSpace.DiscreteUnif.cauchy_le_pure](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/UniformSpace/Cauchy.html#UniformSpace.DiscreteUnif.cauchy_le_pure).

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: I'm not sure, because that forces to name the typeclasses that are used.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: done in the other PR

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: done in the other PR.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: Anyway, I managed to do it in the way you require.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: Why not?

Mathlib/RingTheory/MvPowerSeries/PiTopology.lean
```diff
@@ -127,6 +127,19 @@ theorem continuous_C [Semiring R] :
...
+    ContinuousSMul R (MvPowerSeries σ R) := by
+  suffices (fun (u : R × MvPowerSeries σ R) ↦ (u.1 • u.2 : MvPowerSeries σ R)) =
+    (fun u : MvPowerSeries σ R × MvPowerSeries σ R ↦ u.1 * u.2) ∘ (fun u ↦ ⟨C σ R u.1, u.2⟩) by
+    apply ContinuousSMul.mk
+    rw [this]
+    apply Continuous.comp
+    · exact Continuous.mul continuous_fst continuous_snd
+    · simp only [continuous_prodMk]
+      exact ⟨Continuous.comp continuous_C continuous_fst, continuous_snd⟩
+  ext; simp
```
riccardobrasca, line 141:
```suggestion
instance [Semiring R] [IsTopologicalSemiring R] : ContinuousSMul R (MvPowerSeries σ R) :=
  instContinuousSMulForall
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,977 @@
...
+import Mathlib.Topology.Algebra.Algebra
+import Mathlib.Topology.UniformSpace.DiscreteUniformity
+
+/-! # Substitutions in power series
+
+Here we define the substitution of power series into other power series.
+We follow Bourbaki, Algèbre, chap. 4, §4, n° 3.
+
+Evaluation of a power series at adequate elements has been defined
+in `DividedPowers.ForMathlib.MvPowerSeries.Evaluation.lean`.
```
riccardobrasca, line 23: This probably refers to your repo.

riccardobrasca (commented):
The file is pretty big, but the first declarations I see can be moved to more basic files. Can you please do it, in another PR? You can also but all the prerequisites you need, and feel free to ask my review.

Mathlib/Topology/Algebra/InfiniteSum/Group.lean
```diff
@@ -360,6 +360,13 @@ theorem Multipliable.tendsto_cofinite_one (hf : Multipliable f) : Tendsto f cofi
   refine s.eventually_cofinite_nmem.mono fun x hx ↦ ?_
   · simpa using hs {x} (disjoint_singleton_left.2 hx)
 
+@[to_additive]
```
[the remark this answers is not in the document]
AntoineChambert-Loir: Yes, it was `Summable.finite_support` that I wanted: for the discrete topology, a summable family has finite support. I renamed it as `Summable.finite_support_of_DiscreteTopology`.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,977 @@
...
+    [Semiring S] [Algebra R S] {T : Type*} [Semiring T] [Algebra R T]
+    (φ : S →ₐ[R] T) :
+    MvPowerSeries σ S →ₐ[R] MvPowerSeries σ T where
+  toRingHom   := MvPowerSeries.map σ φ
+  commutes' r := by
+    simp only [RingHom.toMonoidHom_eq_coe, OneHom.toFun_eq_coe, MonoidHom.toOneHom_coe,
+      MonoidHom.coe_coe, MvPowerSeries.algebraMap_apply, map_C, RingHom.coe_coe, AlgHom.commutes]
+
+/-- Change of coefficients in power series, as an `AlgHom` -/
+def PowerSeries.mapAlgHom {R : Type*} [CommSemiring R]
```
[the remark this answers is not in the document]
AntoineChambert-Loir: I still need to do that in the other PR.

Mathlib/Topology/Algebra/InfiniteSum/Group.lean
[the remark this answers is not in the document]
riccardobrasca: I mean that in the multiplicative version there is `mul`, but there is no `add` in the additive version.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,632 @@
...
+variable {a : σ → MvPowerSeries τ S}
+
+lemma coeff_zero_iff :
+    Filter.Tendsto a Filter.cofinite (@nhds _ (@instTopologicalSpace τ S ⊥) 0) ↔
+      ∀ d : τ →₀ ℕ, {s | (a s).coeff S d ≠ 0}.Finite := by
+  letI : UniformSpace S := ⊥
+  simp_rw [tendsto_iff_coeff_tendsto, coeff_zero]
+  apply forall_congr'
+  intro d
+  simp [nhds_discrete]
```
riccardobrasca, line 51:
```suggestion
local instance foo : UniformSpace S := ⊥

lemma coeff_zero_iff :
    Filter.Tendsto a Filter.cofinite (nhds 0) ↔
      ∀ d : τ →₀ ℕ, {s | (a s).coeff S d ≠ 0}.Finite := by
  simp_rw [tendsto_iff_coeff_tendsto, coeff_zero]
  apply forall_congr'
  intro d
  simp [nhds_discrete]
```
If you add this local instance (with a better name and a comment explaining its point) you can remove all `let : UniformSpace ...` from the file. It has the effect that all rings become topological spaces *in this file* (this should be explained clearly), but not in files importing it, so it seems a good compromise.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,632 @@
...
+    @HasEval σ (MvPowerSeries τ S) _ (@instTopologicalSpace τ S ⊥) a :=
+  hasSubst_iff_hasEval_of_discreteTopology.mp ha
+
+theorem hasSubst_X :
+    HasSubst (fun (s : σ) ↦ (X s : MvPowerSeries σ S)) := by
+  letI : UniformSpace S := ⊥
+  rw [hasSubst_iff_hasEval_of_discreteTopology]
+  exact HasEval.X
+
+/-   { const_coeff s := by simp only [constantCoeff_X, IsNilpotent.zero]
```
riccardobrasca, line 76: There a couple of commented results that should be removed/uncommented.

AntoineChambert-Loir: That one shall be.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: No, that seem to work.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,632 @@
...
+  · intro ha
+    exact {
+      hpow s := (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mpr (ha.const_coeff s)
+      tendsto_zero := coeff_zero_iff.mpr ha.coeff_zero }
+  · intro ha
+    exact {
+      const_coeff s := (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mp (ha.hpow s)
+      coeff_zero d := (coeff_zero_iff.mp ha.tendsto_zero) d }
+
+theorem HasSubst.hasEval (ha : HasSubst a) :
```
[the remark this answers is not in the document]
AntoineChambert-Loir: I left it explicit.

AntoineChambert-Loir:
Question to @riccardobrasca : in this file, it is more convenient to add `PowerSeries.HasSubst f` for `IsNilpotent f.constantCoeff`, while for evaluation, it was simple enough to enter `IsTopologicallyNilpotent a`. Maybe, I should restore symmetry and define `PowerSeries.HasEval` to be an abbrev of `IsTopologicallyNilpotent.

riccardobrasca:
Can you remove `PowerSeries.Substitution` from this PR? (Or opening a new one with only the `MvPowerSeries` stuff). This should be very quick and it divides the length by 2.

riccardobrasca:
> Question to @riccardobrasca : in this file, it is more convenient to add `PowerSeries.HasSubst f` for `IsNilpotent f.constantCoeff`, while for evaluation, it was simple enough to enter `IsTopologicallyNilpotent a`. Maybe, I should restore symmetry and define `PowerSeries.HasEval` to be an abbrev of `IsTopologicallyNilpotent.

I am not sure I understand the situation, putting the right discrete topology on the coefficients and the right topology on power series, is `HasSubst` equivalent to `HasEval`?

riccardobrasca:
> > Question to @riccardobrasca : in this file, it is more convenient to add `PowerSeries.HasSubst f` for `IsNilpotent f.constantCoeff`, while for evaluation, it was simple enough to enter `IsTopologicallyNilpotent a`. Maybe, I should restore symmetry and define `PowerSeries.HasEval` to be an abbrev of `IsTopologicallyNilpotent.
> 
> I am not sure I understand the situation, putting the right discrete topology on the coefficients and the right topology on power series, is `HasSubst` equivalent to `HasEval`?

Ah yes, this `hasSubst_iff_hasEval_of_discreteTopology`.

riccardobrasca:
> Question to @riccardobrasca : in this file, it is more convenient to add `PowerSeries.HasSubst f` for `IsNilpotent f.constantCoeff`, while for evaluation, it was simple enough to enter `IsTopologicallyNilpotent a`. Maybe, I should restore symmetry and define `PowerSeries.HasEval` to be an abbrev of `IsTopologicallyNilpotent.

Ah the question is about the one variable case! Let's focus on `MvPowerSeries` to start with.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,433 @@
...
+    {ε : MvPowerSeries τ S →ₐ[R] T}
+
+local instance : ContinuousSMul R T := DiscreteTopology.instContinuousSMul R T
+
+theorem comp_substAlgHom (ha : HasSubst a) (hε : Continuous ε) :
+      ε.comp (substAlgHom ha) = aeval (HasEval.map hε ha.hasEval) :=
+  comp_aeval ha.hasEval hε
+
+theorem comp_subst (ha : HasSubst a) (hε : Continuous ε) :
+    ⇑ε ∘ (subst a) = ⇑(aeval (R := R) (HasEval.map hε ha.hasEval)) := by
```
riccardobrasca, line 245:
```suggestion
    ε ∘ (subst a) = (aeval (R := R) (HasEval.map hε ha.hasEval)) := by
```

AntoineChambert-Loir:
Other question: Here, `MvPowerSeries.rescale` is defined using `subst`. In the (existing) docs#PowerSeries.rescale (for the univariate case) it is done by an explicit formula, which I could add here, and which could even serve as a definition. On the other hand, to get good algebraic properties (beyond add, mul), it would be better to know that it is equal to the one defined as `subst`. What should one do? Should one refactor docs#PowerSeries.rescale using this idea?

riccardobrasca:
> Other question: Here, `MvPowerSeries.rescale` is defined using `subst`. In the (existing) docs#PowerSeries.rescale (for the univariate case) it is done by an explicit formula, which I could add here, and which could even serve as a definition. On the other hand, to get good algebraic properties (beyond add, mul), it would be better to know that it is equal to the one defined as `subst`. What should one do? Should one refactor docs#PowerSeries.rescale using this idea?

The approach with `subst` seems better, and I don't think anyone will complain that it doesn't work for `Semiring` or something. Sooner or later everything relative to `PowerSeries` should be refactored using the `MvPowerSeries` version (if possible), but we can live with the two versions (as what happens for polynomials, where `Polynomial` and `MvPolynomial` are just unrelated).

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+  coeff_zero d : {s | (a s).coeff S d ≠ 0}.Finite
+
+variable {a : σ → MvPowerSeries τ S}
+
+lemma coeff_zero_iff :
+    Filter.Tendsto a Filter.cofinite (@nhds _ (@instTopologicalSpace τ S ⊥) 0) ↔
+      ∀ d : τ →₀ ℕ, {s | (a s).coeff S d ≠ 0}.Finite := by
+  simp_rw [tendsto_iff_coeff_tendsto, coeff_zero]
+  apply forall_congr'
+  intro d
```
riccardobrasca, line 70:
```suggestion
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+  simp [nhds_discrete]
+
+/-- A multivariate power series can be substituted if and only if
+it can be evaluated when the topology on the coefficients of its coefficients
+is the discrete topology. -/
+/- We make the arguments of `HasEval` explicit to be sure that Lean infers the correct topology. -/
```
riccardobrasca, line 76:
```suggestion
it can be evaluated when the topology on the coefficients ring
is the discrete topology. -/
/- We make the arguments of `HasEval` explicit to be sure that Lean infers the correct topology. -/
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+lemma hasSubst_iff_hasEval_of_discreteTopology :
+    HasSubst a ↔ @HasEval σ _ _ (@instTopologicalSpace τ S ⊥) a :=
+  ⟨fun ha ↦
+    {
+      hpow s := (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mpr (ha.const_coeff s)
+      tendsto_zero := coeff_zero_iff.mpr ha.coeff_zero },
+    fun ha ↦
+    {
+      const_coeff s := (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mp (ha.hpow s)
+      coeff_zero d := (coeff_zero_iff.mp ha.tendsto_zero) d }⟩
```
riccardobrasca, line 86:
```suggestion
  ⟨fun ha ↦ ⟨fun s ↦ (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mpr (ha.const_coeff s),
    coeff_zero_iff.mpr ha.coeff_zero⟩, fun ha ↦
      ⟨fun s ↦ (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mp (ha.hpow s),
      fun d ↦ (coeff_zero_iff.mp ha.tendsto_zero) d⟩⟩
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+
+theorem HasSubst.hasEval (ha : HasSubst a) :
+    @HasEval σ (MvPowerSeries τ S) _ (@instTopologicalSpace τ S ⊥) a :=
+  hasSubst_iff_hasEval_of_discreteTopology.mp ha
+
+theorem hasSubst_X :
+    HasSubst (fun (s : σ) ↦ (X s : MvPowerSeries σ S)) := by
```
riccardobrasca, line 93:
```suggestion
theorem hasSubst_X : HasSubst (fun (s : σ) ↦ (X s : MvPowerSeries σ S)) := by
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+  rw [hasSubst_iff_hasEval_of_discreteTopology]
+  exact HasEval.X
+
+theorem hasSubst_zero :
+    HasSubst (fun (_ : σ) ↦ (0 : MvPowerSeries τ S)) := by
+  rw [hasSubst_iff_hasEval_of_discreteTopology]
+  exact HasEval.zero
```
riccardobrasca, line 100:
```suggestion
  simpa [hasSubst_iff_hasEval_of_discreteTopology] using HasEval.X
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+
+theorem hasSubst_add {a b : σ → MvPowerSeries τ S}
+    (ha : HasSubst a) (hb : HasSubst b) :
```
riccardobrasca, line 103:
```suggestion
theorem hasSubst_add {a b : σ → MvPowerSeries τ S} (ha : HasSubst a) (hb : HasSubst b) :
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+    HasSubst (a + b) := by
+  rw [hasSubst_iff_hasEval_of_discreteTopology] at ha hb ⊢
+  exact HasEval.add ha hb
+
+@[simp]
+theorem constantCoeff_smul {R : Type*} [Semiring R] {S : Type*} [Semiring S] [Module R S]
```
riccardobrasca, line 109: This should go to a more basic file.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+    constantCoeff σ S (a • φ) = a • constantCoeff σ S φ :=
+  rfl
+
+theorem hasSubst_mul (b : σ → MvPowerSeries τ S)
+    {a : σ → MvPowerSeries τ S} (ha : HasSubst a) : HasSubst (b * a) := by
+  rw [hasSubst_iff_hasEval_of_discreteTopology] at ha ⊢
+  exact HasEval.mul_left b ha
+
+theorem hasSubst_smul (r : MvPowerSeries τ S) {a : σ → MvPowerSeries τ S} (ha : HasSubst a) :
+    HasSubst (r • a) := by convert hasSubst_mul _ ha
```
riccardobrasca, line 120:
```suggestion
    HasSubst (r • a) := hasSubst_mul _ ha
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+
+/-- Families of mv power series that can be substituted, as an `Ideal` -/
```
riccardobrasca, line 122:
```suggestion
/-- Families of `MvPowerSeries` series that can be substituted, as an `Ideal` -/
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+noncomputable def hasSubst.ideal : Ideal (σ → MvPowerSeries τ S) :=
+  { carrier := setOf HasSubst
+    add_mem' := hasSubst_add
+    zero_mem' := hasSubst_zero
+    smul_mem' := hasSubst_mul }
+
+/-- If σ is finite, then the nilpotent condition is enough for HasSubst -/
```
riccardobrasca, line 129:
```suggestion
/-- If `σ` is finite, then the nilpotent condition is enough for `HasSubst` -/
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+theorem hasSubst_of_constantCoeff_nilpotent [Finite σ]
+    {a : σ → MvPowerSeries τ S} (ha : ∀ s, IsNilpotent (constantCoeff τ S (a s))) :
+    HasSubst a where
+  const_coeff := ha
+  coeff_zero _ := Set.toFinite _
+
+/-- If σ is finite, then having zero constant coefficient is enough for HasSubst -/
```
riccardobrasca, line 136:
```suggestion
/-- If `σ` is finite, then having zero constant coefficient is enough for `HasSubst` -/
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+
+variable {a : σ → MvPowerSeries τ S}
+
+/-- Substitution of power series into a power series -/
+noncomputable def substAlgHom (ha : HasSubst a) :
+    MvPowerSeries σ R →ₐ[R] MvPowerSeries τ S :=
+  MvPowerSeries.aeval ha.hasEval
+
+@[simp]
+theorem coe_substAlgHom (ha : HasSubst a) :
```
riccardobrasca, line 155: I think it makes more sense to have an applied version `coe_substAlgHom_apply`. Something like
```lean
@[simp]
theorem substAlgHom_apply (ha : HasSubst a) (f : MvPowerSeries σ R) :
    substAlgHom ha f = subst a f := by
```
(we can also keep both versions)

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,435 @@
...
+  coeff_zero _ := Set.toFinite _
+
+/-- If σ is finite, then having zero constant coefficient is enough for HasSubst -/
+theorem hasSubst_of_constantCoeff_zero [Finite σ]
+    {a : σ → MvPowerSeries τ S} (ha : ∀ s, constantCoeff τ S (a s) = 0) :
+    HasSubst a :=
+  hasSubst_of_constantCoeff_nilpotent (fun s ↦ by simp only [ha s, IsNilpotent.zero])
+
+/-- Substitution of power series into a power series -/
+noncomputable def subst (a : σ → MvPowerSeries τ S) (f : MvPowerSeries σ R) :
```
riccardobrasca, line 143: Can you clarify in the doc that this is always defined but it is not well behaved in general?

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: indeed.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+    exact (continuous_subst hb).continuousAt
+
+theorem substAlgHom_comp_substAlgHom (ha : HasSubst a) (hb : HasSubst b) :
+    ((substAlgHom hb).restrictScalars R).comp (substAlgHom ha) = substAlgHom (ha.comp hb) := by
+  letI : UniformSpace R := ⊥
+  letI : UniformSpace S := ⊥
+  letI : UniformSpace T := ⊥
+  apply comp_aeval (R := R) (ε := (substAlgHom hb).restrictScalars R) ha.hasEval
+  simp only [AlgHom.coe_restrictScalars', coe_substAlgHom]
+  exact (continuous_subst (R := S) hb)
```
riccardobrasca, line 345:
```suggestion
  simpa using (continuous_subst hb)st (R := S) hb)
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+theorem substAlgHom_comp_substAlgHom_apply (ha : HasSubst a) (hb : HasSubst b)
+    (f : MvPowerSeries σ R) :
+    (substAlgHom hb) (substAlgHom ha f) = substAlgHom (ha.comp hb) f :=
+  DFunLike.congr_fun (substAlgHom_comp_substAlgHom ha hb) f
+
+theorem subst_comp_subst (ha : HasSubst a) (hb : HasSubst b) :
+    (subst b) ∘ (subst a) = subst (R := R) (fun s ↦ subst b (a s)) := by
+  simpa only [funext_iff, coe_substAlgHom, DFunLike.ext_iff,
+    AlgHom.coe_comp, AlgHom.coe_restrictScalars', Function.comp_apply]
+    using substAlgHom_comp_substAlgHom (R := R) ha hb
```
riccardobrasca, line 356:
```suggestion
  simpa [funext_iff, DFunLike.ext_iff] using substAlgHom_comp_substAlgHom ha hb
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+
+variable (R) in
+theorem hasSubst_rescale (a : σ → R) :
+    HasSubst ((a • X) : σ → MvPowerSeries σ R) := by
+  convert hasSubst_mul (fun s ↦ algebraMap R (MvPowerSeries σ R) (a s))
+    hasSubst_X using 1
+  rw [funext_iff]
+  intro s
+  simp only [Pi.smul_apply', Pi.mul_apply]
+  rw [algebra_compatible_smul (MvPowerSeries σ R), smul_eq_mul]
```
riccardobrasca, line 380:
```suggestion
  convert hasSubst_mul (fun s ↦ algebraMap R (MvPowerSeries σ R) (a s)) hasSubst_X
  simp [funext_iff, algebra_compatible_smul (MvPowerSeries σ R)]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+
+variable (R) in
+/-- Rescale multivariate power series, as an `AlgHom` -/
+noncomputable def rescale_algHom (a : σ → R) :
+    MvPowerSeries σ R →ₐ[R] MvPowerSeries σ  R :=
+  substAlgHom (hasSubst_rescale R a)
+
+theorem coe_rescale_algHom (a : σ → R) :
+    ⇑(rescale_algHom R a) = rescale a :=
```
riccardobrasca, line 389:
```suggestion
    rescale_algHom R a = rescale a :=
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+  simp only [AlgHom.coe_comp, Function.comp_apply, rescale_algHom]
+  rw [substAlgHom_comp_substAlgHom_apply]
+  congr
+  rw [funext_iff]
+  intro s
+  simp only [Pi.smul_apply', Pi.mul_apply]
+  rw [AlgHom.map_smul_of_tower]
+  rw [← MvPolynomial.coe_X, substAlgHom_coe, MvPolynomial.aeval_X, MvPolynomial.coe_X]
+  simp only [Pi.smul_apply', algebraMap_smul]
+  rw [← mul_smul, mul_comm]
```
riccardobrasca, line 405:
```suggestion
  ext f
  simp only [AlgHom.coe_comp, Function.comp_apply, rescale_algHom]
  rw [substAlgHom_comp_substAlgHom_apply]
  congr
  rw [funext_iff]
  intro s
  simp only [Pi.smul_apply', Pi.mul_apply]
  rw [AlgHom.map_smul_of_tower, ← MvPolynomial.coe_X, substAlgHom_coe]
  simp [algebraMap_smul, ← mul_smul, mul_comm]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+theorem coeff_rescale (r : σ → R) (f : MvPowerSeries σ R) (d : σ →₀ ℕ) :
+    coeff R d (rescale r f) = (d.prod fun s n ↦ r s ^ n) • coeff R d f := by
+  unfold rescale
+  rw [coeff_subst (hasSubst_rescale R _)]
+  simp only [Pi.smul_apply', smul_eq_mul, prod_smul_X_eq_smul_monomial_one]
+  simp only [LinearMap.map_smul_of_tower, Algebra.mul_smul_comm]
+  rw [finsum_eq_single _ d]
+  · simp only [coeff_monomial_same, mul_one, smul_eq_mul]
+  · intro e he
+    simp only [coeff_monomial_ne he.symm, mul_zero, smul_zero]
```
riccardobrasca, line 420:
```suggestion
  unfold rescale
  rw [coeff_subst (hasSubst_rescale R _)]
  simp only [Pi.smul_apply', smul_eq_mul, prod_smul_X_eq_smul_monomial_one]
  simp only [LinearMap.map_smul_of_tower, Algebra.mul_smul_comm]
  rw [finsum_eq_single _ d]
  · simp
  · intro e he
    simp [coeff_monomial_ne he.symm]
```
Also, can you create a `rfl` lemma to rewrite instead of unfolding `rescale`?

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+
+theorem rescale_one :
+    rescale 1 = @id (MvPowerSeries σ R) := by
+  ext f d
+  simp only [coeff_rescale, Finsupp.prod, Pi.one_apply, one_pow, Finset.prod_const_one, smul_eq_mul,
+    one_mul, id_eq]
```
riccardobrasca, line 426:
```suggestion
theorem rescale_one :
    rescale 1 = id (α := MvPowerSeries σ R) := by
  ext f d
  simp [coeff_rescale, Finsupp.prod]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+
+theorem rescale_algHom_one :
+    rescale_algHom R 1 = AlgHom.id R (MvPowerSeries σ R):= by
+  rw [DFunLike.ext_iff]
+  intro f
+  simp only [Function.const_one, coe_rescale_algHom, AlgHom.coe_id, id_eq, rescale_one]
```
riccardobrasca, line 432:
```suggestion
  simp [Function.const_one, coe_rescale_algHom, rescale_one]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+    rw [← rescale_algHom_comp, AlgHom.End_toSemigroup_toMul_mul]
+
+theorem rescale_zero_apply (f : MvPowerSeries σ R) :
+    rescale 0 f = MvPowerSeries.C σ R (constantCoeff σ R f) := by
+  classical
+  ext d
+  simp only [coeff_rescale, coeff_C]
+  by_cases hd : d = 0
+  · simp only [hd, Pi.zero_apply, Finsupp.prod_zero_index, coeff_zero_eq_constantCoeff,
+      smul_eq_mul, one_mul, ↓reduceIte]
```
riccardobrasca, line 449:
```suggestion
  · simp [hd]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+  · simp only [Pi.zero_apply, smul_eq_mul, if_neg hd]
+    convert zero_smul R _
+    simp only [DFunLike.ext_iff, Finsupp.coe_zero, Pi.zero_apply, not_forall] at hd
+    obtain ⟨s, hs⟩ := hd
+    apply Finset.prod_eq_zero (Finsupp.mem_support_iff.mpr hs)
+    simp only [Function.const_apply, zero_pow hs]
```
riccardobrasca, line 455:
```suggestion
    simp [hs]
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+  simp only [coeff_rescale, Finsupp.prod, Pi.one_apply, one_pow, Finset.prod_const_one, smul_eq_mul,
+    one_mul, id_eq]
+
+theorem rescale_algHom_one :
+    rescale_algHom R 1 = AlgHom.id R (MvPowerSeries σ R):= by
+  rw [DFunLike.ext_iff]
+  intro f
+  simp only [Function.const_one, coe_rescale_algHom, AlgHom.coe_id, id_eq, rescale_one]
+
+/-- Scale mv power series, as a `MonoidHom` in the scaling parameters -/
```
riccardobrasca, line 434:
```suggestion
/-- Scale a power series, as a `MonoidHom` in the scaling parameters -/
```

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+lemma rescale_linear_eq_smul (r : R) (f : MvPowerSeries σ R)
+    (hf : ∀ (d : σ →₀ ℕ), (d.sum (fun _ n ↦ n) ≠ 1) → MvPowerSeries.coeff R d f = 0) :
+    MvPowerSeries.rescale (Function.const σ r) f = r • f := by
+  ext e
+  simp only [MvPowerSeries.coeff_rescale, map_smul]
+  simp only [Finsupp.prod, Function.const_apply, Finset.prod_pow_eq_pow_sum, smul_eq_mul]
+  by_cases he : Finsupp.sum e (fun _ n ↦ n) = 1
+  · simp only [Finsupp.sum] at he
+    simp only [he, pow_one]
+  · simp only [hf e he, mul_zero]
```
riccardobrasca, line 467:
```suggestion
  ext e
  simp only [MvPowerSeries.coeff_rescale, map_smul, Finsupp.prod, Function.const_apply,
    Finset.prod_pow_eq_pow_sum, smul_eq_mul]
  by_cases he : Finsupp.sum e (fun _ n ↦ n) = 1
  · simp only [Finsupp.sum] at he
    simp [he]
  · simp [hf e he]
```

riccardobrasca (approved):
Sorry for the (very) large number of comments, they're almost all trivial and you can just accept them, everything should work.

It almost look good to me, just a couple of more serious comments:
* it's not clear to me if a user should use `subst` or `substAlgHom` in practice. If the user is supposed to use `subst` all the time, please make a `simp` lemma that transform `substAlgHom ha f` into `subst f` (and anyway explain this in the doc). Same for `rescale`.
* do you know what's happening with the `convert coe_aeval _` at line 170? I am going to investigate.
* same for the `erw` at line 241.

riccardobrasca:
Concerning the `convert` I suggest to write something like
```lean
theorem substAlgHom_eq_aeval
    [UniformSpace R] [DiscreteUniformity R] [UniformSpace S] [DiscreteUniformity S]
    (ha : HasSubst a) :
    (substAlgHom ha : MvPowerSeries σ R → MvPowerSeries τ S) = MvPowerSeries.aeval ha.hasEval := by
  simp only [substAlgHom, coe_aeval ha.hasEval]
  convert coe_aeval (R := R) (hasSubst_iff_hasEval_of_discreteTopology.mp ha) <;>
  exact DiscreteUniformity.eq_bot.symm
```
with a comment explaining what's going on. In practice I think the usage of this lemma should be discouraged (even if we want it) since morally `substAlgHom` is purely algebraic while `aeval` involves topological stuff.

riccardobrasca:
OK I am happy with everything in the PR, thanks a lot for the patience!

bors d+

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+lemma hasSubst_iff_hasEval_of_discreteTopology [TopologicalSpace S] [DiscreteTopology S] :
+    HasSubst a ↔ HasEval a :=
+  ⟨fun ha ↦ ⟨fun s ↦ (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mpr (ha.const_coeff s),
+      coeff_zero_iff.mpr ha.coeff_zero⟩,
+    fun ha ↦ ⟨fun s ↦ (tendsto_pow_of_constantCoeff_nilpotent_iff (a s)).mp (ha.hpow s),
+      fun d ↦ (coeff_zero_iff.mp ha.tendsto_zero) d⟩⟩
+
+theorem HasSubst.hasEval [TopologicalSpace S] [DiscreteTopology S] (ha : HasSubst a) :
+    HasEval a :=
+  hasSubst_iff_hasEval_of_discreteTopology.mp ha
```
ADedecker, line 93: A random thought that just occured to me: this is still true if `S` has any linear topology, right? After all, `HasEval` for some topology should imply `HasEval` for any weaker topology, and when the topology on `S` gets weaker the associated one on the ring of power series certainly does as well.

So, is there any interest in making this result (and some of the following) work for any linearly topologized `S`? For example, the lemmas linking evaluation and substitution should definitely still hold in that context.

On the other hand, one could argue that if one works with a weaker topology then one should not use `subst` at all, because it discards as "not well-defined" some substitutions that would be well-defined for a weaker topology.

What do you think? Does this make any sense? Of course I'm happy to leave this discussion for later, no need to block the PR.

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
```diff
@@ -0,0 +1,471 @@
...
+
+section rescale
+
+/-- Rescale multivariate power series -/
+noncomputable def rescale (a : σ → R) (f : MvPowerSeries σ R) :
+    MvPowerSeries σ R :=
+  subst (a • X) f
+
+theorem rescale_eq_subst (a : σ → R) (f : MvPowerSeries σ R) :
+    rescale a f = subst (a • X) f := rfl
```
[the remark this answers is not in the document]
AntoineChambert-Loir: isn't it 4 blank chars?

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: it existed already!

Mathlib/RingTheory/MvPowerSeries/Substitution.lean
AntoineChambert-Loir: You're right, and I know I've already proved it somewhere, but.I have no idea where…

AntoineChambert-Loir: it was tricky.

PR #15160, opened 2024-07-26
refactor(CategoryTheory): making morphisms in Fullsubcategory a 1-field structure

joelriou:
---


joelriou:
Probably too much bit rot in this PR. This will have to be attemped again later.

PR #15161, opened 2024-07-26
feat(Linter): flag `intros x y` which can be replaced by `intro x y`

adomani:
---


grunweg:
Coming here from PR triage, my impression is that this PR can be closed. Is this correct?

adomani:
This linter checks that `intro` is used when `intros` is present.  It may have its value, no?  I am not sure if the linter still needs work, though.

grunweg:
Ah, I misunderstood (and believe this to be similar to the forallIntro linter). This sounds like it might be useful (perhaps paired with an automatic refactoring).

PR #15167, opened 2024-07-26
feat (Analysis/NormedSpace/Equidecomposability): Definition of Equidecomposability

Bergschaf:
---

### DIFF
--- a/Mathlib/Analysis/Equidecomposability.lean
+++ b/Mathlib/Analysis/Equidecomposability.lean
@@ -1,7 +1,9 @@
 import Mathlib.Analysis.NormedSpace.AffineIsometry
 import Mathlib.Order.Partition.Finpartition
 
-
+/-
+Definition of Equidecomposability
+-/
 open Function
 
 variable (𝕜 : Type*) {V: Type*}(P: Type*)
--- a/Mathlib/Analysis/Equidecomposability.lean
+++ b/Mathlib/Analysis/Equidecomposability.lean
@@ -1,3 +1,9 @@
+/-
+Copyright (c) 2024 Chiara Cimino. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chiara Cimino, Christian Krause, Victoria Schleis
+-/
+
 import Mathlib.Analysis.NormedSpace.AffineIsometry
 import Mathlib.Order.Partition.Finpartition
 
--- a/Mathlib/Analysis/Equidecomposability.lean
+++ b/Mathlib/Analysis/Equidecomposability.lean
@@ -1,7 +1,7 @@
 import Mathlib.Analysis.NormedSpace.AffineIsometry
 import Mathlib.Order.Partition.Finpartition
 
-/-
+/-!
 Definition of Equidecomposability
 -/
 open Function
@@ -10,7 +10,7 @@ variable (𝕜 : Type*) {V: Type*}(P: Type*)
   [NormedField 𝕜]
   [SeminormedAddCommGroup V] [NormedSpace 𝕜 V] [PseudoMetricSpace P] [NormedAddTorsor V P]
 
-/-
+/--
 Two subsets B₁ B₂ of a pseudometric space P are equidecomposable if B₁ and B₂ can be divided into
 finitely many affinely isometric parts.
 -/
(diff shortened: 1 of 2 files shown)


Bergschaf:
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Banach.20Tarski.20Paradox.20in.20Lean

Felix-Weilacher:
I said this on Zulip but have not gotten a response. This is not a general enough definition of equidecomposability, and even in the general setting, I suspect that involving `Finpartitions` is not the right way to go.

Bergschaf:
I read your response on zulip but I wanted to make sure that I understood everything correctly and tested it before I just copied your solution.  This took some time because I'm currently quite busy and I'm not that good at these math topics.

Felix-Weilacher:
> I read your response on zulip but I wanted to make sure that I understood everything correctly and tested it before I just copied your solution. This took some time because I'm currently quite busy and I'm not that good at these math topics.

If you like, I can PR some general theory about equidecomposability and let you focus on the specific case of Banach-Tarski

Bergschaf:
That would be very nice, I'd appreciate that a lot.

Mathlib/Analysis/Equidecomposability.lean
```diff
@@ -0,0 +1,28 @@
...
+
+variable (𝕜 : Type*) {V: Type*}(P: Type*)
+  [NormedField 𝕜]
+  [SeminormedAddCommGroup V] [NormedSpace 𝕜 V] [PseudoMetricSpace P] [NormedAddTorsor V P]
+
+/--
+Two subsets B₁ B₂ of a pseudometric space P are equidecomposable if B₁ and B₂ can be divided into
+finitely many affinely isometric parts.
+-/
+def equidecomposable (B₁ B₂ : Set P) : Prop :=
```
urkud, line 23:
```suggestion
def Equidecomposable (B₁ B₂ : Set P) : Prop :=
```

urkud: Do you need the sets to be in the same space?

Mathlib/Analysis/Equidecomposability.lean
```diff
@@ -0,0 +1,28 @@
...
+  ∃ (P_B₁ : Finpartition B₁), ∃ (P_B₂ : Finpartition B₂), ∃ (f : P_B₁.parts → P_B₂.parts),
+    P_B₂.parts.val.sizeOf = P_B₁.parts.val.sizeOf ∧
```
urkud, line 25: You don't need this assumption, because you require `f` to be a bijection anyway.

Mathlib/Analysis/Equidecomposability.lean
```diff
@@ -0,0 +1,28 @@
...
+  Bijective f ∧ (∀ p_b₁ : P_B₁.parts, ∃ (a : AffineIsometry 𝕜 P P),
```
urkud, line 26: The part starting with the inner `Exists` deserves to be a separate `def`.

- Do you want to require `AffineIsometry`, or just an `Isometry`?
- Do you want `a` to be bijective? If yes, then `IsometryEquiv.toRealAffineIsometryEquiv` says, that for real normed torsors the previous question is irrelevant.

Mathlib/Analysis/Equidecomposability.lean
```diff
@@ -0,0 +1,28 @@
...
+    ∀ x ∈ p_b₁.val, ∃ y ∈ (f p_b₁).val, a.toFun x = y ∧
```
urkud, line 27: This says `Set.MapsTo a p_b₁ (f p_b₁)`

Mathlib/Analysis/Equidecomposability.lean
```diff
@@ -0,0 +1,28 @@
...
+    ∀ y ∈ (f p_b₁).val, ∃ x ∈ p_b₁.val, a.toFun x = y)
```
urkud, line 28: These two lines say `a '' p_b₁ = f p_b₁`.

urkud:
Generally, we don't accept bare definitions without a minimal theory around them, so I'm marking this as `awaiting-author`. In this case, the minimal theory should include (at least)
- proofs of reflexivity, symmetry, and transitivity;
- proof of `Equidecomposable K s (e '' s)`, where `e` is an `AffineEquiv`;
- equidecomposable sets have the same cardinality;
- two finite sets of the same cardinality are equidecomposable.

Felix-Weilacher:
> That would be very nice, I'd appreciate that a lot.

#16282

PR #15173, opened 2024-07-26
[Merged by Bors] - chore: use double backticks for explicit names, if possible

grunweg:
This PR is very likely not exhaustive: the next step to finding these more exhaustively would be to write an intelligent linter.
(I found these using a simple text-based linter, which has still too many false positive).

---

### DIFF
--- a/Mathlib/Tactic/Widget/GCongr.lean
+++ b/Mathlib/Tactic/Widget/GCongr.lean
@@ -19,7 +19,7 @@ open Lean Meta Server ProofWidgets
 def makeGCongrString (pos : Array Lean.SubExpr.GoalsLocation) (goalType : Expr)
     (_ : SelectInsertParams) : MetaM (String × String × Option (String.Pos × String.Pos)) := do
 let subexprPos := getGoalLocations pos
-unless goalType.isAppOf ``LE.le || goalType.isAppOf ``LT.lt || goalType.isAppOf ``Int.ModEq do
+unless goalType.isAppOf ``LE.le || goalType.isAppOf ``LT.lt || goalType.isAppOf `Int.ModEq do
   panic! "The goal must be a ≤ or < or ≡."
 let mut goalTypeWithMetaVars := goalType
 for pos in subexprPos do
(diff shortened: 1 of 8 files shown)


Mathlib/Tactic/Widget/GCongr.lean
```diff
@@ -3,6 +3,7 @@ Copyright (c) 2023 Patrick Massot. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Patrick Massot
 -/
+import Mathlib.Algebra.ModEq
```
grunweg, line 6: This import is needed so the name `Int.ModEq` is recognised.

joneugster: Am I right that the widget works fine if `Int.ModEq` is unknown? In that case I'm probably not in favour of adding the import which almost doubles the number of files this depends on.

(although, I also think it doesn't matter too much since it's a widget and will not be imported anywhere)

grunweg: Good question - I don't know!

joneugster: Pretty sure that's how it works, as `goalType.isAppOf` works just on `Expr`.

grunweg: :+1: I have reverted that change.

scripts/noshake.json
```diff
@@ -203,6 +203,7 @@
   ["Mathlib.Topology.Algebra.UniformRing"],
   "Mathlib.Tactic.Widget.SelectPanelUtils":
   ["ProofWidgets.Component.OfRpcMethod"],
+  "Mathlib.Tactic.Widget.GCongr": ["Mathlib.Algebra.ModEq"],
```
joneugster, line 206: I don't know how shake works, is this still needed?

joneugster:
Thanks!

maintainer merge

scripts/noshake.json
grunweg: No, it is not!
```suggestion

PR #15179, opened 2024-07-26
[Merged by Bors] - fix: add check for existing filepath

adomani:
This should fix a problem with the online lean server

[Reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/reading.20the.20.22visible.22.20file.2C.20rather.20than.20the.20saved.20one/near/454337536)

---


semorrison:
bors merge

PR #15186, opened 2024-07-27
[Merged by Bors] - feat: generalize `Int.cast_nonneg` and related lemmas

j-loreaux:
---

If this benchmarks fine, I think it's a good change. See also [this comment](https://github.com/leanprover-community/mathlib4/pull/13974#discussion_r1680054076) on #13974

### DIFF
--- a/Mathlib/Algebra/Order/Ring/Cast.lean
+++ b/Mathlib/Algebra/Order/Ring/Cast.lean
@@ -24,7 +24,7 @@ open Function Nat
 variable {R : Type*}
 
 namespace Int
-section OrderedRing
+section OrderedAddCommGroupWithOne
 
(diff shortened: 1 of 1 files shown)


j-loreaux:
!bench

Mathlib/Algebra/Order/Ring/Cast.lean
```diff
@@ -25,14 +25,16 @@ variable {R : Type*}
 
 namespace Int
 section OrderedRing
```
riccardobrasca, line 27: Can you change the name of the section please?

riccardobrasca (approved):
Does it make sense to move this to another file? Maybe it is not worth the trouble, so don't worry too much about this.

Thanks!

bors d+

j-loreaux:
bors merge

j-loreaux:
> Does it make sense to move this to another file? Maybe it is not worth the trouble, so don't worry too much about this.

I don't think it's worth it.

PR #15189, opened 2024-07-27
[Merged by Bors] - chore: removing unused Mathport files

kim-em:
This does not yet remove `Mathport.Syntax`.

We may want to preserve the information about unported tactics implicitly stored there (although this information is quite messy, in large part out of date, and in any case preserved in the git history if anyone wants to do the archaeology).


urkud:
Thanks! :tada:
bors merge

PR #15190, opened 2024-07-27
[Merged by Bors] - fix: re-add the `TextBased` long line linter

adomani:
As pointed out on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Long.20lines.20linter.20regression), the `TextBased` long line linter should not have been removed.

There are two advantages to having also the `TextBased` linter:
* if you accidentally submit code with long lines, the `TextBased` linter will tell you of all long lines at once;
* the linter can only flag lines on files that import the linter, while the `TextBased` linter crawls all `.lean` files in `Mathlib`.

---


semorrison:
bors merge

PR #13795, opened 2024-06-13
perf(Algebra/{Group, GroupWithZero, Ring}/InjSurj): reduce everything

astrainfinita:
---

- [ ] depends on: #15192
- [ ] depends on: #15476

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15192, opened 2024-07-27
[Merged by Bors] - feat: add some term elaborators for reduction

astrainfinita:
In #13795, `zeta%` and `reduceProj%` will be used to improve the performance of unifying some instances.

---


Mathlib/Util/Term/Basic.lean
```diff
@@ -0,0 +1,47 @@
...
+/-- `delta% t` elaborates to a head-delta reduced version of `t`. -/
+elab "delta% " t:term : term <= expectedType => do
+  let t ← withSynthesize do
+    elabTermEnsuringType t expectedType
+  synthesizeSyntheticMVars
+  let t ← instantiateMVars t
+  let some t ← delta? t | throwError "cannot delta reduce {t}"
+  pure t
+
+/-- `zeta% t` elaborates to a head-zeta reduced version of `t`. -/
```
kmill, line 28: `Meta.zetaReduce` completely zeta reduces `t`, not just head `let`s. It also does zeta-delta reduction.

Could you please correct this docstring?

Mathlib/Util/Term/Basic.lean
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Mathlib.Util.Term.Basic
+
+open Lean Elab Term Meta
+
+/-- `delta% t` elaborates to a head-delta reduced version of `t`. -/
+elab "delta% " t:term : term <= expectedType => do
+  let t ← withSynthesize do
+    elabTermEnsuringType t expectedType
+  synthesizeSyntheticMVars
```
kmill, line 23: Any reason you changed the original implementation? I see two changes:
1. `elabTerm` -> `elabTermEnsuringType`
2. this is wrapped in `withSynthesize`.

I don't think using `elabTermEnsuringType` is justified, for any of these term elaborators.

`withSynthesize` is arguably useful, and it does look like you have a test case exercising it. Still, it's rather strong, and it might be better configuring it with `PostponeBehavior.partial`.

Mathlib/Util/Term/Beta.lean
```diff
@@ -16,7 +16,7 @@ The purpose of this is to support substitutions in notations such as
 `∀ i, beta% p i` so that `p i` gets beta reduced when `p` is a lambda.
 -/
 
-namespace Mathlib.Util.TermBeta
+namespace Mathlib.Util.Term.Beta
```
kmill, line 19: Organizationally, I think it would be better to have all of these in the same file. How about `Mathlib.Util.TermReduce` or similar?

Mathlib/Util/Term/Basic.lean
```diff
@@ -0,0 +1,47 @@
...
+/-!
+# Basic term elaborators
+-/
+
+namespace Mathlib.Util.Term.Basic
+
+open Lean Elab Term Meta
+
+/-- `delta% t` elaborates to a head-delta reduced version of `t`. -/
+elab "delta% " t:term : term <= expectedType => do
```
kmill, line 20: The `<= expectedType` syntax causes postponement if the expected type is not known yet. Is that what we want?

The reason the `beta%` elaborator is written the way it is is to elaborate immediately.

kmill (commented):
Could you please make the PR title and summary more informative too?

joneugster:
Thanks!

maintainer merge

kim-em:
bors merge

PR #15195, opened 2024-07-27
[Merged by Bors] - docs: add module docs for `declarations_diff`

adomani:
This PR adds documentation to the `declarations_diff` script.

Consequence of #15149.

---

### DIFF
--- a/scripts/declarations_diff.sh
+++ b/scripts/declarations_diff.sh
@@ -33,7 +33,7 @@ with namespacing, e.g. you could see it with `++-- map_zero`.
 
 The script uses some heuristics to guide this process.
 * It assumes that the keyword (as above) appear on the same line as the corresponding
-  declaration id -- a line break between `theorem` and `riemannHypothesis` fools the script.
+  declaration id --- a line break between `theorem` and `riemannHypothesis` fools the script.
 * It deals with declaration modifiers (such as `noncomputable`, `nonrec`, `protected`) and
   attributes.
 * It is "aware" of "nameless" `instance`s and, rather than looking for a declaration id,
--- a/scripts/declarations_diff.sh
+++ b/scripts/declarations_diff.sh
@@ -18,6 +18,8 @@ and tries to find exact matches between a removed line and an added line
 Exact matches are removed.
 Among the remaining unmatched lines, the script tries to extract the declaration ids and tries to
 find exact matches among those.
+No effort is made to keep track of namespaces: if the declaration is `theorem abc ...`, then
+`abc` is extracted, even if this happens inside `namespace xyz`.
 If a declaration id is added once and removed once, then they are again considered paired and are
 not shown.
 
@@ -30,8 +32,8 @@ This means that the declaration `thmName` was added 3 times and removed twice: t
 with namespacing, e.g. you could see it with `++-- map_zero`.
 
 The script uses some heuristics to guide this process.
-* It assumes that the keyword above appear on the same line as the corresponding declaration id --
-  a line break between `theorem` and `riemannHypothesis` fools the script.
+* It assumes that the keyword (as above) appear on the same line as the corresponding
+  declaration id -- a line break between `theorem` and `riemannHypothesis` fools the script.
 * It deals with declaration modifiers (such as `noncomputable`, `nonrec`, `protected`) and
   attributes.
 * It is "aware" of "nameless" `instance`s and, rather than looking for a declaration id,
--- a/scripts/declarations_diff.sh
+++ b/scripts/declarations_diff.sh
@@ -162,7 +162,7 @@ printf $'<details>
 ## more verbose report:
 ./scripts/declarations_diff.sh long <optional_commit>
 ```
-The doc-module for `script/declarations_diff.sh` contain some details about this script.
+The doc-module for `script/declarations_diff.sh` contains some details about this script.
 </details>'
  : <<ReferenceTest
 theorem oh hello
--- a/scripts/declarations_diff.sh
+++ b/scripts/declarations_diff.sh
@@ -162,8 +162,9 @@ printf $'<details>
 ## more verbose report:
 ./scripts/declarations_diff.sh long <optional_commit>
 ```
-The doc-module for `script/declarations_diff.sh` contains some details about this script.
-</details>'
+</details>
+The doc-module for `script/declarations_diff.sh` contains some details about this script.'
+
  : <<ReferenceTest
 theorem oh hello
 inductive triggers the count even if it is not lean code
--- a/scripts/declarations_diff.sh
+++ b/scripts/declarations_diff.sh
@@ -20,8 +20,8 @@ Among the remaining unmatched lines, the script tries to extract the declaration
 find exact matches among those.
 No effort is made to keep track of namespaces: if the declaration is `theorem abc ...`, then
 `abc` is extracted, even if this happens inside `namespace xyz`.
-If a declaration id is added once and removed once, then they are again considered paired and are
-not shown.
+If a declaration id is added exactly once *and* removed exactly once,
+then they are again considered paired and are not shown.
 
 If a declaration id is either only added or removed or it is added or removed more than once, then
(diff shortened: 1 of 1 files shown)


adomani:
@riccardobrasca @semorrison I am pinging you, since the motivation for writing this doc-module came from your comments on #15149.

scripts/declarations_diff.sh
```diff
@@ -1,5 +1,44 @@
...
+If a declaration id is added once and removed once, then they are again considered paired and are
+not shown.
+
+If a declaration id is either only added or removed or it is added or removed more than once, then
+the script will return a count such as
+
+++--+ thmName
+
+This means that the declaration `thmName` was added 3 times and removed twice: this can happen
+with namespacing, e.g. you could see it with `++-- map_zero`.
```
riccardobrasca, line 30:
```suggestion
with namespacing, e.g. you could see it with `+++-- map_zero`.
```
With `++--` it would be silent, right?

adomani: No, it is silent exactly with one of `+-` or `-+`.  As soon as there are more matches (or exactly one match), it reports everything

scripts/declarations_diff.sh
```diff
@@ -1,5 +1,44 @@
 #!/usr/bin/env bash
 
+ : <<'BASH_DOC_MODULE'
+
+#  The `declarations_diff` script
+
+The `declarations_diff` script is a text-based script that attempts to find which declarations
+have been removed and which declarations have been added in the current PR with respect to `master`.
+
+In essence, it looks at the output of `git diff origin/master...HEAD`, it extracts the lines that
```
grunweg, line 10:
```suggestion
In essence, it looks at the output of `git diff origin/master...HEAD`, extracts the lines that

File: scripts/declarations_diff.sh [hunk 3]
```diff
@@ -1,5 +1,44 @@
...
+find exact matches among those.
+If a declaration id is added once and removed once, then they are again considered paired and are
+not shown.
+
+If a declaration id is either only added or removed or it is added or removed more than once, then
+the script will return a count such as
+
+++--+ thmName
+
+This means that the declaration `thmName` was added 3 times and removed twice: this can happen
```
grunweg, line 29:
```suggestion
This means that a declaration `thmName` was added 3 times and removed twice: this can happen

File: scripts/declarations_diff.sh [hunk 4]
```diff
@@ -1,5 +1,44 @@
...
+In essence, it looks at the output of `git diff origin/master...HEAD`, it extracts the lines that
+contain one of the keywords
+
+`theorem` `lemma` `inductive` `structure` `def` `class` `instance` `alias`
+
+and tries to find exact matches between a removed line and an added line
+(e.g. when moving a declaration from one file to another or restructuring within the same file).
+
+Exact matches are removed.
+Among the remaining unmatched lines, the script tries to extract the declaration ids and tries to
```
grunweg, line 19: You could mention namespacing already up here, say like "the declaration ids as given by the diff --- in particular, not considering namespacing". What do you think?

scripts/declarations_diff.sh
```diff
@@ -1,5 +1,44 @@
...
+If a declaration id is either only added or removed or it is added or removed more than once, then
+the script will return a count such as
+
+++--+ thmName
+
+This means that the declaration `thmName` was added 3 times and removed twice: this can happen
+with namespacing, e.g. you could see it with `++-- map_zero`.
+
+The script uses some heuristics to guide this process.
+* It assumes that the keyword above appear on the same line as the corresponding declaration id --
```
grunweg, line 33: It took me a moment to parse this sentence: is this one clearer to you?
```
* It assumes that the keyword (as above) appears on the same line as the corresponding declaration id --- for instance, a line break between `theorem` and `riemannHypothesis` fools the script.

File: scripts/declarations_diff.sh [hunk 6]
```diff
@@ -1,5 +1,44 @@
 #!/usr/bin/env bash
 
+ : <<'BASH_DOC_MODULE'
```
grunweg, line 3: Do you know where this is documented? I'm unable to find this in the bash manual.

grunweg (commented):
I also gave this a look - a few minor suggestions only. Thanks for documenting this!

scripts/declarations_diff.sh
adomani: This is a trick that I learned by googling for multiline comments in bash.  The colon `:` is a no-op command and `<<TEXT...TEXT` is how you can pass the command inside `TEXT` to a previous one, similar to a redirection: you redirect random text as an argument to a command that does nothing with it.  Finally, the single quotes around the first `TEXT` avoid some parsing of the intervening text, but I forgot exactly what.

scripts/declarations_diff.sh
adomani: Good point: I added a sentence mentioning that the script ignores namespaces.

scripts/declarations_diff.sh
adomani: Thanks: your formulation does seem clearer.

adomani (commented):
Michael, thank you for your comments: I implemented them all!

scripts/declarations_diff.sh
```diff
@@ -1,5 +1,46 @@
...
+If a declaration id is either only added or removed or it is added or removed more than once, then
+the script will return a count such as
+
+++--+ thmName
+
+This means that a declaration `thmName` was added 3 times and removed twice: this can happen
+with namespacing, e.g. you could see it with `++-- map_zero`.
+
+The script uses some heuristics to guide this process.
+* It assumes that the keyword (as above) appear on the same line as the corresponding
```
grunweg, line 35:
```suggestion
* It assumes that the keyword (as above) appear on the same line as the corresponding

File: scripts/declarations_diff.sh (hunk 6 above)
Comment (grunweg) [reply scripts/declarations_diff.sh:3]: Thanks for explaining. In essence, `BASH_DOC_MODULE` is not a magic variable (you could have written `DamianoWritesRandomStuff` instead), has no further effect - and this is basically a multi-line comment?

File: scripts/declarations_diff.sh [hunk 8]
```diff
@@ -1,5 +1,46 @@
...
+  declaration id -- a line break between `theorem` and `riemannHypothesis` fools the script.
```
grunweg, line 36:
```suggestion

  declaration id --- a line break between `theorem` and `riemannHypothesis` fools the script.

File: scripts/declarations_diff.sh [hunk 9]
```diff
@@ -1,5 +1,46 @@
...
+
+and tries to find exact matches between a removed line and an added line
+(e.g. when moving a declaration from one file to another or restructuring within the same file).
+
+Exact matches are removed.
+Among the remaining unmatched lines, the script tries to extract the declaration ids and tries to
+find exact matches among those.
+No effort is made to keep track of namespaces: if the declaration is `theorem abc ...`, then
+`abc` is extracted, even if this happens inside `namespace xyz`.
+If a declaration id is added once and removed once, then they are again considered paired and are
```
grunweg, line 23: Would writing "exactly once" here alleviate Riccardo's concern below?

grunweg (commented):
Two last, very easy, comments - feel free to maintainer merge on my behalf once you have considered them.

riccardobrasca:
Thanks!

bors d+

scripts/declarations_diff.sh
adomani: Yes, exactly: any text that does not "randomly appear" in what follows, until you repeat it gets ignored.  I chose `BASH_DOC_MODULE` to communicate what the intent of the following text was.

scripts/declarations_diff.sh
adomani: Done!

scripts/declarations_diff.sh
adomani: I also changed `the keyword` to `each keyword: I am not committing this suggestion to avoid merge conflicts.

PR #15201, opened 2024-07-27
[Merged by Bors] - feat: Rename `CompactlyGeneratedSpace` to `UCompactlyGeneratedSpace`

EtienneC30:
`CompactlyGeneratedSpace` is defined in `Mathlib.Topology.Category.CompactlyGenerated`. Due to functors considerations, the definition involves a free universe parameter, which is not desirable for topological purposes. Therefore we rename `CompactlyGeneratedSpace` to `UCompactlyGeneratedSpace` so that `CompactlyGeneratedSpace` can be defined as a special case. This was discussed [here](https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Compactly.20generated.20spaces/near/453957506).

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #15202, opened 2024-07-27
[Merged by Bors] - feat: The subtype coercion associated to a closed subset is a closed map

EtienneC30:
Prove that the subtype coercion associated to a closed subset is a closed map.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15176, opened 2024-07-26
[Merged by Bors] - feat: ULift of a compact space is compact.

EtienneC30:
`ULift` of a compact space is compact.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15203, opened 2024-07-27
[Merged by Bors] - feat(Topology): add API for compactly generated spaces

EtienneC30:
Define compactly generated topological spaces, which are spaces whose topology is coinduced by continuous maps whose domain is compact Hausdorff.
Prove basic properties and instances.
Provide instances for sequential spaces and Hausdorff weakly locally compact spaces.

---

- [x] depends on: #15201
- [x] depends on: #15202
- [x] depends on: #15176


Mathlib/Topology/Compactness/CompactlyGeneratedSpace.lean
```diff
@@ -0,0 +1,177 @@
...
+  refine compactlyGeneratedSpace_of_isClosed fun s h' ↦ isOpen_compl_iff.1 <| h _ fun f hf ↦ ?_
+  rw [preimage_compl, isOpen_compl_iff]
+  exact h' f hf
+
+/-- In a compactly generated space `X`, a set `s` is closed when `f ⁻¹' s` is
+closed for every continuous map `f : K → X`, where `K` is compact Hausdorff. -/
+theorem CompactlyGeneratedSpace.isClosed' [CompactlyGeneratedSpace X] {s : Set X}
+    (hs : ∀ {K : Type u} [TopologicalSpace K], [CompactSpace K] → [T2Space K] →
+      ∀ (f : K → X), Continuous f → IsClosed (f ⁻¹' s)) : IsClosed s := by
+  rw [eq_compactlyGenerated (X := X), TopologicalSpace.compactlyGenerated, isClosed_coinduced,
```
dagurtomas, line 73: This theorem should hold for more general universes and `UCompactlyGeneratedSpace` as well. Could you try to prove it in the proper generality? If it's annoying to apply it (e.g. if Lean can't figure out the universes automatically in the proof of the theorem below), then you can state this special case as well.

Mathlib/Topology/Compactness/CompactlyGeneratedSpace.lean
```diff
@@ -0,0 +1,177 @@
...
+theorem CompactlyGeneratedSpace.isOpen [CompactlyGeneratedSpace X] {s : Set X}
+    (hs : ∀ ⦃K⦄, IsCompact K → IsOpen (s ∩ K)) : IsOpen s := by
+  refine isOpen' fun f hf ↦ ?_
+  rw [← Set.preimage_inter_range]
+  exact (hs (isCompact_range hf)).preimage hf
+
+/-- If the topology of `X` is coinduced by a continuous function whose domain is
+compactly generated, then so is `X`. -/
+theorem compactlyGeneratedSpace_of_coinduced {Y : Type u}
+    [tX : TopologicalSpace X] [tY : TopologicalSpace Y]
```
dagurtomas, line 105: I believe this theorem and the instances below are true for `UCompactlyGeneratedSpace` with any universe. Could you try to generalise?

Mathlib/Topology/Compactness/CompactlyGeneratedSpace.lean
```diff
@@ -0,0 +1,177 @@
...
+universe u v
+
+open TopologicalSpace Filter Topology Set
+
+variable {X : Type u} {Y : Type v} [TopologicalSpace X] [TopologicalSpace Y]
+
+/-- A topological space `X : Type u` is compactly generated if its topology is coinduced by
+continuous maps `f : K → X`, where `K : Type u` is compact Hausdorff. -/
+abbrev CompactlyGeneratedSpace (X : Type u) [TopologicalSpace X] : Prop :=
+  UCompactlyGeneratedSpace.{u} X
```
dagurtomas, line 45: The definition of `UCompactlyGeneratedSpace` could be moved to this file, along with the part of the API surrounding it that is not specific to the category `CompactlyGenerated`. Then the file `Topology.Category.CompactlyGenerated` should import this file.

EtienneC30:
Generalizations to `UCompactlyGeneratedSpace` are quite straightforward. I'm pretty sure they work well for `CompactlyGeneratedSpace`, without having to specialize them. However I think I should still add the specialized versions, because in my view someone who is just interested in topology does not have to know about `UCompactlyGeneratedSpace` nor to care about universes. Is it okay if I do it that way?

dagurtomas:
To answer your question (if I understand it correctly), yes, it's completely fine and useful to have both the general and specialized version of the theorems when that applies. What I mean is that you can prove theorems for `UCompactlyGeneratedSpace.{u, w}`, and then prove the specialized versions for `CompactlyGeneratedSpace` just by applying the general version with `u = w` (don't duplicate the proof).

EtienneC30:
Yes, that's what I intended to do. Thanks!

riccardobrasca:
Thanks!

bors merge

PR #15214, opened 2024-07-27
[Merged by Bors] - feat: add lemma relating Set.encard to tsum over ennreal 

edegeltje:
two lemmas: `s.tsum (1:ENNReal) = s.encard` and `s.tsum c = s.encard * c`, where `c:ENNReal` and `s : Set α`.

---

### DIFF
--- a/Mathlib/Topology/Algebra/InfiniteSum/ENNReal.lean
+++ b/Mathlib/Topology/Algebra/InfiniteSum/ENNReal.lean
@@ -15,7 +15,8 @@ This file provides lemmas relating sums of constants to the cardinality of the d
 ## TODO
 
 + Once we have a topology on `ENat`, provide an `ENat` valued version
-+ Once we replace `PartENat` entirely with `ENat` (and get a `ENat.card` to replace `PartENat.card`), provide versions which sum over the whole type.
++ Once we replace `PartENat` entirely with `ENat` (and replace `PartENat.card` with a `ENat.card`),
+  provide versions which sum over the whole type.
 -/
 
 open Set Function
@@ -32,9 +33,10 @@ lemma tsum_set_one_eq : ∑' (_ : s), (1 : ℝ≥0∞) = s.encard := by
   · have : Infinite s := infinite_coe_iff.mpr hinf
     rw [tsum_const_eq_top_of_ne_zero one_ne_zero, encard_eq_top hinf, ENat.toENNReal_top]
 
-lemma tsum_const_eq' {α : Type*} (s : Set α) (c : ℝ≥0∞) :
+@[simp]
+lemma tsum_set_const_eq' (c : ℝ≥0∞) :
     ∑' (_:s), (c : ℝ≥0∞) = s.encard * c := by
   nth_rw 1 [← one_mul c]
-  rw [ENNReal.tsum_mul_right,tsum_one_eq']
+  rw [ENNReal.tsum_mul_right,tsum_set_one_eq]
 
 end ENNReal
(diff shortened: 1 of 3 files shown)


Mathlib/Topology/Algebra/InfiniteSum/ENNReal.lean
```diff
@@ -0,0 +1,61 @@
...
+  else
+  have : Infinite s := by exact infinite_coe_iff.mpr hfin
+  rw [tsum_const_eq_top_of_ne_zero (by norm_num)]
+  rw [Set.encard_eq_top_iff.mpr hfin]
+  simp only [ENat.toENNReal_top]
+
+lemma tsum_const_eq' {α : Type*} (s : Set α) (c : ℝ≥0∞) :
+    ∑' (_:s), (c : ℝ≥0∞) = s.encard * c := by
+  nth_rw 1 [← one_mul c]
+  rw [ENNReal.tsum_mul_right,tsum_one_eq']
```
j-loreaux, line 59: Similar ideas, but note how we're now using the `simp` lemma from above
```suggestion
@[simp]
lemma tsum_set_const_eq (c : ℝ≥0∞) :
    ∑' (_ : s), (c : ℝ≥0∞) = s.encard * c := by
  rw [← one_mul c, ENNReal.tsum_mul_right]
  simp
```

j-loreaux (commented):
Thanks! There's some simplification we can do here. I tried to explain my thought process. Let me know if you have questions.

Mathlib/Topology/Algebra/InfiniteSum/ENNReal.lean
edegeltje: surely not having a simp makes the proof more robust?

j-loreaux: In this particular case, it's not really a big deal, you can leave it how you had it if you prefer.

However, to answer your question, I would say two things:

1. Given the choice between say `rw [lem1, lem2, lem3, lem4, lem5]` and `simp`, I would greatly prefer the latter for readability and proof comprehension.
2. Given the same choice, I would again prefer `simp` (as a *closing* tactic only) for robustness, because it is not sensitive to changes in lemma names. Name changes are much more common than moving a given lemma into or out of the `simp` set.

In this case, I was opting for `simp` because it conveniently avoided the need to do an `nth_rw` (because even though it rewrites the rhs too, that gets `simp`ed away). But again, the proof you had here was fine.

Mathlib/Topology/Algebra/InfiniteSum/ENNReal.lean
```diff
@@ -0,0 +1,42 @@
...
+lemma tsum_set_one_eq : ∑' (_ : s), (1 : ℝ≥0∞) = s.encard := by
+  obtain (hfin | hinf) := Set.finite_or_infinite s
+  · lift s to Finset α using hfin
+    simp [tsum_fintype]
+  · have : Infinite s := infinite_coe_iff.mpr hinf
+    rw [tsum_const_eq_top_of_ne_zero one_ne_zero, encard_eq_top hinf, ENat.toENNReal_top]
+
+@[simp]
+lemma tsum_set_const_eq' (c : ℝ≥0∞) :
+    ∑' (_:s), (c : ℝ≥0∞) = s.encard * c := by
```
j-loreaux, line 38:
```suggestion
lemma tsum_set_const_eq (c : ℝ≥0∞) : ∑' (_:s), (c : ℝ≥0∞) = s.encard * c := by
```

j-loreaux:
bors d+

PR #15224, opened 2024-07-28
feat(CategoryTheory/Sites): covering families and their associated Grothendieck topology

AnthonyBordg:
Define covering families on a category and their associated Grothendieck topology by using the API for `Coverage`.
Give an explicit characterization of the covering sieves of the said topology. 

---


Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
```diff
@@ -0,0 +1,116 @@
...
+
+## Tags
+
+Grothendieck topology, sieve, generated topology, coverage, covering families
+-/
+
+namespace CategoryTheory
+
+universe u v
+
```
dagurtomas, line 35:
```suggestion
universe v u

```

This is the convention in the category theory library. I don't think you will ever need to explicitly specify the universes, so alternatively you may delete them and change the line below to `variable (C : Type*) [Category C]` (that way, Lean introduces arbitrary universe levels automatically).

dagurtomas (commented):
Could you say a bit more about potential future applications of this API? Possibly as `TODO`s in the module docstring and/or in the PR description

Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
```diff
@@ -0,0 +1,116 @@
...
+variable (C : Type u) [Category.{v} C]
+
+/-! ### The notion of covering families -/
+
+open Sieve
+
+/-- A collection of `CoveringFamilies` on `C` is a `Coverage` on `C` satisfying three additional
+axioms. -/
+structure CoveringFamilies extends Coverage C where
```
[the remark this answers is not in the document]
dagurtomas: I mean that for example, a coverage is also a "collection of covering families", but this is a coverage with more properties, so it feels weird to give it that name

Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
```diff
@@ -0,0 +1,116 @@
...
+namespace CoveringFamilies
+
+/-- The topology `GeneratedTopology` generated by a collection `cf` of covering families is the
+topology generated by its underlying coverage. -/
+def GeneratedTopology : GrothendieckTopology C := cf.toGrothendieck
+
+/-- `covering_of_generatedTopology_iff` gives an explicit characterization of the covering sieves
+of the topology generated by a collection `cf` of covering families, those are exactly the sieves
+that contain the sieve generated by a covering presieve. -/
+theorem covering_of_generatedTopology_iff {X : C} (S : Sieve X) :
```
joelriou, line 65: From this, I am able to show this:
```lean
theorem covering_of_generatedTopology_iff' {X : C} (S : Sieve X) :
    (GeneratedTopology C cf).sieves X S ↔ S.1 ∈ cf.covering X := by
  rw [covering_of_generatedTopology_iff]
  constructor
  · rintro ⟨P, hP₁, hP₂⟩
    refine cf.heredity S.arrows ⟨_, hP₁,
      by simpa only [generate_sieve, Sieve.generate_le_iff] using hP₂⟩
  · rintro h
    exact ⟨_, h, by simp⟩
```

I am afraid the notion you are trying to define is not significantly different from what we already have in mathlib (because when restricted to sieves, the data in `cf` is already a Grothendieck topology, and the axioms are stated in very similar fashions).

AnthonyBordg: @joelriou From the `Coverage` alone and its associated topology J, can you prove that the J-covering sieves are **exactly** the sieves generated by the elements of `covering`?

dagurtomas: No, that's not true in general. What I think Joël means is that your definition is too close to the existing notion of `GrothendieckTopology`

AnthonyBordg:
> Could you say a bit more about potential future applications of this API? Possibly as `TODO`s in the module docstring and/or in the PR description

Regarding applications, according to a duality theorem of Caramello, knowing the covering sieves of a topology is equivalent to provability in first-order geometric theories.

Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
```diff
@@ -0,0 +1,116 @@
...
+
+/-! ### The notion of covering families -/
+
+open Sieve
+
+/-- A collection of `CoveringFamilies` on `C` is a `Coverage` on `C` satisfying three additional
+axioms. -/
+structure CoveringFamilies extends Coverage C where
+  /-- Every presieve generated by a morphism that has a section is a covering presieve. -/
+  retraction_mem : ∀ ⦃X Y : C⦄ ⦃f : Y ⟶ X⦄, IsSplitEpi f → Presieve.singleton f ∈ covering X
```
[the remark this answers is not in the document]
AnthonyBordg: The axiom that every morphism that has a section is covering is indeed a consequence of the other axioms plus the simple axiom that identity morphisms are covering. However, the _raison d'être_  of this axiom is that it corresponds to the description of covering morphisms wrt the discrete topology. In the example of the discrete topology, one sees that the notion of covering sieve can be trivial while the notion of covering morphism (or family of covering morphisms) can be highly non-trivial. For instance, take the category of affine schemes on $\mathbb{Q}$ equipped with the discrete topology, and consider the projection morphism on Spec($\mathbb{Q}$) of Spec(A), where A is the commutative algebra on  $\mathbb{Q}$ generated by the variables $X,X’,Y,Y’$ satisfying the equations $X^n + Y^n = 1$, $X.X’=1$, $Y.Y’=1$. Knowing if the projection is covering wrt the discrete topology is Fermat's problem ...

Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
```diff
@@ -0,0 +1,116 @@
...
+
+## Main Results
+
+⋆ `CategoryTheory.CoveringFamilies.covering_of_generatedTopology_iff` : we give an explicit
+  characterization of the covering sieves of the topology generated by covering families.
+
+## References
+
+⋆ Le problème de l'engendrement des topologies de Grothendieck, slides (not publicly available yet)
+  of a talk given by Laurent Lafforgue at Centre Lagrange on 26th April 2024.
```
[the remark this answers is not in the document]
AnthonyBordg: The only reference I'm aware of.

Mathlib/CategoryTheory/Sites/CoveringFamilies.lean
AnthonyBordg: Indeed, that was my point. In general, the generated topology from a coverage has a very different form. This notion is not operative for working on the generated topology, except for the notions which are linked to it by a Galois correspondence. For example, for a presheaf to be a sheaf for the generated topology, it is sufficient that it be so for the families of a "coverage" which generates this topology. 
The notion of Grothendieck topology is formulated in an equivalent way in terms of sieves or in terms of covering families. The passage from one to the other is immediate in theory for a mathematician (but a little less immediate in a formalized system), so Joël is right, and yet these are two different points of view: one categorical and the other geometric.
Depending on what one seeks to do and depending on the topologies considered, it is more convenient to think and express oneself in one or the other point of view.
Whether one works with the usual topologies of algebraic geometry or with the syntactic topologies of syntactic categories, their natural expression is generally in terms of covering families.

AnthonyBordg: @joelriou Do you think this piece of formalisation would be more suitable as a section of `Mathlib.CategoryTheory.Sites.Coverage`? It would be completely fine with me.

joelriou: I am not convinced that this fits into mathlib. "Covering families" (whatever that means) are already part of mathlib in the sense that we have the notion of a sieve generated by a family of morphisms (this is `Sieve.ofArrows`), and I have recently added some API in order to phrase the sheaf condition in terms of 1-hypercovers (which at the 0th level correspond to the data of a family of morphisms with fixed target such that the generated sieve is covering, see `Sites.OneHypercover`).

Note also that in terms of formalization, there is a difference between (pre)sieves and families of morphisms, because in the latter, there can be duplicate maps, while no duplicated are allowed with presieves. For example, `X` is covered by itself, but it is also covered by the family of two identical identity maps. I would very much welcome more development of the API for families of maps `∀ i, Y i ⟶ X` instead of presieves only...

In my opinion, the only way this could find its way into mathlib is if you were to introduce a new constructor for Grothendieck topologies (or `Coverage`) which would not take a family of presieves as an input but a predicate on families of morphisms `∀ (i : I), Y i ⟶ X`, with `I` in a certain universe `w`, similarly as I have done for `1`-hypercovers. (The structure could be named `CoveringArrowFamilies`?)

In other words, I suggest you use `Sieve.ofArrows` rather than `Sieve.generate`.

adamtopaz:
> > Could you say a bit more about potential future applications of this API? Possibly as `TODO`s in the module docstring and/or in the PR description
> 
> Regarding applications, according to a duality theorem of Caramello, knowing the covering families of a topology is equivalent to provability in first-order geometric theories.

As Joel suggests above, this can almost surely already be formulated with mathlib's existing `GrothendieckTopology`'s and/or `Coverage`s.

joneugster:

PR #15210, opened 2024-07-27
[Merged by Bors] - chore: move Init.Order.Defs to Order.Defs

Ruben-VandeVelde:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15228, opened 2024-07-28
[Merged by Bors] - chore(Init/Order): Rework files

YaelDillies:
Merge `Init.Order.Lemmas` and the part of `Init.Algebra.Classes` about relations into `Order.Defs`.

---

- [x] depends on: #16022
- [x] depends on: #16059


Mathlib/Init/Order/Defs.lean
```diff
@@ -1,411 +0,0 @@
...
-  Preorder.le_trans _ _ _
-
-theorem lt_iff_le_not_le : ∀ {a b : α}, a < b ↔ a ≤ b ∧ ¬b ≤ a :=
-  Preorder.lt_iff_le_not_le _ _
-
-theorem lt_of_le_not_le : ∀ {a b : α}, a ≤ b → ¬b ≤ a → a < b
-  | _a, _b, hab, hba => lt_iff_le_not_le.mpr ⟨hab, hba⟩
-
-theorem le_not_le_of_lt : ∀ {a b : α}, a < b → a ≤ b ∧ ¬b ≤ a
-  | _a, _b, hab => lt_iff_le_not_le.mp hab
```
Ruben-VandeVelde, line 65: If you want to remove random theorems, please do that in a separate PR, after they've been deprecated for a while

YaelDillies: This one is a useless combination of other lemmas and uses are anecdotic. Am happy to see it restored deprecated, though

Ruben-VandeVelde (changes requested):
We should not move code and reformat it at the same time; this makes review impossible. (I also don't think there's consensus we should go around reformatting code anyway.)

YaelDillies:
Well at least @digama0 wrote in the header of this file that a rework was needed 🤷

grunweg:
You have a merge conflict; can you merge master, please? Thanks!

YaelDillies:
@Vierkantor offered to do it, but looks like they're on holiday. Will do it tomorrow evening if they don't beat me to it

Mathlib/Data/DFinsupp/Lex.lean
```diff
@@ -60,10 +60,9 @@ theorem lex_lt_of_lt [∀ i, PartialOrder (α i)] (r) [IsStrictOrder ι r] {x y
 variable [LinearOrder ι]
 
 instance Lex.isStrictOrder [∀ i, PartialOrder (α i)] :
-    IsStrictOrder (Lex (Π₀ i, α i)) (· < ·) :=
-  let i : IsStrictOrder (Lex (∀ i, α i)) (· < ·) := Pi.Lex.isStrictOrder
-  { irrefl := toLex.surjective.forall.2 fun _ ↦ @irrefl _ _ i.toIsIrrefl _
-    trans := toLex.surjective.forall₃.2 fun _ _ _ ↦ @trans _ _ i.toIsTrans _ _ _ }
+    IsStrictOrder (Lex (Π₀ i, α i)) (· < ·) where
+  irrefl _ := lt_irrefl (α := Lex (∀ i, α i)) _
+  trans _ _ _ := lt_trans (α := Lex (∀ i, α i))
```
j-loreaux, line 65: why is code reformatting part of this PR? This is just noise in the diff.

Mathlib/Init/Algebra/Classes.lean
```diff
@@ -142,56 +147,12 @@ class is_idempotent (α : Sort u) (f : α → α) : Prop :=
...
-class IsTotal (α : Sort u) (r : α → α → Prop) : Prop where
-  total : ∀ a b, r a b ∨ r b a
-
-/-- `IsPreorder X r` means that the binary relation `r` on `X` is a pre-order, that is, reflexive
-and transitive. -/
-class IsPreorder (α : Sort u) (r : α → α → Prop) extends IsRefl α r, IsTrans α r : Prop
-
 /-- `IsTotalPreorder X r` means that the binary relation `r` on `X` is total and a preorder. -/
+@[deprecated (since := "2024-07-30")]
 class IsTotalPreorder (α : Sort u) (r : α → α → Prop) extends IsTrans α r, IsTotal α r : Prop
```
j-loreaux, line 156: Lots of things have been deprecated in this PR, with no explanation in the description as to the reasons. Even if they are the right choices, it's hard to understand why in all cases.

Mathlib/Order/Defs.lean
```diff
@@ -16,8 +15,176 @@ Defines classes for preorders, partial orders, and linear orders
 and proves some basic lemmas about them.
 -/
 
+/-! ### Unbundled classes -/
+
 universe u
-variable {α : Type u}
+variable {α : Type*}
```
j-loreaux, line 21: Why the change from `Type u` to `Type*`? Seems just aesthetic.

Mathlib/Order/Defs.lean
```diff
@@ -26,90 +193,59 @@ section Preorder
...
 /-- A preorder is a reflexive, transitive relation `≤` with `a < b` defined in the obvious way. -/
-class Preorder (α : Type u) extends LE α, LT α where
+class Preorder (α : Type*) extends LE α, LT α where
   le_refl : ∀ a : α, a ≤ a
   le_trans : ∀ a b c : α, a ≤ b → b ≤ c → a ≤ c
   lt := fun a b => a ≤ b ∧ ¬b ≤ a
   lt_iff_le_not_le : ∀ a b : α, a < b ↔ a ≤ b ∧ ¬b ≤ a := by intros; rfl
 
-variable [Preorder α]
+variable [Preorder α] {a b c : α}
```
j-loreaux, line 202: I understand the desire to simplify code by pulling out variables, but this just increases the diff and makes review harder.

Mathlib/Order/Defs.lean
```diff
@@ -26,90 +193,59 @@ section Preorder
...
 
 /-- The relation `≤` on a preorder is reflexive. -/
-@[refl]
-theorem le_refl : ∀ a : α, a ≤ a :=
-  Preorder.le_refl
+@[refl] lemma le_refl : ∀ a : α, a ≤ a := Preorder.le_refl
```
j-loreaux, line 205: Again, this is just an aesthetic change and makes it harder to understand what's happening.

j-loreaux (changes requested):
Yaël, I'm appreciative of all the work you do cleaning up these low-level files, but this diff is very difficult to review as-is. There's a ton of noise. You're doing a split / move of multiple files into one, golfing proofs, changing aesthetic-only choices, making stylistic changes (where the previous style wasn't actually wrong according the style guide), adding `deprecation` tags, and maybe more, I'm not sure (because it's too hard to figure out).

Can you please split this into more manageable chunks?

Mathlib/Data/DFinsupp/Lex.lean
YaelDillies: It wasn't showing in the diff before #15210 was merged...

Mathlib/Init/Algebra/Classes.lean
YaelDillies: I am deprecating all the classes that mathlib is not using. This is a long standing TODO in those Init files: Deciding what's useful and what's not. If you scroll up a bit, you will see that Mario has written a fairly detailed list of all those classes, with hints to which ones are useful to keep and which ones are not

Mathlib/Order/Defs.lean
YaelDillies: Again, wasn't showing in the diff before #15210

Mathlib/Data/DFinsupp/Lex.lean
YaelDillies: Actually wait no that's not code reformatting at all. That's necessary because of the argument changes to `irrefl` and `trans`

jcommelin:
Thanks :tada:

bors merge

mattrobball:
@jcommelin I am surprised bors goes forward with changes requested.

grunweg:
That looks surprising indeed. Perhaps because the commented lines are now "outdated" (for github's, but perhaps also bors' purposes)?

PR #15230, opened 2024-07-28
[Merged by Bors] - chore(SobolevInequality): drop a `DecidableEq` assumption

urkud:
---

### DIFF
--- a/Mathlib/Analysis/FunctionalSpaces/SobolevInequality.lean
+++ b/Mathlib/Analysis/FunctionalSpaces/SobolevInequality.lean
@@ -54,7 +54,7 @@ open Set Function Finset MeasureTheory Measure Filter
 
 noncomputable section
 
-variable {ι : Type*} [Fintype ι] [DecidableEq ι]
+variable {ι : Type*} [Fintype ι]
 
 local prefix:max "#" => Fintype.card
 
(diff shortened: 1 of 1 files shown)


Mathlib/Analysis/FunctionalSpaces/SobolevInequality.lean
```diff
@@ -294,6 +294,7 @@ theorem lintegral_pow_le_pow_lintegral_fderiv_aux
     {u : (ι → ℝ) → F} (hu : ContDiff ℝ 1 u)
     (h2u : HasCompactSupport u) :
     ∫⁻ x, (‖u x‖₊ : ℝ≥0∞) ^ p ≤ (∫⁻ x, ‖fderiv ℝ u x‖₊) ^ p := by
+  letI := Classical.decEq ι
```
grunweg, line 297: Would you like to also drop the `DecidableEq ι` assumption on line 57?

grunweg (commented):
One quick question - otherwise, looks good to me. Thanks!

Mathlib/Analysis/FunctionalSpaces/SobolevInequality.lean
urkud: The definition of `T` and some others need decidable equality.

j-loreaux: I guess this is taking advantage of the current behavior of `variable` which drops unnecessary variables after processing the declaration. Can you just shadow the variables instead?

j-loreaux:
bors merge

PR #11746, opened 2024-03-28
[Merged by Bors] - feat: proper group actions

vin-gui:
We define proper action of a group on a topological space, and we prove that in this case the quotient space is T2. We prove that if a Hausdorff group acts properly on $X$, then $X$ is Hausdorff. We also give equivalent definitions of proper action using ultrafilters and show the transfer of proper action to a closed subgroup.
We show that if $X \times X$ is Hausdorff and compactly generated then the action is properly discontinuous if and only if it is continuous in the second variable and proper in the sense defined here. The "compactly generated" assumption is to be rephrased with typeclass inference in the future.

Co-authored-by: @EtienneC30
Co-authored-by: @florestan 
Co-authored-by: @ADedecker 

---

- [x] depends on: #11754
- [x] depends on: #13635

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15239, opened 2024-07-28
[Merged by Bors] - feat: compactly generated spaces for proper maps and preimages of compact sets / proper and properly discontinuous actions

EtienneC30:
Prove that a continuous function whose codomain is T2 and compactly generated is proper if and only if the preimages of compact sets are compact.
This result already existed but was stated without the typeclass CompactlyGeneratedSpace. It is moved to its own file.

When `X \times X` is T2 and compactly generated, a group action is proper if and only if it is properly discontinuous. This was stated without a type class instance for the compactly generated hypothesis. Now it is.

---

- [x] depends on: #15201
- [x] depends on: #15203


loefflerd:
It's a pity that this PR has been left to idle for so long; it looks like a nice contribution.

If I understand correctly, the goal of the PR is to get rid of a number of lemmas which have the words "compactly generated" in their names, but which do not use the CompactlyGeneratedSpace typeclass, instead spelling out the conclusion manually as `IsClosed s ↔ ∀ ⦃K⦄, IsCompact K → IsClosed (s ∩ K)`. I checked and the typeclass machinery is indeed capable of inferring `CompactlyGeneratedSpace X` under the hypotheses of each of these lemmas.

However, I wonder if we should use the deprecation machinery, just in case anyone has written some code (maybe in downstream projects depending on mathlib) which uses these lemmas. Could you perhaps keep them in situ for now, but with a deprecation warning saying "this lemma is deprecated, use the typeclass machinery instead".

(Perhaps it would also be good to have a lemma which precisely reproduces the "old" formulation, `∀ s : Set X, IsClosed s ↔ ∀ ⦃K⦄, IsCompact K → IsClosed (s ∩ K)`, assuming `CompactlyGeneratedSpace X`. I couldn't see a lemma which does exactly that, although it looks like it could be cobbled together pretty easily from other existing lemmas. If we add that, then we can point to it in the deprecation message.)

EtienneC30:
In fact what happened is that the lemmas mentioning `compactlyGenerated` were introduced  by #13635, which was opened before `CompactlyGeneratedSpace` was defined. Then when #13635 and #11746 got merged `CompactlyGeneratedSpace` had been defined so I opened this to update everything, but I had not anticipated that it would be sleeping for so long 😅. A small detail is that the definition of `CompactlyGeneratedSpace` is not the same as the one made in the old lemmas, therefore `∀ s : Set X, IsClosed s ↔ ∀ ⦃K⦄, IsCompact K → IsClosed (s ∩ K)` only holds in a T2 space (both directions are proved in `Topology.Compactness.CompactlyGeneratedSpace`), which is assumed in these lemmas.

Anyway I agree with your suggestions, but I never used deprecation before so I'm not sure what to do. In particular here it's not just a change in the name but a change in the arguments, so things will break anyway. Also lemmas like `compactlyGenerated_of_weaklyLocallyCompactSpace` should just be completely removed as they are now dealt with by typeclass inference. Can/Should I make a deprecation without an alias?

EtienneC30:
Thanks for looking at the PR btw!

EtienneC30:
More questions:

1. It looks like I can't put a deprecation date AND a deprecation message at the same time, so I think I'll just put the message
2. In this PR I moved lemmas related to compactly generated spaces and proper maps in a new file because of the `assert`. So if I want to deprecate `compactlyGenerated_of_weaklyLocallyCompactSpace` I can't point to a lemma in `Topology.Compactness.CompactlyGeneratedSpace` because it is not imported. Should I move the deprecated lemmas from `Topology.Maps.Proper.Basic` to `Topology.Maps.Proper.CompactlyGenerated`?

loefflerd:
> More questions:
> 
> 1. It looks like I can't put a deprecation date AND a deprecation message at the same time, so I think I'll just put the message

Looks like the order of the arguments matters: `@[deprecated "lorem ipsum dolor" (since := "2024-10-10")]` works.

> 2. In this PR I moved lemmas related to compactly generated spaces and proper maps in a new file because of the `assert`. So if I want to deprecate `compactlyGenerated_of_weaklyLocallyCompactSpace` I can't point to a lemma in `Topology.Compactness.CompactlyGeneratedSpace` because it is not imported. Should I move the deprecated lemmas from `Topology.Maps.Proper.Basic` to `Topology.Maps.Proper.CompactlyGenerated`?

I wouldn't advise moving lemmas just to deprecate them. If you give an explicit deprecation-message string, then you can refer to lemmas in files that aren't imported:
```
@[deprecated "Do not use this lemma; use `CompactlyGeneratedSpace.foo` from 
`Topology.Maps.Proper.CompactlyGenerated`" (since := "2024-10-10")]
lemma bar
```

Mathlib/Topology/Compactness/CompactlyGeneratedSpace.lean
```diff
@@ -322,15 +329,15 @@ open scoped Set.Notation in
...
 Then `X` is compactly generated. -/
-theorem compactlyGeneratedSpace_of_isOpen_of_t2 [T2Space X]
+theorem compactlyGeneratedSpace_of_isOpen_of_t2
     (h : ∀ s, (∀ (K : Set X), IsCompact K → IsOpen (K ↓∩ s)) → IsOpen s) :
     CompactlyGeneratedSpace X := by
   refine compactlyGeneratedSpace_of_isOpen fun s hs ↦ h s fun K hK ↦ ?_
   have : CompactSpace ↑K := isCompact_iff_compactSpace.1 hK
   exact hs _ Subtype.val continuous_subtype_val
 
 /-- A Hausdorff and weakly locally compact space and compactly generated. -/
```
loefflerd, line 339: Please fix typo (I know this isn't your fault of course)

loefflerd:
I think the available tooling doesn't provide a good solution for this situation (when we change the formulation of a lemma without changing its name). We just expect downstream projects (or in-progress PR's) to make any necessary adjustments themselves. Realistically it seems very unlikely any downstream projects are using these lemmas anyway, so I wouldn't worry about it too much.

EtienneC30:
I agree they weren't probably used. I'm adding a prime at the end of the name.

EtienneC30:
Another problem: deprecated lemmas use deprecated lemmas so CI can't succeed.

loefflerd:
Hi Etienne, 

I have the impression that my suggestion to use deprecations is causing far more work than it's worth. It's always been my impression with Mathlib that the first priority is to keep on making the library better; maintaining a stable interface for downstream projects is a far, far lower priority. We didn't even have a deprecation system at all until a few months ago.

I suggest you decide first how you want the code to look going forward. Then ask if you can fit in deprecation warnings for the removed lemmas around that. If it's possible and easy, do it. But if it's not, don't worry about it; and definitely don't let deprecation considerations affect your choices about names / locations of new lemmas.

EtienneC30:
In the end I did not put any deprecation. As I kept the names of the old lemmas, I can't deprecate them, so I added a note in the docstring of the new lemmas. That way if anyone encounters an issue they can check the lemma in the docs and see it's still there but was moved, and the note explains why.

loefflerd:
Thanks for your work on this, and for your patience with my (ultimately misguided) suggestions! I am now recommending this PR for merging.

maintainer merge

EtienneC30:
Thank you for looking at the PR!

jcommelin (commented):
Thanks :tada:

bors merge

PR #15222, opened 2024-07-28
[Merged by Bors] - chore: run fix-comments.py on all of mathlib

grunweg:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15242, opened 2024-07-28
[Merged by Bors] - chore: remove fix-comments.py

grunweg:
This script has served its purpose during the port; now that all align information has been removed from mathlib, it is not useful for other projects either.

Projects which wish to use such functionality should check out the v3-eol tag and use the script from that commit.

---

Helps with #14539. PR #15222 (in progress) runs that script on all of mathlib, but is based on the above tag - removing it on `master` does not interfere with this.


j-loreaux:
I agree, now that the `#align`s are actually removed, this has no use.

maintainer merge

bryangingechen:
Thanks!
bors r+

PR #15245, opened 2024-07-29
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 1)

kim-em:
The new variables inclusion mechanism https://github.com/leanprover/lean4/pull/4814 is going to require many changes to Mathlib.

Some of these changes are to make sure typeclass hypotheses are not unnecessarily included. This is achieved either by using `section` to bound `variable [...]`, using `variable ... in`, or reordering theorems. (I think you can see all three being using here.)

Other changes require adding `include ...` to include variables that the new mechanism won't automatically include. These changes can't be backported to master, and instead are accumulating on the `lean-pr-testing-4814` branch.


Mathlib/Logic/Basic.lean
```diff
@@ -951,6 +951,9 @@ either branch to `a`. -/
 theorem ite_apply (f g : ∀ a, σ a) (a : α) : (ite P f g) a = ite P (f a) (g a) :=
   dite_apply P (fun _ ↦ f) (fun _ ↦ g) a
 
+section
+variable [Decidable Q]
```
sgouezel, line 955: This one looks wrong, you're just adding an assumption.

semorrison: Previously we were using an explicit `Decidable P`, and the classical decidable instance for `Q`. Surely this is not what was intended?

sgouezel: Thanks for the explanation, I had not noticed the `attribute [local instance 10] Classical.propDecidable` at the beginning of the file (which we should probably remove, but that's for another PR).

sgouezel:
bors d+

semorrison:
bors merge

PR #15248, opened 2024-07-29
[Merged by Bors] - chore: remove unnecessary implicit arguments in MonoidWithZeroHom.coe_copy

kim-em:
It's perfectly fine for these to just be instance implicit arguments.


j-loreaux:
yep, artifact of porting.

bors merge

PR #4814, opened 2023-06-07
[Merged by Bors] - feat: port MeasureTheory.Integral.DivergenceTheorem

urkud:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #2452, opened 2023-02-22
[Merged by Bors] - feat: Port CategoryTheory.EssentiallySmall

adamtopaz:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15255, opened 2024-07-29
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 4)

kim-em:
See https://github.com/leanprover-community/mathlib4/pull/15245


Mathlib/Order/Bounds/Basic.lean
```diff
@@ -957,91 +957,100 @@ end LinearOrder
...
   ⟨mem_image_of_mem _ Ha.1, Hf.mem_upperBounds_image_self Ha.2 Ha.1⟩
 
 end MonotoneOn
 
 namespace AntitoneOn
 
-variable [Preorder α] [Preorder β] {f : α → β} {s t : Set α} (Hf : AntitoneOn f t) {a : α}
-  (Hst : s ⊆ t)
+variable [Preorder α] [Preorder β] {f : α → β} {s t : Set α} {a : α}
+  -- (Hf : AntitoneOn f t) (Hst : s ⊆ t)
```
sgouezel, line 1012:
```suggestion
```

Mathlib/Order/Bounds/Basic.lean
```diff
@@ -1056,7 +1065,7 @@ theorem mem_upperBounds_image (Ha : a ∈ upperBounds s) : f a ∈ upperBounds (
 theorem mem_lowerBounds_image (Ha : a ∈ lowerBounds s) : f a ∈ lowerBounds (f '' s) :=
   forall_mem_image.2 fun _ H => Hf (Ha H)
 
-theorem image_upperBounds_subset_upperBounds_image (Hf : Monotone f) :
+theorem image_upperBounds_subset_upperBounds_image :
```
sgouezel, line 1068: I'm surprised that you're deleting it here, instead of including it in all the theorems of the section. Is that because you're planning to `include Hf` or something like that? Note that there was a clear majority to say in #2452 that the global (per section) `include / omit` mechanism of Lean 3 was very bad for readability. I don't know if Sebastian is planning to put it in #4814, but if we can avoid it it would probably be better.

sgouezel (commented):
bors d+

Mathlib/Order/Bounds/Basic.lean
semorrison: Yes, I had in mind to use `include`. When I've been working on lean-pr-testing-4814, I was inclined to use `include` once to got to beyond about 5 instances of `include in` or copy-pasting the variable.

But if we really want to avoid section level `include ...`, I am happy to do that!

In fact, it makes life somewhat easier, because we can backport more of the changes to `master` now.

semorrison:
bors merge

PR #15260, opened 2024-07-29
[Merged by Bors] - chore(Idempotents): use `Pairwise`

urkud:
Also generalize a few lemmas from `[Fintype I]` to `{s : Finset I}`.

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #13912, opened 2024-06-17
[Merged by Bors] - refactor(Topology/Category): refactor LightProfinite.Basic

dagurtomas:
This is the second part of the refactor of CompHaus and friends (refactoring the definition of `LightProfinite` in terms of `CompHausLike`, see #12930 for the end result)

---
- [x] depends on: #13904 
- [x] depends on: #13905
- [x] depends on: #13907 
- [x] depends on: #13908 
- [x] depends on: #13909 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15264, opened 2024-07-29
[Merged by Bors] - fix(CategoryTheory): move a general lemma about created limits to the correct file

dagurtomas:
This lemma was accidentally in a file about light profinite sets, but is a completely general category-theoretic statement.

---

- [x] depends on: #13912


joelriou:
Thanks!

bors merge

PR #15269, opened 2024-07-29
feat: Add ENNReal.floor

kkytola:
A right-continuous floor function on `ENNReal`.

---

- [x] depends on: #13938
- [ ] depends on: #15773 (general type class for `ENat`-valued floor functions)
- [x] depends on: #15380 (topology on `ENat` is needed to state right continuity of the floor function)


Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,177 @@
...
+
+section ENNReal_floor
+
+open Filter BigOperators TopologicalSpace Topology Set ENNReal NNReal
+
+namespace ENNReal
+
+/-- The floor function `ℝ≥0∞ → ℝ≥0∞`: the floor of `x` is the supremum of the (coerced) natural
+numbers `n` satisfying `n ≤ x`. -/
+noncomputable def floor (x : ℝ≥0∞) : ℝ≥0∞ :=
```
urkud, line 20: Should it take values in `ENat`?

kkytola: @YaelDillies made the same remark when I mentioned this. My eventual use case would coerce back to `ENNReal` anyway, but I think you have a good point and the floor function might be more naturally be defined as `ENat` valued.

I modified the PR accordingly, except I only made the necessary `ENat` topology instance very quickly in this file locally and `sorry`ed out the proof that it is the order topology. If this approach looks otherwise acceptable, I could see if I can fill in the necessary parts about ENat topology.

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+
+end ENat_topology
+
+section ENNReal_floor
+
+namespace ENNReal
+
+/-- The floor function `ℝ≥0∞ → ℕ∞`: the floor of `x` is the supremum of the extended natural
+numbers `n` satisfying `n ≤ x`. -/
+noncomputable def floor (x : ℝ≥0∞) : ℕ∞ := sSup {n : ℕ∞ | n ≤ x}
```
YaelDillies, line 42: I would tend to think this should be called `ENat.floor` to match `Nat.floor` and `Int.floor`. Other types it could apply to are `ENNRat` (doesn't exist yet), `EReal`, `ERat`

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+  congr
+  ext m
+  simp only [mem_image, mem_setOf_eq, mem_inter_iff, mem_range, mem_Iic]
+  refine ⟨fun hx ↦ ?_, fun hx ↦ ?_⟩
+  · obtain ⟨n, hn⟩ := hx
+    refine ⟨⟨n, hn.2⟩, hn.2 ▸ hn.1⟩
+  · obtain ⟨n, hn⟩ := hx.1
+    refine ⟨n, ⟨hn ▸ hx.2, hn⟩⟩
+
+lemma floor_mono (h : x ≤ y) : x.floor ≤ y.floor := sSup_le_sSup <| fun _ hx ↦ hx.trans h
```
YaelDillies, line 59:
```suggestion
@[gcongr] lemma floor_mono (h : x ≤ y) : x.floor ≤ y.floor := sSup_le_sSup <| fun _ hx ↦ hx.trans h
```

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+  le_antisymm (by exact_mod_cast floor_le n) (le_sSup (by simp))
+
+@[simp] lemma floor_zero_eq : floor 0 = 0 := by exact_mod_cast ENNReal.floor_coe_eq 0
+
+@[simp] lemma floor_one_eq_one : floor 1 = 1 := by exact_mod_cast ENNReal.floor_coe_eq 1
+
+@[simp] lemma floor_top_eq : floor ∞ = ⊤ := floor_coe_eq ⊤
+
+lemma floor_eq_natFloor_toNNReal (x_ne_top : x ≠ ∞) :
+    floor x = Nat.floor x.toNNReal := by
```
YaelDillies, line 91: Maybe worth turning around and simping

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+    have obs := lt_of_le_of_lt hn h
+    cases' n with k
+    · exfalso; apply not_top_lt obs
+    · exact Nat.cast_le.mpr <| Nat.le_of_lt_succ (by exact_mod_cast obs)
+
+@[simp] lemma floor_coe_eq (n : ℕ∞) :
+    floor n = n :=
+  le_antisymm (by exact_mod_cast floor_le n) (le_sSup (by simp))
+
+@[simp] lemma floor_zero_eq : floor 0 = 0 := by exact_mod_cast ENNReal.floor_coe_eq 0
```
YaelDillies, line 84:
```suggestion
@[simp] lemma floor_zero : floor 0 = 0 := by exact_mod_cast ENNReal.floor_coe_eq 0
```

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+
+@[simp] lemma floor_one_eq_one : floor 1 = 1 := by exact_mod_cast ENNReal.floor_coe_eq 1
```
YaelDillies, line 86:
```suggestion
@[simp] lemma floor_one : floor 1 = 1 := by exact_mod_cast ENNReal.floor_coe_eq 1
```

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+
+@[simp] lemma floor_top_eq : floor ∞ = ⊤ := floor_coe_eq ⊤
```
YaelDillies, line 88:
```suggestion
@[simp] lemma floor_top : floor ∞ = ⊤ := floor_coe_eq ⊤
```

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+  cases' m with l
+  · exact le_top
+  · apply sSup_le
+    intro n hn
+    have obs := lt_of_le_of_lt hn h
+    cases' n with k
+    · exfalso; apply not_top_lt obs
+    · exact Nat.cast_le.mpr <| Nat.le_of_lt_succ (by exact_mod_cast obs)
+
+@[simp] lemma floor_coe_eq (n : ℕ∞) :
```
YaelDillies, line 80:
```suggestion
@[simp] lemma floor_coe (n : ℕ∞) :
```

Mathlib/Data/ENNReal/Floor.lean
```diff
@@ -0,0 +1,162 @@
...
+    exact_mod_cast Nat.floor_le (show 0 ≤ x.toNNReal by simp)
+
+lemma floor_floor (x : ℝ≥0∞) : (x.floor : ℝ≥0∞).floor = x.floor := by simp
+
+lemma floor_lt_top {x : ℝ≥0∞} (x_ne_top : x ≠ ∞) :
+    x.floor < ∞ := by
+  simpa [floor_eq_natFloor_toNNReal x_ne_top] using (natCast_ne_top ⌊x.toNNReal⌋₊).symm.lt_top'
+
+@[simp] lemma floor_add_one {x : ℝ≥0∞} :
+    (x + 1).floor = x.floor + 1 := by
```
YaelDillies, line 114: Can you try following the `Nat.floor` API? This involves proving `floor_add_nat` (maybe renamed to `floor_add_natCast` before `floor_add_one`

Mathlib/Data/ENNReal/Floor.lean
kkytola: Yes, that and mimicking the API of `Nat.floor` makes sense now if this is made `ENat`-valued. (The topology on `ENat` is my main concern with this currently. And to a smaller extent the convenience of the required casts back to `ENNReal`, since there does not seem to be a lot of API for `ENat`, and `ENNReal` by itself can already be less-than-pleasant to work with at times.)

I originally thought the `ENNReal.floor` would be a one-off and therefore different from the more standard floor designs, but maybe someone can some day have use for a floor function on the `ENNRat`, too.

I will try to make this more parallel to the existing one. Thanks for the comment and the improvements to code!

Mathlib/Data/ENNReal/Floor.lean
kkytola: This I am slightly unsure of, but I could test. My doubt is because `ENNReal`s can be complicated to work with, and to prove things about them, one still very often breaks to cases and handles the finite case by casting to `NNReal`. That is more or less how this lemma is used already below.

Mathlib/Data/ENNReal/Floor.lean
YaelDillies: The good thing about making your function be `ENat`-valued is that it suddenly forms a Galois coinsertion: `GaloisCoinsertion Int.cast FloorRing.floor` should hold.

Mathlib/Algebra/Order/EFloor.lean
```diff
@@ -0,0 +1,255 @@
...
+We define the extended natural number -valued floor functions.
+
+## Main Definitions
+
+* `EFloorSemiring`: An ordered semiring with natural-valued floor.
+* `Nat.floor a`: Greatest natural `n` such that `n ≤ a`. Equal to `0` if `a < 0`.
+
+## Notations
+
+* `⌊a⌋ᵢ` is `ENat.floor a`.
```
YaelDillies, line 34:
```suggestion
* `⌊a⌋ₑ` is `ENat.floor a`.
```
for "extended" might be clearer?

Mathlib/Algebra/Order/EFloor.lean
```diff
@@ -0,0 +1,255 @@
...
+  floor_lt _ := by rfl
+
+class CastNatENatClass (α : Type*) [Semiring α] [Preorder α] [CoeTC ℕ∞ α] where
+  cast_nat : ∀ (n : ℕ), (n : ℕ∞) = (n : α)
+  cast_enat_lt_iff : ∀ (n m : ℕ∞), (n : α) < (m : α) ↔ n < m
+  cast_enat_le_iff : ∀ (n m : ℕ∞), (n : α) ≤ (m : α) ↔ n ≤ m
+  cast_nonneg {n : ℕ∞} : (0 : α) ≤ n
+
+@[simp, norm_cast]
+lemma ENat.cast_nat [Semiring α] [Preorder α] [CoeTC ℕ∞ α] [CastNatENatClass α] {n : ℕ} :
```
YaelDillies, line 84:
```suggestion
lemma ENat.cast_natCast [Semiring α] [Preorder α] [CoeTC ℕ∞ α] [CastNatENatClass α] {n : ℕ} :
```

YaelDillies:
I should mention that maybe it's a bit premature to generalise everything away from `NNReal`. Certainly for now merely following the `Nat.floor` API (but specialising `α := ENNReal` everywhere) is plenty fine.

joneugster:

Issue #5878, opened 2023-07-13
Normed rings with summable geometric series


urkud:
Currently, we have two mostly independent sets of lemmas about
`∑' n, x ^ n`, `‖x‖ < 1`: lemmas assuming that `x` belongs to a normed
field (division ring?) and lemmas assuming that `x` belongs to a
complete normed ring. I suggest that we introduce a typeclass

``` lean
class GeometricSumRing (E : Type _) [NormedRing E] : Prop where
  summable_pow : ∀ x : E, ‖x‖ < 1 → Summable (fun n : ℕ → x ^ n)
```

then provide instances for a field and for a complete normed ring.

urkud:
We have [HasSummableGeomSeries](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecificLimits/Normed.html#HasSummableGeomSeries) now, so I'm closing this issue.

PR #15270, opened 2024-07-29
[Merged by Bors] - feat: remove completeness assumption in the formula for `∑ n, n * r ^ n`

sgouezel:
Currently, the formula `∑ n, n * r ^ n = r / (1 - r) ^ 2` in a field assumes that the field is complete to get summability of the series from the summability of the norms. However, as `∑ r ^ n` is converging (to `1 / (1 - r)`), the summability of the Cauchy product is automatic thanks to a variation around results we already have, so completeness can be removed. 

In the same way, `∑ n ^ k * r ^ n` is summable without completeness assumptions. The proof is a little bit indirect (first get summability of `∑ n, (n + k).choose k * r ^ n` for each `k`, as this is `(∑ r ^ n) ^ k`, and then deduce the other one), but it gives the nice formula `∑ n, (n + k).choose k * r ^ n = 1 / (1 - r) ^ k` along the way, which is nice to have.

This is a preliminary PR before unifying several results between `[NormedRing R] [CompleteSpace R]` and `[NontriviallyNormedField R]`, which are proved twice with very similar arguments at several places in mathlib.

---


urkud:
Thank you for working on this! It was on my low-priority TODO list for a long time, see #5878

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -128,6 +128,24 @@ theorem sum_Icc_choose (n k : ℕ) : ∑ m ∈ Icc k n, m.choose k = (n + 1).cho
...
 
+/-- Summing `(i+k).choose k` for `i ∈ [0, n]` gives `(n + (k + 1)).choose (k + 1)`.
+
+Combinatorial interpretation: `(i+k).choose k` is the number of decompositions of `[0, i)` in
+`k + 1` (possibly empty) intervals (this follows from a stars and bars description). In particular,
+`(n + (k + 1)).choose (k + 1)` corresponds to decomposing `[0, n)` into `k + 2` intervals. By
+putting away the last interval (of some length `n-i`), we have to decompose the remaining interval
+`[0, i)` into `k+1` intervals, hence the sum.
+-/
+lemma sum_range_add_choose (k n : ℕ) :
```
Parcly-Taxel, line 139: Can you note that this is just a shifted version of my `sum_Icc_choose` right above?

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -128,6 +128,24 @@ theorem sum_Icc_choose (n k : ℕ) : ∑ m ∈ Icc k n, m.choose k = (n + 1).cho
...
+    ∑ i ∈ Finset.range (n + 1), (i + k).choose k = (n + (k + 1)).choose (k + 1) := by
+  induction n with
+  | zero => simp
+  | succ n ih =>
+      have : n + 1 + (k + 1) = (n + (k + 1)) + 1 := by omega
+      rw [this, choose_succ_succ', sum_range_succ, ← ih, add_comm]
+      congr 2
+      omega
```
Parcly-Taxel, line 147:
```suggestion
/-- **Zhu Shijie's identity** aka hockey-stick identity. Version with `range` instead of `Icc`. -/
lemma sum_range_add_choose (n k : ℕ) :
    ∑ i ∈ range (n + 1), (i + k).choose k = (n + k + 1).choose (k + 1) := by
  rw [← sum_Icc_choose (n + k) k, range_eq_Ico]
  convert (sum_map _ (addRightEmbedding k) (·.choose k)).symm using 2
  rw [map_add_right_Ico, zero_add, add_right_comm]; rfl
```

Parcly-Taxel: In fact I've opened #15935 to add this lemma properly.

sgouezel: Thanks for your shorter proof! However, you shouldn't have removed the combinatorial interpretation, because I'm pretty sure people looking for this lemma will grep for "stars and bars", so I'm adding it back.

Mathlib/Analysis/SpecificLimits/Normed.lean
```diff
@@ -327,54 +331,128 @@ end Geometric
...
+          _ = 2 ^ k * n ^ k := Nat.mul_pow 2 n k
+      convert hasSum_sum_range_mul_of_summable_norm' I1 ih.summable
+        (summable_norm_geometric_of_norm_lt_one hr) (summable_geometric_of_norm_lt_one hr) using 1
+      · ext n
+        have : ∑ i ∈ Finset.range (n + 1), ↑((i + k).choose k) * r ^ i * r ^ (n - i) =
+            ∑ i ∈ Finset.range (n + 1), ↑((i + k).choose k) * r ^ n := by
+          apply Finset.sum_congr rfl (fun i hi ↦ ?_)
+          simp only [Finset.mem_range] at hi
+          rw [mul_assoc, ← pow_add, show i + (n - i) = n by omega]
+        simp only [this, ← sum_mul, ← Nat.cast_sum, sum_range_add_choose k n]
```
Parcly-Taxel, line 394:
```suggestion
        simp only [this, ← sum_mul, ← Nat.cast_sum, sum_range_add_choose n k]
```

sgouezel:
@Parcly-Taxel,you shouldn't have pushed to the branch. In general, don't ever push to a branch unless it's yours, or if it's tagged `help-wanted`. The change you have done removing the parentheses doesn't look good, because `k + 1` is a single unit, so it was kept together on purpose.

If you have suggestions to make, the correct way is to suggest these using the github interface, and let the owner of the branch accept them or not.

By the way, thanks a lot for the review, this is much appreciated and very useful!

Mathlib/Algebra/BigOperators/Fin.lean
```diff
@@ -91,6 +91,11 @@ theorem prod_cons [CommMonoid β] {n : ℕ} (x : β) (f : Fin n → β) :
     (∏ i : Fin n.succ, (cons x f : Fin n.succ → β) i) = x * ∏ i : Fin n, f i := by
   simp_rw [prod_univ_succ, cons_zero, cons_succ]
 
+@[to_additive]
+theorem prod_snoc [CommMonoid β] {n : ℕ} (x : β) (f : Fin n → β) :
+    (∏ i : Fin n.succ, (snoc f x : Fin n.succ → β) i) = (∏ i : Fin n, f i) * x := by
```
urkud, line 96: Is it a good `simp` lemma? Same question about `prod_cons`.

Mathlib/Order/Filter/AtTopBot.lean
```diff
@@ -1314,6 +1314,21 @@ theorem tendsto_finset_preimage_atTop_atTop {f : α → β} (hf : Function.Injec
   (Finset.monotone_preimage hf).tendsto_atTop_finset fun x =>
     ⟨{f x}, Finset.mem_preimage.2 <| Finset.mem_singleton_self _⟩
 
+lemma tendsto_finset_prod_atTop :
+    Tendsto (fun (p : Finset ι × Finset ι') ↦ p.1 ×ˢ p.2) (atTop ×ˢ atTop) atTop := by
```
urkud, line 1318: Given the next lemma, should it be `Tendsto _ atTop atTop`?

Mathlib/Data/Nat/Factorial/BigOperators.lean
```diff
@@ -37,4 +38,43 @@ theorem descFactorial_eq_prod_range (n : ℕ) : ∀ k, n.descFactorial k = ∏ i
   | 0 => rfl
   | k + 1 => by rw [descFactorial, prod_range_succ, mul_comm, descFactorial_eq_prod_range n k]
 
+lemma exists_descFactorial_eq_polynomial (k : ℕ) : ∃ (a : Fin k → ℕ), ∀ n,
+    (n + k).descFactorial k = n ^ k + ∑ i, a i * n ^ (i : ℕ) := by
```
urkud, line 42: Does it make sense to define this as a `Polynomial Nat`?

Mathlib/Data/Nat/Choose/Bounds.lean
```diff
@@ -26,13 +26,20 @@ variable {α : Type*} [LinearOrderedSemifield α]
...
   · norm_cast
     rw [← Nat.descFactorial_eq_factorial_mul_choose]
     exact n.descFactorial_le_pow r
   exact mod_cast r.factorial_pos
 
+lemma choose_le_descFactorial (n k : ℕ) : n.choose k ≤ n.descFactorial k := by
+  rw [choose_eq_descFactorial_div_factorial]
+  exact Nat.div_le_self _ _
+
+lemma choose_le_pow (n k : ℕ) : n.choose k ≤ n ^ k :=
```
urkud, line 40: Please add a docstring saying that this name was used for a lemma which is now called `choose_le_pow_div`.

Mathlib/Data/Nat/Factorial/BigOperators.lean
sgouezel: Indeed, it's ascPochhammer up to a shift, so I have deleted this and used directly ascPochhammer in the main proof. Thanks!

Mathlib/Algebra/Polynomial/Degree/Lemmas.lean
```diff
@@ -345,9 +352,8 @@ theorem natDegree_comp : natDegree (p.comp q) = natDegree p * natDegree q := by
       natDegree_C, mul_zero]
   · by_cases p0 : p = 0
     · simp only [p0, zero_comp, natDegree_zero, zero_mul]
-    refine le_antisymm natDegree_comp_le (le_natDegree_of_ne_zero ?_)
-    simp only [coeff_comp_degree_mul_degree q0, p0, mul_eq_zero, leadingCoeff_eq_zero, or_self_iff,
-      ne_zero_of_natDegree_gt (Nat.pos_of_ne_zero q0), pow_ne_zero, Ne, not_false_iff]
+    · simp only [Ne, mul_eq_zero, leadingCoeff_eq_zero, p0, natDegree_comp_eq_of_mul_ne_zero,
+      ne_zero_of_natDegree_gt (Nat.pos_of_ne_zero q0), not_false_eq_true, pow_ne_zero, or_self]
```
Ruben-VandeVelde, line 356:
```suggestion
        ne_zero_of_natDegree_gt (Nat.pos_of_ne_zero q0), not_false_eq_true, pow_ne_zero, or_self]
```

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -126,6 +126,21 @@ theorem sum_Icc_choose (n k : ℕ) : ∑ m ∈ Icc k n, m.choose k = (n + 1).cho
...
+`k + 1` (possibly empty) intervals (this follows from a stars and bars description). In particular,
+`(n + k + 1).choose (k + 1)` corresponds to decomposing `[0, n)` into `k + 2` intervals.
+By putting away the last interval (of some length `n - i`),
+we have to decompose the remaining interval `[0, i)` into `k + 1` intervals, hence the sum. -/
+lemma sum_range_add_choose (n k : ℕ) :
+    ∑ i ∈ Finset.range (n + 1), (i + k).choose k = (n + k + 1).choose (k + 1) := by
+  rw [← sum_Icc_choose, range_eq_Ico]
+  convert (sum_map _ (addRightEmbedding k) (·.choose k)).symm using 2
+  rw [map_add_right_Ico, zero_add, add_right_comm]
+  rfl
```
Ruben-VandeVelde, line 142: What's the `rfl` doing? Is there a missing lemma?

sgouezel: It's for `Icc k (n + k) = Ico k (n + k + 1)`. We have the `Set` version, in `Order.Ico_succ_right` (not imported by this file, though), but not the `Finset` version, so I'm happy to leave that to `rfl`. The other option is to port the material on set intervals from `Mathlib.Order.SuccPred.Basic`  to finset intervals, but I think it's a little bit too much for the current PR.

Ruben-VandeVelde (approved):
maintainer merge

Mathlib/Data/Nat/Choose/Sum.lean
urkud: Is it `Nat.Ico_succ_right`?

urkud: Otherwise LGTM. Thanks a lot!
bors d+

sgouezel:
bors r+
Thanks for the review!

PR #15274, opened 2024-07-29
feat(CStarAlgebra): vectors with entries in a C⋆-algebra

dupuisf:
This PR defines the type `CstarVec n A` as a copy of `n → A` meant for the case where `A` is a C⋆-algebra. This type is then given a Hilbert C⋆-module structure, namely an `A`-valued inner product and the norm `‖v‖ = √‖⟪v, v⟫‖`.

- [ ] depends on: #15244

---


Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+  inner v w := ∑ i : n, star (v i) * w i
+
+noncomputable instance instNorm : Norm (CStarVec n A) where
+  norm v := √‖⟪v, v⟫_A‖
+
+lemma inner_eq_sum {v w : CStarVec n A} : ⟪v, w⟫_A = ∑ i : n, star (v i) * w i := rfl
+
+lemma norm_eq_sum {v : CStarVec n A} : ‖v‖ = √‖∑ i : n, star (v i) * v i‖ := rfl
+
+lemma inner_single_eq_entry [DecidableEq n] {v : CStarVec n A} {i : n} {a : A} :
```
j-loreaux, line 129: How about adding the `right` version of this too?
```suggestion
lemma inner_single_left [DecidableEq n] {v : CStarVec n A} {i : n} {a : A} :
```

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+
+instance instInner : Inner A (CStarVec n A) where
+  inner v w := ∑ i : n, star (v i) * w i
+
+noncomputable instance instNorm : Norm (CStarVec n A) where
+  norm v := √‖⟪v, v⟫_A‖
+
+lemma inner_eq_sum {v w : CStarVec n A} : ⟪v, w⟫_A = ∑ i : n, star (v i) * w i := rfl
+
+lemma norm_eq_sum {v : CStarVec n A} : ‖v‖ = √‖∑ i : n, star (v i) * v i‖ := rfl
```
j-loreaux, line 127: And also add `norm_sq_eq_sum`?
```suggestion
lemma norm_eq_sqrt_sum {v : CStarVec n A} : ‖v‖ = √‖∑ i : n, star (v i) * v i‖ := rfl
```

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+    apply Eq.symm
+    simp [Finset.sum_mul, mul_assoc]
+  inner_smul_right_complex {z} {x} {y} := by
+    simp only [inner_eq_sum, smul_apply, smul_mul_assoc]
+    apply Eq.symm
+    simp only [Finset.smul_sum, mul_smul_comm]
+  star_inner x y := by simp only [inner, star_sum, star_mul, star_star]
+  norm_eq_sqrt_norm_inner_self v := rfl
+
+lemma norm_entry_le_norm [DecidableEq n] [CompleteSpace A] {v : CStarVec n A} {i : n} :
```
j-loreaux, line 168: Is this "entry" terminology used in other parts of the library? 
```suggestion
lemma norm_apply_le_norm [DecidableEq n] [CompleteSpace A] {v : CStarVec n A} {i : n} :
```

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+-/
+
+variable {A : Type*} [NonUnitalNormedRing A] [StarRing A] [CStarRing A] [PartialOrder A]
+  [CompleteSpace A] [StarOrderedRing A] [NormedSpace ℂ A] [StarModule ℂ A] [IsScalarTower ℂ A A]
+  [SMulCommClass ℂ A A]
+
+variable {n : Type*} [Fintype n]
+
+/-- A temporary "bad" instance of `NormedAddCommGroup` that induces a uniformity that is not
+defeq with the product uniformity. -/
```
j-loreaux, line 207:
```suggestion
/-- A temporary "bad" instance of `NormedAddCommGroup` that induces a uniformity that is not
defeq with the product uniformity.

This is not intended for use outside of this file. -/
```

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+private noncomputable abbrev normedAddCommGroupAux : NormedAddCommGroup (CStarVec n A) :=
+  NormedAddCommGroup.ofCore CStarModule.normedSpaceCore
+
+attribute [local instance] normedAddCommGroupAux
+
+/-- A temporary "bad" instance of `NormedSpace`. -/
```
j-loreaux, line 213: Ditto

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+private noncomputable abbrev normedSpaceAux : NormedSpace ℂ (CStarVec n A) :=
+  .ofCore CStarModule.normedSpaceCore
+
+attribute [local instance] normedSpaceAux
+
+open Finset Classical in
+private lemma lipschitzWith_ofFun_symm_aux :
+    LipschitzWith 1 (ofFun.symm : CStarVec n A → n → A) := by
```
j-loreaux, line 221: Are these proofs essentially just copied from the analogous ones for `EuclideanSpace`?

j-loreaux (commented):
Nice! Mainly naming issues. Some of the proofs seem overly hard, but I'm on mobile, so maybe I'm wrong!

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
dupuisf: It's also used a bit in `Matrix`, but it arguably makes more sense there. But yeah, `apply` is a bit more consistent with the rest of mathlib here.

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
dupuisf: No, I mostly looked at `PiLp` for inspiration for this file actually. I followed the same general strategy, but I reproved the Lipschitz/Antilipschitz from scratch.

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,329 @@
...
+replace the uniformity and bornology by the Pi ones when registering the
+`NormedAddCommGroup (CStarVec n A)` instance. See the docstring of the `TopologyAux` section below
+for more details.
+
+-/
+
+open scoped ComplexOrder RightActions Topology Uniformity Bornology
+
+/-- A vector with entries in a C⋆-algebra and a Hilbert C⋆-module structure. -/
+def CStarVec (n : Type*) (A : Type*) := n → A
```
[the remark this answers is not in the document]
dupuisf: Yeah, I'm still not too sure what the best approach is here. I went with `irreducible` for `E →WOT[𝕜] F`, but then I saw that e.g. `PiLp` is even an abbrev to inherit all the instances from `WithLp`, and it doesn't seem to cause major problems (but maybe I just didn't see them). I actually went with `abbrev` for `CStarMatrix` also to inherit everything from matrices, basically copying the `PiLp` design.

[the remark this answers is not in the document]
j-loreaux: Well, the `PiLp` bit is basically just to have nicer syntax. It's `WithLp` that is the type synonym barrier. That is a default def, neither abbrev nor irreducible. This is, essentially exactly what caused the issues in `EuclideanSpace` which is itself another `abbrev` for a `WithLp` variant. The problem is that Lean would see through `EuclideanSpace K n`, which is abbreviated for `WithLp 2 (n -> K)`, to bare functions `n -> K` if the user was ever so slightly sloppy. My thinking is that this is basically the same situation. So I'm convincing myself to advocate for irreducible. Especially success we can avoid pain whenever we need to with `unseal`. This way we keep the barrier tight and force users to use `ofFun` and the `CoeFun` instance, which is what they should be doing already to avoid defeq abuse.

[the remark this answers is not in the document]
dupuisf: I tried to convert this to `irreducible` and it's surprisingly painful. You can't just unseal every lemma that gives you trouble (stuff like `zero_apply` and friends), because once those things are defined via defeq with functions, you can't use them later when resealed. So you have to redo the whole hierarchy of `SubNegMonoid` and friends by hand with `ofFun`. I think this would work fine if we could state lemmas in "sealed mode" and then unseal only in the proof.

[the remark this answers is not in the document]
j-loreaux: Aha, I see. Thanks for trying.

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,340 @@
...
+@[simp] theorem sub_apply [Sub A] {i : n} : (v - w) i = v i - w i := rfl
+
+@[simp] theorem smul_apply (c : R) [SMul R A] {i : n} : (c • v) i = c • v i := rfl
+
+@[simp] theorem neg_apply [Neg A] {i : n} : (-v) i = -v i := rfl
+
+theorem finset_sum_fn {ι : Type*} {s : Finset ι} [AddCommMonoid A] {f : ι → CStarVec n A} :
+    ∑ i ∈ s, ofFun (f i) = (fun j => ∑ i ∈ s, f i j) := by
+  rw [Finset.sum_fn]
+  rfl
```
Ruben-VandeVelde, line 109: What's this `rfl` unfolding?

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,340 @@
...
+
+lemma inner_single_left [DecidableEq n] {v : CStarVec n A} {i : n} {a : A} :
+    ⟪ofFun (Pi.single i a), v⟫_A = star a * v i := by
+  simp [inner_eq_sum, ofFun_apply, Pi.single_apply]
+  have hmain :
+      (fun x => star (if x = i then a else 0) * v x)
+        = fun x => if x = i then star a * v x else 0 := by
+    ext
+    split_ifs with h <;> simp [h, mul_assoc]
+  simp [hmain, Finset.sum_ite]
```
Ruben-VandeVelde, line 140: Does this work?
```suggestion
  have hmain :
      (fun x => star (if x = i then a else 0) * v x)
        = fun x => if x = i then star a * v x else 0 := by
    ext
    split_ifs with h <;> simp [h, mul_assoc]
  simp [inner_eq_sum, ofFun_apply, Pi.single_apply, hmain, Finset.sum_ite]
```

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,340 @@
...
+
+lemma inner_single_right [DecidableEq n] {v : CStarVec n A} {i : n} {a : A} :
+    ⟪v, ofFun (Pi.single i a)⟫_A = star (v i) * a := by
+  simp [inner_eq_sum, ofFun_apply, Pi.single_apply]
+
+@[simp] lemma norm_single {i : n} {a : A} [DecidableEq n] : ‖ofFun (Pi.single i a)‖ = ‖a‖ := by
+  simp [norm_eq_sqrt_sum, Pi.single_apply, CStarRing.norm_star_mul_self]
+
+instance instCStarModule : CStarModule A (CStarVec n A) where
+  inner_add_right {v} {w₁} {w₂} := by
```
Ruben-VandeVelde, line 150: This could be
```suggestion
  inner_add_right {v w₁ w₂} := by
```
according to your preference

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
```diff
@@ -0,0 +1,340 @@
...
+      exact (CStarRing.star_mul_self_eq_zero_iff (x i)).mp <| h i (Finset.mem_univ i)
+    · rintro rfl; simp only [inner, CStarVec.zero_apply, star_zero, mul_zero, Finset.sum_const_zero]
+  inner_op_smul_right {a} {x} {y} := by
+    simp only [inner, smul_apply, smul_eq_mul, mul_assoc]
+    apply Eq.symm
+    simp [Finset.sum_mul, mul_assoc]
+  inner_smul_right_complex {z} {x} {y} := by
+    simp only [inner_eq_sum, smul_apply, smul_mul_assoc]
+    apply Eq.symm
+    simp only [Finset.smul_sum, mul_smul_comm]
```
Ruben-VandeVelde, line 171: I'm surprised you need the `apply Eq.symm` here. Am I missing something?

Mathlib/Analysis/CStarAlgebra/CStarVec.lean
j-loreaux: I think there's some defeq abuse here. I have a PR forthcoming which will change the scope of this PR, and it should address this defeq abuse too. The key is to use the coercion to function manually, because otherwise Lean sees a function application, asks itself "is the type of the thing in the function position a function?", finds the answer is "yes", but commits defeq abuse doing so.

j-loreaux:
@dupuisf is there anything else that we should snag from this PR that I missed? Or should we close it? Or do you think we should still have `CStarVec` as an `abbrev` for `C⋆ᵐᵒᵈ (n → A)` (which would be weird because it's actually longer given the notation!)?

dupuisf:
No, I think we can just close this one.

PR #15276, opened 2024-07-29
[Merged by Bors] - chore(scripts/macos_install.sh): avoid using `brew` completely 

mattrobball:
This script is intended as a single step for a beginner user to get started with Lean. As such, we should avoid third-party packages from package managers whose provenance can be dubious. We should also avoid requiring the user to install a lot of software. We have already removed homebrew from the `elan` install. This PR removes it from the VS Code step as well, avoiding the dependency on `brew` completely. We download from the official Microsoft site instead.

---

### DIFF
--- a/scripts/install_macos.sh
+++ b/scripts/install_macos.sh
@@ -2,23 +2,24 @@
 
 set -exo pipefail
 
-# Install Homebrew
-if ! which brew > /dev/null; then
-    # Install Homebrew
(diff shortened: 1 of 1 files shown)


joneugster (approved):
Haven't tested it, but looks good to me!

maintainer merge

semorrison:
Tested locally, worked great!

bors merge

PR #15277, opened 2024-07-29
[Merged by Bors] - feat(CStarAlgebra): matrices with entries in a C⋆-algebra

dupuisf:
This PR creates the type `CstarMatrix m n A` as a copy of `Matrix m n A` meant for the case where `A` is a C⋆-algebra. Most notably, we show that square matrices with entries in a C⋆-algebra form a C⋆-algebra.

- [x] depends on: #15244 
- [x] depends on: #15274 

---


Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+Since `Matrix m n` does not have a norm instance on it, we make `CStarMatrix` reducible (i.e. an
+`abbrev`) in order to get access to the matrix API, which doesn't conflict with the additional
+structure we are constructing here.
+
+## Notation
+
++ We locally use the notation `n →C⋆ A` for `WithCStarModule (n → A)`.
+-/
+
+set_option maxSynthPendingDepth 2
```
dupuisf, line 40: Note that this was needed for several lemmas. I also added several shortcut instances in the `WithCStarModule` material to deal with instance synthesis problems.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+    · obtain ⟨i, _, hi⟩ := exists_mem_eq_sup (univ : Finset m) (univ_nonempty_iff.mpr hm_triv)
+        fun b => Finset.univ.sup fun b_1 => ‖toMatrixAux M b b_1‖₊
+      obtain ⟨j, _, hj⟩ := exists_mem_eq_sup (univ : Finset n) (univ_nonempty_iff.mpr hn_triv)
+        fun b_1 => ‖toMatrixAux M i b_1‖₊
+      rw [hi, hj, one_mul]
+      exact CStarMatrix.norm_entry_le_norm
+    · simp only [not_nonempty_iff] at hn_triv
+      simp [Finset.sup_eq_bot_of_isEmpty, bot_eq_zero]
+  · simp only [not_nonempty_iff] at hm_triv
+    simp [Finset.sup_eq_bot_of_isEmpty, bot_eq_zero]
```
j-loreaux, line 313:
```suggestion
  rw [one_mul, ← NNReal.coe_le_coe, coe_nnnorm, coe_nnnorm, Matrix.norm_le_iff (norm_nonneg _)]
  exact fun _ _ ↦ CStarMatrix.norm_entry_le_norm
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+        _ = Dn * Dm * ‖M i j‖₊ * ‖v‖₊ * ‖w‖₊ := by ring
+    · simp only [not_nonempty_iff] at hn_triv
+      simp only [Finset.sup_eq_bot_of_isEmpty, Finset.sup_bot]
+      simp only [bot_eq_zero', NNReal.coe_zero, mul_zero, norm_le_zero_iff]
+      ext _ j
+      exact False.elim <| IsEmpty.false j
+  · simp only [not_nonempty_iff] at hm_triv
+    simp [Finset.sup_eq_bot_of_isEmpty, bot_eq_zero]
+    ext i j
+    exact False.elim <| IsEmpty.false i
```
j-loreaux, line 372: This needs, or else the proof breaks, see for yourself.
```lean
attribute [-instance] CStarMatrix.instNorm -- to avoid Lean seeing this instance on `Matrix m n A`?
```
This actually seems like a big problem. Lean sees the `CStarMatrix` norm on `Matrix` because it's an `abbrev`, even when the bit in question is `toMatrixAux A`, which is declared to have type `Matrix m n A`.

It's possible that this won't come up enough in our own work to matter, and probably in other places in the library Lean would have no business inferring `CStarMatrix` in place of `Matrix`, *but* I could imagine people would be pretty upset if they were using their `Matric n n ℂ` and all of a sudden our `CStarMatrix` norm came waltzing in.

It's possible we can't use `abbrev`. What do you think?

```suggestion
open CStarMatrix WithCStarModule in
private lemma antilipschitzWith_toMatrixAux :
    AntilipschitzWith (Fintype.card m * Fintype.card n)
      (toMatrixAux : CStarMatrixAux m n A → Matrix m n A) := by
  refine AddMonoidHomClass.antilipschitz_of_bound _ fun M => ?_
  calc
    ‖M‖ ≤ ∑ i, ∑ j, ‖M i j‖ := by
      rw [norm_def]
      refine (toCLM A M).opNorm_le_bound (by positivity) fun v => ?_
      simp only [toCLM_apply_eq_sum, equiv_symm_pi_apply, Finset.sum_mul]
      apply pi_norm_le_sum_norm _ |>.trans
      gcongr with i _
      apply norm_sum_le _ _ |>.trans
      gcongr with j _
      apply norm_mul_le _ _ |>.trans
      gcongr
      exact norm_apply_le_norm v j
    _ ≤ ∑ _ : m, ∑ _ : n, ‖toMatrixAux M‖ := by
      gcongr with i _ j _
      exact toMatrixAux M |>.norm_entry_le_entrywise_sup_norm
    _ = _ := by simp [mul_assoc]
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+    {w : m →C⋆ A} : ⟪v, toCLM A Mᴴ w⟫_A = ⟪toCLM A M v, w⟫_A := by
+  have : M = Mᴴᴴ := by simp
+  nth_rewrite 2 [this]
+  rw [toCLM_inner_right_eq_left]
+
+/-- The operator norm on `CStarMatrix m n A`. -/
+noncomputable instance instNorm : Norm (CStarMatrix m n A) where
+  norm M := ‖toCLM A M‖
+
+lemma normedSpaceCore [DecidableEq n]: NormedSpace.Core ℂ (CStarMatrix m n A) where
```
j-loreaux, line 155:
```suggestion

lemma norm_def {M : CStarMatrix m n A} : ‖M‖ = ‖toCLM A M‖ := rfl

lemma norm_def' {M : CStarMatrix n n A} : ‖M‖ = ‖toCLMNonUnitalAlgHom A M‖ := rfl

lemma normedSpaceCore [DecidableEq n]: NormedSpace.Core ℂ (CStarMatrix m n A) where
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+      fun _ => Filter.ext_iff.1 CStarMatrixAux.cobounded_eq_aux.symm _
+
+instance instNormedSpace [DecidableEq n] : NormedSpace ℂ (CStarMatrix m n A) :=
+  .ofCore CStarMatrix.normedSpaceCore
+
+protected lemma norm_mul {M₁ M₂ : CStarMatrix n n A} : ‖M₁ * M₂‖ ≤ ‖M₁‖ * ‖M₂‖ := by
+  change ‖toCLMNonUnitalAlgHom A (M₁ * M₂)‖
+    ≤ ‖toCLMNonUnitalAlgHom A M₁‖ * ‖toCLMNonUnitalAlgHom A M₂‖
+  rw [map_mul]
+  exact NormedRing.norm_mul ((toCLMNonUnitalAlgHom A) M₁) ((toCLMNonUnitalAlgHom A) M₂)
```
j-loreaux, line 427: See below
```suggestion
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+
+noncomputable instance instNonUnitalNormedRing [DecidableEq n] :
+    NonUnitalNormedRing (CStarMatrix n n A) where
+  dist_eq _ _ := rfl
+  norm_mul _ _ := CStarMatrix.norm_mul
```
j-loreaux, line 432: norm_mul _ _ := by simpa only [norm_def', map_mul] using norm_mul_le _ _

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+section unital
+
+variable {A : Type*} [NormedRing A] [StarRing A] [CStarRing A] [PartialOrder A]
+  [CompleteSpace A] [StarOrderedRing A] [NormedAlgebra ℂ A] [StarModule ℂ A]
+
+variable {n : Type*} [Fintype n] [DecidableEq n]
+
+noncomputable instance instNormedRing : NormedRing (CStarMatrix n n A) where
+  dist_eq _ _ := rfl
+  norm_mul _ _  := CStarMatrix.norm_mul
```
j-loreaux, line 474:
```suggestion
  norm_mul := norm_mul_le
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+
+noncomputable instance instNormedAlgebra : NormedAlgebra ℂ (CStarMatrix n n A) where
+  norm_smul_le r M := by
+    change ‖toCLM A (r • M)‖ ≤ ‖r‖ * ‖toCLM A M‖
+    simp only [map_smul]
+    exact ContinuousLinearMap.opNorm_smul_le r ((toCLM A) M)
```
j-loreaux, line 480:
```suggestion
  norm_smul_le r M := by simpa only [norm_def, map_smul] using (toCLM A M).opNorm_smul_le r 
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+
+-- TODO: make this non-unital
+instance instPartialOrder : PartialOrder (CStarMatrix n n A) := CStarRing.spectralOrder _
+instance instStarOrderedRing : StarOrderedRing (CStarMatrix n n A) :=
+  CStarRing.spectralOrderedRing _
```
j-loreaux, line 485: Oof, yeah, that's going to be annoying, but sure, I should do that soon.

j-loreaux (commented):
I have lots of simplifications here.

One big issue: I'm not sure that we can use an `abbrev`. See my comment on the proof of `antilipschitzWith_toMatrixAux`.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
dupuisf: I agree this is horrible. I didn't think Lean would try to go "backwards" like this to find an instance (meaning going to `CStarMatrix` to find something about `Matrix`). I'll switch the whole thing from `abbrev` to `def` when I get a bit of time.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+lemma toMatrix_apply {M : Matrix m n A} {i : m} : (toMatrix M) i = M i := rfl
+
+@[ext]
+lemma ext {M₁ M₂ : CStarMatrix m n A} (h : ∀ i j, M₁ i j = M₂ i j) : M₁ = M₂ := Matrix.ext h
+
+variable [Fintype m] [Fintype n] [NonUnitalNormedRing A] [StarRing A] [NormedSpace ℂ A]
+  [PartialOrder A] [CStarRing A] [StarOrderedRing A] [SMulCommClass ℂ A A] [StarModule ℂ A]
+  [IsScalarTower ℂ A A] [CompleteSpace A]
+
+variable (A) in
```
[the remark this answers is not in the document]
dupuisf: I guess not.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,489 @@
...
+      _ ≤ 1 * (‖M‖ * ‖v‖) := by gcongr
+      _ = ‖M‖ * ‖v‖ := by simp
+  case nonempty => exact ⟨0, 0, by simp, by simp⟩
+  case bound =>
+    intro b hb
+    obtain ⟨w, hw₁, hw₂⟩ := hb
+    rw [← hw₂]
+    calc _ ≤ C * ‖v‖ * ‖w‖ := h v w
+      _ ≤ C * ‖v‖ * 1 := by gcongr
+      _ = C * ‖v‖ := by simp
```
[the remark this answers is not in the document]
j-loreaux: This one: `lemma x = 0 ∨ 0 < ‖x‖` for `NormedAddCommGroup`.

I want to be able to write `rfl` in the first argument of `obtain`. I know it's minor, but I've definitely run into this before, and this time it clicked because I did it twice back-to-back.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
j-loreaux: I didn't think it would do that either. This sucks. ☹️

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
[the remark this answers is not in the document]
dupuisf: I just added this lemma.

Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
```diff
@@ -151,13 +158,32 @@ theorem equiv_symm_smul : (equiv E).symm (c • x') = c • (equiv E).symm x' :=
...
+  ext x
+  refine ⟨fun _  => trivial, fun _ => ?_⟩
+  rw [Submodule.mem_map]
+  exact ⟨linearEquiv R E x, by simp⟩
+
+instance instModuleFinite [Semiring R] [AddCommGroup E] [Module R E] [Module.Finite R E] :
+    Module.Finite R (WithCStarModule E) where
+  out := by
+    rw [← map_top_submodule]
+    exact Submodule.FG.map _ ‹Module.Finite R E›.out
```
j-loreaux, line 185: Does `inferInstanceAs` not work?

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+-/
+
+set_option maxSynthPendingDepth 2
+
+open scoped ComplexOrder Topology Uniformity Bornology Matrix NNReal InnerProductSpace
+  WithCStarModule
+
+/-- Type copy `Matrix m n A` meant for matrices with entries in a C⋆-algebra. This is
+a C⋆-algebra when `m = n`. This is an abbrev in order to inherit all instances from `Matrix`,
+which includes the product uniformity, but not a norm. -/
```
j-loreaux, line 39:
```suggestion
a C⋆-algebra when `m = n`. -/
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+
+section basic
+
+variable (m n A) in
+/-- The equivalence between `Matrix m n A` and `CStarMatrix m n A`. -/
+def ofMatrix {m n A : Type*} : Matrix m n A ≃ CStarMatrix m n A := Equiv.refl _
+
+lemma ofMatrix_apply {M : Matrix m n A} {i : m} : (ofMatrix M) i = M i := rfl
+
+lemma ofMatrix_symm_apply {M : CStarMatrix m n A} {i : m} : (ofMatrix.symm M) i = M i := rfl
```
j-loreaux, line 54: Not `@[simp]`?

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+    map_mul' := fun _ _ => rfl
+    map_add' := fun _ _ => rfl }
+
+instance instStarRing [Fintype n] [NonUnitalSemiring A] [StarRing A] :
+    StarRing (CStarMatrix n n A) where
+  star_mul := star_mul (R := Matrix n n A)
+  star_add := star_add (R := Matrix n n A)
+
+instance instAlgebra [Fintype n] [DecidableEq n] [CommSemiring R] [Semiring A] [Algebra R A] :
+    Algebra R (CStarMatrix n n A) where
```
j-loreaux, line 333: `inferInstanceAs` doesn't work?

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+
+lemma toCLM_apply {M : CStarMatrix m n A} {v : C⋆ᵐᵒᵈ (n → A)} :
+    toCLM M v = (WithCStarModule.equiv _).symm (M.mulVec v) := rfl
+
+lemma toCLM_apply_eq_sum {M : CStarMatrix m n A} {v : C⋆ᵐᵒᵈ (n → A)} :
+    toCLM M v = (WithCStarModule.equiv _).symm (fun i => ∑ j, M i j * v j) := by
+  ext i
+  simp [toCLM_apply, Matrix.mulVec, Matrix.dotProduct]
+
+/-- Interpret a `CStarMatrix m n A` as a continuous linear map acting on `n →C⋆ A`. This
```
j-loreaux, line 386: I don't think we should use the local notation in docstrings.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+section
+
+variable {m n A : Type*}
+
+lemma uniformEmbedding_ofMatrix [NonUnitalNormedRing A] :
+    UniformEmbedding (ofMatrix : Matrix m n A → CStarMatrix m n A) where
+  comap_uniformity := Filter.comap_id'
+  inj := fun ⦃_ _⦄ a ↦ a
+
+/-- `ofMatrix` bundled as a continuous linear map. -/
```
j-loreaux, line 671:
```suggestion
/-- `ofMatrix` bundled as a continuous linear equivalence. -/
```

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,694 @@
...
+## Implementation notes
+
+The norm on this type induces the product uniformity and bornology, but these are not defeq to
+`Pi.uniformSpace` and `Pi.instBornology`. Hence, we prove the equality to the Pi instances and
+replace the uniformity and bornology by the Pi ones when registering the
+`NormedAddCommGroup (CStarMatrix m n A)` instance. See the docstring of the `TopologyAux` section
+below for more details.
+-/
+
+set_option maxSynthPendingDepth 2
```
j-loreaux, line 32: Where exactly is this needed? Can you just activate it locally (even in the proof) so we can see exactly where the problems occur?

j-loreaux (commented):
Sorry I let this sit.

Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
dupuisf: It does.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
dupuisf: I've localized it to the handful of places where it's needed.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
dupuisf: Yes, I just had to use it a bit more systematically in the file. Fixed.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,680 @@
...
+  rw [injective_iff_map_eq_zero]
+  intro M h
+  ext i j
+  rw [Matrix.zero_apply, ← norm_eq_zero, ← sq_eq_zero_iff, sq, ← CStarRing.norm_self_mul_star,
+    ← toCLM_apply_single_apply]
+  simp [h]
+
+open WithCStarModule in
+lemma toCLM_inner_right_eq_left {M : CStarMatrix m n A} {v : C⋆ᵐᵒᵈ (m → A)}
+    {w : C⋆ᵐᵒᵈ (n → A)} : ⟪v, toCLM M w⟫_A = ⟪toCLM Mᴴ v, w⟫_A := by
```
j-loreaux, line 412: Shouldn't this be called:
```suggestion
lemma inner_toCLM_right {M : CStarMatrix m n A} {v : C⋆ᵐᵒᵈ (m → A)}
    {w : C⋆ᵐᵒᵈ (n → A)} : ⟪v, toCLM M w⟫_A = ⟪toCLM Mᴴ v, w⟫_A := by
```
Or maybe the direction should be reversed and it should called `inner_toCLM_conjTranspose_left`.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,680 @@
...
+  simp only [toCLM_apply_eq_sum, pi_inner, equiv_symm_pi_apply, inner_def, Finset.mul_sum,
+    Matrix.conjTranspose_apply, star_sum, star_mul, star_star, Finset.sum_mul]
+  rw [Finset.sum_comm]
+  simp_rw [mul_assoc]
+
+lemma toCLM_inner_conjTranspose_right_eq_left {M : CStarMatrix m n A} {v : C⋆ᵐᵒᵈ (n → A)}
+    {w : C⋆ᵐᵒᵈ (m → A)} : ⟪v, toCLM Mᴴ w⟫_A = ⟪toCLM M v, w⟫_A := by
```
j-loreaux, line 419: Same comments as above (except this one seems to be the right way round).

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,680 @@
...
+    _ = _ := by simp [mul_assoc]
+
+private lemma uniformInducing_toMatrixAux [DecidableEq m] :
+    IsUniformInducing (ofMatrix.symm : CStarMatrix m n A → Matrix m n A) :=
+  AntilipschitzWith.isUniformInducing antilipschitzWith_toMatrixAux
+    lipschitzWith_toMatrixAux.uniformContinuous
+
+private lemma uniformity_eq_aux [DecidableEq m] :
+    𝓤 (CStarMatrix m n A) = (𝓤[Pi.uniformSpace _] :
+      Filter (CStarMatrix m n A × CStarMatrix m n A)) := by
```
j-loreaux, line 538: I think you likely shouldn't even need this lemma and the next one (or else you can make the proofs much easier) if you use the results in https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/MetricSpace/Bilipschitz.html

j-loreaux (commented):
About providing all the instances: do you expect this to be a performance benefit or drag? From a mathematical point of view, most of the lower level ones are pointless because we'll never be in a situation where we'll have only those but not the higher ones (because our entries will be C⋆-algebras).

Or am I missing something?

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
dupuisf: I went for the second suggestion.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,680 @@
...
+  __ := instAddMonoidWithOne
+
+instance instAddCommGroupWithOne [AddCommGroupWithOne A] :
+    AddCommGroupWithOne (CStarMatrix n n A) where
+  __ := instAddCommGroup
+  __ := instAddGroupWithOne
+
+@[default_instance 100]
+instance {l : Type*} [Fintype m] [Mul A] [AddCommMonoid A] :
+    HMul (CStarMatrix l m A) (CStarMatrix m n A) (CStarMatrix l n A) where
```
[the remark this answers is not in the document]
dupuisf: Yeah. I just copied over the comment from the corresponding instance on `Matrix`.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
```diff
@@ -0,0 +1,680 @@
...
+@[simp] theorem smul_apply [SMul B A] (r : B) (M : Matrix m n A) (i : m) (j : n) :
+    (r • M) i j = r • (M i j) := rfl
+
+@[simp] theorem sub_apply [Sub A] (M N : Matrix m n A) (i : m) (j : n) :
+    (M - N) i j = (M i j) - (N i j) := rfl
+
+@[simp] theorem neg_apply [Neg A] (M : Matrix m n A) (i : m) (j : n) :
+    (-M) i j = -(M i j) := rfl
+
+/-! simp-normal form pulls `of` to the outside. -/
```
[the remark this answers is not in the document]
dupuisf: This is also just copied over from `Matrix`.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
dupuisf: That would entail using `comap` to define the topology, uniformity and bornology on `CStarMatrix` instead of just using `Pi.topologicalSpace` and so on as is done for `Matrix`, `piLp` etc. We've had this debate before, but I still feel that this pattern is so prevalent in mathlib that we would've noticed by now if this were a major problem.

dupuisf:
> About providing all the instances: do you expect this to be a performance benefit or drag? From a mathematical point of view, most of the lower level ones are pointless because we'll never be in a situation where we'll have only those but not the higher ones (because our entries will be C⋆-algebras).
> 
> Or am I missing something?

My guess is that there might be a performance benefit to doing this (at least I had convinced myself that this was the case a few months ago): whenever using a lemma that requires one of the lower classes, we don't have to go around the whole hierarchy to derive the instance from `[CStarAlgebra A]`.

Mathlib/Analysis/CStarAlgebra/CStarMatrix.lean
j-loreaux: Oh that's right, I forgot. Sorry

j-loreaux:
Thanks!

bors merge

PR #15290, opened 2024-07-30
[Merged by Bors] - feat(FieldTheory/Normal): Intersection of normal extensions is normal

tb65536:
This PR proves that an intersection of normal extensions is normal (we already have that a compositum of normal extensions is normal).

---


Ruben-VandeVelde (approved):
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15291, opened 2024-07-30
[Merged by Bors] - feat: Add partialSups lemmas

RemyDegenne:
---


Mathlib/Order/PartialSups.lean
```diff
@@ -94,6 +94,9 @@ theorem Monotone.partialSups_eq {f : ℕ → α} (hf : Monotone f) : (partialSup
 theorem partialSups_mono : Monotone (partialSups : (ℕ → α) → ℕ →o α) := fun _f _g h _n ↦
   partialSups_le_iff.2 fun k hk ↦ (h k).trans (le_partialSups_of_le _ hk)
 
+lemma monotone_partialSups (f : ℕ → α) : Monotone (partialSups f) :=
```
sgouezel, line 97:
```suggestion
lemma partialSups_monotone (f : ℕ → α) : Monotone (partialSups f) :=
```
There is a special naming convention for `monotone` and friends...

sgouezel (commented):
bors d+

PR #15292, opened 2024-07-30
[Merged by Bors] - feat: add `Pairwise.pairwiseDisjoint`

RemyDegenne:
`Pairwise (Disjoint on f)` implies `s.PairwiseDisjoint f` for any set `s`.

---

### DIFF
--- a/Mathlib/Data/Set/Pairwise/Basic.lean
+++ b/Mathlib/Data/Set/Pairwise/Basic.lean
@@ -293,6 +293,9 @@ lemma pairwiseDisjoint_range_iff {α β : Type*} {f : α → (Set β)} :
   · rintro h _ ⟨x, rfl⟩ _ ⟨y, rfl⟩ hxy
     exact (h x y).resolve_left hxy
 
+lemma _root_.Pairwise.pairwiseDisjoint (h : Pairwise (Disjoint on f)) (s : Set ι) :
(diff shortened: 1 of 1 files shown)


grunweg (commented):
Looks useful, thanks!
maintainer merge

Mathlib/Data/Set/Pairwise/Basic.lean
```diff
@@ -293,6 +293,9 @@ lemma pairwiseDisjoint_range_iff {α β : Type*} {f : α → (Set β)} :
   · rintro h _ ⟨x, rfl⟩ _ ⟨y, rfl⟩ hxy
     exact (h x y).resolve_left hxy
 
+lemma _root_.Pairwise.pairwiseDisjoint (h : Pairwise (Disjoint on f)) (s : Set ι) :
```
Vierkantor, line 296: I had to think for a moment what the differences are between the hypothesis and conclusion. A docstring can't hurt :)
```suggestion
/-- If the range of `f` is pairwise disjoint, then the image of any set `s` under `f` is as well. -/
lemma _root_.Pairwise.pairwiseDisjoint (h : Pairwise (Disjoint on f)) (s : Set ι) :
```

Vierkantor (approved):
Thanks!

bors d+

PR #15294, opened 2024-07-30
[Merged by Bors] - feat: two Finset lemmas

RemyDegenne:
Add `prod_image_of_disjoint` (and the `to_additive` version `sum_image_of_disjoint`) and `sUnion_disjiUnion`

---


j-loreaux (commented):
Thanks! We can simplify at bit. There's definitely a missing `@[congr]` attribute.

Mathlib/Algebra/BigOperators/Group/Finset.lean
```diff
@@ -1178,6 +1178,21 @@ lemma mulSupport_prod (s : Finset ι) (f : ι → α → β) :
   simp only [mulSupport_subset_iff', Set.mem_iUnion, not_exists, nmem_mulSupport]
   exact fun x ↦ prod_eq_one
 
+@[to_additive]
```
sgouezel, line 1181: Could you add a docstring to this lemma (and to its additivized version)? I had to read it several times to understand what it means. Thanks!

j-loreaux: That's a good point. While I didn't have too much trouble understanding the idea, it *did* take me some time to understand why `PairwiseDisjoint` is needed, so commenting on that would be especially useful (especially since we already have [Finset.prod_map](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/Group/Finset.html#Finset.prod_map), where this assumption is not necessary, since `Finset.map` uses an embedding but `Finset.image` doesn't).

Ruben-VandeVelde (approved):
This seems okay to me now.

maintainer merge

Mathlib/Algebra/BigOperators/Group/Finset.lean
```diff
@@ -1179,6 +1179,31 @@ lemma mulSupport_prod (s : Finset ι) (f : ι → α → β) :
   simp only [mulSupport_subset_iff', Set.mem_iUnion, not_exists, nmem_mulSupport]
   exact fun x ↦ prod_eq_one
 
+/-- The product of a function `g` over the image of a finset `I` by a function `f` is equal to
+the product over `I` of `g ∘ f`.
+This lemma supposes that the images by `f` of two distinct elements of `I` are different,
+unless they are both `⊥` (hypothesis `hf_disj`), and that `g ⊥ = 1`.
```
urkud, line 1185: `hf_disj` is a stronger assumption than you say here.

Mathlib/Algebra/BigOperators/Group/Finset.lean
```diff
@@ -1179,6 +1179,31 @@ lemma mulSupport_prod (s : Finset ι) (f : ι → α → β) :
...
+If `f` is injective on `I`, use `Finset.prod_image` instead. See also `Finset.prod_map`. -/
+@[to_additive
+"The sum of a function `g` over the image of a finset `I` by a function `f` is equal to
+the sum over `I` of `g ∘ f`.
+This lemma supposes that the images by `f` of two distinct elements of `I` are different,
+unless they are both `⊥` (hypothesis `hf_disj`), and that `g ⊥ = 0`.
+If `f` is injective on `I`, use `Finset.sum_image` instead. See also `Finset.sum_map`."]
+lemma prod_image_of_disjoint [PartialOrder α] [OrderBot α] [DecidableEq α]
+    (hg_bot : g ⊥ = 1) {f : ι → α} {I : Finset ι} (hf_disj : (I : Set ι).PairwiseDisjoint f) :
```
eric-wieser, line 1194: I think using the ambient `g` variable (but not also `f`) makes this lemma confusing
```suggestion
     {f : ι → α} {g : α → β} (hg_bot : g ⊥ = 1) {I : Finset ι} (hf_disj : (I : Set ι).PairwiseDisjoint f) :
```

eric-wieser (commented):
Is this more general lemma true?
```lean
/-- If the images only overlap where the value is 1, then the function can move inside the binder -/
lemma prod_image_of_pairwise [PartialOrder α] [OrderBot α] [DecidableEq α]
    {f : ι → α} {g : α → β} {I : Finset ι}
    (hf_disj : (I : Set ι).Pairwise fun i j => f i = f j → g (f i) = 1) :
    ∏ s in I.image f, g s = ∏ i in I, g (f i) := by
```
I think the order is sort of a distraction here. Your lemma then follows as
```lean
by
  refine prod_image_of_pairwise <| hf_disj.imp fun i j (hdisj : Disjoint _ _) hfij => ?_
  rw [← hfij, disjoint_self] at hdisj
  rw [hdisj, hg_bot]
```

Mathlib/Algebra/BigOperators/Group/Finset/Basic.lean
```diff
@@ -1459,6 +1434,71 @@ theorem prod_unique_nonempty {α β : Type*} [CommMonoid β] [Unique α] (s : Fi
...
+
+/-- If the images of `f` only overlap where `g (f i) = c` , then `g (f j) = c` whenever
+`g (f j) = g (f n)` for some `n ≠ j`.-/
+lemma filter_erase_of_pairwise [CommMonoid β] {f : ι → α} {g : α → β}
+    {n : ι} {I : Finset ι} (hn : n ∈ I) (c : β)
+    (hf : (I.toSet).Pairwise fun i j => f i = f j → g (f i) = c) :
+    ∀ j ∈ (filter (fun i => f i = f n) I).erase n, g (f j) = c := by
+  intro x hx
+  simp only [mem_erase, ne_eq, mem_filter] at hx
+  exact hf hx.2.1 hn hx.1 hx.2.2
```
eric-wieser, line 1460: Presumably if you swap the order, the first can be proved in terms of the second?

pfaffelh: Actually, I tried this. However, using the second lemma to prove the first breaks the `to_additive` attribute, since `to_additive` does not work for the second lemma.

eric-wieser: I pushed some commits that do this, it worked without any issue.

Mathlib/Algebra/BigOperators/Group/Finset/Basic.lean
```diff
@@ -1434,6 +1434,50 @@ theorem prod_unique_nonempty {α β : Type*} [CommMonoid β] [Unique α] (s : Fi
...
+    (hn : n ∈ I) (hf : (I : Set ι).Pairwise fun i j ↦ f i = f j → g (f i) = 1) :
+    ∏ j ∈ filter (fun j ↦ f j = f n) I, g (f j) = g (f n) := by
+  have h j (hj : j ∈ (filter (fun i ↦ f i = f n) I).erase n) : g (f j) = 1 := by
+    simp only [mem_erase, mem_filter] at hj
+    exact hf hj.2.1 hn hj.1 hj.2.2
+  rw [← mul_one (g (f n)), ← prod_eq_one h,
+    ← mul_prod_erase (filter (f · = f n) I) (fun i ↦ g (f i)) <| mem_filter.mpr ⟨hn, by rfl⟩]
+
+/-- A version of `Finset.prod_map` and `Finset.prod_image`, but we do not assume that `f` is
+injective. Rather, we assume that the image of `f`  on `I` only overlaps where `g (f i) = 1`.
```
Ruben-VandeVelde, line 1452:
```suggestion
injective. Rather, we assume that the image of `f` on `I` only overlaps where `g (f i) = 1`.
```

Mathlib/Algebra/BigOperators/Group/Finset/Basic.lean
```diff
@@ -1434,6 +1434,50 @@ theorem prod_unique_nonempty {α β : Type*} [CommMonoid β] [Unique α] (s : Fi
...
+The conclusion is the same as in `prod_image`.-/
+@[to_additive (attr := simp)
+"A version of `Finset.sum_map` and `Finset.sum_image`, but we do not assume that `f` is
+injective. Rather, we assume that the image of `f`  on `I` only overlaps where `g (f i) = 0`.
```
Ruben-VandeVelde, line 1456:
```suggestion
injective. Rather, we assume that the image of `f` on `I` only overlaps where `g (f i) = 0`.
```

Ruben-VandeVelde (approved):
maintainer merge

jcommelin:
Thanks :tada:

bors merge

PR #15296, opened 2024-07-30
[Merged by Bors] - feat: add `OuterMeasure.ofFunction_eq_iInf_mem`

RemyDegenne:
---


Mathlib/MeasureTheory/OuterMeasure/OfFunction.lean
```diff
@@ -90,6 +90,25 @@ theorem ofFunction_apply (s : Set α) :
     OuterMeasure.ofFunction m m_empty s = ⨅ (t : ℕ → Set α) (_ : s ⊆ iUnion t), ∑' n, m (t n) :=
   rfl
 
+/-- Same as `ofFunction_apply`, except that the sets `t i` satisfy a predicate `P`,
```
sgouezel, line 93: Can you also explain in the docstring what the theorem does, without referring to the other lemma? When they appear in tooltips, it's better if the docstrings are self-contained.

sgouezel:
Can you fix the merge conflict?

RemyDegenne:
Thanks for telling me. It would be nice to get an email notification in that case. The bot could post a comment on the PR for example.

urkud:
Thanks! :tada:
bors merge

PR #15297, opened 2024-07-30
[Merged by Bors] - feat(Algebra/Category/Ring): Add dsimp lemmas for `CommRingCat`.

erdOne:
---


erdOne:
!bench

PR #15300, opened 2024-07-30
[Merged by Bors] - chore: split conj into two files to streamline imports

emilyriehl:
This splits the file CategoryTheory.Conj into CategoryTheory.HomCongr and CategoryTheory.Conj. The point is that the latter requires much heavier imports than the former, while the former is all that is needed for CategoryTheory.Adjunction.Mates, which is used elsewhere.

While making those changes, I added one theorem to the HomCongr file:

* A theorem `map_isoCongr` which parallels the existing `map_homCongr`.

The proof is by `ext; simp` once the existing `isoCongr` is tagged with simps. 

In parallel, I've tagged the definition `homCongr` with simps and removed the theorem `homCongr_apply` which is automatically generated by simps.

---

This was done with help from @jcommelin who tells me I should apologize for polluting the PR!

### DIFF
--- a/Mathlib/CategoryTheory/Conj.lean
+++ b/Mathlib/CategoryTheory/Conj.lean
@@ -5,6 +5,7 @@ Authors: Yury Kudryashov
 -/
 import Mathlib.Algebra.Group.Units.Equiv
 import Mathlib.CategoryTheory.Endomorphism
+import Mathlib.CategoryTheory.HomCongr
 
 /-!
 # Conjugate morphisms by isomorphisms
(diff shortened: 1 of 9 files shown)


emilyriehl:
@semorrison this is the PR I promised I'd create for you, following your suggestions.

Mathlib/CategoryTheory/Conj.lean
```diff
@@ -15,13 +16,12 @@ An isomorphism `α : X ≅ Y` defines
...
   `α.conjAut f = α.symm ≪≫ f ≪≫ α`.
 
-For completeness, we also define
-`CategoryTheory.Iso.homCongr : (X ≅ X₁) → (Y ≅ Y₁) → (X ⟶ Y) ≃ (X₁ ⟶ Y₁)`,
-cf. `Equiv.arrowCongr`,
-and `CategoryTheory.Iso.isoCongr : (f : X₁ ≅ X₂) → (g : Y₁ ≅ Y₂) → (X₁ ≅ Y₁) ≃ (X₂ ≅ Y₂)`.
+using
+`CategoryTheory.Iso.homCongr : (X ≅ X₁) → (Y ≅ Y₁) → (X ⟶ Y) ≃ (X₁ ⟶ Y₁)`
+and `CategoryTheory.Iso.isoCongr : (f : X₁ ≅ X₂) → (g : Y₁ ≅ Y₂) → (X₁ ≅ Y₁) ≃ (X₂ ≅ Y₂)`
+which are defined in  `CategoryTheory.HomCongr`.
```
semorrison, line 22: Something seems to have gone wrong here: sentence begins with "using"?

semorrison:
bors d+

Thanks!

Mathlib/CategoryTheory/Conj.lean
emilyriehl: It's all one sentence, starting with "An isomorphism". I was hoping this comment would be helpful for folks familiar with the old file "conj".

PR #15311, opened 2024-07-30
[Merged by Bors] - feat: Scalar multiplication of an indicator function

james18lpc:
from GibbsMeasure

Co-authored-by: Yael Dillies <yael.dillies@gmail.com>

---

### DIFF
--- a/Mathlib/Algebra/Module/Basic.lean
+++ b/Mathlib/Algebra/Module/Basic.lean
@@ -191,6 +191,16 @@ lemma indicator_smul_const (s : Set α) (r : α → R) (m : M) :
   funext <| indicator_smul_const_apply _ _ _
 
 end SMulWithZero
+
+section MulZeroOneClass
+
+variable [MulZeroOneClass R]
+
+lemma smul_indicator_one_apply (s : Set α) (r : R) (a : α) :
+    r • s.indicator (1 : α → R) a = s.indicator (fun _ ↦ r) a := by
+  simp_rw [← indicator_const_smul_apply, Pi.one_apply, smul_eq_mul, mul_one]
+
+end MulZeroOneClass
 end Set
 
 assert_not_exists Multiset


Mathlib/Algebra/Module/Basic.lean
```diff
@@ -191,6 +191,16 @@ lemma indicator_smul_const (s : Set α) (r : α → R) (m : M) :
...
 
 end SMulWithZero
+
+section MulZeroOneClass
+
+variable [MulZeroOneClass R]
+
+lemma smul_indicator_one_apply (s : Set α) (r : R) (a : α) :
+    r • s.indicator (1 : α → R) a = s.indicator (fun _ ↦ r) a := by
+  simp_rw [← indicator_const_smul_apply, Pi.one_apply, smul_eq_mul, mul_one]
```
j-loreaux, line 201: This is backwards from the way the other declarations in this file are written, so we should flip it and rename. Also, most of these have "unapplied" versions as well. Finally, if you don't know yet about the `simpa` tactic, you should learn it, as it's very useful.
```suggestion
lemma indicator_smul_one_apply (s : Set α) (r : R) (a : α) :
    s.indicator (fun _ ↦ r) a = r • s.indicator (1 : α → R) a := by
  simpa using indicator_const_smul_apply _ r (1 : α → R) _

lemma indicator_smul_one (s : Set α) (r : R) :
    s.indicator (fun _ ↦ r) =  (r • s.indicator (1 : α → R) ·) :=
  funext <| indicator_smul_one_apply _ _
```

j-loreaux (commented):
Thanks for your contribution! I see you're relatively new to Mathlib, so I've tried to provide a few pointers. Once you update the PR, please remove the `awaiting-author` tag so that people know it's ready for review again.

Mathlib/Algebra/Module/Basic.lean
YaelDillies: Hey Jireh. I have co-authored these lemmas and I think they are particularly useful this way around (+ note that your proposed lemma names are incorrect, which is something I have taken into account). The point is to "swallow" operations inside the indicator so as to later apply lemmas like `lintegral_indicator`. I am not opposed to turning the other lemmas around if you really value consistency (although personally I think inconsistency is fine here).

Regarding the unapplied version, I find these lemmas mostly useless here, but if you want to add one it would need to use the pointwise operation, namely
```suggestion
lemma smul_indicator_one_apply (s : Set α) (r : R) (a : α) :
    r • s.indicator (1 : α → R) a = s.indicator (fun _ ↦ r) a := by
  simp_rw [← indicator_const_smul_apply, Pi.one_apply, smul_eq_mul, mul_one]

lemma smul_indicator_one (s : Set α) (r : R) :
    r • s.indicator (1 : α → R) = s.indicator (fun _ ↦ r) := by
  funext <| smul_indicator_one_apply _ _
```

j-loreaux: Okay, if you have a use case in mind already, I'm fine with the current version. (And sorry, you're right I screwed up the lemma names)

j-loreaux:
bors merge

PR #14027, opened 2024-06-22
[Merged by Bors] - feat(Condensed): characterisation of discrete (light) condensed sets and modules

dagurtomas:
This PR provides a characterization of discrete condensed sets and modules, both in the light setting and the classical one. 

* Six equivalent conditions on a condensed set X to be discrete
  1. The counit of the discrete-underlying adjunction applied to X is an isomorphism
  2. X is in the essential image of the constant sheaf functor `Type (u+1) ⥤ CondensedSet.{u}`
  3. X is in the essential image of the functor `Type (u+1) ⥤ CondensedSet.{u}` which takes a set to the sheaf of locally             constant maps into it.
  4. The counit of the locally-constant-underlying adjunction applied to X is an isomorphism.
  5. X restricted to the coherent site of profinite sets is discrete as a sheaf
  6. For every profinite set S written as a limit of finite sets, X maps S to the corresponding colimit.
* A condensed module over a ring is discrete if and only if its underlying condensed set is, and therefore the analogues of the equivalent conditions above also characterize condensed modules as discrete
* The analogues for the above for light condensed sets and modules 

---

- [x] depends on: #12930 
- [x] depends on: #13333 
- [x] depends on: #13486 
- [x] depends on: #13947 
- [x] depends on: #14017 
- [x] depends on: #14382
- [x] depends on: #14406
- [x] depends on: #15321 
- [x] depends on: #15566 
- [x] depends on: #15569 
- [x] depends on: #17537 

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15321, opened 2024-07-30
[Merged by Bors] - feat(Condensed): discrete condensed sets are given by locally constant maps

dagurtomas:
This PR proves that under suitable conditions, the functor from the category of sets to the
category of sheaves for the coherent topology on `CompHausLike P`, given by mapping a set to the
sheaf of locally constant maps to it, is left adjoint to the "underlying set" functor (evaluation
at the point).

We apply this to prove that the constant sheaf functor into (light) condensed sets is isomorphic to
the functor of sheaves of locally constant maps described above.

This is the first part of the characterization of discrete condensed objects, see #14027.

---

- [x] depends on: #15401 
- [x] depends on: #15402
- [x] depends on: #15525  
- [x] depends on: #15527 


Mathlib/Condensed/Discrete/LocallyConstant.lean
```diff
@@ -0,0 +1,563 @@
+/-
+Copyright (c) 2024 Dagur Asgeirsson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dagur Asgeirsson
+-/
+import Mathlib.Data.Countable.Small
+import Mathlib.Topology.Category.CompHausLike.Limits
+import Mathlib.Topology.LocallyConstant.Basic
+import Mathlib.Condensed.Discrete
```
dagurtomas, line 9:
```suggestion
```

Mathlib/Condensed/Discrete/LocallyConstant.lean
```diff
@@ -0,0 +1,442 @@
...
+/--
+`CondensedSet.LocallyConstant.functor` is isomorphic to `Condensed.discrete`
+(by uniqueness of adjoints).
+-/
+noncomputable def CondensedSet.LocallyConstant.iso :
+    CondensedSet.LocallyConstant.functor ≅ discrete (Type (u+1)) :=
+  (LocallyConstant.adjunction _ _).leftAdjointUniq (discreteUnderlyingAdj _)
+
+/-- `CondensedSet.LocallyConstant.functor` is fully faithful. -/
+noncomputable def fullyFaithfulCondensedSetLocallyConstantFunctor :
```
jcommelin, line 390: What do you think of a name like this? Some of the other names could maybe also be shortened-modulo-namespacing in a similar way.
```suggestion
noncomputable def CondensedSet.LocallyConstant.functor_fullyFaithful :
```

dagurtomas: Yes that makes sense, but I guess it would be `CondensedSet.LocallyConstant.functorFullyFaithful` because `FullyFaithful` isn't a prop. I'll make these changes tomorrow.

Mathlib/Condensed/Discrete/LocallyConstant.lean
```diff
@@ -0,0 +1,429 @@
...
+    exact image_eq_image_mk (g := g.unop) (a := a)
+
+variable (P) (X : TopCat.{max u w})
+    [HasExplicitFiniteCoproducts.{0} P] [HasExplicitPullbacks P]
+    (hs : ∀ ⦃X Y : CompHausLike P⦄ (f : X ⟶ Y), EffectiveEpi f → Function.Surjective f)
+
+/-- `locallyConstantIsoContinuousMap` is a natural isomorphism. -/
+noncomputable def functorToPresheavesIsoTopCatToSheafCompHausLike (X : Type (max u w)) :
+    functorToPresheaves.{u, w}.obj X ≅
+      ((topCatToSheafCompHausLike P hs).obj (TopCat.discrete.obj X)).val :=
```
jcommelin, line 226: I'm going to merge this as is. But I think a follow-up PR could rename `topCatToSheafCompHausLike` to `TopCat.toSheafCompHausLike`. That would allow you to make several names a bit shorter.

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/Condensed/Discrete/LocallyConstant.lean
dagurtomas: The name `TopCat.toSheafCompHausLike` already exists as the object part of that functor (which is more useful because of dot notation). But there are indeed some places in this file where that can be used, I opened #17075

PR #15322, opened 2024-07-30
[Merged by Bors] - feat: `setLintegral` for indicator function and `le_meas`

james18lpc:
from GibbsMeasure

Co-authored-by: Yael Dillies <yael.dillies@gmail.com>

---


Mathlib/MeasureTheory/Integral/Lebesgue.lean
```diff
@@ -769,6 +769,11 @@ theorem lintegral_indicator (f : α → ℝ≥0∞) {s : Set α} (hs : Measurabl
   refine ⟨⟨φ.restrict s, fun x => ?_⟩, le_rfl⟩
   simp [hφ x, hs, indicator_le_indicator]
 
+lemma setLintegral_indicator (f : α → ℝ≥0∞) {s t : Set α} (hs : MeasurableSet s)
+    (ht : MeasurableSet t) : ∫⁻ a in t, s.indicator f a ∂μ = ∫⁻ a in s ∩ t, f a ∂μ := by
```
urkud, line 773: I think that measurability of one of the sets should be enough. E.g., you can `rw [lintegral_indicator, Measure.restrict_restrict]` (or `restrict_restrict'` - not sure). Also, it should be true if the set is only null measurable.

james18lpc: Hello Yury,

I just made changes as you suggested. Thanks!

Mathlib/MeasureTheory/Integral/Lebesgue.lean
```diff
@@ -769,12 +769,21 @@ theorem lintegral_indicator (f : α → ℝ≥0∞) {s : Set α} (hs : Measurabl
   refine ⟨⟨φ.restrict s, fun x => ?_⟩, le_rfl⟩
   simp [hφ x, hs, indicator_le_indicator]
 
+lemma setLintegral_indicator (f : α → ℝ≥0∞) {s t : Set α} (hs : MeasurableSet s) :
```
urkud, line 772: It looks like `setLIntegral` is a more popular capitalization. Could you please change it in all 3 lemmas? Otherwise LGTM. Thanks!
bors d+

james18lpc: Hello Yury,

Apologies for the delay - just made changes as you suggested. Thanks!

Mathlib/MeasureTheory/Integral/Lebesgue.lean
```diff
@@ -864,6 +873,12 @@ lemma lintegral_le_meas {s : Set α} {f : α → ℝ≥0∞} (hf : ∀ a, f a 
   · simpa [hx] using hf x
   · simpa [hx] using h'f x hx
 
+lemma setLIntegral_le_meas {s t : Set α} (hs : MeasurableSet s)
+    {f : α → ℝ≥0∞} (hf : ∀ a ∈ s, f a ≤ 1)
+    (hf' : ∀ a ∈ s \ t, f a = 0) : ∫⁻ a in s, f a ∂μ ≤ μ t := by
```
urkud, line 878: I'm sorry for the delay. One last request: please assume `∀ a ∈ s, a ∈ t → f a ≤ 1` and `∀ a ∈ s, a ∉ t → f a = 0`. This is equivalent to your assumptions, but those who apply your lemma won't need to deal with the case `a ∈ s, a ∉ t` twice. Thanks!
bors d+

urkud:
bors merge

PR #15323, opened 2024-07-30
[Merged by Bors] - chore(Topology/LocallyClosed): move around

urkud:
- move definitions to `Topology/Defs/Basic`;
- move the last lemma to `Topology/LocalAtTarget`.

---

### DIFF
--- a/Mathlib/Topology/Defs/Basic.lean
+++ b/Mathlib/Topology/Defs/Basic.lean
@@ -98,6 +98,12 @@ class IsClosed (s : Set X) : Prop where
 def IsClopen (s : Set X) : Prop :=
   IsClosed s ∧ IsOpen s
 
+/--
+A set is locally closed if it is the intersection of some open set and some closed set.
+Also see `isLocallyClosed_tfae`.
+-/
+def IsLocallyClosed (s : Set X) : Prop := ∃ (U Z : Set X), IsOpen U ∧ IsClosed Z ∧ s = U ∩ Z
+
 /-- The interior of a set `s` is the largest open subset of `s`. -/
 def interior (s : Set X) : Set X :=
   ⋃₀ { t | IsOpen t ∧ t ⊆ s }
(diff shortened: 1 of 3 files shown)


Mathlib/Topology/Defs/Basic.lean
```diff
@@ -98,6 +98,12 @@ class IsClosed (s : Set X) : Prop where
 def IsClopen (s : Set X) : Prop :=
   IsClosed s ∧ IsOpen s
 
+/--
+A set is locally closed if it is the intersection of some open set and some closed set.
+Also see `isLocallyClosed_tfae`.
```
erdOne, line 103: Can you mention the file `Mathlib/Topology/LocallyClosed.lean` where the API lemmas are? Thanks.

urkud:
```suggestion
Also see `isLocallyClosed_tfae` and other lemmas in `Mathlib/Topology/LocallyClosed`.
```

Mathlib/Topology/Defs/Basic.lean
```diff
@@ -110,6 +116,18 @@ def closure (s : Set X) : Set X :=
 def frontier (s : Set X) : Set X :=
   closure s \ interior s
 
+/--
+The coborder is defined as the complement of `closure s \ s`,
+or the union of `s` and the complement of `∂(s)`.
+This is the largest set such that `s` is closed in, and `s` is locally closed if and only if
```
grunweg, line 122: I know you just moved the comment, but: I'm stumbling over the grammar of that sentence.
Do you mean "This is the largest set `t` such that `s` is closed in `t`", or "This is the largest set in which `s` is closed"?

grunweg (commented):
Thanks for doing this! I have one pre-existing comment on a docstring.
maintainer merge

Mathlib/Topology/Defs/Basic.lean
erdOne: I added the comment in the first place. To me the the two clauses mean the same thing though. Can you please clarify?

j-loreaux: I agree with Michael. As written, the sentence doesn't parse.
```suggestion
This is the largest set in which `s` is closed, and `s` is locally closed if and only if
```

j-loreaux:
Forgive my ignorance, but is there an obvious reason the import chain should be reversed in the case of `LocallyClosed` and `LocalAtTarget`?

urkud:
I think that `LocallyClosed` is a basic predicate like `IsOpen` or `IsClosed` while `LocalAtTarget` depends on `Topology.Sets.Opens` which depends on a lot of stuff.

j-loreaux:
bors d+

urkud:
bors merge

PR #14873, opened 2024-07-18
[Merged by Bors] - feat(RingTheory/Radical): Radical of an element in a unique factorization normalization monoid

seewoo5:
For a unique factorization normalization monoid, define a radical of an element as a product of normalized prime factors (without duplication).

---

This is part of our project to port [Mason-Stothers Theorem](https://github.com/seewoo5/lean-poly-abc) in mathlib @jcpaik. You can find more theorems that will be added in future PRs [here](https://github.com/seewoo5/lean-poly-abc/blob/main/LeanPolyABC/Lib/Radical.lean).

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15331, opened 2024-07-30
[Merged by Bors] - feat(RingTheory/Radical): Theorems for coprime elements + alpha

seewoo5:
---

More theorems on element radical, especially multiplicativity of radical for coprime elements.
- [x] depends on: #14873 
- [x] depends on: #15826 
- [x] depends on: #15818 


Mathlib/RingTheory/Radical.lean
```diff
@@ -0,0 +1,155 @@
...
+
+open scoped Classical
+
+open UniqueFactorizationMonoid
+
+variable {k : Type _} [Field k]
+-- `CancelCommMonoidWithZero` is required by `UniqueFactorizationMonoid`
+variable {α : Type _} [CancelCommMonoidWithZero α] [NormalizationMonoid α]
+  [UniqueFactorizationMonoid α]
+variable {R : Type _} [CommRing R] [IsDomain R] [NormalizationMonoid R] [UniqueFactorizationMonoid R]
```
erdOne, line 44: Rather than declaring these at the front, I'd prefer if the contents are split into sections so that users will not have to distinguish the typeclass assumptions by looking at the type name and guessing. 
Also please use `Type*` instead of `Type _`.

Mathlib/RingTheory/Radical.lean
```diff
@@ -0,0 +1,155 @@
...
+theorem radical_eq_of_associated {a b : α} (h : Associated a b) : radical a = radical b := by
+  rcases iff_iff_and_or_not_and_not.mp h.eq_zero_iff with (⟨rfl, rfl⟩ | ⟨ha, hb⟩)
+  · rfl
+  · simp_rw [radical, primeFactors]
+    rw [(associated_iff_normalizedFactors_eq_normalizedFactors ha hb).mp h]
+
+theorem radical_unit_eq_one {a : α} (h : IsUnit a) : radical a = 1 :=
+  (radical_eq_of_associated (associated_one_iff_isUnit.mpr h)).trans radical_one_eq
+
+theorem radical_unit_hMul {u : αˣ} {a : α} : radical ((↑u : α) * a) = radical a :=
```
erdOne, line 74:
```suggestion
theorem radical_unit_mul {u : αˣ} {a : α} : radical ((↑u : α) * a) = radical a :=
```
This is not `mul` as both operands are of the same type.

Mathlib/RingTheory/Radical.lean
```diff
@@ -0,0 +1,155 @@
...
+theorem radical_dvd_self (a : α) : radical a ∣ a := by
+  by_cases ha : a = 0
+  · rw [ha]
+    apply dvd_zero
+  · rw [radical, ← Finset.prod_val, ← (normalizedFactors_prod ha).dvd_iff_dvd_right]
+    apply Multiset.prod_dvd_prod_of_le
+    rw [primeFactors, Multiset.toFinset_val]
+    apply Multiset.dedup_le
+
+theorem radical_prime {a : α} (ha : Prime a) : radical a = normalize a := by
```
erdOne, line 95:
```suggestion
theorem radical_of_prime {a : α} (ha : Prime a) : radical a = normalize a := by
```

Mathlib/RingTheory/Radical.lean
```diff
@@ -0,0 +1,155 @@
...
+  rw [radical, primeFactors]
+  rw [normalizedFactors_irreducible ha.irreducible]
+  simp only [Multiset.toFinset_singleton, id, Finset.prod_singleton]
+
+theorem radical_prime_pow {a : α} (ha : Prime a) {n : ℕ} (hn : 1 ≤ n) :
```
erdOne, line 100:
```suggestion
theorem radical_pow_of_prime {a : α} (ha : Prime a) {n : ℕ} (hn : 1 ≤ n) :
```

Mathlib/RingTheory/Radical.lean
```diff
@@ -0,0 +1,155 @@
...
+private theorem IsCoprime.disjoint_primeFactors {a b : R} (hc : IsCoprime a b) :
+    Disjoint (primeFactors a) (primeFactors b) :=
+  Multiset.disjoint_toFinset.mpr (hc.disjoint_normalizedFactors)
+
+private theorem IsCoprime.hMul_primeFactors_disjUnion {a b : R} (ha : a ≠ 0) (hb : b ≠ 0)
+    (hc : IsCoprime a b) :
+    primeFactors (a * b) = (primeFactors a).disjUnion (primeFactors b) hc.disjoint_primeFactors := by
+  rw [Finset.disjUnion_eq_union]
+  simp_rw [primeFactors]
+  rw [normalizedFactors_mul ha hb, Multiset.toFinset_add]
```
erdOne, line 127: Why are these private?

seewoo5:
@erdOne Thank you for the reviews! It seems that most of them are for #14873. Could you leave the comments there? Then I'll updated in that PR accordingly.

Mathlib/RingTheory/Radical.lean
```diff
@@ -104,4 +106,57 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
   rw [radical_pow a hn]
   exact radical_of_prime ha
 
+
+-- Theorems for commutative rings
+
+-- TODO: This holds for "nontrivial" monoids - do not need ring assumption.
+theorem radical_ne_zero (a : R) : radical a ≠ 0 := by
```
seewoo5, line 113: As in comment, we don't need ring structure for this; the theorem holds for "nontrivial" monoid. Should I update it by adding the nontriviality assumption, or define another variable $M_1$ for nontrivial monoid, or just leave it as-is?

Mathlib/RingTheory/Radical.lean
```diff
@@ -104,4 +106,57 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
...
+  rw [Finset.disjUnion_eq_union]
+  simp_rw [primeFactors]
+  rw [normalizedFactors_mul ha hb, Multiset.toFinset_add]
+
+@[simp]
+theorem radical_neg_one : radical (-1 : R) = 1 :=
+  radical_unit_eq_one isUnit_one.neg
+
+/-- Radical is multiplicative for coprime elements. -/
+theorem radical_hMul {a b : R} (hc : IsCoprime a b) :
```
Ruben-VandeVelde, line 147: Is there a reason to use `hMul` rather than `mul` in this name?

seewoo5: Actually not, I think `mul` is better in this case. I just updated.

Ruben-VandeVelde (approved):

Mathlib/RingTheory/Radical.lean
```diff
@@ -104,4 +106,57 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
...
+    simp only [MulZeroClass.zero_mul, radical_zero_eq, one_mul, radical_unit_eq_one hc]
+  by_cases hb : b = 0
+  · subst hb; rw [isCoprime_zero_right] at hc
+    simp only [MulZeroClass.mul_zero, radical_zero_eq, mul_one, radical_unit_eq_one hc]
+  simp_rw [radical]
+  rw [mul_primeFactors_disjUnion ha hb hc]
+  rw [Finset.prod_disjUnion (disjoint_primeFactors hc)]
+
+theorem radical_neg {a : R} : radical (-a) = radical a :=
+  neg_one_mul a ▸ (radical_eq_of_associated <| associated_unit_mul_left a (-1) isUnit_one.neg)
```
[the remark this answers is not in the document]
seewoo5: I'm looking at [Algebra.Associated.Basic](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Associated/Basic.html) now, and I can find `neg_left` or `neg_right` which can be used combined with `rfl`, but not exactly something like `Associated (-a) a`. Should I look for different file?

Mathlib/RingTheory/Radical.lean
seewoo5: Updated, wonder if this fits into conventions. For example, is there any convention for the order of arguments, like `(a : M) (h: Nontrivial M)` vs `(h: Nontrivial M) (a : M)`?

seewoo5:
@Ruben-VandeVelde Hi, let me know if there are anything that I may update, thanks!

Mathlib/RingTheory/Radical.lean
```diff
@@ -104,4 +106,55 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
...
+
+/-- Coprime elements have disjoint prime factors (as finsets). -/
+theorem disjoint_primeFactors {a b : R} (hc : IsCoprime a b) :
+    Disjoint (primeFactors a) (primeFactors b) :=
+  Multiset.disjoint_toFinset.mpr (disjoint_normalizedFactors hc)
+
+theorem mul_primeFactors_disjUnion {a b : R} (ha : a ≠ 0) (hb : b ≠ 0)
+    (hc : IsCoprime a b) :
+    primeFactors (a * b) = (primeFactors a).disjUnion (primeFactors b) (disjoint_primeFactors hc)
+    := by
```
erdOne, line 135: There is a lint failure here. Please put `:=` before line-breaks. If the previous line is too long, split it somewhere (e.g. after `=`)

Mathlib/RingTheory/Radical.lean
```diff
@@ -41,6 +41,8 @@ namespace UniqueFactorizationMonoid
 -- `CancelCommMonoidWithZero` is required by `UniqueFactorizationMonoid`
 variable {M : Type*} [CancelCommMonoidWithZero M] [NormalizationMonoid M]
   [UniqueFactorizationMonoid M]
+variable {R : Type*} [CommRing R] [IsDomain R] [NormalizationMonoid R]
+  [UniqueFactorizationMonoid R]
```
erdOne, line 45: Please put lemmas about UFDs into a separate namespace.

Mathlib/RingTheory/Radical.lean
```diff
@@ -104,4 +106,55 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
   rw [radical_pow a hn]
   exact radical_of_prime ha
 
+theorem radical_ne_zero (a : M) (h: Nontrivial M) : radical a ≠ 0 := by
```
erdOne, line 109:
```suggestion
theorem radical_ne_zero (a : M) [Nontrivial M] : radical a ≠ 0 := by
```

Mathlib/RingTheory/Radical.lean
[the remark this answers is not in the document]
erdOne: `radical_eq_of_associated (.neg_right .rfl)` (or some thing similar; not tested) is still shorter than this though.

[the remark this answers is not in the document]
seewoo5: @erdOne Just updated - `radical_eq_of_associated Associated.rfl.neg_left` worked!

Mathlib/RingTheory/Radical.lean
```diff
@@ -108,4 +108,63 @@ theorem radical_pow_of_prime {a : M} (ha : Prime a) {n : ℕ} (hn : 0 < n) :
...
+  · subst ha; rw [isCoprime_zero_left] at hc
+    simp only [zero_mul, radical_zero_eq, one_mul, radical_unit_eq_one hc]
+  by_cases hb : b = 0
+  · subst hb; rw [isCoprime_zero_right] at hc
+    simp only [mul_zero, radical_zero_eq, mul_one, radical_unit_eq_one hc]
+  simp_rw [radical]
+  rw [mul_primeFactors_disjUnion ha hb hc]
+  rw [Finset.prod_disjUnion (disjoint_primeFactors hc)]
+
+theorem radical_neg {a : R} : radical (-a) = radical a :=
```
jcommelin, line 167:
```suggestion
@[simp]
theorem radical_neg {a : R} : radical (-a) = radical a :=
```

jcommelin (commented):
bors d+

PR #15335, opened 2024-07-30
[Merged by Bors] - chore(Tactic/CategoryTheory): fix meta code for monoidal categories

yuma-mizuno:
Several bug fixes and simple refactoring. The main changes:

- `Monoidal.lean`: In `structural?`, the number of arguments in `getAppFnArgs` matching was wrong (I only considered explicit ones there). Infact, `structural?` is no longer needed since the matching is performed in `eval`. Now the `monoidalCoherence` is `StructuralAtom`, not `Structural`.
- `Coherence.lean`: Unfolding `MonoidalCoherence.hom` by `dsimp` at the beginning of the coherence tactic.

---


Mathlib/Tactic/CategoryTheory/BicategoryCoherence.lean
```diff
@@ -99,21 +99,21 @@ def exception' (msg : MessageData) : TacticM Unit := do
...
 /-- Auxiliary definition for `bicategorical_coherence`. -/
--- We could construct this expression directly without using `elabTerm`,
--- but it would require preparing many implicit arguments by hand.
-def mkLiftMap₂LiftExpr (e : Expr) : TermElabM Expr := do
-  Term.elabTerm
-    (← ``((FreeBicategory.lift (Prefunctor.id _)).map₂ (LiftHom₂.lift $(← Term.exprToSyntax e))))
-    none
+def mkLiftMap₂LiftExpr (e : Expr) : MetaM Expr := do
+  let f ← mkAppOptM ``LiftHom₂.lift #[none, none, none, none, none, none, none, none, e, none]
+  mkAppM ``mkLiftMap₂LiftExprAux #[f]
```
YaelDillies, line 109: I would strongly suggest using Qq to avoid the kind of errors you have spotted and make sure they don't reappear

semorrison: Pushing in a different direction, I generally advise to avoid using `Qq` unless it is for functionality that is really missing elsewhere. 🤷‍♀️

YaelDillies: Not using Qq already caused bugs here though

yuma-mizuno: From another perspective, the bug should have been caught by tests. This PR improves the tests as well.

eric-wieser: One argument for `Qq` here is that ```` ``LiftHom₂.lift #[none, none, none, none, none, none, none, none, e, none] ``` is super fragile, and fails only in the test rather than in the elaborator if the argument order changes.

yuma-mizuno: I tried to use `Qq` this part, but I was not completely successful https://github.com/leanprover-community/mathlib4/pull/15335/commits/d886e88ba98dc4b05cc46e426f8dd7bfd1b277f0.

Instead, I would like to suggest using `MkAppM` here to avoid the `none` sequence, which I believe reasonable solutions for this situation.

yuma-mizuno: It seems that the current `Qq` doesn't work well with matching universe variables needed here. See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2315335.20meta.20code.20for.20monoidal.20categories).

Actually, this part is independent of the other fixes, so I would like to separate it into a future PR to make it easier to review.

YaelDillies: Feel free to make it a separate PR yes

Mathlib/Tactic/CategoryTheory/BicategoryCoherence.lean
```diff
@@ -99,6 +99,14 @@ def exception' (msg : MessageData) : TacticM Unit := do
     -- There might not be any goals
     throwError msg
 
+/-- Sending 2-morphisms in `FreeBicategory B` to those in `B`. -/
+def mkLiftMap₂LiftExprAux {a b : FreeBicategory B} {f g : a ⟶ b} (η : f ⟶ g) :=
+  (FreeBicategory.lift (Prefunctor.id _)).map₂ η
+
+/-- Same as `LiftHom₂.lift`, but the `LiftHom₂ η` instance is explicit. -/
+abbrev LiftHom₂.lift' {f g : a ⟶ b} [LiftHom f] [LiftHom g] (η : f ⟶ g) (inst : LiftHom₂ η) :=
+  inst.lift
```
YaelDillies, line 108: Why is this in the diff?

yuma-mizuno: It was an oversight in reverting, thanks.

Mathlib/Tactic/CategoryTheory/BicategoryCoherence.lean
YaelDillies: Will you split the PR?

yuma-mizuno: I have made #16570.

YaelDillies: Are you reverting stuff from this PR accordingly?

yuma-mizuno: Yes, I separated the codes related to `Qq` discussion into #16570.

semorrison:
bors merge

PR #15339, opened 2024-07-30
[Merged by Bors] - chore (Topology.Algebra.Algebra): name `DiscreteTopology.instContinuousSMul` and de-instance

mattrobball:
This instance is reasonable mathematically but creates complexity in the typeclass search. It is currently not used so we de-instance it. 

---


mattrobball:
!bench

mattrobball:
!bench

urkud:
`DiscreteTopology` is not one of commonly used scopes, so I suggest that we just turn it into a `lemma`.

j-loreaux:
bors d+

mattrobball:
bors merge

PR #15342, opened 2024-07-30
[Merged by Bors] - feat(ENat): not_lt_zero, coe_lt_top

nomeata:
from the Carleson project

### DIFF
--- a/Mathlib/Data/ENat/Basic.lean
+++ b/Mathlib/Data/ENat/Basic.lean
@@ -248,6 +248,14 @@ theorem le_of_lt_add_one (h : m < n + 1) : m ≤ n :=
 theorem le_coe_iff {n : ℕ∞} {k : ℕ} : n ≤ ↑k ↔ ∃ (n₀ : ℕ), n = n₀ ∧ n₀ ≤ k :=
   WithTop.le_coe_iff
 
+@[simp]
+lemma not_lt_zero (n : ℕ∞) : ¬ n < 0 := by
+  cases n <;> simp
+
+@[simp]
+lemma coe_lt_top (n : ℕ) : (n : ℕ∞) < ⊤ :=
+  WithTop.coe_lt_top n
+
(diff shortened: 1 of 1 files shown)


Mathlib/Data/ENat/Basic.lean
```diff
@@ -248,6 +248,14 @@ theorem le_of_lt_add_one (h : m < n + 1) : m ≤ n :=
 theorem le_coe_iff {n : ℕ∞} {k : ℕ} : n ≤ ↑k ↔ ∃ (n₀ : ℕ), n = n₀ ∧ n₀ ≤ k :=
   WithTop.le_coe_iff
 
+@[simp]
+lemma not_lt_zero (n : ℕ∞) : ¬ n < 0 := by
```
b-mehta, line 252: Hmm, this looks like it should be written as `0 ≤ n`, then `simp` can prove `not_lt_zero` since it already uses `not_lt`. What do you think?

b-mehta (approved):

Mathlib/Data/ENat/Basic.lean
nomeata: Thanks for the suggestion, will try in my development

nomeata: No, that doesn't seem to work, a `simpa` proof that used to work fails with
```
type mismatch
  h✝
has type
  height x = 0 ↔ ∀ y < x, height y < 0 : Prop
but is expected to have type
  height x = 0 ↔ ∀ (y : α), ¬y < x : Prop
```
if I change this to
```
@[simp] lemma ENat.zero_le (n : ℕ∞) : 0 ≤ n := by
  cases n <;> simp
```

Ruben-VandeVelde (approved):
Seems fine then

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15346, opened 2024-07-30
[Merged by Bors] - feat(RingTheory/LinearDisjoint): properties of linearly disjoint of subalgebras (2)

acmepjz:
Some new results:

- `Subalgebra.LinearDisjoint.rank_sup_of_free`,
  `Subalgebra.LinearDisjoint.finrank_sup_of_free`:
  if subalgebras `A` and `B` are linearly disjoint and they are
  free modules, then the rank of `A ⊔ B` is equal to the product of the rank of `A` and `B`.

- `Subalgebra.LinearDisjoint.of_finrank_sup_of_free`:
  conversely, if `A` and `B` are subalgebras which are free modules of finite rank,
  such that rank of `A ⊔ B` is equal to the product of the rank of `A` and `B`,
  then `A` and `B` are linearly disjoint.

- `Subalgebra.LinearDisjoint.adjoin_rank_eq_rank_left`:
  `Subalgebra.LinearDisjoint.adjoin_rank_eq_rank_right`:
  if `A` and `B` are linearly disjoint, if `A` is free and `B` is flat (resp. `B` is free and
  `A` is flat), then `[B[A] : B] = [A : R]` (resp. `[A[B] : A] = [B : R]`).
  See also `Subalgebra.adjoin_rank_le`.

- `Subalgebra.LinearDisjoint.of_finrank_coprime_of_free`:
  if the rank of `A` and `B` are coprime, and they satisfy some freeness condition,
  then `A` and `B` are linearly disjoint.

---

- [x] depends on: #15360
- [x] depends on: #17820

discussion: <https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Linearly.20disjoint>


Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,603 @@
...
+
+variable [CommSemiring R] [Semiring S] [Algebra R S]
+
+variable (A B : Subalgebra R S)
+
+/-- If `A` and `B` are subalgebras of `S / R`,
+then `A` and `B` are linearly disjoint, if they are linearly disjoint as submodules of `S`. -/
+@[mk_iff]
+protected structure LinearDisjoint : Prop where
+  submodule : (toSubmodule A).LinearDisjoint (toSubmodule B)
```
erdOne, line 129: Any reason why this is a one-field structure and not an abbrev?

I'd suppose a lot of these lemmas will be unnecessary if it is an abbrev.

Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,603 @@
...
+    (H : LinearIndependent R fun (i : κ × ι) ↦ (a i.1).1 * (b i.2).1) : A.LinearDisjoint B :=
+  ⟨.of_basis_mul _ _ a b H⟩
+
+variable {A B}
+
+variable (H : A.LinearDisjoint B)
+
+theorem of_le_left_of_flat {A' : Subalgebra R S}
+    (h : A' ≤ A) [Module.Flat R B] : A'.LinearDisjoint B := by
+  haveI : Module.Flat R (toSubmodule B) := ‹_›
```
erdOne, line 340: Can you make this into an instance? (`instance [Module.Flat R A] : Module.Flat R A.toSubmodule`)

acmepjz: Do you mean make it global for mathlib (which I think will have a performance impact), or only locally in this file?

erdOne: I don't think there will be a performance impact.

acmepjz: PR created as #15360.

acmepjz: Done.

Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,604 @@
...
+import Mathlib.LinearAlgebra.Dimension.FreeAndStrongRankCondition
+import Mathlib.LinearAlgebra.FreeModule.StrongRankCondition
+import Mathlib.Algebra.Algebra.Subalgebra.MulOpposite
+import Mathlib.Algebra.Algebra.Subalgebra.Rank
+import Mathlib.RingTheory.IntegralClosure.Algebra.Defs
+import Mathlib.RingTheory.IntegralClosure.IsIntegral.Basic
+
+/-!
+
+# Linearly disjoint of subalgebras
```
kbuzzard, line 17:
```suggestion
# Linearly disjoint subalgebras
```

Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,604 @@
...
+
+This file contains basics about the linearly disjoint of subalgebras.
```
kbuzzard, line 19:
```suggestion
This file contains basics about linearly disjoint subalgebras.
```

Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,604 @@
...
+## Main results
+
+### Equivalent characterization of linearly disjointness
+
+- `Subalgebra.LinearDisjoint.linearIndependent_left_op_of_flat`:
+  `Subalgebra.LinearDisjoint.linearIndependent_left_of_flat_of_commute`:
+  `Subalgebra.LinearDisjoint.linearIndependent_left_of_flat`:
+  if `A` and `B` are linearly disjoint, if `B` is a flat `R`-module, then for any family of
+  `R`-linearly independent elements of `A`, they are also `B`-linearly independent
+  (in the opposite ring for the first one).
```
kbuzzard, line 38: Why not drop the `op` and `of_commute` lemmas here -- this is Main results, not necessarily all results. Then you can just explain one lemma clearly instead of adding brackets to try and explain three lemmas in one sentence (and similarly for the next three).

Mathlib/RingTheory/LinearDisjoint.lean
```diff
@@ -0,0 +1,604 @@
...
+  See also `Subalgebra.adjoin_rank_le`.
+
+- `Subalgebra.LinearDisjoint.of_finrank_coprime_of_free`:
+  if the rank of `A` and `B` are coprime, and they satisfy some freeness condition,
+  then `A` and `B` are linearly disjoint.
+
+- `Subalgebra.LinearDisjoint.inf_eq_bot_of_commute`, `Subalgebra.LinearDisjoint.inf_eq_bot`:
+  if `A` and `B` are linearly disjoint, under suitable technical conditions, they are disjoint.
+
+The results with name containing "of_commute" also have corresponding specified versions
```
kbuzzard, line 99:
```suggestion
The results with name containing "of_commute" also have corresponding specialized versions
```

Mathlib/RingTheory/LinearDisjoint.lean
kbuzzard: So why not try making it an `abbrev`? Maybe it will reduce a bunch of code duplication? It's hard to get a PR merged if it's very long and relies on definitions which need some discussion.

kbuzzard (commented):
This PR is very long. Are you sure that you want a one field structure for your definition of `LinearDisjoint`?

Mathlib/RingTheory/LinearDisjoint.lean
acmepjz: OK let me think and try it again.

Mathlib/RingTheory/LinearDisjoint.lean
acmepjz: OK.

Mathlib/RingTheory/LinearDisjoint.lean
acmepjz: Seems that it's working; I can't use dot notation in a few places, see commit https://github.com/leanprover-community/mathlib4/pull/15346/commits/0e6961518ae56ef52f77fd8cab40953d6e1cc380

acmepjz:
> This PR is very long.

Fine. I'll try to split this as two separate PRs. Note that the first one can only contain very basic results which can almost do nothing.

acmepjz:
OK I've split half of them to #17820. It's 450 lines but I think it can't be shorten. Please review it first.

erdOne:
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15350, opened 2024-07-31
[Merged by Bors] - perf (Site.DenseSubsite): speed up file 

mattrobball:
`rintro` and `obtain` are misbehaving here in a way `rcases` is not. We cut the instructions by ~40% by prefacing some `rintro` calls with a `dsimp` and by replacing `obtain`s with `have; rcases`.

---


mattrobball:
!bench

grunweg (approved):
Nice find! Are there any other files with the same issue/is this indicate of a bug in rintro/rcases?

mattrobball:
For 5 minutes, I stared at the implementations of `obtain` to try to figure why `have h; rcases h with` is much faster here and couldn't immediately find it. 

I think prefacing `rintro` with `dsimp` might be a good idea generally but there is probably something more surgical that can be done internal to the implementation.

j-loreaux:
Can you please add some comments (preferably in a way that we can track as technical debt, but I'm not sure what that is) near at least one `dsimp; rintro` and one `have; rcases` explaining the change and the performance change?

bors d+

mattrobball:
bors merge

PR #15353, opened 2024-07-31
[Merged by Bors] - chore(MeasureTheory/Prod): minor golf, add a docstring

urkud:
---


Mathlib/MeasureTheory/Constructions/Prod/Basic.lean
```diff
@@ -779,22 +779,29 @@ namespace MeasurePreserving
...
 
+/-- Let `f : α → β` be a measure preserving map.
+For a.e. all `a`, let `g a : γ → δ` be a measure preserving map.
+Also suppose that `g` is measurable as a function of two arguments.
+Then the map `fun (a, c) ↦ (f a, g a c)` is a measure preserving map
+for the product measures on `α × γ` and `β × δ`.
+
+Some authors call a map of the form `fun (a, c) ↦ (f a, g a c)` a *skew product* over `f`,
+thus the choice of a name.
+-/
```
grunweg, line 790: You're using `a` here, and the theorem statement uses `x`. I like using `a` (or using `x` and renaming \alpha to `X`)... can you rename the variable in the hypothesis accordingly, please?

grunweg (commented):
Thanks, looks good. I have one comment (and am not a measure theorist, so cannot comment on whether the term *skew product* actually exists) - after that, I'd maintainer merge this.

urkud:
> Thanks, looks good. I have one comment (and am not a measure theorist, so cannot comment on whether the term _skew product_ actually exists) - after that, I'd maintainer merge this.

I know that my PhD advisor calls them "skew products". However, he is Russian, and sometimes he uses translations of Russian terms instead of existing English terms.

Mathlib/MeasureTheory/Constructions/Prod/Basic.lean
urkud: Done.

grunweg:
Thanks!
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15360, opened 2024-07-31
[Merged by Bors] - feat(RingTheory/Flat/Basic): if a subalgebra is flat, then it's also flat as a submodule

acmepjz:
---

This is suggested in https://github.com/leanprover-community/mathlib4/pull/15346#discussion_r1698051003.

I don't know if it is `easy` since it adds an instance, which may have performance impact on the whole mathlib.


jcommelin (commented):
Thanks :tada:

bors merge

PR #15364, opened 2024-07-31
[Merged by Bors] - feat: gcongr works with eta and proofs

fpvandoorn:
* `gcongr` now recognizes eta-expanded variables as variables
* `gcongr` now recognizes all proofs as "defeq" even if the proofs have different types (in e.g. `Finset.sup'_mono`
* Give a bit more information in the error message when not accepting a `gcongr`-lemma.

---

### DIFF
--- a/test/GCongr/inequalities.lean
+++ b/test/GCongr/inequalities.lean
@@ -5,6 +5,7 @@ Authors: Heather Macbeth
 -/
 import Mathlib.Algebra.Order.BigOperators.Ring.Finset
 import Mathlib.Algebra.Order.Field.Basic
+import Mathlib.Data.Finset.Lattice
 import Mathlib.Tactic.Linarith
 import Mathlib.Tactic.GCongr
 import Mathlib.Tactic.SuccessIfFailWithMsg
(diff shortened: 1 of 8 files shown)


test/GCongr/inequalities.lean
```diff
@@ -169,6 +170,12 @@ example {a b c d e : ℝ} (_h1 : 0 ≤ b) (_h2 : 0 ≤ c) (hac : a * b + 1 ≤ c
   guard_target =ₛ a * b ≤ c * d
   linarith
 
+-- test big operators
+example (f g : ℕ → ℕ) (s : Finset ℕ) (h : ∀ i ∈ s, f i ≤ g i) :
```
grunweg, line 174: Do f, g and s need to be explicit? (I am aware the same question applies to the next example also.)

grunweg (commented):
It's great to see tactic improvements, thanks for doing this!
I have one tiny question about a test. And can you add a descriptive PR title, please? Thanks!

test/GCongr/inequalities.lean
fpvandoorn: No, that is irrelevant

grunweg: :+1: In this case: do you mind switching them to implicit? (If this were a mathlib file, this is what I'd ask a PR author to do.)

j-loreaux: I don't think this really matters for test files. This is not something we're careful about at all (see for instance, the rest of the this file).

j-loreaux:
This looks simple enough for even me to understand.

maintainer merge

test/GCongr/inequalities.lean
grunweg: Sure, I'm happy to adjust to that policy as well.

dupuisf:
Looks good!

bors r+

PR #15368, opened 2024-07-31
[Merged by Bors] - chore: backport some changes for lean v4.11.0

Ruben-VandeVelde:
---


Mathlib/CategoryTheory/IsomorphismClasses.lean
```diff
@@ -42,14 +42,14 @@ def isomorphismClasses : Cat.{v, u} ⥤ Type u where
   map {C D} F := Quot.map F.obj fun X Y ⟨f⟩ => ⟨F.mapIso f⟩
   map_id {C} := by  -- Porting note: this used to be `tidy`
     dsimp; apply funext; intro x
-    apply x.recOn  -- Porting note: `induction x` not working yet
+    apply @Quot.recOn _ _ _ x
```
kmill, line 45: I think the `intro x; apply` sequence should just be `rintro ⟨x⟩`, here and at 52.

(`refine x.recOn ?_ ?_` probably would work too, but `rintro` is cleaner, if it works.)

Ruben-VandeVelde: Neither of those work.

semorrison:
bors merge

PR #15370, opened 2024-07-31
[Merged by Bors] - feat(Subgroup/IsCommutative): add instance for MonoidHom.range

yu-yama:
This PR adds:

- `instance Subgroup.commGroup_isCommutative [CommGroup G] (H : Subgroup G) : H.IsCommutative` for subgroups of an abelian group; and
- `instance MonoidHom.range_isCommutative [CommGroup G] (f : G →* N) : f.range.IsCommutative`, for the range of a `MonoidHom` whose domain is abelian.

It also adds a lemma `Subgroup.mul_comm_of_mem_isCommutative` to allow applying the commutativity without constructing terms of a `Subgroup`. (This lemma is included in this same PR as I mentioned it in the same Zulip thread, but is independent of the two instances above.)

Please refer to the [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/CommGroup.20MonoidHom.2Erange.20if.20the.20domain.20is.20a.20CommGroup) for discussions.

---

### DIFF
--- a/Mathlib/Algebra/Group/Subgroup/Basic.lean
+++ b/Mathlib/Algebra/Group/Subgroup/Basic.lean
@@ -1730,6 +1730,11 @@ attribute [class] AddSubgroup.IsCommutative
 instance IsCommutative.commGroup [h : H.IsCommutative] : CommGroup H :=
   { H.toGroup with mul_comm := h.is_comm.comm }
 
+/-- A subgroup of a commutative group is commutative. -/
+@[to_additive "A subgroup of a commutative group is commutative."]
+instance commGroup_isCommutative {G : Type*} [CommGroup G] (H : Subgroup G) : H.IsCommutative :=
+  ⟨CommMagma.to_isCommutative⟩
+
 @[to_additive]
 instance map_isCommutative (f : G →* G') [H.IsCommutative] : (H.map f).IsCommutative :=
   ⟨⟨by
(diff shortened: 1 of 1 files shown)


Mathlib/Algebra/Group/Subgroup/Basic.lean
```diff
@@ -1751,6 +1756,13 @@ theorem comap_injective_isCommutative {f : G' →* G} (hf : Injective f) [H.IsCo
 instance subgroupOf_isCommutative [H.IsCommutative] : (H.subgroupOf K).IsCommutative :=
   H.comap_injective_isCommutative Subtype.coe_injective
 
+@[to_additive]
+lemma mul_comm_of_mem_isCommutative [H.IsCommutative] {a b : G} (ha : a ∈ H) (hb : b ∈ H) :
+    a * b = b * a := by
+  have := mul_comm (⟨a, ha⟩ : H) (⟨b, hb⟩ : H)
+  simp only [Submonoid.mk_mul_mk, Subtype.mk.injEq] at this
+  exact this
```
kbuzzard, line 1764: simpa only [Submonoid.mk_mul_mk, Subtype.mk.injEq] using mul_comm (⟨n₁, h₁⟩ : f.range) ⟨n₂, h₂⟩

kbuzzard, line 1764:
```suggestion
  have := mul_comm (⟨a, ha⟩ : H) (⟨b, hb⟩ : H)
  simpa only [Submonoid.mk_mul_mk, Subtype.mk.injEq] using this
```
Does that work? Maybe you can even do `simpa only ... using mul_comm ...`?

kbuzzard (approved):
LGTM!

Mathlib/Algebra/Group/Subgroup/Basic.lean
yu-yama: Thank you for the suggestion.

```suggestion
  simpa only [Submonoid.mk_mul_mk, Subtype.mk.injEq] using mul_comm (⟨a, ha⟩ : H) (⟨b, hb⟩ : H)
```

Yes, it does work (at least on my local environment).

kbuzzard:
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15371, opened 2024-07-31
[Merged by Bors] - Chore: Remove some uses of `open Classical`

sgouezel:
It is dangerous to `open Classical`, as it can hide that some theorem statements would be better stated with explicit decidability assumptions. This PR removes a few of them, using instead `classical` in proofs and `open Classical in ...` when needed for the statement.

---


j-loreaux:
Thanks! This addresses some untracked technical debt!

bors merge

PR #15377, opened 2024-07-31
[Merged by Bors] - feat: define `rpow` and `sqrt` based on the continuous functional calculus

dupuisf:
This PR defines `CFC.nnrpow`, `CFC.sqrt` and `CFC.rpow` based on the continuous functional calculus. We define two separate versions `CFC.nnrpow` (based on the non-unital calculus) and `CFC.rpow` (unital) due to what happens at 0. Since `NNReal.rpow 0 0 = 1`, this means that this function does not map zero to zero when the exponent
is zero, and hence `CFC.nnrpow a 0 = 0` whereas `CFC.rpow a 0 = 1`.

---


Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+    exact ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const <| Or.inl <| by aesop
+  rw [← cfc_mul _ _ a (h x) (h y)]
+  refine cfc_congr ?_
+  intro z hz
+  have : z ≠ 0 := by aesop
+  simp [NNReal.rpow_add this _ _]
+
+-- TODO: relate to a strict positivity condition
+lemma rpow_rpow_of_spectrum_pos [UniqueContinuousFunctionalCalculus ℝ≥0 A]
+    {a : A} {x y : ℝ} (ha₁ : ∀ z ∈ spectrum ℝ≥0 a, 0 < z) (hx : x ≠ 0) (ha₂ : 0 ≤ a := by cfc_tac) :
```
j-loreaux, line 223: Since you already have the `0 ≤ a` assumption, you can get away with:
```suggestion
    {a : A} {x y : ℝ} (ha₁ : 0 ∉ spectrum ℝ≥0 a) (hx : x ≠ 0) (ha₂ : 0 ≤ a := by cfc_tac) :
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+
+@[simp]
+lemma sqrt_algebraMap {r : ℝ≥0} : sqrt (algebraMap ℝ≥0 A r) = algebraMap ℝ≥0 A (NNReal.sqrt r) := by
+  rw [sqrt_eq_cfc, cfc_algebraMap]
+
+@[simp]
+lemma sqrt_one : sqrt (1 : A) = 1 := by simp [sqrt_eq_cfc]
+
+-- TODO: relate to a strict positivity condition
+lemma sqrt_rpow_of_spectrum_pos {a : A} {x : ℝ} (h : ∀ z ∈ spectrum ℝ≥0 a, 0 < z)
```
j-loreaux, line 299: I didn't say this before, but could have: since this is the `ℝ≥0`-spectrum, not including zero is equivalent to being positive.
```suggestion
lemma sqrt_rpow_of_spectrum_pos {a : A} {x : ℝ} (h : 0 ∉ spectrum ℝ≥0 a)
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+    (hx : x ≠ 0) : sqrt (a ^ x) = a ^ (x / 2) := by
+  by_cases hnonneg : 0 ≤ a
+  case pos =>
+    simp only [sqrt_eq_rpow, div_eq_mul_inv, one_mul, rpow_rpow_of_spectrum_pos h hx]
+  case neg =>
+    simp [sqrt_eq_cfc, rpow_def, cfc_apply_of_not_predicate a hnonneg]
+
+-- TODO: relate to a strict positivity condition
+lemma rpow_sqrt_of_spectrum_pos {a : A} {x : ℝ} (h : ∀ z ∈ spectrum ℝ≥0 a, 0 < z)
```
j-loreaux, line 308:
```suggestion
lemma rpow_sqrt_of_spectrum_pos {a : A} {x : ℝ} (h : 0 ∉ spectrum ℝ≥0 a)
```

j-loreaux (commented):
This isn't a thorough look, but I wanted to have a first pass.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+## Implementation notes
+
+We define two separate versions `CFC.rpowₙ` and `CFC.rpow` due to what happens at 0. Since
+`NNReal.rpow 0 0 = 1`, this means that this function does not map zero to zero when the exponent
+is zero, and hence `CFC.rpowₙ a 0 = 0` whereas `CFC.rpow a 0 = 1`.
+
+## Notation
+
++ We define a `Pow A ℝ` instance for `CFC.rpow`, i.e `a ^ y` with `A` an operator and `y : ℝ` works
+  as expected.
```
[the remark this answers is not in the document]
dupuisf: I'm afraid that in the unital case both of them will apply, and then it'll be painful to distinguish `rpowₙ` from `rpow`.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+/-!
+# Real powers defined via the continuous functional calculus
+
+This file defines real powers via the continuous functional calculus (CFC) and builds its API.
+This allows one to take real powers of matrices, operators, elements of a C⋆-algebra, etc. The
+square root is also defined via the non-unital CFC.
+
+## Main declarations
+
++ `CFC.rpowₙ`: the real power function based on the non-unital CFC, i.e. `cfcₙ NNReal.rpow`
```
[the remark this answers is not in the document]
dupuisf: Yeah that sounds good actually.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,339 @@
...
+  rw [cfc_pow (id : ℝ≥0 → ℝ≥0) n a, cfc_id ℝ≥0 a]
+
+@[simp]
+lemma rpow_algebraMap {x : ℝ≥0} {y : ℝ} :
+    (algebraMap ℝ≥0 A x) ^ y = algebraMap ℝ≥0 A (x ^ y) := by
+  simp only [rpow_def]
+  rw [cfc_algebraMap x (fun z : ℝ≥0 => z ^ y)]
+
+lemma rpow_add_of_zero_not_mem_spectrum {a : A} {x y : ℝ} (ha : 0 ∉ spectrum ℝ≥0 a) :
+    a ^ (x + y) = a ^ x * a ^ y := by
```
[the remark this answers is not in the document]
dupuisf: And also, we should figure out how to deal with strict positivity at some point.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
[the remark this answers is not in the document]
dupuisf: Actually I just tried putting the `Pow A ℝ≥0` instance and it seems fine.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
dupuisf: Fair enough.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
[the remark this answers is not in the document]
j-loreaux: That's what I expected from the elaborator for pow. The type of the exponent tells it what to look for. It's the same reason we don't have issues with `Nat` and `Int` powers in groups.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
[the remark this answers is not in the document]
j-loreaux: I have a plan for that. In progress.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,347 @@
...
+lemma nnrpow_def {a : A} {y : ℝ≥0} : a ^ y = cfcₙ (fun x : ℝ≥0 => x ^ (y : ℝ)) a := rfl
+
+lemma nnrpow_add {a : A} {x y : ℝ≥0} (hx : 0 < x) (hy : 0 < y) :
+    a ^ (x + y) = a ^ x * a ^ y := by
+  simp only [nnrpow_def]
+  rw [← cfcₙ_mul (fun z : ℝ≥0 => z ^ (x : ℝ)) (fun z : ℝ≥0 => z ^ (y : ℝ)) a]
+  congr
+  ext z
+  have : (x : ℝ) + (y : ℝ) ≠ 0 := by exact_mod_cast (ne_of_gt (add_pos hx hy))
+  simp [NNReal.rpow_add' _ this]
```
j-loreaux, line 79: This `rw` was *extremely* slow (thank goodness for incremental compilation!). I spent a short amount of time debugging, and about an hour coming up with a solution. The issue seems to be the following: `fun_prop (disch := aesop)` succeeded, but was very slow. Moreover, it was clear `aesop` was doing some sort of lifting because `fun_prop` alone failed. So I figured, I'll add an `@[fun_prop]` lemma of the form `∀ y : ℝ≥0, Continuous fun x : ℝ≥0 ↦ x ^ (y : ℝ)` because then it will find this and we'll be set. That *did* allow `fun_prop` alone to succeed, but it was still slow. However, `apply Continuous.continuousOn; fun_prop` was instant. This suggested to me that `fun_prop` was going down some rabbit hole involving the coercion because too much was unfolded already. So, my solution is this: we add
```
namespace NNReal

noncomputable
abbrev nnrpow (a : ℝ≥0) (b : ℝ≥0) : ℝ≥0 := a ^ (b : ℝ)

-- maybe this could even be marked `@[simp]`
lemma nnrpow_def (a b : ℝ≥0) : nnrpow a b = a ^ (b : ℝ) := rfl

@[fun_prop]
lemma continuous_nnrpow_const (y : ℝ≥0) : Continuous (nnrpow · y) :=
  continuous_rpow_const zero_le_coe

end NNReal
```
Then change the following def and lemma to:
```
noncomputable def nnrpow (a : A) (y : ℝ≥0) : A := cfcₙ (NNReal.nnrpow · y) a

lemma nnrpow_def {a : A} {y : ℝ≥0} : a ^ y = cfcₙ (NNReal.nnrpow · y) a := rfl
```
It is a bit interesting that `abbrev NNReal.nnrpow` is enough, but it seems to be. Indeed, this allows for the quite fast and nice proof of this lemma:
```suggestion
lemma nnrpow_add {a : A} {x y : ℝ≥0} (hx : 0 < x) (hy : 0 < y) :
    a ^ (x + y) = a ^ x * a ^ y := by
  simp only [nnrpow_def]
  rw [← cfcₙ_mul _ _ a]
  congr! 2 with z
  exact_mod_cast NNReal.rpow_add' z <| ne_of_gt (add_pos hx hy)
```
(Note: it is possible to increase the speed further by providing a few `simp` lemmas for `NNReal.nnrpow` that allow us to prove `NNReal.nnrpow 0 x = 0` faster, because at this point the slowest part of the proof is the `aesop` calls to show this in the auto-params for `cfcₙ_mul`, but the `fun_prop` calls are now nearly instantaneous.)

This suggestion requires adding `NNReal.nnrpow_def` to some of the `simp` calls elsewhere.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,347 @@
...
+          Continuous.continuousOn <| NNReal.continuous_rpow_const (le_of_lt hx)
+        have hmapzero : ∀ r : ℝ≥0, 0 < r → (fun z : ℝ≥0 => z ^ (r : ℝ)) 0 = 0 := fun r hr => by
+          ext
+          simp only [NNReal.coe_rpow, NNReal.coe_zero, le_refl]
+          rw [Real.zero_rpow (by exact_mod_cast ne_of_gt hr)]
+        rw [← cfcₙ_comp (fun z : ℝ≥0 => z ^ (y : ℝ)) (fun z : ℝ≥0 => z ^ (x : ℝ)) a h₁
+          (hmapzero y hy) h₂ (hmapzero x hx)]
+        have : (fun z : ℝ≥0 => z ^ (y : ℝ)) ∘ (fun z : ℝ≥0 => z ^ (x : ℝ))
+            = fun z => z ^ ((x : ℝ) * y) := by ext; simp [Real.rpow_mul]
+        simp [this]
```
j-loreaux, line 132: With the above suggestion, we get:
```suggestion
        rw [← cfcₙ_comp _ _ a]
        congr! 2 with u
        ext
        simp [NNReal.nnrpow_def, Real.rpow_mul]
```
and it's pretty fast.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,347 @@
...
+  [UniqueNonUnitalContinuousFunctionalCalculus ℝ≥0 A]
+
+lemma nnrpow_eq_rpow {a : A} {x : ℝ≥0} (hx : 0 < x) : a ^ x = a ^ (x : ℝ) := by
+  have hzero : NNReal.rpow 0 x = 0 := by
+    have : x ≠ 0 := ne_of_gt hx
+    simp [this]
+  have hcont : ContinuousOn (fun z : ℝ≥0 => z ^ (x : ℝ)) (quasispectrum ℝ≥0 a) :=
+    fun _ _ => ContinuousAt.continuousWithinAt
+      <| NNReal.continuousAt_rpow_const (Or.inr NNReal.zero_le_coe)
+  rw [nnrpow_def, rpow_def, cfcₙ_eq_cfc hcont hzero]
```
j-loreaux, line 284: also with the above:
```suggestion
  rw [nnrpow_def, rpow_def, cfcₙ_eq_cfc]
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,347 @@
...
+  case pos =>
+    by_cases hx : 0 < x
+    case neg =>
+      replace hx : x = 0 := eq_of_le_of_not_lt (le_of_not_lt hx) not_lt_zero'
+      simp [hx]
+    case pos =>
+      by_cases hy : 0 < y
+      case neg =>
+        replace hy : y = 0 := eq_of_le_of_not_lt (le_of_not_lt hy) not_lt_zero'
+        simp [hy]
```
j-loreaux, line 116: You can deal with all the `0` cases like this quickly:
```
    obtain (rfl | hx) := eq_zero_or_pos x <;> obtain (rfl | hx) := eq_zero_or_pos y
    all_goals try simp
```
By the way, I knew we had to have that lemma: `eq_zero_or_pos`, and it was found instantly with a quick [Loogle](https://loogle.lean-lang.org/?q=%3Fx+%3D+0+∨+0+<+%3Fx)

j-loreaux (commented):
This is all I have the time or energy for right now. I think there is some more to be cleaned up yet regarding automation. I think we're missing a few more `fun_prop` lemmas, or possibly even some continuity lemmas about `rpow` (the OG `NNReal.rpow` that is, not `CFC.rpow`).

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,333 @@
...
+  simp only [rpow_def]
+  rw [cfc_algebraMap x (fun z : ℝ≥0 => z ^ y)]
+
+lemma rpow_add_of_zero_not_mem_spectrum {a : A} {x y : ℝ} (ha : 0 ∉ spectrum ℝ≥0 a) :
+    a ^ (x + y) = a ^ x * a ^ y := by
+  simp only [rpow_def, NNReal.rpow_eq_pow]
+  have h : ∀ r, ContinuousOn (fun z : ℝ≥0 => z ^ (r : ℝ)) (spectrum ℝ≥0 a) := by
+    intro r z hz
+    exact ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const <| Or.inl <| by aesop
+  rw [← cfc_mul _ _ a (h x) (h y)]
```
j-loreaux, line 219: It's not that the proof is super long, but we should definitely have this lemma, and we can make the automation work for us:
```lean
@[fun_prop]
theorem _root_.NNReal.continuousOn_rpow_const_compl_zero {r : ℝ} :
    ContinuousOn (fun z : ℝ≥0 => z ^ r) {0}ᶜ :=
  fun _ h => ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const (.inl h)

-- even though this follows from `ContinuousOn.mono` and the previous lemma, we include it for
-- automation purposes with `fun_prop`, because the side goal `0 ∉ s ∨ 0 ≤ r` is often easy to check
@[fun_prop]
theorem _root_.NNReal.continuousOn_rpow_const {r : ℝ} {s : Set ℝ≥0}
    (h : 0 ∉ s ∨ 0 ≤ r) : ContinuousOn (fun z : ℝ≥0 => z ^ r) s :=
  h.elim (fun _ ↦ ContinuousOn.mono (s := {0}ᶜ) (by fun_prop) (by aesop))
    (NNReal.continuous_rpow_const · |>.continuousOn)
```
and then we get:
```suggestion
  rw [← cfc_mul _ _ a]
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,333 @@
...
+    (a ^ x) ^ y = a ^ (x * y) := by
+  simp only [rpow_def]
+  have h₁ : ContinuousOn (fun z : ℝ≥0 => z ^ (y : ℝ))
+      ((fun z : ℝ≥0 => z ^ (x : ℝ)) '' spectrum ℝ≥0 a) := by
+    intro z hz
+    exact ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const <| Or.inl <| by aesop
+  have h₂ : ContinuousOn (fun z : ℝ≥0 => z ^ (x : ℝ)) (spectrum ℝ≥0 a) := by
+    intro z hz
+    exact ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const <| Or.inl <| by aesop
+  rw [← cfc_comp _ _ a ha₂ h₁ h₂]
```
j-loreaux, line 237: With the above:
```suggestion
  rw [← cfc_comp _ _ a]
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,333 @@
...
+
+lemma rpow_rpow_of_exponent_nonneg {a : A} {x y : ℝ} (hx : 0 ≤ x) (hy : 0 ≤ y)
+    (ha₂ : 0 ≤ a := by cfc_tac) : (a ^ x) ^ y = a ^ (x * y) := by
+  simp only [rpow_def]
+  have h₁ : ContinuousOn (fun z : ℝ≥0 => z ^ (y : ℝ))
+      ((fun z : ℝ≥0 => z ^ (x : ℝ)) '' spectrum ℝ≥0 a) :=
+    fun _ _ => ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const (Or.inr hy)
+  have h₂ : ContinuousOn (fun z : ℝ≥0 => z ^ (x : ℝ)) (spectrum ℝ≥0 a) :=
+    fun _ _ => ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const (Or.inr hx)
+  rw [← cfc_comp _ _ a ha₂ h₁ h₂]
```
j-loreaux, line 249: With the above
```suggestion
  rw [← cfc_comp _ _ a]
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,333 @@
...
+  refine cfc_congr fun _ _ => ?_
+  simp [NNReal.rpow_mul]
+
+lemma rpow_one {a : A} (ha : 0 ≤ a := by cfc_tac) : a ^ (1 : ℝ) = a := by
+  simp only [rpow_def, NNReal.coe_one, NNReal.rpow_eq_pow, NNReal.rpow_one]
+  change cfc (id : ℝ≥0 → ℝ≥0) a = a
+  rw [cfc_id ℝ≥0 a]
```
j-loreaux, line 256: We have a variant lemma for this:
```suggestion
  simp only [rpow_def, NNReal.coe_one, NNReal.rpow_eq_pow, NNReal.rpow_one, cfc_id' ℝ≥0 a]
```

j-loreaux (commented):
Here's another set of simplifications by making appropriate `fun_prop` lemmas. As I said before, `fun_prop` is not perfect, but it's quite good overall, and if you learn not to lead it astray, then I think we can make it succeed most of the time.

Note: the `∨` I used in the statement is precisely to avoid leading it astray in this way. Both of those conditions should be relatively easy for a smart discharger like `aesop` to either succeed in one, or fail in both. And in the latter case, the result will likely not be true (or else it will require a much more sophisticated human-supplied argument).

(I'll probably have one last round of review after this.)

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,314 @@
...
+lemma rpow_one {a : A} (ha : 0 ≤ a := by cfc_tac) : a ^ (1 : ℝ) = a := by
+  simp only [rpow_def, NNReal.coe_one, NNReal.rpow_eq_pow, NNReal.rpow_one, cfc_id' ℝ≥0 a]
+
+@[simp]
+lemma one_rpow {x : ℝ} : (1 : A) ^ x = (1 : A) := by simp [rpow_def]
+
+lemma rpow_zero {a : A} (ha : 0 ≤ a := by cfc_tac) : a ^ (0 : ℝ) = 1 := by
+  simp [rpow_def, cfc_const_one ℝ≥0 a]
+
+lemma zero_rpow {x : ℝ} (hx : x ≠ 0) : rpow (0 : A) x = 0 := by simp [rpow, NNReal.zero_rpow hx]
```
j-loreaux, line 245: can you move these up to immediately after `rpow_def`?

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,314 @@
...
+  rw [← cfc_pow_id (R := ℝ≥0) a n, rpow_def]
+  congr
+  simp
+
+@[simp]
+lemma rpow_algebraMap {x : ℝ≥0} {y : ℝ} :
+    (algebraMap ℝ≥0 A x) ^ y = algebraMap ℝ≥0 A (x ^ y) := by
+  rw [rpow_def, cfc_algebraMap ..]
+
+lemma rpow_add_of_zero_not_mem_spectrum {a : A} {x y : ℝ} (ha : 0 ∉ spectrum ℝ≥0 a) :
```
j-loreaux, line 211: I think this should be the default assumption here.
```suggestion
lemma rpow_add {a : A} {x y : ℝ} (ha : 0 ∉ spectrum ℝ≥0 a) :
```
and likewise for other lemmas, I think we can drop the hypothesis from the name.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,314 @@
...
+  rw [← cfc_comp _ _ a ha₂]
+  refine cfc_congr fun _ _ => ?_
+  simp [NNReal.rpow_mul]
+
+lemma rpow_rpow_of_exponent_nonneg {a : A} {x y : ℝ} (hx : 0 ≤ x) (hy : 0 ≤ y)
+    (ha₂ : 0 ≤ a := by cfc_tac) : (a ^ x) ^ y = a ^ (x * y) := by
+  simp only [rpow_def]
+  rw [← cfc_comp _ _ a]
+  refine cfc_congr fun _ _ => ?_
+  simp [NNReal.rpow_mul]
```
j-loreaux, line 234: somewhere in this section (maybe around here?) it would be nice to have at least these lemmas:
```
lemma rpow_mul_rpow_neg {a : A} (x : ℝ) (ha : 0 ∉ spectrum ℝ≥0 a)
    (ha' : 0 ≤ a := by cfc_tac) : a ^ x * a ^ (-x) = 1 := by
  rw [← rpow_add ha, add_right_neg, rpow_zero a]

lemma rpow_neg_mul_rpow {a : A} (x : ℝ) (ha : 0 ∉ spectrum ℝ≥0 a)
    (ha' : 0 ≤ a := by cfc_tac) : a ^ (-x) * a ^ x = 1 := by
  rw [← rpow_add ha, add_left_neg, rpow_zero a]

lemma inv_nonneg_of_nonneg (a : Aˣ) (ha : (0 : A) ≤ a := by cfc_tac) : (0 : A) ≤ a⁻¹ :=
  cfc_inv_id (R := ℝ≥0) a ▸ cfc_predicate _ (a : A)

lemma inv_nonneg (a : Aˣ)  : (0 : A) ≤ a⁻¹ ↔ (0 : A) ≤ a :=
  ⟨fun _ ↦ inv_inv a ▸ inv_nonneg_of_nonneg a⁻¹, fun _ ↦ inv_nonneg_of_nonneg a⟩

lemma rpow_neg_one_eq_inv (a : Aˣ) (ha : (0 : A) ≤ a := by cfc_tac) :
    a ^ (-1 : ℝ) = (↑a⁻¹ : A) := by
  refine a.inv_eq_of_mul_eq_one_left ?_ |>.symm
  simpa [rpow_one (a : A)] using rpow_neg_mul_rpow 1 (spectrum.zero_not_mem ℝ≥0 a.isUnit)

lemma rpow_neg {a : Aˣ} {x : ℝ} (ha' : (0 : A) ≤ a := by cfc_tac) :
    (a : A) ^ (-x) = (↑a⁻¹ : A) ^ x := by
  sorry -- possibly combine with the previous lemma and `rpow_rpow`

lemma rpow_intCast {a : Aˣ} {n : ℤ} (ha : (0 : A) ≤ a := by cfc_tac) :
    (a : A) ^ (n : ℝ) = (↑(a ^ n) : A) := by
  sorry -- follows easily from the previous lemma?
```

Mathlib/Analysis/SpecialFunctions/Pow/Continuity.lean
```diff
@@ -417,6 +417,19 @@ theorem continuousAt_rpow_const {x : ℝ≥0} {y : ℝ} (h : x ≠ 0 ∨ 0 ≤ y
...
+    ContinuousOn (fun z : ℝ≥0 => z ^ r) {0}ᶜ :=
+  fun _ h => ContinuousAt.continuousWithinAt <| NNReal.continuousAt_rpow_const (.inl h)
+
+-- even though this follows from `ContinuousOn.mono` and the previous lemma, we include it for
+-- automation purposes with `fun_prop`, because the side goal `0 ∉ s ∨ 0 ≤ r` is often easy to check
+@[fun_prop]
+theorem continuousOn_rpow_const {r : ℝ} {s : Set ℝ≥0}
+    (h : 0 ∉ s ∨ 0 ≤ r) : ContinuousOn (fun z : ℝ≥0 => z ^ r) s :=
+  h.elim (fun _ ↦ ContinuousOn.mono (s := {0}ᶜ) (by fun_prop) (by aesop))
+    (NNReal.continuous_rpow_const · |>.continuousOn)
```
j-loreaux, line 431: So I found out something *very* unfortunate here. The stuff that was supposed to be fast from this is slow! :fearful: It turns out that I had declared this and the attribute inside the file we were working on, so it was tried first and was fast. But at this location, it is tried later, and it's very slow (and seemingly for no good reason if you look at the trace). We desperately need to figure that out. Sorry this PR is turning out to be so much work, but I think we're learning some valuable lessons here, and fixing this kind of stuff now is important before we get in too deep.

j-loreaux (commented):
gah, more stuff! And some `fun_prop` problems. This is one of the first times I've really been bitten hard by it. It also makes me question whether the `NNReal.nnrpow` was really a true solution or it just got lucky from being defined close.

Mathlib/Analysis/SpecialFunctions/Pow/Continuity.lean
j-loreaux: By the way, you can tell it's just about the location where it's declared, because you can "redeclare" the attribute with `attribute [fun_prop] continuousOn_rpow_const` in the `CFC.rpow` file and the immediately become fast again.

dupuisf: I'm starting to think that `fun_prop` isn't the problem at all, *everything* is slow in this file (and pretty much every other file about the CFC), especially in the last section where we use both the unital and non-unital CFC. Try sorry-ing out a bunch of lemmas and you'll see what I mean: even then it's slow!

dupuisf: I think whatever is causing that problem is what is slowing down fun_prop: if it takes ~0.5s when it immediately picks the right lemma, it's not surprising that it takes 4s when it has to try 8 things.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,314 @@
...
+  simp only [nnrpow_def, NNReal.nnrpow_def, NNReal.coe_one, NNReal.rpow_one]
+  change cfcₙ (id : ℝ≥0 → ℝ≥0) a = a
+  rw [cfcₙ_id ℝ≥0 a]
+
+lemma nnrpow_two {a : A} (ha : 0 ≤ a := by cfc_tac) : a ^ (2 : ℝ≥0) = a * a := by
+  simp only [nnrpow_def, NNReal.nnrpow_def, NNReal.coe_ofNat, NNReal.rpow_ofNat, pow_two]
+  change cfcₙ (fun z : ℝ≥0 => id z * id z) a = a * a
+  rw [cfcₙ_mul id id a, cfcₙ_id ℝ≥0 a]
+
+lemma nnrpow_three {a : A} (ha : 0 ≤ a := by cfc_tac) : a ^ (3 : ℝ≥0) = a * a * a := by
```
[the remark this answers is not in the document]
dupuisf: Well, it's done so we might as well keep it!

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,314 @@
...
+We define two separate versions `CFC.nnrpow` and `CFC.rpow` due to what happens at 0. Since
+`NNReal.rpow 0 0 = 1`, this means that this function does not map zero to zero when the exponent
+is zero, and hence `CFC.nnrpow a 0 = 0` whereas `CFC.rpow a 0 = 1`. Note that the non-unital version
+only makes sense for nonnegative exponents, and hence we define it such that the exponent is in
+`ℝ≥0`.
+
+## Notation
+
++ We define a `Pow A ℝ` instance for `CFC.rpow`, i.e `a ^ y` with `A` an operator and `y : ℝ` works
+  as expected. Likewise, we define a `Pow A ℝ≥0` instance for `CFC.nnrpow`.
```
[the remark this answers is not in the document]
dupuisf: Yeah, that's what I had in mind when I wrote about "other instances when they are available" in the docstring.

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,388 @@
...
+      aesop
+    rw [Set.mem_image] at hz
+    obtain ⟨t, ht⟩ := hz
+    have h₂ := this t ht.1
+    rw [← ht.2]
+    exact inv_ne_zero h₂
+  rw [← cfc_inv_id (R := ℝ≥0) a, rpow_def, rpow_def,
+      ← cfc_comp' (fun z => z ^ x) (Inv.inv : ℝ≥0 → ℝ≥0) (a : A) h₁]
+  refine cfc_congr fun _ _ => ?_
+  simp [NNReal.rpow_neg, NNReal.inv_rpow]
```
j-loreaux, line 313: :golf: You can just use the `NNReal.continuousOn_rpow_const` lemma we had here, plus some convenience lemmas concerning zero in the spectrum.
```suggestion
  suffices h₁ : ContinuousOn (fun z ↦ z ^ x) (Inv.inv '' (spectrum ℝ≥0 (a : A))) by
    rw [← cfc_inv_id (R := ℝ≥0) a, rpow_def, rpow_def,
        ← cfc_comp' (fun z => z ^ x) (Inv.inv : ℝ≥0 → ℝ≥0) (a : A) h₁]
    refine cfc_congr fun _ _ => ?_
    simp [NNReal.rpow_neg, NNReal.inv_rpow]
  refine NNReal.continuousOn_rpow_const (.inl ?_)
  rintro ⟨z, hz, hz'⟩
  exact spectrum.zero_not_mem ℝ≥0 a.isUnit <| inv_eq_zero.mp hz' ▸ hz
```

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
```diff
@@ -0,0 +1,380 @@
...
+
+lemma rpow_neg_mul_rpow {a : A} (x : ℝ) (ha : 0 ∉ spectrum ℝ≥0 a)
+    (ha' : 0 ≤ a := by cfc_tac) : a ^ (-x) * a ^ x = 1 := by
+  rw [← rpow_add ha, add_left_neg, rpow_zero a]
+
+lemma inv_nonneg_of_nonneg (a : Aˣ) (ha : (0 : A) ≤ a := by cfc_tac) : (0 : A) ≤ a⁻¹ :=
+  cfc_inv_id (R := ℝ≥0) a ▸ cfc_predicate _ (a : A)
+
+lemma inv_nonneg (a : Aˣ)  : (0 : A) ≤ a⁻¹ ↔ (0 : A) ≤ a :=
+  ⟨fun _ ↦ inv_inv a ▸ inv_nonneg_of_nonneg a⁻¹, fun _ ↦ inv_nonneg_of_nonneg a⟩
```
j-loreaux, line 289: Can you move these into `Analysis.CStarAlgebra.ContinuousFunctionalCalculus.Unital`?

j-loreaux (commented):
Please merge master, as I expect there may be some breakage. Also, now that #11092 is merged, it would be interesting to see if it helps at all without that "redeclaration" of the `fun_prop` attribute (although given our other testing, I doubt it). In any case:

bors d+

Mathlib/Analysis/SpecialFunctions/ContinuousFunctionalCalculus/Rpow.lean
dupuisf: I did, but I couldn't generalize them properly because `OrderedSemifield` doesn't exist, so I left them specialized to `ℝ≥0`.

PR #15384, opened 2024-07-31
[Merged by Bors] - feat(RelSeries): head_singleton, last_singleton

nomeata:
from the Carleson project.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15385, opened 2024-07-31
feat(RelSeries): replaceLast

nomeata:
Replaces the last element in a series with a (non-strictly) larger element.
Essentially `p.eraseLast.snoc x`, but also works when `p` is a singleton.

From the Carleson project.

---

- [ ] depends on: #15384


nomeata:
Actually, I might be able to do without this maybe too obscure operation. Let me check.

PR #15391, opened 2024-08-01
[Merged by Bors] - feat: add `OrderedSMul ℝ A` instance for `StarOrderedRing A`

j-loreaux:
This instance is already available in the scope `ComplexOrder`, but for `ℝ` that is unnecesary, so we make it a global instance.

This instance could be moved earlier (so that we don't make use of `RCLike`), but it's probably not that important to do so.

---


urkud:
If there are no benchmark regressions, then please merge it. Thanks!
bors d+

j-loreaux:
bors merge

PR #15401, opened 2024-08-01
[Merged by Bors] - refactor(Condensed): reorganize TopComparison

dagurtomas:
Define a general functor from `TopCat` to sheaves on `CompHausLike P` satisfying some properties, and redefine `topCatToCondensedSet` and `topCatToLightCondSet` in terms of that functor.

---

This simplifies #15321, which the characterization of discrete condensed objects (#14027) depends on.

- [x] depends on: #15518


joelriou:
bors d+

As the variable inclusion has recently changed, it may be safer to sync with master.

dagurtomas:
bors merge

PR #15403, opened 2024-08-01
[Merged by Bors] - feat: Essential supremum according to the counting measure

YaelDillies:
... is just the usual supremum

From LeanAPAP

---


j-loreaux:
I'll let someone with more familiarity with the measure theory library make the final decision here.

maintainer merge

Mathlib/MeasureTheory/OuterMeasure/AE.lean
```diff
@@ -106,15 +106,25 @@ theorem ae_ball_iff {ι : Type*} {S : Set ι} (hS : S.Countable) {p : α → ∀
     (∀ᵐ x ∂μ, ∀ i (hi : i ∈ S), p x i hi) ↔ ∀ i (hi : i ∈ S), ∀ᵐ x ∂μ, p x i hi :=
   eventually_countable_ball hS
 
-theorem ae_eq_refl (f : α → β) : f =ᵐ[μ] f :=
-  EventuallyEq.rfl
+lemma ae_eq_refl (f : α → β) : f =ᵐ[μ] f := EventuallyEq.rfl
+lemma ae_eq_rfl {f : α → β} : f =ᵐ[μ] f := EventuallyEq.rfl
```
semorrison, line 110: Why this duplication? Could one be deprecated?

semorrison:
Please redo `maintainer-merge` afterwards.

Mathlib/MeasureTheory/OuterMeasure/AE.lean
YaelDillies: We have this duplication for every concrete reflexive relation, eg `le_refl`/`le_rfl`. I'm not sure this PR is the place to upturn such a widespread pattern and naming convention.

YaelDillies:
Like so?

maintainer merge

b-mehta:
Removing my request for review here, since I don't know this part of the library well enough to comment.

Mathlib/MeasureTheory/OuterMeasure/AE.lean
semorrison: :-( This makes me really sad, but okay.

RemyDegenne:
Thanks!
bors r+

PR #15404, opened 2024-08-01
[Merged by Bors] - feat: Adds a variant of lemma about eigenvectors.

adamtopaz:
---


b-mehta (approved):
Thanks!

bors merge

PR #15407, opened 2024-08-01
[Merged by Bors] - feat(Topology): set of generic points of generic components

erdOne:
---


Mathlib/Topology/Sober.lean
```diff
@@ -112,6 +113,11 @@ theorem IsIrreducible.genericPoint_spec [QuasiSober α] {S : Set α} (hS : IsIrr
     IsGenericPoint hS.genericPoint (closure S) :=
   (QuasiSober.sober hS.closure isClosed_closure).choose_spec
 
+theorem IsIrreducible.genericPoint_spec' [QuasiSober α] {S : Set α}
+    (hS : IsIrreducible S) (hS' : IsClosed S) :
+    IsGenericPoint hS.genericPoint S := by
+  convert hS.genericPoint_spec; exact hS'.closure_eq.symm
```
callesonne, line 119: The following might be out of scope of this PR, but to me it feels a bit odd that this version has the `'`, and that it is not immediate by definition, given how `QuasiSober` is defined (that closed irr sets have generic points).

I understand that it is useful for `IsIrreducible.genericPoint` to make sense for non-closed sets as well, but then shouldn't the definition of `QuasiSober` already reflect that? So that its defining property would instead be
` sober : ∀ {S : Set α}, IsIrreducible S → ∃ x, IsGenericPoint x (closure S)`
Then it would make more sense to me that the prime versions are for the case of closed sets.

Mathlib/Topology/Sober.lean
```diff
@@ -112,6 +113,11 @@ theorem IsIrreducible.genericPoint_spec [QuasiSober α] {S : Set α} (hS : IsIrr
...
+
 @[simp]
 theorem IsIrreducible.genericPoint_closure_eq [QuasiSober α] {S : Set α} (hS : IsIrreducible S) :
```
callesonne, line 122: Could you add a prime version of this too, so the fact that when `S` is closed the closure of its generic point equals S.

callesonne (commented):
I only have some comments about `IsIrreducible.genericPoint_spec'`, other than that it all looks good!

Mathlib/Topology/Sober.lean
erdOne: I changed the names a bit. Please check if you like them.

ADedecker:
I am by no means used to the conventions around these things, but isn't `genericPoints` a bit too vague of a name ? As in, I would expect it to be the set of points generic in the whole space (granted, this is useless when you have a non-irreducible space). But if you think there's no risk of misunderstanding I'll believe you.

erdOne:
I might be misunderstanding you, but isn't the definition proposed in this PR exactly what you described (when the space is irreducible)?

ADedecker:
Yes it is, it's just that, purely from the name `genericPoints`, I was expecting it to be empty in the non-irreducible setting 😅 

Really the only thing is that it sounds surprising to have `IsGenericPoint` and `genericPoints` not being as trivially related as the names might suggest. But since one needs 10 seconds to understand the subtlety I think it doesn't matter much and you can ignore my comments.

callesonne (approved):
LGTM!

Mathlib/Topology/Sober.lean
```diff
@@ -222,3 +238,66 @@ instance (priority := 100) T2Space.quasiSober [T2Space α] : QuasiSober α where
...
+noncomputable
+def ofComponent [QuasiSober α] (x : irreducibleComponents α) : genericPoints α :=
+  ⟨x.2.1.genericPoint, show _ ∈ irreducibleComponents α from
+    (x.2.1.isGenericPoint_genericPoint (isClosed_of_mem_irreducibleComponents x.1 x.2)).symm ▸ x.2⟩
+
+lemma isGenericPoint_ofComponent [QuasiSober α] (x : irreducibleComponents α) :
+    IsGenericPoint (ofComponent x).1 x :=
+    x.2.1.isGenericPoint_genericPoint (isClosed_of_mem_irreducibleComponents x.1 x.2)
+
+lemma component_ofComponent [QuasiSober α] (x : irreducibleComponents α) :
```
jcommelin, line 269:
```suggestion
@[simp]
lemma component_ofComponent [QuasiSober α] (x : irreducibleComponents α) :
```

Mathlib/Topology/Sober.lean
```diff
@@ -222,3 +238,66 @@ instance (priority := 100) T2Space.quasiSober [T2Space α] : QuasiSober α where
...
+    component (ofComponent x) = x :=
+  Subtype.ext (isGenericPoint_ofComponent x)
+
+lemma ofComponent_component [T0Space α] [QuasiSober α] (x : genericPoints α) :
```
jcommelin, line 273:
```suggestion
@[simp]
lemma ofComponent_component [T0Space α] [QuasiSober α] (x : genericPoints α) :
```

jcommelin (commented):
bors d+

erdOne:
bors merge

PR #15409, opened 2024-08-01
chore (Nat.Find): avoid well-founded recursion 

mattrobball:
Since well-founded recursion is irreducible and not straightforward to unfold manually, it is better to avoid it if possible. We use @kmill's implementation from [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.E2.9C.94.20Computability.20and.20Nat.2Efind/near/442669686) for `Nat.findX`.

---

### DIFF
--- a/Mathlib/Data/Nat/Find.lean
+++ b/Mathlib/Data/Nat/Find.lean
@@ -24,29 +24,27 @@ private def lbp (m n : ℕ) : Prop :=
 
 variable [DecidablePred p] (H : ∃ n, p n)
 
-private def wf_lbp : WellFounded (@lbp p) :=
-  ⟨let ⟨n, pn⟩ := H
-    suffices ∀ m k, n ≤ k + m → Acc lbp k from fun a => this _ _ (Nat.le_add_left _ _)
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Nat/Find.lean
```diff
@@ -24,29 +24,27 @@ private def lbp (m n : ℕ) : Prop :=
...
+where
+  go (H : ∃ n, p n) (n : ℕ) (H' : ∀ n' < n, ¬ p n') : { n // p n ∧ ∀ m < n, ¬p m } :=
+    if h : p n then ⟨n, h, H'⟩
+    else
+      go H (n + 1) <| by
+        intro n' hn'
+        obtain hn' | rfl := lt_or_eq_of_le (le_of_lt_succ hn')
+        · exact H' n' hn'
+        · exact h
+  termination_by H.choose - n
```
eric-wieser, line 38: This adds a dependency on the axiom of choice, which is perhaps unfortunate and something @digama0  has opinions about.

PR #15413, opened 2024-08-01
[Merged by Bors] - chore: remove some uses of `open Classical`, part 2

sgouezel:
See #15371.

---


Mathlib/Algebra/GroupWithZero/Units/Basic.lean
```diff
@@ -454,10 +453,9 @@ end CommGroupWithZero
 
 section NoncomputableDefs
 
-open scoped Classical
-
 variable {M : Type*} [Nontrivial M]
 
+open Classical in
```
eric-wieser, line 458:
```suggestion
open scoped Classical in
```

Mathlib/Algebra/MonoidAlgebra/Degree.lean
```diff
@@ -232,13 +233,15 @@ theorem supDegree_sub_le {f g : R'[A]} :
...
-  (Finset.sup_mono Finsupp.support_finset_sum).trans_eq (Finset.sup_biUnion _ _)
+    (∑ i ∈ s, f i).supDegree D ≤ s.sup (fun i => (f i).supDegree D) := by
+  classical
+  exact (Finset.sup_mono Finsupp.support_finset_sum).trans_eq (Finset.sup_biUnion _ _)
 
 theorem supDegree_single_ne_zero (a : A) {r : R} (hr : r ≠ 0) :
     (single a r).supDegree D = D a := by
   rw [supDegree, Finsupp.support_single_ne_zero a hr, Finset.sup_singleton]
 
+open Classical in
```
grunweg, line 244: Is this needed, or can you add suitable `DecidableEq` hypotheses?

Mathlib/Algebra/Order/CauSeq/Completion.lean
```diff
@@ -185,6 +183,7 @@ instance instRatCast : RatCast (Cauchy abv) where ratCast q := ofRat q
 @[simp, norm_cast] lemma ofRat_nnratCast (q : ℚ≥0) : ofRat (q : β) = (q : Cauchy abv) := rfl
 @[simp, norm_cast] lemma ofRat_ratCast (q : ℚ) : ofRat (q : β) = (q : Cauchy abv) := rfl
 
+open Classical in
```
grunweg, line 186: Same question here.

Mathlib/AlgebraicTopology/TopologicalSimplex.lean
```diff
@@ -49,6 +50,7 @@ def toTopMap {x y : SimplexCategory} (f : x ⟶ y) (g : x.toTopObj) : y.toTopObj
       simp [Finset.eq_univ_iff_forall]
     · apply Set.pairwiseDisjoint_filter⟩
 
+open Classical in
```
grunweg, line 53: Same question here.

Mathlib/Analysis/Analytic/Basic.lean
```diff
@@ -80,10 +80,7 @@ noncomputable section
 
 variable {𝕜 E F G : Type*}
 
-open scoped Classical
-open Topology NNReal Filter ENNReal
-
-open Set Filter Asymptotics
+open Topology NNReal Filter ENNReal Set Asymptotics
```
grunweg, line 83: Nice clean-up!

Mathlib/Algebra/GroupWithZero/Units/Basic.lean
```diff
@@ -72,8 +72,7 @@ theorem not_isUnit_zero [Nontrivial M₀] : ¬IsUnit (0 : M₀) :=
 
 namespace Ring
 
-open scoped Classical
-
+open Classical in
```
grunweg, line 75: Does `open scoped Classical` also work? Would you like to use that instead? (Also applies in ~30 other places.)

Mathlib/MeasureTheory/MeasurableSpace/Defs.lean
```diff
@@ -195,6 +193,7 @@ protected theorem MeasurableSet.ite {t s₁ s₂ : Set α} (ht : MeasurableSet t
     (h₁ : MeasurableSet s₁) (h₂ : MeasurableSet s₂) : MeasurableSet (t.ite s₁ s₂) :=
   (h₁.inter ht).union (h₂.diff ht)
 
+open Classical in
```
grunweg, line 196: Same comment as above.

Mathlib/MeasureTheory/Measure/GiryMonad.lean
```diff
@@ -30,11 +30,7 @@ giry monad
 
 noncomputable section
 
-open scoped Classical
-open ENNReal
-
-open scoped Classical
-open Set Filter
+open ENNReal Set Filter
```
grunweg, line 33: Nice find!

grunweg (commented):
Thanks for doing this! I have a few comments, which should be quick to address. 
maintainer delegate, i.e. feel free to maintainer merge on my behalf once you have addressed them.

Mathlib/Algebra/MonoidAlgebra/Degree.lean
eric-wieser: I suspect the answer is "let's scoped all the `Classical`s first, and remove them later"

Mathlib/Algebra/GroupWithZero/Units/Basic.lean
sgouezel: I have used `open Classical in` instead of `open scoped Classical in` as it doesn't make a real difference (if it created a name conflict, one would spot it right away) and is a little bit shorter. If you want, I can change  all over the PR, just tell me.

Mathlib/Algebra/MonoidAlgebra/Degree.lean
sgouezel: For each statement, I have tried to balance between ease of use and adding decidability assumptions. If we notice that some of my choices create friction elsewhere, we can go back and fix them later.

Mathlib/Algebra/Order/CauSeq/Completion.lean
sgouezel: Here it's an instance, so I definitely don't want decidability assumptions, to make sure it's available all the time. In general, in definitions and instances, `open Classical in` is always the way to go, while in theorems one has to double-think.

Mathlib/AlgebraicTopology/TopologicalSimplex.lean
sgouezel: It's for rewriting from left to right, as above. In general, if decidability only appears on the right of an equality, I went for `open Classical in` for this reason.

Mathlib/Algebra/GroupWithZero/Units/Basic.lean
sgouezel: Yes, it works, but doesn't make a difference and is a little bit longer.

grunweg: Fair enough; I won't insist on that one.

Mathlib/Algebra/Order/CauSeq/Completion.lean
grunweg: We also have instancen which explicitly take a `DecidableXX` assumption...

grunweg (commented):
You have well-thought answers to all comments: let's have a second pair of eyes on this.
maintainer merge

Mathlib/Algebra/Category/Grp/FilteredColimits.lean
```diff
@@ -26,11 +26,7 @@ universe v u
 
 noncomputable section
 
-open scoped Classical
-
-open CategoryTheory
-
-open CategoryTheory.Limits
+open CategoryTheory CategoryTheory.Limits
```
PatrickMassot, line 29: Is there any reason to avoid the modern style ``` open CategoryTheory Limits` ?

PatrickMassot:
bors merge

PR #15414, opened 2024-08-01
[Merged by Bors] - fix(lint-style): mention the correct size limit in 'file too long' errors

grunweg:
The previous error message (for github consumption) was wrong in two ways:
- it would always print the size limit, even when there was no previous exception for this file
- it would print the *new* limit, not the previous one To fix this, we need to make an existing limit part of the error data.

---


semorrison:
bors merge

grunweg:
Thank you for the fast review!

PR #15416, opened 2024-08-01
[Merged by Bors] - feat: `RCLike` versions of Hahn-Banach separation theorems

JonBannon:
Generalized first definition in `Mathlib.Analysis.NormedSpace.Extend`, and used this to prove `RCLike` generalizations of Hahn-Banach separation theorems.

---

### DIFF
--- a/Mathlib/Analysis/NormedSpace/Extend.lean
+++ b/Mathlib/Analysis/NormedSpace/Extend.lean
@@ -35,11 +35,10 @@ open RCLike
 
 open ComplexConjugate
 
-variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*} [SeminormedAddCommGroup F]
-
+variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*}
 namespace LinearMap
 
-variable [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
+variable  [AddCommGroup F] [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
 
 /-- Extend `fr : F →ₗ[ℝ] ℝ` to `F →ₗ[𝕜] 𝕜` in a way that will also be continuous and have its norm
 bounded by `‖fr‖` if `fr` is continuous. -/
@@ -85,8 +84,8 @@ theorem extendTo𝕜'_apply_re (fr : F →ₗ[ℝ] ℝ) (x : F) : re (fr.extendT
   simp only [extendTo𝕜'_apply, map_sub, zero_mul, mul_zero, sub_zero,
     rclike_simps]
 
-theorem norm_extendTo𝕜'_apply_sq [NormedSpace 𝕜 F] (fr : F →ₗ[ℝ] ℝ) (x : F) :
-    ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
+theorem norm_extendTo𝕜'_apply_sq [SeminormedAddCommGroup F] [NormedSpace 𝕜 F]
+    (fr : F →ₗ[ℝ] ℝ) (x : F) : ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
   calc
     ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = re (conj (fr.extendTo𝕜' x) * fr.extendTo𝕜' x : 𝕜) := by
       rw [RCLike.conj_mul, ← ofReal_pow, ofReal_re]
@@ -95,7 +94,7 @@ theorem norm_extendTo𝕜'_apply_sq [NormedSpace 𝕜 F] (fr : F →ₗ[ℝ] ℝ
 
 end LinearMap
 
-variable [NormedSpace 𝕜 F]
+variable [SeminormedAddCommGroup F] [NormedSpace 𝕜 F]
 namespace ContinuousLinearMap
 
 variable [NormedSpace ℝ F] [IsScalarTower ℝ 𝕜 F]
--- a/Mathlib/Analysis/NormedSpace/Extend.lean
+++ b/Mathlib/Analysis/NormedSpace/Extend.lean
@@ -84,7 +84,7 @@ theorem extendTo𝕜'_apply_re (fr : F →ₗ[ℝ] ℝ) (x : F) : re (fr.extendT
   simp only [extendTo𝕜'_apply, map_sub, zero_mul, mul_zero, sub_zero,
     rclike_simps]
 
-theorem norm_extendTo𝕜'_apply_sq [SeminormedAddCommGroup F] [NormedSpace 𝕜 F]
+theorem norm_extendTo𝕜'_apply_sq [SeminormedAddCommGroup F]
     (fr : F →ₗ[ℝ] ℝ) (x : F) : ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
   calc
     ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = re (conj (fr.extendTo𝕜' x) * fr.extendTo𝕜' x : 𝕜) := by
--- a/Mathlib/Analysis/NormedSpace/Extend.lean
+++ b/Mathlib/Analysis/NormedSpace/Extend.lean
@@ -38,7 +38,7 @@ open ComplexConjugate
 variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*}
 namespace LinearMap
 
-variable  [AddCommGroup F] [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
+variable [AddCommGroup F] [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
 
 /-- Extend `fr : F →ₗ[ℝ] ℝ` to `F →ₗ[𝕜] 𝕜` in a way that will also be continuous and have its norm
 bounded by `‖fr‖` if `fr` is continuous. -/
--- a/Mathlib/Analysis/NormedSpace/Extend.lean
+++ b/Mathlib/Analysis/NormedSpace/Extend.lean
@@ -84,7 +84,7 @@ theorem extendTo𝕜'_apply_re (fr : F →ₗ[ℝ] ℝ) (x : F) : re (fr.extendT
   simp only [extendTo𝕜'_apply, map_sub, zero_mul, mul_zero, sub_zero,
     rclike_simps]
 
-theorem norm_extendTo𝕜'_apply_sq [SeminormedAddCommGroup F]
+theorem norm_extendTo𝕜'_apply_sq
     (fr : F →ₗ[ℝ] ℝ) (x : F) : ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
   calc
     ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = re (conj (fr.extendTo𝕜' x) * fr.extendTo𝕜' x : 𝕜) := by
--- a/Mathlib/Analysis/NormedSpace/Extend.lean
+++ b/Mathlib/Analysis/NormedSpace/Extend.lean
@@ -35,11 +35,11 @@ open RCLike
 
 open ComplexConjugate
 
-variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*} [SeminormedAddCommGroup F] [NormedSpace 𝕜 F]
+variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*} [SeminormedAddCommGroup F]
 
 namespace LinearMap
 
-variable [Module ℝ F] [IsScalarTower ℝ 𝕜 F]
+variable [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
 
 /-- Extend `fr : F →ₗ[ℝ] ℝ` to `F →ₗ[𝕜] 𝕜` in a way that will also be continuous and have its norm
 bounded by `‖fr‖` if `fr` is continuous. -/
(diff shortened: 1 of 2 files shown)


Mathlib/Analysis/NormedSpace/Extend.lean
```diff
@@ -35,11 +35,10 @@ open RCLike
 
 open ComplexConjugate
 
-variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*} [SeminormedAddCommGroup F] [NormedSpace 𝕜 F]
-
+variable {𝕜 : Type*} [RCLike 𝕜] {F : Type*}
 namespace LinearMap
 
-variable [Module ℝ F] [IsScalarTower ℝ 𝕜 F]
+variable  [AddCommGroup F] [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
```
Ruben-VandeVelde, line 41:
```suggestion
variable [AddCommGroup F] [Module ℝ F] [Module 𝕜 F] [IsScalarTower ℝ 𝕜 F]
```

Mathlib/Analysis/NormedSpace/Extend.lean
```diff
@@ -85,8 +84,8 @@ theorem extendTo𝕜'_apply_re (fr : F →ₗ[ℝ] ℝ) (x : F) : re (fr.extendT
   simp only [extendTo𝕜'_apply, map_sub, zero_mul, mul_zero, sub_zero,
     rclike_simps]
 
-theorem norm_extendTo𝕜'_apply_sq (fr : F →ₗ[ℝ] ℝ) (x : F) :
-    ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
+theorem norm_extendTo𝕜'_apply_sq
+    (fr : F →ₗ[ℝ] ℝ) (x : F) : ‖(fr.extendTo𝕜' x : 𝕜)‖ ^ 2 = fr (conj (fr.extendTo𝕜' x : 𝕜) • x) :=
```
Ruben-VandeVelde, line 88: Why this change?

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+    { toLinearMap := LinearMap.extendTo𝕜' fr
+      cont := show Continuous fun x ↦ (fr x : 𝕜) - (I : 𝕜) * (fr ((I : 𝕜) • x) : 𝕜) by fun_prop }
+  have h fr x : to𝕜 fr x = ((fr x : 𝕜) - (I : 𝕜) * (fr ((I : 𝕜) • x) : 𝕜)) := rfl
+  { toFun := to𝕜
+    map_add' := by intros; ext; simp [h]; ring
+    map_smul' := by intros; ext; simp [h, real_smul_eq_coe_mul]; ring }
+
+@[simp]
+lemma real_of_LinTo𝕜' (g : E →L[ℝ] ℝ) (x : E) :  re ((LinTo𝕜' g) x : 𝕜) = g x := by
+  have h g (x : E) : LinTo𝕜' g x = ((g x : 𝕜) - (I : 𝕜) * (g ((I : 𝕜) • x) : 𝕜)) := rfl
```
Ruben-VandeVelde, line 225: This should be a lemma, possibly generated by `@[simps]` on the definition

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+  letI to𝕜 (fr : (E →L[ℝ] ℝ)) : (E →L[𝕜] 𝕜) :=
+    { toLinearMap := LinearMap.extendTo𝕜' fr
+      cont := show Continuous fun x ↦ (fr x : 𝕜) - (I : 𝕜) * (fr ((I : 𝕜) • x) : 𝕜) by fun_prop }
+  have h fr x : to𝕜 fr x = ((fr x : 𝕜) - (I : 𝕜) * (fr ((I : 𝕜) • x) : 𝕜)) := rfl
+  { toFun := to𝕜
+    map_add' := by intros; ext; simp [h]; ring
+    map_smul' := by intros; ext; simp [h, real_smul_eq_coe_mul]; ring }
+
+@[simp]
+lemma real_of_LinTo𝕜' (g : E →L[ℝ] ℝ) (x : E) :  re ((LinTo𝕜' g) x : 𝕜) = g x := by
```
Ruben-VandeVelde, line 224: I think this should be called `re_LinTo𝕜'` or `re_linTo𝕜'`

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+  have h g (x : E) : LinTo𝕜' g x = ((g x : 𝕜) - (I : 𝕜) * (g ((I : 𝕜) • x) : 𝕜)) := rfl
+  simp only [h , map_sub, ofReal_re, mul_re, I_re, zero_mul, ofReal_im, mul_zero,
+  sub_self, sub_zero]
```
Ruben-VandeVelde, line 227:
```suggestion
    sub_self, sub_zero]
```

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+
+variable [ContinuousSMul ℝ E]
+
+theorem separate_convex_open_set_RCLike  {s : Set E}
```
Ruben-VandeVelde, line 231:
```suggestion
theorem separate_convex_open_set_RCLike {s : Set E}
```

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+    (hs₀ : (0 : E) ∈ s) (hs₁ : Convex ℝ s) (hs₂ : IsOpen s) {x₀ : E} (hx₀ : x₀ ∉ s) :
+    ∃ f : E →L[𝕜] 𝕜, re (f x₀) = 1 ∧ ∀ x ∈ s, re (f x) < 1 := by
+  obtain ⟨g, hg⟩ := separate_convex_open_set hs₀ hs₁ hs₂ hx₀; use LinTo𝕜' g
+  simp only [real_of_LinTo𝕜']; exact hg
```
Ruben-VandeVelde, line 235: I don't think there's a need to put multiple tactics on a line here or below

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+
+theorem geometric_hahn_banach_open_RCLike (hs₁ : Convex ℝ s) (hs₂ : IsOpen s) (ht : Convex ℝ t)
```
Ruben-VandeVelde, line 237: Maybe these should go into the RCLike namespace

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+  use LinTo𝕜' f; simp only [real_of_LinTo𝕜']; exact Exists.intro u h
+
+variable [LocallyConvexSpace ℝ E]
+
+theorem geometric_hahn_banach_compact_closed_RCLike (hs₁ : Convex ℝ s) (hs₂ : IsCompact s)
+    (ht₁ : Convex ℝ t) (ht₂ : IsClosed t) (disj : Disjoint s t) :
+    ∃ (f : E →L[𝕜] 𝕜) (u v : ℝ), (∀ a ∈ s, re (f a) < u) ∧ u < v ∧ ∀ b ∈ t, v < re (f b) := by
+  obtain ⟨g, u, v, h1⟩ := geometric_hahn_banach_compact_closed hs₁ hs₂ ht₁ ht₂ disj; use LinTo𝕜' g
+  simp only [real_of_LinTo𝕜', exists_and_left]; use u; constructor; exact h1.1; use v
+  exact h1.2
```
Ruben-VandeVelde, line 266: Probably this works:
```suggestion
  simp only [real_of_LinTo𝕜', exists_and_left]
  exact ⟨u, h1⟩
```
or this:
```suggestion
  simp only [real_of_LinTo𝕜', exists_and_left]
  exact ⟨u, h1.1, h1.2⟩
```

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,103 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
+  exact ⟨f, by linarith [hs x rfl, ht y rfl]⟩
+
+theorem iInter_halfspaces_eq_RCLike (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
+    ⋂ l : E →L[𝕜] 𝕜, { x | ∃ y ∈ s, re (l x) ≤ re (l y) } = s := by
+  rw [Set.iInter_setOf]
+  refine Set.Subset.antisymm (fun x hx => ?_) fun x hx l => ⟨x, hx, le_rfl⟩
+  by_contra h
+  obtain ⟨l, s, hlA, hl⟩ := geometric_hahn_banach_closed_point_RCLike (𝕜 := 𝕜) hs₁ hs₂ h
+  obtain ⟨y, hy, hxy⟩ := hx l
+  exact ((hxy.trans_lt (hlA y hy)).trans hl).not_le le_rfl
```
Ruben-VandeVelde, line 302:
```suggestion
  exact ((hxy.trans_lt (hlA y hy)).trans hl).false
```

Ruben-VandeVelde (commented):
Some small comments:

Mathlib/Analysis/NormedSpace/Extend.lean
JonBannon: The linter spotted that the typeclass assumptions weren't needed, and it turns out that it was right.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: Tried @[simps] on the definition, and it fails, suggesting @[simps!]. When I tried that, it broke the definition and several things below.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: Somewhat irritatingly, neither works.

JonBannon: Thanks, by the way, for looking over this!!

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: By the way, though, if I give the definition `@[simp]` this works. Is that an ok change?

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
dupuisf: If they're in the `RCLike` namespace, then there is no reason to append `RCLike` to the names of the lemmas.

dupuisf (commented):
I just have a few comments mostly about naming, but it's looking good!

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
```diff
@@ -200,3 +202,114 @@ theorem iInter_halfspaces_eq (hs₁ : Convex ℝ s) (hs₂ : IsClosed s) :
...
   obtain ⟨y, hy, hxy⟩ := hx l
   exact ((hxy.trans_lt (hlA y hy)).trans hl).not_le le_rfl
+
+namespace RCLike
+
+variable [RCLike 𝕜] [TopologicalSpace E] [AddCommGroup E] [TopologicalAddGroup E]
+  [Module 𝕜 E] [Module ℝ E] [ContinuousSMul 𝕜 E] [IsScalarTower ℝ 𝕜 E]
+
+/--Real linear extension of continuous extension of `LinearMap.extendTo𝕜'` -/
+noncomputable def LinTo𝕜' : (E →L[ℝ] ℝ) →ₗ[ℝ] (E →L[𝕜] 𝕜) :=
```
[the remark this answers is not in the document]
JonBannon: Excellent, thanks!

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: Interesting. What is the better style move, here? Should I just remove the `RCLike` from the names of the lemmas and put these (named identically) lemmas, or should I keep them out of the namespace and include the underscored qualification? (I imagine this is sorted by the naming convention somehow, but I'm still internalizing this...)

JonBannon: By the way, I tried removing the `_RCLike` from these names, but Lean then is still identifying the results with those named identically (aside from the `RCLike.` appearing above.) I get structural recursion errors.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: Actually, I just tried @[simps!] again, and it didn't break this time...and I have no idea why. I noticed that this generates a longer list of lemmas to apply in the call than if I stick with the single lemma I introduced. Is it better to use the auto generated lemmas if it leads to using more rewrites or is it better to introduce the new lemma. Thanks again!

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
dupuisf: Those errors come from the fact that what you invoke (say) `geometric_hahn_banach_open` from `RCLike.geometric_hahn_banach_open`, it looks like a recursive call. To make sure that Lean interprets it as invoking the original theorem, you can write `_root_.geometric_hahn_banach_open` to specify you want the one from the root namespace.

As for whether we want the lemmas in the `RCLike` namespace or with `rclike` appended to the names (it should probably be spelled like that in this case), I don't really have a strong preference.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
dupuisf: You can use `@[simps!?]` or `@[simps?]` to make it tell you what lemmas it generated. In this case, I think the one you wrote by hand is better.

JonBannon:
I like the namespace a little better, since these really aren't different
results, but are RCLike versions. I made that fix. Thanks!

JonBannon:
Ok cool. I'll roll this back, then.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
JonBannon: Cool. I rolled those changes back.

Mathlib/Analysis/NormedSpace/HahnBanach/Separation.lean
dupuisf: `exact ⟨u, h1.1, v, h1.2⟩` does the job!

JonBannon:
Awesome! I didn't even *think* to do that. Thank you!

dupuisf (commented):
Thanks, looks good!

bors r+

PR #15417, opened 2024-08-01
[Merged by Bors] - chore: remove some open Classical, part 3

grunweg:
See #15371.

---


Mathlib/Analysis/BoxIntegral/Basic.lean
```diff
@@ -259,6 +260,7 @@ theorem Integrable.of_neg (hf : Integrable I l (-f) vol) : Integrable I l f vol
 theorem integrable_neg : Integrable I l (-f) vol ↔ Integrable I l f vol :=
   ⟨fun h => h.of_neg, fun h => h.neg⟩
 
+open Classical in
 @[simp]
 theorem integral_neg : integral I l (-f) vol = -integral I l f vol :=
   if h : Integrable I l f vol then h.hasIntegral.neg.integral_eq
```
sgouezel, line 266: Better use `classical` in the proof, if it is not  needed for the statement, like:
```suggestion
@[simp]
theorem integral_neg : integral I l (-f) vol = -integral I l f vol := by
  classical
  exact if h : Integrable I l f vol then h.hasIntegral.neg.integral_eq
```

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -165,6 +165,7 @@ theorem inner_single_left (i : ι) (a : G i) (f : lp G 2) : ⟪lp.single 2 i a,
   · subst h; rfl
   · simp
 
+open Classical in
```
sgouezel, line 168: When the decidability assumption is used for the left hand side of the equality in the theorem, can you instead add a decidability assumption?

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -211,6 +212,7 @@ protected theorem hasSum_linearIsometry (f : lp G 2) :
     HasSum (fun i => V i (f i)) (hV.linearIsometry f) :=
   (hV.summable_of_lp f).hasSum
 
+open Classical in
 @[simp]
 protected theorem linearIsometry_apply_single {i : ι} (x : G i) :
```
sgouezel, line 217: ditto

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -222,6 +224,7 @@ protected theorem linearIsometry_apply_single {i : ι} (x : G i) :
   · subst h; simp
   · simp [h]
 
+open Classical in
 protected theorem linearIsometry_apply_dfinsupp_sum_single (W₀ : Π₀ i : ι, G i) :
```
sgouezel, line 228: ditto

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -234,6 +237,7 @@ protected theorem range_linearIsometry [∀ i, CompleteSpace (G i)] :
     -- Porting note: dot notation broken
   refine le_antisymm ?_ ?_
   · rintro x ⟨f, rfl⟩
+    classical
```
sgouezel, line 240: just use one `classical`, at the beginning of the proof?

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -309,6 +314,7 @@ protected theorem IsHilbertSum.hasSum_linearIsometryEquiv_symm (hV : IsHilbertSu
     (w : lp G 2) : HasSum (fun i => V i (w i)) (hV.linearIsometryEquiv.symm w) := by
   simp [IsHilbertSum.linearIsometryEquiv, OrthogonalFamily.hasSum_linearIsometry]
 
+open Classical in
```
sgouezel, line 317: ditto (here and below)

Mathlib/FieldTheory/Finite/Polynomial.lean
```diff
@@ -179,12 +178,14 @@ variable [CommRing K]
 
 noncomputable instance decidableRestrictDegree (m : ℕ) :
     DecidablePred (· ∈ { n : σ →₀ ℕ | ∀ i, n i ≤ m }) := by
+  classical
```
sgouezel, line 181: This one looks wrong: if `classical` is allowed, there is no need to unfold anything, and the instance doesn't make sense. Instead, could you add suitable decidability assumptions for the instance?

Mathlib/FieldTheory/Normal.lean
```diff
@@ -61,6 +61,7 @@ theorem Normal.exists_isSplittingField [h : Normal F K] [FiniteDimensional F K]
     ⟨∏ x, minpoly F (s x), splits_prod _ fun x _ => h.splits (s x),
       Subalgebra.toSubmodule.injective ?_⟩
   rw [Algebra.top_toSubmodule, eq_top_iff, ← s.span_eq, Submodule.span_le, Set.range_subset_iff]
+  classical
```
sgouezel, line 64:
```suggestion
```

sgouezel (commented):
Thanks!

Mathlib/AlgebraicTopology/TopologicalSimplex.lean
```diff
@@ -73,6 +73,7 @@ def toTop : SimplexCategory ⥤ TopCat where
     ext f
     apply toTopObj.ext
     funext i
+    classical
```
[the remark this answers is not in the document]
grunweg: The proof requires `classical`, but that was already present --- reverted.

Mathlib/FieldTheory/Finite/Polynomial.lean
grunweg: Good catch! I'm not sure what a suitable decidability assumption is... I have used `open Classical in` for now (which is less wrong, but perhaps not "right" yet).

grunweg (commented):
Thanks for the careful review. That is exactly the feedback I was looking for! I have addressed all comments.

Mathlib/Analysis/InnerProductSpace/l2Space.lean
```diff
@@ -380,26 +382,31 @@ namespace HilbertBasis
 instance {ι : Type*} : Inhabited (HilbertBasis ι 𝕜 ℓ²(ι, 𝕜)) :=
   ⟨ofRepr (LinearIsometryEquiv.refl 𝕜 _)⟩
 
+open Classical in
 /-- `b i` is the `i`th basis vector. -/
 instance instCoeFun : CoeFun (HilbertBasis ι 𝕜 E) fun _ => ι → E where
   coe b i := b.repr.symm (lp.single 2 i (1 : 𝕜))
 
+open Classical in
```
sgouezel, line 390: use decidability assumptions?

Mathlib/AlgebraicTopology/TopologicalSimplex.lean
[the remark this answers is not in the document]
grunweg: Yes, I did! (I stashed those change temporarily, and forgot to pop my stash.) Added now.

Mathlib/Analysis/InnerProductSpace/l2Space.lean
grunweg: I added them - this makes the proofs require `convert rfl`, though.

sgouezel:
bors r+
Thanks!

PR #15424, opened 2024-08-02
[Merged by Bors] - feat(Valued/LocallyCompact): `totallyBounded_iff_finite_residueField`

pechersky:
---

- [x] depends on: #14768
- [x] depends on: #15425
- [x] depends on: #15426
- [x] depends on: #17561

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15425, opened 2024-08-02
[Merged by Bors] - feat(QuotientGroup,Ideal/Quotient): group is finite if quotient is finite

pechersky:
Does not actually require the subgroup to be normal Also provides a breakdown of the group as the disjoint indexed union of cosets

---


Mathlib/GroupTheory/QuotientGroup.lean
```diff
@@ -132,7 +133,7 @@ theorem eq_iff_div_mem {N : Subgroup G} [nN : N.Normal] {x y : G} :
 
 @[to_additive]
 instance Quotient.commGroup {G : Type*} [CommGroup G] (N : Subgroup G) : CommGroup (G ⧸ N) :=
-  { toGroup := have := N.normal_of_comm; QuotientGroup.Quotient.group N
+  { toGroup := @QuotientGroup.Quotient.group _ _ N N.normal_of_comm
```
erdOne, line 136: Why is this? I think the original version is preferred, because the order of implicit arguments are implementation detail and can easily change.

Mathlib/GroupTheory/QuotientGroup.lean
```diff
@@ -596,6 +597,17 @@ theorem comap_comap_center {H₁ : Subgroup G} [H₁.Normal] {H₂ : Subgroup (G
   simp only [mk'_apply, Subgroup.mem_comap, Subgroup.mem_center_iff, forall_mk, ← mk_mul,
     eq_iff_div_mem, mk_div]
 
+/-- A group is made up of a disjoint union of cosets of a subgroup. -/
+@[to_additive "A additive group is made up of a disjoint union of cosets of an additive
```
erdOne, line 601:
```suggestion
@[to_additive "An additive group is made up of a disjoint union of cosets of an additive
```

This should go in `Coset.lean`. Or just get the next lemma from `groupEquivQuotientProdSubgroup` and ditch this.

Mathlib/RingTheory/Ideal/Quotient.lean
```diff
@@ -348,4 +347,13 @@ theorem map_pi {ι : Type*} [Finite ι] {ι' : Type w} (x : ι → R) (hi : ∀
 
 end Pi
 
+/-- A ring is made up of a disjoint union of cosets of an image. -/
+lemma univ_eq_iUnion_image_add {R : Type*} [Ring R] (I : Ideal R) :
+    (Set.univ (α := R)) = ⋃ x : R ⧸ I, (x.out' + ·) '' I :=
+  QuotientAddGroup.univ_eq_iUnion_image_add I.toAddSubgroup
```
erdOne, line 353: I'm not sure if this is useful. `.out'` is quite hard to work with.

Mathlib/GroupTheory/QuotientGroup.lean
pechersky: This change came from a bad git merge of master, will fix

Mathlib/RingTheory/Ideal/Quotient.lean
pechersky: One can use `Ideal.Quotient.eq` to relate to `out'`. I'll link how I use it in the larger proof.

pechersky: https://github.com/leanprover-community/mathlib4/pull/15424/files#diff-e2953739e89168dcbb0040fe18b7996a54734ee000cfa73813c6a324d3c2f8a9R186-R192

This way, it discusses balls around `i.out` and `y.out`, because it's an easy way to talk about representatives of the residue field (or other quotients) inside the ring itself.

Mathlib/GroupTheory/QuotientGroup.lean
```diff
@@ -631,4 +631,10 @@ noncomputable def fintypeOfKerOfCodom [Fintype g.ker] : Fintype G :=
 noncomputable def fintypeOfDomOfCoker [Normal f.range] [Fintype <| G ⧸ f.range] : Fintype G :=
   fintypeOfKerLeRange _ (mk' f.range) fun x => (eq_one_iff x).mp
 
+@[to_additive]
+lemma _root_.Finite.of_finite_quot_finite_subgroup {H : Subgroup G}
+    (hH : Set.Finite (H : Set G)) (h : Finite (G ⧸ H)) : Finite G := by
```
tb65536, line 636: Is there a reason to prefer `Set.Finite (H : Set G)` over `Finite H`?

pechersky: No reason, ends up being the same. I thought that since a subgroup is often treated more like a set like term rather than a type, that the Set level finite predicate made more sense. While for the ambient group and the quotient group, since they are treated more as types, the Finite predicate made more sense. Which would you prefer? I have no resistance to switching.

tb65536: I think `Finite` is probably preferable, even for subgroups, since it allows you to apply results from the group theory library directly to the subgroup without having to convert between the two notions of finite.

pechersky: I've changed it to take `[Finite H]`, which made the Ideal version require some more assistance. Please let me know if this works for you!

tb65536: Yes, great!

Mathlib/GroupTheory/Coset/Basic.lean
```diff
@@ -686,4 +686,15 @@ noncomputable def preimageMkEquivSubgroupProdSet (s : Subgroup α) (t : Set (α
   left_inv := fun ⟨a, ha⟩ => Subtype.eq <| show _ * _ = a by simp
   right_inv := fun ⟨⟨a, ha⟩, ⟨x, hx⟩⟩ => by ext <;> simp [ha]
 
+/-- A group is made up of a disjoint union of cosets of a subgroup. -/
+@[to_additive "An additive group is made up of a disjoint union of cosets of an additive
+subgroup. -/"]
+lemma univ_eq_iUnion_image_mul (H : Subgroup α) :
+    (Set.univ (α := α)) = ⋃ x : α ⧸ H, (x.out' * ·) '' H := by
```
urkud, line 693: Should we use pointwise `smul` here?

Mathlib/GroupTheory/Coset/Basic.lean
```diff
@@ -686,4 +686,15 @@ noncomputable def preimageMkEquivSubgroupProdSet (s : Subgroup α) (t : Set (α
   left_inv := fun ⟨a, ha⟩ => Subtype.eq <| show _ * _ = a by simp
   right_inv := fun ⟨⟨a, ha⟩, ⟨x, hx⟩⟩ => by ext <;> simp [ha]
 
+/-- A group is made up of a disjoint union of cosets of a subgroup. -/
+@[to_additive "An additive group is made up of a disjoint union of cosets of an additive
+subgroup. -/"]
```
urkud, line 691:
```suggestion
subgroup."]
```

urkud:
Thanks! :tada:
bors merge

PR #15428, opened 2024-08-02
[Merged by Bors] - chore(VitaliFamily): rename `α` → `X`

urkud:
The module docstring uses `X` while the code used `α`.

---


grunweg (commented):
Thank you for doing this. Prior to this PR, `x` was often used as an element of \alpha - this notation is much better!
I have one follow-up question: this file now uses `x : X`, and `a` if there's another variable in `X`. Would you like to rename that one as well? (Apart from that, it's good to go!)

urkud:
> Thank you for doing this. Prior to this PR, `x` was often used as an element of \alpha - this notation is much better! I have one follow-up question: this file now uses `x : X`, and `a` if there's another variable in `X`. Would you like to rename that one as well? (Apart from that, it's good to go!)

`a` was used for variables of type `Set X`. This doesn't follow our naming convention, so renamed to `s` or `t`.

grunweg:
Thanks, looks good to me now.
maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15431, opened 2024-08-02
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 18)

kim-em:
see https://github.com/leanprover-community/mathlib4/pull/15245

### DIFF
--- a/Mathlib/CategoryTheory/Galois/Prorepresentability.lean
+++ b/Mathlib/CategoryTheory/Galois/Prorepresentability.lean
@@ -230,6 +230,8 @@ theorem AutGalois.π_surjective [FiberFunctor F] (A : PointedGaloisObject F) :
   exact eval_section_surjective_of_surjective
     (autGaloisSystem F ⋙ forget _) (autGaloisSystem_map_surjective F) A σ
 
+variable [FiberFunctor F]
+
 section EndAutGaloisIsomorphism
 
 /-!
--- a/Mathlib/CategoryTheory/Galois/Prorepresentability.lean
+++ b/Mathlib/CategoryTheory/Galois/Prorepresentability.lean
@@ -49,8 +49,7 @@ namespace PreGaloisCategory
 open Limits Functor
 
 variable {C : Type u₁} [Category.{u₂} C] [GaloisCategory C]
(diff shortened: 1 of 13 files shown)


Mathlib/CategoryTheory/Galois/Prorepresentability.lean
```diff
@@ -229,6 +222,16 @@ lemma AutGalois.ext {f g : AutGalois F}
   ext A
   exact h A
 
+variable [FiberFunctor F]
+
+/-- `autGalois.π` is surjective for every pointed Galois object. -/
+theorem AutGalois.π_surjective (A : PointedGaloisObject F) :
```
sgouezel, line 228:
```suggestion
/-- `autGalois.π` is surjective for every pointed Galois object. -/
theorem AutGalois.π_surjective [FiberFunctor F] (A : PointedGaloisObject F) :
```

Mathlib/RingTheory/Ideal/Maps.lean
```diff
@@ -59,38 +59,44 @@ theorem mem_map_of_mem (f : F) {I : Ideal R} {x : R} (h : x ∈ I) : f x ∈ map
 theorem apply_coe_mem_map (f : F) (I : Ideal R) (x : I) : f x ∈ I.map f :=
   mem_map_of_mem f x.2
 
-theorem map_le_iff_le_comap : map f I ≤ K ↔ I ≤ comap f K :=
+theorem map_le_iff_le_comap [RingHomClass F R S] : map f I ≤ K ↔ I ≤ comap f K :=
```
sgouezel, line 62: I don't understand this change. There is `[RingHomClass F R S]` in scope, and the statement uses `f : F` and `I : Ideal R` and `K : Ideal S`, so normally `[RingHomClass F R S]` should be pulled in by the new algorithm. If it isn't, it is probably a bug, that should be fixed instead of adding the assumption (unless I'm misunderstanding something). 

Same thing for the next statements.

semorrison: Good catch. This was working around an earlier bug that is now fixed in 4814. (Not pulling in things mentioned via the type of something pulled in.)

PR #15436, opened 2024-08-02
[Merged by Bors] - feat: Add LinearOrderedCommMonoidWithZero and lemmas for unitInterval

PhoenixIra:

---
This PR adds various lemmas about the `unitInterval` that might be of interest. 

It would be nice to also have lemmas similar to [Mathlib/Algebra/Order/GroupWithZero/Unbundled](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/GroupWithZero/Unbundled.html), but instantiation `unitInterval` as `PosMulMono` and similar feels weird as unitInterval has no nonnegative values. Maybe there is a clever way to generalize these lemmas? Or we can provide a similar API for `OrderedCommMonoid`? Unfortunately, that would not help with strictness and reflection lemmas. All of these feel not fitting for this PR, however.

This [zulip thread](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/eq.20one.20of.20le.20mul.20in.20the.20unit) is related.

### DIFF
--- a/Mathlib/Topology/UnitInterval.lean
+++ b/Mathlib/Topology/UnitInterval.lean
@@ -142,7 +142,7 @@ theorem half_le_symm_iff (t : I) : 1 / 2 ≤ (σ t : ℝ) ↔ (t : ℝ) ≤ 1 / 
   rw [coe_symm_eq, le_sub_iff_add_le, add_comm, ← le_sub_iff_add_le, sub_half]
 
 @[simp]
-lemma eq_zero_iff_symm_eq_one {i : I} : σ i = 1 ↔ i = 0 := by
+lemma symm_eq_one_iff_eq_zero {i : I} : σ i = 1 ↔ i = 0 := by
   apply Iff.intro
   · intro h
     rw [← unitInterval.symm_zero] at h
@@ -151,7 +151,7 @@ lemma eq_zero_iff_symm_eq_one {i : I} : σ i = 1 ↔ i = 0 := by
     rw [h, symm_zero]
 
 @[simp]
-lemma eq_one_iff_symm_eq_zero {i : I} : σ i = 0 ↔ i = 1 := by
+lemma symm_eq_zero_iff_eq_one {i : I} : σ i = 0 ↔ i = 1 := by
   apply Iff.intro
   · intro h
     rw [← unitInterval.symm_one] at h
@@ -159,39 +159,39 @@ lemma eq_one_iff_symm_eq_zero {i : I} : σ i = 0 ↔ i = 1 := by
   · intro h
     rw [h, symm_one]
 
-theorem le_symm_if_le_symm {i j : I} (h : i ≤ σ j) : j ≤ σ i := by
+theorem le_symm_of_le_symm {i j : I} (h : i ≤ σ j) : j ≤ σ i := by
   rw [Subtype.mk_le_mk, coe_symm_eq] at h ⊢
   apply le_sub_left_of_add_le
   rw [add_comm]
   exact add_le_of_le_sub_left h
 
 theorem le_symm_iff_le_symm {i j : I} : i ≤ σ j ↔ j ≤ σ i :=
-  ⟨le_symm_if_le_symm, le_symm_if_le_symm⟩
+  ⟨le_symm_of_le_symm, le_symm_of_le_symm⟩
 
-theorem symm_le_if_symm_le {i j : I} (h : σ i ≤ j) : σ j ≤ i := by
+theorem symm_le_of_symm_le {i j : I} (h : σ i ≤ j) : σ j ≤ i := by
   rw [Subtype.mk_le_mk, coe_symm_eq] at h ⊢
   rw [sub_le_iff_le_add, add_comm, ← sub_le_iff_le_add]
   exact h
 
 theorem symm_le_iff_symm_le {i j : I} : σ i ≤ j ↔ σ j ≤ i :=
-  ⟨symm_le_if_symm_le, symm_le_if_symm_le⟩
+  ⟨symm_le_of_symm_le, symm_le_of_symm_le⟩
 
-theorem lt_symm_if_lt_symm {i j : I} (h : i < σ j) : j < σ i := by
+theorem lt_symm_of_lt_symm {i j : I} (h : i < σ j) : j < σ i := by
   rw [Subtype.mk_lt_mk, coe_symm_eq] at h ⊢
   apply lt_sub_left_of_add_lt
   rw [add_comm]
   exact add_lt_of_lt_sub_left h
 
 theorem lt_symm_iff_lt_symm {i j : I} : i < σ j ↔ j < σ i :=
-  ⟨lt_symm_if_lt_symm, lt_symm_if_lt_symm⟩
+  ⟨lt_symm_of_lt_symm, lt_symm_of_lt_symm⟩
 
-theorem symm_lt_if_symm_lt {i j : I} (h : σ i < j) : σ j < i := by
+theorem symm_lt_of_symm_lt {i j : I} (h : σ i < j) : σ j < i := by
   rw [Subtype.mk_lt_mk, coe_symm_eq] at h ⊢
   rw [sub_lt_iff_lt_add, add_comm, ← sub_lt_iff_lt_add]
   exact h
 
 theorem symm_lt_iff_symm_lt {i j : I} : σ i < j ↔ σ j < i :=
-  ⟨symm_lt_if_symm_lt, symm_lt_if_symm_lt⟩
+  ⟨symm_lt_of_symm_lt, symm_lt_of_symm_lt⟩
 
 instance : ConnectedSpace I :=
   Subtype.connectedSpace ⟨nonempty_Icc.mpr zero_le_one, isPreconnected_Icc⟩
--- a/Mathlib/Topology/UnitInterval.lean
+++ b/Mathlib/Topology/UnitInterval.lean
@@ -67,11 +67,19 @@ instance : BoundedOrder I := Set.Icc.boundedOrder zero_le_one
 
 lemma univ_eq_Icc : (univ : Set I) = Icc (0 : I) (1 : I) := Icc_bot_top.symm
 
+@[norm_cast]
 theorem coe_ne_zero {x : I} : (x : ℝ) ≠ 0 ↔ x ≠ 0 :=
   not_iff_not.mpr coe_eq_zero
 
+@[norm_cast]
 theorem coe_ne_one {x : I} : (x : ℝ) ≠ 1 ↔ x ≠ 1 :=
   not_iff_not.mpr coe_eq_one
+
+@[norm_cast]
+lemma coe_pos {x : I} : (0 : ℝ) < x ↔ 0 < x := Iff.rfl
+
+@[norm_cast]
+lemma coe_lt_one {x : I} : (x : ℝ) < 1 ↔ x < 1 := Iff.rfl
 
 instance : Nonempty I :=
   ⟨0⟩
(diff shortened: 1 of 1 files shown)


Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,62 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
 theorem half_le_symm_iff (t : I) : 1 / 2 ≤ (σ t : ℝ) ↔ (t : ℝ) ≤ 1 / 2 := by
   rw [coe_symm_eq, le_sub_iff_add_le, add_comm, ← le_sub_iff_add_le, sub_half]
 
+@[simp]
+lemma eq_zero_iff_sym_eq_one (i : I) : σ i = 1 ↔ i = 0 := by
```
Ruben-VandeVelde, line 145: Why are these `sym` rather than `symm`?

Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,62 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+lemma eq_one_iff_sym_eq_zero (i : I) : σ i = 0 ↔ i = 1 := by
+  apply Iff.intro
+  · intro h
+    rw [← unitInterval.symm_one] at h
+    exact symm_bijective.injective h
+  · intro h
+    rw [h, symm_one]
+
+theorem le_symm_if_le_symm (i j : I) : i ≤ σ j → j ≤ σ i := by
+  intro h
```
Ruben-VandeVelde, line 163:
```suggestion
theorem le_symm_if_le_symm (i j : I) (h : i ≤ σ j) : j ≤ σ i := by
```
Though it seems like i and j are determined by h, so should be implicit as well

Also below

Mathlib/Topology/UnitInterval.lean
Ruben-VandeVelde, line 145: I think we prefer the `i` to be implicit when the conclusion is an iff

Mathlib/Topology/UnitInterval.lean
```diff
@@ -160,6 +224,19 @@ theorem nonneg' {t : I} : 0 ≤ t :=
 theorem le_one' {t : I} : t ≤ 1 :=
   t.2.2
 
+@[simp]
+lemma ne_zero_iff_pos {x : I} : x ≠ 0 ↔ 0 < x := by
```
Ruben-VandeVelde, line 228: Hmm, there's some `pos_iff_ne_zero` lemmas already; you might match those. Does this need to be simp, though? lt doesn't obviously seem simpler than ne

Mathlib/Topology/UnitInterval.lean
```diff
@@ -172,6 +249,25 @@ theorem mul_pos_mem_iff {a t : ℝ} (ha : 0 < a) : a * t ∈ I ↔ t ∈ Set.Icc
...
+  mul_le_mul_left i j h_ij k := by
+    simp only [← Subtype.coe_le_coe, coe_mul]
+    apply mul_le_mul (by rfl) ?_ (nonneg i) (nonneg k)
+    simp [h_ij]
+  le_total := LinearOrder.le_total
+  decidableLE := LinearOrder.decidableLE
+  min_def := LinearOrder.min_def
+  max_def := LinearOrder.max_def
+  compare_eq_compareOfLessAndEq i j := by
+    simp [compare, compareOfLessAndEq]
```
Ruben-VandeVelde, line 265: Replace by the output of `simp? [compare, compareOfLessAndEq]` please

Ruben-VandeVelde (commented):
Thanks, some small comments.

Mathlib/Topology/UnitInterval.lean
PhoenixIra: I think there is argument why lt is more useful, but I won't fight for that. Changed it to `pos_iff_ne_zero` and removed the simp attribute.

Mathlib/Topology/UnitInterval.lean
```diff
@@ -67,12 +67,20 @@ instance : BoundedOrder I := Set.Icc.boundedOrder zero_le_one
...
 
+@[norm_cast]
 theorem coe_ne_one {x : I} : (x : ℝ) ≠ 1 ↔ x ≠ 1 :=
   not_iff_not.mpr coe_eq_one
 
+@[norm_cast]
+lemma coe_pos {x : I} : (0 : ℝ) < x ↔ 0 < x := Iff.rfl
+
+@[norm_cast]
+lemma coe_lt_one {x : I} : (x : ℝ) < 1 ↔ x < 1 := Iff.rfl
```
YaelDillies, line 82:
```suggestion
@[simp, norm_cast] lemma coe_pos {x : I} : (0 : ℝ) < x ↔ 0 < x := Iff.rfl
@[simp, norm_cast] lemma coe_lt_one {x : I} : (x : ℝ) < 1 ↔ x < 1 := Iff.rfl
```
Same above

Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,58 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
 theorem half_le_symm_iff (t : I) : 1 / 2 ≤ (σ t : ℝ) ↔ (t : ℝ) ≤ 1 / 2 := by
   rw [coe_symm_eq, le_sub_iff_add_le, add_comm, ← le_sub_iff_add_le, sub_half]
 
+@[simp]
+lemma symm_eq_one_iff_eq_zero {i : I} : σ i = 1 ↔ i = 0 := by
```
YaelDillies, line 145:
```suggestion
lemma symm_eq_one {i : I} : σ i = 1 ↔ i = 0 := by
```
The RHS is implied

Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,58 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+  apply Iff.intro
+  · intro h
+    rw [← unitInterval.symm_zero] at h
+    exact symm_bijective.injective h
+  · intro h
+    rw [h, symm_zero]
```
YaelDillies, line 151:
```suggestion
  rw [← symm_zero, symm_bijective.injective.eq_iff]
```
and please make `symm_bijective.injective.eq_iff` a lemma called `symm_inj`

Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,58 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+  · intro h
+    rw [h, symm_one]
+
+theorem le_symm_of_le_symm {i j : I} (h : i ≤ σ j) : j ≤ σ i := by
+  rw [Subtype.mk_le_mk, coe_symm_eq] at h ⊢
+  apply le_sub_left_of_add_le
+  rw [add_comm]
+  exact add_le_of_le_sub_left h
+
+theorem le_symm_iff_le_symm {i j : I} : i ≤ σ j ↔ j ≤ σ i :=
```
YaelDillies, line 168:
```suggestion
theorem le_symm_comm {i j : I} : i ≤ σ j ↔ j ≤ σ i :=
```

Mathlib/Topology/UnitInterval.lean
```diff
@@ -133,6 +141,58 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+    rw [← unitInterval.symm_one] at h
+    exact symm_bijective.injective h
+  · intro h
+    rw [h, symm_one]
+
+theorem le_symm_of_le_symm {i j : I} (h : i ≤ σ j) : j ≤ σ i := by
+  rw [Subtype.mk_le_mk, coe_symm_eq] at h ⊢
+  apply le_sub_left_of_add_le
+  rw [add_comm]
+  exact add_le_of_le_sub_left h
```
YaelDillies, line 166: Can you first show `σ j ≤ σ i ↔ i ≤ j` and `σ j < σ i ↔ i < j`?

Mathlib/Topology/UnitInterval.lean
PhoenixIra: If you mean with 'same above' the theorems `coe_ne_zero` and `coe_ne_one` then those are according to the linter not in simp-normal form. Or am I misunderstanding something?

Mathlib/Topology/UnitInterval.lean
PhoenixIra: I changed it to `symm_eq_one_iff`. Are you fine with that as well?

YaelDillies: I would rather remove the `_iff`. It's clearly implied

PhoenixIra: Alright, I removed it.

Mathlib/Topology/UnitInterval.lean
PhoenixIra: `@unitInterval.coe_ne_zero` simp can prove this: `by simp only [@ne_eq, @Set.Icc.coe_eq_zero]`
`@unitInterval.coe_ne_one` simp can prove this: `by simp only [@ne_eq, @Set.Icc.coe_eq_one]`

Mathlib/Topology/UnitInterval.lean
```diff
@@ -67,19 +67,17 @@ instance : BoundedOrder I := Set.Icc.boundedOrder zero_le_one
 
 lemma univ_eq_Icc : (univ : Set I) = Icc (0 : I) (1 : I) := Icc_bot_top.symm
 
-theorem coe_ne_zero {x : I} : (x : ℝ) ≠ 0 ↔ x ≠ 0 :=
-  not_iff_not.mpr coe_eq_zero
-
-theorem coe_ne_one {x : I} : (x : ℝ) ≠ 1 ↔ x ≠ 1 :=
-  not_iff_not.mpr coe_eq_one
+@[norm_cast] theorem coe_ne_zero {x : I} : (x : ℝ) ≠ 0 ↔ x ≠ 0 := not_iff_not.mpr coe_eq_zero
+@[norm_cast] theorem coe_ne_one {x : I} : (x : ℝ) ≠ 1 ↔ x ≠ 1 := not_iff_not.mpr coe_eq_one
```
YaelDillies, line 71:
```suggestion
@[norm_cast] theorem coe_ne_zero {x : I} : (x : ℝ) ≠ 0 ↔ x ≠ 0 := coe_eq_zero.not
@[norm_cast] theorem coe_ne_one {x : I} : (x : ℝ) ≠ 1 ↔ x ≠ 1 := coe_eq_one.not
```

Mathlib/Topology/UnitInterval.lean
```diff
@@ -130,9 +128,40 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+@[simp]
+lemma symm_eq_one {i : I} : σ i = 1 ↔ i = 0 := by
+  rw [← symm_zero, symm_inj]
+
+@[simp]
+lemma symm_eq_zero {i : I} : σ i = 0 ↔ i = 1 := by
+  rw [← symm_one, symm_inj]
+
+@[simp]
+theorem symm_le_symm_iff {i j : I} : σ i ≤ σ j ↔ j ≤ i := by
```
YaelDillies, line 146:
```suggestion
theorem symm_le_symm {i j : I} : σ i ≤ σ j ↔ j ≤ i := by
```

Mathlib/Topology/UnitInterval.lean
```diff
@@ -130,9 +128,40 @@ theorem strictAnti_symm : StrictAnti σ := fun _ _ h ↦ sub_lt_sub_left (α :=
...
+  simp only [symm, Subtype.mk_le_mk, sub_le_sub_iff, add_le_add_iff_left, Subtype.coe_le_coe]
+
+theorem le_symm_comm {i j : I} : i ≤ σ j ↔ j ≤ σ i := by
+  rw [← symm_le_symm_iff, symm_symm]
+
+theorem symm_le_comm {i j : I} : σ i ≤ j ↔ σ j ≤ i := by
+  rw [← symm_le_symm_iff, symm_symm]
+
+@[simp]
+theorem symm_lt_symm_iff {i j : I} : σ i < σ j ↔ j < i := by
```
YaelDillies, line 156:
```suggestion
theorem symm_lt_symm {i j : I} : σ i < σ j ↔ j < i := by
```

Mathlib/Topology/UnitInterval.lean
```diff
@@ -160,6 +189,18 @@ theorem nonneg' {t : I} : 0 ≤ t :=
 theorem le_one' {t : I} : t ≤ 1 :=
   t.2.2
 
+lemma pos_iff_ne_zero {x : I} : 0 < x ↔ x ≠ 0 := by
+  rw [← coe_ne_zero, ← coe_pos, lt_iff_le_and_ne, and_iff_right (nonneg x), ne_comm]
+
+lemma lt_one_iff_ne {x : I} : x < 1 ↔ x ≠ 1 := by
+  rw [← coe_lt_one, ← coe_ne_one, lt_iff_le_and_ne, and_iff_right (le_one x)]
```
YaelDillies, line 196:
```suggestion
lemma lt_one_iff_ne_ {x : I} : x < 1 ↔ x ≠ 1 := lt_top_iff_ne_top
```

Mathlib/Topology/UnitInterval.lean
```diff
@@ -172,6 +213,25 @@ theorem mul_pos_mem_iff {a t : ℝ} (ha : 0 < a) : a * t ∈ I ↔ t ∈ Set.Icc
...
+  le_total := LinearOrder.le_total
+  decidableLE := LinearOrder.decidableLE
+  min_def := LinearOrder.min_def
+  max_def := LinearOrder.max_def
+  compare_eq_compareOfLessAndEq i j := by
+    simp only [compare, compareOfLessAndEq, Subtype.coe_lt_coe]
+    split_ifs
+    all_goals try rfl
+    case pos h_ne h_eq => exact h_ne <| Subtype.coe_inj.mpr h_eq
+    case neg h_eq h_ne => exact h_ne <| Subtype.coe_inj.mp h_eq
```
YaelDillies, line 233: You should be able to just not specify this

Mathlib/Topology/UnitInterval.lean
PhoenixIra: I assume the extra underscore is a typo?

Mathlib/Topology/UnitInterval.lean
PhoenixIra: I should, and I assumed this as well.
However, leaving `compare_eq_compareOfLessAndEq` away leaves me with the unsolved case `mk.mk` with goal `compare ⟨val✝¹, property✝¹⟩ ⟨val✝, property✝⟩ = compareOfLessAndEq ⟨val✝¹, property✝¹⟩ ⟨val✝, property✝⟩`.

Mathlib/Topology/UnitInterval.lean
```diff
@@ -160,6 +189,17 @@ theorem nonneg' {t : I} : 0 ≤ t :=
 theorem le_one' {t : I} : t ≤ 1 :=
   t.2.2
 
+lemma pos_iff_ne_zero {x : I} : 0 < x ↔ x ≠ 0 := by
+  rw [← coe_ne_zero, ← coe_pos, lt_iff_le_and_ne, and_iff_right (nonneg x), ne_comm]
+
```
YaelDillies, line 194:
```suggestion
protected lemma pos_iff_ne_zero {x : I} : 0 < x ↔ x ≠ 0 := bot_lt_iff_ne_bot
```

Mathlib/Topology/UnitInterval.lean
YaelDillies: Whoops yes

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/Topology/UnitInterval.lean
```diff
@@ -172,6 +212,25 @@ theorem mul_pos_mem_iff {a t : ℝ} (ha : 0 < a) : a * t ∈ I ↔ t ∈ Set.Icc
 theorem two_mul_sub_one_mem_iff {t : ℝ} : 2 * t - 1 ∈ I ↔ t ∈ Set.Icc (1 / 2 : ℝ) 1 := by
   constructor <;> rintro ⟨h₁, h₂⟩ <;> constructor <;> linarith
 
+instance : LinearOrderedCommMonoidWithZero I where
```
urkud, line 215: Please reuse the existing `LinearOrder` instance. Something like
```
  __ := inferInstanceAs (LinearOrder I)
```
should work. Otherwise we risk introducing two incompatible `LinearOrder I` structures.

PhoenixIra: Thanks, that worked!

urkud:
Thanks! :tada:
bors merge

PR #15438, opened 2024-08-02
[Merged by Bors] - fix(lint-style): correct error message

grunweg:
---


Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -443,7 +443,7 @@ def lintModules (moduleNames : Array String) (mode : OutputSetting) : IO UInt32
   | OutputSetting.print style =>
     formatErrors allUnexpectedErrors style
     if numberErrorFiles > 0 && mode matches OutputSetting.print _ then
```
adomani, line 445:
```suggestion
    if allUnexpectedErrors.size > 0 && mode matches OutputSetting.print _ then
```
Should this then also be `allUnexpectedErrors.size`?

adomani (commented):
Does this look correct?

Mathlib/Tactic/Linter/TextBased.lean
grunweg: This would be equally fine, but I'd say it doesn't matter.
`numberErrorFiles` is the *number* of files with an error, **allUnexpectedErrors.size** is the number of overall errors.

Do you think using the latter is clearer?

j-loreaux:
@adomani I'm delegating this to you, since you have a question to answer. Otherwise I think this change is fine.

bors d=adomani

Mathlib/Tactic/Linter/TextBased.lean
adomani: Ok, I understand now: leaving what you initially had is fine, thanks!

adomani:
Thanks!

bors merge

PR #15439, opened 2024-08-02
[Merged by Bors] - chore(Geometry/Manifold): split variable statements

grunweg:
Make sure to perform binder annotation updates and declaring new variables in separate variable commands.
This is currently recommended because of leanprover/lean4#2789: mixing these can yield confusing errors.

Found by a linter in #15400.

---


urkud:
Thanks! :tada:
bors merge

PR #15441, opened 2024-08-02
[Merged by Bors] - feat(AddChar): Additive group structure and double dual embedding

YaelDillies:
This PR constructs the additive group structure on `AddChar A G` (a copy of its multiplicative group structure) and leverages it to define the double dual embedding as a morphism `A →+ AddChar (AddChar A M) M`.

From LeanAPAP

---

### DIFF
--- a/Mathlib/Algebra/Group/AddChar.lean
+++ b/Mathlib/Algebra/Group/AddChar.lean
@@ -23,6 +23,12 @@ We also include some constructions specific to the case when `A = R` is a ring; 
 For more refined results of a number-theoretic nature (primitive characters, Gauss sums, etc)
 see `Mathlib.NumberTheory.LegendreSymbol.AddCharacter`.
 
+# Implementation notes
+
+Due to their role as the dual of an additive group, additive characters must themselves be an
+additive group. This contrasts to their pointwise operations which make them a multiplicative group.
+We simply define both the additive and multiplicative group structures and prove them equal.
+
 ## Tags
 
 additive character
@@ -33,6 +39,8 @@ additive character
 -/
 
 open Function Multiplicative
+open Finset hiding card
+open Fintype (card)
 
 section AddCharDef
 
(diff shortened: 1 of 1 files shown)


loefflerd:
I understand the issue you're facing here, but I am not terribly happy with the proposed solution: having a structure with "addition" and "multiplication" operations which are actually two different names for the same operation seems confusing and I fear it would be awkward to maintain.

Have you considered the alternative of defining the dual of `G` in purely additive language, as `G →+ UnitAddCircle`? Then you never step outside the additive world, and the definition of the double-dual embedding is completely transparent.

loefflerd:
> > Have you considered the alternative of defining the dual of `G` in purely additive language, as `G →+ UnitAddCircle`?
> 
> Yes, and it has the following problem: I can't directly apply `f : G →+ UnitAddCircle` at `a : G` to get `f a : ℂ`

`(f a).toCircle`? Or you can define a local notation for `UnitAddCircle.toCircle` (or the bundled version `toCircle_addChar`) and shorten things further (analogous to the use of `𝐞` in the Fourier-transform theory over the reals).

> (or `dft f : (G →+ UnitAddCircle) →+ UnitAddCircle` at `g : G →+ UnitAddCircle` to get `dft f g : ℂ`

I don't think you mean this, because `dft f` won't take values in any kind of circle. But if we let `𝐞` be notation for `UnitAddCircle.toCircle`, you can define `dft f` to be the function sending `h : G →+ UnitAddCircle` to `∑ g : G, 𝐞 (h g) • f g`, with optional signs and/or factors of `G.card` according to taste.

YaelDillies:
I don't think those extra characters everywhere are counterbalancing not needing a one-time duplication of a small part of the API.

I have built all of APAP on this design choice after [this discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Additive.20characters), in particular [this message](https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Additive.20characters/near/378870289) by @MichaelStollBayreuth who agreed with having both notations.

loefflerd:
Fair enough; if the alternatives are worse, then we can probably live with the duplication of additive / multiplicative structures.

MichaelStollBayreuth:
> who agreed with having both notations

My point was that apparently both notations are commonly used (in different parts of mathematics), so it makes sense for Mathlib to support both.

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -236,15 +253,39 @@ end Basic
 
 section toCommMonoid
 
-variable {A M : Type*} [AddMonoid A] [CommMonoid M]
+variable {ι A M : Type*} [AddMonoid A] [CommMonoid M]
 
 /-- When `M` is commutative, `AddChar A M` is a commutative monoid. -/
 instance instCommMonoid : CommMonoid (AddChar A M) := toMonoidHomEquiv.commMonoid
+instance instAddCommMonoid : AddCommMonoid (AddChar A M) := Additive.addCommMonoid
```
MichaelStollBayreuth, line 260:
```suggestion
/-- When `M` is commutative, `AddChar A M` is an additive commutative monoid. -/
instance instAddCommMonoid : AddCommMonoid (AddChar A M) := Additive.addCommMonoid
```
For reasons of symmetry?

MichaelStollBayreuth:
LGTM.
maintainer merge

Mathlib/Algebra/Group/AddChar.lean
```diff
@@ -23,6 +23,12 @@ We also include some constructions specific to the case when `A = R` is a ring;
 For more refined results of a number-theoretic nature (primitive characters, Gauss sums, etc)
 see `Mathlib.NumberTheory.LegendreSymbol.AddCharacter`.
 
+# Implementation notes
+
+Due to their role as the dual of an additive group, additive characters must themselves be an
+additive group. This contrasts to their pointwise operations which make them a multiplicative group.
+We simply define both the additive and multiplicative group structures and prove them equal.
```
jcommelin, line 30:
```suggestion
We simply define both the additive and multiplicative group structures and prove them equal.

For more information on this design decision, see the following zulip thread:
https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Additive.20characters


```

jcommelin (commented):
bors d+

YaelDillies:
bors merge

PR #15443, opened 2024-08-02
feat: Marcinkiewicz-Zygmund inequality

YaelDillies:
Prove the Marcinkiewicz-Zygmund inequality.

The Marcinkiewicz-Zygmund inequality states that, if `X₁, ..., Xₐ ∈ L^p` are independent random
variables of mean zero valued in some inner product space, then the `L^p`-norm of `X₁ + ... + Xₐ` is
at most `Cₚ` times the `L^(p/2)`-norm of `|X₁|² + ... + |Xₐ|²`, where `Cₚ` is a constant depending on
`p` alone.

From LeanAPAP

---

- [x] depends on: #15832
- [x] depends on: #21033

### DIFF
--- a/Mathlib/Probability/MarcinkiewiczZygmund.lean
+++ b/Mathlib/Probability/MarcinkiewiczZygmund.lean
@@ -1,0 +1,199 @@
+/-
+Copyright (c) 2023 Yaël Dillies, Bhavik Mehta. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yaël Dillies, Bhavik Mehta
+-/
+module
+
+public import Mathlib.Data.Nat.Choose.Multinomial
+public import Mathlib.Probability.IdentDistrib
+
+/-!
+# The Marcinkiewicz-Zygmund inequality
+
+This file proves the Marcinkiewicz-Zygmund inequality.
+
+The Marcinkiewicz-Zygmund inequality states that, if `X₁, ... Xₐ ∈ L^p` are independent random
+variables of mean zero valued in some inner product space, then the `L^p`-norm of `X₁ + ... + Xₐ` is
+at most `Cₚ` times the `L^(p/2)`-norm of `|X₁|² + ... + |Xₐ|²`, where `Cₚ` is a constant depending
+on `p` alone.
+
+## Notation
+
+Throughout this file, `A ^^ n` denotes `A × ... × A` (with `n` factors). Formally, this is
+`Fintype.piFinset fun _ : Fin n ↦ A`.
+
+## TODO
+
+We currently only prove the inequality for `p = 2 * m` an even natural number. The general `p` case
+can be obtained from this specific one by nesting of Lp norms.
+-/
+
+public section
+
+open Finset Fintype Function Nat MeasureTheory ProbabilityTheory Real
+open scoped NNReal ENNReal
+
+variable {ι Ω E : Type*} {A : Finset ι} {m n : ℕ} [MeasurableSpace Ω] {μ : Measure Ω}
+  [IsFiniteMeasure μ] [mE : MeasurableSpace E] [NormedAddCommGroup E] [InnerProductSpace ℝ E]
+  {X : ι → Ω → E}
+
+local notation:70 A:70 " ^^ " n:71 => Fintype.piFinset fun _ : Fin n ↦ A
+
+/-- The constant appearing in the Marcinkiewicz-Zygmund inequality for symmetric random variables.
+-/
+noncomputable def marcinkiewiczZygmundSymmConst (p : ℝ≥0) : ℝ := (p / 2) ^ (p / 2 : ℝ)
+
+/-- The **Marcinkiewicz-Zygmund inequality** for symmetric random variables, with a slightly better
+constant than `marcinkiewicz_zygmund`. -/
+theorem marcinkiewicz_zygmund_symmetric (iIndepFun_X : iIndepFun X μ)
+    (identDistrib_neg_X : ∀ i, IdentDistrib (X i) (-X i) μ μ)
+    (memLp_X : ∀ i ∈ A, MemLp (X i) (2 * m) μ) :
+    ∫ ω, ‖∑ i ∈ A, X i ω‖ ^ (2 * m) ∂μ ≤
+      marcinkiewiczZygmundSymmConst (2 * m) * ∫ ω, (∑ i ∈ A, ‖X i ω‖ ^ 2) ^ m ∂μ := by
+  have : DecidableEq ι := Classical.decEq _
+  -- Turn the `L^p` assumption on the `X i` into various integrability conditions.
+  have integrable_prod_norm_X I (hI : I ∈ A ×ˢ A ^^ m) :
+      Integrable (fun ω ↦ ∏ k, ‖X (I k).1 ω‖ * ‖X (I k).2 ω‖) μ := by
+    obtain rfl | hm := eq_or_ne m 0
+    · simp
+    simp_rw [Finset.prod_mul_distrib]
+    rw [← memLp_one_iff_integrable]
+    have aux : (∑ _k : Fin m, (2 * (m : ℝ≥0∞))⁻¹)⁻¹ = 2 := by
+      rw [ENNReal.mul_inv (a := 2) (.inl <| by norm_num) (.inl <| by norm_num)]
+      simp [hm, mul_comm, ← mul_assoc, ENNReal.mul_inv_cancel]
+    refine .mul' (p := 2) (q := 2) ?_ ?_ <;>
+    · rw [← aux]
+      exact .prod' fun k _ ↦ (memLp_X _ <| by simp_all).norm
+  have integrable_prod_inner_X I (hI : I ∈ A ×ˢ A ^^ m) :
+    Integrable (fun ω ↦ ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω)) μ := sorry
+  -- Call a family of indices `i₁, ..., iₙ, j₁, ..., jₙ` *even* if each `i ∈ A` appears an even
+  -- number of times among the `2n` indices.
+  let EvenIndex (I : Fin m → ι × ι) : Prop :=
+    ∀ i ∈ A, Even (#{k | (I k).1 = i} + #{k | (I k).2 = i})
+  -- Now, let the calculation begin...
+  calc
+    ∫ ω, ‖∑ i ∈ A, X i ω‖ ^ (2 * m) ∂μ
+    -- Expand out the power of the sum into a sum over families of indices
+    -- `i₁, ..., iₙ, j₁, ..., jₙ` of `∏ k, ⟨X iₖ, X jₖ⟩`. Push the integral inside the sum.
+    _ = ∑ I ∈ A ×ˢ A ^^ m, ∫ ω, ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ := by
+      simp_rw [pow_mul, ← real_inner_self_eq_norm_sq, sum_inner, inner_sum, ← sum_product',
+        Finset.sum_pow', integral_finsetSum _ integrable_prod_inner_X]
+    -- Show that the terms coming from odd families of indices `i₁, ..., iₙ, j₁, ..., jₙ` integrate
+    -- to zero.
+    _ = ∑ I ∈ A ×ˢ A ^^ m with EvenIndex I, ∫ ω, ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ := by
+      rw [Finset.sum_filter_of_ne]
+      -- Assume that `I = (i₁, ..., iₙ, j₁, ..., jₙ)` is an odd family.
+      -- Say `i` appears an odd number of times in it.
+      rintro I hI hI' i hi
+      contrapose! hI'
+      replace hI' : Odd (#{k | (I k).1 = i} + #{k | (I k).2 = i}) := by simpa using hI'
+      -- Let `Y` be the family of random variables `X` where `X i` has been replaced by `-X i`.
+      let Y : ι → Ω → E := update X i (-X i)
+      -- By the assumption that `X i` is symmetric, we get that `X j` and `Y j` are identically
+      -- distributed for all `j`.
+      have identDistrib_X_Y j : IdentDistrib (X j) (Y j) μ μ := by
+        obtain rfl | hji := eq_or_ne j i
+        · simpa [Y] using identDistrib_neg_X _
+        · simpa [Y, hji] using .refl (identDistrib_neg_X _).aemeasurable_fst
+      -- To show that `𝔼 ∏ k, ⟨X iₖ, X jₖ⟩ = 0`, we will show
+      -- `𝔼 ∏ k, ⟨X iₖ, X jₖ⟩ = 𝔼 ∏ k, ⟨Y iₖ, Y jₖ⟩ = -𝔼 ∏ k, ⟨X iₖ, X jₖ⟩`.
+      rw [← neg_eq_self, ← integral_neg, eq_comm]
+      calc
+        -- `𝔼 ∏ k, ⟨X iₖ, X jₖ⟩ = 𝔼 ∏ k, ⟨Y iₖ, Y jₖ⟩` because `𝔼 ∏ k, ⟨X iₖ, X jₖ⟩` and
+        -- `∏ k, ⟨Y iₖ, Y jₖ⟩` are identically distributed.
+        ∫ ω, ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ
+        _ = ∫ ω, ∏ k, inner ℝ (Y (I k).1 ω) (Y (I k).2 ω) ∂μ := by
+          refine IdentDistrib.integral_eq ?_
+          sorry -- TODO: Upstream result from PFR
+        -- `𝔼 ∏ k, ⟨Y iₖ, Y jₖ⟩ = -𝔼 ∏ k, ⟨X iₖ, X jₖ⟩` by the assumption that `i` appears an odd
+        -- number of times in `I`.
+        _ = ∫ ω, -∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ := by
+          congr with ω
+          calc
+            ∏ k, inner ℝ (Y (I k).1 ω) (Y (I k).2 ω)
+            _ = ∏ k, (if (I k).1 = i then -1 else 1) * (if (I k).2 = i then -1 else 1) *
+                inner ℝ (X (I k).1 ω) (X (I k).2 ω) := by
+              congr! with k; split_ifs with hk₁ hk₂ hk₂ <;> simp [hk₁, hk₂, Y]
+            _ = -∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) := by
+              rw [prod_mul_distrib, prod_mul_distrib]
+              simp [prod_ite, ← pow_add, hI']
+    -- Upper bound the sum by its absolute value and push the absolute value inside.
+    _ ≤ |∑ I ∈ A ×ˢ A ^^ m with EvenIndex I, ∫ ω, ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ| :=
+      le_abs_self _
+    _ ≤ ∑ I ∈ A ×ˢ A ^^ m with EvenIndex I, |∫ ω, ∏ k, inner ℝ (X (I k).1 ω) (X (I k).2 ω) ∂μ| :=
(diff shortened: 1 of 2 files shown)


Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,274 @@
...
+  simp only [step_four, mul_ite, mul_zero, mul_one, ← sum_filter, ← sum_mul]
+  rw [mul_comm, mul_div_mul_left]
+  swap; · positivity
+  simpa only [even_iff_two_dvd, ← map_nsmul_piAntidiag_univ _ two_ne_zero, sum_map,
+    Function.Embedding.coeFn_mk] using end_step hm' hA
+
+/-- The **Marcinkiewicz-Zygmund inequality** for real-valued functions, with a slightly easier to
+bound constant than `Real.marcinkiewicz_zygmund'`.
+
+Note that `RCLike.marcinkiewicz_zygmund` is another  -/
```
YaelDillies, line 214:
```suggestion
Note that `RCLike.marcinkiewicz_zygmund` is another version that works for both `ℝ` and `ℂ` at the
expense of a slightly worse constant. -/
```

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,275 @@
...
+  rw [← mul_pow, ← mul_pow, ← mul_pow, mul_assoc, mul_sum _ _ (2 : ℝ)]
+  gcongr with i
+  exact add_sq_le.trans_eq (by simp)
+
+private lemma step_eight {f : α → ℝ} {a b : Fin n → α} :
+    m ^ m * 2 ^ m * (∑ i, (f (a i) ^ 2 + f (b i) ^ 2) : ℝ) ^ m ≤
+      m ^ m * 2 ^ (m + (m - 1)) *
+        ((∑ i, f (a i) ^ 2) ^ m + (∑ i, f (b i) ^ 2) ^ m) := by
+  rw [pow_add, ← mul_assoc _ _ (2 ^ _), mul_assoc _ (2 ^ (m - 1)), sum_add_distrib]
+  gcongr
```
b-mehta, line 161: I know this isn't enforced, but I feel pretty strongly that `gcongr` should always have an explicit pattern to make it more robust. At some point I'll make a stronger case for this on zulip, but in the mean time would you mind squeezing the `gcongr`s in this PR?

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,275 @@
...
+Authors: Yaël Dillies, Bhavik Mehta
+-/
+import Mathlib.Analysis.MeanInequalitiesPow
+import Mathlib.Data.Nat.Choose.Multinomial
+import Mathlib.Tactic.Positivity.Finset
+
+/-!
+# The Marcinkiewicz-Zygmund inequality
+
+This file proves the Marcinkiewicz-Zygmund inequality. This is the statement that, for a real- or
```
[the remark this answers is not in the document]
YaelDillies: Yes we did

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
YaelDillies: I don't really fancy duplicating all the LHS or RHS every time I use `gcongr`, especially since all `gcongr`s here are maximal (in the sense that the goals it leaves are not of the form `r (f a) (f b)` where `r` is a concrete relation and `f` a concrete operation) so I will wait until you have convinced the rest of the community that this is worth doing

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,275 @@
...
+
+-- works with this
+-- private lemma step_two_aux' {β γ : Type*} [AddCommMonoid β] [CommRing γ]
+--   (f : (Fin n → α) → (Fin n → γ)) (ε : Fin n → γ)
+--   (hε : ∀ i, ε i = -1 ∨ ε i = 1) (g : (Fin n → γ) → β) :
+--   ∑ a b in A ^^ n, g (ε * (f a - f b)) = ∑ a b in A ^^ n, g (f a - f b) :=
+-- feels like could generalise more...
+-- the key point is that you combine the double sums into a single sum, and do a pair swap
+-- when the corresponding ε is -1
+-- but the order here is a bit subtle (ie this explanation is an oversimplification)
```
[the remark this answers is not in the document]
YaelDillies: You did. I would never have written that because I would have found the generalisation instead 😉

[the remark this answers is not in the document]
b-mehta: Okay cool, it's nonsense, let's delete it. The generalisation given here doesn't work.

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
b-mehta: Eh, we really should have `gcongr?`

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,275 @@
...
+      simp only [mul_add, sum_add_distrib, sum_const, nsmul_eq_mul, ← mul_sum]
+      rw [← mul_add, ← two_mul, ← mul_assoc 2, ← mul_assoc 2, mul_right_comm 2, ← _root_.pow_succ',
+        add_assoc, Nat.sub_add_cancel hm, pow_add, ← mul_pow, ← mul_pow, card_piFinset, prod_const,
+        card_univ, Fintype.card_fin, Nat.cast_pow, mul_div_cancel_left₀]
+      norm_num
+      positivity
+
+namespace Real
+
+/-- The **Marcinkiewicz-Zygmund inequality** for real-valued functions, with a slightly better
```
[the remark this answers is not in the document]
YaelDillies: I've added a todo in the module doc

Mathlib/Analysis/Complex/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,248 @@
...
+# The Marcinkiewicz-Zygmund inequality
+
+This file proves the Marcinkiewicz-Zygmund inequality. This is the statement that, for a real- or
+complex-valued random variable `f`, the $$L^p$$-norm of the sum of `n` samples of `f` is bounded
+by the $$L^p$$-norm of `f` up to some multiplicative constant.
+
+## TODO
+
+We currently only prove the inequality for `p = 2 * m` an even natural number. The general `p` case
+can be obtained from this specific one by nesting of LP norms.
```
b-mehta, line 20:
```suggestion
can be obtained from this specific one by nesting of Lp norms.
```

hrmacbeth:
Does this work for Euclidean space, by expressing the norm-square of a vector as the sum of the squares of its components?  You do this for the real+complex decomposition of a complex number, but I don't immediately see anything special to two summands.  (Of course, the constants would be adjusted.)

If that works, then you can extend to a finite-dimensional inner product space by using `stdOrthonormalBasis`.

YaelDillies:
I asked the people around me (Bhavik, Thomas Bloom, my PhD advisor), and we all agree that your generalisation should work, but also that we have no use for it. I am therefore leaving a TODO.

hrmacbeth:
I pushed a cheap generalization to inner product spaces (cheap because I kept the dimension restriction, so your original proof goes through without change).  OK?

hrmacbeth:
... at which point I figured I might as well do the general case.  Are you willing to keep it part of the PR?

YaelDillies:
Ahah, that's great! I was not willing to generalise because I have so many other things to push from APAP, but if you have already done it I won't complain 😅 Thanks a lot!

Mathlib/Analysis/InnerProductSpace/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,286 @@
...
+    ∑ a in A ^^ n, ‖∑ i, f (a i)‖ ^ (2 * m) ≤
+      (4 * finrank ℝ 𝕜 * m) ^ m * card n ^ (m - 1) * ∑ a in A ^^ n, ∑ i, ‖f (a i)‖ ^ (2 * m) := by
+  obtain ht | ht := Nat.eq_zero_or_pos (finrank ℝ 𝕜)
+  · rw [FiniteDimensional.finrank_zero_iff] at ht
+    have : 2 * m ≠ 0 := by positivity
+    simp [norm_of_subsingleton, zero_pow this]
+  let b := stdOrthonormalBasis ℝ 𝕜; clear_value b
+  set B := A ^^ n
+  simp only [pow_mul, ← b.repr.norm_map, PiLp.norm_sq_eq_of_L2, map_sum, norm_eq_abs, sq_abs,
+    ge_iff_le, Fintype.sum_apply (γ := n)]
```
b-mehta, line 227: This looks like another case of `simp` giving `ge_iff_le` for no good reason

YaelDillies:
I have decided to move the new file to a new folder `Analysis.Discrete` because I have more discrete analysis in the making and Marcinkiewicz-Zygmund is usually classified as a discrete analysis result.

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,286 @@
...
+-/
+import Mathlib.Analysis.InnerProductSpace.PiL2
+import Mathlib.Analysis.MeanInequalitiesPow
+import Mathlib.Data.Nat.Choose.Multinomial
+import Mathlib.Tactic.Positivity.Finset
+
+/-!
+# The Marcinkiewicz-Zygmund inequality
+
+This file proves the Marcinkiewicz-Zygmund inequality. This is the statement that, for a random
```
sgouezel, line 14: I am really surprised by this version: for me, one of the important points of Marcinkiewicz-Zygmund is that you can take random variables which are not identically distributed. Also, you are working with the specific case of a fintype, while the inequality in true for a general probability space. I'd rather deduce the particular version from the general case (which is not really harder, although I'd rather get the more precise Rosenthal or Burkholder inequalities), but if you want to focus on the discrete space you should emphasize that in the file-level docstring.

b-mehta: Indeed this is a special case, but it's precisely the special case which arises as a very early dependence in our work. Would you mind if we stayed with this version for now, with a comment to generalise it in the way you describe?

sgouezel: I absolutely don't mind, as long as it's explained clearly in the file-level docstring.

YaelDillies: Tried adding something

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,288 @@
...
+variable `f` valued in a real inner product space, the $$L^p$$-norm of the sum of `n` samples of `f`
+is bounded by the $$L^p$$-norm of `f` up to some multiplicative constant.
+
+## TODO
+
+* We currently only prove the inequality for `p = 2 * m` an even natural number. The general `p`
+  case can be obtained from this specific one by nesting of Lp norms.
+* We currently only prove the case of a uniformly distributed discrete random variable. This needs
+  to be generalised! (and when it is, this file should move out of the `Algebra.Discrete`) folder.
```
Parcly-Taxel, line 23:
```suggestion
  to be generalised! (and when it is, this file should move out of the `Analysis.Discrete`) folder.
```

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,290 @@
...
+  rw [pow_add, ← mul_assoc _ _ (2 ^ _), mul_assoc _ (2 ^ (m - 1)), sum_add_distrib]
+  gcongr
+  exact add_pow_le (by positivity) (by positivity) m
+
+variable [DecidableEq n]
+
+local notation:70 s:70 " ^^ " n:71 => Fintype.piFinset fun _ : n ↦ s
+
+private lemma step_one (hA : A.Nonempty) (f : α → ℝ) (a : n → α)
+    (hf : ∀ i, ∑ a in A ^^ n, f (a i) = 0) :
```
sgouezel, line 49:
```suggestion
    (hf : ∀ i, ∑ a ∈ A ^^ n, f (a i) = 0) :
```
The `in` notation is deprecated, can you use `∈` instead throughout?

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,290 @@
...
+    _ = _ := by
+      simp only [mul_add, sum_add_distrib, sum_const, nsmul_eq_mul, ← mul_sum]
+      rw [← mul_add, ← mul_add, ← two_mul, card_piFinset, prod_const, Finset.card_univ,
+        Nat.cast_pow, mul_div_cancel_left₀ _ (by positivity), ← mul_assoc, mul_assoc _ _ 2,
+        mul_comm 4, mul_pow, ← pow_succ, add_assoc, Nat.sub_add_cancel hm, ← two_mul, pow_mul]
+      norm_num
+
+namespace Real
+
+/-- The **Marcinkiewicz-Zygmund inequality** for real-valued functions, with a slightly better
```
sgouezel, line 164: Can you spell out what the inequality says, in layman terms -- in particular since there is no explicit independence in the statement, while it's clearly the intuition to have in mind? Same thing for the other main theorems.

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,290 @@
...
+  rw [mul_comm, mul_div_mul_left _ _ (by positivity)]
+  simpa only [even_iff_two_dvd, ← map_nsmul_piAntidiag_univ _ two_ne_zero, sum_map,
+    Function.Embedding.coeFn_mk] using end_step hm' hA
+
+/-- The **Marcinkiewicz-Zygmund inequality** for real-valued functions, with a slightly easier to
+bound constant than `Real.marcinkiewicz_zygmund'`.
+
+Note that `InnerProductSpace.marcinkiewicz_zygmund` and `RCLike.marcinkiewicz_zygmund` are other
+versions that works for more general target spaces, at the expense of a slightly worse constant. -/
+theorem marcinkiewicz_zygmund (hm : m ≠ 0) (f : α → ℝ) (hf : ∀ i, ∑ a in A ^^ n, f (a i) = 0) :
```
sgouezel, line 194: Isn't it also true for `m = 0`? Same thing below.

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,290 @@
...
+* We currently only prove the inequality for `p = 2 * m` an even natural number. The general `p`
+  case can be obtained from this specific one by nesting of Lp norms.
+* We currently only prove the case of a uniformly distributed discrete random variable. This needs
+  to be generalised! (and when it is, this file should move out of the `Analysis.Discrete`) folder.
+-/
+
+open Finset Fintype Nat Real
+variable {α : Type*} {n : Type*} [Fintype n] {A : Finset α} {m : ℕ}
+
+private lemma step_seven {f : α → ℝ} {a b : n → α} :
```
sgouezel, line 29: To help your reader follow the argument, it would help a lot to have a short proof sketch in the file-level docstring, mentioning where the various auxiliary lemmas are relevant.

Mathlib/Analysis/InnerProductSpace/Basic.lean
```diff
@@ -2257,6 +2257,27 @@ instance : InnerProductSpace ℝ ℂ := InnerProductSpace.complexToReal
...
 end RCLikeToReal
 
+/-- An `RCLike` field is a real inner product space. This is not registered as an instance since it
+creates problems with the case `𝕜 = ℝ`, but in can be used in a proof to obtain a real inner product
+space structure on an `RCLike` field.
+
+This is different from the inner product space structure obtained from
+`InnerProductSpace.rclikeToReal 𝕜 𝕜` because of a diamond in the `ℝ`-normed space instances:
+`NormedSpace.restrictScalars ℝ 𝕜 𝕜` ≠ `NormedAlgebra.toNormedSpace' (𝕜 := ℝ) (𝕜' := 𝕜)`. -/
+noncomputable def RCLike.innerProductSpaceReal : InnerProductSpace ℝ 𝕜 :=
```
sgouezel, line 2267: can you use `where` instead?

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
sgouezel: No, it's not, when `n` is empty. Sorry for the noise.

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,290 @@
...
+import Mathlib.Analysis.InnerProductSpace.PiL2
+import Mathlib.Data.Nat.Choose.Multinomial
+import Mathlib.Tactic.Positivity.Finset
+
+/-!
+# The Marcinkiewicz-Zygmund inequality
+
+This file proves the Marcinkiewicz-Zygmund inequality. This is the statement that, for a random
+variable `f` valued in a real inner product space, the $$L^p$$-norm of the sum of `n` samples of `f`
+is bounded by the $$L^p$$-norm of `f` up to some multiplicative constant.
```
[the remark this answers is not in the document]
YaelDillies: I am not sure I understand this quibble. Isn't the point to get *any* constant? At least, [Wikipedia](https://en.wikipedia.org/wiki/Marcinkiewicz%E2%80%93Zygmund_inequality) only claims the result with a constant, without specifying anything about its size and relationship to the central limit theorem.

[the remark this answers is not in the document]
b-mehta: I think the quibble here is exactly the opposite! The usual MZ inequality allows any constant, but in our specific application the shape of the constant is actually important. And so what LeanAPAP needs as "the MZ inequality" is a bit different from the usual formulation, since we require better control on the constants. And then the docstring should say a bit more about what the control needed is

YaelDillies:
Thank you Heather and Sébastien for the reviews. I have decided that the current proof was too ugly and specific to be worth spending more time on it, so I have started formalising the general random variable version.

Along the way, I've made another big change: Instead of generalising to inner product spaces by taking a basis (bootstrapping), I rewrote the entire proof to handle products of inner products rather than products. This makes the proof a bit more complicated in places, but overall seems to work nicer AND gives a constant that does not depend on the dimension of the space.

There are still a few sorries left (10 at the time of writing), but none of them should be particularly challenging.

Mathlib/Analysis/Discrete/MarcinkiewiczZygmund.lean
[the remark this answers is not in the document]
b-mehta: (imo, your new docstring satisfies this)

Mathlib/Probability/MarcinkiewiczZygmund.lean
```diff
@@ -0,0 +1,187 @@
...
+    (identDistrib_neg_X : ∀ i, IdentDistrib (X i) (-X i) μ μ)
+    (memℒp_X : ∀ i ∈ A, Memℒp (X i) (2 * m) μ) :
+    ∫ ω, ‖∑ i ∈ A, X i ω‖ ^ (2 * m) ∂μ ≤
+      marcinkiewiczZygmundSymmConst (2 * m) * ∫ ω, (∑ i ∈ A, ‖X i ω‖ ^ 2) ^ m ∂μ := by
+  have : DecidableEq ι := Classical.decEq _
+  -- Turn the `L^p` assumption on the `X i` into various integrability conditions.
+  have integrable_prod_norm_X I (hI : I ∈ A ×ˢ A ^^ m) :
+    Integrable (fun ω ↦ ∏ k, ‖X (I k).1 ω‖ * ‖X (I k).2 ω‖) μ := sorry
+  have integrable_prod_inner_X I (hI : I ∈ A ×ˢ A ^^ m) :
+    Integrable (fun ω ↦ ∏ k, inner (𝕜 := ℝ) (X (I k).1 ω) (X (I k).2 ω)) μ := sorry
```
YaelDillies, line 56: @sgouezel, any idea for how to prove these integrability conditions swiftly? It's hopefully a simple application of Hölder

YaelDillies:
I will reopen this once I am done generalising to arbitrary measures.

PR #15445, opened 2024-08-02
[Merged by Bors] - feat(Analysis.Normed.Ring.IsPowMulFaithful): prove eq_of_powMul_faithful

mariainesdff:
We add `AlgebraNorm` and `MulAlgebraNorm`.

We prove Bosch-Güntzer-Remmert, Proposition 3.1.5/1: if `R` is a normed
commutative ring and `f₁` and `f₂` are two power-multiplicative `R`-algebra norms on `S`, then if
`f₁` and `f₂` are equivalent on every subring `R[y]` for `y : S`, it follows that `f₁ = f₂`.

---


Mathlib/Analysis/Normed/Ring/Seminorm.lean
```diff
@@ -151,6 +151,26 @@ end Ring
...
+    (a : α) : ∀ {n : ℕ}, n ≠ 0 → f (a ^ n) ≤ f a ^ n
+  | 0, h => absurd rfl h
+  | 1, _ => by simp only [pow_one, le_refl]
+  | n + 2, _ => by
+    simp only [pow_succ _ (n + 1)];
+      exact
+        le_trans (map_mul_le_mul f _ a)
+          (mul_le_mul_of_nonneg_right (map_pow_le_pow _ _ n.succ_ne_zero) (apply_nonneg f a))
+
+/-- If `f` is a ring seminorm on `a` with `f 1 ≤ `, then `∀ (n : ℕ), f (a ^ n) ≤ f a ^ n`. -/
```
adamtopaz, line 165:
```suggestion
/-- If `f` is a ring seminorm on `a` with `f 1 ≤ 1`, then `∀ (n : ℕ), f (a ^ n) ≤ f a ^ n`. -/
```

Mathlib/Analysis/Normed/Ring/Seminorm.lean
```diff
@@ -151,6 +151,26 @@ end Ring
 
 end RingSeminorm
 
+/-- If `f` is a ring seminorm on `a`, then `∀ {n : ℕ}, n ≠ 0 → f (a ^ n) ≤ f a ^ n`. -/
+theorem map_pow_le_pow {F α : Type _} [Ring α] [FunLike F α ℝ] [RingSeminormClass F α ℝ] (f : F)
```
adamtopaz, line 155: It's better to use `Type*` as opposed to `Type _` , and not just here but most other places in this PR.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
```diff
@@ -0,0 +1,120 @@
...
+  ext x
+  exact le_antisymm (contraction_of_isPowMul_of_boundedWrt g hfpm hge x)
+    (contraction_of_isPowMul_of_boundedWrt f hgpm hle x)
+
+variable {R S : Type _} [NormedCommRing R] [CommRing S] [Algebra R S]
+
+/-- The restriction of a power-multiplicative function to a subalgebra is power-multiplicative. -/
+theorem IsPowMul.restriction (A : Subalgebra R S) {f : S → ℝ} (hf_pm : IsPowMul f) :
+    IsPowMul fun x : A => f x.val := fun x n hn => by
+  simpa [SubsemiringClass.coe_pow] using hf_pm (↑x) hn
```
adamtopaz, line 99: This should probably go somewhere earlier as well...

adamtopaz (commented):
Overall this looks good. I suggested a few things to rearrange in the comments. Also, would you mind splitting off the prerequisites of the main results into a separate PR? (By "prerequisites" I mean essentially everything you do outside of the `IsPowMulFaithful.lean` file.)

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
```diff
@@ -0,0 +1,120 @@
...
+def RingHom.IsBounded {α : Type _} [SeminormedRing α] {β : Type _} [SeminormedRing β]
+    (f : α →+* β) : Prop :=
+  ∃ C : ℝ, 0 < C ∧ ∀ x : α, norm (f x) ≤ C * norm x
+
+/-- A ring homomorphism `f : α →+* β` is bounded with respect to the functions `nα : α → ℝ` and
+  `nβ : β → ℝ` if there exists a positive constant `C` such that for all `x` in `α`,
+  `nβ (f x) ≤ C * nα x`. -/
+def RingHom.IsBoundedWrt {α : Type _} [Ring α] {β : Type _} [Ring β] (nα : α → ℝ) (nβ : β → ℝ)
+    (f : α →+* β) : Prop :=
+  ∃ C : ℝ, 0 < C ∧ ∀ x : α, nβ (f x) ≤ C * nα x
```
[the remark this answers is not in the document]
mariainesdff: I have moved them to `Analysis.Normed.Field.Basic` and `Data.Real.Basic`, respectively.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
mariainesdff: I moved it to `Analysis.Normed.Field.Basic`.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-!
+# Equivalent power-multiplicative norms
+
+In this file, we prove [BGR, Proposition 3.1.5/1][bosch-guntzer-remmert]: if `R` is a normed
+commutative ring and `f₁` and `f₂` are two power-multiplicative `R`-algebra norms on `S`, then if
+`f₁` and `f₂` are equivalent on every subring `R[y]` for `y : S`, it follows that `f₁ = f₂`.
+
+## Main Definitions
+* `AlgebraNorm.restriction`: The restriction of an algebra norm to a subalgebra.
```
jcommelin, line 17: It seems this moved the the `Basic` file.

Mathlib/Analysis/Normed/Algebra/Norm.lean
```diff
@@ -0,0 +1,186 @@
...
+/-- An `R`-algebra norm such that `f 1 = 1` extends the norm on `R`. -/
+theorem extends_norm' (hf1 : f 1 = 1) (a : R) : f (a • (1 : S)) = ‖a‖ := by
+  rw [← mul_one ‖a‖, ← hf1]; exact f.smul' _ _
+
+/-- An `R`-algebra norm such that `f 1 = 1` extends the norm on `R`. -/
+theorem extends_norm (hf1 : f 1 = 1) (a : R) : f (algebraMap R S a) = ‖a‖ := by
+  rw [Algebra.algebraMap_eq_smul_one]; exact extends_norm' hf1 _
+
+/-- The restriction of an algebra norm to a subalgebra. -/
+def restriction (A : Subalgebra R S) (f : AlgebraNorm R S) : AlgebraNorm R A where
```
jcommelin, line 92: Would it make sense to generalize this to any tower of algebras?

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
```diff
@@ -0,0 +1,102 @@
...
+
+variable {R S : Type*} [NormedCommRing R] [CommRing S] [Algebra R S]
+
+/-- If `R` is a normed commutative ring and `f₁` and `f₂` are two power-multiplicative `R`-algebra
+  norms on `S`, then if `f₁` and `f₂` are equivalent on every  subring `R[y]` for `y : S`, it
+  follows that `f₁ = f₂` [BGR, Proposition 3.1.5/1][bosch-guntzer-remmert]. -/
+theorem eq_of_powMul_faithful (f₁ : AlgebraNorm R S) (hf₁_pm : IsPowMul f₁) (f₂ : AlgebraNorm R S)
+    (hf₂_pm : IsPowMul f₂)
+    (h_eq : ∀ y : S, ∃ (C₁ C₂ : ℝ) (_ : 0 < C₁) (_ : 0 < C₂),
+      ∀ x : Algebra.adjoin R {y}, f₁ x.val ≤ C₁ * f₂ x.val ∧ f₂ x.val ≤ C₂ * f₁ x.val) :
```
jcommelin, line 89: Is it possible to phrase this as actual equivalence of the `AlgebraNorm.restriction`s? Would that be helpful, or rather annoying?

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
mariainesdff: Thanks! I have deleted this comment and the next.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
mariainesdff: I don't think that would simplify applications of this lemma.

Mathlib/Analysis/Normed/Algebra/Norm.lean
mariainesdff: This is possible if adding `[Nontrivial S]  [NoZeroSMulDivisors A S]` as in the code below, but then I would also have to add hypotheses `[Nontrivial S] [NoZeroDivisors S]` to `eq_of_powMul_faithful`. Do you think this is worth it?

```lean
/-- The restriction of an algebra norm in a scalar tower. -/
def AlgebraNorm.restriction {A : Type*} [CommRing A] [Algebra R A] [Algebra A S]
  [Nontrivial S]  [NoZeroSMulDivisors A S] [IsScalarTower R A S] (f : AlgebraNorm R S) : 
    AlgebraNorm R A where
  toFun       := fun x : A => f (algebraMap A S x)
  map_zero'   := by simp only [map_zero]
  add_le' x y := by simp only [map_add, map_add_le_add]
  neg' x      := by simp only [map_neg, map_neg_eq_map]
  mul_le' x y := by simp only [map_mul, map_mul_le_mul]
  eq_zero_of_map_eq_zero' x hx := by
    rw [← map_eq_zero_iff (algebraMap A S) (NoZeroSMulDivisors.algebraMap_injective A S)]
    exact eq_zero_of_map_eq_zero f hx
  smul' r x := by
    simp only [Algebra.smul_def, map_mul, ← IsScalarTower.algebraMap_apply]
    simp only [← smul_eq_mul, algebraMap_smul, map_smul_eq_mul]
```

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
jcommelin: I see. Because I think it *would* be a bit easier to parse. Now I was mentally checking whether this was just what I expected, or whether there was some subtle difference with "equivalence of restrictions".

Mathlib/Analysis/Normed/Algebra/Norm.lean
jcommelin: Hmm, probably that isn't worth it. I could imagine it is useful to have a version for algebras where you explicitly ask for the assumption `Function.Injective (algebraMap A S)`. Feel free to add that if you think it's useful. Otherwise, let's just leave it like this.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
mariainesdff: Are you proposing to define something like [MulRingNorm.equiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Ring/Seminorm.html#MulRingNorm.equiv)[MulRingNorm.equiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Normed/Ring/Seminorm.html#MulRingNorm.equiv) for `AlgebraNorm`?

Mathlib/Analysis/Normed/Algebra/Norm.lean
mariainesdff: I have added it as `AlgebraNorm.isScalarTower_restriction`.

Mathlib/Analysis/Normed/Ring/IsPowMulFaithful.lean
jcommelin: Yes, that's right. But maybe you would rather do that in follow up work? I will let you decide.

bors d+

mariainesdff: Yes, I prefer to leave that for later. Thanks!

PR #15447, opened 2024-08-02
[Merged by Bors] - feat(Data/Set/Finite): union of infinitely many sets is infinite

apnelson1:
We add three separate versions of the statement that the union of an infinite collection of sets is infinite. 

---


urkud:
Thanks! :tada:
bors merge

PR #15454, opened 2024-08-03
feat: linter for bare `open (scoped) Classical`

grunweg:
Inspired by #15371: `open (scoped) Classical in` is allowed.

---

Does not pass CI yet, as part 3 (and future parts 4 and perhaps 5) still need to land.
- [x] depends on: #15413
- [x] depends on: #15417
- [ ] depends on: #15480
- [x] depends on: #15481

### DIFF
--- a/Mathlib/Tactic/Linter/Lint.lean
+++ b/Mathlib/Tactic/Linter/Lint.lean
@@ -273,6 +273,25 @@ namespace OpenClassical
 /-- Gets the value of the `linter.openClassical` option. -/
 def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.openClassical o
 
+/-- If `stx` is syntax describing an `open` command, `extractOpenNames stx`
+returns syntax for the opened names as well as an array of all names opened
+(discarding renamed or hidden items). -/
+def extractOpenNames : Syntax → Option (Syntax × (Array Name)) := fun stx ↦ do
+  match stx with
+    | `(command|open $name hiding $_) => some (name.raw, #[name.getId])
+    | `(command|open $name renaming $_) => some (name.raw, #[name.getId])
+    -- XXX Is there a nicer way to parse "open Foo (bar)" syntax?
+    -- | `(command|open $names:ident "(" $_ ")") => some names--.raw.getArgs
+    | `(command|open $openDecl) =>
+      let inner := openDecl.raw.getArgs
+      let names := s!"{inner.get! 0}"
+      -- remove [ and ], and a leading "`" of each name.
+      -- TODO: find a cleaner parser! was `inner.map (fun s ↦ s.getId)` doesn't quite work...
+      let names := ((names.stripPrefix "[").stripSuffix "]").split (· == ' ')
+      let names := (names.map (·.stripPrefix "`")).map String.toName
+      some (openDecl.raw, names.toArray)
+    | _ => none
+
 @[inherit_doc Mathlib.Linter.linter.openClassical]
 def openClassicalLinter : Linter where run := withSetOptionIn fun stx ↦ do
     unless getLinterHash (← getOptions) do
@@ -283,24 +302,11 @@ def openClassicalLinter : Linter where run := withSetOptionIn fun stx ↦ do
     unless #[`Mathlib, `test, `Archive, `Counterexamples].contains (← getMainModule).getRoot do
       return
     -- If `stx` describes an `open` command, extract the list of opened namespaces.
-    let names : Option (Array Name) := match stx with
-    | `(command|open $name hiding $_) => some #[name.getId]
-    | `(command|open $name renaming $_) => some #[name.getId]
-    --| `(command|open $names:ident "(" $_ ")") => some names--.raw.getArgs
-    | `(command|open $openDecl) =>
-      let inner := openDecl.raw.getArgs
-      let names := s!"{inner.get! 0}"
-      -- remove [ and ], and a leading "`" of each name.
-      -- TODO: find a cleaner parser! was `inner.map (fun s ↦ s.getId)` doesn't quite work...
-      let names := ((names.stripPrefix "[").stripSuffix "]").split (· == ' ')
-      let names := (names.map (·.stripPrefix "`")).map String.toName
-      some names.toArray
-    | _ => none
-    if let some names := names then
+    if let some (stxN, names) := extractOpenNames stx then
       -- We need to handle `open foo (bar)` commands also: ignore all anonymous names.
       let names := names.filter fun n ↦ !(n matches .anonymous)
       if names.contains `Classical then
-        Linter.logLint linter.openClassical stx "\
+        Linter.logLint linter.openClassical stxN "\
         please avoid 'open (scoped) Classical' statements: this can hide theorem statements\n\
         which would be better stated with explicit decidability statements.\n\
         Instead, use `open Classical in` for definitions or instances, the `classical` tactic \
--- a/Mathlib/Tactic/Linter/Lint.lean
+++ b/Mathlib/Tactic/Linter/Lint.lean
@@ -290,12 +290,11 @@ def openClassicalLinter : Linter where run := withSetOptionIn fun stx ↦ do
     | `(command|open $openDecl) =>
       let inner := openDecl.raw.getArgs
       let names := s!"{inner.get! 0}"
-      -- remove [ and ]. TODO: also remove `... want a better parser!
-      let names := (names.stripPrefix "[").stripSuffix "]"
-      dbg_trace names
-      let names := names.split (· == ' ')
-    -- todo: turn into each part into a `Name, and return these
-      some ((openDecl.raw.getArgs).map (fun s ↦ s.getId))
+      -- remove [ and ], and a leading "`" of each name.
+      -- TODO: find a cleaner parser! was `inner.map (fun s ↦ s.getId)` doesn't quite work...
+      let names := ((names.stripPrefix "[").stripSuffix "]").split (· == ' ')
+      let names := (names.map (·.stripPrefix "`")).map String.toName
+      some names.toArray
     | _ => none
     if let some names := names then
       -- We need to handle `open foo (bar)` commands also: ignore all anonymous names.
--- a/Mathlib/Tactic/Linter/Lint.lean
+++ b/Mathlib/Tactic/Linter/Lint.lean
@@ -280,16 +280,19 @@ def extractOpenNames : Syntax → Option (Syntax × (Array Name)) := fun stx ↦
   match stx with
     | `(command|open $name hiding $_) => some (name.raw, #[name.getId])
     | `(command|open $name renaming $_) => some (name.raw, #[name.getId])
-    -- XXX Is there a nicer way to parse "open Foo (bar)" syntax?
-    -- | `(command|open $names:ident "(" $_ ")") => some names--.raw.getArgs
+    | `(command|open $openOnly:openOnly) =>
+      -- The first argument is the identifier we care about.
+      some (openOnly.raw, #[(openOnly.raw.getArgs.get! 0).getId])
+    | `(command|open $openScoped:openScoped) =>
+      -- The first argument of `openScoped` is "scoped",
+      -- the second one is an array containing all the names we are interested in.
+      let namesSyntax := (openScoped.raw.getArgs).get! 1
+      some (namesSyntax, (namesSyntax.getArgs).map fun a ↦ a.getId)
     | `(command|open $openDecl) =>
-      let inner := openDecl.raw.getArgs
-      let names := s!"{inner.get! 0}"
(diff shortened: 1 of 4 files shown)


Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+      -- The first and only argument of `openDecl` is an array containing all the names we
+      -- are interested in.
+      let namesSyntax := (openDecl.raw.getArgs).get! 0
+      some (namesSyntax, (namesSyntax.getArgs).map fun a ↦ a.getId)
+    | `(command|open $openScoped:openScoped) =>
+      -- The first argument of `openScoped` is "scoped",
+      -- the second one is an array containing all the names we are interested in.
+      let namesSyntax := (openScoped.raw.getArgs).get! 1
+      some (namesSyntax, (namesSyntax.getArgs).map fun a ↦ a.getId)
+    | _ => none
```
adomani, line 314:
```suggestion
def extractOpenNames : Syntax → Option (Syntax × (Array Name))
  | `(command|open $openHiding:openHiding) =>
    -- The first argument is the identifier we care about.
    let arg := openHiding.raw[0]
    some (arg, #[arg.getId])
  | `(command|open $openRenaming:openRenaming) =>
    -- The first argument is the identifier we care about.
    let arg := openRenaming.raw[0]
    some (arg, #[arg.getId])
  | `(command|open $openOnly:openOnly) =>
    -- The first argument is the identifier we care about.
    some (openOnly.raw, #[openOnly.raw[0].getId])
  | `(command|open $openDecl:openSimple) =>
    -- The first and only argument of `openDecl` is an array containing all the names we
    -- are interested in.
    let namesSyntax := (openDecl.raw.getArgs).get! 0
    some (namesSyntax, (namesSyntax.getArgs).map (·.getId))
  | `(command|open $openScoped:openScoped) =>
    -- The first argument of `openScoped` is "scoped",
    -- the second one is an array containing all the names we are interested in.
    let namesSyntax := openScoped.raw[1]
    some (namesSyntax, (namesSyntax.getArgs).map (·.getId))
  | _ => none
```

Just a suggestion.

adomani (approved):
Thanks: looks good to me!

I'll think about it a little longer before maintainer merging, but I like it.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+      return
+    if (← MonadState.get).messages.hasErrors then
+     return
+    -- TODO: once mathlib's Lean version includes leanprover/lean4#4741, make this configurable
+    unless #[`Mathlib, `test, `Archive, `Counterexamples].contains (← getMainModule).getRoot do
+      return
+    -- If `stx` describes an `open` command, extract the list of opened namespaces.
+    if let some (stxN, names) := extractOpenNames stx then
+      if names.contains `Classical then
+        Linter.logLint linter.openClassical stxN "\
```
adomani, line 328: If in `extractOpenNames` you return an `Array Syntax` as second factor, then you can log the warning just at the `Classical`, rather than the full list of `open`ed namespaces (assuming I am parsing the code correctly).

test/Lint.lean
```diff
@@ -116,3 +116,112 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+
+/--
+warning: please avoid 'open (scoped) Classical' statements: this can hide theorem statements
+which would be better stated with explicit decidability statements.
+Instead, use `open Classical in` for definitions or instances, the `classical` tactic for proofs.
+For theorem statements, either add missing decidability assumptions or use `open Classical in`.
+note: this linter can be disabled with `set_option linter.openClassical false`
+-/
+#guard_msgs in
+open scoped Classical
```
adomani, line 202: This one is repeated.  Is this intended?  Did you want to test something different instead?  Is it just a copy-paste oversight?

grunweg: Good catch! I think I wanted to adapt the tests for `open` (which handle several cases), but for `open scoped` there is no variant for hiding or renaming items, and this got left over. I've removed this.

Mathlib/Tactic/Linter/Lint.lean
grunweg: You are, this is possible. I considered it - since you also asked, I just changed the code accordingly. The diff is neutral, so this is probably a good change.

test/Lint.lean
```diff
@@ -116,3 +116,93 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+
```
adomani, line 194:
```suggestion

```
This one might also be superfluous, then!

grunweg: Well, this and the next test verify slightly different things: this one has only a single namespace, the other is about multiple of them. If you think this difference is really inessential, I can remove this one.

adomani: No, it's ok to keep it.  When I looked at it, I think that GitHub was still showing the test that you had meanwhile removed.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+  | `(command|open $openDecl:openSimple) =>
+    -- The first and only argument of `openDecl` is an array containing all the names we
+    -- are interested in.
+    let namesSyntax := ((openDecl.raw.getArgs).get! 0).getArgs
+    some (namesSyntax.map (fun arg ↦ (arg, arg.getId)))
+  | `(command|open $openScoped:openScoped) =>
+    -- The first argument of `openScoped` is "scoped",
+    -- the second one is an array containing all the names we are interested in.
+    let namesSyntax := (openScoped.raw[1]).getArgs
+    some (namesSyntax.map fun arg ↦ (arg, arg.getId))
```
adomani, line 312:
```suggestion
def extractOpenNames : Syntax → Option (Array Syntax)
  | `(command|open $arg hiding $_*)    => some #[arg]
  | `(command|open $arg renaming $_,*) => some #[arg]
  | `(command|open $arg ($_*))         => some #[arg]  -- `openOnly`, hence the parens in `($_*)`
  | `(command|open $args*)             => some args
  | `(command|open scoped $args*)      => some args
```
If you compress this like above...

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+    unless getLinterHash (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+     return
+    -- TODO: once mathlib's Lean version includes leanprover/lean4#4741, make this configurable
+    unless #[`Mathlib, `test, `Archive, `Counterexamples].contains (← getMainModule).getRoot do
+      return
+    -- If `stx` describes an `open` command, extract the list of opened namespaces.
+    if let some namesSyntaxes := extractOpenNames stx then
+      if let some (stxN, _name) := namesSyntaxes.find? (fun (_, name) ↦ name == `Classical) then
```
adomani, line 326:
```suggestion
      if let some stxN := namesSyntaxes.find? (·.getId == `Classical) then
`Classical) then
```
... then this change is enough to make everything parse again!

test/Lint.lean
```diff
@@ -116,3 +116,93 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+/--
+warning: please avoid 'open (scoped) Classical' statements: this can hide theorem statements
+which would be better stated with explicit decidability statements.
+Instead, use `open Classical in` for definitions or instances, the `classical` tactic for proofs.
+For theorem statements, either add missing decidability assumptions or use `open Classical in`.
+note: this linter can be disabled with `set_option linter.openClassical false`
+-/
+#guard_msgs in
+open scoped Int Classical Nat
```
adomani, line 203:
```suggestion
---
warning: please avoid 'open (scoped) Classical' statements: this can hide theorem statements
which would be better stated with explicit decidability statements.
Instead, use `open Classical in` for definitions or instances, the `classical` tactic for proofs.
For theorem statements, either add missing decidability assumptions or use `open Classical in`.
note: this linter can be disabled with `set_option linter.openClassical false`
-/
#guard_msgs in
open scoped Int Classical Nat Classical
```
And this tests the reporting of every instance of `Classical` that is open.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+    -- The first and only argument of `openDecl` is an array containing all the names we
+    -- are interested in.
+    let namesSyntax := ((openDecl.raw.getArgs).get! 0).getArgs
+    some (namesSyntax.map (fun arg ↦ (arg, arg.getId)))
+  | `(command|open $openScoped:openScoped) =>
+    -- The first argument of `openScoped` is "scoped",
+    -- the second one is an array containing all the names we are interested in.
+    let namesSyntax := (openScoped.raw[1]).getArgs
+    some (namesSyntax.map fun arg ↦ (arg, arg.getId))
+  | _ => none
```
adomani, line 313:
```suggestion
def extractOpenNames : Syntax → Array Syntax
  | `(command|open $arg hiding $_*)    => #[arg]
  | `(command|open $arg renaming $_,*) => #[arg]
  | `(command|open $arg ($_*))         => #[arg]  -- openOnly, hence the parens in `($_*)`
  | `(command|open $args*)             => args
  | `(command|open scoped $args*)      => args
  | _ => #[]
```
If you compress this like above...

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+      return
+    -- If `stx` describes an `open` command, extract the list of opened namespaces.
+    if let some namesSyntaxes := extractOpenNames stx then
+      if let some (stxN, _name) := namesSyntaxes.find? (fun (_, name) ↦ name == `Classical) then
+        Linter.logLint linter.openClassical stxN "\
+        please avoid 'open (scoped) Classical' statements: this can hide theorem statements\n\
+        which would be better stated with explicit decidability statements.\n\
+        Instead, use `open Classical in` for definitions or instances, the `classical` tactic \
+        for proofs.\nFor theorem statements, \
+        either add missing decidability assumptions or use `open Classical in`."
```
adomani, line 332:
```suggestion
    for stxN in (extractOpenNames stx).filter (·.getId == `Classical) do
      Linter.logLint linter.openClassical stxN "\
      please avoid 'open (scoped) Classical' statements: this can hide theorem statements\n\
      which would be better stated with explicit decidability statements.\n\
      Instead, use `open Classical in` for definitions or instances, the `classical` tactic \
      for proofs.\nFor theorem statements, \
      either add missing decidability assumptions or use `open Classical in`."
```
then this fixes the parsing!

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -269,4 +269,70 @@ initialize addLinter longLineLinter
...
+    some (namesSyntax.map (fun arg ↦ (arg, arg.getId)))
+  | `(command|open $openScoped:openScoped) =>
+    -- The first argument of `openScoped` is "scoped",
+    -- the second one is an array containing all the names we are interested in.
+    let namesSyntax := (openScoped.raw[1]).getArgs
+    some (namesSyntax.map fun arg ↦ (arg, arg.getId))
+  | _ => none
+
+@[inherit_doc Mathlib.Linter.linter.openClassical]
+def openClassicalLinter : Linter where run := withSetOptionIn fun stx ↦ do
```
adomani, line 316:
```suggestion
def openClassicalLinter : Linter where run stx := do
```
Using a `set_option in` in front of a non-`in`-`set_option` seems like it would achieve to effective cancel the inner `set_option`.  Maybe this could be another linter, but here I don't think that `withSetOptionIn` is really needed.

Please double check, as I'm not completely sure!

grunweg: I just tried it out: making this change breaks the standard guard_msg dance: then, the following passes (i.e., the linter is silent)

set_option linter.openClassical false in
#guard_msgs in
set_option linter.openClassical true in
open Classical
```
Removing the `in` in the last line makes the linter complain, but the guard_msg empty - that's fun, but not really what I would expect.

grunweg (commented):
Thanks for the simplifications; that is much nicer!

grunweg:
Re-based in #21947.

grunweg:
Closing in favour of the rebased version.

PR #15455, opened 2024-08-03
[Merged by Bors] - chore(*): use `getElem` instead of `List.get`/`List.nthLe`

urkud:
Deletions:
- List.toFinsupp_apply_lt'

---


Mathlib/Order/RelSeries.lean
```diff
@@ -105,8 +105,10 @@ lemma toList_ne_nil (x : RelSeries r) : x.toList ≠ [] := fun m =>
 /-- Every nonempty list satisfying the chain condition gives a relation series-/
 @[simps]
 def fromListChain' (x : List α) (x_ne_nil : x ≠ []) (hx : x.Chain' r) : RelSeries r where
-  length := x.length.pred
-  toFun := x.get ∘ Fin.cast (Nat.succ_pred_eq_of_pos <| List.length_pos.mpr x_ne_nil)
+  length := x.length - 1
+  toFun i := x[i]'(by
```
eric-wieser, line 109: Since this takes in a Fin, would this be better as
```suggestion
  toFun i := x[Fin.cast (Nat.succ_pred_eq_of_pos <| List.length_pos.mpr x_ne_nil) i]
```
or similar?

eric-wieser (approved):
This looks OK to me, but I think @semorrison or @fgdorais should check this matches their expectations.

fgdorais (approved):
All changes here look good to me. Eric's proposed edit also looks good.

Mathlib/Analysis/Calculus/FDeriv/Mul.lean
```diff
@@ -570,7 +570,7 @@ theorem hasStrictFDerivAt_list_prod_finRange' {n : ℕ} {x : Fin n → 𝔸} :
 theorem hasStrictFDerivAt_list_prod_attach' [DecidableEq ι] {l : List ι} {x : {i // i ∈ l} → 𝔸} :
     HasStrictFDerivAt (𝕜 := 𝕜) (fun x ↦ (l.attach.map x).prod)
       (∑ i : Fin l.length, ((l.attach.take i).map x).prod •
-        smulRight (proj (l.attach.get (i.cast l.length_attach.symm)))
+        smulRight (proj (l.attach[i]'(by rw [l.length_attach]; exact i.2)))
```
eric-wieser, line 573: I guess the same question arises here; should `a[Fin.cast i h]` be preferred to `a[i]'(by something [h])`?

fgdorais: I've encountered some awkwardness with `a[i]'(h)` when `i` is `Fin _` and `h` is nontrivial. I expect this to be fixed sooner than later but, since we happen to be in the present, I think `Fin.cast` has better support.

urkud: My main concern is that `Finc.cast` is not the simp normal form here.

fgdorais: That is a very legitimate concern. I think the simp normal form of `a[i]'(h)` is still fluid in some cases so it's a gamble either way.

j-loreaux:
@semorrison can you please confirm that this is the correct direction? I'm delegating this to you (even though you could just merge it yourself).

bors d=semorrison

semorrison:
bors merge

PR #15462, opened 2024-08-03
chore(ZMod/Basic): rewrite proof without `ModEq`, part 1

rwst:
Rewrite `natCast_zmod_eq_zero_iff_dvd` without using `ModEq`. Part of the effort of reducing the dependency of `ZMod/Basic` on `Data/*/ModEq`.

Zulip: https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/rw.20in.20ZMOD

---


urkud:
It looks like I've missed the discussion. What's wrong with `Int.ModEq`? Could you please link the relevant Zulip thread?

PR #15465, opened 2024-08-03
[Merged by Bors] - feat(Combinatorics/SimpleGraph): `dist_bot` and `dist_top`

Rida-Hamadani:
---


Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -128,6 +128,15 @@ theorem edist_eq_one_iff_adj : G.edist u v = 1 ↔ G.Adj u v := by
     exact w.adj_of_length_eq_one <| Nat.cast_eq_one.mp <| h ▸ hw
   · exact le_antisymm (edist_le h.toWalk) (ENat.one_le_iff_pos.mpr <| edist_pos_of_ne h.ne)
 
+lemma edist_bot [DecidableEq V] :
+    ∀ u v, (⊥ : SimpleGraph V).edist u v = (if u = v then 0 else ⊤) := by
```
IvanRenison, line 132: Why not?
```lean
lemma edist_bot [DecidableEq V] (u v : V) : (⊥ : SimpleGraph V).edist u v = (if u = v then 0 else ⊤)
```
(And the same in the others)

Rida-Hamadani: Good point, thank you! ^^

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -128,6 +128,13 @@ theorem edist_eq_one_iff_adj : G.edist u v = 1 ↔ G.Adj u v := by
     exact w.adj_of_length_eq_one <| Nat.cast_eq_one.mp <| h ▸ hw
   · exact le_antisymm (edist_le h.toWalk) (ENat.one_le_iff_pos.mpr <| edist_pos_of_ne h.ne)
 
+lemma edist_bot [DecidableEq V] : (⊥ : SimpleGraph V).edist u v = (if u = v then 0 else ⊤) := by
```
b-mehta, line 131: I think the "special" case of `edist u v = ⊤` in the case that u and v are distinct should be a lemma, in the same way that eg `Function.update_noteq` is. (The case that they're equal is fine since it's in `edist_self` already, but probably that should be a simp-lemma?)

b-mehta: Oh sorry, I didn't mean to say you should get rid of `edist_bot`, instead that we should have both! `edist_bot_of_ne` is a fine name

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -256,6 +259,9 @@ lemma Connected.exists_path_of_dist (hconn : G.Connected) (u v : V) :
 lemma dist_bot : (⊥ : SimpleGraph V).dist u v = 0 := by
```
b-mehta, line 259: This one can probably be a simp lemma

b-mehta (commented):
bors d+

PR #15466, opened 2024-08-03
[Merged by Bors] - feat(Probability/Kernel): `Kernel.sectL` and `sectR`

LorenzoLuccioli:
Define a `Kernel α γ` and a `Kernel β γ` from a `Kernel (α × β) γ` by taking the comap of `fun a ↦ (a, b)` and `fun b ↦ (a, b)` respectively.

- Add definitions of `Kernel.sectL` and `Kernel.sectR`.
- Add some API lemmas: `sectL_apply`, `sectL_zero`, `comap_sectL`, `sectL_prodMkLeft`, `sectL_prodMkRight` and the equivalent lemmas for `sectR`.
- Add instances for `IsMarkovKernel`, `IsFiniteKernel`, `IsSFiniteKernel` and `NeZero`.
- Add `sectL_swapRight` and `sectR_swapRight`.
- Add `compProd_apply_eq_compProd_sectR`.

---


LorenzoLuccioli:
The names are `Kernel.fst'` and `Kernel.snd'`, since `Kernel.fst` and `Kernel.snd` are already taken for the case where the codomain is a product space. Suggestions for a better name are welcome.

Mathlib/Probability/Kernel/Composition.lean
```diff
@@ -1026,6 +1026,96 @@ lemma snd_swapRight (κ : Kernel α (β × γ)) : snd (swapRight κ) = fst κ :=
 
 end FstSnd
 
+section Fst'Snd'
+
+variable {γ δ : Type*} {mγ : MeasurableSpace γ} {mδ : MeasurableSpace δ}
+
+/-- Define a `Kernel α γ` from a `Kernel (α × β) γ` by taking the comap of `fun a ↦ (a, b)` for
+a given `b : β`. -/
+noncomputable def fst' (κ : Kernel (α × β) γ) (b : β) : Kernel α γ :=
```
YaelDillies, line 1035: Could you rename that to
```suggestion
noncomputable def domFst (κ : Kernel (α × β) γ) (b : β) : Kernel α γ :=
```
? Then `Kernel.fst` can be renamed to `Kernel.codFst`

LorenzoLuccioli: Thank you for the suggestion!
Among the proposed names this is the one I prefer, so for now I am going to use it.

LorenzoLuccioli: Regarding [`Kernel.fst`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Probability/Kernel/Composition.html#ProbabilityTheory.Kernel.fst), I think the current name is fitting, since it is defined using [`Prod.fst`](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Prod.fst) and pointwise it is equal to the first marginal of the measure [Measure.fst](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Measure/Prod.html#MeasureTheory.Measure.fst), as can be seen from the following equality:

```
example {α β γ : Type*} [MeasurableSpace α] [MeasurableSpace β] [MeasurableSpace γ] 
    (κ : Kernel α (β × γ)) (a : α) : κ.fst a = (κ a).fst := rfl
```

YaelDillies: Yes, so the situation here is that `Kernel` is functorial in both its arguments, contravariantly in the first and covariantly in the second. This means that `Prod.fst : α × β → α` gives rise to two natural maps that could "fittingly" be called `Kernel.fst`:
* `Kernel α γ → Kernel (α × β) γ`, which you are introducing here
* `Kernel γ (α × β) → Kernel γ α`, which is the existing `Kernel.fst`

YaelDillies: ... and now I realise that you are not introducing the map `Kernel α γ → Kernel (α × β) γ` but rather `β → Kernel (α × β) γ → Kernel α γ` which is functoriality in the first argument from the map `Prod.sectL b : α → α × β`.

Following my naming convention suggestion, this should be called `domSectL`.

LorenzoLuccioli: I see, this is similar to a previous suggestion to use `sectl`; it still feels a bit obscure as a name to me, but since this definition is not really linked to `Prod.fst` I guess that it really makes no sense for `fst` to appear in the name, so `sectl` will do.

I have doubts about 2 details:
- Is the `dom` part actually useful? With a quick search into the library I didn't find any occurrence of this convention and I'm not sure whether we will ever need `codSecL`.
- Should `L` at the end be capitalized (`sectL` vs `sectl`)? In Mathlib it seems to always appear non capitalized (e.g. [`Prod.sectl`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Products/Basic.html#CategoryTheory.Prod.sectl)).

YaelDillies: * I don't know whether we will ever need `codSectL`, no. Feel free to keep the `dom` out of the name for now
* I personally find `sectl` and `sectr` to be quite obscure names, while `sectL` and `sectR` are much clearer in their intent. I think it's okay to use `sectl` for now, but I would prefer `sectL`, and even better would be to first rename all the `sectl` to `sectL`

YaelDillies: I've opened #18907 to rename the (few) existing uses of the name

YaelDillies:
Can you please update the PR description?

YaelDillies (approved):
Thanks! 🚀

maintainer merge

YaelDillies:
Looks like this is ready to be merged again?

maintainer merge

PR #15473, opened 2024-08-03
[Merged by Bors] - feat (MeasureTheory.Decomposition.RadonNikodym): `rnDeriv_le_one_iff_le`

LorenzoLuccioli:
Add `rnDeriv_le_one_iff_le` and `rnDeriv_eq_one_iff_eq`.

---

### DIFF
--- a/Mathlib/MeasureTheory/Decomposition/RadonNikodym.lean
+++ b/Mathlib/MeasureTheory/Decomposition/RadonNikodym.lean
@@ -78,7 +78,7 @@ lemma rnDeriv_pos [HaveLebesgueDecomposition μ ν] (hμν : μ ≪ ν) :
     ae_withDensity_iff (Measure.measurable_rnDeriv _ _), Measure.withDensity_rnDeriv_eq _ _  hμν]
   exact ae_of_all _ (fun x hx ↦ lt_of_le_of_ne (zero_le _) hx.symm)
 
-lemma rnDeriv_pos' [SigmaFinite μ] [SFinite ν] (hμν : μ ≪ ν) :
+lemma rnDeriv_pos' [HaveLebesgueDecomposition ν μ] [SigmaFinite μ] (hμν : μ ≪ ν) :
     ∀ᵐ x ∂μ, 0 < ν.rnDeriv μ x := by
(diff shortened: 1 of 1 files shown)


Mathlib/MeasureTheory/Decomposition/RadonNikodym.lean
```diff
@@ -405,6 +405,17 @@ lemma rnDeriv_le_one_of_le (hμν : μ ≤ ν) [SigmaFinite ν] : μ.rnDeriv ν
   simp only [Pi.one_apply, MeasureTheory.setLIntegral_one]
   exact (Measure.setLIntegral_rnDeriv_le s).trans (hμν s)
 
+lemma rnDeriv_le_one_iff_le [SigmaFinite μ] [SigmaFinite ν] (hμν : μ ≪ ν) :
```
JasonKYi, line 408: Does it work if you replace the assumption by this?
```suggestion
lemma rnDeriv_le_one_iff_le [μ.HaveLebesgueDecomposition ν] (hμν : μ ≪ ν) :
```

LorenzoLuccioli: It does not work, since `rnDeriv_le_one_of_le` and `withDensity_apply'` need `[SigmaFinite ν]`. 
`[μ.HaveLebesgueDecomposition ν] [SigmaFinite ν]` works, should we leave like it is, or use these hypotheses?

LorenzoLuccioli: It does not work, since `rnDeriv_le_one_of_le` and `withDensity_apply'` need `[SigmaFinite ν]`. 
`[μ.HaveLebesgueDecomposition ν] [SigmaFinite ν]` works, should we leave like it is, or use these hypotheses?

LorenzoLuccioli: Ok, I generalized the hypotheses of the other lemmas in the file, where possible.

JasonKYi:
Thanks!
maintainer merge

semorrison:
bors merge

PR #15478, opened 2024-08-03
[Merged by Bors] - feat: in separable Borel spaces, convergence in distribution is metrizable (not just pseudometrizable)

kkytola:
In earlier PRs it was shown that the Lévy-Prokhorov distance pseudometrizes the convergence in distribution (i.e. weak convergence of probability measures) on separable pseudometric spaces. This PR strengthens the conclusion: the LP distance on probability measures on Borel spaces is a metric (not just a pseudometric), and therefore in separable Borel spaces, convergence in distribution is metrizable (not just pseudometrizable).

---


Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
```diff
@@ -198,6 +194,24 @@ lemma levyProkhorovDist_triangle [OpensMeasurableSpace Ω] (μ ν κ : Measure 
 when they are to be equipped with the Lévy-Prokhorov distance. -/
 def LevyProkhorov (α : Type*) := α
 
+/-- Coercion from the type synonym `LevyProkhorov (ProbabilityMeasure Ω)`
+to `ProbabilityMeasure Ω`. -/
```
j-loreaux, line 198: If you want this, and the other three `def`s below to be coercions, then you need to do two things:

1. Define appropriate `Coe` instances for each. (this allows Lean to look for the coercion during elaboration, so that you don't have to manually insert it)
2. add the `@[coe]` attribute, which allows Lean to display a `↑` in the infoview and makes `norm_cast` aware that this is a coercion for its purposes.

However, that being said, I tend to think that you don't want *any* of these. Instead, it seems like you want a "parameterized type synonym", much like we have for `WithLp` or `Lex`. So I think you just want a *single* `def`, and it should be an `Equiv`, not a plain function. That is:
```lean
-- maybe you could make `α` implicit?
def LevyProkhorov.equiv (α : Type*) : LevyProkhorov α ≃ α where
  ...
```

Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
```diff
@@ -233,6 +280,30 @@ noncomputable instance levyProkhorovDist_pseudoMetricSpace_probabilityMeasure :
...
+`ProbabilityMeasure Ω` a metric space. The instance is recorded on the type synonym
+`LevyProkhorov (ProbabilityMeasure Ω) := ProbabilityMeasure Ω`.
+
+Note: For this metric to give the topology of convergence in distribution, one must
+furthermore assume that `Ω` is separable. -/
+noncomputable instance levyProkhorovDist_metricSpace_probabilityMeasure [BorelSpace Ω] :
+    MetricSpace (LevyProkhorov (ProbabilityMeasure Ω)) where
+  eq_of_dist_eq_zero := by
+    intro μ ν h
+    have same : μ.toProbabilityMeasure = ν.toProbabilityMeasure := by
```
j-loreaux, line 293:
```suggestion
  show μ.toProbabilityMeasure = ν.toProbabilityMeasure
```
Then you can unindent and remove the `exact same` at the end.

Or better yet, you can use my `equiv` suggestion above and just make this line:
```suggestion
  apply LevyProkhorov.equiv.injective
```
and remove the defeq abuse in its entirety.

Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
```diff
@@ -219,9 +233,42 @@ noncomputable instance levyProkhorovDist_pseudoMetricSpace_finiteMeasure :
...
+lemma measure_eq_measure_of_isClosed_of_levyProkhorovEDist_eq_zero {μ ν : Measure Ω}
+    (hLP : levyProkhorovEDist μ ν = 0) {s : Set Ω} (s_closed : IsClosed s)
+    (hμs : ∃ δ > 0, μ (thickening δ s) ≠ ∞) (hνs : ∃ δ > 0, ν (thickening δ s) ≠ ∞) :
+    μ s = ν s := by
+  apply le_antisymm
+  · exact (measure_le_measure_closure_of_levyProkhorovEDist_eq_zero
+            hLP s_closed.measurableSet hνs).trans (le_of_eq (congr_arg _ s_closed.closure_eq ))
+  · exact (measure_le_measure_closure_of_levyProkhorovEDist_eq_zero
+              (levyProkhorovEDist_comm μ ν ▸ hLP) s_closed.measurableSet hμs).trans
+            (le_of_eq (congr_arg _ s_closed.closure_eq ))
```
j-loreaux, line 264:
```suggestion
  · exact measure_le_measure_closure_of_levyProkhorovEDist_eq_zero
      hLP s_closed.measurableSet hνs |>.trans <|
      le_of_eq (congr_arg _ s_closed.closure_eq)
  · exact measure_le_measure_closure_of_levyProkhorovEDist_eq_zero
      (levyProkhorovEDist_comm μ ν ▸ hLP) s_closed.measurableSet hμs |>.trans <|
      le_of_eq (congr_arg _ s_closed.closure_eq)
```

j-loreaux (commented):
I may have more to say after some of this has been addressed.

Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
kkytola: Thanks! I did this and it looks cleaner indeed --- maybe a little less descriptive as naming, but much better that the "coercion" is not special-cased to both cases of interest.

kkytola: I noticed that I should still update naming of the lemmas to refer to this equiv, so I marked as "awaiting author" again until then...

Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
```diff
@@ -198,6 +194,9 @@ lemma levyProkhorovDist_triangle [OpensMeasurableSpace Ω] (μ ν κ : Measure 
 when they are to be equipped with the Lévy-Prokhorov distance. -/
 def LevyProkhorov (α : Type*) := α
 
+/-- The "identity" equivalence between the type synonym `LevyProkhorov α` and `α`. -/
+def LevyProkhorov.equiv {α : Type*} : LevyProkhorov α ≃ α := Equiv.refl _
```
JasonKYi, line 198:
```suggestion
def LevyProkhorov.equiv (α : Type*) : LevyProkhorov α ≃ α := Equiv.refl _
```
I think this is better

RemyDegenne:
> I may have more to say after some of this has been addressed.

@j-loreaux you wrote that last month. Do you have more to say?

Mathlib/MeasureTheory/Measure/LevyProkhorovMetric.lean
```diff
@@ -219,9 +218,43 @@ noncomputable instance levyProkhorovDist_pseudoMetricSpace_finiteMeasure :
...
+  have obs := Tendsto.add key (tendsto_nhdsWithin_of_tendsto_nhds tendsto_id)
+  simp only [id_eq, add_zero] at obs
+  apply ge_of_tendsto (b := μ s) obs
+  filter_upwards [self_mem_nhdsWithin] with ε ε_pos
+  exact left_measure_le_of_levyProkhorovEDist_lt (B_mble := s_mble) (hLP ▸ ε_pos)
+
+/-- Two measures at vanishing Lévy-Prokhorov distance from each other assign the same values to all
+closed sets. -/
+lemma measure_eq_measure_of_isClosed_of_levyProkhorovEDist_eq_zero {μ ν : Measure Ω}
+    (hLP : levyProkhorovEDist μ ν = 0) {s : Set Ω} (s_closed : IsClosed s)
```
RemyDegenne, line 241: The `of_X_of_Y` in the name should be in the same order as `X` and `Y` in the hypotheses. So with this hypothesis order it should be `measure_eq_measure_of_levyProkhorovEDist_eq_zero_of_isClosed`.

RemyDegenne (approved):
Looks good to me. Please apply the suggestion of Jason and let's wait a bit for Jireh, but otherwise the PR seems good to go.

RemyDegenne:
If you find time to fix the two little things mentioned above, we could then merge this PR.

RemyDegenne (approved):
Thanks!
bors r+

PR #15480, opened 2024-08-04
chore(NumberTheory): replace remaining occurrences of open Classical

grunweg:
-------

I'm very open to fixing these in a different way: for instance, in RamificationInertia, the FactorsMap section can *almost* be changed to add a [DecidableEq (Ideal S)] variable instead.

---


Mathlib/NumberTheory/LSeries/HurwitzZetaEven.lean
```diff
@@ -189,6 +189,7 @@ lemma hasSum_int_cosKernel (a : ℝ) {t : ℝ} (ht : 0 < t) :
   simp only [this]
   exact hasSum_jacobiTheta₂_term _ (by rwa [I_mul_im, ofReal_re])
 
+open Classical in
```
sgouezel, line 192: Instead of these `open Classical in`, I think you should register a `DecidableEq` instance for `UnitAddCircle`, once and for all (using classical itself).

grunweg: Good idea! Done.

Mathlib/NumberTheory/NumberField/CanonicalEmbedding/Basic.lean
```diff
@@ -401,15 +402,14 @@ end norm
...
-open scoped Classical
-
 open Complex MeasureTheory MeasureTheory.Measure Zspan Matrix ComplexConjugate
 
 variable [NumberField K]
 
 /-- The type indexing the basis `stdBasis`. -/
 abbrev index := {w : InfinitePlace K // IsReal w} ⊕ ({w : InfinitePlace K // IsComplex w}) × (Fin 2)
 
+open Classical in
```
sgouezel, line 412: Instead of adding `open Classical in` whenever you want to have fintype instances and normed space structure on `E K`, I'd add somewhere the lines
```
instance : DecidablePred (fun (w : InfinitePlace K) ↦ IsReal w) := Classical.decPred _
instance : DecidablePred (fun (w : InfinitePlace K) ↦ IsComplex w) := Classical.decPred _
```
(maybe just after `NumberField.InfinitePlace.fintype`?)

In this way, you can remove most `open Classical` from this file and the other one.

grunweg:
This work has been done: this PR is now obsolete.

PR #15481, opened 2024-08-04
[Merged by Bors] - chore(Topology): remove some uses of open Classical

grunweg:
---

Again, comments with a different preference on how to fix these are welcome.


riccardobrasca:
Thanks!

bors merge

PR #15482, opened 2024-08-04
[Merged by Bors] - chore(GroupTheory/Coset): deprecate some lemma

astrainfinita:
They were different in mathlib3 because of coercions.

---

### DIFF
--- a/Mathlib/GroupTheory/Coset.lean
+++ b/Mathlib/GroupTheory/Coset.lean
@@ -408,13 +408,13 @@ variable (s)
   stated in terms of an arbitrary `h : s`, rather than the specific `h = g⁻¹ * (mk g).out'`. -/
 @[to_additive QuotientAddGroup.mk_out'_eq_mul]
 theorem mk_out'_eq_mul (g : α) : ∃ h : s, (mk g : α ⧸ s).out' = g * h :=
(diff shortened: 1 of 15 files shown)


eric-wieser (approved):
bors merge

Thanks!

Mathlib/GroupTheory/Coset.lean
```diff
@@ -371,9 +371,7 @@ theorem induction_on {C : α ⧸ s → Prop} (x : α ⧸ s) (H : ∀ z, C (Quoti
 instance : Coe α (α ⧸ s) :=
   ⟨mk⟩
 
-@[to_additive (attr := elab_as_elim)]
-theorem induction_on' {C : α ⧸ s → Prop} (x : α ⧸ s) (H : ∀ z : α, C z) : C x :=
-  Quotient.inductionOn' x H
+@[to_additive (attr := deprecated (since := "2024-08-04"))] alias induction_on' := induction_on
```
kbuzzard, line 374: I know this has been merged now but did we just lose an `elab_as_elim` attribute on the primed declaration?

eric-wieser: Good catch, though it probably doesn't matter much since this is now deprecated.

PR #15485, opened 2024-08-04
Feat (Mathlib.Data.Finset.NoncommProd) : universal coproduct property of finite products

AntoineChambert-Loir:
- Add the `Finset.noncommProd` versions of `Finset.prod_eq_single` and `Finset.prod_eq_one`

- `MonoidHom.pi_lift` proves the universal coproduct property of finite products of monoids
---


AntoineChambert-Loir:
I discovered that it already existed (under the name noncommPiCoprod)

PR #15488, opened 2024-08-04
[Merged by Bors] - perf(Geometry/Manifold/Instances/Sphere): speed up

grunweg:
Speed up the file by 9% by doing the following
- squeeze simps where the full form is short and this is a clear win
- "squeeze" three really slow field_simp calls: replace them by the equivalent simp invocation and squeeze that
- avoid one slow `convert`
- squeeze two really slow nlinarith calls (they're still slow, but much better)
- replace one congr! by congr

This is a reduced version of #13456, which is hopefully less controversial.

---


grunweg:
!bench

jcommelin:
Thanks :tada:

bors merge

PR #15490, opened 2024-08-04
[Merged by Bors] - chore: include bare 'open Classical' in tech debt metrics

grunweg:
And remove one commented counter: there is now a lint against deprecations without a date, so keeping it is probably not useful.

---

### DIFF
--- a/scripts/technical-debt-metrics.sh
+++ b/scripts/technical-debt-metrics.sh
@@ -55,9 +55,9 @@ done
 
 printf '%s|%s\n' "$(grep -c 'docBlame' scripts/nolints.json)" "documentation nolint entries"
 printf '%s|%s\n' "$(grep -c 'ERR_NUM_LIN' scripts/style-exceptions.txt)" "large files"
+print '%s|%s\n' "$(git grep "open " | grep " Classical" | grep -v " in$" -c)" "bare open (scoped) Classical"
 # We print the number of files, not the number of matches --- hence, the nested grep.
 printf '%s|%s\n' "$(git grep -c 'autoImplicit true' | grep -c -v 'test')" "non-test files with autoImplicit true"
-#printf '%s|%s\n' "$(git grep '@\[.*deprecated' | grep -c -v 'deprecated .*(since := "')" "deprecations without a date"
 
 initFiles="$(git ls-files '**/Init/*.lean' | xargs wc -l | sed 's=^ *==')"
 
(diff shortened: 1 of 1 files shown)


scripts/technical-debt-metrics.sh
```diff
@@ -55,9 +55,9 @@ done
 
 printf '%s|%s\n' "$(grep -c 'docBlame' scripts/nolints.json)" "documentation nolint entries"
 printf '%s|%s\n' "$(grep -c 'ERR_NUM_LIN' scripts/style-exceptions.txt)" "large files"
+print '%s|%s\n' "$(git grep "open " | grep " Classical" | grep -v " in$" -c)" "bare open (scoped) Classical"
```
adomani, line 58:
```suggestion
print '%s|%s\n' "$(git grep "^open .*Classical" | grep -v " in$" -c)" "bare open (scoped) Classical"
```

scripts/technical-debt-metrics.sh
```diff
@@ -55,9 +55,9 @@ done
...
 # We print the number of files, not the number of matches --- hence, the nested grep.
 printf '%s|%s\n' "$(git grep -c 'autoImplicit true' | grep -c -v 'test')" "non-test files with autoImplicit true"
-#printf '%s|%s\n' "$(git grep '@\[.*deprecated' | grep -c -v 'deprecated .*(since := "')" "deprecations without a date"
```
adomani, line 60: I think that the reports of the script have motivated people to take action to reduce some of the debts.  So, while I agree that the linter tracks un-dated `deprecation`s, it might still be useful to include it in the reports.  Maybe adding a comment to the extent that this is tracked by a linter.

grunweg: Indeed; this is one reason I'm proposing to track bare "open Classical" statements :-)

I just checked: that linter is never allowed; the current 11 hits are all false positives. Do we ever expect the linter to be allowed? I don't, so I'm not sure if including this actually helps anybody. (Reporting a line that is always zero, if we don't expect that to change, does not seem helpful to me.)
I can certainly add a comment that some previous technical debt, which is now fixed and (if applicable) linted against, is not tracked any more.

scripts/technical-debt-metrics.sh
```diff
@@ -55,9 +55,9 @@ done
 
 printf '%s|%s\n' "$(grep -c 'docBlame' scripts/nolints.json)" "documentation nolint entries"
 printf '%s|%s\n' "$(grep -c 'ERR_NUM_LIN' scripts/style-exceptions.txt)" "large files"
+print '%s|%s\n' "$(git grep "^open .*Classical" | grep -v " in$" -c)" "bare open (scoped) Classical"
```
eric-wieser, line 58:
```suggestion
printf '%s|%s\n' "$(git grep "^open .*Classical" | grep -v " in$" -c)" "bare open (scoped) Classical"
```

scripts/technical-debt-metrics.sh
adomani: No, you convinced me: I'm happy to remove this counter.

adomani:
Thanks a lot!

maintainer merge

riccardobrasca:
Thanks!

bors merge

PR #15491, opened 2024-08-04
Feat(Mathlib/Algebra/Group/Prod): add noncomm coproduct property

AntoineChambert-Loir:
We add the noncomm analogues of `MonoidHom.coprod`

---


AntoineChambert-Loir:
I discover that it existed.

PR #15493, opened 2024-08-04
chore(RingTheory/Valuation): actually deprecate `Valuation.ext_iff`

grunweg:
Its deprecation is commented, with an outdated porting note. Deprecate it for real.

---


grunweg:
This is obsoleted by the Lean bump, closing.

PR #15501, opened 2024-08-04
[Merged by Bors] - feat(Combinatorics/SimpleGraph): Add definition of circulant graph and cycle graph

IvanRenison:

---
This pr replaces #14364

### DIFF
--- a/Mathlib/Combinatorics/SimpleGraph/Circulant.lean
+++ b/Mathlib/Combinatorics/SimpleGraph/Circulant.lean
@@ -15,7 +15,7 @@ are adjacent if and only if `u - v ∈ s` or `v - u ∈ s`. The elements of `s` 
 
 ## Main declarations
 
-* `SimpleGraph.CirculantGraph s`: the circulant graph over `G` with jumps `s`.
+* `SimpleGraph.circulantGraph s`: the circulant graph over `G` with jumps `s`.
 * `SimpleGraph.cycleGraph n`: the cycle graph over `Fin n`.
 -/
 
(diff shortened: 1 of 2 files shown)


Mathlib/Combinatorics/SimpleGraph/Circulant.lean
```diff
@@ -0,0 +1,136 @@
...
+  exact Or.symm h
+
+instance [DecidableEq G] [DecidablePred (· ∈ s)] : DecidableRel (CirculantGraph s).Adj :=
+  fun _ _ => inferInstanceAs (Decidable (_ ∧ _))
+
+theorem circulantGraph_adj_translate {s : Set G} {u v d : G} :
+    (CirculantGraph s).Adj (u + d) (v + d) ↔ (CirculantGraph s).Adj u v := by simp
+
+/-- Cycle graph over `Fin n` -/
+def cycleGraph : (n : ℕ) → SimpleGraph (Fin n)
```
b-mehta, line 59:
```suggestion
def CycleGraph : (n : ℕ) → SimpleGraph (Fin n)
```

Mathlib/Combinatorics/SimpleGraph/Circulant.lean
```diff
@@ -0,0 +1,136 @@
+/-
+Copyright (c) 2024 Iván Renison, Bhavik Mehta. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Iván Renison, Bhavik Mehta
+-/
+import Mathlib.Combinatorics.SimpleGraph.Hasse
+import Mathlib.Data.Set.Pointwise.Basic
+
+/-!
+# Definition of circulants graphs
```
b-mehta, line 10:
```suggestion
# Definition of circulant graphs
```
and similar below

b-mehta (commented):
I think there's lots we could add about these graphs, but this is definitely a great start, and further things can be added in later PRs.
I've written too much of this to merge it however, so I'll ask another maintainer to take a look too.

maintainer merge

Mathlib/Combinatorics/SimpleGraph/Circulant.lean
IvanRenison: I used `cycleGraph` for using the same style that `pathGraph`, you think that `CycleGraph` is better?

IvanRenison: Also, `completeGraph`, `emptyGraph`, `completeBipartiteGraph`, `hasse` are in lower case, so I will change `CirculantGraph`

b-mehta:
good point, my mistake

riccardobrasca:
Thanks!

bors merge

PR #15508, opened 2024-08-05
[Merged by Bors] - feat(Order/Lattice): add Monotone.of_map_inf_le_left etc

urkud:
- Add `Monotone.of_map_inf_le_left` etc.

- Use `Monotone.of_left_le_map_sup` to reorganize the proof of
  `SimpleFunc.lintegral_mono`.

- Add `gcongr` attrs.

---


Mathlib/Order/Lattice.lean
```diff
@@ -902,9 +902,27 @@ theorem map_inf_le [SemilatticeInf α] [SemilatticeInf β] {f : α → β} (h :
...
+  apply h
+
+theorem of_map_inf_le [SemilatticeInf α] [SemilatticeInf β] {f : α → β}
+    (h : ∀ x y, f (x ⊓ y) ≤ f x ⊓ f y) : Monotone f :=
+  of_map_inf_le_left fun x y ↦ (h x y).trans inf_le_left
+
 theorem of_map_inf [SemilatticeInf α] [SemilatticeInf β] {f : α → β}
     (h : ∀ x y, f (x ⊓ y) = f x ⊓ f y) : Monotone f :=
-  fun x y hxy => inf_eq_left.1 <| by rw [← h, inf_eq_left.2 hxy]
+  of_map_inf_le fun x y ↦ (h x y).le
```
Ruben-VandeVelde, line 917: Is this one still worth keeping, then? If not, please deprecate

Mathlib/MeasureTheory/Function/SimpleFunc.lean
```diff
@@ -919,27 +919,28 @@ theorem restrict_const_lintegral (c : ℝ≥0∞) {s : Set α} (hs : MeasurableS
     ((const α c).restrict s).lintegral μ = c * μ s := by
   rw [restrict_lintegral_eq_lintegral_restrict _ hs, const_lintegral_restrict]
 
+@[gcongr]
+theorem lintegral_mono_fun {f g : α →ₛ ℝ≥0∞} (h : f ≤ g) : f.lintegral μ ≤ g.lintegral μ := by
+  refine Monotone.of_left_le_map_sup (f := (lintegral · μ)) (fun f g ↦ ?_) h
+  change ((pair f g).map Prod.fst).lintegral μ ≤ ((pair f g).map fun p ↦ p.1 ⊔ p.2).lintegral μ
```
Ruben-VandeVelde, line 925: What's the `change` doing? Please try to replace it

Mathlib/Order/Lattice.lean
urkud: This is a natural condition. It is used twice in the library. In both cases, it's possible to use `of_map_inf_le` but it turns `of_map_inf h` into `of_map_inf_le fun _ _ ↦ (h _ _).le`.

Mathlib/MeasureTheory/Function/SimpleFunc.lean
urkud: I turned this into `calc` and removed one defeq (ab)use. For the other one, we would need more `rfl` lemmas to rewrite and I think that this is an unrelated refactor (this defeq abuse was in the original proof).

RemyDegenne:
LGTM!
bors r+

PR #15511, opened 2024-08-05
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 22)

kim-em:


Mathlib/Analysis/BoxIntegral/Partition/Filter.lean
```diff
@@ -173,8 +173,7 @@ noncomputable section
 
 namespace BoxIntegral
 
-variable {ι : Type*} [Fintype ι] {I J : Box ι} {c c₁ c₂ : ℝ≥0} {r r₁ r₂ : (ι → ℝ) → Ioi (0 : ℝ)}
-  {π π₁ π₂ : TaggedPrepartition I}
+variable {ι : Type*} [Fintype ι] {I J : Box ι} {c c₁ c₂ : ℝ≥0}
```
sgouezel, line 176: I don't understand why this change is needed: these are variables that can only be imported if they're mentioned in the type, right?

Mathlib/RingTheory/SimpleModule.lean
```diff
@@ -217,19 +219,21 @@ protected theorem range (f : M →ₗ[R] N) : IsSemisimpleModule R (range f) :=
 
 section
 
-variable [Module S N] {σ : R →+* S} [RingHomSurjective σ] (l : M →ₛₗ[σ] N)
+variable {M'} [AddCommGroup M'] [Module R M'] {N'} [AddCommGroup N'] [Module S N']
```
sgouezel, line 222:
```suggestion
variable {M' : Type*} [AddCommGroup M'] [Module R M'] {N'} [AddCommGroup N'] [Module S N']
```
Otherwise, I fear it is interpreted as `Type _`, which can (seldom, but still) cause some weird things.

sgouezel (commented):
bors d+

Mathlib/Analysis/BoxIntegral/Partition/Filter.lean
semorrison: Yes, it's not strictly necessary. I found it less confusing having these closer to the point of us.

PR #15514, opened 2024-08-05
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 23)

kim-em:


Mathlib/Algebra/DirectSum/LinearMap.lean
```diff
@@ -122,7 +122,8 @@ lemma trace_comp_eq_zero_of_commute_of_trace_restrict_eq_zero
   rw [restrict_comp, trace_comp_eq_mul_of_commute_of_isNilpotent μ h_comm
     (f.isNilpotent_restrict_iSup_sub_algebraMap μ), hg, mul_zero]
 
-lemma mapsTo_biSup_of_mapsTo (s : Set ι) {f : Module.End R M} (hf : ∀ i, MapsTo f (N i) (N i)) :
+lemma mapsTo_biSup_of_mapsTo {ι} {N : ι → Submodule R M}
```
sgouezel, line 125:
```suggestion
lemma mapsTo_biSup_of_mapsTo {ι : Type*} {N : ι → Submodule R M}
```
Better be explicit (is `Sort` allowed? Will Lean decide that it has to be `Prop`?)

Mathlib/LinearAlgebra/BilinearForm/Properties.lean
```diff
@@ -426,7 +426,8 @@ theorem apply_dualBasis_right (B : BilinForm K V) (hB : B.Nondegenerate) (sym :
   rw [sym.eq, apply_dualBasis_left]
 
 @[simp]
-lemma dualBasis_dualBasis_flip (B : BilinForm K V) (hB : B.Nondegenerate) {ι}
+lemma dualBasis_dualBasis_flip [FiniteDimensional K V]
+    (B : BilinForm K V) (hB : B.Nondegenerate) {ι}
```
sgouezel, line 430:
```suggestion
    (B : BilinForm K V) (hB : B.Nondegenerate) {ι : Type*}
```

sgouezel (commented):
bors d+

Mathlib/Algebra/DirectSum/LinearMap.lean
semorrison: This problem is not as severe as it used to be: Lean can only change the meaning based on the signature, not only the body, now. But sure, it doesn't hurt to be explicit!

There are going to be more of these in future PRs!

semorrison:
bors merge

PR #15518, opened 2024-08-05
[Merged by Bors] - feat(CategoryTheory): opposite functors preserving finite products

dagurtomas:
If `F` preserves finite coproducts, then `F.op` preserves finite products, and the seven other variants for `leftOp`, `unop` etc.

Also remove some lean3-isms from docstrings

---


joelriou:
Thanks!

bors merge

PR #15519, opened 2024-08-05
[Merged by Bors] - fix: put `Nat.primeCounting` in a narrower scope

eric-wieser:
Without this, it is not possible to mix `!` notation for factorial and `π` notation for `Real.pi`.

More precisely; while the following works in `Mathlib/Analysis/SpecialFunctions/Stirling.lean`
```lean
lemma factorial_isEquivalent_stirling :
    (fun n ↦ n ! : ℕ → ℝ) ~[atTop] fun n ↦ Real.sqrt (2 * n * π) * (n / exp 1) ^ n := by
```
it produces an error about ambiguous syntax in a standalone file with `import Mathlib`:
```lean
import Mathlib

open scoped Real Nat Asymptotics
open Real

lemma factorial_isEquivalent_stirling :
    (fun n ↦ n ! : ℕ → ℝ) ~[atTop] fun n ↦ Real.sqrt (2 * n * π) * (n / exp 1) ^ n := by
```

Follows on from #13473

---

### DIFF
--- a/Mathlib/NumberTheory/PrimeCounting.lean
+++ b/Mathlib/NumberTheory/PrimeCounting.lean
@@ -27,8 +27,8 @@ are not prime, and so only at most `φ(k)/k` fraction of the numbers from `k` to
 
 ## Notation
 
-We use the standard notation `π` to represent the prime counting function (and `π'` to represent
-the reindexed version).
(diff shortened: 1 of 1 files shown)


riccardobrasca:
Thanks!

bors merge

PR #15521, opened 2024-08-05
[Merged by Bors] - chore: enable the setOption, longLine, missingEnd linters globally

grunweg:
With Lean 4.11.0-rc1, downstream projects can disable these linters by setting (e.g.) linter.setOption false in their `lakefile.lean`.

------

Not tested on downstream projects yet: help doing so is welcome.

- [x] depends on: #15616


joneugster:
I added the dependency that was suggested in the other PR. Will revisit once the other one is reviewed!

semorrison:
bors d+

grunweg:
Thanks!
bors merge

PR #15522, opened 2024-08-05
[Merged by Bors] - feat: monotonicity lemmas for OrdinalApprox

PhoenixIra:
This PR adds two lemmas about monotonicity for `lfpApprox` and `gfpApprox` each. I found them helpful when working with the API.

---

I am a bit unsure about the names and would welcome any suggestions. Especially to make them fit with the already present monotonicity lemma.

~~Moreover, I could now express the monotonicity for the functions in `lfpApprox_mono₁` in terms of `Monotone`. I welcome any suggestions to express it with `Monotone`.~~


Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -110,6 +110,41 @@ theorem lfpApprox_add_one (h : x ≤ f x) (a : Ordinal) :
     simp only [Set.mem_setOf_eq]
     use a
 
+theorem lfpApprox_mono₁ (h : f ≤ g) : lfpApprox f ≤ lfpApprox g := by
```
eric-wieser, line 113: Could you state this with `Monotone` too? Perhaps
```suggestion
theorem lfpApprox_mono_left : Monotone (lfpApprox : (α →o α) → _) := by
  intros f g h
```
is enough to make Lean happy.

eric-wieser (commented):
Thanks!

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
PhoenixIra: Yes, thanks! That works.

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -110,6 +110,41 @@ theorem lfpApprox_add_one (h : x ≤ f x) (a : Ordinal) :
     simp only [Set.mem_setOf_eq]
     use a
 
+theorem lfpApprox_mono_left : Monotone (lfpApprox : (α →o α) → _) := by
+  intro f g h x a
+  induction a using Ordinal.induction with
+  | h i ih =>
+    unfold lfpApprox
```
vihdzp, line 117: Having to unfold a definition over and over is somewhat of an antipattern. The issue here is that since your definition of `lfpApprox` is recursive, it's not unfolded by default, causing it not to be definitionally equal. A simple fix is to introduce a lemma like
```
theorem lfpApprox_def (a : Ordinal) :
    lfpApprox f x a = sSup ({ f (lfpApprox f x b) | (b : Ordinal) (_ : b < a) } ∪ {x}) := by
  rw [lfpApprox]
```
and rewrite using that lemma (which notably, is now searchable) instead of using `unfold`.

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -110,6 +110,41 @@ theorem lfpApprox_add_one (h : x ≤ f x) (a : Ordinal) :
...
+      forall_eq_or_imp, le_sup_left, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂,
+      true_and]
+    intro i' h_lt
+    apply le_sup_of_le_right
+    apply le_sSup_of_le
+    · use i'
+    · apply le_trans (h _)
+      simp only [OrderHom.toFun_eq_coe]
+      apply g.monotone
+      exact ih i' h_lt
```
vihdzp, line 129:
```suggestion
      exact g.monotone (ih i' h_lt)
```

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -230,6 +265,41 @@ theorem gfpApprox_add_one (h : f x ≤ x) (a : Ordinal) :
...
+      forall_eq_or_imp, inf_le_left, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂,
+      true_and]
+    intro i' h_lt
+    apply inf_le_of_right_le
+    apply sInf_le_of_le
+    · use i'
+    · apply le_trans (h _)
+      simp only [OrderHom.toFun_eq_coe]
+      apply g.monotone
+      exact ih i' h_lt
```
vihdzp, line 284:
```suggestion
  intro f g h
  have : (OrderHom.dual g) ≤ (OrderHom.dual f) := h
  exact lfpApprox_mono_left this
```

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -230,6 +265,41 @@ theorem gfpApprox_add_one (h : f x ≤ x) (a : Ordinal) :
...
+    simp only [exists_prop, Set.union_singleton, Set.mem_insert_iff, Set.mem_setOf_eq, sInf_insert,
+      forall_eq_or_imp, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂]
+    constructor
+    · exact inf_le_of_left_le h
+    · intro i' h_i'
+      apply inf_le_of_right_le
+      apply sInf_le_of_le
+      · use i'
+      · apply f.monotone
+        exact ih i' h_i'
```
vihdzp, line 301:
```suggestion
theorem gfpApprox_mono_mid : Monotone (gfpApprox f) :=
  fun _ _ h => lfpApprox_mono_mid (OrderHom.dual f) h
```

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -110,6 +110,41 @@ theorem lfpApprox_add_one (h : x ≤ f x) (a : Ordinal) :
...
+    simp only [exists_prop, Set.union_singleton, Set.mem_insert_iff, Set.mem_setOf_eq, sSup_insert,
+      forall_eq_or_imp, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂]
+    constructor
+    · exact le_sup_of_le_left h
+    · intro i' h_i'
+      apply le_sup_of_le_right
+      apply le_sSup_of_le
+      · use i'
+      · apply f.monotone
+        exact ih i' h_i'
```
vihdzp, line 146:
```suggestion
      · exact f.monotone (ih i' h_i')
```

vihdzp (commented):
Sorry this took so long. I would've reviewed earlier, but I had never heard about "ordinal approximants" to fixed points. For a good while I thought the ordinals themselves were the ones being approximated!

PhoenixIra:
Thank you very much for the review!
I think usually the theorem is called `constructive version of Tarski's fixed point theorem` as the documentation mentions. But since mathlib does not define ordinals constructively, we decided against that name and choose to highlight the approximation scheme via ordinals instead.

vihdzp:
As a separate thing, I've skimmed through the two references provided, and neither calls these values "ordinal approximants". I find the name to be quite misleading, as really, the approximations are simply ordinal-*indexed*.

Moreover, the file never mentions the actual result being built towards, that the fixed points of a monotone function on a complete lattice form a complete lattice themselves.

Do you think you could add this documentation to the file in a separate PR?

Do you think you could explain th

PhoenixIra:
I can do create a separate PR for the naming confusion. Do you expect this separate PR to only change the wording of "ordinal approximants" to "ordinal-indexed approximants", or would you also like to see some more details (which?) regarding the chosen name?

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
PhoenixIra: I added that theorem and refactored it in the previous existing theorems. Is this fine or should the second part be a separate PR?

vihdzp:
Just that change, and a better docstring for `lfpApprox` and `gfpApprox`, would go a long way.

By the way, I just noticed that [`fixedPoints.completeLattice`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/FixedPoints.html#fixedPoints.completeLattice) already exists in Mathlib. Is there some other result you're planning to build towards?

vihdzp:
FYI I just opened #16361 which simplifies a lot of the proofs that were already in the file.

PhoenixIra:
I have created the doc PR here: #16406

I do not plan to further make any connections between this theorem and [fixedPoints.completeLattice](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/FixedPoints.html#fixedPoints.completeLattice). I believe both proofs are used for different applications. It is sometimes nice to argue about pre fixed points (as Tarki's Theorem is enabling), but it is sometimes nice to be able to apply transfinite induction on fixed points (as Cousot-Cousot's Theorem is enabling).

I considered to maybe generalize the ordinal-indexed approximations to domains with a complete partial order instead of complete lattices - but I have no plan to do it yet.

PhoenixIra:
Moreover, I think it is redundant to give another construction of a complete lattice of fixed points in this file, as the supremum of fixed points is constructed analogous to Tarski's theorem, as far as I remember. The connection to the least fixed point and the greatest fixed point however are created in `lfpApprox_ord_eq_lfp` and `gfpApprox_ord_eq_gfp`.

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -177,7 +213,7 @@ theorem lfpApprox_le_of_mem_fixedPoints {a : α}
     (h_a : a ∈ fixedPoints f) (h_le_init : x ≤ a) (i : Ordinal) : lfpApprox f x i ≤ a := by
   induction i using Ordinal.induction with
   | h i IH =>
-    unfold lfpApprox
+    rw [lfpApprox_def]
```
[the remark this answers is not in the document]
PhoenixIra: It does. I have really no preference which to prefer here. Should I use `rw [lfpApprox]`?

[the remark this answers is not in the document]
YaelDillies: If it works, then I would suggest you use that instead of declaring a new lemma `lfpApprox_def`

YaelDillies (approved):
maintainer merge

Mathlib/SetTheory/Ordinal/FixedPointApproximants.lean
```diff
@@ -230,6 +261,14 @@ theorem gfpApprox_add_one (h : f x ≤ x) (a : Ordinal) :
     gfpApprox f x (a+1) = f (gfpApprox f x a) :=
   lfpApprox_add_one (OrderHom.dual f) x h a
 
+theorem gfpApprox_mono_left : Monotone (gfpApprox : (α →o α) → _) := by
+  intro f g h
+  have : (OrderHom.dual g) ≤ (OrderHom.dual f) := h
```
fpvandoorn, line 266:
```suggestion
  have : g.dual ≤ f.dual := h
```
Does this work? If not, at least remove the parentheses

fpvandoorn (commented):
One comment, otherwise LGTM.

bors d+

PR #15526, opened 2024-08-05
docs: tweak description about `MonoidHomClass`

alissa-tung:
---

---

I am curious about why exist two same description about ``When possible, instead of parametrizing results over `(f : M →* N)`,`` if my understanding is not wrong, the `→+` could be replaced by `→*`, if my understanding is wrong, I think we can put more description in where `AddMonoidHomClass` to classify the difference


joneugster:
Thanks for the PR! Unfortunately, our setup requires PRs to be made from a branch of the repo, not a fork, see [Creating a PR](https://leanprover-community.github.io/contribute/index.html#making-a-pull-request-pr).

As it's a simple typo I opened #15804 with the change you spotted to replace this PR.

PR #15527, opened 2024-08-05
[Merged by Bors] - feat(Topology): the partition of a space into fibers of a function

dagurtomas:
---

On its own this PR may seem strange, but this API is useful in #15321 to prove that discrete condensed sets are given by locally constant maps.


Mathlib/Logic/Function/FiberPartition.lean
```diff
@@ -0,0 +1,66 @@
...
+set, and the forgetful functor from the category of condensed sets to the category of sets
+(see PR #14027).
+-/
+
+variable {X Y Z : Type*} (f : Y → Z) (g : X → Y)
+
+namespace Function
+
+/-- The indexing set of the partition. -/
+def Fibers : Type _ := Set.range (fun (x : Set.range f) ↦ f ⁻¹' {x.val})
```
jcommelin, line 23: This won't work with the new `variable` algorithm. Please copy `f` syntax-explicitly into the statement:
```suggestion
def Fibers (f : Y → Z) : Type _ := Set.range (fun (x : Set.range f) ↦ f ⁻¹' {x.val})
```

Mathlib/Logic/Function/FiberPartition.lean
```diff
@@ -0,0 +1,66 @@
...
+  convert x.prop
+
+lemma mk_image (y : Y) : (Fibers.mk f y).image = f y :=
+  (map_eq_image (x := mkSelf f y)).symm
+
+lemma mem_iff_eq_image (y : Y) (a : Fibers f) : y ∈ a.val ↔ f y = a.image :=
+  ⟨fun h ↦ a.map_eq_image _ ⟨y, h⟩, fun h ↦ by rw [a.eq_fiber_image]; exact h⟩
+
+/-- An arbitrary element of `a : Fibers f`. -/
+noncomputable def preimage (a : Fibers f) : Y := a.2.choose.2.choose
```
jcommelin, line 55: Can you please add a proof of `Set.Nonempty a.1`?

Mathlib/Logic/Function/FiberPartition.lean
```diff
@@ -0,0 +1,66 @@
...
+
+lemma map_preimage_eq_image (a : Fibers f) : f a.preimage = a.image := a.2.choose.2.choose_spec
```
jcommelin, line 57: Should this be simp-lemma?

dagurtomas: I don't think so. It's used once in each directions in the application, as far as I can tell

jcommelin (commented):
Thanks :tada:

bors merge

PR #15533, opened 2024-08-05
[Merged by Bors] - fix: `withMainContext` in `type_check` tactic

znssong:
Fixing a problem about `type_check` tactic: If you don't use `withMainContext`, you can't use local hypotheses and definitions produced by tactics before the `type_check` tactic, only can use hypotheses written on `theorem` header.

### DIFF
--- a/Mathlib/Tactic/TypeCheck.lean
+++ b/Mathlib/Tactic/TypeCheck.lean
@@ -4,17 +4,19 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Jireh Loreaux
 -/
 import Lean.Elab.Tactic.Basic
+import Lean.Elab.SyntheticMVars
 
 /-!
 # The `type_check` tactic
 Define the `type_check` tactic: it type checks a given expression, and traces its type.
 -/
 
-open Lean Meta
+open Lean Elab Meta
 
 /-- Type check the given expression, and trace its type. -/
 elab tk:"type_check " e:term : tactic => do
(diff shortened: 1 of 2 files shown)


Mathlib/Tactic/TypeCheck.lean
```diff
@@ -4,17 +4,19 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+open Lean Elab Meta
 
 /-- Type check the given expression, and trace its type. -/
 elab tk:"type_check " e:term : tactic => do
-  let e ← Lean.Elab.Term.elabTerm e none
-  check e
-  let type ← inferType e
-  Lean.logInfoAt tk f!"{← Lean.instantiateMVars type}"
+  Tactic.withMainContext do
+    let e ← Term.elabTermAndSynthesize e none
```
eric-wieser, line 19: Can you explain the change to `andSynthesize`?

meow-sister: For example, if you use `elabTerm`, `type_check [1]` will get `List ?m.25`, not `List Nat`.

eric-wieser: Could you add a test for that?

meow-sister: OK, thanks.

j-loreaux:
Thanks! I wrote this before I understood anything about `withMainContext`!

bors merge

PR #15539, opened 2024-08-06
[Merged by Bors] - feat(Combinatorics/SimpleGraph): Add lemma about number of odd subcomponents

pimotte:
Added lemma that the number of odd components contained within a component
in a bigger graph is odd iff the bigger component is odd. In preparation for Tuttes.

---

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Tutte's.20theorem)


Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
```diff
@@ -184,6 +184,22 @@ lemma odd_card_iff_odd_components : Odd (Nat.card V) ↔
   rw [Nat.card_eq_fintype_card, Fintype.card_ofFinset]
   exact (Finset.odd_sum_iff_odd_card_odd (fun x : G.ConnectedComponent ↦ Nat.card x.supp))
 
+lemma ConnectedComponent.odd_card_supp_iff_odd_subcomponents {G'} [DecidableRel G.Adj]
+    [DecidableRel G'.Adj] (h : G ≤ G') (c' : ConnectedComponent G') :
+    Odd (Nat.card c'.supp) ↔ Odd (Nat.card
+    ({(c : {c : ConnectedComponent G | c.supp ⊆ c'.supp}) | Odd (Nat.card c.1.supp) })) := by
```
YaelDillies, line 190: Any reason why you are nesting set comprehensions here?
```suggestion
    {c : ConnectedComponent G | c.supp ⊆ c'.supp ∧ Odd (Nat.card c.1.supp)}) := by
```

Mathlib/Combinatorics/SimpleGraph/Path.lean
```diff
@@ -961,6 +961,24 @@ lemma mem_coe_supp_of_adj {v w : V} {H : Subgraph G} {c : ConnectedComponent H.c
...
 
+lemma connectedComponentMk_supp_subset_supp {G'} {v : V} (h : G ≤ G') (c' : G'.ConnectedComponent)
+    (hc' : v ∈ c'.supp) : (G.connectedComponentMk v).supp ⊆ c'.supp := by
+  intro v' hv'
+  simp only [mem_supp_iff, ConnectedComponent.eq] at hv' ⊢
+  rw [ConnectedComponent.sound (hv'.mono h)]
+  exact hc'
+
+lemma union_supp_eq_supp {G G' : SimpleGraph V} (h : G ≤ G') (c' : ConnectedComponent G') :
+    ⋃ c : {c : ConnectedComponent G | c.supp ⊆ c'.supp}, c.1.supp = c'.supp := by
```
YaelDillies, line 972: Maybe
```suggestion
lemma biUnion_supp_eq_supp {G G' : SimpleGraph V} (h : G ≤ G') (c' : ConnectedComponent G') :
    ⋃ (c : ConnectedComponent G) (hc : c.supp ⊆ c'.supp}, c.1.supp = c'.supp := by
```

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
YaelDillies, line 190: Do you think you could golf the lemma above using this one? (by setting `G' := ⊤`

pimotte: Mainly because I couldn't figure out how to rewrite between these two forms in using it :see_no_evil:

YaelDillies: I think you never need your original form. You must have thought you needed it because of your statement of `union_supp_eq_supp`, but my suggestion below should fix this

Mathlib/Combinatorics/SimpleGraph/Path.lean
pimotte: I tried this, two remarks:

- Is biUnion the right name for this? This is just a double union, whereas the biUnion is over a set, right?
- I'm having trouble using the lemma because the motive is incorrect. Previously I managed to fix this by adding the `DecidablePred`, but I don't really know how to get there now.

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
pimotte: Done. In number of lines it's not much better, but I'm guessing there's some golfing possibilities

Mathlib/Combinatorics/SimpleGraph/Path.lean
pimotte: Managed to get this to work, but I'm guessing we want to do some golfing/extraction before merging. In particular the argument to Finset.filter_congr seems a prime candidate. Any thoughts what we would want to do there?

YaelDillies: > * Is biUnion the right name for this? This is just a double union, whereas the biUnion is over a set, right?

`biUnion` is precisely the special case of `iUnion₂` where the inner `iUnion` runs over a Prop

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
```diff
@@ -172,17 +172,56 @@ instance instDecidableMemSupp (c : G.ConnectedComponent) (v : V) : Decidable (v
...
-  rw [Nat.card_eq_fintype_card]
-  simp only [← (set_fintype_card_eq_univ_iff _).mpr G.iUnion_connectedComponentSupp,
-    ConnectedComponent.mem_supp_iff, Fintype.card_subtype_compl,
-    ← Set.toFinset_card, Set.toFinset_iUnion ConnectedComponent.supp]
-  rw [Finset.card_biUnion
-    (fun x _ y _ hxy ↦ Set.disjoint_toFinset.mpr (pairwise_disjoint_supp_connectedComponent _ hxy))]
-  simp_rw [Set.toFinset_card, ← Nat.card_eq_fintype_card]
-  rw [Nat.card_eq_fintype_card, Fintype.card_ofFinset]
-  exact (Finset.odd_sum_iff_odd_card_odd (fun x : G.ConnectedComponent ↦ Nat.card x.supp))
+  if h : Nonempty V then
```
YaelDillies, line 212:
```suggestion
  cases isEmpty_or_nonempty V
```
is more idiomatic

pimotte: Fixed:)

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
```diff
@@ -172,17 +172,55 @@ instance instDecidableMemSupp (c : G.ConnectedComponent) (v : V) : Decidable (v
...
-  rw [Nat.card_eq_fintype_card]
-  simp only [← (set_fintype_card_eq_univ_iff _).mpr G.iUnion_connectedComponentSupp,
-    ConnectedComponent.mem_supp_iff, Fintype.card_subtype_compl,
-    ← Set.toFinset_card, Set.toFinset_iUnion ConnectedComponent.supp]
-  rw [Finset.card_biUnion
-    (fun x _ y _ hxy ↦ Set.disjoint_toFinset.mpr (pairwise_disjoint_supp_connectedComponent _ hxy))]
-  simp_rw [Set.toFinset_card, ← Nat.card_eq_fintype_card]
-  rw [Nat.card_eq_fintype_card, Fintype.card_ofFinset]
-  exact (Finset.odd_sum_iff_odd_card_odd (fun x : G.ConnectedComponent ↦ Nat.card x.supp))
+  cases isEmpty_or_nonempty V with
```
YaelDillies, line 212: I guess here I meant `cases'` since we don't want to name the hypothesis
```suggestion
  cases' isEmpty_or_nonempty V
```

pimotte: Changed, do I didn't know how to do the second part without naming the hypothesis, so I added a `with` to it

pimotte:
Just backmerged master. Originally I refactored `odd_card_iff_odd_components`, but it had been shown that it could do without [DecidableEq V], and I do think it's needed for the subcomponent variant, so I kept the old stronger result.

YaelDillies (commented):
Sorry, I started a review two weeks ago, but it was apparently never published

Mathlib/Combinatorics/SimpleGraph/Connectivity/WalkCounting.lean
```diff
@@ -209,11 +209,10 @@ lemma ConnectedComponent.odd_card_supp_iff_odd_subcomponents {G'} [DecidableRel
 
 lemma odd_card_iff_odd_components : Odd (Nat.card V) ↔
     Odd (Nat.card ({(c : ConnectedComponent G) | Odd (Nat.card c.supp)})) := by
-  cases isEmpty_or_nonempty V with
-  | inl h => simp only [Nat.card_of_isEmpty, Nat.card_eq_fintype_card, Fintype.card_eq_zero,
+  cases' isEmpty_or_nonempty V with _ h
```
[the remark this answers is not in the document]
pimotte: I don't think this is relevant anymore since I reverted back on the old proof to get rid of the dependency on `[DecidableEq V]`, but I'll keep it in mind for future applications

YaelDillies (approved):
Golfed the proof a bit. Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15542, opened 2024-08-06
[Merged by Bors] - doc(Probability/Notation): tell user to open ProbabilityTheory to use these notations

henkkuli:
Previously this requirement was not documented anywhere apart from the source.

---

### DIFF
--- a/Mathlib/Probability/Notation.lean
+++ b/Mathlib/Probability/Notation.lean
@@ -25,7 +25,8 @@ We note that the notation `∂P/∂Q` applies to three different cases, namely,
 
 - `ℙ` is a notation for `volume` on a measured space.
 
-To use these notations, you need to use `open ProbabilityTheory`.
+To use these notations, you need to use `open scoped ProbabilityTheory`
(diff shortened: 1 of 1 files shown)


Mathlib/Probability/Notation.lean
```diff
@@ -24,6 +24,8 @@ We note that the notation `∂P/∂Q` applies to three different cases, namely,
 `MeasureTheory.ComplexMeasure.rnDeriv`.
 
 - `ℙ` is a notation for `volume` on a measured space.
+
+To use these notations, you need to use `open ProbabilityTheory`.
```
adomani, line 28:
```suggestion
To use these notations, you need to use `open scoped ProbabilityTheory`
or `open ProbabilityTheory`.
```

henkkuli: Right, I guess it is better to tell the user to use the more restricted `open scoped` over `open`.

b-mehta:
Thanks!

bors merge

PR #9606, opened 2024-01-10
[Merged by Bors] - chore: refactor of Algebra/Group/Defs to reduce imports

kim-em:
We are not that far from the point that `Algebra/Group/Defs` depends on nothing significant besides `simps` and `to_additive`.

This removes from `Mathlib.Algebra.Group.Defs` the dependencies on
* `Mathlib.Tactic.Basic` (which is a grab-bag of random stuff)
* `Mathlib.Init.Algebra.Classes` (which is ancient and half-baked)
* `Mathlib.Logic.Function.Basic` (not particularly important, but it is barely used in this file)

The goal is to avoid all unnecessary imports to set up the *definitions* of basic algebraic structures. 

We also separate out `Mathlib.Tactic.TypeStar` and `Mathlib.Tactic.Lemma` as prerequisites to `Mathlib.Tactic.Basic`, but which can be imported separately when the rest of `Mathlib.Tactic.Basic` is not needed.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15544, opened 2024-08-06
[Merged by Bors] - refactor(Cyclic): redefine, move definition

urkud:
Redefine `IsCyclic` without `Subgroup` dependency, move the definition to `Algebra/Group/Defs`.

---
I'm going to prove that certain groups aren't cyclic, and the relevant file doesn't import `SpecificGroups/Cyclic`.
I added `import Mathlib.Logic.Function.Defs` to `Algebra.Group.Defs` to have `Surjective`.


mattrobball:
@semorrison what was the rationale for `assert_not_exists Function.Injective`?

urkud:
It was added in #9606 while removing `import Mathlib.Logic.Function.Basic` which is a bit heavier than `Mathlib.Logic.Function.Defs`.

j-loreaux:
We've worked hard to get the dependencies for `Algebra.Group.Defs` very low. While I agree `Logic.Function.Defs` is not heavy, it's relative weight for such a low-level file is still relatively high. So, personally, I'd prefer to have another solution here.

Maybe expand the definition of `Function.Surjective` locally in this file, and add the actual `Function.Surjective` lemma in a later file?

urkud:
@j-loreaux 
What if I move everything but the *definitions* from `Logic/Function/Defs`, so that it becomes a very light dependency?

mattrobball:
Is there harm in just making a new file?

urkud:
BTW, the only new Mathlib imports that come with `Function.Defs` are `Tactic.AdaptationNote` and `Attr.Register`. Probably, the script reports Batteries imports too.

urkud:
> Is there harm in just making a new file?

What do you suggest to move to a new file?

mattrobball:
`IsCyclic/IsAddCyclic` to `Algebra.Group.IsCyclic`?

urkud:
I would expect that `Algebra/Group/IsCyclic` is a file with basic properties of `IsCyclic`, not a file with the definition.

urkud:
Another option is to move everything but definitions from `Logic/Function/Defs` to a new file (or to `Function/Basic` if every file that needs them imports `Basic`).

urkud:
@j-loreaux What are the benefits of not importing `Function/Defs` in `Algebra/Group/Defs`?

j-loreaux:
Ah, I didn't see your earlier comment. I think I'm okay with the new imports pulled in by `Function.Defs`, but I would like other input.

My goal here is just to be very stingy with respect to imports of low-level files so that we don't end up slowly migrating back to the old paradigm.

j-loreaux:
Whoever looks at this, tha main question is should we import `Logic.Function.Defs` into `Algebra.Group.Defs`. Ultimately, I'm okay with it, but @mattrobball suggested moving to another file, which I also think is a good option. Please weigh in or merge.

maintainer merge

semorrison:
bors merge

PR #15546, opened 2024-08-06
[Merged by Bors] - feat(Combinatorics/SimpleGraph): Add definition of disjoint sum of graphs

IvanRenison:
---


Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,77 @@
...
+* `G + H`: The disjoint sum of `G` and `H`.
+-/
+
+variable {α β γ : Type*}
+
+namespace SimpleGraph
+
+/-- Disjoint sum of `G` and `H`. -/
+@[simps!]
+def Sum (G : SimpleGraph α) (H : SimpleGraph β) : SimpleGraph (α ⊕ β) where
```
Command-Master, line 31: I think this should be protected, as it can collide with the builtin `Sum` if you open `SimpleGraph`

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,77 @@
...
+    | Sum.inr u, Sum.inr v => H.Adj u v
+    | _, _ => false
+  symm u v := match u, v with
+    | Sum.inl u, Sum.inl v => G.adj_symm
+    | Sum.inr u, Sum.inr v => H.adj_symm
+    | Sum.inl _, Sum.inr _ | Sum.inr _, Sum.inl _ => id
+  loopless u := by cases u <;> simp
+
+/-- Disjoint sum of `G` and `H`. -/
+infixl:60 " + " => SimpleGraph.Sum
```
b-mehta, line 43: I don't think it's a good idea to overload the + operator in this way. Instead, does it work to give an instance for `HAdd (SimpleGraph α) (SimpleGraph β) (SimpleGraph (α ⊕ β))`?

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,77 @@
...
+  intro u v
+  cases u <;> cases v <;> rename_i u v
+  · cases u <;> cases v <;> simp
+  · cases u <;> simp
+  · cases v <;> simp
+  · simp⟩
+
+/-- The embedding of `G` into `G + H`. -/
+@[simps]
+def SumLeft : G ↪g G + H where
```
b-mehta, line 65: I'd prefer a name like `SimpleGraph.inl` here, or maybe `SimpleGraph.Sum.inl`

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,77 @@
...
+`H : SimpleGraph β` is a graph on `α ⊕ β` where `u` and `v` are adjacent if and only if they are
+both in `G` and adjacent in `G`, or they are both in `H` and adjacent in `H`.
+
+## Main declarations
+
+* `SimpleGraph.Sum`: The disjoint sum of graphs.
+
+## Notation
+
+* `G + H`: The disjoint sum of `G` and `H`.
```
YaelDillies, line 22: I don't really like the `+` notation. . Could you use `⊎` or something similar instead?

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
```
YaelDillies, line 22: You forgot to update this!

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+  intro u v
+  cases u <;> cases v <;> rename_i u v
+  · cases u <;> cases v <;> simp
+  · cases u <;> simp
+  · cases v <;> simp
+  · simp⟩
+
+/-- The embedding of `G` into `G ⊎ H`. -/
+@[simps]
+protected def Sum.inl : G ↪g G ⊎ H where
```
YaelDillies, line 64: `protected` only prevents opening the last namespace, so it's useless here.
```suggestion
def SumInl : G ↪g G ⊎ H where
```
Furthermore, this name doesn't match the other ones above

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+  toFun u := _root_.Sum.inl u
+  inj' u v := by simp
+  map_rel_iff' := by simp
+
+/-- The embedding of `H` into `G ⊎ H`. -/
+@[simps]
+protected def Sum.inr : H ↪g G ⊎ H where
```
YaelDillies, line 71: Same here

YaelDillies (approved):
Great, thanks! Once you have solved those comments,

maintainer merge

Mathlib/Combinatorics/SimpleGraph/Sum.lean
b-mehta: I think `Sum.inl` is better here, so that once we're in the `SimpleGraph.Sum` namespace, we can write `inl` in analogy to regular `Sum.inl`. See also [CategoryTheory.Sum.inr_](https://leanprover-community.github.io/mathlib4_docs/Mathlib/CategoryTheory/Sums/Basic.html#CategoryTheory.Sum.inr_)

b-mehta: (But I agree wrt `protected`, and making sure the names are consistent through the file)

IvanRenison: I have changed it to `SumInl`, but if you prefer I can use the `.` version everywhere.

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+    | Sum.inl _, Sum.inr _ | Sum.inr _, Sum.inl _ => id
+  loopless u := by cases u <;> simp
+
+@[inherit_doc] infixl:60 " ⊎ " => SimpleGraph.Sum
+
+variable {G : SimpleGraph α} {H : SimpleGraph β}
+
+/-- The disjoint sum is commutative up to isomorphism. `Equiv.sumComm` as a graph isomorphism. -/
+@[simps!]
+def SumComm : G ⊎ H ≃g H ⊎ G := ⟨Equiv.sumComm α β, by
```
urkud, line 48:
```suggestion
def Iso.sumComm : G ⊎ H ≃g H ⊎ G := ⟨Equiv.sumComm α β, by
```

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+    | Sum.inl u, Sum.inl v => G.Adj u v
+    | Sum.inr u, Sum.inr v => H.Adj u v
+    | _, _ => false
+  symm u v := match u, v with
+    | Sum.inl u, Sum.inl v => G.adj_symm
+    | Sum.inr u, Sum.inr v => H.adj_symm
+    | Sum.inl _, Sum.inr _ | Sum.inr _, Sum.inl _ => id
+  loopless u := by cases u <;> simp
+
+@[inherit_doc] infixl:60 " ⊎ " => SimpleGraph.Sum
```
urkud, line 42: Should we use `⊕g` instead? @YaelDillies What do you think? Internet search says that `⊎` is called "multiset union" symbol and is used for `+` on `Multiset`s. Will it cause confusion?

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+
+/-- The disjoint sum is commutative up to isomorphism. `Equiv.sumComm` as a graph isomorphism. -/
+@[simps!]
+def SumComm : G ⊎ H ≃g H ⊎ G := ⟨Equiv.sumComm α β, by
+  intro u v
+  cases u <;> cases v <;> simp⟩
+
+/-- The disjoint sum is associative up to isomorphism. `Equiv.sumAssoc` as a graph isomorphism. -/
+@[simps!]
+def SumAssoc {I : SimpleGraph γ} : (G ⊎ H) ⊎ I ≃g G ⊎ (H ⊎ I) := ⟨Equiv.sumAssoc α β γ, by
```
urkud, line 54:
```suggestion
def Iso.sumAssoc {I : SimpleGraph γ} : (G ⊎ H) ⊎ I ≃g G ⊎ (H ⊎ I) := ⟨Equiv.sumAssoc α β γ, by
```

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+  intro u v
+  cases u <;> cases v <;> rename_i u v
+  · cases u <;> cases v <;> simp
+  · cases u <;> simp
+  · cases v <;> simp
+  · simp⟩
+
+/-- The embedding of `G` into `G ⊎ H`. -/
+@[simps]
+def SumInl : G ↪g G ⊎ H where
```
urkud, line 64:
```suggestion
def Embedding.sumInl : G ↪g G ⊎ H where
```

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+  toFun u := _root_.Sum.inl u
+  inj' u v := by simp
+  map_rel_iff' := by simp
+
+/-- The embedding of `H` into `G ⊎ H`. -/
+@[simps]
+def SumInr : H ↪g G ⊎ H where
```
urkud, line 71:
```suggestion
def Embedding.sumInr : H ↪g G ⊎ H where
```

urkud:
Thanks! :tada: I'm not sure about notation and suggested another round of renames. Sorry that we (reviewers) failed to quickly agree on this. Delegating to Yael.
bors d=@YaelDillies

Mathlib/Combinatorics/SimpleGraph/Sum.lean
b-mehta: I prefer `⊕g` over `⊎` here too - it looks closer to the informal notation, and aligns with existing Lean notation for disjoint union.

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+* `G ⊕g H`: The disjoint sum of `G` and `H`.
+-/
+
+variable {α β γ : Type*}
+
+namespace SimpleGraph
+
+/-- Disjoint sum of `G` and `H`. -/
+@[simps!]
+protected def Sum (G : SimpleGraph α) (H : SimpleGraph β) : SimpleGraph (α ⊕ β) where
```
urkud, line 31: One last rename:
```suggestion
protected def sum (G : SimpleGraph α) (H : SimpleGraph β) : SimpleGraph (α ⊕ β) where
```

Mathlib/Combinatorics/SimpleGraph/Sum.lean
```diff
@@ -0,0 +1,76 @@
...
+    | Sum.inl u, Sum.inl v => G.Adj u v
+    | Sum.inr u, Sum.inr v => H.Adj u v
+    | _, _ => false
+  symm u v := match u, v with
+    | Sum.inl u, Sum.inl v => G.adj_symm
+    | Sum.inr u, Sum.inr v => H.adj_symm
+    | Sum.inl _, Sum.inr _ | Sum.inr _, Sum.inl _ => id
+  loopless u := by cases u <;> simp
+
+@[inherit_doc] infixl:60 " ⊕g " => SimpleGraph.Sum
```
urkud, line 42:
```suggestion
@[inherit_doc] infixl:60 " ⊕g " => SimpleGraph.sum
```

PR #15548, opened 2024-08-06
[Merged by Bors] - feat (MeasureTheory.Decomposition.RadonNikodym): `rnDeriv_mul_rnDeriv'`

LorenzoLuccioli:
Add `rnDeriv_mul_rnDeriv'`, an alternative version of `rnDeriv_mul_rnDeriv`. The difference is that the new version requires the absolute continuity of the second measure wrt the third (as opposed to the first wrt the second), but the thesis is true a.e. wrt the second measure (as opposed to wrt the third).


JasonKYi:
Thanks!
maintainer merge

PR #15552, opened 2024-08-06
[Merged by Bors] - feat(Topology/Maps): preimage of a dense set

urkud:
... under an open map is dense.

---


j-loreaux:
Thanks! 

bors merge

PR #15563, opened 2024-08-06
[Merged by Bors] - feat(Topology/Algebra): add `smul_mem_nhds_smul`

urkud:
- add `smul_mem_nhds_smul_iff` and `smul_mem_nhds_smul`, deprecate `smul_mem_nhds`;
- move 3 lemmas up in the import graph;
- add `smul_mem_nhds_self`;
- rename `set_smul_mem_nhds_smul_iff` to `smul_mem_nhds_smul_iff₀`, deprecate the old lemma;
- similarly with `set_smul_mem_nhds_smul` and `smul_mem_nhds_smul₀`;

---


Mathlib/Topology/Algebra/ConstMulAction.lean
```diff
@@ -240,6 +240,19 @@ theorem Dense.smul (c : G) {s : Set α} (hs : Dense s) : Dense (c • s) := by
...
+theorem IsOpen.smul_left {s : Set G} {t : Set α} (ht : IsOpen t) : IsOpen (s • t) := by
+  rw [← iUnion_smul_set]
+  exact isOpen_biUnion fun a _ => ht.smul _
+
+@[to_additive]
+theorem subset_interior_smul_right {s : Set G} {t : Set α} : s • interior t ⊆ interior (s • t) :=
+  interior_maximal (Set.smul_subset_smul_left interior_subset) isOpen_interior.smul_left
+
+@[to_additive (attr := simp)]
+theorem smul_mem_nhds {t : Set α} (g : G) {a : α} : g • t ∈ 𝓝 (g • a) ↔ t ∈ 𝓝 a :=
```
eric-wieser, line 253: Should this be `set_smul_mem_nhds_smul_iff`? The one below could gain a subscript zero suffix.

urkud: We have both `set_smul` and `smul_set`. `Data/Set/Pointwise/SMul` uses both. Should we normalize?

eric-wieser: Ah sorry, I meant `smul_mem_nhds_smul_iff`. I think we can put off normalizing the `set` part

urkud: What about the new names?

Mathlib/Topology/Algebra/ConstMulAction.lean
```diff
@@ -474,19 +494,18 @@ section MulAction
 variable {G₀ : Type*} [GroupWithZero G₀] [MulAction G₀ α] [TopologicalSpace α]
   [ContinuousConstSMul G₀ α]
 
--- Porting note: generalize to a group action + `IsUnit`
```
eric-wieser, line 477: I think this should be kept as a TODO about `IsUnit`

Mathlib/Topology/Algebra/ConstMulAction.lean
```diff
@@ -474,19 +494,18 @@ section MulAction
...
-/-- Scalar multiplication preserves neighborhoods. -/
-theorem set_smul_mem_nhds_smul {c : G₀} {s : Set α} {x : α} (hs : s ∈ 𝓝 x) (hc : c ≠ 0) :
-    c • s ∈ 𝓝 (c • x : α) := by
-  rw [mem_nhds_iff] at hs ⊢
-  obtain ⟨U, hs', hU, hU'⟩ := hs
-  exact ⟨c • U, Set.smul_set_mono hs', hU.smul₀ hc, Set.smul_mem_smul_set hU'⟩
+/-- Scalar multiplication by a nonzero scalar preserves neighborhoods. -/
```
urkud, line 497:
```suggestion
-- Porting note: generalize to a monoid action + `IsUnit`
/-- Scalar multiplication by a nonzero scalar preserves neighborhoods. -/
```

Mathlib/Topology/Algebra/ConstMulAction.lean
urkud: Restored.

Mathlib/Topology/Algebra/ConstMulAction.lean
```diff
@@ -240,6 +240,26 @@ theorem Dense.smul (c : G) {s : Set α} (hs : Dense s) : Dense (c • s) := by
...
+
+@[to_additive]
+theorem subset_interior_smul_right {s : Set G} {t : Set α} : s • interior t ⊆ interior (s • t) :=
+  interior_maximal (Set.smul_subset_smul_left interior_subset) isOpen_interior.smul_left
+
+@[to_additive (attr := simp)]
+theorem smul_mem_nhds_smul {t : Set α} (g : G) {a : α} : g • t ∈ 𝓝 (g • a) ↔ t ∈ 𝓝 a :=
+  (Homeomorph.smul g).openEmbedding.image_mem_nhds
+
+@[deprecated (since := "2024-08-06")] alias ⟨_, smul_mem_nhds⟩ := smul_mem_nhds_smul
```
j-loreaux, line 256: I think I would like to keep this around instead of deprecating.

Mathlib/Topology/Algebra/ConstMulAction.lean
```diff
@@ -474,19 +494,19 @@ section MulAction
...
-  refine ⟨fun h => ?_, fun h => set_smul_mem_nhds_smul h hc⟩
-  rw [← inv_smul_smul₀ hc x, ← inv_smul_smul₀ hc s]
-  exact set_smul_mem_nhds_smul h (inv_ne_zero hc)
+@[deprecated (since := "2024-08-06")]
+alias set_smul_mem_nhds_smul_iff := smul_mem_nhds_smul₀
+
+@[deprecated  smul_mem_nhds_smul₀ (since := "2024-08-06")]
+theorem set_smul_mem_nhds_smul {c : G₀} {s : Set α} {x : α} (hs : s ∈ 𝓝 x) (hc : c ≠ 0) :
+    c • s ∈ 𝓝 (c • x : α) :=
+  (smul_mem_nhds_smul₀ hc).2 hs
```
j-loreaux, line 509: can we just deprecate this, but leave an undeprecated alias with the name `smul_mem_nhds₀`?

j-loreaux (commented):
otherwise LGTM

urkud:
@j-loreaux Do you mean that you want to have implications as well as iffs? Then I can add `_iff` to new lemma names. I think that we should use`smul` twice in the name.

j-loreaux:
Right, I want implications too (at least the ones that go the usual direction)

j-loreaux (commented):
bors d+

PR #15565, opened 2024-08-06
[Merged by Bors] - feat: Set.iUnion over Sum

scholzhannah:
---

### DIFF
--- a/Mathlib/Data/Set/Lattice.lean
+++ b/Mathlib/Data/Set/Lattice.lean
@@ -686,11 +686,9 @@ theorem iInter_iInter_eq_or_left {b : β} {p : β → Prop} {s : ∀ x : β, x =
     ⋂ (x) (h), s x h = s b (Or.inl rfl) ∩ ⋂ (x) (h : p x), s x (Or.inr h) := by
   simp only [iInter_or, iInter_inter_distrib, iInter_iInter_eq_left]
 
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Set/Lattice.lean
```diff
@@ -686,6 +686,12 @@ theorem iInter_iInter_eq_or_left {b : β} {p : β → Prop} {s : ∀ x : β, x =
     ⋂ (x) (h), s x h = s b (Or.inl rfl) ∩ ⋂ (x) (h : p x), s x (Or.inr h) := by
   simp only [iInter_or, iInter_inter_distrib, iInter_iInter_eq_left]
 
+lemma iUnion_sum {s : α ⊕ β → Set γ} : ⋃ x, s x = (⋃ x, s (.inl x)) ∪ ⋃ x, s (.inr x) := by
+  ext; simp only [mem_iUnion, Sum.exists, mem_union]
```
urkud, line 690: Didn't test but this should work, same below with `iInf_sum`. I think that set lemmas that are special cases of complete lattice lemmas should reuse the proofs.
```suggestion
lemma iUnion_sum {s : α ⊕ β → Set γ} : ⋃ x, s x = (⋃ x, s (.inl x)) ∪ ⋃ x, s (.inr x) :=
  iSup_sum
```

scholzhannah: That makes sense, thank you!

j-loreaux:
bors merge

PR #15566, opened 2024-08-06
[Merged by Bors] - feat(Condensed): colimit characterization of discrete condensed sets

dagurtomas:
This PR provides the necessary API to prove that a condensed set `X` is discrete if and only if
for every profinite set `S = limᵢSᵢ`, `X(S) ≅ colimᵢX(Sᵢ)`, and the analogous result for light
condensed sets.

---

- [x] depends on: #14382 
- [x] depends on: #15321 
- [x] depends on: #17075 
- [x] depends on: #17380 
- [x] depends on: #17381 
- [x] depends on: #17383 


dagurtomas:
!bench

Mathlib/Condensed/Discrete/Colimit.lean
```diff
@@ -0,0 +1,373 @@
...
+    obtain ⟨j, h⟩ := Profinite.exists_locallyConstant.{_, u} c hc f
+    exact ⟨⟨j⟩, h⟩
+  · intro ⟨i⟩ ⟨j⟩ (fi : LocallyConstant _ _) (fj : LocallyConstant _ _)
+      (h : fi.comap (c.π.app i) = fj.comap (c.π.app j))
+    obtain ⟨k, ki, kj, _⟩ := IsCofilteredOrEmpty.cone_objs i j
+    refine ⟨⟨k⟩, ki.op, kj.op, ?_⟩
+    dsimp only [comp_obj, op_obj, functorToPresheaves_obj_obj, CompHausLike.coe_of,
+      Functor.comp_map, op_map, Quiver.Hom.unop_op, functorToPresheaves_obj_map]
+    -- Note: we might want to remove the `simps` attribute from `FintypeCat.toProfinite`; keeping
+    -- `toProfinite_obj` in the `dsimp` block above causes the following `ext` to fail.
```
dagurtomas, line 56: I have a followup PR #17155 which removes the simp lemma `toProfinite_obj`

dagurtomas:
!bench

Mathlib/Condensed/Discrete/Colimit.lean
```diff
@@ -0,0 +1,373 @@
...
+    simp only [comp_obj, op_obj, finYoneda_obj, Functor.comp_map, op_map]
+    ext
+    simp only [types_comp_apply, isoFinYonedaComponents_hom, finYoneda_map, op_obj,
+      Function.comp_apply, Types.productLimitCone, const_obj_obj, fintypeCatAsCofan, Cofan.mk_pt,
+      cofan_mk_inj, Fan.mk_pt, Fan.mk_π_app, ← FunctorToTypes.map_comp_apply]
+    rfl
+
+/--
+A presheaf `F`, which takes a profinite set written as a cofiltered limit to the corresponding
+colimit, is isomorphic to the presheaf `LocallyConstant - F(*)`.
```
jcommelin, line 187:
```suggestion
colimit, is isomorphic to the presheaf `LocallyConstant _ F(*)`.
```
Or did you mean something else?

Mathlib/Condensed/Discrete/Colimit.lean
```diff
@@ -0,0 +1,373 @@
...
+    change fi ((c.π.app k ≫ (F ⋙ toProfinite).map _) x) =
+      fj ((c.π.app k ≫ (F ⋙ toProfinite).map _) x)
+    have h := LocallyConstant.congr_fun h x
+    rwa [c.w, c.w]
+
+/--
+The functor `locallyConstantPresheaf` takes a profinite set written as the cofiltered limit of its
+discrete quotients (`S.asLimit`) to the corresponding colimit.
+-/
+noncomputable def isColimit (S : Profinite) :
```
jcommelin, line 69: Most of the `def`s in this PR do not come with `@[simps]` or other forms of equation lemmas/API.

Do you think `@[simps]` just doesn't work well enough? I'm a bit worried that these defn's will be be hard to work with if there are no supporting lemmas at all.

jcommelin:
What do you think of removing the `t-category-theory` label? Since `t-condensed` is an "official" top-level label.

Mathlib/Condensed/Discrete/Colimit.lean
dagurtomas: The `-` was intended as a "blank". So changed to `LocallyConstant · F(*)`.

Mathlib/Condensed/Discrete/Colimit.lean
dagurtomas: My idea was that one shouldn't interact that much with the `def`s in this file, but rather use the `Condensed.IsDiscrete` typeclass introduced in #14027 and access the equivalent characterizations via the `tfae` statements introduced in that same PR.

But I'll go through this and see if there are some useful simp lemmas to be extracted.

Mathlib/Condensed/Discrete/Colimit.lean
```diff
@@ -0,0 +1,535 @@
...
+
+/-- The presheaf on `LightProfinite` of locally constant functions to `X`. -/
+abbrev locallyConstantPresheaf : LightProfiniteᵒᵖ ⥤ Type u :=
+  CompHausLike.LocallyConstant.functorToPresheaves.{u, u}.obj X
+
+/--
+The functor `locallyConstantPresheaf` takes sequential limits of finite sets with surjective
+projection maps to colimits.
+-/
+noncomputable def isColimitLocallyConstantPresheaf (hc : IsLimit c) [∀ i, Epi (c.π.app i)] :
```
jcommelin, line 293: Should this be accompanied by simp lemmas?

Mathlib/Condensed/Discrete/Colimit.lean
```diff
@@ -0,0 +1,535 @@
...
+    ext x
+    obtain ⟨x, hx⟩ := ((LightProfinite.epi_iff_surjective (c.π.app k)).mp inferInstance) x
+    rw [← hx]
+    change fi ((c.π.app k ≫ (F ⋙ toLightProfinite).map _) x) =
+      fj ((c.π.app k ≫ (F ⋙ toLightProfinite).map _) x)
+    have h := LocallyConstant.congr_fun h x
+    rwa [c.w, c.w]
+
+/-- `isColimitLocallyConstantPresheaf` in the case of `S.asLimit`. -/
+noncomputable def isColimitLocallyConstantPresheafDiagram (S : LightProfinite) :
```
jcommelin, line 315: And this?

Mathlib/Condensed/Discrete/Colimit.lean
jcommelin: Thanks, and fair point.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15567, opened 2024-08-06
[Merged by Bors] - chore: move binary recursion on `Nat` into a new file

astrainfinita:
This is #13649 with fewer definition changes.

---

- [x] depends on: #17897
- [x] depends on: #17924
- [x] depends on: #17925
- [x] depends on: #17926


Mathlib/Data/Nat/BinaryRec.lean
```diff
@@ -0,0 +1,132 @@
...
+  For a predicate `C : Nat → Sort u`, if instances can be
+  constructed for natural numbers of the form `bit b n`,
+  they can be constructed for all natural numbers. -/
+@[specialize]
+def binaryRec {C : Nat → Sort u} (z : C 0) (f : ∀ b n, C n → C (bit b n)) (n : Nat) : C n :=
+  if n0 : n = 0 then n0 ▸ z
+  else
+    let x := f (1 &&& n != 0) (n >>> 1) (binaryRec z f (n >>> 1))
+    n.bit_testBit_zero_shiftRight_one ▸ x
+decreasing_by exact bitwise_rec_lemma n0
```
eric-wieser, line 54: This definition isn't the same as the original, right?

Mathlib/Data/Nat/Bits.lean
```diff
@@ -288,50 +230,26 @@ theorem bit_ne_zero (b) {n} (h : n ≠ 0) : bit b n ≠ 0 := by
...
-  intros; rfl
-
-@[simp]
-theorem bitCasesOn_bit0 {motive : ℕ → Sort u} (H : ∀ b n, motive (bit b n)) (n : ℕ) :
+theorem bitCasesOn_bit0 {C : ℕ → Sort u} (H : ∀ b n, C (bit b n)) (n : ℕ) :
     bitCasesOn (2 * n) H = H false n :=
   bitCasesOn_bit H false n
 
 @[simp]
-theorem bitCasesOn_bit1 {motive : ℕ → Sort u} (H : ∀ b n, motive (bit b n)) (n : ℕ) :
```
eric-wieser, line 314: This should remain as `motive`

eric-wieser:
I'm not sure whether I trust the state of this PR at this point; since you've been cleaning everything else up in place (which is great), it might be best to do the move from a clean checkout once the dust settles, to ensure that what ends up in the new file is what you ended up submitting in the other PRs.

Mathlib/Data/Nat/BinaryRec.lean
```diff
@@ -0,0 +1,157 @@
+/-
+Copyright (c) 2022 Praneeth Kolichala. All rights reserved.
```
eric-wieser, line 2: I think this should read "2017 Microsoft Corporation", `binary_rec` came from https://github.com/leanprover-community/lean/blob/master/library/init/data/nat/bitwise.lean authored by @digama0.

eric-wieser (approved):
bors merge

Thanks!

PR #15568, opened 2024-08-06
chore: rename `testBit_bit_{zero, succ}` -> `bit_testBit_{zero, succ}`

astrainfinita:
These are consistent with `Nat.zero_testBit` in core.

---


eric-wieser (changes requested):
Looking at https://loogle.lean-lang.org/?q=Nat.testBit, I think `Nat.zero_testBit` is the exception rather than the rule. Notably, things like `Nat.testBit_or` would need to be renamed to `Nat.or_testBit` for this change to be consistent.

PR #15569, opened 2024-08-06
[Merged by Bors] - feat(Condensed): discrete condensed modules are given by locally constant maps

dagurtomas:
This PR provides the necessary API to prove that a condensed `R`-module is discrete if and only
if the underlying condensed set is (both for light condensed and condensed).

That is, it defines the functor `CondensedMod.LocallyConstant.functor` which takes an `R`-module to
the condensed `R`-modules given by locally constant maps to it, and proves that this functor is
naturally isomorphic to the constant sheaf functor (and the analogues for light condensed modules).

The actual result that a condensed module is discrete if and only if the underlying condensed set is follows easily and is added in #14027.

---

- [x] depends on: #15321 
- [x] depends on: #16385  
- [x] depends on: #17075 


jcommelin (commented):
Thanks :tada:

bors merge

PR #15570, opened 2024-08-06
perf: work with `Expr`s rather than `Term`s in `linear_combination`

hrmacbeth:
This is a rewrite of the proof-construction in the `linear_combination` tactic to work around the issue https://github.com/leanprover/lean4/issues/4861.  Previously, a proof was built up as a `Term`, like
```
``(pf_add_c $p₁ $e₂)
```
The change is to instead build it up as an `Expr`, like `q(pf_add_c $p₁ $e₂)`.

I don't fully understand why, but this gives a 10x speedup on (at least some) large examples.  I have added one such example (which arose in real life as a `polyrith` output) to the test file.  There doesn't seem to be any big change on "normal-sized" examples -- it's possibly a little faster, but not much.

Trade-offs in this implementation:
- The tactic no longer solves equality goals in which the type is a metavariable.
- The default interpretation of coercions in coefficients of a `linear_combination` is now "deeper in", so coercions "on the outside" of a coefficient need to be written manually.  Example: for `a : Kˣ`, `x y : K`, and `h : x = y`, the coefficient `a⁻¹` in `linear_combination a⁻¹ * h` is now parsed as `(a:K)⁻¹`, so it is necessary to write `↑a⁻¹` to have the coercion happen outside the inversion. 

---

### DIFF
--- a/Mathlib/Tactic/LinearCombination.lean
+++ b/Mathlib/Tactic/LinearCombination.lean
@@ -130,6 +130,14 @@ theorem eq_of_add_pow [Ring α] [NoZeroDivisors α] (n : ℕ) (p : (a:α) = b)
     (H : (a' - b')^n - (a - b) = 0) : a' = b' := by
   rw [← sub_eq_zero] at p ⊢; apply pow_eq_zero (n := n); rwa [sub_eq_zero, p] at H
 
+-- /-- Get the mvarid of the main goal, run the given `tactic`,
+-- then set the new goals to be the resulting goal list.-/
+-- @[inline] def liftMetaTactic (tactic : MVarId → MetaM (List MVarId)) : TacticM Unit :=
+--   withMainContext do
+--     let mvarIds ← tactic (← getMainGoal)
+--     replaceMainGoal mvarIds
+--     pure ()
+
 /-- Implementation of `linear_combination` and `linear_combination2`. -/
 def elabLinearCombination
     (norm? : Option Syntax.Tactic) (exp? : Option Syntax.NumLit) (input : Option Syntax.Term)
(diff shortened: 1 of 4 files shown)


hrmacbeth:
!bench

test/linear_combination.lean
```diff
@@ -230,3 +235,39 @@ example (h : g a = g b) : a ^ 4 = b ^ 4 := by
...
+                6561 * q ^ 6 * p ^ 3 * x * y +
+              19683 / 2 * q ^ 7 * p ^ 2 * z * y) *
+            h₁ +
+        (-(128 / 3 * p ^ 12 * x * z) - 192 * q * p ^ 10 * x ^ 2 - 864 * q ^ 2 * p ^ 9 * x * z -
+                    3888 * q ^ 3 * p ^ 7 * x ^ 2 -
+                  5832 * q ^ 4 * p ^ 6 * x * z -
+                26244 * q ^ 5 * p ^ 4 * x ^ 2 -
+              13122 * q ^ 6 * p ^ 3 * x * z -
+            59049 * q ^ 7 * p * x ^ 2) *
+          h₂
```
eric-wieser, line 273: an aside: this pretty-printing/line-wrapping behavior of lean is pretty nuts, and perhaps worth opening an RFC about

eric-wieser:
> * The tactic no longer solves equality goals in which the type is a metavariable.

Out of scope for this PR, but is a version of `synthInstanceQ` that creates a metavariable and askes lean to synthesize the instance later possible?

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -119,25 +141,46 @@ theorem eq_of_add_pow [Ring α] [NoZeroDivisors α] (n : ℕ) (p : (a:α) = b)
...
+          let mvar ← mkFreshExprMVar q(($a' - $b') ^ $n - ($a - $b) = 0)
+          pure (q(eq_of_add_pow (a' := $a') (b' := $b') $n $p), #[mvar])
+      | none =>
+        let _i ← synthInstanceQ q(AddGroup ($α : Type v))
+        let mvar ← mkFreshExprMVar q($a' - $b' - ($a - $b) = 0)
+        pure (q(eq_of_add (a' := $a') (b' := $b') $p), #[mvar])
+  (← Tactic.getMainGoal).assign (mkAppN e mvars)
+  for mvar in mvars do
+    Tactic.setGoals [mvar.mvarId!]
+    Tactic.evalTactic norm
```
eric-wieser, line 183: Should this assert that no goals remain? I think we might have a combinator for "run `tac` on `mvar` and make it close the goal"

hrmacbeth: No, for example quite often one wants to run `linear_combination (norm := skip)`, precisely to see the remaining goals.

eric-wieser: Perhaps I'm missing something; how do these unsolved goals end up back in the tactic list? Doesn't `setGoals` leave only the last one around, forcing the user to `recover`?

hrmacbeth: Ah, you're right: the tests don't contain an example in which simultaneously
- `mvars` contains more than one goal;
- `norm` does not close the first goal.

Here's an example like this:
```lean
example {a b c : ℤ} (h1 : a = 2) (h2 : b = c) : a + b = 2 + c := by
  linear_combination2 (norm := skip) h1 + h2
```
This indeed now produces an error "(kernel) declaration has metavariables '_example'".

Maybe you can tell me how to pick up the lost goal and add it back to the list -- I don't know how.

(By the way, the `linear_combination2` variant of the tactic is the only setting in which more than one goal would be generated, and as it happens, I've been planning to make a case for the deletion of that variant ....)

hrmacbeth: OK, 5483c674ecf8e95b622117ea1765df150217d0ff is to address this, not sure if it's the idiomatic way but it seems to work.

test/linear_combination.lean
```diff
@@ -230,3 +237,40 @@ example (h : g a = g b) : a ^ 4 = b ^ 4 := by
...
```
kmill, line 275: I just measured my caching experiment on this example, and it doesn't do too much: 3.91s on master, vs 2.58s with lean4#4938. I didn't measure this PR's changes.

The problem is the caching only helps large expressions, but `elabTerm` is being called many many times here. Still, something along these lines might improve performance...

kmill:
I haven't built all of mathlib yet, but over at #15599 I adapted the approach here to the original macro-based approach.

The macros aren't the slow part exactly, but instead it's mostly a matter of (1) double elaboration and (2) slowness in elaborating exponentiation. An unfortunate complexity with exponentiation is that in programming contexts you want `x ^ 2.1` with `x : Float` to end up with `2.1 : Float`, but if you use the default instances for monoids it would try to make it be `2.1 : Nat` and then end with an elaboration failure. What I did was force the exponentiation to elaborate in the math-friendly way by fully elaborating exponents ahead of time.

test/linear_combination.lean
eric-wieser: Is there an argument that the elaboration should happen all at once? That is, something like:
* build a `Syntax` with the equalities replaced with holes of the right type
* elaborate this expression
* parse the result with `Qq`, assemble a proof without invoking the elaborator

I think the current version is slow because it assembles the proof as `Syntax` too.

kmill: The current version on master is slow because of exponentiation elaboration relying on default instances (I mention this in the other comment). Fixing this makes the performance comparable to this Qq version.

The caching I was trying to do was to try to re-use these default instance problems as much as possible, but the caching has too small of a scope to be useful.

kmill:
It turns out re-using elaboration wasn't actually very important, and plus it made compiling mathlib fail. The most significant effect was in helping exponentiation elaborate quickly, bringing the time from 3.91s to 0.3s in the big example.

eric-wieser:
What makes elaboration of `^` in `linear_combination` different to elaboration of `^` elsewhere? Or is it slow everywhere?

kmill:
It's slow everywhere unfortunately.

hrmacbeth:
Closing in favour of #15599.

PR #15571, opened 2024-08-06
[Merged by Bors] - chore: redefine `Nat.bitCasesOn` `Nat.binaryRec`

astrainfinita:
The new definitions of `Nat.bitCasesOn` and `Nat.binaryRec*` have better runtime performance. Also they have lighter weight dependencies now and may be upstreamed to `Batteries` later.

---


Mathlib/Data/Nat/Bits.lean
```diff
@@ -121,10 +121,20 @@ lemma bit_val (b n) : bit b n = 2 * n + cond b 1 0 := by
...
+  cases mod_two_eq_zero_or_one n with | _ h => simpa [h] using Nat.div_add_mod n 2
+
 /-- For a predicate `C : Nat → Sort*`, if instances can be
   constructed for natural numbers of the form `bit b n`,
   they can be constructed for any given natural number. -/
-def bitCasesOn {C : Nat → Sort u} (n) (h : ∀ b n, C (bit b n)) : C n := bit_decomp n ▸ h _ _
+def bitCasesOn {C : Nat → Sort u} (n) (h : ∀ b n, C (bit b n)) : C n :=
+  -- `1 &&& n != 0` is faster than `n.testBit 0`. This may change when we have faster `testBit`.
+  let x := h (1 &&& n != 0) (n >>> 1)
+  -- `congrArg C _` is `rfl` in non-dependent case
```
eric-wieser, line 136: What are you trying to convey here? Is the point that when you unfold `bitCasesOn` in the dependent case, this is easier to eliminate?

Mathlib/Data/Nat/Bits.lean
```diff
@@ -121,10 +121,20 @@ lemma bit_val (b n) : bit b n = 2 * n + cond b 1 0 := by
...
+
+theorem bit_testBit_zero_shiftRight_one (n : Nat) : bit (n.testBit 0) (n >>> 1) = n := by
+  simp only [bit, shiftRight_one, testBit_zero]
+  cases mod_two_eq_zero_or_one n with | _ h => simpa [h] using Nat.div_add_mod n 2
+
 /-- For a predicate `C : Nat → Sort*`, if instances can be
   constructed for natural numbers of the form `bit b n`,
   they can be constructed for any given natural number. -/
-def bitCasesOn {C : Nat → Sort u} (n) (h : ∀ b n, C (bit b n)) : C n := bit_decomp n ▸ h _ _
+def bitCasesOn {C : Nat → Sort u} (n) (h : ∀ b n, C (bit b n)) : C n :=
```
eric-wieser, line 133: While you're here, and before you move things; could you rename `C` to `motive` in these recursors too?

eric-wieser (approved):
This generally looks good to me.

Can you add a sentence or two in the PR description explaining the motivation (which is some combination of performance, and wanting to drop bodd / div2, right?)

Mathlib/Data/Nat/Bits.lean
```diff
@@ -122,10 +122,20 @@ lemma bit_val (b n) : bit b n = 2 * n + cond b 1 0 := by
 lemma bit_decomp (n : Nat) : bit (bodd n) (div2 n) = n :=
   (bit_val _ _).trans <| (Nat.add_comm _ _).trans <| bodd_add_div2 _
 
-/-- For a predicate `C : Nat → Sort*`, if instances can be
+theorem shiftRight_one (n) : n >>> 1 = n / 2 := rfl
+
+theorem bit_testBit_zero_shiftRight_one (n : Nat) : bit (n.testBit 0) (n >>> 1) = n := by
```
eric-wieser, line 127:
```suggestion
@[simp]
theorem bit_testBit_zero_shiftRight_one (n : Nat) : bit (n.testBit 0) (n >>> 1) = n := by
```

eric-wieser (approved):
bors d+

Thanks!

Mathlib/Data/Nat/Bits.lean
eric-wieser: Ah, I now see that `testBit` is not simp-normal; perhaps better to just ignore this suggestion.

PR #15582, opened 2024-08-07
[Merged by Bors] - feat (LinearAlgebra/RootSystem) : infinitely many roots when two are parallel

ScottCarnahan:
This PR has a proof that if two roots are linearly independent and "parallel" (meaning their pairings multiply to 4), then the root system containing them is infinite.  This is mostly done by refactoring an existing proof.

---


Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -174,33 +174,47 @@ lemma Dual.eq_of_preReflection_mapsTo' [CharZero R] [NoZeroSMulDivisors R M]
 variable {y}
 variable {g : Dual R M}
 
+/-- Composite of reflections in "parallel" hyperplanes is a shear (special case). -/
+lemma reflection_reflection_iterate (hfx : f x = 2) (hgy : g y = 2) (hgxfy : f y * g x = 4)
+    (n : ℕ) : ((Module.reflection hgy).trans (Module.reflection hfx))^[n] y =
+      y + (n : ℤ) • (f y • x - (2 : ℤ) • y) := by
```
ocfnash, line 180: I think it will be slightly more convenient to use `(2 : R)`. I also prefer not to break place new lines in a term (the equalty in this case) unless necessary. Thus I suggest:
```suggestion
lemma reflection_reflection_iterate
    (hfx : f x = 2) (hgy : g y = 2) (hgxfy : f y * g x = 4) (n : ℕ) :
    ((reflection hgy).trans (reflection hfx))^[n] y = y + n • (f y • x - (2 : R) • y) := by
```

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -174,33 +174,47 @@ lemma Dual.eq_of_preReflection_mapsTo' [CharZero R] [NoZeroSMulDivisors R M]
...
+  induction n with
+  | zero => simp
+  | succ n ih =>
+    have hz : ∀ z : M, f y • g x • z = 2 • 2 • z := by
+      intro z
+      rw [smul_smul, hgxfy, smul_smul, ← Nat.cast_smul_eq_nsmul R (2 * 2), Nat.cast_eq_ofNat]
+    simp only [iterate_succ', comp_apply, ih, map_add, LinearEquiv.trans_apply, sub_smul, map_smul,
+      reflection_apply_self, map_neg, reflection_apply, hz, two_smul, neg_sub, map_sub, smul_sub,
+      map_nsmul, map_zsmul, smul_neg, smul_add, Nat.cast_succ, mul_add, mul_one, add_smul, two_mul]
```
ocfnash, line 189: I think this fixes the proof for the tweaked statement:
```suggestion
    simp only [iterate_succ', comp_apply, ih, two_smul, smul_sub, smul_add, map_add,
      LinearEquiv.trans_apply, reflection_apply_self, map_neg, reflection_apply, neg_sub, map_sub,
      map_nsmul, map_smul, smul_neg, hz, add_smul]
```

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -174,33 +174,47 @@ lemma Dual.eq_of_preReflection_mapsTo' [CharZero R] [NoZeroSMulDivisors R M]
...
+    abel
+
+lemma infinite_range_of_reflection_reflection_iterate [NoZeroSMulDivisors ℤ M] (hfx : f x = 2)
+    (hgy : g y = 2) (hgxfy : f y * g x = 4) (hxy : ¬ f y • x = (2 : ℤ) • y) : Set.Infinite (range
+      (fun n => ((Module.reflection hgy).trans (Module.reflection hfx))^[n] y)) := by
+  refine infinite_range_of_injective fun n m hnm => ?_
+  simp only [reflection_reflection_iterate hfx hgy hgxfy] at hnm
+  rw [add_right_inj, ← sub_eq_zero, ← sub_smul, smul_eq_zero, sub_eq_zero, sub_eq_zero] at hnm
+  simpa using hnm.resolve_right hxy
```
ocfnash, line 198: I think there is a tiny result worth factoring out here. To be precise I suggest putting these lemmas at the bottom of `Mathlib.Algebra.Module.Torsion`:
```lean
@[simp]
lemma infinite_range_add_smul_iff
    [AddCommGroup M] [Ring R] [Module R M] [Infinite R] [NoZeroSMulDivisors R M] (x y : M) :
    (Set.range <| fun r : R ↦ x + r • y).Infinite ↔ y ≠ 0 := by
  refine ⟨fun h hy ↦ by simp [hy] at h, fun h ↦ Set.infinite_range_of_injective fun r s hrs ↦ ?_⟩
  rw [add_right_inj] at hrs
  exact smul_left_injective _ h hrs

@[simp]
lemma infinite_range_add_nsmul_iff
    [AddCommGroup M] [NoZeroSMulDivisors ℤ M] (x y : M) :
    (Set.range <| fun n : ℕ ↦ x + n • y).Infinite ↔ y ≠ 0 := by
  refine ⟨fun h hy ↦ by simp [hy] at h, fun h ↦ Set.infinite_range_of_injective fun r s hrs ↦ ?_⟩
  rw [add_right_inj, ← natCast_zsmul, ← natCast_zsmul] at hrs
  simpa using smul_left_injective _ h hrs
```
so that we can shorten the proof here (and also upgrade it to an iff) as follows:
```suggestion
lemma infinite_range_reflection_reflection_iterate_iff [NoZeroSMulDivisors ℤ M]
    (hfx : f x = 2) (hgy : g y = 2) (hgxfy : f y * g x = 4) :
    (range <| fun n ↦ ((reflection hgy).trans (reflection hfx))^[n] y).Infinite ↔
    f y • x ≠ (2 : R) • y := by
  simp only [reflection_reflection_iterate hfx hgy hgxfy, infinite_range_add_nsmul_iff, sub_ne_zero]
```
I was 50-50 on whether we should state these results or their contrapositive but given your ultimate application I was persuaded this was the most convenient.

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -174,33 +174,47 @@ lemma Dual.eq_of_preReflection_mapsTo' [CharZero R] [NoZeroSMulDivisors R M]
...
-    rw [← IsFixedPt.image_iterate hb.image_eq n]; exact mem_image_of_mem _ hyΦ⟩
-  have : ¬ Injective f' := not_injective_infinite_finite f'
-  contrapose! this
-  intros n m hnm
-  rw [Subtype.mk_eq_mk, hsxy, hsxy, add_right_inj, ← sub_eq_zero, ← sub_smul, smul_eq_zero,
-    sub_eq_zero, sub_eq_zero] at hnm
-  simpa using hnm.resolve_right this
+  suffices h : f y • x = (2 : ℤ) • y by
+    rw [hfy, two_smul R x, ← two_zsmul, ← sub_eq_zero, ← smul_sub, smul_eq_zero, sub_eq_zero] at h
+    simpa using h.resolve_left
```
ocfnash, line 208: Again I suggest sticking to `(2 : R)` here:
```suggestion
  suffices h : f y • x = (2 : R) • y by
    rw [hfy, two_smul R x, two_smul R y, ← two_zsmul, ← two_zsmul] at h
    exact smul_right_injective _ two_ne_zero h
```

Mathlib/LinearAlgebra/Reflection.lean
```diff
@@ -174,33 +174,47 @@ lemma Dual.eq_of_preReflection_mapsTo' [CharZero R] [NoZeroSMulDivisors R M]
...
+  rw [← not_infinite] at hΦ
+  contrapose! hΦ
+  refine Infinite.mono ?_
+    (infinite_range_of_reflection_reflection_iterate hfx hgy (by rw [hfy, hgx]; norm_cast) hΦ)
```
ocfnash, line 212: And then given the changes above this becomes:
```suggestion
apply ((infinite_range_reflection_reflection_iterate_iff hfx hgy
    (by rw [hfy, hgx]; norm_cast)).mpr hΦ).mono
```

Mathlib/LinearAlgebra/RootSystem/Basic.lean
```diff
@@ -74,6 +71,27 @@ protected def equiv_of_mapsTo :
 
 end reflection_perm
 
+lemma infinite_of_linearly_independent_coxeterWeight_four [CharZero R] [NoZeroSMulDivisors ℤ M]
+    (P : RootPairing ι R M N) (i j : ι) (hl : LinearIndependent R ![P.root i, P.root j])
+    (hc : P.coxeterWeight i j = 4) : Infinite ι := by
+  refine (infinite_range_iff (Embedding.injective P.root)).mp (Infinite.mono ?_
+    (infinite_range_of_reflection_reflection_iterate (P.coroot_root_two i)
```
ocfnash, line 78: Some small change will be necessary here too which I will leave to you (assuming you agree to the suggestions above).

ocfnash:
Thanks!

bors merge

PR #15583, opened 2024-08-07
[Merged by Bors] - fix: add `all` flag to import summary script

adomani:
Summary of the PR

* Add the `all` flag to the import diff script: running
  ```bash
  ./scripts/import_trans_difference.sh all
  ```
  will not limit the output to ~~200 modified files~~ ~130k characters, but will print the whole import diff.

* Increase the size cut-off for printing the full output of the `import-diff script` to roughly half the size allowed in GitHub comments.

* Add a comment that you can run the script locally.
  The script switches over between commits and creates new files to store the information that it gathers from the various steps and can be intrusive: the automation is designed more to be ran in CI, with a fresh clone, than locally with possibly uncommitted changes, so I am not too sure that this is a good idea.

#15584 shows the new script in action: the printed comment there exceeds the previous limits of the script.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/There.20are.201617.20files.20with.20changed.20transitive.20imports.20.2E.2E.2E)

---

### DIFF
--- a/scripts/import_trans_difference.sh
+++ b/scripts/import_trans_difference.sh
@@ -72,18 +72,18 @@ git checkout "${currCommit}"
 
 printf '\n\n<details><summary>Import changes for all files</summary>\n\n%s\n\n</details>\n' "$(
   printf "|Files|Import difference|\n|-|-|\n"
-  (awk -F, -v all="${all}" '{ diff[$1]+=$2 } END {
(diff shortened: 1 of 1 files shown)


joneugster:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15587, opened 2024-08-07
[Merged by Bors] - doc(LinearAlgebra/FreeProduct): remove untrue TODOs

robertmaxton42:
Neither TODO turned out to be mathematically true: `FreeProduct` imposes more structure than `FreeAlgebra`, specifically combining adjacent elements from the same algebra, and my categorical argument for the isomorphism with `PiTensorProduct` in the commutative case was faulty.

---

### DIFF
--- a/Mathlib/LinearAlgebra/FreeProduct/Basic.lean
+++ b/Mathlib/LinearAlgebra/FreeProduct/Basic.lean
@@ -40,8 +40,6 @@ is just `PiTensorProduct`.)
 
 ## TODO
 - Induction principle for `FreeProduct`
-- Equivalence of `FreeProduct` with `PiTensorProduct` in the commutative case and with
-  `FreeAlgebra R ⨁ (i : ι), A i`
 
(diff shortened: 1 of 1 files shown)


Ruben-VandeVelde (approved):
Thanks!

maintainer merge

mattrobball:
bors merge

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -40,8 +40,6 @@ is just `PiTensorProduct`.)
 
 ## TODO
 - Induction principle for `FreeProduct`
-- Equivalence of `FreeProduct` with `PiTensorProduct` in the commutative case and with
-  `FreeAlgebra R ⨁ (i : ι), A i`
```
eric-wieser, line 44: Would `MonoidAlgebra R (Monoid.CoprodI A)` be correct here instead?

PR #15588, opened 2024-08-07
[Merged by Bors] - chore: remove some useless `simp`s

kmill:
They are just instantiating some metavariables that are incidentally there.

---


eric-wieser (approved):
bors merge

Thanks!

PR #15592, opened 2024-08-07
perf(CategoryTheory.Sites.Sheaf): use `suppress_compilation` 

mattrobball:
After moving to `v4.11.0-rc1`, we acquired 40 additional seconds of attempts to compile a declaration in a file wrapped in a `noncomputable section`. We use `suppress_compilation` to avoid this.

---


mattrobball:
!bench

PR #15595, opened 2024-08-07
[Merged by Bors] - golf: found by the have vs let linter

adomani:
The `have` vs `let` linter had flagged these `have` as "should be `let`s" and changing them made the proof simpler.

---

I could not get `!bench` to work, but these are the timings locally:
```lean
count_heartbeats in version with `let`
Used 49399 heartbeats, which is less than the current maximum of 200000.

count_heartbeats in version with `have`
Used 55316 heartbeats, which is less than the current maximum of 200000.
```


adomani:
!bench

Ruben-VandeVelde:
Not sure what happened to the bench, but feel free to maintainer merge on my behalf if it doesn't show anything nasty

adomani:
!bench

adomani:
Ruben, I'm trying bench again, but the local timings are in favour of the change (very slightly, but still, in the correct direction!).

adomani:
Since bench says that there are no significant changes and the local timings are slightly in favour of the change, I'm going to

maintainer merge

on Ruben's behalf!

j-loreaux:
bors merge

PR #15597, opened 2024-08-07
[Merged by Bors] - chore: add ascription for delicate-to-elaborate expression

kmill:
The expression `2 • (Finsupp.single 0 1 : α →₀ F) = Finsupp.single 0 1` only accidentally elaborates. The binop elaborator fails to compute a "max type" for this, since the right-hand `Finsupp.single 0 1` does not have a fully specified type yet, so it falls back to a strategy where it propagates the LHS type to the RHS. However, the rest of the process then depends on the exact order that default instances are applied. If the default instances for the RHS apply first, it specializes to `ℕ →₀ ℕ` and elaboration fails, but if the LHS ones apply first, then the SMul manages to specialize to `ℕ → (α →₀ F) → (α →₀ F)`, and it succeeds.

This came up while experimenting with performance improvements to the binop elaborator.

---


urkud:
Thanks! :tada:
bors merge

PR #15599, opened 2024-08-07
[Merged by Bors] - perf: eagerly elaborate leaf nodes in `linear_combination`

kmill:
Since the arithmetic type is known from the equality goal, we can eagerly elaborate all leaf nodes in `expandLinearCombo` while also avoiding re-elaborating them when deciding whether they are proofs or not. This brings the elaboration of a complicated example (see tests) from 3.91s to 0.29s.

The main issue is that large expressions can be slow to elaborate. In particular, exponentiation relies on the default instance mechanism, which appears can take quadratic time to resolve. Forcing the resolution of default instances within each coefficient of the linear combination bounds the default instance problems.

This is similar to PR #15570, but it sticks with the syntax transformation paradigm.

Co-authored-by: Heather Macbeth <25316162+hrmacbeth@users.noreply.github.com>

---

### DIFF
--- a/Mathlib/Tactic/LinearCombination.lean
+++ b/Mathlib/Tactic/LinearCombination.lean
@@ -105,18 +105,21 @@ partial def expandLinearCombo (stx : Syntax.Term) : TermElabM Expanded := withRe
     | .const c₁, .proof p₂ => .proof <$> ``(c_div_pf $p₂ $c₁)
     | .proof p₁, .proof p₂ => .proof <$> ``(div_pf $p₁ $p₂)
   | `($e ^ $n) => do
-    -- Fully pre-elaborate n:
+    -- Fully pre-elaborate n to speed up exponentiation elaboration:
     let n ← (withSynthesize <| elabTerm n none) >>= Expr.toSyntax
     match ← expandLinearCombo e with
     | .const c => .const <$> ``($c ^ $n)
     | .proof p => .proof <$> ``(pow_pf $p $n)
   | e => do
-    let c ← Term.elabTerm e none
+    let s ← saveState
+    let c ← withSynthesizeLight <| Term.elabTerm e none
     let eType ← inferType c
     if (← withReducible do whnf eType).isEq then
       .proof <$> c.toSyntax
     else
-      .const <$> c.toSyntax
+      -- Restore state to remove all pending instance problems.
+      restoreState s
+      return .const e
 
 theorem eq_trans₃ (p : (a:α) = b) (p₁ : a = a') (p₂ : b = b') : a' = b' := p₁ ▸ p₂ ▸ p
 
--- a/Mathlib/Tactic/LinearCombination.lean
+++ b/Mathlib/Tactic/LinearCombination.lean
@@ -47,9 +47,14 @@ theorem pf_mul_c [Mul α] (p : a = b) (c : α) : a * c = b * c := p ▸ rfl
 theorem c_mul_pf [Mul α] (p : b = c) (a : α) : a * b = a * c := p ▸ rfl
 theorem mul_pf [Mul α] (p₁ : (a₁:α) = b₁) (p₂ : a₂ = b₂) : a₁ * a₂ = b₁ * b₂ := p₁ ▸ p₂ ▸ rfl
 theorem inv_pf [Inv α] (p : (a:α) = b) : a⁻¹ = b⁻¹ := p ▸ rfl
+theorem pow_pf {γ : Type*} [Pow α γ] (p : (a:α) = b) (n : γ) : a ^ n = b ^ n := p ▸ rfl
 theorem pf_div_c [Div α] (p : a = b) (c : α) : a / c = b / c := p ▸ rfl
 theorem c_div_pf [Div α] (p : b = c) (a : α) : a / b = a / c := p ▸ rfl
 theorem div_pf [Div α] (p₁ : (a₁:α) = b₁) (p₂ : a₂ = b₂) : a₁ / a₂ = b₁ / b₂ := p₁ ▸ p₂ ▸ rfl
+
+inductive Expanded
+  | proof (pf : Syntax.Term)
+  | const (c : Syntax.Term)
 
 /--
 Performs macro expansion of a linear combination expression,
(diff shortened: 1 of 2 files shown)


hrmacbeth:
This speedup is confusing to me, because, if I understand correctly, it's not just about the `^` elaboration!  Case in point: in this branch
https://github.com/leanprover-community/mathlib4/tree/HM-linear-combination-variant
I tried to make the diff of this PR more minimal, by throwing away the `Expanded` data structure and using `Option` as before.  In that variant, even using the tweaks to `^` elaboration, the big example still takes 58,000 heartbeats.  So how is the `Expanded` structure helping?

kmill:
@hrmacbeth I think the deal is that in your branch in the exponentiation case, it's only committing to the fully-elaborated exponent if the base of the exponent is a proof. Mine commits even if the base isn't a proof.

To spell out my understanding of the problem a bit more, when you have `x ^ 2` for example, it might need to go through multiple rounds of default instances until it commits to `2 : Nat`, and from there it's immediate that there's a `Pow _ Nat` instance since fields are monoids. The thing to keep in mind here is that the term elaborator state contains a big list of pending instance problems, and the elaborator is not very smart about how it makes progress: it iteratively goes through this entire list finding any that can make progress. Default instances only apply once it's gone through the entire list (and failed), and then once it finds any default instance, it returns to the beginning of the list and starts iterating again. This can have quadratic behavior (or worse?).

With my version of the code, we fully elaborate `2`, which defaults to `Nat` fairly quickly. This prevents it from ever being added to the main list of instance problems, and we can see it turns out to have a significant effect.

hrmacbeth:
Thanks, that's instructive.  So here's another version to consider: don't special-case exponentiation, just pre-elaborate *every* leaf term (noting that we know the expected type, because we can take it from the goal).
https://github.com/leanprover-community/mathlib4/tree/HM-linear-combination-variant-2

This is a little slower, 8,374 rather than 7,763 heartbeats, but not by much.  And perhaps it might also avoid similar problems in situations other than exponentiation?

kmill:
@hrmacbeth Good idea. I've updated the PR, and managed to re-use the elaboration in leaf nodes to bring down the heartbeats further. Let's see if mathlib builds!

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -60,51 +65,61 @@ using `+`/`-`/`*`/`/` on equations and values.
...
+    -- Assumption: mathlib only uses Pow instances that respect the default instances
+    -- for the exponent. For example, if the exponent is `37` it should be a `Nat`.
+    let n ← (withSynthesize <| elabTerm n none) >>= Expr.toSyntax
+    match ← expandLinearCombo ty e with
+    | .const c => .const <$> ``($c ^ $n)
+    | .proof p => .proof <$> ``(pow_pf $p $n)
+  | e =>
+    -- We have the expected type from the goal, so we can fully synthesize this leaf node.
+    withSynthesize do
+      let c ← withSynthesizeLight <| Term.elabTerm e ty
```
hrmacbeth, line 118: In the "proof" case, the expected type being provided here as `ty` will be wrong -- that is, the type will actually be `(a : ty) = (b : ty)`, not `ty`.  Since it's still very fast, I assume this doesn't matter?

hrmacbeth:
By the way, in the long term I'm not in favour of the "exponentiation" capability this PR adds (like `linear_combination h ^ 3`) -- eventually I'd also advocate for restricting the capabilities of `linear_combination` to only genuine *linear* operations, which would also involve disabling other capabilities Mario added in the port, like `linear_combination h1 * h2` and `linear_combination `5 / h`.

Short version of my rationale: the `congr` term elaborator already provides support for arbitrary operations, but the particular way hypotheses are combined in `linear_combination` (apply the operation, move everything to LHS, convert, then normalize) becomes much less canonical/predictable for non-linear operations.

I'm happy for this PR to be merged as-is ... I can open another PR later to allow that discussion to happen separately.

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -51,60 +51,70 @@ theorem pf_div_c [Div α] (p : a = b) (c : α) : a / c = b / c := p ▸ rfl
...
+  | proof (pf : Syntax.Term)
+  /-- A value, equivalently a proof of `c = c`. -/
+  | const (c : Syntax.Term)
+
 /--
 Performs macro expansion of a linear combination expression,
 using `+`/`-`/`*`/`/` on equations and values.
-* `some p` means that `p` is a syntax corresponding to a proof of an equation.
+* `.proof p` means that `p` is a syntax corresponding to a proof of an equation.
   For example, if `h : a = b` then `expandLinearCombo (2 * h)` returns `some (c_add_pf 2 h)`
```
hrmacbeth, line 65:
```suggestion
  For example, if `h : a = b` then `expandLinearCombo (2 * h)` returns `.proof (c_add_pf 2 h)`
```

Mathlib/Tactic/LinearCombination.lean
kmill: One weird (but useful) thing about the `elabTerm` protocol is that the expected type is merely a suggestion. You have to use `elabTermEnsuringType` if you want to be sure the result has the given type.

This might in fact be a bad idea, but it seems to work. It definitely works if each proof is an fvar, but it gets less and less likely to work as the proof term gets more and more complicated (though perhaps those proofs had little hope of elaborating anyway when the expected type is none? and maybe we want do discourage complicated proofs in a `linear_combination` expression?)

kmill: It occurred to me that it would be good to ensure the type like in `elabTermEnsuringType` in the `.const` case, just to give the proofs an easier time elaborating. This inserts coercions if necessary.

kmill:
@hrmacbeth This is *so* close to just generating expressions instead of going through macros, so I implemented a version that calls elaboration functions directly to do this, and for the big example it just saves 300 heartbeats, without measurably taking any less time. It adds a lot of complexity for no reason, so I'm not pushing it.

I think this proves that the fundamental slowdown is really just from instance synthesis and bounding the elaboration problems rather than using a macro-based approach. I benchmarked your Qq version too, and it takes 0.29s on my computer, with 7500 heartbeats. What do you think? Should we go with this version? Or go with the Qq one, and hopefully fix how it elaborates expressions so that you don't need to change any of mathlib?

Mathlib/Tactic/LinearCombination.lean
kmill: Oops, ensuring the type changed how elaboration works in the Weierstrass file. I guess it wasn't a good idea.

hrmacbeth:
> What do you think? Should we go with this version? Or go with the Qq one, and hopefully fix how it elaborates expressions so that you don't need to change any of mathlib?

@kmill This has been a very interesting exploration for me to follow, but I don't feel equipped to judge between the two implementations we have reached: the crude metric of speed is the only one I know for judging meta code, and they seem to be the same on this. So, your call -- do you think one version has an edge over the other in simplicity or in maintainability?

Mathlib/Tactic/LinearCombination.lean
hrmacbeth: > One weird (but useful) thing about the `elabTerm` protocol is that the expected type is merely a suggestion.

Thanks for the explanation!

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -51,60 +51,70 @@ theorem pf_div_c [Div α] (p : a = b) (c : α) : a / c = b / c := p ▸ rfl
 theorem c_div_pf [Div α] (p : b = c) (a : α) : a / b = a / c := p ▸ rfl
 theorem div_pf [Div α] (p₁ : (a₁:α) = b₁) (p₂ : a₂ = b₂) : a₁ / a₂ = b₁ / b₂ := p₁ ▸ p₂ ▸ rfl
 
+/-- Result of `expandLinearCombo`, either a  -/
```
hrmacbeth, line 54: Maybe you meant to write more here?

kmill, line 54:
```suggestion
/-- Result of `expandLinearCombo`, either an equality proof or a value. -/
```

kmill:
@hrmacbeth I think that this non-Qq version might be easier to maintain, and it has the bonus of being a drop-in replacement. I still don't completely feel like I have a complete handle on the issue... but I think we've pretty well localized it to elaborating large arithmetic expressions being slow, and the `withSynthesize` in `expandLinearCombo` really is the performance fix here.

hrmacbeth:
Sounds good to me!  You made me a co-author on this, so I won't merge directly :)

maintainer merge

semorrison:
bors merge

PR #15602, opened 2024-08-08
feat(Data/Nat/Factorial): add `descFactorial_le` and `descFactorial_mul_descFactorial`

nmouha:
Add two basic theorems involving falling factorials. Note that `descFactorial_mul_descFactorial` collapses to `Nat.factorial_mul_descFactorial` when `m = n`.

---


nmouha:
I'm closing this PR and https://github.com/leanprover-community/mathlib4/pull/15665 so that I can reopen then the right way.

PR #15603, opened 2024-08-08
[Merged by Bors] - feat: better proof of spectral permanence

j-loreaux:
This uses a much cleaner argument than the previous one in order to prove the *spectral permanence* property of C*-algebras. The argument now goes through some generic results for Banach algebras, a corollary of which applies to selfadjoint elements.

The old results which led to spectral permanence have been removed, although the spectral permanence results themselves remain unchanged (except that they have been moved into `Analysis.CStarAlgebra.Spectrum` from `Analysis.CStarAlgebra.ContinuousFunctionalCalculus.Basic`, which seems a better home).

Some module documentation has also been updated.

---

- [x] depends on: #14925 


Mathlib/Analysis/Complex/UpperHalfPlane/Topology.lean
```diff
@@ -131,3 +131,17 @@ lemma comp_ofComplex (f : ℍ → ℂ) (z : ℍ) : (↑ₕ f) z = f z := by
...
 end UpperHalfPlane
+
+lemma Complex.isConnected_of_upperHalfPlane {s : Set ℂ} (hs₁ : {z | 0 < z.im} ⊆ s)
+    (hs₂ : s ⊆ {z | 0 ≤ z.im}) : IsConnected s := by
+  refine IsConnected.subset_closure ?_ hs₁ (by simpa using hs₂)
+  rw [isConnected_iff_connectedSpace]
+  exact inferInstanceAs (ConnectedSpace UpperHalfPlane)
+
+lemma Complex.isConnected_of_lowerHalfPlane {s : Set ℂ} (hs₁ : {z | z.im < 0} ⊆ s)
+    (hs₂ : s ⊆ {z | z.im ≤ 0}) : IsConnected s := by
```
dupuisf, line 142: Can't you just basically copy-paste the proof from  `isConnected_of_upperHalfPlane` since it's shorter anyway?

dupuisf:
Looks good!

bors d+

Mathlib/Analysis/Complex/UpperHalfPlane/Topology.lean
j-loreaux: No, because we would need a `LowerHalfPlane` type on which we have a `ConnectedSpace` instance.

PR #15604, opened 2024-08-08
[Merged by Bors] - chore: rm relaxedAutoImplicit from lakefile

digama0:
The `relaxedAutoImplicit` option has no effect if `autoImplicit` is false, so this has no impact on most of mathlib. However, when enabling `autoImplicit` locally, it is annoying (and slightly non-discoverable) that one must also enable `relaxedAutoImplicit` in many cases to actually get auto-implicits.

---


semorrison:
bors merge

PR #15605, opened 2024-08-08
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 33)

kim-em:
---

### DIFF
--- a/Mathlib/Combinatorics/SimpleGraph/LapMatrix.lean
+++ b/Mathlib/Combinatorics/SimpleGraph/LapMatrix.lean
@@ -28,7 +28,14 @@ open Finset Matrix
 namespace SimpleGraph
 
 variable {V : Type*} (R : Type*)
-variable [Fintype V] [DecidableEq V] (G : SimpleGraph V) [DecidableRel G.Adj]
+variable [Fintype V] (G : SimpleGraph V) [DecidableRel G.Adj]
+
+theorem degree_eq_sum_if_adj {R} [AddCommMonoidWithOne R] (i : V) :
+    (G.degree i : R) = ∑ j : V, if G.Adj i j then 1 else 0 := by
+  unfold degree neighborFinset neighborSet
+  rw [sum_boole, Set.toFinset_setOf]
+
+variable [DecidableEq V]
 
 /-- The diagonal matrix consisting of the degrees of the vertices in the graph. -/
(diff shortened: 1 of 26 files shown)


Mathlib/Combinatorics/SimpleGraph/LapMatrix.lean
```diff
@@ -28,7 +28,14 @@ open Finset Matrix
 namespace SimpleGraph
 
 variable {V : Type*} (R : Type*)
-variable [Fintype V] [DecidableEq V] (G : SimpleGraph V) [DecidableRel G.Adj]
+variable [Fintype V] (G : SimpleGraph V) [DecidableRel G.Adj]
+
+theorem degree_eq_sum_if_adj {R} [AddCommMonoidWithOne R] (i : V) :
```
sgouezel, line 33:
```suggestion
theorem degree_eq_sum_if_adj {R : Type*} [AddCommMonoidWithOne R] (i : V) :
```

Mathlib/NumberTheory/Cyclotomic/Basic.lean
```diff
@@ -578,10 +579,12 @@ instance algebraBase : Algebra A (CyclotomicRing n A K) :=
 -- but there is at `reducible_and_instances` #10906
 example {n : ℕ+} : CyclotomicRing.algebraBase n ℤ ℚ = Ring.toIntAlgebra _ := rfl
 
-instance : NoZeroSMulDivisors A (CyclotomicRing n A K) :=
+instance [NoZeroSMulDivisors A (CyclotomicField n K)] :
```
sgouezel, line 582: I'm confused by this one: it seems that you're adding a typeclass assumption, but that you haven't deleted it in this file. Same thing below.

sgouezel (commented):
bors d+

Mathlib/NumberTheory/Cyclotomic/Basic.lean
semorrison: Oops, this had previously been supplied indirectly via the `[IsFractionRing A K]` deleted above, but I guess I just added the typeclass that Lean reported was missing, rather than the stronger deleted assumption.

I'll change these back to `[IsFractionRing A K]`, as it seems unlikely that any added generality is actually useful or helpful here!!

semorrison:
bors merge

semorrison:
bors merge

PR #15606, opened 2024-08-08
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 35)

kim-em:
This one is a bit awkward / done badly. I reordered theorems in order to avoid pulling in `SmoothManifoldWithCorners` unnecessarily, but I didn't preserve the section headings properly. :-(


semorrison:
If either @sgouezel or @fpvandoorn would be able to do another pass on this, either starting over or just getting this one in shape, I'd be much appreciative.

semorrison:
(I've cherry picked this onto lean-pr-testing-4814, so whatever we end up needs to go there as well.)

fpvandoorn:
Thanks!

bors merge

PR #15609, opened 2024-08-08
test: weak option for hashCommandLinter

adomani:
---


grunweg:
It appears to me that this PR is obsolete now (as the change tested here has landed on master). Is this correct?

PR #15610, opened 2024-08-08
[Merged by Bors] - feat: the longFile linter

adomani:
The "longFile" linter is a syntax linter. It emits a warning on files which are longer than a certain number of lines
(1500 by default). A value of `0` silences the linter entirely.

This PR also modifies the `TextBased` `fileTooLong` linter: now it only lints files that do not contain `set_option linter.style.longFile `, and removes the exceptions that are now covered by the "longFile" linter.

The `TextBased` longFile linter stays (for now), though it is going to be removed soon.

---

Once #15845 lands, the text-based linter can be removed.

I tried setting the default option for the linter to 0 and then modifying it in the lakefile to be 1500 for Mathlib, but I could not get it to work.


adomani:
What I feel is really missing for this linter is a test.  However, I do not know how to test it, since `eoi` is not a "normal" syntax node and `#guard_msgs in` does not latch onto it.

grunweg:
Two quick thoughts (I can review properly next week):
- can you add a comment to the text-based linter about this, including that many style exceptions are not necessary as these are managed by the syntax based linter, and why both linter will co-exist?
- can you update the technical debt counter as well? Currently, it counts the number of "long file" style exceptions; can you ensure to also count the number of "set_option linter.longFile" (and add these... if a shell script allows this)?
- as I said already: a syntax linter has much nicer user experience; thanks for upgrading the text-based linter even further!

adomani:
Michael, thank you for the review!

I expanded the docs and comment to the text-based linters, mentioning the syntax one.

I also added the `set_option linter.longFile` count to the technical debts metrics.  This is what I see locally:
```bash
$ printf '%s|%s\n' "$(
  { grep 'ERR_NUM_LIN' scripts/style-exceptions.txt
    grep '^set_option linter.longFile [0-9]*' $(git ls-files '*.lean') ; } | wc -l
)" "large files"
66|large files
```
and it matches the number that the Zulip post has.

grunweg (commented):
Thanks for rewriting this linter! I took the liberty to propose some wording changes directly - feel free to revert any you dislike. I have one minor comment - apart from that, this looks good to go.

grunweg:
(Long-term, it would be nicest to have this PR *remove* the text-based linter instead, and import this linter in Mathlib.Init instead. We'll see in which order things land.)

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -218,6 +218,56 @@ initialize addLinter cdotLinter
...
+
+namespace LongFile
+
+@[inherit_doc Mathlib.Linter.linter.longFile]
+def longFileLinter : Linter where run := withSetOptionIn fun stx ↦ do
+  let linterBound := linter.longFile.get (← getOptions)
+  if linterBound == 0 then
+    return
+  unless stx.isOfKind ``Lean.Parser.Command.eoi do return
+  if (← getMainModule) == `Mathlib then return
```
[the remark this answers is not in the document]
adomani: Added a comment and a "future" TODO.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -218,6 +218,56 @@ initialize addLinter cdotLinter
...
+    -- `candidate` is divisible by `100` and satisfies `lastLine + 100 < candidate ≤ lastLine + 200`
+    let candidate := (lastLine / 100) * 100 + 200
+    let endMsg := m!"using `set_option linter.longFile {candidate}`.\n\
+                    You can disable completely this linter by setting the length limit to `0`."
+    if linterBound < lastLine then
+      logWarningAt stx <| .tagged linter.longFile.name
+        m!"This file is {lastLine} lines long, but the limit is {linterBound}.\n\n\
+          You can extend the default length of the file {endMsg}"
+    else
+    let candidate := max candidate 1500
```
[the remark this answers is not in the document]
adomani: I disallowed forcing "short" files.  I agree that currently no file is required to be short, so allowing to set a value of less than 1500 is not really justified.

adomani (commented):
Thanks Michael!

I accepted all your suggestions!

grunweg (commented):
Thanks! I took a look again, realised the logic wasn't fully doing what it was supposed to do, and ended up staring at this for long enough that I fixed it myself. Careful checking is welcome :-)

Mathlib/Tactic/Linter/Lint.lean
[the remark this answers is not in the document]
grunweg: That is reasonable, good idea!

I took a detailed look at this logic again: it would still *suggest* a new limit shorter than the default.
At the same time, the equality case also needs to be handled: if a file has 300 lines, but a limit of 2000, something should get printed --- ideally, a suggestion to remove the set_option. I added both.

Yes, this code is more subtle than I would like...

grunweg:
Given that I re-wrote some of the code, I prefer to wait for a moment to give you the chance to review these changes.
If you're also happy, one of us can maintainer merge on behalf of both, or so :-)

adomani:
I did not like using `default` because `default : Nat` is already 0: Lean did not get confused, but I prefer to avoid even giving it a chance of getting it wrong.

Also, I clearly did not think enough about the conditions.  I pushed something, but I'm not sure if it is correct now.

adomani:
I did not like using `default` because `default : Nat` is already 0: Lean did not get confused, but I prefer to avoid even giving it a chance of getting it wrong.

Also, I clearly did not think enough about the conditions.  I pushed something, but I'm not sure if it is correct now.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,73 @@ initialize addLinter cdotLinter
...
+def longFileLinter : Linter where run := withSetOptionIn fun stx ↦ do
+  let linterBound := linter.longFile.get (← getOptions)
+  if linterBound == 0 then
+    return
+  let defValue := linter.longFile.defValue
+  let setO := match stx with
+      | `(set_option linter.longFile $x) =>
+        if TSyntax.getNat (⟨x.raw⟩) ≤ 1500 then
+          true
+        else false
```
adomani, line 243:
```suggestion
      | `(set_option linter.longFile $x) => TSyntax.getNat ⟨x.raw⟩ ≤ 1500
```

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,73 @@ initialize addLinter cdotLinter
...
+  if setO then logWarningAt stx m!"The default value of the `longFile` linter is {defValue}.\n\
+                                  The current bound of {linterBound} smaller than allowed. \
+                                  Please, remove the `set_option linter.longFile {linterBound}`."
+
+  -- thanks to the above check, the linter option is either not set (and hence equal
+  -- to the default) or set to some value *larger* than the default
+  unless stx.isOfKind ``Lean.Parser.Command.eoi do return
+  -- We exclude `Mathlib.lean` from the linter: it exceeds linter's default number of allowed
+  -- lines, and it is an auto-generated import-only file.
+  -- TODO: if there be more such files, revise the implementation.
```
adomani, line 254:
```suggestion
  -- TODO: if there are more such files, revise the implementation.
```

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,65 @@ initialize addLinter cdotLinter
...
+  let linterBound := linter.longFile.get (← getOptions)
+  if linterBound == 0 then
+    return
+  let defVal := linter.longFile.defValue
+  if linterBound < linter.longFile.defValue then
+    logWarningAt stx
+        m!"The default value of the `longFile` is {defVal}. \
+          The current bound is {linterBound} and it is smaller than the allowed bound.\n\n\
+          Please, remove the `set_option linter.longFile {linterBound}`.\n\
+          You can completely disable this linter by setting the length limit to `0`."
```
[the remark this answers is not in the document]
adomani: I have a slight preference for keeping the "disable with `0`": this is what `logLint` would say.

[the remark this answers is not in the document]
grunweg: Sure, I agree in general! Just in this case, suggesting the disabling logic feels weird when the `set_option` can just be *removed*: that is even simpler...

grunweg:
> I did not like using `default` because `default : Nat` is already 0: Lean did not get confused, but I prefer to avoid even giving it a chance of getting it wrong.
> 
> Also, I clearly did not think enough about the conditions. I pushed something, but I'm not sure if it is correct now.

:scream: I had no idea about `default`. Can you tell me more what I should be concerned about?
(What about `defaultValue`) then? I just do not find `defValue` very nice to read...

Mathlib/Tactic/Linter/Lint.lean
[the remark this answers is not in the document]
adomani: I think that this is fixed by the current iteration of the linter.

Mathlib/Tactic/Linter/Lint.lean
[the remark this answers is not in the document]
adomani: Ok, this feature is likely only useful in the test files: I added a comment whenever I used it there.

adomani:
`default` is simply the name of the unique field of the `Inhabited` class (which is `export`ed to the root namespace).  Since `Nat` has an `Inhabited` instance (with default value `0`), writing `let default := (x : Nat)` creates a shadowing between `(Inhabited.)default` and the local `let`-bound `default`.

We can use `defaultValue`, but I preferred `defValue` since that is the name of the `linter.longLine` field, there is no actual shadowing in this case, since the option is not exported and the hypothetically shadowed values would be equal anyways!  It just felt like less baggage to carry: this natural number already has (almost) this name.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -247,6 +247,80 @@ initialize addLinter dollarSyntaxLinter
...
+The "longFile" linter emits a warning on files which are longer than a certain number of lines
+(1500 by default). If this option is set to `N` lines, the linter warns once a file has more than
+`N` lines. A value of `0` silences the linter entirely.
+-/
+register_option linter.longFile : Nat := {
+  defValue := 1500
+  descr := "enable the longFile linter"
+}
+
+namespace LongFile
```
grunweg: Per upcoming PRs, I think it's better to move this to the style namespace.
```suggestion
namespace Style.longFile

File: Mathlib/Tactic/Linter/Lint.lean [hunk 7]
```diff
@@ -247,6 +247,80 @@ initialize addLinter dollarSyntaxLinter
...
+      if linterBound != candidate then
+        logWarningAt stx <| .tagged linter.longFile.name
+          m!"For this file, the recommended limit for the number of lines is {candidate}, \
+            instead of {linterBound}.\n\n\
+            Please adjust the limit to the recommended value using \
+            `set_option linter.longFile {candidate}`."
+
+initialize addLinter longFileLinter
+
+end LongFile
```
grunweg: Per the above:
```suggestion
end Style.longFile

File: Mathlib/Tactic/Linter/Lint.lean [hunk 8]
```diff
@@ -247,6 +247,80 @@ initialize addLinter dollarSyntaxLinter
...
+}
+
+namespace LongFile
+
+@[inherit_doc Mathlib.Linter.linter.longFile]
+def longFileLinter : Linter where run := withSetOptionIn fun stx ↦ do
+  let linterBound := linter.longFile.get (← getOptions)
+  if linterBound == 0 then
+    return
+  let defValue := linter.longFile.defValue
```
grunweg:
```suggestion
  let defValue := linter.style.longFile.defValue

File: Mathlib/Tactic/Linter/Lint.lean [hunk 9]
```diff
@@ -247,6 +247,80 @@ initialize addLinter dollarSyntaxLinter
...
+register_option linter.longFile : Nat := {
+  defValue := 1500
+  descr := "enable the longFile linter"
+}
+
+namespace LongFile
+
+@[inherit_doc Mathlib.Linter.linter.longFile]
+def longFileLinter : Linter where run := withSetOptionIn fun stx ↦ do
+  let linterBound := linter.longFile.get (← getOptions)
```
grunweg:
```suggestion
  let linterBound := linter.style.longFile.get (← getOptions)

**Comment by grunweg** [comment]
I just merged master, and pushed a commit moving the linter into the `Style` namespace. Full review incoming...

File: Mathlib/Tactic/Linter/Lint.lean [hunk 10]
```diff
@@ -293,6 +293,80 @@ initialize addLinter lambdaSyntaxLinter
...
+  if smallOption then
+    logWarningAt stx <| .tagged linter.style.longFile.name
+      m!"The default value of the `longFile` linter is {defValue}.\n\
+        The current value of {linterBound} does not exceed the allowed bound.\n\
+        Please, remove the `set_option linter.style.longFile {linterBound}`."
+  else
+  -- Thanks to the above check, the linter option is either not set (and hence equal
+  -- to the default) or set to some value *larger* than the default.
+  -- `Parser.isTerminalCommand` allows `stx` to be `#exit`: this is useful for tests.
+  unless Parser.isTerminalCommand stx do return
```
grunweg, line 333: Great that you found a way to test this! Let's add a comment about user implications: how about the following?
```
-- (This also means the linter will stop at an #exit command: if the user sets this linter option *before* the #exit, this could generate slightly confusing warnings if the linter suggests lowering the value.
-- Hopefully, in practice the linter is only changed at the end of the file, so this will not matter.)
```

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -293,6 +293,80 @@ initialize addLinter lambdaSyntaxLinter
...
+  -- We exclude `Mathlib.lean` from the linter: it exceeds linter's default number of allowed
+  -- lines, and it is an auto-generated import-only file.
+  -- TODO: if there are more such files, revise the implementation.
+  if (← getMainModule) == `Mathlib then return
+  if let some init := stx.getTailPos? then
+    -- the last line: we subtract 1, since the last line is expected to be empty
+    let lastLine := ((← getFileMap).toPosition init).line
+    if lastLine ≤ defValue && defValue < linterBound then
```
grunweg, line 341: And the equality case (of setting the linter to the default value) is handled above. I am fully happy with this check now!

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -293,6 +293,80 @@ initialize addLinter lambdaSyntaxLinter
...
+      logWarningAt stx <| .tagged linter.style.longFile.name
+        m!"The default value of the `longFile` linter is {defValue}.\n\
+          This file is {lastLine} lines long which does not exceed the allowed bound.\n\
+          Please, remove the `set_option linter.style.longFile {linterBound}`."
+    else
+    -- `candidate` is divisible by `100` and satisfies `lastLine + 100 < candidate ≤ lastLine + 200`
+    -- note that `candidate` is necessarily bigger than `lastLine` and hence bigger than `defValue`
```
grunweg, line 348: This is not fully true: if linterBound == defValue and lastLine <= defValue, we also end up here.
(In this case, we will not warn and all is fine. But the comment is inaccurate. This is also why the `candidate := max candidate defValue` is *necessary*.)

How about moving this comment, and the `max` line into the `else` branch below? In the first branch, it is not needed. (Did I recommend to move this check out, back then? If so, I'm sorry, and need not move it back (and just fix the comment). I do not intend to be Goldilocks' reviewer.)

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -293,6 +293,80 @@ initialize addLinter lambdaSyntaxLinter
...
+    let candidate := (lastLine / 100) * 100 + 200
+    let candidate := max candidate defValue
+    if linterBound < lastLine then
+      logWarningAt stx <| .tagged linter.style.longFile.name
+        m!"This file is {lastLine} lines long, but the limit is {linterBound}.\n\n\
+          You can extend the allowed length of the file using \
+          `set_option linter.style.longFile {candidate}`.\n\
+          You can completely disable this linter by setting the length limit to `0`."
+    else -- note that the inequality `linterBound ≤ defValue` holds
+    if lastLine + 200 < linterBound then
```
grunweg, line 358: In the text-based linter, I used a slightly different algorithm --- which is more sticky, as it only lowers the option in steps of at least 200 lines. (Currently, this linter will warn more often, e.g. a file going from 2520 lines (with candidate of 2700) to 2480 will yield a warning. I think that's slightly overzealous.) What do you think?
```suggestion
    if candidate + 200 <= linterBound then

File: test/Lint.lean [hunk 14]
```diff
@@ -254,3 +254,64 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+
+/--
+warning: using 'exit' to interrupt Lean
+---
+warning: The default value of the `longFile` linter is 1500.
+This file is 166 lines long which does not exceed the allowed bound.
+Please, remove the `set_option linter.style.longFile 1600`.
+-/
+#guard_msgs in
+-- Do not allow increasing unnecessarily the `longFile` linter option
```
grunweg, line 289: Not a native speaker, but is this better grammar? (If you know better, feel to ignore this/enlighten me!)
```suggestion
-- Do not allow unnecessarily increasing the `longFile` linter option

File: test/Lint.lean [hunk 15]
```diff
@@ -254,3 +254,64 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+/--
+warning: using 'exit' to interrupt Lean
+---
+warning: This file is 181 lines long, but the limit is 10.
+
+You can extend the allowed length of the file using `set_option linter.style.longFile 1500`.
+You can completely disable this linter by setting the length limit to `0`.
+-/
+#guard_msgs in
+-- First, we silence the linter, so that we can set a default value smaller than 1500
```
grunweg, line 302:
```suggestion
-- First, we silence the linter, so that we can set a default value smaller than 1500.

File: test/Lint.lean [hunk 16]
```diff
@@ -254,3 +254,64 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+set_option linter.style.longFile 0 in
+-- Next, we test that the `longFile` linter warns when a file exceeds the allowed value
```
grunweg, line 304:
```suggestion
-- Next, we test that the `longFile` linter warns when a file exceeds the allowed value.

File: test/Lint.lean [hunk 17]
```diff
@@ -254,3 +254,64 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+
+/--
+warning: using 'exit' to interrupt Lean
+---
+warning: The default value of the `longFile` linter is 1500.
+This file is 196 lines long which does not exceed the allowed bound.
+Please, remove the `set_option linter.style.longFile 1700`.
+-/
+#guard_msgs in
+-- First, we silence the linter, so that we can set a default value smaller than 1500
```
grunweg, line 316:
```suggestion
-- First, we silence the linter, so that we can set a default value smaller than 1500.

File: test/Lint.lean [hunk 18]
```diff
@@ -254,3 +254,64 @@ info: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
...
+set_option linter.style.longFile 0 in
+-- If we set the allowed bound for the `longFile` linter that is too large,
+-- then the linter tells us to use a smaller bound
```
grunweg, line 319:
```suggestion
-- the linter tells us to use a smaller bound.

**Review by grunweg — COMMENTED** [review]
Sorry for the delay; I now gave this a final look. I agree with the logic now; I merely suggested a few minor phrasing tweaks, one tweak to the logic (to mirror the text-based linter) and found one comment that seems outdated.

File: Mathlib/Tactic/Linter/Lint.lean (hunk 10 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/Lint.lean:332]: I do not understand what this comment is clarifying: this is probably due to the fact that I have thought quite a bit about the linter and I am failing to see what can be confusing about it!

Can you explain what confusion this comment alleviates?  :smile:

File: Mathlib/Tactic/Linter/Lint.lean (hunk 12 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/Lint.lean:347]: `candidate` is used in this branch, since the linter suggests a value for its option with which it will be happy.

I updated the comment to include the other possibility, though.

File: Mathlib/Tactic/Linter/Lint.lean (hunk 13 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/Lint.lean]: Great, I like this!

**Review by adomani — COMMENTED** [review]
Michael, thanks a lot for the careful review!  I left a couple of comments and implemented the rest of the changes!

**Comment by adomani** [comment]
Btw, we could *enforce* that `set_option linter.style.longFile x` is only followed by `#exit` or the end of the file by making the linter say something if it parses a non-terminal command when the option is not set to the default value.

Would this be an improvement?  Or is it ok if the option can be set anywhere in the file?

File: Mathlib/Tactic/Linter/Lint.lean (hunk 2 above)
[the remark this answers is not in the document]
Comment (grunweg) [reply Mathlib/Tactic/Linter/Lint.lean]: So did I!

File: Mathlib/Tactic/Linter/Lint.lean (hunk 10 above)
Comment (grunweg) [reply Mathlib/Tactic/Linter/Lint.lean:332]: I worried about the following scenario:
- in a long file, the set_option is set "early" (say, at the beginning of the file)
- the user temporarily writes #exit, say, to debug something
This exit is placed *after* the set_option, but early enough that *at this point*, the linter option is set too high.

Then, the linter will complain about the option being set too high --- but that complain is spurious (it will disappear once the #exit is removed) and following the warning will yield *another* (correct, non-spurious) warning when the #exit is removed again.

Granted, this scenario is a bit convoluted (and only a temporary state).
Given this, your suggestion of warning about the option being set early seems useful (if perf. allows it; otherwise, a comment should suffice).

File: Mathlib/Tactic/Linter/Lint.lean (hunk 12 above)
Comment (grunweg) [reply Mathlib/Tactic/Linter/Lint.lean]: Thanks. This looks correct now.

**Comment by grunweg** [comment]
As I said above: checking for an early set_option sounds useful --- but only if this is fine performance-wise. The gain (avoiding a confusing, but presumably very rare scenario) is not *that* large, after all.

**Comment by grunweg** [comment]
Otherwise: would you like to remove the text-based linter now, or shall e.g. I submit a follow-up?

File: Mathlib/Tactic/Linter/Lint.lean (hunk 10 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/Lint.lean:332]: I see.  Honestly, if someone knows that `#exit` shielded inside a `xxx in #exit` allows Lean to continue parsing the file after `#exit`, I think that they will not be confused by the linter for too long!  I would leave this as is, though I can implement the forcing the `set_option` to be the "last" command of a file.

**Comment by adomani** [comment]
I usually prefer to do the removal in a separate PR, since if, for any reason, it needs to be reverted, you can more easily revert the removal and decide if you also want to revert the addition of the new linter.

File: Mathlib/Tactic/Linter/Lint.lean (hunk 10 above)
Comment (grunweg) [reply Mathlib/Tactic/Linter/Lint.lean:332]: Sure - this is not something I consider essential. While minimising surprises is nice and these things add up, there are probably more useful changes one could make in that regard.

**Comment by grunweg** [comment]
In this case: assuming CI passes now, let me
maintainer merge

Thanks a lot for your patience throughout this long review process. I am (still) very happy for the text-based linter to get rewritten in nicer form, and look forward to removing the removing ones over time :-)

File: Mathlib/Tactic/Linter/Lint.lean [hunk 19]
```diff
@@ -292,6 +292,81 @@ initialize addLinter lambdaSyntaxLinter
...
+        m!"The default value of the `longFile` linter is {defValue}.\n\
+          This file is {lastLine} lines long which does not exceed the allowed bound.\n\
+          Please, remove the `set_option linter.style.longFile {linterBound}`."
+    else
+    -- `candidate` is divisible by `100` and satisfies `lastLine + 100 < candidate ≤ lastLine + 200`
+    -- note that either `lastLine ≤ defValue` and `defValue = linterBound` hold or
+    -- `candidate` is necessarily bigger than `lastLine` and hence bigger than `defValue`
+    let candidate := (lastLine / 100) * 100 + 200
+    let candidate := max candidate defValue
+    if linterBound < lastLine then
```
mattrobball, line 351: Maybe move this above the definition of `candidate` since it is independent?

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -292,6 +292,81 @@ initialize addLinter lambdaSyntaxLinter
...
+      logWarningAt stx <| .tagged linter.style.longFile.name
+        m!"This file is {lastLine} lines long, but the limit is {linterBound}.\n\n\
+          You can extend the allowed length of the file using \
+          `set_option linter.style.longFile {candidate}`.\n\
+          You can completely disable this linter by setting the length limit to `0`."
+    else -- note that the inequality `linterBound ≤ defValue` holds
+    if candidate + 200 <= linterBound then
+      if linterBound != candidate then
```
mattrobball, line 359: I don't think x + 200 \leq x.

Mathlib/Tactic/Linter/Lint.lean
adomani: I think that `candidate` is used in the `if` to print something in the warning.  Or am I misunderstanding what you are suggesting?

Mathlib/Tactic/Linter/Lint.lean
adomani: Oops!  The logic behind when to lint what changed so many times that this is probably a leftover from a version where it made sense!  😊

Mathlib/Tactic/Linter/Lint.lean
mattrobball: Sorry. I didn't read carefully.

Mathlib/Tactic/Linter/Lint.lean
adomani: I just removed the final `else ...` block.

mattrobball:
bors merge

PR #15611, opened 2024-08-08
[Merged by Bors] - perf(Algebra.Ring.BooleanRing): scope simp theorems with weak keys but complex parameters 

mattrobball:
The keys of these are the form `HAdd.hAdd _ _ _ _ _ _` etc. but they require a typeclass search for `BooleanRing` which extends `Ring`. As they are only meant to apply to `BooleanRing`s, we add a namespace and scope them to `BooleanRing`.

---


mattrobball:
Actually I should just add the attribute to the field itself.

bryangingechen:
(I think I wrote some of the lean3 version of this code, but I haven't looked at it since before the port.) What's the reason for adding a prime to `mul_self`?

j-loreaux:
@mattrobball it looks like you gave yourself a TODO here, so I added `awaiting-author`

bryangingechen:
Thanks!
bors r+

PR #15612, opened 2024-08-08
[Merged by Bors] - chore: backports for leanprover/lean4#4814 (part 36)

Parcly-Taxel:
---

### DIFF
--- a/Archive/Imo/Imo2024Q2.lean
+++ b/Archive/Imo/Imo2024Q2.lean
@@ -47,6 +47,9 @@ namespace Condition
 
 variable {a b : ℕ} (h : Condition a b)
 
+section
+include h
+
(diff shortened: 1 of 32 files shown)


Archive/Imo/Imo2024Q2.lean
```diff
@@ -47,6 +47,8 @@ namespace Condition
 
 variable {a b : ℕ} (h : Condition a b)
 
+section
```
semorrison, line 50: What does this `section` achieve? It seems it could be removed.

semorrison: Oh, I see, it has an `include` that wasn't backported. I wasn't doing these, but sure, no problem.

semorrison:
bors merge

PR #15616, opened 2024-08-08
[Merged by Bors] - feat: reset options using moreServerArgs

adomani:
The change sets the default value of several linters to be false, so that they are not imposed on dependencies of mathlib, but sets them on for mathlib.

#15521 removes the ad-hoc crippling of some of the linters, to prevent them from acting on downstream projects.  Even though #15521 is not technically a dependency of this one, it probably makes sense to not merge #15521 before this one (and it should definitely be merged simultaneously/just after this one!).

The [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Improper.20firing.20of.20linter.2Ehashcommand/near/457284896) about this settled on the following choices for global vs mathlib-only linter options:

Global:

* `globalAttributeIn`
* `haveLetLinter`
* `dupNamespace`
* `unusedTacticLinter`

Mathlib local:

* `hashCommandLinter`
* `missingEndLinter`
* `cdotLinter`
* `longLineLinter`
* `oldObtainLinter`
* `refineLinter`
* `setOptionLinter`

---


grunweg:
Thanks for doing this! A few quick questions/requests; I'm happy to review in more detail next week.
- Can you document (say, in the commit message, or somewhere in an overview comment) which linters are enabled by default and which ones are not? Eric's comment on zulip seems spot on, copy-pasting that is fine :-) I'd just like to retain that knowledge somewhere besides "some zulip thread".
- Does this still enable the linters for `Archive` and `Counterexamples`? What about `tests`? (I quite liked that the linters previously did apply to them...)

adomani:
I updated the message above, adding Mario's triaging of the linter options.

As for `Archive` and `Counterexamples`, the linters are not active there by default.  I agree that this is undesirable, but I am not sure how to make them act there, except by adding each set_option on each file :disappointed:

joneugster:
> As for `Archive` and `Counterexamples`, the linters are not active there by default. I agree that this is undesirable, but I am not sure how to make them act there, except by adding each set_option on each file 😞

I think you can just add this to the lakefile, can't you? Like this:

```
lean_lib Archive where
  leanOptions := mathlibOnlyLinters.map fun s => { s with name := `weak ++ s.name }
  moreServerOptions := mathlibOnlyLinters
```

grunweg:
> > As for `Archive` and `Counterexamples`, the linters are not active there by default. I agree that this is undesirable, but I am not sure how to make them act there, except by adding each set_option on each file 😞
> 
> I think you can just add this to the lakefile, can't you? Like this:
> 
> ```
> lean_lib Archive where
>   leanOptions := mathlibOnlyLinters.map fun s => { s with name := `weak ++ s.name }
>   moreServerOptions := mathlibOnlyLinters
> ```
> 
> (and same for `test`)
> 
> I suggest, you abstract the part
> 
> ```
> #[ ⟨`pp.unicode.fun, true⟩, -- pretty-prints `fun a ↦ b`
>    ⟨`autoImplicit, false⟩ ] ++ mathlibOnlyLinters.map fun s => { s with name := `weak ++ s.name }
> ```
> 
> as well and add the options to all 3 as @grunweg suggested.

I was not aware that worked - this is great! (Otherwise, I would have suggested adding these options in the .yml files, which is clearly not desirable.

grunweg:
(I'm not sure if tests require `autoImplicit false`, but adding it as necessary should not be too hard.

grunweg:
I took the liberty to implement @joneugster's suggestion. I am still amazed that lakefiles allow this!

grunweg:
I did some local testing: this works well for `Archive` and `Counterexamples`.
It does not work for `tests`; the linters are disabled there --- but this is the status quo: this appears to be because `tests` is a `lean_exe`; making it a `lean_lib` makes the configuration take effect. (I remember a known bug along these lines; I'm not sure if there is an issue for this.)

joneugster (approved):
Looks good to me!

If somebody wants to search for a lean4 issue about `lean_exe` to link... but I think the comment is good enough for now.

maintainer merge

grunweg:
Note to self: make sure to adjust for whichever of #15896 and this one lands first.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15620, opened 2024-08-08
perf(Algebra.CharTwo/Zero): scope lemmas 

mattrobball:
These have very general keys but narrow application. We scope them to their namespaces.

In particular, the `no_index`ed `ofNat` simp theorems are now explicitly opted into. If the handling of `ofNat` changes in core, this makes it easy to update.

---

### DIFF
--- a/Archive/Imo/Imo1960Q2.lean
+++ b/Archive/Imo/Imo1960Q2.lean
@@ -36,6 +36,7 @@ structure IsGood (x : ℝ) : Prop where
   /-- The number belongs to the domain of the denominator. -/
   denom_dom : (1 - sqrt (2 * x + 1)) ^ 2 ≠ 0
 
+open scoped OfNat in -- to use `no_index`ed simp lemmas for `ofNat`
 /-- Solution of IMO 1960 Q2: solutions of the inequality
 are the numbers of the half-closed interval \([-1/2, 45/8)\) except for the number zero. -/
 theorem isGood_iff {x} : IsGood x ↔ x ∈ Ico (-1/2) (45/8) \ {0} := by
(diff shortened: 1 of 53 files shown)


mattrobball:
!bench

mattrobball:
> Here are the [benchmark results](http://speed.lean-fro.org/mathlib4/run-detail/c535d7a7-2c9f-4d9d-afa2-b23ef2a17e69) for commit [754a204](https://github.com/leanprover-community/mathlib4/commit/754a2040263dd3a4978471f27cafcbbb63b33b81). There were [no significant changes](http://speed.lean-fro.org/mathlib4/compare/085156a8-725b-499f-b0b5-4b868c90fc3b/to/c535d7a7-2c9f-4d9d-afa2-b23ef2a17e69) against commit [2edef7d](https://github.com/leanprover-community/mathlib4/commit/2edef7d509a7aeba89793a84905121a9f19171b6).

Dropped total instructions by 200B and linting instructions by 100B/1%

j-loreaux:
While I like the idea of this, I feel like it's going to be hard to remember, and annoying to need to explicitly open this scope. What I suspect is likely is that people are going to expect a `simp` lemma to be capable of handling something, it will fail, and they'll have no idea why.

It was necessary to `open scoped ofNat in` many places.

mattrobball:
Let me split the CharTwo/CharZero scoping and the OfNat issues. 

There is some hope that the core issues with handling of OfNat will be resolved at some point. I will say more on the split.

mattrobball:
#15631 does everything but the `OfNat` simp theorems and should be straightforward

mattrobball:
#15631 and #15676 achieve most of the performance gains with no downsides. 

Closing this

PR #15622, opened 2024-08-08
[Merged by Bors] - perf(RepresentationTheory.LowDegree): scope simp theorems with weak keys

mattrobball:
Both of these have `HSMul.hSMul _ _ _ _ _ _` as keys so apply to any occurrence of scalar multiplication.

---


j-loreaux:
bors merge

PR #15623, opened 2024-08-08
[Merged by Bors] - perf(NumberTheory.MulChar): scope simp theorem with weak keys

mattrobball:
The keys for this `DFunLike.coe _ _ _ _ _ _` so applly to any expression that has a `DFunLike.coe` subexpression.

---


j-loreaux:
bors merge

PR #15625, opened 2024-08-08
[Merged by Bors] - feat(MeasureTheory): add `ae_restrict_of_forall_mem`

dupuisf:
I found myself repeatedly needing this lemma.

---


j-loreaux:
maintainer merge

urkud:
Why not use `rw [ae_restrict_iff']` instead?

dupuisf:
I was using `ae_restrict_iff_subtype` directly at the beginning, but after four or five times I thought it was time to write the lemma I actually wanted.

urkud:
Another option is `(ae_restrict_mem _).mono`. If you prefer your lemma, then we can add it, but please move it near `ae_restrict_mem`.
bors d+

PR #15626, opened 2024-08-08
[Merged by Bors] - perf(Data.Nat.Cast.Basic): de-simp theorem with weak keys

mattrobball:
The keys are `DFunLike.coe _ _ _ _ _ _` which means we are trying to synthesize two `NonAssocSemiring`'s each time any expression has this as a subterm.

---


mattrobball:
!bench

j-loreaux:
ooo, I like these PRs.

bors merge

PR #15628, opened 2024-08-08
[Merged by Bors] - perf(Algebra.GradedMonoid): scope theorem with weak keys 

mattrobball:
The keys for this theorem are `HSMul.hSMul _ _ _ _ _ _` so will be attempted on any term which has an `SMul` in it.

---


eric-wieser (approved):
Sure, this GradeZero stuff is dubious anyway.

Just to check, what scope does this end up going in?

mattrobball:
`GradedMonoid`

joneugster:
maintainer merge

PR #15629, opened 2024-08-08
refactor(Analysis/CStarAlgebra/Module): change structure fields for convenience

j-loreaux:
---


j-loreaux:
After playing with this further, I think I've decided it was a terrible idea.

PR #15632, opened 2024-08-09
feat(ContinuousMap): A family `f : α → C(β, γ)` is continuous if it is pointwise continuous and `β` is compact

dupuisf:
A family of continuous maps `f : α → C(β, γ)` is continuous on an open set if it is pointwise continuous, provided that `β` is compact and `α` is locally compact. 

---


Mathlib/Topology/ContinuousFunction/Compact.lean
```diff
@@ -330,6 +330,17 @@ theorem dist_lt_of_dist_lt_modulus (f : C(α, β)) (ε : ℝ) (h : 0 < ε) {a b
     (w : dist a b < f.modulus ε h) : dist (f a) (f b) < ε :=
   (Classical.choose_spec (uniform_continuity f ε h)).2 w
 
+/-- A family of continuous maps `f : α → C(β, γ)` is continuous on an open set if it is
+  pointwise continuous, provided that `β` is compact and `α` is locally compact. -/
+lemma continuousOn_of_pointwise_continuousOn_of_compact {α : Type*} {β : Type*} {γ : Type*}
+    [UniformSpace α] [UniformSpace β] [LocallyCompactSpace α] [CompactSpace β] [UniformSpace γ]
```
urkud, line 336: Is it true for topological spaces?

dupuisf: I don't know. If it is I have no idea how to prove it.

AntoineChambert-Loir: It is not important that `α` be a uniform space, because it the continuity is only used locally, and it is locally compact, and compact spaces have a canonical uniform structure.

AntoineChambert-Loir: Same for `β`.

dupuisf: I see, thanks!

urkud:
`ContinuousMap.curry` says that you don't need any assumptions on the topologies or on the set `U`.

dupuisf:
> `ContinuousMap.curry` says that you don't need any assumptions on the topologies or on the set `U`.

I was not aware of  `ContinuousMap.curry`; I'll just close this PR now since I can just use that instead of this lemma.

PR #15638, opened 2024-08-09
[Merged by Bors] - chore: cleanup Init imports

kim-em:


j-loreaux:
I think the 2 increases are probably okay, but why are they necessary?

semorrison:
> I think the 2 increases are probably okay, but why are they necessary?

I dug around for a moment (manually, not actually listing the transitive imports...), but I'm failing to understand what has increased.

I'm not too worried here, the imports are very basic and well understood.

j-loreaux:
bors merge

PR #15641, opened 2024-08-09
[Merged by Bors] - feat(CategoryTheory): Finality of functors from filtered categories into PUnit

javra:
---

### DIFF
--- a/Mathlib/CategoryTheory/Filtered/Final.lean
+++ b/Mathlib/CategoryTheory/Filtered/Final.lean
@@ -230,8 +230,8 @@ theorem Functor.initial_iff_isCofiltered_costructuredArrow [IsCofilteredOrEmpty 
 
 end LocallySmall
 
-/-- If `C` is filtered, then every functor `F : ⥤ Discrete PUnit` is final. -/
-theorem Functor.final_of_isFiltered_pUnit {C : Type u₁} [Category.{v₁} C]
+/-- If `C` is filtered, then every functor `F : C ⥤ Discrete PUnit` is final. -/
+theorem Functor.final_of_isFiltered_of_pUnit {C : Type u₁} [Category.{v₁} C]
     [IsFiltered C] (F : C ⥤ Discrete PUnit) :
     Final F := by
   refine final_of_exists_of_isFiltered F (fun _ => ?_) (fun {_} {c} _ _ => ?_)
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Filtered/Final.lean
```diff
@@ -230,4 +230,24 @@ theorem Functor.initial_iff_isCofiltered_costructuredArrow [IsCofilteredOrEmpty
 
 end LocallySmall
 
+/-- If `C` is filtered, then every functor `F : ⥤ Discrete PUnit` is final. -/
```
TwoFX, line 233:
```suggestion
/-- If `C` is filtered, then every functor `F : C ⥤ Discrete PUnit` is final. -/
```

Mathlib/CategoryTheory/Filtered/Final.lean
```diff
@@ -230,4 +230,24 @@ theorem Functor.initial_iff_isCofiltered_costructuredArrow [IsCofilteredOrEmpty
...
+theorem Functor.final_of_isFiltered_pUnit {C : Type u₁} [Category.{v₁} C]
+    [IsFiltered C] (F : C ⥤ Discrete PUnit) :
+    Final F := by
+  refine final_of_exists_of_isFiltered F (fun _ => ?_) (fun {_} {c} _ _ => ?_)
+  · use Classical.choice IsFiltered.nonempty
+    exact ⟨Discrete.eqToHom (by simp)⟩
+  · use c; use 𝟙 c
+    apply Subsingleton.elim
+
+/-- If `C` is cofiltered, then every functor `F : ⥤ Discrete PUnit` is initial. -/
```
TwoFX, line 243:
```suggestion
/-- If `C` is cofiltered, then every functor `F : C ⥤ Discrete PUnit` is initial. -/
```

Mathlib/CategoryTheory/Filtered/Final.lean
```diff
@@ -230,4 +230,24 @@ theorem Functor.initial_iff_isCofiltered_costructuredArrow [IsCofilteredOrEmpty
 
 end LocallySmall
 
+/-- If `C` is filtered, then every functor `F : ⥤ Discrete PUnit` is final. -/
+theorem Functor.final_of_isFiltered_pUnit {C : Type u₁} [Category.{v₁} C]
```
TwoFX, line 234: I would call this `final_of_isFiltered_of_pUnit` I think.

TwoFX (approved):
Thanks!
bors r+

PR #15644, opened 2024-08-09
[Merged by Bors] - feat(Combinatorics/Enumerative/Bell): number of ways of partitioning a set into subsets of given cardinalities

AntoineChambert-Loir:
* `Multiset.bell m` is the number of possibilities of dividing a set with `m.sum` elements into disjoint subsets with cardinalities indicated by `m`.
* `Nat.uniformBell m n` is `Multiset.bell (replicate m n)`, the number of possibilities of dividing a set with `m * n` elements into `m` groups of `n`-element subsets

These integers are defined as finite products of integers.

* `Multiset.bell_eq` shows that it equals some expression involving fractions of factorials, `Multiset.bell_mul_eq` shows the equivalent formula without division.

* `Nat.uniformBell_eq` shows that it equals `(m * n).factorial / (m.factorial * n.factorial ^ m)`
and `Nat.uniformBell_mul_factorial_mul_pow_factorial` establishes the equality
  `mchoose m n * m.factorial * n.factorial ^ m = (m * n).factorial`.

* `Nat.choose_mul_add` and `Nat.choose_mul_right` are lemmas for `Nat.choose`
that couldn't lie in `Mathlib.Data.Nat.Choose.Basic` because their proof uses the `ring` tactic.

Co-authored with [María Inés de Frutos-Fernández](https://github.com/mariainesdff) 

TODO : solve the enumerative problem.

---


Mathlib/Data/Nat/Choose/Mchoose.lean
```diff
@@ -0,0 +1,106 @@
...
+  by_cases hm : m = 0
+  · simp only [hm, zero_mul, choose_eq_zero_iff]
+    exact Nat.pos_of_ne_zero hn
+  · set p := m - 1; have hp : m = p + 1 := (succ_pred_eq_of_ne_zero hm).symm
+    simp only [hp]
+    rw [add_mul, one_mul, choose_mul_add _ hn]
+
+/-- Number of possibilities of dividing a set with `m * n` elements into `m` groups
+of `n`-element subsets. -/
+def mchoose (m n : ℕ) : ℕ :=
```
eric-wieser, line 57: Is this different from `Nat.multichoose`?

AntoineChambert-Loir: This has nothing to do. Maybe we should find a better name. It is a particular case of the following concept : given a finite set s and a partition p = (p_1,...,p_k) of Card(s), how many (unordered) partitions of the set are there such that its members have sizes p_1,...,p_k. Here we take s of cardinality n * m and p = (n,...,n).

AntoineChambert-Loir: (I would likely call them restricted Bell numbers, but that does not seem to be the accepted terminology.)
A definition (using a division) is 
```quote
def rBell (m : Multiset ℕ) :=  (m.sum)! / ((m.map fun n ↦ n !).prod * (∏ n in m.toFinset, (m.count n)!))
```
and then, one has
```
mchoose m n = rBell (Multiset.replicate m n)
```

AntoineChambert-Loir: (I have formalized that this works, but this is clearly off topic, and, unfortunately uses the properties of that PR. The difficulty is to prove that the denominator divides the numerator.)

eric-wieser: To be clear, my comment was primarily "we either need a better name, or documentation to help disambiguate the names, or both". I agree that this is not the same thing!

AntoineChambert-Loir: Now we have `Multiset.bell` and `Nat.uniformBell`.

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+    exact Nat.pos_of_ne_zero hn
+  · set p := m - 1; have hp : m = p + 1 := (succ_pred_eq_of_ne_zero hm).symm
+    simp only [hp]
+    rw [add_mul, one_mul, choose_mul_add _ hn]
+
+namespace Multiset
+
+/-- Number of partitions of a set of cardinality `m.sum` 
+whose parts have cardinalities given by `m` -/
+def bell (m : Multiset ℕ) : ℕ :=
```
eric-wieser: I'm now realizing that #16647 is relevant here. No action needed in this PR though.

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+
+* `Multiset.bell_mul_eq` shows that
+    `m.bell * (m.map (fun j ↦ j !)).prod *
+      Finset.prod (m.toFinset.erase 0) (fun j ↦ (m.count j)!) = m.sum !`
+
+* `Nat.uniformBell_mul_eq`  shows that
+    `uniformBell m n * n.factorial ^ m * m.factorial = (m * n).factorial`
+
+* `Nat.uniformBell_zero`, `Nat.uniformBell_zero'`, `Nat.uniformBell_one`, `Nat.uniformBell_one':
+compute `Nat.uniformBell` when one of the parameters is `0` or `1`
```
YaelDillies, line 32:
```suggestion
  compute `Nat.uniformBell` when one of the parameters is `0` or `1`
```

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+For `n : ℕ`, the `n`th Bell number is the number of partitions of a set of cardinality `n`.
+Here, we define a refinement of these numbers, that count, for any `m : Multiset ℕ`,
+the number of partitions of a set of cardinality `m.sum` whose parts have cardinalities
+given by `m`.
+
+The definition presents it as an integer.
+
+* `Multiset.bell`: number of partitions of a set whose parts have cardinalities a given multiset
+
+* `Multiset.bell_zero`: `Multiset.bell 0 = 1`
```
YaelDillies, line 20: I'm not super convinced very basic lemmas like `bell_zero` need mentioning in the module doc. They can already efficiently be found through the symbol-reading convention.

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+    _ = (m + 1) * (m * (p + 1) + p).choose p * ((m * (p + 1))! * (p + 1)!) := by ring
+
+theorem Nat.choose_mul_right (m) {n : ℕ} (hn : n ≠ 0) :
+    (m * n).choose n = m * (m * n - 1).choose (n - 1) := by
+  by_cases hm : m = 0
+  · simp only [hm, zero_mul, choose_eq_zero_iff]
+    exact Nat.pos_of_ne_zero hn
+  · set p := m - 1; have hp : m = p + 1 := (succ_pred_eq_of_ne_zero hm).symm
+    simp only [hp]
+    rw [add_mul, one_mul, choose_mul_add _ hn]
```
YaelDillies, line 71: Can these move to an earlier file?

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+
+namespace Multiset
+
+/-- Number of partitions of a set of cardinality `m.sum`
+whose parts have cardinalities given by `m` -/
+def bell (m : Multiset ℕ) : ℕ :=
+  Nat.multinomial m.toFinset (fun k ↦ k * m.count k) *
+    (m.toFinset.erase 0).prod
+      (fun k ↦ Finset.prod (Finset.range (m.count k)) fun j ↦ (j * k + k - 1).choose (k-1))
```
YaelDillies, line 80:
```suggestion
    ∏ k ∈ m.toFinset.erase 0, ∏ j ∈ .range (m.count k), (j * k + k - 1).choose (k - 1)
```

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+
+@[simp]
+theorem bell_zero : bell 0 = 1 := by unfold bell; simp
```
YaelDillies, line 83: I kind of suspect this is refl
```suggestion
theorem bell_zero : bell 0 = 1 := rfl
```

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+      rw [← Nat.choose_mul_add c hx]
+      rw [mul_comm c x]
+      rw [Nat.add_choose_mul_factorial_mul_factorial (x * c) x]
+      rw [mul_add, mul_one]
+    rw [factorial_succ, pow_succ]
+    ring_nf
+
+theorem bell_mul_eq (m : Multiset ℕ) :
+    m.bell * (m.map (fun j ↦ j !)).prod *
+      Finset.prod (m.toFinset.erase 0) (fun j ↦ (m.count j)!) = m.sum ! := by
```
YaelDillies, line 114: Similarly here and below, can you use the `∏ _ ∈ _, _` notation?

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+  · by_cases hn : n = 0
+    · simp [hn]
+    · rw [show ({n} : Finset ℕ).erase 0 = {n} by simp [Ne.symm hn]]
+      simp [count_replicate]
+
+theorem uniformBell_zero (n : ℕ) : uniformBell 0 n = 1 := by
+  simp [uniformBell_eq]
+
+theorem uniformBell_zero' (m : ℕ) : uniformBell m 0 = 1 := by
+  simp [uniformBell_eq]
```
YaelDillies, line 184: Can these be
```suggestion
theorem uniformBell_zero_left (n : ℕ) : uniformBell 0 n = 1 := by
  simp [uniformBell_eq]

theorem uniformBell_zero_right (m : ℕ) : uniformBell m 0 = 1 := by
  simp [uniformBell_eq]
```
and same below?

Mathlib/Data/Finset/Basic.lean
```diff
@@ -3026,6 +3026,13 @@ theorem disjoint_toFinset {m1 m2 : Multiset α} :
     rw [Multiset.mem_toFinset] at ha hb
     exact h ha hb
 
+@[simp]
+lemma replicate_toFinset (n : ℕ) (a : α) :
+    (replicate n a).toFinset = if n = 0 then ∅ else {a} := by
```
YaelDillies, line 3031:
```suggestion
lemma toFinset_replicate (n : ℕ) (a : α) :
    (replicate n a).toFinset = if n = 0 then ∅ else {a} := by
```

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,217 @@
...
+  · simp only [replicate_toFinset]
+    split_ifs with hm
+    · rw [hm, factorial_zero, eq_comm]
+      rw [show (∅ : Finset ℕ).erase 0 = ∅ from rfl,  Finset.prod_empty]
+    · rw [show ({n} : Finset ℕ).erase 0 = {n} by simp [Ne.symm hn]]
+      simp only [Finset.prod_singleton, count_replicate_self]
+  · simp
+
+theorem uniformBell_eq_div (m : ℕ) {n : ℕ} (hn : n ≠ 0) :
+    uniformBell m n = (m * n).factorial / (n.factorial ^ m * m.factorial) := by
```
YaelDillies, line 211:
```suggestion
    uniformBell m n = (m * n) ! / (n ! ^ m * m !) := by
```
and above as well

YaelDillies (commented):
Nice work! Can you merge master? The primed lemma names are now linted against

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: I remember I had tried, but that the use of the tactic calc was not possible in the file that defined `Nat.Choose`. I'll check again.

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: Done for the third factor (the `Finset.prod`), is there a notation for Multiset.prop?

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: Another issue is that I would have to import Mathlib.Algebra.GroupWithZero.Defs

AntoineChambert-Loir: That one could be OK, actually, this is `ring` which is missing. If I import `Tactic.Ring.RingNF`, that works.

AntoineChambert-Loir: That was eventually not possible, so I added a new file.

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,184 @@
...
+    ∏ k ∈ m.toFinset.erase 0, ∏ j ∈ .range (m.count k), (j * k + k - 1).choose (k - 1)
+
+@[simp]
+theorem bell_zero : bell 0 = 1 := rfl
+
+private theorem bell_mul_eq_lemma {x : ℕ} (hx : x ≠ 0) (c : ℕ) :
+    x ! ^ c * c ! * ∏ j ∈ Finset.range c, (j * x + x - 1).choose (x - 1) = (x * c)! := by
+  induction c with
+  | zero => simp
+  | succ c hrec =>
```
YaelDillies, line 56: You can use pattern-matching here
```suggestion
private theorem bell_mul_eq_lemma {x : ℕ} (hx : x ≠ 0) :
    ∀ c, x ! ^ c * c ! * ∏ j ∈ Finset.range c, (j * x + x - 1).choose (x - 1) = (x * c)!
  | 0 => by simp
  | c + 1 =>
```

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,184 @@
...
+      have : (x * c)! * ((c * x + x - 1).choose (x - 1)) * x ! * (c + 1)
+        = ((c + 1) * ((c * x + x - 1).choose (x - 1))) * (x * c)! *  x ! := by
+        ring_nf
+      rw [this]
+      rw [← Nat.choose_mul_add c hx]
+      rw [mul_comm c x]
+      rw [Nat.add_choose_mul_factorial_mul_factorial (x * c) x]
+      rw [mul_add, mul_one]
+    rw [factorial_succ, pow_succ]
+    ring_nf
```
YaelDillies, line 77: This looks like a good candidate for a `calc` proof. Do you mind trying?

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,184 @@
...
+
+theorem bell_mul_eq (m : Multiset ℕ) :
+    m.bell * (m.map (fun j ↦ j !)).prod *
+    ∏ j ∈ (m.toFinset.erase 0), (m.count j)! = m.sum ! := by
+  unfold bell
+  rw [← Nat.mul_right_inj]
+  · simp only [← mul_assoc]
```
YaelDillies, line 84:
```suggestion
  rw [← Nat.mul_right_inj (by positivity)]
  simp only [← mul_assoc]
```
and everything below unindented

Mathlib/Data/Finset/Basic.lean
```diff
@@ -3116,6 +3116,13 @@ theorem disjoint_toFinset {m1 m2 : Multiset α} :
     _root_.Disjoint m1.toFinset m2.toFinset ↔ Disjoint m1 m2 := by
   simp [disjoint_left, Finset.disjoint_left]
 
+@[simp]
+lemma toFinset_replicate (n : ℕ) (a : α) :
+    (replicate n a).toFinset = if n = 0 then ∅ else {a} := by
+  ext x
+  simp only [mem_toFinset, Finset.mem_singleton, mem_replicate]
```
YaelDillies, line 3123:
```suggestion
```

Mathlib/Data/Nat/Choose/Mul.lean
```diff
@@ -0,0 +1,48 @@
...
+theorem choose_mul_right (m) {n : ℕ} (hn : n ≠ 0) :
+    (m * n).choose n = m * (m * n - 1).choose (n - 1) := by
+  by_cases hm : m = 0
+  · simp only [hm, zero_mul, choose_eq_zero_iff]
+    exact Nat.pos_of_ne_zero hn
+  · set p := m - 1; have hp : m = p + 1 := (succ_pred_eq_of_ne_zero hm).symm
+    simp only [hp]
+    rw [add_mul, one_mul, choose_mul_add _ hn]
+
+
```
YaelDillies, line 47:
```suggestion
```

Mathlib/Data/Nat/Choose/Mul.lean
```diff
@@ -0,0 +1,48 @@
...
+/-! # Two lemmas on choose
+
+The proofs of these lemmas use the `ring` tactic
+and can't be given in `Mathlib.Data.Nat.Choose.Basic`
+
+-/
+
+namespace Nat
+
+theorem choose_mul_add (m) {n : ℕ} (hn : n ≠ 0) :
```
YaelDillies, line 19: I would suggest putting `hn` first since `m` will usually be inferrable from context

Mathlib/Data/Nat/Choose/Mul.lean
```diff
@@ -0,0 +1,48 @@
...
+    _ = ((m * (p + 1) + p) + 1) * (m * (p + 1) + p)! := by rw [factorial_succ]
+    _ = (m * (p + 1) + p)! * ((p + 1) * (m + 1)) := by ring
+    _ = ((m * (p + 1) + p).choose p * (m * (p+1))! * (p)!) * ((p + 1) * (m + 1)) := by
+      rw [add_choose_mul_factorial_mul_factorial]
+    _ = (m * (p + 1) + p).choose p * (m * (p+1))! * (((p + 1) * (p)!) * (m + 1)) := by ring
+    _ = (m * (p + 1) + p).choose p * (m * (p+1))! * ((p + 1)! * (m + 1)) := by rw [factorial_succ]
+    _ = (m + 1) * (m * (p + 1) + p).choose p * ((m * (p + 1))! * (p + 1)!) := by ring
+
+theorem choose_mul_right (m) {n : ℕ} (hn : n ≠ 0) :
+    (m * n).choose n = m * (m * n - 1).choose (n - 1) := by
```
YaelDillies, line 39: Same here

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,184 @@
...
+theorem uniformBell_one_left (n : ℕ) : uniformBell 1 n = 1 := by
+  simp only [uniformBell_eq, Finset.range_one, Finset.prod_singleton, zero_mul,
+    zero_add, choose_self]
+
+theorem uniformBell_one_right (m : ℕ) : uniformBell m 1 = 1 := by
+  simp only [uniformBell_eq, mul_one, add_tsub_cancel_right, ge_iff_le, le_refl,
+    tsub_eq_zero_of_le, choose_zero_right, Finset.prod_const_one]
+
+theorem uniformBell_mul_eq (m : ℕ) {n : ℕ} (hn : n ≠ 0) :
+    uniformBell m n * n.factorial ^ m * m.factorial = (m * n).factorial := by
```
YaelDillies, line 166: Can you use the factorial notation?

Mathlib/Combinatorics/Enumerative/Bell.lean
```diff
@@ -0,0 +1,184 @@
...
+      rw [← Nat.choose_mul_add c hx]
+      rw [mul_comm c x]
+      rw [Nat.add_choose_mul_factorial_mul_factorial (x * c) x]
+      rw [mul_add, mul_one]
+    rw [factorial_succ, pow_succ]
+    ring_nf
+
+theorem bell_mul_eq (m : Multiset ℕ) :
+    m.bell * (m.map (fun j ↦ j !)).prod *
+    ∏ j ∈ (m.toFinset.erase 0), (m.count j)! = m.sum ! := by
```
YaelDillies, line 81:
```suggestion
    m.bell * (m.map (fun j ↦ j !)).prod * ∏ j ∈ (m.toFinset.erase 0), (m.count j)!
      = m.sum ! := by
```
or something like that?

Mathlib/Data/Finset/Basic.lean
AntoineChambert-Loir: I tried, this doesn't work.

Mathlib/Data/Nat/Choose/Mul.lean
AntoineChambert-Loir: I made the `m` argument implicit.

Mathlib/Data/Nat/Choose/Mul.lean
AntoineChambert-Loir: I made the `m` argument implicit. That works, and I preferred having the arguments in the same order as they appear in the formula.

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: No I can't, i need the induction hypothesis.

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: done

Mathlib/Combinatorics/Enumerative/Bell.lean
AntoineChambert-Loir: done

Mathlib/Combinatorics/Enumerative/Bell.lean
YaelDillies: You can 😉 The induction is simply called `bell_mul_eq_lemma` (it's a recursive call)

Mathlib/Data/Nat/Choose/Mul.lean
YaelDillies: I think it's "suboptimal but okay"

YaelDillies (approved):
maintainer merge

apart from [that](https://github.com/leanprover-community/mathlib4/pull/15644/files#r1831728215). Thanks!

kim-em:
bors d+

kim-em:
Please merge after addressing @YaelDillies's final comment above.

PR #15645, opened 2024-08-09
[Merged by Bors] - feat: Multiset lemmas

haitian-yuki:
---

These are used to prove the Dershowitz-Manna ordering theorem.

### DIFF
--- a/Mathlib/Data/Multiset/Basic.lean
+++ b/Mathlib/Data/Multiset/Basic.lean
@@ -996,6 +996,19 @@ theorem card_erase_eq_ite {a : α} {s : Multiset α} :
   · rwa [card_erase_of_mem h, if_pos]
   · rwa [erase_of_not_mem h, if_neg]
 
+-- 1 of 5
+@[simp]
+theorem mul_singleton_erase (a : α) (s : Multiset α) : s - {a} = s.erase a := by
+  ext
+  simp [Multiset.erase_singleton, Multiset.count_singleton]
+  split <;> simp_all
+
+-- 2 of 5
+theorem in_of_ne_of_cons_erase {a a0: α} {M X : Multiset α}
+    (H : M = (a0 ::ₘ X).erase a) (hyp : ¬ a = a0) : a0 ∈ M := by
+  rw [H, Multiset.mem_erase_of_ne fun h ↦ hyp <| id <| Eq.symm h, Multiset.mem_cons]
+  tauto
+
 end Erase
 
 @[simp]
@@ -1470,6 +1483,45 @@ theorem sub_eq_fold_erase (s t : Multiset α) : s - t = foldl erase erase_comm s
 theorem card_sub {s t : Multiset α} (h : t ≤ s) : card (s - t) = card s - card t :=
   Nat.eq_sub_of_add_eq $ by rw [← card_add, tsub_add_cancel_of_le h]
 
+-- 3 of 5
+theorem singleton_add_sub_of_cons_add [DecidableEq α] {a a0: α} {M X : Multiset α}
+    (H : a ::ₘ M = X + {a0}) : M = X + {a0} - {a} := by
+  by_cases hyp : a = a0
+  · rw [hyp, add_comm] at H
+    simp_all [Multiset.singleton_add]
+  · have a0_a: a0 ≠ a := by rw [eq_comm] at hyp; exact hyp
+    ext b
+    simp [Multiset.count_cons, Multiset.count_singleton, Multiset.count_add]
+    have H : Multiset.count b (a ::ₘ M) = Multiset.count b (X + {a0}) := by simp_all only
+    [Multiset.count_add]
+    by_cases ba : b = a
+    · rw [ba] at *
+      have : (a ::ₘ M).count a = M.count a + 1 := by simp
+      simp_all
+    by_cases ba0 : b = a0
+    · have : (a ::ₘ M).count a0 = X.count a0 + 1 := by
+        subst_eqs
+        rw [add_comm, Multiset.singleton_add] at H
+        simp_all
+      have : M.count a0 = Multiset.count a0 (a ::ₘ M) := by
+        have : a0 ≠ a := by simp_all
+        rw [Multiset.count_cons_of_ne this M]
+      simp_all
+    · have : M.count b = (a ::ₘ M).count b := by
+        have : b ≠ a := by simp_all
+        rw [Multiset.count_cons_of_ne this M]
+      rw [this]
+      simp_all
+
+-- 5 of 5
+theorem mem_sub_of_not_mem_of_mem {α} [DecidableEq α] (x : α) (X Y: Multiset α) (x_in_X : x ∈ X)
+    (x_notin_Y : x ∉ Y) : x ∈ X - Y  := by
+  have : Multiset.count x X ≥ 1 := by
+    rw [← Multiset.one_le_count_iff_mem] at x_in_X
+    exact x_in_X
+  rw [← Multiset.one_le_count_iff_mem, Multiset.count_sub]
+  simp_all only [not_false_eq_true, Multiset.count_eq_zero_of_not_mem, tsub_zero]
+
 /-! ### Union -/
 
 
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2333,70 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
...
+  have h0 : M.count x + N.count x = P.count x + Q.count x := by
+    rw [Multiset.ext] at h
+    simp_all only [Multiset.mem_add, Multiset.count_add]
+  by_cases l_u : M.count x ≤ P.count x
+  · have : N.count x ≥ Q.count x := by sorry -- linarith?
+    simp_all only [ge_iff_le, min_eq_right]
+    -- have := tsub_add_cancel_of_le l_u
+    sorry --linarith
+  · simp_all only [not_le, gt_iff_lt]
+    have : Multiset.count x N ≤ Multiset.count x Q := by sorry --linarith?
```
YaelDillies, line 2396: You can use `le_or_le_of_add_le_add h0.ge` here

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -996,6 +996,11 @@ theorem card_erase_eq_ite {a : α} {s : Multiset α} :
   · rwa [card_erase_of_mem h, if_pos]
   · rwa [erase_of_not_mem h, if_neg]
 
+theorem in_of_ne_of_cons_erase {a a0: α} {M X : Multiset α}
+    (H : M = (a0 ::ₘ X).erase a) (hyp : ¬ a = a0) : a0 ∈ M := by
+  rw [H, Multiset.mem_erase_of_ne fun h ↦ hyp <| id <| Eq.symm h, Multiset.mem_cons]
+  tauto
```
YaelDillies, line 1002: I don't think this lemma is worth having:
1. We almost never have lemmas taking an equality as an assumption. The reasons are that
  a. The user is (almost) always able to substitute that equality before trying to apply any lemma
  b. Adding all possible such lemmas would lead to combinatorial explosion
2. Once you substitute `M` in, the proof is structurally made of existing lemmas, namely `(mem_erase_of_ne hyp.symm).2 (mem_cons_self ..)`

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2332,57 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
   simp only [count_eq_card_filter_eq, filter_map, card_map, Function.id_comp,
     eq_true_eq_id, Function.comp_apply]
 
+theorem sub_singleton [DecidableEq α] (a : α) (s : Multiset α) : s - {a} = s.erase a := by
```
YaelDillies, line 2335:
```suggestion
@[simp] theorem sub_singleton [DecidableEq α] (a : α) (s : Multiset α) : s - {a} = s.erase a := by
```

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2332,57 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
...
+        rw [Multiset.count_cons_of_ne this M]
+      simp_all
+    · have : M.count b = (a ::ₘ M).count b := by
+        have : b ≠ a := by simp_all
+        rw [Multiset.count_cons_of_ne this M]
+      rw [this]
+      simp_all
+
+theorem mem_sub_of_not_mem_of_mem {α} [DecidableEq α] (x : α) (X Y: Multiset α) (x_in_X : x ∈ X)
+    (x_notin_Y : x ∉ Y) : x ∈ X - Y  := by
```
YaelDillies, line 2370:
```suggestion
theorem mem_sub [DecidableEq α] : a ∈ s - t ↔ t.count a < s.count a := by
  rw [← count_pos, count_sub, Nat.sub_pos_iff_lt]
```
then your original lemma can be obtained using `count_pos` again and `count_eq_zero_of_not_mem`

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2332,57 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
...
+  have : Multiset.count x X ≥ 1 := by
+    rw [← Multiset.one_le_count_iff_mem] at x_in_X
+    exact x_in_X
+  rw [← Multiset.one_le_count_iff_mem, Multiset.count_sub]
+  simp_all
+
+theorem inter_add_sub_of_eq {α} [dec : DecidableEq α] {M N P Q: Multiset α} (h : M + N = P + Q) :
```
YaelDillies, line 2377:
```suggestion
theorem inter_add_sub_of_eq [DecidableEq α] {M N P Q : Multiset α} (h : M + N = P + Q) :
```

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2332,57 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
...
+    N = N ∩ Q + (P - M) := by
```
YaelDillies, line 2378: Can you put a few more brackets so that it's clear whether it's supposed to be read as
```suggestion
    N = N ∩ (Q + (P - M)) := by
```
or
```suggestion
    N = (N ∩ Q) + (P - M) := by
```

YaelDillies (commented):
Please mark this PR as ready for review (instead of draft) when it is

Mathlib/Data/Multiset/Basic.lean
haitian-yuki: Indeed, (also as a side note to myself ;-)) the original lemma can be proved as follows:

`
theorem mem_sub_of_not_mem_of_mem {α} [DecidableEq α] (x : α) (X Y: Multiset α) (x_in_X : x ∈ X)
    (x_notin_Y : x ∉ Y) : x ∈ X - Y := by
  rwa [mem_sub, count_eq_zero_of_not_mem, count_pos]
  assumption
`

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2327,53 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
...
+        simp_all
+      have : M.count a0 = Multiset.count a0 (a ::ₘ M) := by
+        have : a0 ≠ a := by simp_all
+        rw [Multiset.count_cons_of_ne this M]
+      simp_all
+    · have : M.count b = (a ::ₘ M).count b := by
+        have : b ≠ a := by simp_all
+        rw [Multiset.count_cons_of_ne this M]
+      rw [this]
+      simp_all
```
YaelDillies, line 2362: Given that this works, I think we don't need this lemma
```suggestion
  rw [← H, sub_singleton, erase_cons_head]
```

YaelDillies:
The rest looks good

maintainer merge

Mathlib/Data/Multiset/Basic.lean
```diff
@@ -2327,6 +2327,24 @@ theorem map_count_True_eq_filter_card (s : Multiset α) (p : α → Prop) [Decid
   simp only [count_eq_card_filter_eq, filter_map, card_map, Function.id_comp,
     eq_true_eq_id, Function.comp_apply]
 
+@[simp] theorem sub_singleton [DecidableEq α] (a : α) (s : Multiset α) : s - {a} = s.erase a := by
+  ext
+  simp [Multiset.erase_singleton, Multiset.count_singleton]
```
urkud, line 2332: `simp only ...`?

YaelDillies: I think `split` should be considered a fancy tactic here

m4lvin: What does "fancy" mean here? That we do not need to change this to `simp only`?

YaelDillies: Yes, basically. See eg https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Terminal.20tidy/near/293210921

YaelDillies (commented):
maintainer merge

Mathlib/Data/Multiset/Basic.lean
semorrison: I disagree here, I've seen plenty on nonterminal simps before a split break.

m4lvin: I pushed a change for this.

b-mehta:
Thanks!

bors merge

PR #15646, opened 2024-08-09
[Merged by Bors] - chore: avoid importing `Cardinal` in basic group theory

kim-em:


Mathlib/GroupTheory/Commutator/Basic.lean
```diff
@@ -229,3 +209,6 @@ theorem mem_commutatorSet_iff : g ∈ commutatorSet G ↔ ∃ g₁ g₂ : G, ⁅
 
 theorem commutator_mem_commutatorSet : ⁅g₁, g₂⁆ ∈ commutatorSet G :=
   ⟨g₁, g₂, rfl⟩
+
+assert_not_exists Cardinal
+assert_not_exists Multiset
```
Ruben-VandeVelde, line 214: Can you please put those at the top?

Mathlib/GroupTheory/Coset/Basic.lean
```diff
@@ -722,3 +686,5 @@ noncomputable def preimageMkEquivSubgroupProdSet (s : Subgroup α) (t : Set (α
   right_inv := fun ⟨⟨a, ha⟩, ⟨x, hx⟩⟩ => by ext <;> simp [ha]
 
 end QuotientGroup
+
+assert_not_exists Cardinal
```
Ruben-VandeVelde, line 690: Ditto

Mathlib/Topology/Algebra/Ring/Basic.lean
```diff
@@ -354,3 +354,5 @@ def AbsoluteValue.comp {R S T : Type*} [Semiring T] [Semiring R] [OrderedSemirin
   add_le' _ _ := (congr_arg v (map_add f _ _)).trans_le (v.add_le _ _)
 
 end AbsoluteValue
+
+assert_not_exists Cardinal
```
Ruben-VandeVelde, line 358: Ditto

Ruben-VandeVelde (approved):
lgtm, though I'd prefer you keep the import statements sorted

semorrison:
> lgtm, though I'd prefer you keep the import statements sorted

I think this is a problem with `shake`, actually. When splitting, I try to leave imports "as is" (i.e. I replace the split file with the "last" of its split children), and then just let `shake` deal with minimising imports.

My preference is not to add the busy work for humans to resort imports, and either automate resorting imports periodically, or change `shake`.

Mathlib/GroupTheory/Commutator/Basic.lean
semorrison: We're actually already split in Mathlib as to whether these go at the top or bottom. I'm unsure myself which is better. At the top means they are closer to the import statements, at the bottom they are less clutter in the way of a normal reader who just wants to start reading declarations.

semorrison: Looking for consensus at https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/.60assert_not_exists.60.20at.20top.20or.20bottom.20of.20file.3F/near/459960523. I'll mark these as resolved in the meantime and we can cleanup later as required.

tb65536:
maintainer merge

mattrobball:
Yes. 

bors merge

PR #12648, opened 2024-05-04
feat(Computability): regular expression nfa equivalence

TpmKranz:
Proofs that (formal language) regular expressions and NFA's describe the same class of languages

- every NFA has a regular expression describing the same language, implemented via state ripping in generalized nondeterministic finite automata
- every regular expression has an NFA describing the same language via Thompson's construction, adapted to not require ε-transitions

Very much based on https://github.com/leanprover-community/mathlib/pull/14250, thus:
Co-authored-by: Russell Emerine <rre0520@gmail.com>

---

- [ ] depends on: #15647
- [ ] depends on: #15649
- [ ] depends on: #15651
- [ ] depends on: #15654

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15651, opened 2024-08-09
feat(Computability/NFA): operations for Thompson's construction

TpmKranz:
Lays the groundwork for a proof of equivalence of RE and NFA, w.r.t. described language. Actual connection to REs comes later, after the groundwork for the opposite direction has been laid.

Third chunk of #12648

---

### DIFF
--- a/Mathlib/Computability/NFA.lean
+++ b/Mathlib/Computability/NFA.lean
@@ -21,7 +21,7 @@ open Set
 
 open Computability
 
-universe u v
+universe u v w
 
 
 /-- An NFA is a set of states (`σ`), a transition function from state to state labelled by the
@@ -33,7 +33,7 @@ structure NFA (α : Type u) (σ : Type v) where
   start : Set σ
   accept : Set σ
 
-variable {α : Type u} {σ σ' : Type v} (M : NFA α σ)
+variable {α : Type u} {σ : Type v} {σ' : Type w} (M : NFA α σ)
 
 namespace NFA
 
@@ -67,6 +67,18 @@ theorem evalFrom_singleton (S : Set σ) (a : α) : M.evalFrom S [a] = M.stepSet 
 theorem evalFrom_append_singleton (S : Set σ) (x : List α) (a : α) :
     M.evalFrom S (x ++ [a]) = M.stepSet (M.evalFrom S x) a := by
   simp only [evalFrom, List.foldl_append, List.foldl_cons, List.foldl_nil]
+
+/-- Evaluations can be staggered. -/
+theorem evalFrom_append (S : Set σ) (x y : List α) :
+    M.evalFrom S (x ++ y) = M.evalFrom (M.evalFrom S x) y := by
+  simp only [evalFrom, List.foldl_append]
+
+/-- Evaluations are monotone. -/
+theorem evalFrom_subset {σ : Type v} (M : NFA α σ) (x : List α) {qs ps : Set σ}
+    (subs : qs ⊆ ps) : M.evalFrom qs x ⊆ M.evalFrom ps x := x.reverseRecOn subs <| by
+  simp only [evalFrom_append_singleton, mem_stepSet, subset_def]
+  rintro _ _ ih _ ⟨t, eval, step⟩
+  exact ⟨t, ih t eval, step⟩
 
 /-- `M.eval x` computes all possible paths though `M` with input `x` starting at an element of
   `M.start`. -/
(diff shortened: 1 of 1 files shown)


Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +124,296 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+
+Since NFAs are models for regular languages, they provide decision procedures for the empty set,
+singleton sets with members of at most length 1 as well as a notion of closedness under union,
+concatenation and application of the Kleene star.
+
+Cf. <https://en.wikipedia.org/wiki/Thompson%27s_construction> for the basic idea in a setting
+permitting ε-moves.
+-/
+instance : Zero (NFA α σ) :=
+  ⟨default⟩
```
dupuisf, line 137: I think it's a bit strange to have this defined in terms of `default`, which is meant to be an arbitrary element of the type with no special meaning. I think it would be best to move this instance to the top of the file and then define the `Inhabited` instance as `⟨0⟩` and replace uses of `default` further down by `0`.

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +124,296 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+    rcases x.eq_nil_or_concat with rfl | ⟨_, _, rfl⟩
+    · exact rfl
+    rw [List.concat_eq_append, eval_append_singleton, mem_stepSet] at eval
+    rcases eval with ⟨_, _, ⟨⟩⟩
+  · rintro rfl
+    exact ⟨default, Set.mem_univ _, Set.mem_univ _⟩
+
+/-- An NFA accepting just a single word of length 1, given by the underlying character. -/
+def char (a : α) : (NFA α (σ ⊕ σ')) :=
+  ⟨fun p c q ↦ p.isLeft ∧ a = c ∧ q.isRight, (Sum.isLeft ·), (Sum.isRight ·)⟩
```
dupuisf, line 163:
```suggestion
/-- An NFA accepting just a single word of length 1, given by the underlying character. -/
def char (a : α) : NFA α (σ ⊕ σ') :=
  ⟨fun p c q ↦ p.isLeft ∧ a = c ∧ q.isRight, (Sum.isLeft ·), (Sum.isRight ·)⟩
```

dupuisf: Also, I don't really understand why we need `σ` and `σ'` around here. Why not just `Fin 2` as the set of states?

Mathlib/Computability/NFA.lean
TpmKranz: Thank you for pointing that out! I didn't want to upend the sectioning of the file, so I instead moved the `Inhabited` instance down so it can use `⟨0⟩`. The uses of `default` further down refer to the states, which can, inherently, not depend on implementation details of `σ/σ'`.

A silly thought that occured to me to really drive the point of not relying on the particulars of `default` home was to just use `1`, which doesn't need `σ` to be `Inhabited` to be defined but has differing semantics depending on whether or not `σ` actually is.

Mathlib/Computability/NFA.lean
TpmKranz: It is a bit awkward, to be sure. I wanted to stay as generic as possible and what really matters about the `char` automaton's state type is that it has at least two distinct elements. Unfortunately, there doesn't seem to be a data-carrying pendant to `Nontrivial`, like `Inhabited` is to `Nonempty`. There is `TwoPointed`, but that's not a class…

`Fin 2` (or `Bool` or whatever) would lock the state space in place.
This way, one can't just conjure up a three-stated `char` automaton and re-use the `char_correct` proof (if that is ever on someone's mind).
It also raises the question of why the `One` instance doesn't just work with `Fin 1`/`Unit` or `Zero` with `Fin 0`/`Empty`.

But in the end, I'm really curious to find out what “the mathlib approach”™ would be. I see four different ways to proceed here:
1. make `TwoPointed` a class (adding another PR?)
2. use an `Inhabited σ` and let `char` work with `Option σ`
3. keep the sum
4. use a specific state type (and have a think about `Zero`/`One`?)

I would be glad to receive some input and/or additions!

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+  ⟨fun p c q ↦
+    match (p, q) with
+    | (some p, some q) => P.step p c q ∨ ∃ r, P.accept r ∧ P.step p c r ∧ P.start q
+    | _ => False,
+    Option.rec True P.start,
+    Option.rec True P.accept⟩
+
+/-- All start states of an unstarred NFA can be reached in a starred NFA by a string in the
+original's language. -/
+lemma kstar_eval_aux {P : NFA α σ} {y} (h : y ∈ P.accepts) :
```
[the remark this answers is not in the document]
TpmKranz: Thanks, I didn't know about the `private` keyword and thought `lemma` would be enough to signify the intention of being only immediately useful. I shall add the keyword, then. The other `lemma`ta might come in useful in other places though, since they describe interactions with `eval` somewhat more completely. Or would you prefer them being made private, as well?

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+
+instance : HAdd (NFA α σ) (NFA α σ') (NFA α (σ ⊕ σ')) :=
+  ⟨fun P₁ P₂ ↦
+    ⟨fun p c q ↦
+      match (p, q) with
+      | (.inl p, .inl q) => P₁.step p c q
+      | (.inr p, .inr q) => P₂.step p c q
+      | _ => False,
+      Sum.elim P₁.start P₂.start,
+      Sum.elim P₁.accept P₂.accept⟩⟩
```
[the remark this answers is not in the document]
TpmKranz: I've leaned on the notation for Kleene algebras, where + is the union and * is concatenation. `hadd_correct` and `hmul_correct` illustrate this. Or would you like me to avoid notation in general? I could certainly make `union` and `concat` (or whatever) functions out of these definitions.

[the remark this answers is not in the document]
YaelDillies: Heterogeneous addition is so far not used in mathlib, so I am a bit wary of introducing a use of it. I don't want to be dogmatic, so my personal opinion is that removing the notation might be the best thing to do. I would suggest calling the definition `sum` to match the `⊕` that appears in the return type.

Mathlib/Computability/NFA.lean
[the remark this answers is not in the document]
YaelDillies: I think the other ones shouldn't be private

meithecatte:
Relevant Zulip discussion: [#mathlib4 > Regular languages: the review queue](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Regular.20languages.3A.20the.20review.20queue/with/514669532)

leanprover-community-bot-assistant:

rudynicolop:
I have conflicting PRs #31610 (for kleene closure) and  #30872 (concatenation closure), that this PR subsumes. If this PR is preferred and merged, I will close my #31610 and #30872, otherwise I am happy to work with the maintainers and @TpmKranz, and credit @TpmKranz in my PRs.

Mathlib/Computability/NFA.lean
```diff
@@ -33,13 +33,10 @@ structure NFA (α : Type u) (σ : Type v) where
   start : Set σ
   accept : Set σ
 
-variable {α : Type u} {σ σ' : Type v} (M : NFA α σ)
+variable {α : Type u} {σ : Type v} {σ' : Type w} (M : NFA α σ)
```
YaelDillies, line 36: Can we please just avoid naming the universe variables?
```suggestion
variable {α σ σ' : Type*} (M : NFA α σ)
```

Mathlib/Computability/NFA.lean
```diff
@@ -68,6 +65,18 @@ theorem evalFrom_append_singleton (S : Set σ) (x : List α) (a : α) :
     M.evalFrom S (x ++ [a]) = M.stepSet (M.evalFrom S x) a := by
   simp only [evalFrom, List.foldl_append, List.foldl_cons, List.foldl_nil]
 
+/-- Evaluations can be staggered. -/
```
YaelDillies, line 68: I find this docstring more confusing than it helps. Maybe just drop it?

Mathlib/Computability/NFA.lean
```diff
@@ -68,6 +65,18 @@ theorem evalFrom_append_singleton (S : Set σ) (x : List α) (a : α) :
...
+theorem evalFrom_append (S : Set σ) (x y : List α) :
```
YaelDillies, line 69: Should this be simp?

Mathlib/Computability/NFA.lean
```diff
@@ -68,6 +65,18 @@ theorem evalFrom_append_singleton (S : Set σ) (x : List α) (a : α) :
...
+    M.evalFrom S (x ++ y) = M.evalFrom (M.evalFrom S x) y := by
+  simp only [evalFrom, List.foldl_append]
+
+/-- Evaluations are monotone. -/
+theorem evalFrom_subset {σ : Type v} (M : NFA α σ) (x : List α) {qs ps : Set σ}
```
YaelDillies, line 74:
```suggestion
@[gcongr]
theorem evalFrom_subset {σ : Type v} (M : NFA α σ) (x : List α) {qs ps : Set σ}
```

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
 
+/-! ### Regex-like operations
+
+Since NFAs are models for regular languages, they provide decision procedures for the empty set,
+singleton sets with members of at most length 1 as well as a notion of closedness under union,
+concatenation and application of the Kleene star.
+
+Cf. <https://en.wikipedia.org/wiki/Thompson%27s_construction> for the basic idea in a setting
+permitting ε-moves.
+-/
```
YaelDillies, line 132:
```suggestion
-/

```

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
   rw [← toDFA_correct] at hx ⊢
   exact M.toDFA.pumping_lemma hx hlen
 
+/-! ### Regex-like operations
+
+Since NFAs are models for regular languages, they provide decision procedures for the empty set,
+singleton sets with members of at most length 1 as well as a notion of closedness under union,
```
YaelDillies, line 127:
```suggestion
singleton sets with members of length at most 1 as well as a notion of closure under union,
```

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+concatenation and application of the Kleene star.
+
+Cf. <https://en.wikipedia.org/wiki/Thompson%27s_construction> for the basic idea in a setting
+permitting ε-moves.
+-/
+instance : Zero (NFA α σ) :=
+  ⟨NFA.mk (fun _ _ ↦ ∅) ∅ ∅⟩
```
YaelDillies, line 134:
```suggestion
  ⟨.mk (fun _ _ ↦ ∅) ∅ ∅⟩
```

Mathlib/Computability/NFA.lean
```diff
@@ -112,6 +121,299 @@ theorem pumping_lemma [Fintype σ] {x : List α} (hx : x ∈ M.accepts)
...
+
+instance : Inhabited (NFA α σ) :=
+  ⟨0⟩
+
+@[simp]
+theorem zero_correct : (0 : NFA α σ).accepts = 0 := by
```
YaelDillies, line 140:
```suggestion
theorem accepts_zero : (0 : NFA α σ).accepts = 0 := by
```
Same below

YaelDillies:
The current PR will need to be reopened from a fork. Given the lack of reply from @TpmKranz, I believe it would be okay for you to be the one reopening the PR.

rudynicolop:
> The current PR will need to be reopened from a fork. Given the lack of reply from @TpmKranz, I believe it would be okay for you to be the one reopening the PR.

(hopefully I'm not being presumptuous), does it make sense for me to re-open this PR, or to try to get my other regular language PRs merged (#31610, #31247, #30872)?

Thanks!

YaelDillies:
Ah, if your own PRs replace this one, then there's nothing more to be done for now!

joneugster:

PR #15652, opened 2024-08-09
[Merged by Bors] - feat(Data/Set/Finite): added lemmas about cardinality of finite sets

pimotte:
Added some lemmas in preparation for Tuttes theorem

---

[Zulip thread on Tuttes](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Tutte's.20theorem)

I put this in a new file, because Data.Set.Card is already fairly long and this seems somewhat coherent, but I'm also open to adding it there if that makes more sense:)


Mathlib/Data/Set/Finite/Parity.lean
```diff
@@ -0,0 +1,58 @@
...
+    intro h
+    rw [@Set.nonempty_iff_ne_empty] at hn
+    exact hn <| (Set.ncard_eq_zero hs).mp h
+  have : s.ncard ≠ 1 := by
+    intro h
+    rw [Set.Nat.card_coe_set_eq, h] at he
+    simp at he
+  omega
+
+lemma two_of_even_of_nonempty (hs : Set.Finite s) (hn : Set.Nonempty s) (he : Even (Nat.card s)) :
```
AntoineChambert-Loir, line 31: This lemma is very specific and doesn't seem much easier to apply that the one with `two_le`.

Mathlib/Data/Set/Finite/Parity.lean
```diff
@@ -0,0 +1,58 @@
...
+
+This file contains lemmas concerning finite sets and parity of their cardinality.
+-/
+
+namespace Set.Finite
+
+variable {V : Type _} {s : Set V}
+
+lemma one_lt_ncard_of_nonempty_of_even (hs : Set.Finite s) (hn : Set.Nonempty s)
+    (he : Even (Nat.card s)) : 1 < s.ncard := by
```
Ruben-VandeVelde, line 20: Why are you using `Nat.card` on sets?

pimotte: Because I standardized on it in my staging area, I guess using `ncard` makes more sense here, changed to that:)

Mathlib/Data/Set/Finite/Parity.lean
pimotte: Fair enough, removed it:)

Mathlib/Data/Set/Finite/Parity.lean
```diff
@@ -0,0 +1,47 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Pim Otte
+-/
+import Mathlib.Data.Set.Card
+import Mathlib.Data.Set.Finite
+
+/-!
+# Finite sets and parity of their cardinality
+
+This file contains lemmas concerning finite sets and parity of their cardinality.
```
vihdzp, line 12:
```suggestion
This file contains lemmas concerning finite sets and the parity of their cardinality.
```
Perhaps this could go a bit more in depth about the scope of the file? Or, if these are the two only theorems you plan on adding, maybe they can be moved elsewhere? Perhaps to the same file of the rest of your proof.

Mathlib/Data/Set/Finite/Parity.lean
```diff
@@ -0,0 +1,47 @@
...
+  have : s.ncard ≠ 0 := by
+    intro h
+    rw [@Set.nonempty_iff_ne_empty] at hn
+    exact hn <| (Set.ncard_eq_zero hs).mp h
+  have : s.ncard ≠ 1 := by
+    intro h
+    simp [h] at he
+  omega
+
+lemma even_card_diff_pair [DecidableEq V] {x y : V} {u : Set V} (hu : Set.Finite u)
```
vihdzp, line 30: My first instinct seeing this theorem is that it might be better stated in the other direction. That is, `Even u → Even (insert x (insert y u))`. Of course, this is pretty trivial as the latter set has cardinality `2` more than the former. I think the usefulness of this is contingent on whichever proof you want to use it on, and even then, I'd ask if the `insert` form can't be used instead.

AntoineChambert-Loir: Given the arguments, I'd suggest two lemmas, relating Even s or Odd s to the opposite for ‘insert x s‘.
And maybe an additional that proves that y does not belong to insert x s iff...

Mathlib/Data/Set/Finite/Parity.lean
pimotte: The vast majority of the rest of the proof would be in Combinatorics.SimpleGraph, so if you think a separate file is overkill, I'd say Data.Set.Card is the most likely alternative.

Mathlib/Data/Set/Finite/Parity.lean
pimotte: Fixed this:)

Mathlib/Data/Set/Finite/Parity.lean
pimotte: In refactoring the code became even smaller, so I moved it to Data.Set.Card

Mathlib/Data/Set/Card.lean
```diff
@@ -614,6 +614,16 @@ theorem ncard_exchange' {a b : α} (ha : a ∉ s) (hb : b ∈ s) :
...
 
+lemma odd_card_insert_iff {a : α} (hs : s.Finite := by toFinite_tac) (ha : a ∉ s) :
+    Odd (insert a s).ncard ↔ Even s.ncard := by
+  rw [ncard_insert_of_not_mem ha hs, Nat.odd_add]
+  simp only [Nat.odd_add, Nat.odd_iff_not_even, Nat.not_even_one, iff_false, Decidable.not_not]
+
+lemma even_card_insert_iff {a : α} (hs : s.Finite := by toFinite_tac) (ha : a ∉ s) :
+    Even (insert a s).ncard ↔ Odd s.ncard := by
+  rw [ncard_insert_of_not_mem ha hs, Nat.even_add]
+  simp only [Nat.not_even_one, iff_false, Nat.odd_iff_not_even]
```
Ruben-VandeVelde, line 625:
```suggestion
  rw [ncard_insert_of_not_mem ha hs, Nat.even_add_one, Nat.not_even_iff_odd]
```

Ruben-VandeVelde (commented):
Please merge master

pimotte:
> Please merge master

Done:)

Mathlib/Data/Set/Card.lean
```diff
@@ -991,6 +1000,15 @@ theorem one_lt_ncard_iff (hs : s.Finite := by toFinite_tac) :
   rw [one_lt_ncard hs]
   simp only [exists_prop, exists_and_left]
 
+lemma one_lt_ncard_of_nonempty_of_even (hs : Set.Finite s) (hn : Set.Nonempty s := by toFinite_tac)
+    (he : Even (s.ncard)) : 1 < s.ncard := by
+  have : s.ncard ≠ 0 := by
+    intro h
+    rw [Set.nonempty_iff_ne_empty] at hn
+    exact hn <| (Set.ncard_eq_zero hs).mp h
```
Ruben-VandeVelde, line 1008:
```suggestion
  have : 0 < s.ncard := (Set.ncard_pos hs).mpr hn
```
or
```suggestion
  rw [← Set.ncard_pos hs] at hn
```

Ruben-VandeVelde (approved):
One more; LGTM otherwise

Mathlib/Data/Set/Card.lean
pimotte: Fixed:)

Ruben-VandeVelde (approved):
Thanks! 

maintainer merge

PR #15660, opened 2024-08-09
[Merged by Bors] - refactor: deprecate stdBasis

PatrickMassot:
Deprecate stdBasis in favor of its duplicate LinearMap.single. Also changes the binder types in the later.

Note the `simp` normal form of an applied (or even simply coerced to function) `LinearMap.single` is a `Pi.single`, as discussed during the community meeting.

See the [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/LinearMap.2EstdBasis/near/457515331).

---


Mathlib/LinearAlgebra/FinsuppVectorSpace.lean
```diff
@@ -154,9 +154,15 @@ theorem _root_.Finset.sum_single_ite [Fintype n] (a : R) (i : n) :
   simp only [apply_ite (Finsupp.single _), Finsupp.single_zero, Finset.sum_ite_eq,
     if_pos (Finset.mem_univ _)]
 
-theorem equivFun_symm_stdBasis [Finite n] (b : Basis n R M) (i : n) :
-    b.equivFun.symm (LinearMap.stdBasis R (fun _ => R) i 1) = b i := by
+theorem equivFun_symm_single [Finite n] (b : Basis n R M) (i : n) :
+    b.equivFun.symm (LinearMap.single R (fun _ => R) i 1) = b i := by
```
eric-wieser, line 158: I think I'd argue this should be
```suggestion
    b.equivFun.symm (Pi.single i 1) = b i := by
```
which would then make it safe to be `simp` too.

Mathlib/LinearAlgebra/Matrix/BilinearForm.lean
```diff
@@ -153,7 +152,7 @@ theorem BilinForm.toMatrix'_toBilin' (M : Matrix n n R₂) :
 
 @[simp]
 theorem BilinForm.toMatrix'_apply (B : BilinForm R₂ (n → R₂)) (i j : n) :
-    BilinForm.toMatrix' B i j = B (stdBasis R₂ (fun _ => R₂) i 1) (stdBasis R₂ (fun _ => R₂) j 1) :=
+    BilinForm.toMatrix' B i j = B (single R₂ (fun _ => R₂) i 1) (single R₂ (fun _ => R₂) j 1) :=
```
eric-wieser, line 155: Here too I think I'd argue for `Pi.single` on the RHS

Mathlib/LinearAlgebra/Matrix/DotProduct.lean
```diff
@@ -35,21 +35,30 @@ section Semiring
 
 variable [Semiring R] [Fintype n]
 
-@[simp]
+theorem dotProduct_single_eq_mul [DecidableEq n] (v : n → R) (c : R) (i : n) :
+    dotProduct v (LinearMap.single R (fun _ => R) i c) = v i * c := by
```
eric-wieser, line 39: And here

eric-wieser (commented):
What do you think of my suggestions below to replace all occurences of `LinearMap.single R (fun _ => R) i c` with `Pi.single i c`? That makes the spelling simp-normal, which would allow a fair number of lemmas to also be `@[simp]`.

eric-wieser:
> or even simply coerced to function

To me this choice is not obviously the right one (vs only making the applied version `simp`), but nor is it obviously wrong.

mattrobball:
> What do you think of my suggestions below to replace all occurences of `LinearMap.single R (fun _ => R) i c` with `Pi.single i c`? That makes the spelling simp-normal, which would allow a fair number of lemmas to also be `@[simp]`.
> 
> 

I don't really know what API boundaries you want to keep here.

Mathlib/LinearAlgebra/Pi.lean
```diff
@@ -113,25 +115,109 @@ theorem apply_single [AddCommMonoid M] [Module R M] [DecidableEq ι] (f : (i : 
...
+  classical
+    by_cases hiI : i ∈ I
+    · by_cases hiJ : i ∈ J
+      · exact (h.le_bot ⟨hiI, hiJ⟩).elim
+      · exact hJ i hiJ
+    · exact hI i hiI
+
+theorem single_eq_single {a : R} :
+    (fun i : ι => (single R (fun _ : ι => R) i) a) = fun i : ι => ↑(Finsupp.single i a) :=
+  funext fun i => (Finsupp.single_eq_pi_single i a).symm
```
eric-wieser, line 212: This lemma strikes me as very strange; I'd argue that any user of it should be using `Finsupp.single_eq_pi_single` directly.

eric-wieser:
> I don't really know what API boundaries you want to keep here.

I'm deliberately saying that there should be no API boundary between `Pi.single` and `LinearMap.single`; the rule is that the latter should only exist in lemmas where it is not applied, as is the case with most bundled maps with a corresponding unbundled spelling.

Mathlib/LinearAlgebra/Pi.lean
```diff
@@ -113,25 +114,105 @@ theorem apply_single [AddCommMonoid M] [Module R M] [DecidableEq ι] (f : (i : 
...
 
-variable (R φ)
+variable [DecidableEq ι]
+
+@[simp]
+theorem single_same (i : ι) (b : φ i) : single R φ i b i = b :=
+  Pi.single_eq_same i b
+
+theorem single_ne (i j : ι) (h : j ≠ i) (b : φ i) : single R φ i b j = 0 :=
+  Pi.single_eq_of_ne h b
```
eric-wieser, line 146: I'd propose we delete these and `single_apply'` and `single_apply` too, and poitn the deprecations directly at the `Pi.single` lemmas. I'll push a change to that effect once everything else is passing.

Mathlib/LinearAlgebra/StdBasis.lean
```diff
@@ -154,32 +146,32 @@ section Module
 
 variable {η : Type*} {ιs : η → Type*} {Ms : η → Type*}
 
-theorem linearIndependent_stdBasis [Ring R] [∀ i, AddCommGroup (Ms i)] [∀ i, Module R (Ms i)]
```
eric-wieser, line 157: It looks like the deprecation for this one is missing

eric-wieser (approved):
bors d+

Please check the changes I pushed look ok. Thanks for doing the bulk of the work here!

Mathlib/LinearAlgebra/Pi.lean
eric-wieser: (I did this)

PatrickMassot:
bors merge

PR #15663, opened 2024-08-09
[Merged by Bors] - chore: turn some `induction'` to `induction`

pitmonticone:
This PR turns some (soon to be deprecated?) `induction'` into `induction`.

---

### DIFF
--- a/Mathlib/Algebra/Polynomial/Eval.lean
+++ b/Mathlib/Algebra/Polynomial/Eval.lean
@@ -104,10 +104,9 @@ def eval₂AddMonoidHom : R[X] →+ S where
 
 @[simp]
 theorem eval₂_natCast (n : ℕ) : (n : R[X]).eval₂ f x = n := by
-  induction' n with n ih
-  -- Porting note: `Nat.zero_eq` is required.
-  · simp only [eval₂_zero, Nat.cast_zero, Nat.zero_eq]
-  · rw [n.cast_succ, eval₂_add, ih, eval₂_one, n.cast_succ]
+  induction n with -- Porting note: `Nat.zero_eq` is required.
+  | zero => simp only [eval₂_zero, Nat.cast_zero, Nat.zero_eq]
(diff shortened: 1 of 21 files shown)


Mathlib/Algebra/Polynomial/Eval.lean
```diff
@@ -104,10 +104,9 @@ def eval₂AddMonoidHom : R[X] →+ S where
 
 @[simp]
 theorem eval₂_natCast (n : ℕ) : (n : R[X]).eval₂ f x = n := by
-  induction' n with n ih
-  -- Porting note: `Nat.zero_eq` is required.
-  · simp only [eval₂_zero, Nat.cast_zero, Nat.zero_eq]
-  · rw [n.cast_succ, eval₂_add, ih, eval₂_one, n.cast_succ]
+  induction n with -- Porting note: `Nat.zero_eq` is required.
```
j-loreaux, line 107: Is this porting note still relevant? Either way, I think we can delete it.

j-loreaux (commented):
Thanks for doing this!

bors d+

Mathlib/Algebra/Polynomial/Eval.lean
```diff
@@ -104,10 +104,9 @@ def eval₂AddMonoidHom : R[X] →+ S where
...
+  | zero => simp only [eval₂_zero, Nat.cast_zero, Nat.zero_eq]
```
eric-wieser, line 108:
```suggestion
  | zero => simp only [eval₂_zero, Nat.cast_zero]
```
I think

pitmonticone: Thank you Eric. I'm going to test your suggested change as soon as I can.

pitmonticone: Tested. You are right. Thanks.

Mathlib/Algebra/Polynomial/Eval.lean
pitmonticone: Ok, done. Thanks.

pitmonticone:
bors merge

PR #15667, opened 2024-08-10
[Merged by Bors] - feat: linter for `admit` (use `sorry` instead)

adomasbaliuka:
The "admit" linter flags usages of the `admit` tactic.

The tactics `admit` and `sorry` are synonyms.
The use of `sorry` is much more common and should be preferred.

---
See also [zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/.E2.9C.94.20difference.20between.20sorry.20and.20admit/near/459519146)

### DIFF
--- a/Mathlib/Tactic/Linter/AdmitLinter.lean
+++ b/Mathlib/Tactic/Linter/AdmitLinter.lean
@@ -1,5 +1,5 @@
 /-
-Copyright (c) 2024 Damiano Testa. All rights reserved.
+Copyright (c) 2024 Adomas Baliuka. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Damiano Testa, Adomas Baliuka
 -/
@@ -23,7 +23,7 @@ namespace Mathlib.Linter.admit
 
 /-- The admit linter emits a warning on usages of `admit`. -/
 register_option linter.admit : Bool := {
-  defValue := true
+  defValue := false
   descr := "enable the admit linter"
 }
 
--- a/Mathlib/Tactic/Linter/AdmitLinter.lean
+++ b/Mathlib/Tactic/Linter/AdmitLinter.lean
@@ -27,17 +27,13 @@ register_option linter.admit : Bool := {
   descr := "enable the admit linter"
 }
 
-syntax (name := admit) "admit " : tactic
-/-- `admit` is a shorthand for `exact sorry`. -/
-macro "admit" : tactic => `(tactic| exact @sorryAx _ false)
-
-/-- `getadmit t` returns all usages of the `admit` tactic in the input syntax `t`. -/
+/-- `getAdmit t` returns all usages of the `admit` tactic in the input syntax `t`. -/
 partial
-def getadmit : Syntax → Array Syntax
-  | stx@(.node _ kind args) =>
-    let rargs := (args.map getadmit).flatten
-    if kind == ``admit then rargs.push stx else rargs
-  | _ => default
+def getAdmit (stx : Syntax) : Array Syntax :=
+  if let `(tactic| admit) := stx then
+    #[stx]
+  else
+    stx.foldArgs (fun arg r => r ++ getAdmit arg) #[]
 
 /-- The "admit" linter flags usages of the `admit` tactic.
 
@@ -47,15 +43,14 @@ The use of `sorry` is much more common and should be preferred.
 def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.admit o
 
 @[inherit_doc getLinterHash]
-def admitLinter : Linter where run := withSetOptionIn fun _stx => do
+def admitLinter : Linter where run := withSetOptionIn fun stx => do
   unless getLinterHash (← getOptions) do
     return
   if (← MonadState.get).messages.hasErrors then
     return
-  for stx in (getadmit _stx) do
-    Linter.logLint linter.admit stx
-      "The `admit` tactic is discouraged: \
-      please consider using the synonymous `sorry` instead."
+  for stxAdmit in (getAdmit stx) do
+    Linter.logLint linter.admit stxAdmit
+      "The `admit` tactic is discouraged: please consider using the synonymous `sorry` instead."
 
 initialize addLinter admitLinter
 
--- a/Mathlib/Tactic/Linter/AdmitLinter.lean
+++ b/Mathlib/Tactic/Linter/AdmitLinter.lean
@@ -1,0 +1,62 @@
+/-
+Copyright (c) 2024 Damiano Testa. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Damiano Testa, Adomas Baliuka
+-/
+import Lean.Elab.Command
+import Lean.Linter.Util
+
+/-!
+# The "admit" linter
+
+The "admit" linter flags usages of the `admit` tactic.
+
+The tactics `admit` and `sorry` are synonyms.
+The use of `sorry` is much more common and should be preferred.
+
+This linter is an incentive to discourage uses of `admit`, without being a ban.
+-/
+
+open Lean Elab
+
+namespace Mathlib.Linter.admit
+
+/-- The admit linter emits a warning on usages of `admit`. -/
+register_option linter.admit : Bool := {
+  defValue := true
+  descr := "enable the admit linter"
+}
+
+syntax (name := admit) "admit " : tactic
+/-- `admit` is a shorthand for `exact sorry`. -/
+macro "admit" : tactic => `(tactic| exact @sorryAx _ false)
+
+/-- `getadmit t` returns all usages of the `admit` tactic in the input syntax `t`. -/
+partial
+def getadmit : Syntax → Array Syntax
+  | stx@(.node _ kind args) =>
+    let rargs := (args.map getadmit).flatten
+    if kind == ``admit then rargs.push stx else rargs
+  | _ => default
+
+/-- The "admit" linter flags usages of the `admit` tactic.
+
+The tactics `admit` and `sorry` are synonyms.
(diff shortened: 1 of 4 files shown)


Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+
+/-- The "admit" linter flags usages of the `admit` tactic.
+
+The tactics `admit` and `sorry` are synonyms.
+The use of `sorry` is much more common and should be preferred.
+-/
+def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.admit o
+
+@[inherit_doc getLinterHash]
+def admitLinter : Linter where run := withSetOptionIn fun _stx => do
```
grunweg, line 50: Prefixing an actually used variable with an underscore looks slightly funny. What do you think about renaming `_stx` to `stx` instead, and using a different name below (or even just `stx` again, shadowing it - if that works)?
Also, use \maps, please.

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+  unless getLinterHash (← getOptions) do
+    return
+  if (← MonadState.get).messages.hasErrors then
+    return
+  for stx in (getadmit _stx) do
+    Linter.logLint linter.admit stx
+      "The `admit` tactic is discouraged: \
```
grunweg, line 57: Does this fit on one line? (The message looks good to me.)

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+
+This linter is an incentive to discourage uses of `admit`, without being a ban.
+-/
+
+open Lean Elab
+
+namespace Mathlib.Linter.admit
+
+/-- The admit linter emits a warning on usages of `admit`. -/
+register_option linter.admit : Bool := {
```
grunweg, line 25: For the other linters, this option is usually put outside of the `linterName` namespace. Did you change this for a reason?

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+  defValue := true
+  descr := "enable the admit linter"
+}
+
+syntax (name := admit) "admit " : tactic
```
grunweg, line 30: Can you use the definition of admit instead of rolling your own? (Or is that the issue you're referring to in the description?)

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
+/-
+Copyright (c) 2024 Damiano Testa. All rights reserved.
```
grunweg, line 2: Should you be the copyright holder? @adomani what do you think?

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+This linter is an incentive to discourage uses of `admit`, without being a ban.
+-/
+
+open Lean Elab
+
+namespace Mathlib.Linter.admit
+
+/-- The admit linter emits a warning on usages of `admit`. -/
+register_option linter.admit : Bool := {
+  defValue := true
```
grunweg, line 26: This setting enables the linter in *any* file which imports this (or a file importing this one). As you want ~all of mathlib to be linted, that implies projects *depending* on mathlib will also see this linter. The question to decide is whether this is what we want.
I tend towards "no": this is purely a decision about code style, which mathlib should not impose on other projects.
Thus, the ideal long-term fix is to *disable* this linter by default, but *enable* it in mathlib (and the archive and counterexamples).
That said: there's another PR open about doing exactly this - so you're very welcome to argue the other PR should modify this linter instead!

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+  descr := "enable the admit linter"
+}
+
+syntax (name := admit) "admit " : tactic
+/-- `admit` is a shorthand for `exact sorry`. -/
+macro "admit" : tactic => `(tactic| exact @sorryAx _ false)
+
+/-- `getadmit t` returns all usages of the `admit` tactic in the input syntax `t`. -/
+partial
+def getadmit : Syntax → Array Syntax
```
grunweg, line 36: Do you think this should be `getAdmit` instead?

grunweg (commented):
Thanks for writing this linter!
I see this is in draft status, but took a look anyway. I have a couple of comments, but all of them are minor/should be quick to address.

grunweg:
Oh, and another question I realised (sorry!, this is really the last one): would you like to add a test for this linter, i.e. demonstrate that the "real" admit is actually linted? (`test/lint.lean` has a few examples to copy from; feel free to ask if you're not sure how to add one and I (and I presume Damiano as well) will be happy to help.

Mathlib/Tactic/Linter/AdmitLinter.lean
```diff
@@ -0,0 +1,62 @@
...
+  | stx@(.node _ kind args) =>
+    let rargs := (args.map getadmit).flatten
+    if kind == ``admit then rargs.push stx else rargs
```
kmill, line 39:
```suggestion
    if kind == ``Lean.Parser.Tactic.tacticAdmit then rargs.push stx else rargs
```
Then you can delete your `admit` definition.

Or, even better, use a syntax quotation like
```lean
partial
def getadmit (stx : Syntax) : Array Syntax :=
  if let `(tactic| admit) := stx then
    #[stx]
  else if let .node _ _ args := stx then
    (args.map getadmit).flatten
  else
    #[]
```
Even shorter:
```lean
partial
def getadmit (stx : Syntax) : Array Syntax :=
  if let `(tactic| admit) := stx then
    #[stx]
  else
    stx.foldArgs (fun arg r => r ++ getadmit arg) #[]
```

Mathlib/Tactic/Linter/AdmitLinter.lean
adomasbaliuka: I copied this from the refineLinter

Mathlib/Tactic/Linter/AdmitLinter.lean
adomasbaliuka: done

Mathlib/Tactic/Linter/AdmitLinter.lean
adomasbaliuka: Mathlib does not allow commits that contain `sorry` or `admit` anyway. So I'm not sure what the linter would be doing in that case?

Mathlib/Tactic/Linter/AdmitLinter.lean
adomasbaliuka: Used the last suggestion.

adomasbaliuka:
> Oh, and another question I realised (sorry!, this is really the last one): would you like to add a test for this linter, i.e. demonstrate that the "real" admit is actually linted? (`test/lint.lean` has a few examples to copy from; feel free to ask if you're not sure how to add one and I (and I presume Damiano as well) will be happy to help.


It seems I have to import something into `test/Lint.lean`? The `refine'` linter that I copied this from also doesn't seem to have a test?

Mathlib/Tactic/Linter/AdmitLinter.lean
adomani: I agree: the copyright holder should probably be Adomas.  I also do not really pay much attention to the copyright section of the files anyway.

Mathlib/Tactic/Linter/AdmitLinter.lean
adomani: The linter would not affect code that gets merged into mathlib, but it would still highlight "in-development" `admit`s.  I also agree that the default value should be `false` anyway: the admit command is defined in core and it should really be the decision of the core developers if they want the `sorry`-synonym to continue existing or not.  The linter gives the opportunity to projects downstream of mathlib to choose whether or not they allow it.

adomasbaliuka: Set default to `false`.

Mathlib/Tactic/Linter/AdmitLinter.lean
adomasbaliuka: @adomani was there a reason this was done for the refineLinter? Should we change it there, too? Or just here?

adomani: I don't have an opinion: I think that in copy-pasting something changed at one point and one of the two options became dominant!

Feel free to uniformize, if you want!

adomasbaliuka: I adjusted this PR to handle namespaces like what most other linters seem to do.

grunweg:
I took a last look: this PR looks good to me. I directly pushed a commit adapting this PR to the (very soon landing) PR #15771, and added a test.
maintainer merge

grunweg:
> > Oh, and another question I realised (sorry!, this is really the last one): would you like to add a test for this linter, i.e. demonstrate that the "real" admit is actually linted? (`test/lint.lean` has a few examples to copy from; feel free to ask if you're not sure how to add one and I (and I presume Damiano as well) will be happy to help.
> 
> @grunweg It seems I have to import something into `test/Lint.lean`? Right now the linter doesn't work in that file. The `refine'` linter that I copied this from also doesn't seem to have a test?

Sorry, it seems my comment was misleading. The `refine'` linter is tested `tests/RefineLinter`: I just pushed a commit adding a test for this linter.

b-mehta (approved):
Nice!

bors merge

PR #15669, opened 2024-08-10
[Merged by Bors] - chore(Geometry): remove remaining `open Classical`

grunweg:
---

For two theorems in `PartitionOfUnity`, I'm unsure if this is the best fix, or adding a suitable decidability assumption is better. Feedback on this is welcome!


Mathlib/Geometry/Manifold/LocalInvariantProperties.lean
```diff
@@ -42,13 +42,9 @@ coincide on `s`, then `LiftPropWithinAt P g' s x` holds. We can't call it
...
 -/
 
-
 noncomputable section
 
-open scoped Classical
-open Manifold Topology
-
-open Set Filter TopologicalSpace
+open Manifold Topology Set Filter TopologicalSpace
```
YaelDillies, line 47: Same complaint as on the other PR: You're probably mangling `open` and `open_locale`

grunweg: Fixed: in fact, the *first* line ought to be *open scoped*.

jcommelin:
LGTM, but is it intentional that you are also deleting a lot of stuff from the nolints file?

bors d+

grunweg:
Not really: this removal is from #15752 (which I mistakenly included... the easiest way forward could be to merge that and then merge master here; that PR should be really easy to review).

grunweg:
Thanks for the review!
bors merge

PR #15671, opened 2024-08-10
[Merged by Bors] - chore(Topology): remove more bare open Classical

grunweg:
The only remaining occurrences are in Topology/Algebra.

---

### DIFF
--- a/Mathlib/Topology/ContinuousFunction/Compact.lean
+++ b/Mathlib/Topology/ContinuousFunction/Compact.lean
@@ -26,12 +26,7 @@ you should restate it here. You can also use
 
 noncomputable section
 
-open scoped Classical
-open Topology NNReal BoundedContinuousFunction
-
-open Set Filter Metric
-
-open BoundedContinuousFunction
+open Topology NNReal BoundedContinuousFunction Set Filter Metric
 
 namespace ContinuousMap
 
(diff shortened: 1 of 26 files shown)


Mathlib/Topology/ContinuousFunction/Compact.lean
```diff
@@ -26,12 +26,7 @@ you should restate it here. You can also use
 
 noncomputable section
 
-open scoped Classical
-open Topology NNReal BoundedContinuousFunction
-
-open Set Filter Metric
-
-open BoundedContinuousFunction
+open Topology NNReal BoundedContinuousFunction Set Filter Metric
```
YaelDillies, line 29: I think you're mangling what was originally `open_locale` with the bare `open`. Can you restore what was there before, but replacing one of the `open` with `open scoped`

Mathlib/Topology/LocallyConstant/Basic.lean
```diff
@@ -19,15 +19,11 @@ This file sets up the theory of locally constant function from a topological spa
...
-
 -/
 
-
 variable {X Y Z α : Type*} [TopologicalSpace X]
 
-open Set Filter
-
-open Topology
+open Set Filter Topology
```
YaelDillies, line 26: Same here

Mathlib/Topology/ContinuousFunction/Compact.lean
grunweg: I am aware of this issue in general. In this case, replacing the `open` by `open scoped` errors.

grunweg:

YaelDillies (approved):
Thanks! 🚀

maintainer merge

PR #15674, opened 2024-08-10
chore(CategoryTheory/Sites): remove unused lemmas and simplify some proofs in Sites.Discrete

dagurtomas:
This PR does two things: 
- `Sheaf.IsDiscrete` was always meant to require the constant sheaf functor to be fully faithful. This PR adds that requirement.
- It removes two lemmas which were essentially duplicates of other ones, and simplifies a proof to be able to remove an auxiliary lemma. 

---

I added these lemmas initially because I used them in #14027, but it turns out they're essentially duplicates of other lemmas, now merged into one lemma `constantSheafAdj_counit_w` which is used in this file and in #14027.

Regarding requiring the constant sheaf functor to be fully faithful: this agrees for example with the definition of discrete object [here](https://ncatlab.org/nlab/show/discrete+object). If one only requires the counit to be an isomorphism and not full faithfulness of the constant sheaf functor, then `X^cst` will in general not be discrete for a set `X`


Mathlib/CategoryTheory/Sites/Discrete.lean
```diff
@@ -41,22 +41,32 @@ namespace CategoryTheory.Sheaf
...
-
 /--
 A sheaf is discrete if it is a discrete object of the "underlying object" functor from the sheaf
 category to the target category.
 -/
-abbrev IsDiscrete (F : Sheaf J A) : Prop := IsIso ((constantSheafAdj J A ht).counit.app F)
+class IsDiscrete (F : Sheaf J A) : Prop :=
+  isIso_counit : IsIso ((constantSheafAdj J A ht).counit.app F)
+  faithful : (constantSheaf J A).Faithful := by infer_instance
+  full : (constantSheaf J A).Full := by infer_instance
```
joelriou, line 51: I am not super happy that a typeclass about `F` contains information that are specific to `J` and `A`. The previous design seemed better to me. Also, the notion of constant sheaf is relevant even when the functor is not fully faithful.

dagurtomas: That's fair. My arguments for this approach are that it just feels weird that in general, an object in the image of the constant sheaf functor is not discrete, and that this terminology follows the literature (if [nLab](https://ncatlab.org/nlab/show/discrete+object) counts as literature).

dagurtomas: I have some code on a different branch where I do the same cleanup for the old approach. There are some lemmas which don't require the fully faithful condition. I'll open a different PR later today or tomorrow and compare the effect on #14027 (it shouldn't matter much because there the constant sheaf functor is fully faithful).

dagurtomas: @joelriou do you prefer #15985?

joelriou: Yes, this seems better to me :-)

dagurtomas:
Closing in favour of #15985

PR #15675, opened 2024-08-10
[Merged by Bors] - feat(Combinatorics/SimpleGraph): bot is not connected

Rida-Hamadani:
---

### DIFF
--- a/Mathlib/Combinatorics/SimpleGraph/Path.lean
+++ b/Mathlib/Combinatorics/SimpleGraph/Path.lean
@@ -724,6 +724,9 @@ lemma bot_preconnected_iff_subsingleton : (⊥ : SimpleGraph V).Preconnected ↔
   contrapose h
   simp [nontrivial_iff.mp <| not_subsingleton_iff_nontrivial.mp h, Preconnected, reachable_bot, h]
 
+lemma bot_preconnected [Subsingleton V] : (⊥ : SimpleGraph V).Preconnected :=
+  bot_preconnected_iff_subsingleton.mpr ‹_›
+
 lemma bot_not_preconnected [Nontrivial V] : ¬(⊥ : SimpleGraph V).Preconnected :=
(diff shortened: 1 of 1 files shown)


YaelDillies (commented):
Can we also have `bot_not_preconnected`?

YaelDillies:
Thanks!

maintainer merge

Mathlib/Combinatorics/SimpleGraph/Path.lean
```diff
@@ -719,6 +719,14 @@ theorem Preconnected.map {G : SimpleGraph V} {H : SimpleGraph V'} (f : G →g H)
 protected lemma Preconnected.mono  {G G' : SimpleGraph V} (h : G ≤ G') (hG : G.Preconnected) :
     G'.Preconnected := fun u v => (hG u v).mono h
 
+lemma bot_preconnected_iff_subsingleton : (⊥ : SimpleGraph V).Preconnected ↔ Subsingleton V := by
+  refine ⟨fun h ↦ ?_, fun h ↦ by simpa [subsingleton_iff, ← reachable_bot] using h⟩
+  contrapose h
+  simp [nontrivial_iff.mp <| not_subsingleton_iff_nontrivial.mp h, Preconnected, reachable_bot, h]
+
+lemma bot_not_preconnected [Nontrivial V] : ¬(⊥ : SimpleGraph V).Preconnected :=
```
b-mehta, line 727: Let's also add that it's connected under the typeclass assumption of subsingleton

b-mehta (approved):
bors d+

Mathlib/Combinatorics/SimpleGraph/Path.lean
Rida-Hamadani: We will need to assume non-emptiness too, right?
We can however do preconnectedness without more assumptions. Is that enough?

b-mehta: Yup, go ahead

PR #15676, opened 2024-08-10
perf(CharZero): de-simp `OfNat.ofNat_eq_ofNat` 

mattrobball:
This is attempted 46478 times for `simp` and is successful 0 times.

---

### DIFF
--- a/Mathlib/Algebra/CharZero/Defs.lean
+++ b/Mathlib/Algebra/CharZero/Defs.lean
@@ -111,7 +111,7 @@ variable [AddMonoidWithOne R] [CharZero R]
   (ofNat_ne_one n).symm
 
 lemma ofNat_eq_ofNat {m n : ℕ} [m.AtLeastTwo] [n.AtLeastTwo] :
-    (ofNat m : R) = (ofNat n) ↔ (ofNat m : ℕ) = ofNat n :=
+    (no_index (ofNat m : R)) = no_index (ofNat n) ↔ (ofNat m : ℕ) = ofNat n :=
   Nat.cast_inj
 
 end OfNat
--- a/Mathlib/Algebra/CharZero/Defs.lean
+++ b/Mathlib/Algebra/CharZero/Defs.lean
@@ -93,6 +93,9 @@ end Nat
 
 namespace OfNat
 
+-- TODO: This whole section is draped in `no_index` because of its special casing in
+-- core, see leanprover/lean4#2867. Hopefully this can be fixed in the future.
+
 variable [AddMonoidWithOne R] [CharZero R]
 
 @[simp] lemma ofNat_ne_zero (n : ℕ) [n.AtLeastTwo] : (no_index (ofNat n) : R) ≠ 0 :=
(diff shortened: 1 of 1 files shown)


mattrobball:
!bench

MichaelStollBayreuth:
<b>General information:</b>
```text
lint:                                                  -9.6715 * 10⁹ (-0.138%)
build:                                                 -68.003 * 10⁹ (-0.0552%)
```


<b>2 files got slower by at least 10⁹ instructions:</b>
```text
Mathlib.NumberTheory.RamificationInertia:              +5.4142 * 10⁹ (+1.79%)
Mathlib.Algebra.Exact:                                 +4.4683 * 10⁹ (+2.99%)
```


<b>10 files got faster by at least 10⁹ instructions:</b>
```text
Mathlib.Algebra.Homology.DifferentialObject:           -3.1134 * 10⁹ (-2.02%)
Mathlib.CategoryTheory.Triangulated.Functor:           -2.6756 * 10⁹ (-0.777%)
Mathlib.GroupTheory.Coxeter.Matrix:                    -2.5101 * 10⁹ (-2.39%)
Mathlib.Algebra.Homology.HomotopyCategory.SingleFunctors:
                                                       -1.8658 * 10⁹ (-2.17%)
Mathlib.AlgebraicTopology.SimplicialObject:            -1.3443 * 10⁹ (-0.640%)
Mathlib.CategoryTheory.Monoidal.Internal.Module:       -1.2882 * 10⁹ (-0.730%)
Mathlib.Topology.Sheaves.SheafCondition.OpensLeCover:  -1.2523 * 10⁹ (-1.31%)
Mathlib.LinearAlgebra.Multilinear.Basic:               -1.2491 * 10⁹ (-0.842%)
Mathlib.CategoryTheory.Idempotents.HomologicalComplex: -1.0946 * 10⁹ (-2.54%)
Mathlib.RepresentationTheory.Action.Limits:            -1.0013 * 10⁹ (-1.47%)
```


MichaelStollBayreuth:
There is a net positive, but it is a bit strange that two files got a bit slower. (Not sure if this is within the "noise" regime...)

mattrobball:
Probably noise. Removing something rejiggers simp priorities slightly so maybe that flaps its wings and ... 

It is worth noting there are 100+ more of these non-applying simp theorems with more than 100 uses. Some have some pretty big type classes assumptions. I am interested to see the effect of bulk removal

eric-wieser:
> This is attempted 46478 times for `simp` and is successful 0 times.

The fact a lemma is never used by mathlib does not mean it is not a good simp lemma; we should consider downstream projects too.

That being said, this particular case is probably more in scope for norm_num. Can you add a test that norm_num can solve this, and a comment on the lemma explaining that it is not `simp` as `norm_num` can handle it?

mattrobball:
I think that being attempted in rewrite almost 50000 times and failing every single time is the definition of a bad simp theorem. 

I'll check norm_num.

eric-wieser:
> I think that being attempted in rewrite almost 50000 times and failing every single time is the definition of a bad simp theorem.

I think the true blame here falls on `OfNat` not being used in DiscrTrees, but that requires a change in core.
A comment referencing the lean4 issue would be fine instead, if the `norm_num` comment is false.

Mathlib/Algebra/CharZero/Defs.lean
```diff
@@ -93,6 +93,9 @@ end Nat
 
 namespace OfNat
 
+-- TODO: This whole section is draped in `no_index` because of its special casing in
+-- core, see leanprover/lean4#2867. Hopefully this can be fixed in the future.
+
```
eric-wieser, line 98: The standard comment we use (on every lemma) is `-- See note [no_index around OfNat.ofNat]`

PR #15678, opened 2024-08-10
[Merged by Bors] - feat(CategoryTheory): `IsIso.comp_isIso` with explicit arguments

dagurtomas:
Often when one wants to prove that a composition is an isomorphism, one proves that each component is. The instance `IsIso.comp_isIso` can be annoying to because of its instance arguments, this PR adds a lemma `IsIso.comp_isIso'` which is exactly the same as the instance, but with explicit arguments.

This avoids having to guess the correct number of underscores in `@IsIso.comp_isIso` or doing `apply ( config := { allowSynthFailures := true } ) IsIso.comp_isIso` 

---


joelriou:
Thanks!

bors merge

PR #15686, opened 2024-08-11
refactor(LinearAlgebra/FreeProduct): move PowerAlgebra construction to its own namespace, minor simp opts

robertmaxton42:
* Make the namespace `asPowers` to simplify/shorten names when working primarily with the power algebra construction of the free product
    * Also adds related some convenience API to `rel'`
* Add simp shortcut attribute to `lift_comp_ι` and add simp shortcut lemma `lift_algebraMap`
* Replaces some `aesop` calls with `simp_rw` for performance

---


kbuzzard:
Just looking at the module docstring of this file -- it claims that in the category of commutative $R$-algebras, the coproduct is just `PiTensorProduct`. I think PiTensorProduct is much too big to be the coproduct if the indexing set is infinite. For example I think a countably infinite PiTensorProduct of 1-dimensional vector spaces over the rationals or reals is uncountably infinite-dimensional, because e.g. 1 tensor 1 tensor 1 tensor ... and 2 tensor 2 tensor 2 tensor ... are linearly independent (one is "infinitely bigger than the other").

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -116,37 +116,97 @@ abbrev FreeTensorAlgebra := TensorAlgebra R (⨁ i, A i)
...
 @[reducible] def _root_.LinearAlgebra.FreeProduct := RingQuot <| FreeProduct.rel R A
 
 /--The free product of the collection of `R`-algebras `A i`, as a quotient of `PowerAlgebra R A`-/
-@[reducible] def _root_.LinearAlgebra.FreeProduct_ofPowers := RingQuot <| FreeProduct.rel' R A
+@[reducible] def asPowers := RingQuot <| FreeProduct.rel' R A
 
-/--The `R`-algebra equivalence relating `FreeProduct` and `FreeProduct_ofPowers`-/
-noncomputable def equivPowerAlgebra : FreeProduct_ofPowers R A ≃ₐ[R] FreeProduct R A :=
+/--The `R`-algebra equivalence relating `FreeProduct` and `FreeProduct.asPowers`-/
+@[simps (config := {isSimp := false, rhsMd := .default, simpRhs := true})]
```
kbuzzard, line 208: Why `isSimp := false`?

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -226,10 +353,13 @@ to a unique arrow `π` from `FreeProduct R A` such that  `π ∘ ι i = maps i`.
 /--Universal property of the free product of algebras, property:
 for every `R`-algebra `B`, every family of maps `maps : (i : I) → (A i →ₐ[R] B)` lifts
 to a unique arrow `π` from `FreeProduct R A` such that  `π ∘ ι i = maps i`.-/
-theorem lift_comp_ι : (lift R A maps) ∘ₐ (ι R A i) = maps := by
+@[simp↓] theorem lift_comp_ι : (lift R A maps) ∘ₐ (ι R A i) = maps := by
```
kbuzzard, line 356: Why `simp↓`?

YaelDillies, line 356:
```suggestion
@[simp↓] theorem lift_comp_ι : lift R A maps ∘ₐ ι R A i = maps := by
```

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -116,37 +116,97 @@ abbrev FreeTensorAlgebra := TensorAlgebra R (⨁ i, A i)
...
 @[reducible] noncomputable def powerAlgebra_equiv_freeAlgebra :
     PowerAlgebra R A ≃ₐ[R] FreeTensorAlgebra R A :=
   TensorAlgebra.equivDirectSum.symm
 
+theorem PowerAlgebra.ofDirectSum_ι :
+    (ofDirectSum.toLinearMap ∘ₗ PowerAlgebra.ι R A) = TensorAlgebra.ι R := by
+  ext i a
+  simp_rw [ι, LinearMap.comp_apply, lof_eq_of, MultilinearMap.ofSubsingleton_symm_apply_apply,
+            ofDirectSum.toLinearMap_apply, ofDirectSum_of_tprod, tprod_apply,
+           List.ofFn_const, List.replicate_one, List.prod_singleton]
```
YaelDillies, line 136: We don't match the indent of the start of the list in `simp_rw` calls
```suggestion
    ofDirectSum.toLinearMap_apply, ofDirectSum_of_tprod, tprod_apply, List.ofFn_const,
    List.replicate_one, List.prod_singleton]
```

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -116,37 +116,97 @@ abbrev FreeTensorAlgebra := TensorAlgebra R (⨁ i, A i)
...
+      |>.symm (PiTensorProduct.tprod R))
+
 /--The free tensor algebra and its representation as an infinite direct sum
 of tensor powers are (noncomputably) equivalent as `R`-algebras.-/
 @[reducible] noncomputable def powerAlgebra_equiv_freeAlgebra :
     PowerAlgebra R A ≃ₐ[R] FreeTensorAlgebra R A :=
   TensorAlgebra.equivDirectSum.symm
 
+theorem PowerAlgebra.ofDirectSum_ι :
+    (ofDirectSum.toLinearMap ∘ₗ PowerAlgebra.ι R A) = TensorAlgebra.ι R := by
```
YaelDillies, line 132:
```suggestion
    ofDirectSum.toLinearMap ∘ₗ PowerAlgebra.ι R A = TensorAlgebra.ι R := by
```

YaelDillies (commented):
Thanks for your PR! Could you please address Kevin's comments and clarify the kind of this PR? You have currently written "refactor", but it adds a net 130 lines. That seems more like a "feat".

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
```diff
@@ -116,37 +116,97 @@ abbrev FreeTensorAlgebra := TensorAlgebra R (⨁ i, A i)
...
   | prod : ∀ {i : I} {a₁ a₂ : A i},
       rel
         (tprod R (⨁ i, A i) 2 (fun | 0 => lof R I A i a₁ | 1 => lof R I A i a₂))
         (ι R <| lof R I A i (a₁ * a₂))
 
-/--The generating equivalence relation for elements of the power algebra
-that are identified in the free product. -/
-@[reducible, simp] def rel' := rel R A on ofDirectSum
-
+-- We reproduce `rel.id` and `.prod here to make `R, A, i...` explicit.
```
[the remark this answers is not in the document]
robertmaxton42: `rel.id` and `rel.prod` are automatically defined by the compiler when I define `rel` above; however, they aren't terribly ergonomic to use, as Lean automatically makes `R` and `A` implicit, and I can't figure out how to override that. Rather than fight the compiler too much, I just defined a second copy with more useful binders.

Mathlib/LinearAlgebra/FreeProduct/Basic.lean
robertmaxton42: The lhs of this definition is not in simp-normal form, but it is in, shall we say, "human-normal form" -- it's something that I expect actual users of this file to type out at some point while using it, certainly it's something I used often enough. Meanwhile, the actual simp-normal form for the lhs is an illegible mess. `@[simp↓]` marks the resulting simp lemma as a "shortcut simp" that should be applied immediately on its lhs at a high priority, before applying other simp lemmas, so that I don't need to define a separate lemma with a messy lhs just for `simp` to use.

robertmaxton42:
> Just looking at the module docstring of this file -- it claims that in the category of commutative R -algebras, the coproduct is just `PiTensorProduct`. I think PiTensorProduct is much too big to be the coproduct if the indexing set is infinite. For example I think a countably infinite PiTensorProduct of 1-dimensional vector spaces over the rationals or reals is uncountably infinite-dimensional, because e.g. 1 tensor 1 tensor 1 tensor ... and 2 tensor 2 tensor 2 tensor ... are linearly independent (one is "infinitely bigger than the other"). What do you think?

That ... is an interesting argument. I think I would argue that _either_ we must be working over the extended reals, in which case there's something in the ring of the "correct" infinite size, _or_ the very fact that our ring doesn't have infinity in it implies that any such infinite product must "really" be an infinite limit and should be handled accordingly -- but I will freely admit I am confused on the point and I'm not sure that argument would go through in Lean.

Instead, I will retreat to the direct categorical argument. Let $A_1, A_2, B$ be $R$-algebras, then we can define the tensor product with a universal property: there exist canonical injections $\iota_1: A_1 \to A\otimes B, \iota_2: A_2 \to A\otimes B$ s.t. for all $B$ and for any $f:A_1 \to B, g:A_2\to B$ whose images in $B$ commute, there exists a unique lift $h:A_1 \otimes A_2 \to B$ s.t. $h\circ\iota_1 = f$ and $h\circ\iota_2 = g$. (We have this in Mathlib as [Algebra.TensorProduct.liftEquiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/TensorProduct/Basic.html#Algebra.TensorProduct.liftEquiv).) But the universal property of the coproduct is precisely the same, except without the requirement that the images commute. Thus if the category consists only of commutative $R$-algebras, the binary tensor product coincides with the binary coproduct, and in particular this is true of the category $R\mathbf{\text{-}CAlg}$ itself. 

I am unsure how well this argument extends to arbitrarily infinite index sets, but it should certainly extend as far as countably infinite sets at least.

robertmaxton42:
> Thanks for your PR! Could you please address Kevin's comments and clarify the kind of this PR? You have currently written "refactor", but it adds a net 130 lines. That seems more like a "feat".

I mean, it can be a 'feat' if you want it to be! The biggest effect of this PR IMO is that there are now a bunch of lemmas that are spelled `asPowers.foo` instead of `foo_asPowers`, so I think of it as a 'refactor'; there's not much mathematical content, it's just providing different access to an equivalent formulation, plus a bunch of usability stuff like simp annotations.

YaelDillies:
> I mean, it can be a 'feat' if you want it to be! The biggest effect of this PR IMO is that there are now a bunch of lemmas that are spelled `asPowers.foo` instead of `foo_asPowers`, so I think of it as a 'refactor'; there's not much mathematical content

I suspect the confusion come from the fact that you have four bullet points in your description, but the first one (ie the titular) only makes up for a few lines of changes, while the other three make up for the bulk of the PR. I'd suggest splitting this PR in two: one "refactor" PR for the first bullet point, one "feat" PR for the other three

PR #15690, opened 2024-08-11
[Merged by Bors] - feat(SetTheory/Game/Ordinal): game product of ordinals is natural product

vihdzp:
I also flipped and renamed `toPGame_nadd` at the request of Yaël.

---

- [x] depends on: #15708

It's been a long while since I wrote any Lean code, so it might be possible to golf my proofs a bit.


Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -389,6 +399,50 @@ theorem quot_mul_comm (x y : PGame.{u}) : (⟦x * y⟧ : Game) = ⟦y * x⟧ :=
 theorem mul_comm_equiv (x y : PGame) : x * y ≈ y * x :=
   Quotient.exact <| quot_mul_comm _ _
 
+instance isEmpty_leftMoves_mul (x y : PGame.{u}) [IsEmpty x.LeftMoves] [IsEmpty y.RightMoves] :
+    IsEmpty (x * y).LeftMoves := by
```
YaelDillies, line 403:
```suggestion
instance isEmpty_leftMoves_mul (x y : PGame.{u})
    [IsEmpty (x.LeftMoves × y.RightMoves ⊕ x.RightMoves × y.LeftMoves] :
    IsEmpty (x * y).LeftMoves := ‹_›
```
seems more principled (and avoids the instance duplication, and forgetting some combinations of instances as you have done now)

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -180,4 +180,32 @@ termination_by a b => (a, b)
 theorem toPGame_add_mk' (a b : Ordinal) : (⟦a.toPGame⟧ + ⟦b.toPGame⟧ : Game) = ⟦(a ♯ b).toPGame⟧ :=
   Quot.sound (toPGame_add a b)
 
+/-- The product of ordinals as games corresponds to natural multiplication of ordinals. -/
+theorem toPGame_mul : ∀ a b : Ordinal.{u}, a.toPGame * b.toPGame ≈ (a ⨳ b).toPGame
```
YaelDillies, line 184: Can you turn it around and make it simp?

Mathlib/SetTheory/Game/Basic.lean
vihdzp: The combinations I omitted correspond to when either game is identical to the zero game. So actually, your approach should generalize those instances too.

YaelDillies: Yes, that is indeed my point

Mathlib/SetTheory/Game/Ordinal.lean
vihdzp: This is an equivalence, not an equality. As for the subsequent result written for quotients, I'd argue that it should either stay in its current direction or not be a `simp` lemma at all. The intention of these theorems is to turn the more complex operations on games into comparatively well-characterized operations on ordinals.

vihdzp: Actually, I think it makes sense to simply remove the `simp` lemma. Even if the RHS is simpler to work with, it's not structurally simpler in the way `simp` is supposed to be.

YaelDillies: > The intention of these theorems is to turn the more complex operations on games into comparatively well-characterized operations on ordinals.

Yes, but it is also a distributivity lemma, and distributivity lemmas are always written in the form "non-distributed = distributed"

vihdzp: I'll rewrite the lemmas, but keep the simps out then.

Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -101,20 +101,24 @@ If `0 ⧏ x` (less or fuzzy with), then Left can win `x` as the first player. -/
 def LF : Game → Game → Prop :=
   Quotient.lift₂ PGame.LF fun _ _ _ _ hx hy => propext (lf_congr hx hy)
 
-local infixl:50 " ⧏ " => LF
-
 /-- On `Game`, simp-normal inequalities should use as few negations as possible. -/
 @[simp]
-theorem not_le : ∀ {x y : Game}, ¬x ≤ y ↔ y ⧏ x := by
+theorem not_le : ∀ {x y : Game}, ¬x ≤ y ↔ Game.LF y x := by
```
YaelDillies, line 106: I don't understand this change. The statement is the same as before, right?

YaelDillies (commented):
Do you mind fixing the `Game.PGame` namespace issue in a preliminary PR? I have a hard time understanding what's going on

Mathlib/SetTheory/Game/Basic.lean
vihdzp: Since this is now in a separate code section, the infix notation is no longer there. I didn't think it was worth it to enable it for two statements.

Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -149,6 +149,9 @@ theorem lt_iff_game_lt {x y : PGame} : x < y ↔ (⟦x⟧ : Game) < ⟦y⟧ :=
 theorem equiv_iff_game_eq {x y : PGame} : x ≈ y ↔ (⟦x⟧ : Game) = ⟦y⟧ :=
   (@Quotient.eq' _ _ x y).symm
 
+theorem game_eq {x y : PGame} : x ≈ y → (⟦x⟧ : Game) = ⟦y⟧ :=
+  equiv_iff_game_eq.1
```
YaelDillies, line 153:
```suggestion
alias ⟨game_eq, _⟩ := equiv_iff_game_eq
```

Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -396,21 +399,19 @@ theorem quot_mul_comm (x y : PGame.{u}) : (⟦x * y⟧ : Game) = ⟦y * x⟧ :=
...
+    [IsEmpty (x.LeftMoves × y.LeftMoves ⊕ x.RightMoves × y.RightMoves)] :
+    IsEmpty (x * y).LeftMoves := by
   cases x
-  apply instIsEmptySum
-
-instance isEmpty_zero_mul_leftMoves (x : PGame.{u}) : IsEmpty (0 * x).LeftMoves := by
-  cases x
-  apply instIsEmptySum
+  cases y
+  assumption
```
YaelDillies, line 407: Do you even need the `cases x; cases y`?

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
   map_rel_iff' := @toPGame_le_iff
 
 /-- The sum of ordinals as games corresponds to natural addition of ordinals. -/
-theorem toPGame_add : ∀ a b : Ordinal.{u}, a.toPGame + b.toPGame ≈ (a ♯ b).toPGame
+theorem toPGame_add : ∀ a b : Ordinal.{u}, (a ♯ b).toPGame ≈ a.toPGame + b.toPGame
   | a, b => by
```
YaelDillies, line 159:
```suggestion
theorem toPGame_nadd (a b : Ordinal.{u}) : (a ♯ b).toPGame ≈ a.toPGame + b.toPGame := by
```

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
+      (try rw [add_moveLeft_inl]) <;>
+      (try rw [add_moveLeft_inr]) <;>
+      rw [toPGame_moveLeft', ← lf_congr_left (toPGame_add _ _), toPGame_lf_iff]
+      · exact nadd_lt_nadd_right wf _
+      · exact nadd_lt_nadd_left wf _
 termination_by a b => (a, b)
 
-@[simp]
-theorem toPGame_add_mk' (a b : Ordinal) : (⟦a.toPGame⟧ + ⟦b.toPGame⟧ : Game) = ⟦(a ♯ b).toPGame⟧ :=
+theorem toPGame_add_mk' (a b : Ordinal) : (⟦(a ♯ b).toPGame⟧ : Game) = ⟦a.toPGame⟧ + ⟦b.toPGame⟧ :=
```
YaelDillies, line 179:
```suggestion
theorem toPGame_nadd_mk' (a b : Ordinal) : (⟦(a ♯ b).toPGame⟧ : Game) = ⟦a.toPGame⟧ + ⟦b.toPGame⟧ :=
```

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
   Quot.sound (toPGame_add a b)
 
+/-- The product of ordinals as games corresponds to natural multiplication of ordinals. -/
+theorem toPGame_mul : ∀ a b : Ordinal.{u}, (a ⨳ b).toPGame ≈ a.toPGame * b.toPGame
+  | a, b => by
```
YaelDillies, line 184:
```suggestion
theorem toPGame_nmul (a b : Ordinal.{u}) (a ⨳ b).toPGame ≈ a.toPGame * b.toPGame := by
```

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
+      intro i j
+      rw [mul_moveLeft_inl, toPGame_moveLeft', toPGame_moveLeft', lf_iff_game_lf,
+        quot_sub, quot_add, ← Game.not_le, le_sub_iff_add_le]
+      repeat rw [← game_eq (toPGame_mul _ _)]
+      repeat rw [← toPGame_add_mk']
+      apply toPGame_lf (nmul_nadd_lt _ _) <;>
+      exact toLeftMovesToPGame_symm_lt _
+termination_by a b => (a, b)
+
+theorem toPGame_mul_mk' (a b : Ordinal) : (⟦(a ⨳ b).toPGame⟧ : Game) = ⟦a.toPGame * b.toPGame⟧ :=
```
YaelDillies, line 206:
```suggestion
theorem toPGame_nmul_mk' (a b : Ordinal) : (⟦(a ⨳ b).toPGame⟧ : Game) = ⟦a.toPGame * b.toPGame⟧ :=
```

Mathlib/SetTheory/Game/Basic.lean
vihdzp: Yes, for whatever reason statements like `(mk xl xr xL xR + mk yl yr yL yR).LeftMoves = xl ⊕ yl` are def-eq, while `(x + y).LeftMoves = x.LeftMoves ⊕ y.LeftMoves` aren't.

Mathlib/SetTheory/Game/Ordinal.lean
vihdzp: Putting `a` and `b` before the colon doesn't work, since it breaks the `termination_by` syntax.

YaelDillies: Ugh, annoying. @nomeata? 😇

Mathlib/SetTheory/Game/Basic.lean
YaelDillies: That's quite weird. Eta for structures should have fixed this... unless it's something to do with well-founded recursion being irreducible by default now? I am confused

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
-      rw [toPGame_moveLeft', lf_congr_left (toPGame_add _ _), toPGame_lf_iff]
+      rw [toPGame_moveLeft', ← lf_congr_left (toPGame_nadd _ _), toPGame_lf_iff]
       · exact nadd_lt_nadd_right wf _
       · exact nadd_lt_nadd_left wf _
+termination_by a b => (a, b)
+
+theorem toPGame_nadd_mk' (a b : Ordinal) : (⟦(a ♯ b).toPGame⟧ : Game) = ⟦a.toPGame⟧ + ⟦b.toPGame⟧ :=
+  Quot.sound (toPGame_nadd a b)
+
+/-- The product of ordinals as games corresponds to natural multiplication of ordinals. -/
```
YaelDillies, line 182:
```suggestion
/-- Natural multiplication of ordinals corresponds to the product of ordinals as games. -/
```
since we turned the lemma around. Same above

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -155,29 +155,55 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
+      repeat rw [← game_eq (toPGame_nmul _ _)]
+      repeat rw [← toPGame_nadd_mk']
+      apply toPGame_lf (nmul_nadd_lt _ _) <;>
+      exact toLeftMovesToPGame_symm_lt _
 termination_by a b => (a, b)
 
-@[simp]
-theorem toPGame_add_mk' (a b : Ordinal) : (⟦a.toPGame⟧ + ⟦b.toPGame⟧ : Game) = ⟦(a ♯ b).toPGame⟧ :=
-  Quot.sound (toPGame_add a b)
+theorem toPGame_nmul_mk' (a b : Ordinal) : (⟦(a ⨳ b).toPGame⟧ : Game) = ⟦a.toPGame * b.toPGame⟧ :=
```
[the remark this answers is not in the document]
vihdzp: Be careful! Multiplication isn't defined on games, as products of equivalent games aren't always equivalent.

We don't have a name for that map yet. What's the notation for defining an abbreviation like that?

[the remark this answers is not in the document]
YaelDillies: Ah! Yeah that's annoying. Then my suggestion doesn't really make sense.

The notation is `abbrev Ordinal.toGame (o : Ordinal) : Game := ⟦o.toPGame⟧` or `@[reducible, inline] def Ordinal.toGame (o : Ordinal) : Game := ⟦o.toPGame⟧`

Mathlib/SetTheory/Game/Ordinal.lean
nomeata: Stuff before the colon is in scope for `termination_by`, so just don't bind it there again?

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -154,30 +154,57 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
+    (try rw [add_moveLeft_inr]) <;>
+    rw [toPGame_moveLeft', ← lf_congr_left (toPGame_nadd _ _), toPGame_lf_iff]
+    · exact nadd_lt_nadd_right wf _
+    · exact nadd_lt_nadd_left wf _
+termination_by (a, b)
+
+theorem toGame_nadd (a b : Ordinal) : (a ♯ b).toGame = a.toGame + b.toGame :=
+  Quot.sound (toPGame_nadd a b)
+
+/-- The natural multiplication of ordinals corresponds to their product as games. -/
```
YaelDillies, line 184:
```suggestion
/-- The natural multiplication of ordinals corresponds to their product as pre-games. -/
```

YaelDillies (approved):
Thanks!

maintainer merge

eric-wieser:
bors merge

Thanks!

PR #15691, opened 2024-08-11
[Merged by Bors] - doc(Tactic): expand `ring` and `ring_nf` docstrings

rwst:
Copy some explanations from other comments into the actual `ring_nf` docstring. Add details to `ring` docstring.

---

### DIFF
--- a/Mathlib/Tactic/Ring/RingNF.lean
+++ b/Mathlib/Tactic/Ring/RingNF.lean
@@ -196,6 +196,12 @@ which rewrites all ring expressions into a normal form.
   * `recursive`: if true, `ring_nf` will also recurse into atoms
 * `ring_nf` works as both a tactic and a conv tactic.
   In tactic mode, `ring_nf at h` can be used to rewrite in a hypothesis.
+
+
+This can be used non-terminally to normalize ring expressions in the goal such as
+`⊢ P (x + x + x)` ~> `⊢ P (x * 3)`, as well as being able to prove some equations that
+`ring` cannot because they involve ring reasoning inside a subterm, such as
+`sin (x + y) + sin (y + x) = 2 * sin (x + y)`.
 -/
 elab (name := ringNF) "ring_nf" tk:"!"? cfg:(config ?) loc:(location)? : tactic => do
   let mut cfg ← elabConfig cfg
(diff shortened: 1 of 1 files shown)


Mathlib/Tactic/Ring/RingNF.lean
```diff
@@ -249,6 +249,11 @@ example (n : ℕ) (m : ℤ) : 2^(n+1) * m = 2 * 2^n * m := by ring
 example (a b : ℤ) (n : ℕ) : (a + b)^(n + 2) = (a^2 + b^2 + a * b + b * a) * (a + b)^n := by ring
 example (x y : ℕ) : x + id y = y + id x := by ring!
 ```
+
+This can be used non-terminally to normalize ring expressions in the goal such as
```
bryangingechen, line 253: What's the referent of "This" here? It doesn't seem like it is the `ring` tactic, since the following sentence says it can be used to prove things that `ring` cannot.

Since this text was copied from the module docstring, maybe it applies to `ring_nf` instead? But the module docstring is also confusing to me since it states that `ring_nf` uses `ring`, but as far as I can tell, it's actually the other way around?

cc: @digama0

rwst: I agree. There should be some fixes. Let's use this PR for it.

rwst: Okay, moved the passage to the `ring_nf` docstring. Thanks.

digama0: @bryangingechen `ring_nf` uses `ring` infrastructurally (`ring_nf` is the more complicated tactic, built on `ring`), but it also installs itself as a hook to `ring` so that `ring` will try `ring_nf` on failure and suggest that.

Mathlib/Tactic/Ring/RingNF.lean
```diff
@@ -238,7 +243,8 @@ elab (name := ring1NF) "ring1_nf" tk:"!"? cfg:(config ?) : tactic => do
 
 /--
 Tactic for evaluating expressions in *commutative* (semi)rings, allowing for variables in the
-exponent.
+exponent. It uses `ring1` and `ring_nf` in succession so, if `ring1` fails, `ring_nf` will
+not be called.
```
Ruben-VandeVelde, line 247: Speaking for myself, I'm not sure this addition actually clarifies anything. (I have a reasonable intuition for what `ring` does, but don't have a clue about `ring1`.) I'm not blocking this PR, but please think about whether this is the most useful thing to show users?

digama0: the user-level explanation of `ring1`  is: it's `ring`, but it fails instead of trying and suggesting `ring_nf` if the goal is not appropriate for `ring`.

rwst: Changed accordingly and added example. I think it's good to have, as I myself was puzzled for a long time by the `ring_nf` suggestion.

joneugster:
Looks good to me.

maintainer merge

mattrobball:
bors merge

PR #15694, opened 2024-08-11
[Merged by Bors] - feat(Set/Prod): add `dcomp_image_pi`

urkud:
---


Mathlib/Data/Set/Prod.lean
```diff
@@ -802,6 +802,23 @@ theorem eval_image_univ_pi (ht : (pi univ t).Nonempty) :
     (fun f : ∀ i, α i => f i) '' pi univ t = t i :=
   eval_image_pi (mem_univ i) ht
 
+theorem dcomp_image_pi {f : ∀ i, α i → β i} (hf : ∀ i ∉ s, Surjective (f i)) (t : ∀ i, Set (α i)) :
```
urkud, line 805: Should we use `dcomp` or `piMap` as the name (both def and theorems)?

jcommelin (commented):
`dcomp` seems fine with me

bors merge

PR #15695, opened 2024-08-11
[Merged by Bors] - feat: the monoidal category structure on homological complexes

joelriou:
Let `c : ComplexShape I` with `I` an additive monoid. If `c` is equipped with the data and axioms `c.TensorSigns`, then the category `HomologicalComplex C c` can be eqquiped with a monoidal category structure if `C` is a monoidal category such that `C` has certain coproducts and both left/right tensoring commute with these.

In particular, we obtain a monoidal category structure on `ChainComplex C ℕ` when `C` is an additive monoidal category.

Co-authored-by: Kim Morrison <kim@tqft.net>

---

- [x] depends on: #15719
- [x] depends on: #15712
- [x] depends on: #15740
- [x] depends on: #15225


jcommelin:
Thanks :tada:

bors merge

PR #15708, opened 2024-08-11
[Merged by Bors] - chore(SetTheory/Game/Basic): fix Game.PGame namespace

vihdzp:
We move the `Fuzzy` definition so that the `PGame` namespace doesn't need to be opened twice. We also move the infix notation definition a bit later down the file.

---


YaelDillies (approved):
Thanks for fixing the mess

maintainer merge

semorrison:
bors merge

PR #15714, opened 2024-08-12
[Merged by Bors] - chore: polishing the IMO solution 2024Q1

madvorak:
---


joneugster (approved):
Looks good, thanks!

maintainer merge

Archive/Imo/Imo2024Q1.lean
```diff
@@ -68,31 +70,31 @@ lemma mem_Ico_one_of_mem_Ioo (h : α ∈ Set.Ioo 0 2) : α ∈ Set.Ico 1 2 := by
...
       calc 1 ≤ α⁻¹ * α := by simp [h0.ne']
-        _ ≤ ⌈α⁻¹⌉₊ * α := by gcongr; exact Nat.le_ceil _
-    · calc ⌈α⁻¹⌉₊ * α < (α⁻¹ + 1) * α := by gcongr; exact Nat.ceil_lt_add_one (inv_nonneg.2 h0.le)
+        _ ≤ ⌈α⁻¹⌉₊ * α := (mul_le_mul_right h0).2 (Nat.le_ceil α⁻¹)
+    · calc ⌈α⁻¹⌉₊ * α
+        _ < (α⁻¹ + 1) * α := (mul_lt_mul_right h0).2 (Nat.ceil_lt_add_one (inv_nonneg.2 h0.le))
         _ = 1 + α := by field_simp [h0.ne']
-        _ ≤ (1 : ℕ) + 1 := by gcongr; norm_cast
-  · refine Finset.sum_eq_zero ?_
+        _ ≤ (1 : ℕ) + 1 := add_le_add Nat.cast_one.symm.le hn.le
```
jcommelin, line 87: I personally prefer the `gcongr; norm_cast` version.

Archive/Imo/Imo2024Q1.lean
```diff
@@ -30,21 +30,23 @@ namespace Imo2024Q1
 /-- The condition of the problem. -/
 def Condition (α : ℝ) : Prop := (∀ n : ℕ, 0 < n → (n : ℤ) ∣ ∑ i ∈ Finset.Icc 1 n, ⌊i * α⌋)
 
+/-- This is to be determined by the solver of the original problem. -/
+def solutionSet : Set ℝ := {α : ℝ | ∃ m : ℤ, α = 2 * m}
```
jcommelin, line 34: Is there a reason for moving this to the top? I think it also make sense to have it really close to the main result at the bottom.

jcommelin:
I would like to hear @jsm28 's opinion on these changes.

Archive/Imo/Imo2024Q1.lean
madvorak: I thought it might be better to readers to first see what we are proving and then start building lemmas towards it. If you want to keep it at the end, ok, but don't you want to inline it into the theorem?

Archive/Imo/Imo2024Q1.lean
madvorak: Only here, or everywhere?

jsm28 (commented):
I don't have any comments on these changes.

Archive/Imo/Imo2024Q1.lean
joneugster: Is there generally any preference in imo files to use tactic- or term-mode?

Maybe

``` ... := by
  gcongr
  norm_cast
  apply ...
```

in that case?

Archive/Imo/Imo2024Q1.lean
madvorak: If you don't want it on the top, it would better be inlined, right?

Archive/Imo/Imo2024Q1.lean
madvorak: Undo this change, or all changes to terms in `calc` ?

Archive/Imo/Imo2024Q1.lean
jcommelin: Maybe this is a nice use case for `recall`. So that you can recall both the `Condition` and `solutionSet` just before the final result.

Archive/Imo/Imo2024Q1.lean
jcommelin: I would prefer to revert all of them. `gcongr` is a very natural atomic step in my opinion. I think we should showcase it.
But my opinion isn't very strong.

madvorak: All right, I will revert them all.

jcommelin (commented):
Thanks :tada:

bors merge

jcommelin:
Can you please merge master? Maybe that will make CI happy.

jcommelin:
Thanks :tada:

bors merge

PR #15716, opened 2024-08-12
[Merged by Bors] - feat(SetTheory/Game/Birthday): Define birthday of a game

vihdzp:
The updated module description explains how this differs from the birthday of a pre-game.

---

- [x] depends on: #15690 (this PR introduces some API changes to the Ordinal file)


Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +159,112 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+  sInf (PGame.birthday '' (Quotient.mk' ⁻¹' {x}))
 
-end PGame
+/-- The set in the definition of birthday is nonempty. -/
+theorem birthday_nonempty (x : Game) : (PGame.birthday '' (Quotient.mk' ⁻¹' {x})).Nonempty := by
+  rw [Set.image_nonempty]
+  exact ⟨_, x.out_eq⟩
+
+theorem birthday_eq_pGame_birthday (x : Game) : ∃ y : PGame.{u},
+    ⟦y⟧ = x ∧ y.birthday = birthday x :=
```
eric-wieser, line 186: Best to wrap at the `:` before wrapping elsewhere:
```suggestion
theorem birthday_eq_pGame_birthday (x : Game) :
    ∃ y : PGame.{u}, ⟦y⟧ = x ∧ y.birthday = birthday x :=
```

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +159,112 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+  apply le_antisymm
+  · conv_rhs => rw [← PGame.toPGame_birthday o]
+    apply birthday_le_pGame_birthday
+  · let ⟨x, hx₁, hx₂⟩ := birthday_eq_pGame_birthday o.toGame
+    rw [← hx₂, ← toPGame_le_iff]
+    rw [← PGame.equiv_iff_game_eq] at hx₁
+    exact hx₁.2.trans (PGame.le_birthday x)
+
+@[simp]
+theorem birthday_natCast (n : ℕ) : birthday n = n := by
```
eric-wieser, line 219: Could you add `birthday_ofNat` too, using `no_index (OfNat.ofNat n)`?

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +159,112 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+    exact PGame.star_fuzzy_zero.not_equiv
+
+private theorem neg_birthday' (x : Game) : (-x).birthday ≤ x.birthday := by
+  let ⟨y, hy₁, hy₂⟩ := birthday_eq_pGame_birthday x
+  rw [← hy₂, ← PGame.neg_birthday y]
+  conv_lhs => rw [← hy₁]
+  apply birthday_le_pGame_birthday
+
+@[simp]
+theorem neg_birthday (x : Game) : (-x).birthday = x.birthday := by
```
eric-wieser, line 243:
```suggestion
theorem birthday_neg (x : Game) : (-x).birthday = x.birthday := by
```

Mathlib/SetTheory/Game/Birthday.lean
vihdzp: What's `no_index`? Does `Game` even have an `OfNat.ofNat` instance? I don't believe we use it anywhere else in it's API.

eric-wieser: > What's `no_index`?

In practice, a workaround for janky behavior by `simp` around `ofNat`

> Does `Game` even have an `OfNat.ofNat` instance?

Good question; I assumed it would have since it seems to have a `NatCast`, and those usually come together. I didn't actually check though. Is there an `AddMonoidWithOne` instance?

eric-wieser: Yes, it exists:
```
#synth OfNat SetTheory.Game 37
```

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -154,30 +154,68 @@ noncomputable def toPGameEmbedding : Ordinal.{u} ↪o PGame.{u} where
...
+    exact toLeftMovesToPGame_symm_lt _
+termination_by (a, b)
+
+theorem toGame_nmul (a b : Ordinal) : (a ⨳ b).toGame = ⟦a.toPGame * b.toPGame⟧ :=
+  Quot.sound (toPGame_nmul a b)
 
 @[simp]
-theorem toPGame_add_mk' (a b : Ordinal) : (⟦a.toPGame⟧ + ⟦b.toPGame⟧ : Game) = ⟦(a ♯ b).toPGame⟧ :=
-  Quot.sound (toPGame_add a b)
+theorem nat_toGame : ∀ n : ℕ, toGame n = n
```
eric-wieser, line 211:
```suggestion
theorem toGame_natCast : ∀ n : ℕ, toGame n = n
```

and the `ofNat` lemma too

Mathlib/SetTheory/Game/Birthday.lean
vihdzp: Is there any theorem saying `OfNat.ofNat n = ↑n`? There's [Nat.cast_eq_ofNat](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Cast/Defs.html#Nat.cast_eq_ofNat) but only for naturals greater than 2.

eric-wieser: No, it should assume `[n.AtLeastTwo]`. The theorem you want is then true by rfl, so you can prove the whole thing in terms of the natCast lemma

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +159,119 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+    rw [← PGame.equiv_iff_game_eq] at hx₁
+    exact hx₁.2.trans (PGame.le_birthday x)
+
+@[simp]
+theorem birthday_natCast (n : ℕ) : birthday n = n := by
+  rw [← toGame_natCast]
+  exact toPGame_birthday _
+
+@[simp]
+theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] : birthday (no_index (OfNat.ofNat n)) = n :=
```
eric-wieser, line 224:
```suggestion
theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] :
    birthday (no_index (OfNat.ofNat n)) = OfNat.ofNat n :=
```

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +159,119 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+    rw [← hx₂, ← toPGame_le_iff]
+    rw [← PGame.equiv_iff_game_eq] at hx₁
+    exact hx₁.2.trans (PGame.le_birthday x)
+
+@[simp]
+theorem birthday_natCast (n : ℕ) : birthday n = n := by
+  rw [← toGame_natCast]
+  exact toPGame_birthday _
+
+@[simp]
```
eric-wieser, line 223: Can you add the usual `-- See note` comment that appears before `no_index`/`OfNat` lemmas?

vihdzp: Sorry, I don't think I've ever seen that note. Is there any other theorem I can copy it from?

eric-wieser: Searching for `OfNat.ofNat` should find the comment quite quickly

vihdzp: Oh I found it, I think I might have had caps lock on or something.

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -35,8 +39,8 @@ open scoped NaturalOps PGame
 namespace PGame
 
 /-- The birthday of a pre-game is inductively defined as the least strict upper bound of the
-birthdays of its left and right games. It may be thought as the "step" in which a certain game is
-constructed. -/
+  birthdays of its left and right games. It may be thought as the "step" in which a certain game is
+  constructed. -/
```
eric-wieser, line 43: I think the convention is not to indent here

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -3,27 +3,31 @@ Copyright (c) 2022 Violeta Hernández Palacios. All rights reserved.
...
-The birthday of a game is an ordinal that represents at which "step" the game was constructed. We
-define it recursively as the least ordinal larger than the birthdays of its left and right games. We
-prove the basic properties about these.
+There are two related but distinct notions of a birthday within combinatorial game theory. One is
+the birthday of a pre-game, which represents the "step" at which it is constructed. We define it
+recursively as the least ordinal larger than the birthdays of its left and right options. On the
+other hand, the birthday of a game is the smallest birthday among all pre-games that quotient to it.
+
+The birthday of a pre-game can be understood as representing the depth of its game tree. On the
+other hand, the birthday of a game more closely matches Conway's original description.
```
eric-wieser, line 20: Can you mention the lemma that links the two definitions?

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +161,121 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+namespace Game
 
-theorem birthday_add_nat (n : ℕ) : (a + n).birthday = a.birthday + n := by simp
+/-- The birthday of a game is defined as the least birthday among all pre-games that define it. -/
+noncomputable def birthday (x : Game.{u}) : Ordinal.{u} :=
+  sInf (PGame.birthday '' (Quotient.mk' ⁻¹' {x}))
 
-theorem birthday_nat_add (n : ℕ) : (↑n + a).birthday = a.birthday + n := by simp
+/-- The set in the definition of birthday is nonempty. -/
+theorem birthday_nonempty (x : Game) : (PGame.birthday '' (Quotient.mk' ⁻¹' {x})).Nonempty := by
```
Command-Master, line 183: Should this be a theorem? This is trivial, and you only use it once

vihdzp: The intention of this theorem is that if someone wants to work with `birthday` using the direct infimum definition, they can show that they're not in the trivial and somewhat badly behaved case of `sInf Ø`. Admittedly I'm not sure how often you'll want to do this, but even if it's never, I don't think having this theorem hurts.

Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -78,6 +78,9 @@ instance instAddCommGroupWithOneGame : AddCommGroupWithOne Game where
 instance : Inhabited Game :=
   ⟨0⟩
 
+protected theorem zero_def : (0 : Game) = ⟦0⟧ :=
+  rfl
```
YaelDillies, line 82: Does this need to be `protected`?

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -3,27 +3,33 @@ Copyright (c) 2022 Violeta Hernández Palacios. All rights reserved.
...
-define it recursively as the least ordinal larger than the birthdays of its left and right games. We
-prove the basic properties about these.
+There are two related but distinct notions of a birthday within combinatorial game theory. One is
+the birthday of a pre-game, which represents the "step" at which it is constructed. We define it
+recursively as the least ordinal larger than the birthdays of its left and right options. On the
+other hand, the birthday of a game is the smallest birthday among all pre-games that quotient to it.
+
+The birthday of a pre-game can be understood as representing the depth of its game tree. On the
+other hand, the birthday of a game more closely matches Conway's original description. The lemmas
+`SetTheory.Game.birthday_eq_pGame_birthday` and `SetTheory.Game.birthday_eq_pGame_birthday` link
```
YaelDillies, line 20: Looks like you referenced twice the same lemma

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +161,121 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+  sInf (PGame.birthday '' (Quotient.mk' ⁻¹' {x}))
 
-theorem birthday_nat_add (n : ℕ) : (↑n + a).birthday = a.birthday + n := by simp
+/-- The set in the definition of birthday is nonempty. -/
+theorem birthday_nonempty (x : Game) : (PGame.birthday '' (Quotient.mk' ⁻¹' {x})).Nonempty := by
+  rw [Set.image_nonempty]
+  exact ⟨_, x.out_eq⟩
 
-end PGame
+theorem birthday_eq_pGame_birthday (x : Game) :
```
YaelDillies, line 187:
```suggestion
theorem birthday_eq_pGameBirthday (x : Game) :
```

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +161,121 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+    ∃ y : PGame.{u}, ⟦y⟧ = x ∧ y.birthday = birthday x :=
+  csInf_mem (birthday_nonempty x)
+
+theorem birthday_le_pGame_birthday (x : PGame) : birthday ⟦x⟧ ≤ x.birthday :=
```
YaelDillies, line 191:
```suggestion
theorem birthday_quot_le_pGameBirthday (x : PGame) : birthday ⟦x⟧ ≤ x.birthday :=
```

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +161,121 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+  · intro h
+    let ⟨y, hy₁, hy₂⟩ := birthday_eq_pGame_birthday x
+    rw [← hy₁]
+    rw [h, PGame.birthday_eq_zero] at hy₂
+    exact PGame.game_eq (@PGame.Equiv.isEmpty _ hy₂.1 hy₂.2)
+  · rintro rfl
+    exact birthday_zero
+
+@[simp]
+theorem birthday_toGame (o : Ordinal) : birthday o.toGame = o := by
```
YaelDillies, line 211:
```suggestion
theorem birthday_ordinalToGame (o : Ordinal) : birthday o.toGame = o := by
```
for disambiguation, maybe?

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -157,26 +161,121 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
+  apply le_antisymm
+  · conv_rhs => rw [← PGame.birthday_toPGame o]
+    apply birthday_le_pGame_birthday
+  · let ⟨x, hx₁, hx₂⟩ := birthday_eq_pGame_birthday o.toGame
+    rw [← hx₂, ← toPGame_le_iff]
+    rw [← PGame.equiv_iff_game_eq] at hx₁
+    exact hx₁.2.trans (PGame.le_birthday x)
+
+@[simp]
```
YaelDillies, line 220:
```suggestion
@[simp, norm_cast]
```

Mathlib/SetTheory/Game/Ordinal.lean
```diff
@@ -207,4 +207,15 @@ termination_by (a, b)
 theorem toGame_nmul (a b : Ordinal) : (a ⨳ b).toGame = ⟦a.toPGame * b.toPGame⟧ :=
   Quot.sound (toPGame_nmul a b)
 
+@[simp]
```
YaelDillies, line 210:
```suggestion
@[simp, norm_cast]
```

Mathlib/SetTheory/Game/Birthday.lean
YaelDillies: In those situations, what we usually do is make the lemma `private`. The rationale is that people shouldn't this kind of lemma past the API, else there is a problem with the API.

vihdzp: I don't remember if I was the one who added them, but there do exist other lemmas like this for the ordinal API, like [`Ordinal.sub_nonempty`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Arithmetic.html#Ordinal.sub_nonempty) or [`Ordinal.div_nonempty`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Arithmetic.html#Ordinal.div_nonempty). Should those be privated in a separate PR?

Mathlib/SetTheory/Game/Basic.lean
vihdzp: Probably not.

Mathlib/SetTheory/Game/Birthday.lean
YaelDillies: Yeah probably

Mathlib/SetTheory/Game/Birthday.lean
```diff
@@ -156,26 +159,117 @@ theorem birthday_add : ∀ x y : PGame.{u}, (x + y).birthday = x.birthday ♯ y.
...
-alias birthday_nat_cast := birthday_natCast
+end PGame
 
-theorem birthday_add_nat (n : ℕ) : (a + n).birthday = a.birthday + n := by simp
+namespace Game
 
-theorem birthday_nat_add (n : ℕ) : (↑n + a).birthday = a.birthday + n := by simp
+/-- The birthday of a game is defined as the least birthday among all pre-games that define it. -/
+noncomputable def birthday (x : Game.{u}) : Ordinal.{u} :=
+  sInf (PGame.birthday '' (Quotient.mk' ⁻¹' {x}))
```
YaelDillies, line 178:
```suggestion
  sInf (PGame.birthday '' Quotient.mk' ⁻¹' {x})
```

vihdzp: This... doesn't seem to work.

YaelDillies: Uh okay

Mathlib/SetTheory/Game/Basic.lean
```diff
@@ -225,6 +228,13 @@ namespace PGame
 @[simp] theorem quot_add (a b : PGame) : ⟦a + b⟧ = (⟦a⟧ : Game) + ⟦b⟧ := rfl
 @[simp] theorem quot_sub (a b : PGame) : ⟦a - b⟧ = (⟦a⟧ : Game) - ⟦b⟧ := rfl
 
+@[simp]
+theorem quot_natCast : ∀ n : ℕ, ⟦(n : PGame)⟧ = (n : Game)
+  | 0 => rfl
+  | n + 1 => by
+    rw [PGame.nat_succ, quot_add, Nat.cast_add, Nat.cast_one, quot_natCast]
+    rfl
```
eric-wieser, line 236: You'd do better to replace the `natCast` field in the `AddCommGroupWithOne` instance above. I think your two proofs here become the `natCast_zero` and `natCast_succ` fields, and then this proof becomes `rfl`.

vihdzp: I'd rather not tweak the `NatCast` instances at this point, since there's an issue with the `PGame` instance that needs to be addressed.

The `NatCast` instance on `PGame` doesn't match the actual mathematical definition. It's directly defined as `0 + 1 + 1 + ...`, which due to how game addition works, leads to it having a left set of different relabellings of `n - 1` rather than the expected `0` through `n - 1`. 

The actual definition is given by `Ordinal.toPGame` on a natural argument. But replacing it means `(0 : N)` is no longer equal (not even prop-eq) to `0`. That is, unless, we defined it near the start of the file. Which feels like a messy solution to me.

I made a thread about this on Zulip in the #matlib4 channel, can't link it on mobile unfortunately.

vihdzp: Here's the thread: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Nat.20.E2.86.92.20PGame.20cast.2C.20again/near/469290453

vihdzp: The conclusion from that thread (or at least what I concluded) was that it's probably better to wait for `IGame` before tweaking any of these instances.

YaelDillies:
Eric's comment having been addressed,

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15721, opened 2024-08-12
[Merged by Bors] - chore(SetTheory/Ordinal/FixedPoint): Make sure simp lemma has a reasonable key

vihdzp:
This is the second most called `simp` lemma and never succeeds (see [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Infrastructure.20for.20tracking.20frequently.20applied.20simp.20theorems/near/459926416)]. It's also way too specific to be reasonably useful outside of the file itself.

Also did some pretty minor cleanup.

---

- [x] depends on: #15755


vihdzp:
!bench

Mathlib/SetTheory/Ordinal/FixedPoint.lean
```diff
@@ -368,7 +368,6 @@ def nfp (f : Ordinal → Ordinal) : Ordinal → Ordinal :=
 theorem nfp_eq_nfpFamily (f : Ordinal → Ordinal) : nfp f = nfpFamily fun _ : Unit => f :=
   rfl
 
-@[simp]
 theorem sup_iterate_eq_nfp (f : Ordinal.{u} → Ordinal.{u}) :
     (fun a => sup fun n : ℕ => f^[n] a) = nfp f := by
```
YaelDillies, line 372: Instead of unsimping, do you mind changing the statement to either
```suggestion
theorem sup_iterate_eq_nfp (f : Ordinal.{u} → Ordinal.{u}) :
    sup (fun n ↦ f^[n]) = nfp f := by
```
or
```suggestion
theorem sup_iterate_eq_nfp (f : Ordinal.{u} → Ordinal.{u}) (a : Ordinal.{u}) :
    sup (fun n ↦ f^[n] a) = nfp f a := by
```
? Possibly we want both

vihdzp: The first one is definitely better, though I still don't think it's justified in being a simp lemma. Surely the second theorem is an immediate consequence of rewriting the first, right?

YaelDillies: > I still don't think it's justified in being a simp lemma

I don't mind it not being simp, but at least it won't be a performance penalty since it will be keyed on (the rather uncommon) `Ordinal.sup`.

> Surely the second theorem is an immediate consequence of rewriting the first, right?

It is, but I can imagine you might want to rewrite an expression where eg both `nfp f a` and `nfp f b` appear, and you might not want to rewrite both. Or maybe you want to rewrite it backwards, in which case it's annoying having to call `iSup_apply` explicitly.

vihdzp: I've replaced the current funext version of `sup_iterate_eq_nfp ` by the applied version. It doesn't seem like the original was all that useful.

vihdzp: Oh, that's nice. Even more reason to do that refactor.

YaelDillies (approved):
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15723, opened 2024-08-12
[Merged by Bors] - feat(FieldTheory): add lemmas about minpoly

mariainesdff:

---

### DIFF
--- a/Mathlib/FieldTheory/Minpoly/IsIntegrallyClosed.lean
+++ b/Mathlib/FieldTheory/Minpoly/IsIntegrallyClosed.lean
@@ -186,4 +186,24 @@ theorem _root_.PowerBasis.ofGenMemAdjoin'_gen (B : PowerBasis R S) (hint : IsInt
 
 end AdjoinRoot
 
+section Subring
+
+variable {K L : Type*} [Field K] [Field L] [Algebra K L]
+
+variable (A : Subring K) [IsIntegrallyClosed A] [IsFractionRing A K]
+
+-- Implementation note: `inferInstance` does not work for these.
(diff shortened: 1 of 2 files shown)


Mathlib/FieldTheory/Minpoly/IsIntegrallyClosed.lean
```diff
@@ -186,4 +186,24 @@ theorem _root_.PowerBasis.ofGenMemAdjoin'_gen (B : PowerBasis R S) (hint : IsInt
 
 end AdjoinRoot
 
+section Subring
+
+variable {K L : Type*} [Field K] [Field L] [Algebra K L]
+
+variable (A : Subring K) [IsIntegrallyClosed A] [IsFractionRing A K]
```
kbuzzard, line 193: Why is `A` a subring of `K` here? In all uses below you treat it as a type. Wouldn't it be more general to have `(A : Type*) [CommRing A] [Algebra A K] [Algebra A L] [IsScalarTower A K L]`?

mariainesdff: Because we want to apply this lemma when `A` is the valuation subring of a valued field K (see [here](https://github.com/mariainesdff/LocalClassFieldTheory/blob/df56df16dcdb067fd536758dea68e27b79fd8488/LocalClassFieldTheory/DiscreteValuationRing/DiscreteNorm.lean#L185); this is the only place so where we have used this lemma in our local field project).

kbuzzard:
Thanks and sorry for the hold-up!

bors merge

PR #15724, opened 2024-08-12
test: the `assertNotExists` style linter

adomani:
The "assertNotExists" style linter checks that a file starts with
```lean
import*
doc-module*
assert_not_exists*
[no more assert_not_exists]
```
It emits a warning on each `assert_not_exists` that is not preceded by
* possibly some `import` statements,
* possibly some doc-module strings, and
* possibly some `assert_not_exists` commands
in this order.

---


eric-wieser (commented):
In lean3 we had an `assertNotExists` linter that detected whether the definitions actually exist elsewhere in mathlib. Can we name this one something else, to disambiguate? Perhaps `linters.style.assertNotExists` etc?

adomani:
Eric, your comment made me think that maybe the implementation of `assert_not_exists` could include a persistent change to the environment that an associated linter could then pick up.  Looking at the lean3 code that you mentioned seems like what the lean3 linter was doing as well.

I'll see if I can implement it like this, thanks!

adomani:
Michael, let me try to implement the doc-module *style* linter.  It should be relatively straightforward to make it so that it flags an improperly formatted file when it reaches the `eoi` syntax.  This means that all the file would have to be processed before you get a warning.  Otherwise, it is a little tricky to make sure that it only process the "preamble" once.

adomani:
Oh, I forgot to mention: I inserted `Style` after `Linter` in the namespace: hopefully this makes the intent of this linter clearer.

grunweg:
> Oh, I forgot to mention: I inserted `Style` after `Linter` in the namespace: hopefully this makes the intent of this linter clearer.

You could (additionally) refer to this as the "`assert_not_exists` style" linter, perhaps that would make it *even* clearer. (The style namespace already helps!)

grunweg:
> Eric, your comment made me think that maybe the implementation of `assert_not_exists` could include a persistent change to the environment that an associated linter could then pick up. Looking at the lean3 code that you mentioned seems like what the lean3 linter was doing as well.
> 
> I'll see if I can implement it like this, thanks!

That would be great, and sounds like the best way forward!

Mathlib/Tactic/Linter/AssertNotExists.lean
```diff
@@ -0,0 +1,101 @@
...
+[no more `assert_not_exists`]
+```
+It emits a warning on each `assert_not_exists` that is not preceded by
+* possibly some `import` statements,
+* possibly some doc-module strings, and
+* possibly some `assert_not_exists` commands
+
+in this order.
+-/
+register_option linter.assertNotExists : Bool := {
```
eric-wieser, line 74:
```suggestion
register_option linter.style.assertNotExists : Bool := {
```

Mathlib/Tactic/Linter/AssertNotExists.lean
```diff
@@ -0,0 +1,101 @@
...
+  defValue := true
+  descr := "enable the assertNotExists linter"
+}
+
+namespace Style.AssertNotExists
+
+/-- Gets the value of the `linter.assertNotExists` option. -/
+def getLinterHash (o : Options) : Bool := Linter.getLinterValue linter.assertNotExists o
```
eric-wieser, line 82: What does `Hash` stand for here?

adomani: I copied the name from the first linter that I wrote, probably using code that Mario wrote, created a template for linters and never went back to think about it again.

I do not know what it stands for... :shrug:

eric-wieser: I suspect this is left from the hash command linter?

adomani: #15771

grunweg:
To record what happened on zulip: it seems sensible to require `assert_not_imported` to come right before `assert_not_exists` commands.

PR #15715, opened 2024-08-12
[Merged by Bors] - chore: uncomment some now viable assert_not_exists

kim-em:
Per [zulip](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/.60assert_not_exists.60.20at.20top.20or.20bottom.20of.20file.3F/near/459960523) discussion.

PR #15727, opened 2024-08-12
chore: move some `assert_not_exists`s at the beginning of the file

adomani:
This is in preparation for a linter that will enforce `assert_not_exists`s to be placed right after the initial doc-module strings.

I also guarded against the downstream-existing `ModuleCat.abelian`, instead of the nowhere-existing `Module.abelian`.

---

- [x] depends on: #15715


adomani:
This PR was essentially merged as #15747, hence I am closing it.

PR #15732, opened 2024-08-12
[Merged by Bors] - chore: use down-stream existing declarations when `assert_not_exists`ing them

adomani:
Another PR stemming from the future `assertNotExists` linter.

---


Mathlib/Algebra/Field/Defs.lean
```diff
@@ -47,7 +47,7 @@ field, division ring, skew field, skew-field, skewfield
 assert_not_exists NeZero
 
 -- Check that we have not imported `Mathlib.Tactic.Common` yet.
-assert_not_exists Mathlib.Tactic.LibrarySearch.librarySearch
+assert_not_exists Mathlib.Tactic.scopedNS
```
adomani, line 50: I wonder whether there should be an `assert_not_imported` command for this sort of checks.

grunweg: So do I - but I also think this can be a follow-up change. (Would you like to create a zulip thread or file an issue to record this? That is not something I will insist on.)

Mathlib/Data/Nat/Cast/Basic.lean
```diff
@@ -23,8 +23,9 @@ assert_not_exists OrderedCommGroup
 assert_not_exists Commute.zero_right
 assert_not_exists Commute.add_right
 assert_not_exists abs_eq_max_neg
-assert_not_exists natCast_ne
-assert_not_exists MulOpposite.natCast
+assert_not_exists NeZero.natCast_ne
+-- `MulOpposite.op_natCast` was not intended to be imported
```
grunweg, line 27: So this `assert_not_exists` fails now, right? Can you make the comment mention "TODO" somewhere?

grunweg (approved):
Thanks for doing this, looks great. Two very minor comments - feel free to maintainer merge on my behalf with those addressed.

Mathlib/Data/Nat/Cast/Basic.lean
adomani: TODO added, thanks!

Mathlib/Algebra/Field/Defs.lean
adomani: I started [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/assert_not_imported) about this issue.

adomani (commented):
Thanks for the review!

grunweg:
Thanks!
maintainer merge

mattrobball:
bors merge

grunweg:
Actually: would you also like to replace `Module.abelian` by `ModuleCat.abelian` (picked from #15727)?

PR #15733, opened 2024-08-12
[Merged by Bors] - feat(Algebra/Ring/Subring/IntPolynomiall): add intPolynomial

mariainesdff:

---


jcommelin (commented):
Am I right that this entire file can be generalized from `v.valuationSubring` to an arbitrary `S : Subring K`?
The name could suggestively remain the same... or maybe something like `subPolynomial` or `resPolynomial`. But I don't really like either of those...

faenuccio (commented):
Good point, I generalize everything to an arbitrary subring, keeping the same name. After some thoughts, I believe that keeping `R` explicit is better, otherwise I believe that typeclass inference would struggle too much.

jcommelin (commented):
What do you think of:

1. Moving this file to some folder about (sub)rings? (If so, please update the PR title)
2. Renaming to `Polynomial.int`, and making `P` explicit, so that dot-notation can be used?

faenuccio (commented):
I am happy with 1. but concerning 2. I am a bit sceptical. The hypothesis `hP` that all coefficients are in `R` would be needed anyhow, so unless we create a structure `intPolynomial` we would at any rate end up with something like `P.int R hP`, and at this point I do not see the advantage of dot notation.

jcommelin:
One advantage is that the total expression is shorter than `intPolynomial R hp`. And also, the user is visually reminded of the `P`, which would otherwise only be available via hovering etc...

mariainesdff:
> What do you think of:
> 
>     1. Moving this file to some folder about (sub)rings? (If so, please update the PR title)
> 
>     2. Renaming to `Polynomial.int`, and making `P` explicit, so that dot-notation can be used?

I have implemented both changes.

Mathlib/Algebra/Ring/Subring/IntPolynomial.lean
```diff
@@ -0,0 +1,57 @@
...
+variable {K : Type*} [Field K] (R : Subring K)
+
+open Polynomial
+
+open scoped Polynomial
+
+/-- Given a polynomial in `K[X]` such that all coefficients belong to the subring `R`,
+  `intPolynomial` is the corresponding polynomial in `R[X]`. -/
+def Polynomial.int (P : K[X]) (hP : ∀ n : ℕ, P.coeff n ∈ R) : R[X]
+    where toFinsupp :=
```
jcommelin, line 31:
```suggestion
def Polynomial.int (P : K[X]) (hP : ∀ n : ℕ, P.coeff n ∈ R) : R[X] where
  toFinsupp :=
```

Mathlib/Algebra/Ring/Subring/IntPolynomial.lean
```diff
@@ -0,0 +1,57 @@
...
+theorem int_leadingCoeff_eq : ↑(P.int R hP).leadingCoeff = P.leadingCoeff := rfl
+
+theorem int_monic_iff : (P.int R hP).Monic ↔ P.Monic := by
+  rw [Monic, Monic, ← int_leadingCoeff_eq, OneMemClass.coe_eq_one]
+
+theorem int_natDegree : (P.int R hP).natDegree = P.natDegree := rfl
+
+variable {L : Type*} [Field L] [Algebra K L]
+
+theorem int_eval₂_eq (x : L) :
```
jcommelin, line 52:
```suggestion
@[simp]
theorem int_coeff_eq  (n : ℕ) : ↑((P.int R hP).coeff n) = P.coeff n := rfl

@[simp]
theorem int_leadingCoeff_eq : ↑(P.int R hP).leadingCoeff = P.leadingCoeff := rfl

@[simp]
theorem int_monic_iff : (P.int R hP).Monic ↔ P.Monic := by
  rw [Monic, Monic, ← int_leadingCoeff_eq, OneMemClass.coe_eq_one]

@[simp]
theorem int_natDegree : (P.int R hP).natDegree = P.natDegree := rfl

variable {L : Type*} [Field L] [Algebra K L]

@[simp]
theorem int_eval₂_eq (x : L) :
```

jcommelin (commented):
Thnanks!

bors d+

PR #15737, opened 2024-08-12
CI: add check for assert_not_exists

adomani:
Add a CI step to make sure that if `assert_not_exists X` appears in a file, then `X` is actually defined somewhere in `Mathlib`.

Until #15732 is merged, the test should fail, since there are currently declarations in `assert_not_exists` that not only do not exist in the file, but do not exist at all.

---

- [ ] depends on: #15732 [optional extra text]


adomani:
This has been implemented adding an environment extension tracking the assertions and collecting them in `test/AssertImported.lean`.

PR #15741, opened 2024-08-12
[Merged by Bors] - feat(Data/Int/WithZero): add WithZeroMultIntToNNReal

mariainesdff:
We define  the morphism `withZeroMultIntToNNReal`  from `ℤₘ₀ → ℝ≥0` sending `0 ↦ 0` and  `x ↦ e^(Multiplicative.toAdd (WithZero.unzero hx)` when `x ≠ 0`, for a nonzero `e : ℝ≥0` and prove some of its properties.

---


Mathlib/Data/Int/WithZero.lean
```diff
@@ -0,0 +1,89 @@
...
+WithZero, multiplicative, nnreal
+-/
+
+noncomputable section
+
+open scoped NNReal
+
+open Multiplicative WithZero
+
+local notation "ℤₘ₀" => WithZero (Multiplicative ℤ)
```
jcommelin, line 35: The names of these theorems become pretty long. Would it make sense to make this notation into an `abbrev` and then use the `WithZeroMulInt` namespace. Then the morphism below would become `WithZeroMulInt.toNNReal`.

mariainesdff: Done!

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/Data/Int/WithZero.lean
```diff
@@ -0,0 +1,94 @@
...
+noncomputable section
+
+open scoped NNReal
+
+open Multiplicative WithZero
+
+/-- The type `WithZero (Multiplicative ℤ)`. -/
+abbrev WithZeroMulInt := WithZero (Multiplicative ℤ)
+
+local notation "ℤₘ₀" => WithZeroMulInt
```
YaelDillies, line 38: This notation is already defined by this point (you might have to merge master)
```suggestion
```

sgouezel:
Can you merge master, fix the build, and then send it back to bors?
bors d+

PR #15743, opened 2024-08-12
[Merged by Bors] - feat(NumberTheory/JacobiSum/Basic): add file (def., basic API)

MichaelStollBayreuth:
This adds a new folder `NumberTheory/JacobiSum` and a first file `Basic.lean` in it, which defines the *Jacobi sum* of two multiplicative characters (on a finite commutative ring with values in another commutative ring) and provides some first results.

---

### DIFF
--- a/Mathlib/NumberTheory/JacobiSum/Basic.lean
+++ b/Mathlib/NumberTheory/JacobiSum/Basic.lean
@@ -1,0 +1,120 @@
+/-
+Copyright (c) 2024 Michael Stoll. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Michael Stoll
+-/
+import Mathlib.NumberTheory.MulChar.Basic
+import Mathlib.Algebra.Module.BigOperators
+
+/-!
+# Jacobi Sums
+
+This file defines the *Jacobi sum* of two multiplicative characters `χ` and `ψ` on a finite
+commutative ring `R` with values in another commutative ring `R'`:
+
+`jacobiSum χ ψ = ∑ x : R, χ x * ψ (1 - x)`
+
+(see `jacobiSum`) and provides some basic results and API lemmas on Jacobi sums.
+
+## References
+
+We essentially follow
+* [K. Ireland, M. Rosen, *A classical introduction to modern number theory*
+   (Section 8.3)][IrelandRosen1990]
+
+but generalize where appropriate.
+
+This is based on Lean code written as part of the bachelor's thesis of Alexander Spahl.
+-/
+
+open BigOperators Finset
+
+/-!
+### Jacobi sums: definition and first properties
+-/
+
+section Def
+
+namespace Finset
+
+-- Is this useful enough to be non-private?
+private lemma sum_eq_sum_one_sub {R M : Type*} [Ring R] [Fintype R] [AddCommMonoid M] (f : R → M) :
+    Finset.sum univ f = Finset.sum univ fun x ↦ f (1 - x) := by
+  refine Fintype.sum_bijective (1 - ·) (Function.Involutive.bijective ?_) _ _ fun x ↦ ?_
+  · simp only [Function.Involutive, sub_sub_cancel, implies_true]
+  · simp only [sub_sub_cancel]
+
+end Finset
+
+-- need `Fintype` instead of `Finite` for `Finset.sum` etc.
+variable {R R' : Type*} [CommRing R] [Fintype R] [CommRing R']
+
+/- The *Jacobi sum* of two multiplicative characters on a finite commutative ring. -/
+def jacobiSum (χ ψ : MulChar R R') : R' :=
+  ∑ x : R, χ x * ψ (1 - x)
+
+lemma jacobiSum_comm (χ ψ : MulChar R R') : jacobiSum χ ψ = jacobiSum ψ χ := by
+  simp only [jacobiSum]
+  convert sum_eq_sum_one_sub fun x ↦ χ x * ψ (1 - x) using 2 with x
+  simp only [mul_comm, sub_sub_cancel]
+
+/-- The Jacobi sum is compatible with ring homomorphisms. -/
+lemma jacobiSum_ringHomComp {R'' : Type*} [CommRing R''] (χ ψ : MulChar R R') (f : R' →+* R'') :
+    jacobiSum (χ.ringHomComp f) (ψ.ringHomComp f) = f (jacobiSum χ ψ) := by
+  simp only [jacobiSum, MulChar.ringHomComp, MulChar.coe_mk, MonoidHom.coe_mk, OneHom.coe_mk,
+    map_sum, map_mul]
+
+end Def
+
+/-!
+### Jacobi sums over finite fields
+-/
+
+section FiniteField
+
+variable {F R : Type*} [Field F] [Fintype F] [DecidableEq F] [CommRing R]
+
+/-- The Jacobi sum of two multiplicative characters on a finite field `F` can be written
+as a sum over `F \ {0,1}`. -/
+lemma jacobiSum_eq_sum_sdiff (χ ψ : MulChar F R) :
+    jacobiSum χ ψ = ∑ x ∈ univ \ {0,1}, χ x * ψ (1 - x) := by
+  simp only [jacobiSum, subset_univ, sum_sdiff_eq_sub, mem_singleton, zero_ne_one,
+    not_false_eq_true, sum_insert, isUnit_iff_ne_zero, ne_eq, not_true_eq_false,
+    MulCharClass.map_nonunit, sub_zero, map_one, mul_one, sum_singleton, sub_self, mul_zero,
+    add_zero]
+
+private lemma jacobiSum_eq_aux (χ ψ : MulChar F R) :
+    jacobiSum χ ψ = ∑ x : F, χ x + ∑ x : F, ψ x - Fintype.card F +
+                      ∑ x ∈ univ \ {0, 1}, (χ x - 1) * (ψ (1 - x) - 1) := by
+  rw [jacobiSum]
+  conv =>
+    enter [1, 2, x]
+    rw [show ∀ x y : R, x * y = x + y - 1 + (x - 1) * (y - 1) by intros; ring]
+  rw [sum_add_distrib, sum_sub_distrib, sum_add_distrib, ← sum_eq_sum_one_sub,
+    Fintype.card_eq_sum_ones, Nat.cast_sum, Nat.cast_one, sum_sdiff_eq_sub (subset_univ _),
+    ← sub_zero (_ - _ + _), add_sub_assoc]
+  congr
+  rw [sum_pair zero_ne_one, sub_zero, ψ.map_one, χ.map_one, sub_self, mul_zero, zero_mul, add_zero]
+
+/-- If `1` is the trivial multiplicative character on a finite field `F`, then `J(1,1) = #F-2`. -/
+theorem jacobiSum_triv_triv : jacobiSum (1 : MulChar F R) 1 = Fintype.card F - 2 := by
+  rw [show 1 = MulChar.trivial F R from rfl, jacobiSum_eq_sum_sdiff]
+  have : ∀ x ∈ univ \ {0, 1}, (MulChar.trivial F R) x * (MulChar.trivial F R) (1 - x) = 1 := by
+    intros x hx
+    have hx' : IsUnit (x * (1 - x)) := by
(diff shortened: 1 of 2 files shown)


Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+/-!
+### Jacobi sums: definition and first properties
+-/
+
+section Def
+
+namespace Finset
+
+-- Is this useful enough to be non-private?
+private lemma sum_eq_sum_one_sub {R M : Type*} [Ring R] [Fintype R] [AddCommMonoid M] (f : R → M) :
```
kbuzzard, line 41: Maybe put the more complicated term on the LHS of the inequality?

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+    Finset.sum univ f = Finset.sum univ fun x ↦ f (1 - x) := by
+  refine Fintype.sum_bijective (1 - ·) (Function.Involutive.bijective ?_) _ _ fun x ↦ ?_
+  · simp only [Function.Involutive, sub_sub_cancel, implies_true]
+  · simp only [sub_sub_cancel]
+
+end Finset
+
+-- need `Fintype` instead of `Finite` for `Finset.sum` etc.
+variable {R R' : Type*} [CommRing R] [Fintype R] [CommRing R']
```
kbuzzard, line 50: I thought that the fashion now was `[Finite R]`?

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+
+open BigOperators Finset
+
+/-!
+### Jacobi sums: definition and first properties
+-/
+
+section Def
+
+namespace Finset
```
kbuzzard, line 38: You could just `lemma Finset.sum_eq_sum_one_sub` I guess

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,120 @@
...
+  rw [show 1 = MulChar.trivial F R from rfl, jacobiSum_eq_sum_sdiff]
+  have : ∀ x ∈ univ \ {0, 1}, (MulChar.trivial F R) x * (MulChar.trivial F R) (1 - x) = 1 := by
+    intros x hx
+    have hx' : IsUnit (x * (1 - x)) := by
+      simp only [mem_sdiff, mem_univ, mem_insert, mem_singleton, not_or, ← ne_eq, true_and] at hx
+      simp only [isUnit_iff_ne_zero]
+      exact mul_ne_zero hx.1 <| sub_ne_zero.mpr hx.2.symm
+    rw [← map_mul, MulChar.trivial_apply, if_pos hx']
+  calc ∑ x ∈ univ \ {0, 1}, (MulChar.trivial F R) x * (MulChar.trivial F R) (1 - x)
+  _ = ∑ _ ∈ @univ F _ \ {0, 1}, (1 : R) := sum_congr rfl this
```
kbuzzard, line 110: Maybe
```suggestion
  _ = ∑ _ ∈ (univ \ {0, 1} : Finset F), (1 : R) := sum_congr rfl this
```
? Gets rid of the @ at the expense of more characters...

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: I think I tried that, and then `Finset.sum` complained...

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: Good point. The proof doesn't even need to change...

MichaelStollBayreuth: Actually, it even gets simpler!

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: OK.

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: I think it is because `Finset.univ` requires a `Fintype` instance.

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: (Still keeping it `private`, though.)

Mathlib/NumberTheory/JacobiSum/Basic.lean
kbuzzard: `attribute [local instance] Fintype.ofFinite` fixes this (or `letI := Fintype.ofFinite F` in lots of places)

MichaelStollBayreuth: My understanding is that when it is needed for a statement or definition, requiring `Fintype` should be OK, and if it is just needed in a proof, requiring `Finite` is preferred (and then using `Fintype.ofFinite` in the proof); by analogy with `Decidable`. If desired, one can always use `Fintype.ofFinite` in either way when using the results.

Note that using `Finite` instead of `Fintype`  makes `jacobiSum` noncomputable and does not allow proofs like
```lean
example : jacobiSum ZMod.χ₄ ZMod.χ₄ = 0 := rfl
```
or
```lean
example : jacobiSum ZMod.χ₄ ZMod.χ₄ = 0 := by decide
```

So I think keeping `Fintype` is better here.

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,116 @@
...
+
+open BigOperators Finset
+
+/-!
+### Jacobi sums: definition and first properties
+-/
+
+section Def
+
+-- Is this useful enough to be non-private?
```
jcommelin, line 38: I think so. And I guess you can use `#find_home` to find a better place for it. And maybe generalize `1` to an arbitrary element `r` of `R`?
It could also be a `simp`-lemma, I think.

And if you like it, the version with `x + r` should be just as short, and can maybe be added for completeness?

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,116 @@
...
+variable {R R' : Type*} [CommRing R] [Fintype R] [CommRing R']
+
+/-- The *Jacobi sum* of two multiplicative characters on a finite commutative ring. -/
+def jacobiSum (χ ψ : MulChar R R') : R' :=
+  ∑ x : R, χ x * ψ (1 - x)
+
+lemma jacobiSum_comm (χ ψ : MulChar R R') : jacobiSum χ ψ = jacobiSum ψ χ := by
+  simp only [jacobiSum]
+  convert (sum_one_sub_eq_sum fun x ↦ χ x * ψ (1 - x)).symm using 2 with x
+  simp only [mul_comm, sub_sub_cancel]
```
jcommelin, line 55:
```suggestion
  simp only [jacobiSum, mul_comm (χ _)]
  rw [← sum_one_sub_eq_sum]
  simp only [sub_sub_cancel]
```

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,116 @@
...
+    enter [1, 2, x]
+    rw [show ∀ x y : R, x * y = x + y - 1 + (x - 1) * (y - 1) by intros; ring]
+  rw [sum_add_distrib, sum_sub_distrib, sum_add_distrib, sum_one_sub_eq_sum,
+    Fintype.card_eq_sum_ones, Nat.cast_sum, Nat.cast_one, sum_sdiff_eq_sub (subset_univ _),
+    ← sub_zero (_ - _ + _), add_sub_assoc]
+  congr
+  rw [sum_pair zero_ne_one, sub_zero, ψ.map_one, χ.map_one, sub_self, mul_zero, zero_mul, add_zero]
+
+/-- If `1` is the trivial multiplicative character on a finite field `F`, then `J(1,1) = #F-2`. -/
+theorem jacobiSum_triv_triv : jacobiSum (1 : MulChar F R) 1 = Fintype.card F - 2 := by
```
jcommelin, line 96:
```suggestion
theorem jacobiSum_one_one : jacobiSum (1 : MulChar F R) 1 = Fintype.card F - 2 := by
```
would it make sense to also have the version that syntactically mentions `MulChar.trivial`?

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,116 @@
...
+  rw [show 1 = MulChar.trivial F R from rfl, jacobiSum_eq_sum_sdiff]
+  have : ∀ x ∈ univ \ {0, 1}, (MulChar.trivial F R) x * (MulChar.trivial F R) (1 - x) = 1 := by
+    intros x hx
+    have hx' : IsUnit (x * (1 - x)) := by
+      simp only [mem_sdiff, mem_univ, mem_insert, mem_singleton, not_or, ← ne_eq, true_and] at hx
+      simp only [isUnit_iff_ne_zero]
+      exact mul_ne_zero hx.1 <| sub_ne_zero.mpr hx.2.symm
+    rw [← map_mul, MulChar.trivial_apply, if_pos hx']
```
jcommelin, line 104:
```suggestion
    rw [← map_mul, MulChar.trivial_apply, if_pos]
    simp only [mem_sdiff, mem_univ, mem_insert, mem_singleton, not_or, ← ne_eq, true_and] at hx
    simpa only [isUnit_iff_ne_zero, mul_ne_zero_iff, ne_eq, sub_eq_zero, @eq_comm _ _ x] using hx
```

Mathlib/NumberTheory/JacobiSum/Basic.lean
```diff
@@ -0,0 +1,116 @@
...
+  calc ∑ x ∈ univ \ {0, 1}, (MulChar.trivial F R) x * (MulChar.trivial F R) (1 - x)
+  _ = ∑ _ ∈ univ \ {0, 1}, 1 := sum_congr rfl this
+  _ = Finset.card (univ \ {0, 1}) := (cast_card _).symm
+  _ = Fintype.card F - 2 := by
+    rw [card_sdiff (subset_univ _), card_univ, card_pair zero_ne_one]
+    obtain ⟨m, hm⟩ : ∃ m : ℕ, Fintype.card F = 1 + m + 1 :=
+      Nat.exists_eq_add_of_lt Fintype.one_lt_card
+    rw [show 1 + m + 1 = m + 2 by ring] at hm
+    simp only [hm, add_tsub_cancel_right (α := ℕ), Nat.cast_add, Nat.cast_ofNat,
+      add_sub_cancel_right]
```
jcommelin, line 114: Maybe a bit hacky to abuse `lt/le` like this:
```suggestion
    obtain ⟨m, hm⟩ : ∃ m : ℕ, Fintype.card F = m + 2 :=
      Nat.exists_eq_add_of_le' Fintype.one_lt_card
    rw [hm, add_tsub_cancel_right]
    simp only [Nat.cast_add, Nat.cast_ofNat, add_sub_cancel_right]
```
I guess we should just have `Fintype.two_le_card`.

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: Further inspection shows that this is (up to DefEq / `Equiv.subLeft_apply`) just `(Equiv.subLeft 1).sum_comp f`, so I'm now opting for inlining it.

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: OK; added the version using `MulChar.trivial` as well.

Mathlib/NumberTheory/JacobiSum/Basic.lean
MichaelStollBayreuth: Golfed this a bit further using `Nat.cast_sub <| Nat.add_one_le_of_lt Fintype.one_lt_card`.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15745, opened 2024-08-12
[Merged by Bors] - chore(SetTheory/Ordinal/NaturalOps): cleanup termination_by

vihdzp:
We simplify some of the proofs and definitions that use `termination_by` by moving the arguments before the colon.

The diff might seem large, but all I really did besides that was reindent proofs.

---


Ruben-VandeVelde (approved):
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15757, opened 2024-08-13
[Merged by Bors] - chore(Data/Nat/Factorial): replace all `induction'` by `induction`

nmouha:
---


dupuisf:
Looks good, thanks!

bors r+

PR #15760, opened 2024-08-13
[Merged by Bors] - feat: convenience lemmas for replacing the uniformity and bornology via bilipschitz equivalence

j-loreaux:
---


Mathlib/Topology/MetricSpace/Bilipschitz.lean
```diff
@@ -0,0 +1,95 @@
...
+import Mathlib.Topology.MetricSpace.Lipschitz
+
+/-! # Bilipschitz equivalence
+
+A common pattern in Mathlib is to replace the topology, uniformity and bornology on a type
+synonym with those of the underlying type.
+
+The most common way to do this is to activate a local instance for something which puts a
+`PseudoMetricSpace` structure on the type synonym, prove that this metric is bilipschitz equivalent
+to the metric on th eunderlying type, and then use this to show that the uniformities and
```
eric-wieser, line 16:
```suggestion
to the metric on the underlying type, and then use this to show that the uniformities and
```

Mathlib/Topology/MetricSpace/Bilipschitz.lean
```diff
@@ -0,0 +1,95 @@
...
+-/
+
+open NNReal
+
+section Uniformity
+
+open Uniformity
+
+variable {α β : Type*} [PseudoEMetricSpace α] [PseudoEMetricSpace β]
+variable {K₁ K₂ : ℝ≥0} {f : α → β} (hf₁ : AntilipschitzWith K₁ f) (hf₂ : LipschitzWith K₂ f)
```
eric-wieser, line 37: Should we have `BilipschitzWith` for convenience?

j-loreaux: maybe, but I don't want to add it yet because I'm not ready to develop the API for it. When we get more use cases, then sure. (There certainly are many)

dupuisf (commented):
Looks good!

bors d+

PR #15762, opened 2024-08-13
[Merged by Bors] - chore: delete `Init.Data.Fin.Basic`

Parcly-Taxel:
---


Mathlib/Init/Data/Fin/Basic.lean
```diff
@@ -1,34 +0,0 @@
...
-variable {n : ℕ} {i j : Fin n}
-
-@[deprecated eq_of_val_eq (since := "2024-02-15")]
-theorem eq_of_veq : i.val = j.val → i = j := eq_of_val_eq
-
-@[deprecated val_eq_of_eq (since := "2024-02-15")]
-theorem veq_of_eq : i = j → i.val = j.val := val_eq_of_eq
-
--- These two aren't deprecated because `ne_of_val_ne` and `val_ne_of_ne`
--- use `¬a = b` instead of `a ≠ b`. TODO: fix or rename in Lean core.
```
jcommelin, line 30: This comment has not been addressed. But seemingly this syntactic difference does not cause any problems in mathlib.

Mathlib/Data/List/Nodup.lean
```diff
@@ -3,9 +3,9 @@ Copyright (c) 2018 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro, Kenny Lau
 -/
+import Mathlib.Data.Fin.Basic
```
jcommelin, line 6: Why do you need this import? The init file that was imported before only has deprecated stuff in it. So this file should be able to work with the core versions, right?

Parcly-Taxel: Fixed.

Mathlib/Init/Data/Fin/Basic.lean
eric-wieser: Can we open a PR to core that changes these two lemmas to use `≠`? That seems uncontroversial to me.

Parcly-Taxel: @eric-wieser https://github.com/leanprover/lean4/pull/5011

semorrison: I've merged this now.

semorrison:
bors merge

PR #15763, opened 2024-08-13
[Merged by Bors] - feat: add the `WithCStarModule` type synonym

j-loreaux:
It is often the case that we want to construct a `CStarModule` instance on a type that is already
endowed with a norm, and it has a different norm which makes it into a `CStarModule`. For this
reason, we create a type synonym `WithCStarModule` which is endowed with the requisite
`CStarModule` instance. We also introduce the scoped notation `C⋆ᵐᵒᵈ` for this type synonym.

The common use cases are, when `A` is a C⋆-algebra:

+ `A` itself over `A`
+ `E × F` where `E` and `F` are `CStarModule`s over `A`
+ `Π i, E i` where `E i` is a `CStarModule` over `A` and `i : ι` with `ι` a `Fintype`
+ `E` where `E` is an `InnerProductSpace` over `ℂ`

The `WithCStarModule` synonym is of vital importance, especially because the `CStarModule` class
marks `A` as an `outParam`. Indeed, we want to infer `A` from the type of `E`, but, as with modules,
a type `E` can be a `CStarModule` over different C⋆-algebras. For example, note that if `A` is a
C⋆-algebra, then so is `A × A`, and therefore we may consider both `A` and `A × A` as `CStarModule`s
over themselves, respectively. However, we may *also* consider `A × A` as a `CStarModule` over `A`.
However, by utilizing the type synonym, these actually correspond to *different types*, namely:

+ `A` as a `CStarModule` over `A` corresponds to `C⋆ᵐᵒᵈ A`
+ `A × A` as a `CStarModule` over `A × A` corresponds to `C⋆ᵐᵒᵈ (A × A)`
+ `A × A` as a `CStarModule` over `A` corresponds to `C⋆ᵐᵒᵈ (C⋆ᵐᵒᵈ A × C⋆ᵐᵒᵈ A)`

---

The above is copied from the module documentation.


Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
```diff
@@ -0,0 +1,300 @@
...
+instance [u : UniformSpace E] : UniformSpace (C⋆ᵐᵒᵈ E) := u.comap <| equiv E
+
+instance [Bornology E] : Bornology (C⋆ᵐᵒᵈ E) := Bornology.induced <| equiv E
+
+/-- `WithCStarModule.equiv` as a uniform equivalence between `C⋆ᵐᵒᵈ E` and `E`. -/
+def uniformEquiv [UniformSpace E] : C⋆ᵐᵒᵈ E ≃ᵤ E := equiv E |>.toUniformEquivOfUniformInducing ⟨rfl⟩
+
+instance [UniformSpace E] [CompleteSpace E] : CompleteSpace (C⋆ᵐᵒᵈ E) :=
+  uniformEquiv.completeSpace_iff.mpr inferInstance
+
```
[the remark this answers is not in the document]
j-loreaux: I think this is not necessary because we will always be constructing a `NormedAddCommGroup` instance on the `WithCStarModule E` type, and these tools aren't needed to make that happen. And once we get that, we'll get these for free.

dupuisf:
Looks good. I'll delegate in case you had some last-minute changes to make.

bors d+

Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
```diff
@@ -0,0 +1,320 @@
...
+Authors: Jireh Loreaux
+-/
+import Mathlib.RingTheory.Finiteness
+import Mathlib.Topology.Bornology.Constructions
+import Mathlib.Topology.UniformSpace.Equiv
+
+/-! # Type synonym for types with a `CStarModule` structure
+
+It is often the case that we want to construct a `CStarModule` instance on a type that is already
+endowed with a norm, and it has a different norm which makes it into a `CStarModule`. For this
```
eric-wieser, line 13: I'm a little confused; does this file actually provide such a norm? If not, can you make that clear in the docstring?

j-loreaux: No, although it's always given by `√⟪x, x⟫_A` once you have `A`-valued inner product defined on your type.

You can have a look at #15763 to see how we use this type synonym.

j-loreaux:
bors merge

Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
eric-wieser: You just linked this same PR?

Mathlib/Analysis/CStarAlgebra/Module/Synonym.lean
```diff
@@ -0,0 +1,320 @@
...
+import Mathlib.RingTheory.Finiteness
+import Mathlib.Topology.Bornology.Constructions
+import Mathlib.Topology.UniformSpace.Equiv
+
+/-! # Type synonym for types with a `CStarModule` structure
+
+It is often the case that we want to construct a `CStarModule` instance on a type that is already
+endowed with a norm, but this norm is not the one associated to its `CStarModule` structure. For
+this reason, we create a type synonym `WithCStarModule` which is endowed with the requisite
+`CStarModule` instance. We also introduce the scoped notation `C⋆ᵐᵒᵈ` for this type synonym.
```
eric-wieser, line 15: I'm still very confused here; where is it "endowed with the requisite
`CStarModule` instance"? A type synonym with no new instances strikes me as quite weird.

PR #15766, opened 2024-08-13
[Merged by Bors] - chore(SetTheory/Game/PGame): simplify ofLists API

vihdzp:
The equation `(ofLists L R).LeftMoves = ULift (Fin L.length)` (ditto for `RightMoves`) is entirely def-eq in Lean 4. As such, `toOfListsLeftMoves` doesn't solve the same problem that other `toXLeftMoves` equivalences do, which is provide this correspondence without having to tediously figure out how to rewrite the equality. We instead turn this into an abbreviation for `ULift.up` and `ULift.down`, which has the nice side effect of making `simp` much more capable of simplifying equations involving `ofLists`.

---

To be clear, I don't recall if this was def-eq in Lean 3. I imagine not, as we probably wouldn't have been using `Equiv.cast` otherwise.


jcommelin (commented):
Thanks :tada:

bors merge

PR #15770, opened 2024-08-13
[Merged by Bors] - feat(CategoryTheory/Limits): epi/mono in functor categories

joelriou:
In this file, we obtain new versions of lemmas `NatTrans.mono_iff_mono_app` (resp. `NatTrans.epi_iff_epi_app`) when the category has pullbacks (resp. pushouts). The previous versions of these lemmas were deduced from the existence of adjunctions which relied on the existence of (co)products.

---

### DIFF
--- a/Mathlib.lean
+++ b/Mathlib.lean
@@ -1542,6 +1542,7 @@ import Mathlib.CategoryTheory.Limits.FintypeCat
 import Mathlib.CategoryTheory.Limits.Fubini
 import Mathlib.CategoryTheory.Limits.FullSubcategory
 import Mathlib.CategoryTheory.Limits.FunctorCategory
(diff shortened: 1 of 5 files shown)


joelriou:
The lemmas are duplicates...

dagurtomas:
Do we really want to remove the old lemmas, rather than just rename them, since the new lemmas aren't generalizations of the old ones? Or am I misunderstanding?

joelriou:
> Do we really want to remove the old lemmas, rather than just rename them, since the new lemmas aren't generalizations of the old ones? Or am I misunderstanding?

I have put the old lemmas back, as `epi_iff_epi_app'`, etc.

dagurtomas:
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15771, opened 2024-08-13
[Merged by Bors] - chore: inline `getLinterHash`*

adomani:
A copy-paste error that propagated to almost all linters -- [pointed out by Eric](https://github.com/leanprover-community/mathlib4/pull/15724#discussion_r1715138669).

Resolved by inlining the corresponding definition.

In the case of the `refine'` linter, I also recycled one doc-string and removed a `refine` namespace since every declaration already had `refine` in its name anyway.

---


Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -139,7 +139,7 @@ def missingEndLinter : Linter where run := withSetOptionIn fun stx ↦ do
     -- TODO: once mathlib's Lean version includes leanprover/lean4#4741, make this configurable
     unless #[`Mathlib, `test, `Archive, `Counterexamples].contains (← getMainModule).getRoot do
       return
-    if getLinterHash (← getOptions) && !(← MonadState.get).messages.hasErrors then
+    if getLinterMissingEnd (← getOptions) && !(← MonadState.get).messages.hasErrors then
```
eric-wieser, line 142: Should we just inline this?
```suggestion
    if Linter.getLinterValue linter.missingEnd (← getOptions) && !(← MonadState.get).messages.hasErrors then
```
Maybe we can even put `getLinterValue` in the namespace of `Lean.Option` so that we can write 
```suggestion
    if linter.missingEnd.getLinterValue (← getOptions) && !(← MonadState.get).messages.hasErrors then
```

adomani: Do you really mean this change for just the `MissingEnd` linter, or for all of them?

grunweg:
Much better; thank you for performing this rename! Come to think of it, "getLinter" sounds like a rather odd prefix. What do you think about "statusHashCommandLinter" (with a get prefix is that's house style) and similar?

Mathlib/Tactic/Linter/Lint.lean
eric-wieser: I mean inline all of them

adomani:
Michael, I actually prefer inlining, since these functions appear to be used only once and their names are not really very different/more informative than the code that they hide.

Mathlib/Tactic/Linter/Lint.lean
adomani: Done!  I did not introduce the alias, but I inlined them as is.

joneugster:
maintainer merge

PR #15115, opened 2024-07-24
feat: Generalize assumptions in liminf and limsup results in ENNReals

kkytola:
In a [review comment](https://github.com/leanprover-community/mathlib4/pull/13938#discussion_r1649744441) it was pointed out that results about liminf and limsup in ENNReal hold under milder assumptions. This PR does the generalization.

---

This PR is split off from #13938, where the review comment was made. The changes needed for the suggested generalization were of different kind than the simple PR's content, justifying a separate PR.

- [x] depends on: #13938

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #11549, opened 2024-03-20
[Merged by Bors] - feat: Levy-Prokhorov distance pseudometrizes convergence in distribution

kkytola:
Add the more interesting direction of topology comparison for the Lévy-Prokhorov distance. Conclude that if the underlying space is separable, then the topology of convergence in distribution coincides with the one obtained from the Lévy-Prokhorov (pseudo)metric. In particular the topology of convergence in distribution is (pseudo)metrizable, and so first countable.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15773, opened 2024-08-13
feat: Add type class for ENat-valued floor functions

kkytola:
This PR adds a type class for extended natural number -valued floor functions. This is split off from #15269, where the suggestions were to make the floor function on ENNReal ENat-valued and to have an API mimicling FloorSemirings. Besides ENNReal, at least ENNRat would naturally satisfy the general type class.

---

Two things are worth noting:
 * To fully mimic `FloorSemiring` API, also ceiling functions need to be added. The plan is to do that in a later PR.
 * The PR introduces the class `CastNatENatClass` to ensure that `natCast` and a coercion from `ENat` behave in the natural and order-respecing manner. This feels a little ad hoc, but such assumptions are needed for general `ENat`-valued floor functions to behave well.


urkud:
Even if we add `ENNRat`, this PR won't help deduplicating API between `ENNReal` and `ENNRat`, because Lean unfolds coercions, so generic lemmas will use `coe` while specific types will use `ENat.toENNReal` and `ENat.toENNRat`. Should we stick to `ENNReal` for now and generalize later (probably, by introduce `ENatCast` class), if needed?

urkud:
@kkytola What do you plan to do about this PR and the floor function `ENNReal -> ENat`?

kkytola:
> @kkytola What do you plan to do about this PR and the floor function `ENNReal -> ENat`?

Honestly, I am rather unsure. I created this PR to try to satisfy reviews of PR #15269, which was intended to be one of a handful of trivialities (including #15115) with which one gets the good formulation (i.e., a formulation in terms of general filters) of portmanteau theorem. The good version of portmanteau is essentially a prerequitise for more substantial theory of convergence in distribution, including what is needed for the central limit theorem. (Although in separable spaces it is possible also to work around the general filter portmanteau by using because due to first countability from pseudo-metrizability of the topology of convergence in distribution from #11549... But not embracing general filters just does not feel right for Mathlib.)

If it looks acceptable, I might consider abandoning the present PR and reverting PR #15269 back to a commit where the ENat-valued version of ENNReal.floor was working without the general API that this PR was intended to satisfy. (Perhaps a commit around d77dc74 <https://github.com/leanprover-community/mathlib4/pull/15269/commits/d77dc74c7b55ac7569c0bbb567e012bb2956eb49>, but there have since been other modifications, including Galois coinsertions.)

Is there a consensus among reviewers about what is wanted? I just wanted a floor function on ENNReal, together with its right continuity and a few other simple facts (I actually needed the ceiling function instead, but I PR'd the floor as a probe of an acceptable implementation since both counterparts should exist in the end). I will likely not have time for refactors in the near future, but recording a clear view of what is wanted would be helpful.

YaelDillies:
> If it looks acceptable, I might consider abandoning the present PR and reverting PR #15269 back to a commit where the ENat-valued version of ENNReal.floor was working without the general API that this PR was intended to satisfy. (Perhaps a commit around [d77dc74](https://github.com/leanprover-community/mathlib4/commit/d77dc74c7b55ac7569c0bbb567e012bb2956eb49), but there have since been other modifications, including Galois coinsertions.)

Yes. I think we talked past each other in #15269. I meant to say that you should write the API *as if* it were general, so that it's easier to adapt to `ENNRat` later if we want, but I didn't mean that you should be adding the machinery to actually make it more general. Basically, I want the lemma names and statements to feel like they are part of a general API, but I would rather avoid premature generalisation.

joneugster:

PR #15380, opened 2024-07-31
[Merged by Bors] - feat: topology on `ENat`

Rida-Hamadani:
this is ported from Peter Nelson's [Matroid](https://github.com/apnelson1/Matroid) project

Co-authored-by: Peter Nelson <apnelson@uwaterloo.ca>

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15774, opened 2024-08-13
feat: Topology on `ENat`

kkytola:
This PR may contain some results that are not in the PR that adds a toppology on `ENat`.

---

The immediate motivation is to be able to make the right continuity statement of the extended floor function on `ENNReal`, when this floor function is made `ENat`valued as suggested in the PR review of #15269.

The API here is modelled on the early parts of [the corresponding ENNReal file](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Instances/ENNReal.html#top), although much of this case is more trivial.

- [x] depends on: #15380


Mathlib/Data/ENat/Basic.lean
```diff
@@ -271,6 +271,20 @@ lemma toNat_eq_toNat {a b :ℕ∞} (ha : a ≠ ⊤) (hb : b ≠ ⊤) :
     a.toNat = b.toNat ↔ a = b :=
   ⟨fun h ↦ by simpa [ha, hb] using WithTop.untop'_eq_untop'_iff.mp h, fun h ↦ congrArg toNat h⟩
 
+@[simp] lemma sub_eq_top_iff {a b : ℕ∞} : a - b = ⊤ ↔ a = ⊤ ∧ b ≠ ⊤ := WithTop.sub_eq_top_iff
+
+@[simp] lemma sub_ne_top_iff {a b : ℕ∞} : a - b ≠ ⊤ ↔ a ≠ ⊤ ∨ b = ⊤ := by simp [← not_iff_not]
+
+lemma sub_add_cancel {a b : ℕ∞} (h : b ≤ a) : a - b + b = a := by
```
YaelDillies, line 278:
```suggestion
protected lemma sub_add_cancel {a b : ℕ∞} (h : b ≤ a) : a - b + b = a := by
```

PR #15780, opened 2024-08-13
[Merged by Bors] - chore(Combinatorics/SimpleGraph): rename and add `gcongr` tag to distance lemmas

Rida-Hamadani:
---

### DIFF
--- a/Mathlib/Combinatorics/SimpleGraph/Metric.lean
+++ b/Mathlib/Combinatorics/SimpleGraph/Metric.lean
@@ -268,6 +268,7 @@ lemma dist_top [DecidableEq V] : (⊤ : SimpleGraph V).dist u v = (if u = v then
   by_cases h : u = v <;> simp [h]
 
 /-- Supergraphs have smaller or equal distances to their subgraphs. -/
+@[gcongr]
 protected theorem Reachable.dist_anti {G' : SimpleGraph V} (h : G ≤ G') (hr : G.Reachable u v) :
     G'.dist u v ≤ G.dist u v := by
   obtain ⟨_, hw⟩ := hr.exists_walk_length_eq_dist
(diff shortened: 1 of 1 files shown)


Rida-Hamadani:
@YaelDillies is it a good idea to tag the `dist` lemma with `gcongr` too?

Mathlib/Combinatorics/SimpleGraph/Metric.lean
```diff
@@ -267,7 +268,7 @@ lemma dist_top [DecidableEq V] : (⊤ : SimpleGraph V).dist u v = (if u = v then
   by_cases h : u = v <;> simp [h]
 
 /-- Supergraphs have smaller or equal distances to their subgraphs. -/
-theorem dist_le_subgraph_dist {G' : SimpleGraph V} (h : G ≤ G') (hr : G.Reachable u v) :
+protected theorem Reachable.dist_anti {G' : SimpleGraph V} (h : G ≤ G') (hr : G.Reachable u v) :
     G'.dist u v ≤ G.dist u v := by
```
YaelDillies, line 272: Does `gcongr` accept this?
```suggestion
@[gcongr]
protected theorem Reachable.dist_anti {G' : SimpleGraph V} (h : G ≤ G') (hr : G.Reachable u v) :
    G'.dist u v ≤ G.dist u v := by
```

Rida-Hamadani: Yes!

YaelDillies:
Let's go!

maintainer merge

Mathlib/Combinatorics/SimpleGraph/Metric.lean
b-mehta: I'm surprised! Is it able to actually use this?

b-mehta (commented):
Nice stuff

bors merge

Ruben-VandeVelde:
Deprecated aliases seem to be missing?

Rida-Hamadani:
@Ruben-VandeVelde Sorry, these lemmas were added recently, and so far only used in an open PR by me which I will edit. Should we add deprecated aliases or is it okay if we leave it this way?

PR #15782, opened 2024-08-13
[Merged by Bors] - chore: use `assert_not_exists`

grunweg:
The first location was suggested in #15732; now we can do this.
The other locations had comments also suggesting such functionality: they referred to an outdated module name, which we correct.

---


adomani:
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15783, opened 2024-08-13
[Merged by Bors] - chore(assert_not_exists): tweak message

grunweg:
Also, upgrade the warning about a non-existing module to an error.

------

The latter change seems more logical to me (it is clearly an error, at least in this command) --- but I could be convinced otherwise.


adomani:
Looks good, thanks!

maintainer merge

semorrison:
bors merge

PR #15786, opened 2024-08-13
feat(AlgebraicGeometry/EllipticCurve/NumberField/Height): define heights on Weierstrass curves

Multramate:
---


Multramate:
This is a preliminary investigation into formalising heights (naive, logarithmic, canonical, relative, absolute, etc) on elliptic curves. This is necessary to prove the Mordell--Weil theorem, which is necessary to define the elliptic regulator in the leading term part of the Birch--Swinnerton-Dyer conjecture.

Multramate:
This PR has been migrated to a fork-based workflow: https://github.com/leanprover-community/mathlib4/pull/25986

PR #15791, opened 2024-08-14
[Merged by Bors] - chore: better failures for `linear_combination`

hrmacbeth:
The implementation of `linear_combination` is as a macro performing effectively `refine ******; ring`.  If the term provided to `refine` fails to elaborate, Lean inserts sorries where needed and then continues on to `ring`.  This produces a confusing error message.  For example, the problem
```lean
example (x y : ℤ) (h1 : x * y + 2 * x = 1) (h2 : x = y) : x * y + 2 * x = 1 := by
  linear_combination h1 + (0 : ℝ) * h2
```
produces both the "true" error message
```
application type mismatch
  Mathlib.Tactic.LinearCombination.c_mul_pf h2 0
argument
  0
has type
  ℝ : Type
but is expected to have type
  ℤ : Type
```
and the "tactic is confused because it persevered when it shouldn't have" error message
```
ring failed, ring expressions not equal
x y : ℤ
h1 : x * y + 2 * x = 1
h2 : x = y
⊢ -(x * sorryAx ℤ true) + y * sorryAx ℤ true = 0
```

This PR fixes this by strategically inserting a `Term.withoutErrToSorry`.  In examples such as the above, it now stops after the `refine`, so the second error message does not appear.

---

### DIFF
--- a/Mathlib/Tactic/LinearCombination.lean
+++ b/Mathlib/Tactic/LinearCombination.lean
@@ -138,7 +138,7 @@ def elabLinearCombination
     | .const c => `(Eq.refl $c)
     | .proof p => pure p
   let norm := norm?.getD (Unhygienic.run `(tactic| ring1))
-  Tactic.evalTactic <| ← withFreshMacroScope <|
+  Term.withoutErrToSorry <| Tactic.evalTactic <| ← withFreshMacroScope <|
   if twoGoals then
     `(tactic| (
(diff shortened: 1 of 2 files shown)


kmill:
Seems good. Maybe it could make sense to move withoutErrToSorry earlier, but where you have it currently is definitely justified since it's for elaboration of syntax generated by the tactic itself.

bors r+

PR #15792, opened 2024-08-14
[Merged by Bors] - chore: cleanup in List/Infix

kim-em:


Mathlib/Data/List/Infix.lean
```diff
@@ -186,7 +186,7 @@ instance decidableInfix [DecidableEq α] : ∀ l₁ l₂ : List α, Decidable (l
       (@instDecidableOr _ _ (l₁.decidablePrefix (b :: l₂)) (l₁.decidableInfix l₂))
       infix_cons_iff.symm
 
-theorem prefix_take_le_iff {L : List (List (Option α))} (hm : m < L.length) :
+theorem prefix_take_le_iff {L : List α} (hm : m < L.length) :
```
joneugster, line 189: I don't fully understand why the signature used to be this way before the change... If a maintainer could look at this before

maintainer merge

that would be good.

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/Data/List/Infix.lean
semorrison: Yes, it was just horribly wrong before.

PR #15793, opened 2024-08-14
[Merged by Bors] - feat(SetTheory/ZFC/Ordinal): Initial development of ordinals in ZFSet

vihdzp:
This is ported from the Mathlib 3 branch [`von_neumann_v2`](https://github.com/leanprover-community/mathlib3/blob/5b3c1d3d66838f426b44cae25b615f26438c8acb/src/set_theory/zfc/ordinal.lean). This is meant as a small, initial PR to iron out any kinks (with the definition, notation, auto-ported lemmas, etc.) before fully developing the file.

Of particular note is the quite bare-bones definition of an ordinal. Surprisingly, our weakened transitivity assumption is equivalent to the much stronger claim that ordinals are well-ordered under `∈`; see #17001 for a proof.

---


Mathlib/SetTheory/ZFC/Basic.lean
```diff
@@ -340,11 +340,14 @@ instance (x y : PSet) : Inhabited (insert x y).Type :=
   inferInstanceAs (Inhabited <| Option y.Type)
 
 /-- The n-th von Neumann ordinal -/
+@[deprecated "this will be moved into the ZFC/Ordinal file" (since := "2024-08-18")]
```
b-mehta, line 343: I don't think we should deprecate these until the replacement exists (hopefully this should be soon anyway?)

Mathlib/SetTheory/ZFC/Ordinal.lean
```diff
@@ -49,38 +51,44 @@ theorem isTransitive_iff_mem_trans : z.IsTransitive ↔ ∀ {x y : ZFSet}, x ∈
 alias ⟨IsTransitive.mem_trans, _⟩ := isTransitive_iff_mem_trans
 
 protected theorem IsTransitive.inter (hx : x.IsTransitive) (hy : y.IsTransitive) :
-    (x ∩ y).IsTransitive := fun z hz w hw => by
```
b-mehta, line 52: These changes really aren't necessary! Both the old and new are valid style (I personally prefer the old version, but it's a matter of taste)

Mathlib/SetTheory/ZFC/Ordinal.lean
```diff
@@ -89,4 +97,42 @@ theorem isTransitive_iff_subset_powerset : x.IsTransitive ↔ x ⊆ powerset x :
 
 alias ⟨IsTransitive.subset_powerset, _⟩ := isTransitive_iff_subset_powerset
 
+/-- A set `x` is a von Neumann ordinal when it's a transitive set, that's transitive under `∈`. We
+will prove that this further implies that `x` is well-ordered under `∈`.
+
+The transitivity condition is written in an even weaker form, where `a ∈ b` and `b ∈ c` imply
+`a ∈ c` when only `c ∈ x` and not `a ∈ x` or `b ∈ x` are known a priori. -/
```
b-mehta, line 104: Doesn't `x.IsTransitive` automatically give that `b ∈ c` together with `c ∈ x` imply `b ∈ x`? If so, I'm not sure I'd phrase this as "written in an even weaker form". Perhaps "The transitivity condition can be written without assuming `a ∈ x` and `b ∈ x`, and `isOrdinal_iff_isTrans` shows it is equivalent to the usual one." What do you think?

Mathlib/SetTheory/ZFC/Ordinal.lean
```diff
@@ -89,4 +97,42 @@ theorem isTransitive_iff_subset_powerset : x.IsTransitive ↔ x ⊆ powerset x :
...
+def IsOrdinal (x : ZFSet) : Prop :=
+  x.IsTransitive ∧ ∀ a b c : ZFSet, a ∈ b → b ∈ c → c ∈ x → a ∈ c
+
+namespace IsOrdinal
+
+protected theorem isTransitive (h : x.IsOrdinal) : x.IsTransitive :=
+  h.1
+
+theorem subset_of_mem (h : x.IsOrdinal) : y ∈ x → y ⊆ x :=
```
b-mehta, line 113: Hmm, I'm not entirely convinced this one and the one below should be lemmas since they're just combinations of `isTransitive` and lemmas about transitive sets. `mem_trans'` is fine though

Mathlib/SetTheory/ZFC/Ordinal.lean
vihdzp: I like that phrasing better.

Mathlib/SetTheory/ZFC/Ordinal.lean
vihdzp: There's a bunch of lemmas about `∈` and `⊆` we can state on ordinals that aren't true for general transitive sets, such as
```lean 4
variable {x y z : ZFSet} 

example : IsOrdinal x → IsOrdinal y → x ⊆ y → y ∈ z → x ∈ z := sorry
example : IsOrdinal x → IsOrdinal y → x ⊆ y ∨ y ⊆ x := sorry
example : IsOrdinal x → IsOrdinal y → x ∈ y ∨ x = y ∨ y ∈ x := sorry
```
I think it's convenient to have these small wrappers over `IsTransitive` - that way you don't have to stop and think whether a theorem holds for all transitive sets or is exclusive to ordinals.

b-mehta: Okay, fair enough

vihdzp:
I made `IsOrdinal` into a structure, I felt that having the fields explicitly named was cleaner.

Mathlib/SetTheory/ZFC/Ordinal.lean
```diff
@@ -3,32 +3,33 @@ Copyright (c) 2022 Violeta Hernández Palacios. All rights reserved.
...
-  `SetTheory/Ordinal/Arithmetic.lean`.
+- Define von Neumann ordinals and the von Neumann hierarchy.
+- Build correspondences between these notions and those of the standard `Ordinal` type.
 -/
 
 
 universe u
 
-variable {x y z : ZFSet.{u}}
+variable {x y z w : ZFSet.{u}}
```
YnirPaz, line 32:
```suggestion
variable {x y z : ZFSet.{u}}
```

YnirPaz (approved):
I actually learned this definition as the normal one (an ordinal is a transitive set of transitive sets). Either that or a transitive set linearly ordered by epsilon, but I haven't heard of the definition of a transitive set on which epsilon is a transitive relation. No need to change anything, I'm just interested in the cultural differences.

Command-Master (approved):

YaelDillies (approved):
Thanks! 🚀

maintainer merge

kim-em:
bors merge

PR #15794, opened 2024-08-14
feat: the DocModule linter

adomani:
As is, this linter only works on a "static" file: any edit will cause the linter to flag perfectly reasonable code!

The `DocModule` style linter is a syntax linter that checks if a file starts with
```lean
import*
/-! doc-module -/
```

It emits a warning if a file does not have a doc-module string as its first non-`import` command.

In particular, the linter allows `import`-only files to have no doc-module string.

The linter also allows files in the `test` folder to not have doc-module strings.

---


Mathlib/Tactic/Linter/DocModule.lean
```diff
@@ -0,0 +1,77 @@
...
+  if (← get).messages.hasErrors then
+    return
+  -- `test` files are not required to have a doc-module string
+  if let `test::_ := (← getMainModule).components then return
+  let undoc? ← undocumented.get
+  -- `terminal?` is `true` iff `stx` is the end of the file, or
+  -- (unreachable by the linter) `import X`
+  let terminal? := Parser.isTerminalCommand stx
+  -- if we reached the end of the file, we reset to false the `undocumented` counter and
+  -- report nothing: thus, `import`-only files do not get flagged by the linter
```
grunweg, line 62: `terminal?` is used only once - inline it?

Mathlib/Tactic/Linter/DocModule.lean
```diff
@@ -0,0 +1,77 @@
...
+  if terminal? then
+    undocumented.set false
+    return
+  unless ! undoc? do return
```
grunweg, line 66: I am worried about one edge case: what happens if you have a module docstring *before* imports?
It seems this is allowed by the linter - is that what we want?

grunweg, line 66: Can you avoid the double negation? I had to think twice about whether this was the right way around...

Mathlib/Tactic/Linter/DocModule.lean
```diff
@@ -0,0 +1,77 @@
...
+  if !stx.isOfKind ``Lean.Parser.Command.moduleDoc then
+    Linter.logLint linter.docModule stx
+      m!"Add the doc-module string before this command\n\
```
grunweg, line 69: What do you think about reversing the order, like so?
```
`Mathlib` files must contain a module doc-string between their list of `import`s and the first non-`import` command.
Please add a module doc-string before this command.

**Review by grunweg — COMMENTED** [review]
I look a more detailed look. The overall idea looks nicely done, as usual.
- As you say, it would be good to avoid the flimsy IO.Ref usage, so reloading the file is not necessary.
- ~~I wonder if this linter works as intended if you have a module doc-string *before* imports.~~ impossible, see below
- Can you add a test for this? I am not confident I'd get the detailed logic right on first try, the `undoc?` and `terminal?` both flip twice throughout the file, which is subtle enough to not be "entirely obvious".
Update: I agree that is slightly tricky, as the linter disables itself in the `tests` directory... would the same trick as in the other PR help - disabling this linter by default, enabling it in mathlib, but disabling *this particular* linter in the tests directory?

File: Mathlib/Tactic/Linter/DocModule.lean (hunk 2 above)
Comment (grunweg) [reply Mathlib/Tactic/Linter/DocModule.lean:66]: I just tested this locally: Lean forces the first command in a file to be imports - this means module docstrings cannot go *before* imports. It would be nice to document this - this is implicitly used by the linter.

File: Mathlib/Tactic/Linter/DocModule.lean (hunk 1 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/DocModule.lean:64]: Good catch!

File: Mathlib/Tactic/Linter/DocModule.lean (hunk 3 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/DocModule.lean:71]: I reversed the order, but prefer to talk about the first non-`import` command: this covers the case of import-only files, albeit implicitly.

File: Mathlib/Tactic/Linter/DocModule.lean (hunk 2 above)
Comment (adomani) [reply Mathlib/Tactic/Linter/DocModule.lean:68]: Used a plain `if ... then` instead of double negation.

File: Mathlib/Tactic/Linter/DocModule.lean [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+  unless Linter.getLinterValue linter.docModule (← getOptions) do
+    return
+  if (← get).messages.hasErrors then
+    return
+  -- `test` files are not required to have a doc-module string
+  if let `test::_ := (← getMainModule).components then return
+  let undoc? ← undocumented.get
+  -- `terminal?` is `true` iff `stx` is the end of the file, or
+  -- (unreachable by the linter) `import X`
+  let terminal? := Parser.isTerminalCommand stx
```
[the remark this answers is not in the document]
adomani: This disappeared now, to the comment is resolved!  :smile:

adomani (commented):
Thanks, Michael!  I've thought about this, and I think that this use of an `IO.Ref` is not going to work with the linter.  I'll implement a different version!

adomani:
The linter at #16574 should be a better implementation of this one.

PR #15795, opened 2024-08-14
[Merged by Bors] - fix: `assert_not_exists` takes at least one input

adomani:
[Reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/assert_not_imported/near/462220460).

---


joneugster:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

grunweg:
Thanks for the really fast response! Would you like to add a test/like me to add a one?

grunweg:
I just tested this locally: the wrong code doesn't parse, so no test seems necessary.

adomani:
I had tried testing it, but `#guard_msgs in` does not seem to catch this sort of error.

PR #15796, opened 2024-08-14
[Merged by Bors] - feat(Algebra/EuclideanDomain/Defs): add `IsWellFounded` instance

vihdzp:
This does add an import, but it's a very early one at that.

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #15797, opened 2024-08-14
[Merged by Bors] - feat(RingTheory/Artinian): make `IsArtinian` an abbreviation for `WellFoundedLT`

vihdzp:
We also tweak the `IsNoetherian` API analogously. This all improves code legibility quite a bit, and allows instance inference to be useful.

---

We can't directly make `Noetherian` an abbreviation since it's defined in terms of free submodules instead. However, we can create the corresponding instance, and a few helper lemmas.


semorrison:
Please don't use the `tech-debt` label for things like this. See e.g. [https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Technical.20Debt.20Counters].

grunweg:
> Please don't use the `tech-debt` label for things like this. See e.g. https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Technical.20Debt.20Counters.

@semorrison I took the liberty to update the label description accordingly.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15799, opened 2024-08-14
[Merged by Bors] - feat(CategoryTheory): Finality of the forget functor on over and under categories

javra:
---


TwoFX (approved):
Thanks!
bors r+

PR #13027, opened 2024-05-19
[Merged by Bors] - chore: move `SMulWithZero` with `SMulZeroClass`

YaelDillies:
Move the basic content of `GroupTheory.GroupAction` to new folders `Algebra.Group.Action` and `Algebra.GroupWithZero.Action` depending on whether it is additivisable or not.

---

- [x] depends on: #12974
- [x] depends on: #13029
- [x] depends on: #13030
- [x] depends on: #13031
- [x] depends on: #13140
- [x] depends on: #13161
- [x] depends on: #14677
- [x] depends on: #14705
- [x] depends on: #14706
- [x] depends on: #14851
- [x] depends on: #14870
- [x] depends on: #14871
- [x] depends on: #14937
- [x] depends on: #15430
- [x] depends on: #15806
- [x] depends on: #15983
- [x] depends on: #15984
- [x] depends on: #16440
- [x] depends on: #20773
- [x] depends on: #22121
- [x] depends on: #22395
- [x] depends on: #22512

I am currently splitting one PR off for each new file to discuss the specifics and bisecting out the weird timeouts I'm hitting.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15806, opened 2024-08-14
[Merged by Bors] - chore(GroupTheory/GroupAction/Group): Delete

YaelDillies:
All lemmas can be moved to earlier files

---


Ruben-VandeVelde (approved):
maintainer merge

Mathlib/Algebra/GroupWithZero/Action/Defs.lean
```diff
@@ -3,11 +3,9 @@ Copyright (c) 2018 Chris Hughes. All rights reserved.
...
 Authors: Chris Hughes, Yury Kudryashov
 -/
-import Mathlib.Algebra.Group.Action.Defs
-import Mathlib.Algebra.Group.Hom.Defs
-import Mathlib.Algebra.Group.TypeTags
-import Mathlib.Algebra.Opposites
-import Mathlib.Logic.Embedding.Basic
+import Mathlib.Algebra.Group.Action.Units
+import Mathlib.Algebra.Group.Equiv.Basic
+import Mathlib.Algebra.GroupWithZero.Units.Basic
```
jcommelin, line 8: This seems to amount to a 12% increase of the number transitive imports...

Since this is a `Defs` file, I would rather prefer to see the number go down.

YaelDillies: It is a very recent `Defs` file, however. I am still working on it. Here are my future plans:
1. Move some of the content from `GroupTheory.GroupAction.Group` to this file (this PR)
2. Move most of the content of `Algebra.SMulWithZero` to this file (not yet opened, but you can see the result in #13027)
3. Think about a nice way to split things into `Defs` and `Basic`

I can't really do the last step before the first two, so I would rather stick to my plan and declare this temporary 12% increase as fine.

j-loreaux: Why can't you do (2) first and then (1) and (3) simultaneously?

YaelDillies: (2) will increase imports as well

j-loreaux:
@jcommelin I'll let you decide, since you initially raised the objection. I'm content with Yaël's indication to later split the file into `Defs` + `Basic` in some helpful way.

jcommelin:
Ok, let's go ahead.

bors merge

PR #15812, opened 2024-08-14
chore(Set.Basic): make `Set.inclusion` an `abbrev`

mattrobball:
This change allows cleaning up one `rfl`

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Should.20.60Set.2Einclusion.60.20be.20an.20.60abbrev.60.3F/near/462205622)

---

Let's see the performance impact


mattrobball:
!bench

eric-wieser:
Besides cleaning up one messy proof (that can probably be cleaned up without making `inclusion` an `abbrev`), the win doesn't seem super clear to me. The def seems a little easier to read that a function with an invisible coercion in, and `simp` gets rid of `inclusion` anyway once you case on the right variables.

mattrobball:
Ok, strong majority win then.

MichaelStollBayreuth:
<b>General information:</b>
```text
build:                                                 -2.1430 * 10⁹ (-0.00174%)
```


<b>1 file got slower by at least 10⁹ instructions:</b>
```text
Mathlib.AlgebraicGeometry.Restrict:                    +1.0597 * 10⁹ (+1.23%)
```


<b>1 file got slower by at least 10%:</b>
```text
Mathlib.Topology.Sheaves.Sheafify:                                    +13.3%
```


<b>3 files got faster by at least 10⁹ instructions:</b>
```text
Mathlib.Algebra.Star.NonUnitalSubalgebra:              -2.4904 * 10⁹ (-1.28%)
Mathlib.Algebra.Algebra.NonUnitalSubalgebra:           -1.2540 * 10⁹ (-1.24%)
Mathlib.Algebra.Algebra.Subalgebra.Directed:           -1.1011 * 10⁹ (-6.41%)
```


jcommelin:
What's up with the
```quote
1 file got slower by at least 10%:

    Mathlib.Topology.Sheaves.Sheafify:                                    +13.3%
```

mattrobball:
It went from 7B to 7.9B

Mathlib/Data/Set/Image.lean
```diff
@@ -938,19 +938,7 @@ theorem range_diff_image {f : α → β} (H : Injective f) (s : Set α) : range
...
-    rw [inclusion, Subtype.mk.injEq] at ha
-    rw [mem_setOf, Subtype.coe_mk, ← ha]
-    exact Subtype.coe_prop _
-  · rw [mem_setOf, Subtype.coe_mk, mem_range]
-    intro hx'
-    use ⟨x, hx'⟩
-    trivial
-  -- simp_rw [inclusion, mem_range, Subtype.mk_eq_mk]
-  -- rw [SetCoe.exists, Subtype.coe_mk, exists_prop, exists_eq_right, mem_set_of, Subtype.coe_mk]
+  simp
```
eric-wieser, line 941: I created #15819 with the same proof: the porting note was just wrong, it works fine.

vihdzp:
I think this is a prime use case for an abbreviation, it allows writing a simple yet abstract string of symbols `⟨x, h x.2⟩` in a human readable way. If the performance issues aren't major, then LGTM.

YaelDillies:
I tend to agree with Eric here that the point of `Set.inclusion` is to use it unapplied, else it gets simped away. This matters in simp lemmas, as then being able to index on `Set.inclusion` is important, eg [`Set.range_inclusion`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Image.html#Set.range_inclusion) would see its key become `Set.range ?_` after this PR.

vihdzp:
That's a good point. Approval revoked.

mattrobball:
> I tend to agree with Eric here that the point of `Set.inclusion` is to use it unapplied, else it gets simped away. This matters in simp lemmas, as then being able to index on `Set.inclusion` is important, eg [`Set.range_inclusion`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Image.html#Set.range_inclusion) would see its key become `Set.range ?_` after this PR.

Sure, but when you index correctly on subtypes, it turns out to be more pain that useful.

YaelDillies:
Mind expanding? I'm not sure what "indexing correctly on subtypes" amounts to here.

mattrobball:
leanprover/lean4#5020

eric-wieser:
What's the trick for finding out the index computed for a simp lemma?

mattrobball:
`#discr_tree_simp_key my_lemma` and `#discr_tree_key myInstance`

mattrobball:
There is also a tactic which does it for the goal state which I might PR to mathlib at some point.

PR #15813, opened 2024-08-14
[Merged by Bors] - feat: Actions on and by an order synonym

YaelDillies:
Complete the API about actions on and by `OrderDual`/`Lex`.

---

### DIFF
--- a/Mathlib/Algebra/Order/Group/Action/Synonym.lean
+++ b/Mathlib/Algebra/Order/Group/Action/Synonym.lean
@@ -1,0 +1,88 @@
+/-
+Copyright (c) 2021 Yaël Dillies. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yaël Dillies
+-/
+import Mathlib.Algebra.Group.Action.Defs
+import Mathlib.Algebra.Order.Group.Synonym
+
+/-!
+# Actions by and on order synonyms
+
+This PR transfers group action instances from a type `α` to `αᵒᵈ` and `Lex α`.
+
+## See also
+
+* `Mathlib.Algebra.Order.GroupWithZero.Action.Synonym`
+* `Mathlib.Algebra.Order.Module.Synonym`
+-/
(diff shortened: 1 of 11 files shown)


Mathlib/Algebra/Order/Group/Action/Synonym.lean
```diff
@@ -0,0 +1,81 @@
...
+/-!
+# Actions by and on order synonyms
+-/
+
+variable {M N α : Type*}
+
+namespace OrderDual
+
+@[to_additive]
+instance instMulAction [Monoid M] [MulAction M α] : MulAction (Lex M) α := ‹MulAction M α›
```
eric-wieser, line 18: This is in the `OrderDual` namespace yet is about `Lex`.

Mathlib/Algebra/Order/GroupWithZero/Action/Synonym.lean
```diff
@@ -0,0 +1,70 @@
...
+
+/-!
+# Actions by and on order synonyms
+-/
+
+variable {G₀ M₀ : Type*}
+
+namespace OrderDual
+
+instance instSMulWithZero [Zero G₀] [Zero M₀] [SMulWithZero G₀ M₀] : SMulWithZero G₀ᵒᵈ M₀ :=
```
eric-wieser, line 18: Is the `SMul` instance defined in an earlier file? If so, can you mention where that instance is (and the lemmas around it) in the docstrings of the new files?

YaelDillies: `Algebra.Order.Group.Action.Synonym` is where it is defined

Mathlib/Algebra/Order/Group/Action/Synonym.lean
eric-wieser: This comment still applies to this line

b-mehta (approved):
bors merge

PR #15814, opened 2024-08-14
[Merged by Bors] - chore: `LinearOrderedSemiring α → LinearOrderedCancelAddCommMonoid α`

YaelDillies:
This was accidentally about `LinearOrderedCommSemiring`, probably due to confusion over which operation the `Comm` in the name refers to.

---


j-loreaux:
!bench

YaelDillies:
Very strange that the benchmark fails. Let me try merging master.

mattrobball:
Sorry. Despite authoring to the commit fix this problem with the bench bot, I seem to have completely forgotten. 

Merging master will fix it

YaelDillies:
!bench

vihdzp:
I think the tech debt label isn't intended for general "stuff we've missed over the years", but specifically things left over from the Mathlib 3 port.

YaelDillies:
That would be a valid interpretation, but in practice that's not how we are using it.

PR #15817, opened 2024-08-14
[Merged by Bors] - feat(Data.Fin): Add lemma `Fin.sub_succ_le_sub_of_le`

IvanRenison:
---

### DIFF
--- a/Mathlib/Data/Fin/Basic.lean
+++ b/Mathlib/Data/Fin/Basic.lean
@@ -1484,6 +1484,13 @@ lemma pos_of_ne_zero {n : ℕ} {a : Fin (n + 1)} (h : a ≠ 0) :
     0 < a :=
   Nat.pos_of_ne_zero (val_ne_of_ne h)
 
+lemma sub_succ_le_sub_of_le {n : ℕ} {u v : Fin (n + 2)} (h : u < v) : v - (u + 1) < v - u  := by
+  have h' : u + 1 ≤ v := add_one_le_of_lt h
+  apply lt_def.mpr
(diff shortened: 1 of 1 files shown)


Mathlib/Data/Fin/Basic.lean
```diff
@@ -1484,6 +1484,13 @@ lemma pos_of_ne_zero {n : ℕ} {a : Fin (n + 1)} (h : a ≠ 0) :
     0 < a :=
   Nat.pos_of_ne_zero (val_ne_of_ne h)
 
+lemma sub_succ_le_sub_of_le {n : ℕ} {u v : Fin (n + 2)} (h : u < v) : v - (u + 1) < v - u  := by
```
Ruben-VandeVelde, line 1487:
```suggestion
lemma sub_succ_le_sub_of_le {n : ℕ} {u v : Fin (n + 2)} (h : u < v) : v - (u + 1) < v - u := by
```

Ruben-VandeVelde (approved):
maintainer merge

Thanks!

semorrison:
@IvanRenison, apologies to hijack your PR, but while reviewing your proof I was sad that we didn't have better automation for this, so went away and wrote some!

I've pushed a commit that adds a `fin_omega` macro, which is just a preprocessor for the `omega` tactic, that then handles your new lemma as just `by fin_omega`.

If it is inappropriate for me to hijack your PR, please just let me know and we can merge as is, and then I'll do this in a separate PR. But if you don't object it seems less overhead to do it directly here.

semorrison:
bors merge

PR #15819, opened 2024-08-14
[Merged by Bors] - chore: remove a porting note and simplify

eric-wieser:
The old proof seems to work again. Also tidies some related proofs.

---


j-loreaux (commented):
bors merge

PR #15827, opened 2024-08-15
chore: adaptations for nightly-2024-08-14

kim-em:


Mathlib/Data/List/Basic.lean
```diff
@@ -972,6 +955,7 @@ theorem indexOf_inj [DecidableEq α] {l : List α} {x y : α} (hx : x ∈ l) (hy
       simp only [h]
     simp only [indexOf_get] at x_eq_y; exact x_eq_y, fun h => by subst h; rfl⟩
 
+@[deprecated getElem_reverse_aux₂ (since := "2024-08-15")]
 theorem getElem_reverse_aux₂ :
```
Ruben-VandeVelde, line 959: Is this deprecated in favor of the same lemma?

Mathlib/Data/List/Infix.lean
```diff
@@ -35,36 +33,23 @@ variable {l l₁ l₂ l₃ : List α} {a b : α} {m n : ℕ}
...
-  prefix_refl _
+  prefix_refl
 
+@[deprecated suffix_refl (since := "2024-08-15")]
 theorem suffix_rfl : l <:+ l :=
-  suffix_refl _
+  suffix_refl
 
+@[deprecated infix_refl (since := "2024-08-15")]
 theorem infix_rfl : l <:+: l :=
```
Ruben-VandeVelde, line 47: Wait, is upstream using `_refl` for the implicit argument version? That should be fixed

Mathlib/Data/List/Basic.lean
semorrison: Oops, should just be deprecated without a suggestion: this is just an auxiliary lemma for a now deprecated theorem, `get_reverse`.

Mathlib/Data/List/Infix.lean
semorrison: Ah, indeed. I've removed the deprecations and will fix upstream.

semorrison: https://github.com/leanprover/lean4/pull/5077

jcommelin:
bors d+

PR #15828, opened 2024-08-15
[Merged by Bors] - feat: Compute derivative of `fun x ↦ f (a - x)`

YaelDillies:
... and similar. Also generalise existing lemmas to not require differentiability assumptions.

---

### DIFF
--- a/Mathlib/Analysis/Calculus/Deriv/Add.lean
+++ b/Mathlib/Analysis/Calculus/Deriv/Add.lean
@@ -132,6 +132,14 @@ lemma differentiableAt_comp_add_const {a b : 𝕜} :
     DifferentiableAt 𝕜 (fun x ↦ f (x + b)) a ↔ DifferentiableAt 𝕜 f (a + b) := by
   simpa [add_comm b] using differentiableAt_comp_const_add (f := f) (b := b)
 
+lemma differentiableAt_iff_comp_const_add {a b : 𝕜} :
+    DifferentiableAt 𝕜 f a ↔ DifferentiableAt 𝕜 (fun x ↦ f (b + x)) (-b + a) := by
+  simp [differentiableAt_comp_const_add]
+
+lemma differentiableAt_iff_comp_add_const {a b : 𝕜} :
+    DifferentiableAt 𝕜 f a ↔ DifferentiableAt 𝕜 (fun x ↦ f (x + b)) (a - b) := by
+  simp [differentiableAt_comp_add_const]
+
 end Add
 
 section Sum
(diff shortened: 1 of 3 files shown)


Mathlib/Analysis/Calculus/Deriv/Shift.lean
```diff
@@ -1,48 +1,63 @@
...
 
 /-- The derivative of `x ↦ f (-x)` at `a` is the negative of the derivative of `f` at `-a`. -/
-lemma deriv_comp_neg {𝕜 : Type*} [NontriviallyNormedField 𝕜] {F : Type*} [NormedAddCommGroup F]
-    [NormedSpace 𝕜 F] (f : 𝕜 → F) (a : 𝕜) : deriv (fun x ↦ f (-x)) a = -deriv f (-a) := by
-  by_cases h : DifferentiableAt 𝕜 f (-a)
-  · simpa only [deriv_neg, neg_one_smul] using deriv.scomp a h (differentiable_neg _)
-  · rw [deriv_zero_of_not_differentiableAt (mt differentiableAt_comp_neg_iff.mpr h),
-      deriv_zero_of_not_differentiableAt h, neg_zero]
+lemma deriv_comp_neg : deriv (fun x ↦ f (-x)) x = -deriv f (-x) := by
+  by_cases f : DifferentiableAt 𝕜 f (-x)
```
dupuisf, line 43: Shouldn't all these `deriv` lemmas be simp lemmas?

Mathlib/Analysis/Calculus/Deriv/Add.lean
```diff
@@ -120,6 +120,18 @@ theorem deriv_const_add (c : F) : deriv (fun y => c + f y) x = deriv f x := by
...
+lemma differentiableAt_comp_const_add {a b : 𝕜} :
+    DifferentiableAt 𝕜 (fun x ↦ f (b + x)) a ↔ DifferentiableAt 𝕜 f (b + a) := by
+  refine ⟨fun H ↦ ?_, fun H ↦ H.comp _ (differentiable_id.const_add _).differentiableAt⟩
+  convert DifferentiableAt.comp (b + a) (by simpa)
+    (differentiable_id.const_add (-b)).differentiableAt
+  ext
+  simp
+
+lemma differentiableAt_comp_add_const {a b : 𝕜} :
+    DifferentiableAt 𝕜 (fun x ↦ f (x + b)) a ↔ DifferentiableAt 𝕜 f (a + b) := by
```
dupuisf, line 132: Do we want iff versions like for the other ones below?

dupuisf (commented):
Thanks!

bors d+

Mathlib/Analysis/Calculus/Deriv/Shift.lean
YaelDillies: I doubt so. Since lambdas aren't (yet) indexed in the discrimination tree, their key would all be of the form `deriv ?_ ?_` which is rather generic. Maybe not a performance issue since there aren't so many uses of `deriv`, but for now I would rather not.

Mathlib/Analysis/Calculus/Deriv/Add.lean
YaelDillies: I decided against them because they would involve subtraction, but I could be convinced otherwise

Mathlib/Analysis/Calculus/Deriv/Shift.lean
dupuisf: Fair enough.

Mathlib/Analysis/Calculus/Deriv/Add.lean
dupuisf: Well, I'll leave it up to you.

YaelDillies: Added :)

YaelDillies:
bors merge

PR #15835, opened 2024-08-15
[Merged by Bors] - chore(Archive): remove most bare open Classical

grunweg:
A few files remain, which use it pretty pervasively: I was not sure if removing it there would have been an improvement.

---


urkud:
Thanks! :tada:
bors merge

PR #15836, opened 2024-08-15
[Merged by Bors] - chore(Logic): remove bare open Classical

grunweg:
---

### DIFF
--- a/Mathlib/Logic/Denumerable.lean
+++ b/Mathlib/Logic/Denumerable.lean
@@ -184,21 +184,19 @@ open Function Encodable
 
 /-! ### Subsets of `ℕ` -/
 
-
 variable {s : Set ℕ} [Infinite s]
 
 section Classical
 
 theorem exists_succ (x : s) : ∃ n, (x : ℕ) + n + 1 ∈ s := by
+  by_contra h
+  have : ∀ (a : ℕ) (_ : a ∈ s), a < x + 1 := fun a ha =>
+    lt_of_not_ge fun hax => h ⟨a - (x + 1), by rwa [add_right_comm, Nat.add_sub_cancel' hax]⟩
   classical
-  exact _root_.by_contradiction fun h =>
-    have : ∀ (a : ℕ) (_ : a ∈ s), a < x + 1 := fun a ha =>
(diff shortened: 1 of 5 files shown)


Mathlib/Logic/Nontrivial/Defs.lean
```diff
@@ -44,6 +41,7 @@ protected theorem Decidable.exists_ne [Nontrivial α] [DecidableEq α] (x : α)
     exact ⟨y', h.symm⟩
   · exact ⟨y, Ne.symm hx⟩
 
+open Classical in
 theorem exists_ne [Nontrivial α] (x : α) : ∃ y, y ≠ x := Decidable.exists_ne x
```
grunweg, line 45: This is the only change I'm not fully sure of.

jcommelin: Why?

Mathlib/Logic/Denumerable.lean
```diff
@@ -189,10 +189,9 @@ variable {s : Set ℕ} [Infinite s]
 
 section Classical
 
-open scoped Classical
-
-theorem exists_succ (x : s) : ∃ n, (x : ℕ) + n + 1 ∈ s :=
-  _root_.by_contradiction fun h =>
+theorem exists_succ (x : s) : ∃ n, (x : ℕ) + n + 1 ∈ s := by
+  classical
+  exact _root_.by_contradiction fun h =>
```
jcommelin, line 194: Can this be optimized using `by_contra h` or something like that?

Mathlib/Logic/Nontrivial/Defs.lean
grunweg: I'm unsure in the sense of "should I add a suitable decidability assumption instead?" --- though come to think of it, the answer is "no".

Mathlib/Logic/Denumerable.lean
grunweg: Yes, it can - pushed.

j-loreaux (commented):
Thanks!

bors merge

PR #15838, opened 2024-08-15
[Merged by Bors] - feat(CategoryTheory): Characterization of Structured Arrows of Diagonal and Projections

javra:
---

### DIFF
--- a/Mathlib/CategoryTheory/Comma/Over.lean
+++ b/Mathlib/CategoryTheory/Comma/Over.lean
@@ -673,13 +673,13 @@ def ofStructuredArrowProjEquivalence {D : Type u₂} [Category.{v₂} D] (F : D 
          (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
       (fun f => f.right.hom) (by simp))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.id_obj, Functor.const_obj_obj, proj_obj, left_eq_id, Functor.const_obj_map,
-        proj_map, Functor.comp_obj, Functor.comp_map, eq_mp_eq_cast, id_eq, Under.forget_obj,
-        Under.forget_map, Functor.toStructuredArrow_obj, Functor.toUnder_obj, mk_right,
-        mk_hom_eq_self]
+      dsimp only [Functor.id_obj, Functor.const_obj_obj, proj_obj, left_eq_id,
+        Functor.const_obj_map, proj_map, Functor.comp_obj, Functor.comp_map, eq_mp_eq_cast, id_eq,
+        Under.forget_obj, Under.forget_map, Functor.toStructuredArrow_obj, Functor.toUnder_obj,
+        mk_right, mk_hom_eq_self]
       apply eta) (by simp))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.const_obj_obj, Functor.comp_obj, Under.forget_obj, left_eq_id,
+      dsimp only [Functor.const_obj_obj, Functor.comp_obj, Under.forget_obj, left_eq_id,
         Functor.const_obj_map, Functor.comp_map, Under.forget_map, proj_obj, proj_map,
         eq_mp_eq_cast, id_eq, Functor.toStructuredArrow_obj, Functor.toUnder_obj, mk_right,
         mk_hom_eq_self, Functor.id_obj]
@@ -697,18 +697,24 @@ def ofDiagEquivalence (X : T × T) :
     (Functor.toStructuredArrow (StructuredArrow.proj _ _ ⋙ Under.forget _) _ _
       (fun f => (f.right.hom, f.hom)) (fun m => by have := m.w; aesop_cat))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.id_obj, Functor.const_obj_obj, Functor.diag_obj, prod_Hom, left_eq_id,
+      dsimp only [Functor.id_obj, Functor.const_obj_obj, Functor.diag_obj, prod_Hom, left_eq_id,
         Functor.const_obj_map, prod_comp, prod_id_fst, prod_id_snd, Functor.diag_map, proj_obj,
         proj_map, eq_mp_eq_cast, id_eq, Functor.comp_obj, Under.forget_obj,
         Functor.toStructuredArrow_obj, Functor.toUnder_obj, mk_right, mk_hom_eq_self, Prod.mk.eta]
       apply eta) (by simp))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.comp_obj, proj_obj, Under.forget_obj, Functor.diag_obj,
+      dsimp only [Functor.comp_obj, proj_obj, Under.forget_obj, Functor.diag_obj,
         Functor.const_obj_obj, prod_Hom, left_eq_id, Functor.const_obj_map, prod_comp,
         Functor.id_obj, prod_id_fst, prod_id_snd, Functor.diag_map, proj_map, eq_mp_eq_cast, id_eq,
         Functor.toStructuredArrow_obj, Functor.toUnder_obj, mk_right, mk_hom_eq_self]
       symm
       apply eta) (by aesop_cat))
+
+def ofDiagEquivalence' (X : T × T) :
+    StructuredArrow X (Functor.diag _) ≌ StructuredArrow X.1 (Under.forget X.2) :=
+  (ofDiagEquivalence X).trans
+    <| (ofStructuredArrowProjEquivalence (𝟭 T) X.1 X.2).trans
+    <| StructuredArrow.mapNatIso (eqToIso (Functor.comp_id _))
 
 end StructuredArrow
 
@@ -729,13 +735,13 @@ def ofCostructuredArrowProjEquivalence {D : Type u₂} [Category.{v₂} D]
         (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
       (fun f => f.left.hom) (by simp))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.id_obj, proj_obj, Functor.const_obj_obj,
+      dsimp only [Functor.id_obj, proj_obj, Functor.const_obj_obj,
         proj_map, right_eq_id, Functor.const_obj_map, Functor.comp_obj, Functor.comp_map,
         eq_mp_eq_cast, id_eq, Over.forget_obj, Over.forget_map, Functor.toCostructuredArrow_obj,
         Functor.toOver_obj, mk_left, mk_hom_eq_self]
       apply eta) (by simp))
     (NatIso.ofComponents (fun x => by
-      simp only [Functor.comp_obj, Over.forget_obj, Functor.const_obj_obj, Functor.comp_map,
+      dsimp only [Functor.comp_obj, Over.forget_obj, Functor.const_obj_obj, Functor.comp_map,
         Over.forget_map, right_eq_id, Functor.const_obj_map, proj_obj, proj_map, eq_mp_eq_cast,
         id_eq, Functor.toCostructuredArrow_obj, Functor.toOver_obj, mk_left, mk_hom_eq_self,
         Functor.id_obj]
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,72 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+        (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.left.hom) (by simp))
+    (Functor.toCostructuredArrow
+      (Functor.toCostructuredArrow (CostructuredArrow.proj _ Y ⋙ Over.forget X) _ _
+        (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.left.hom) (by simp))
+    (by aesop_cat)
+    (by aesop_cat)
+
+/-- Characterization of the structured arrow category on the diagonal functor `T ⥤ T × T`. -/
```
TwoFX, line 712:
```suggestion
/-- Characterization of the costructured arrow category on the diagonal functor `T ⥤ T × T`. -/
```

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,72 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+    (Functor.toStructuredArrow
+      (Functor.toUnder (StructuredArrow.proj X _ ⋙ StructuredArrow.proj Y _) _
+        (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.right.hom) (by simp))
+    (Functor.toStructuredArrow
+      (Functor.toStructuredArrow (StructuredArrow.proj Y _ ⋙ Under.forget X) _ _
+         (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.right.hom) (by simp))
+    (by aesop_cat)
+    (by aesop_cat)
```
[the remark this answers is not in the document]
javra: It's kind of slow anyway, will replace it.

javra:
Is the placement of the constructions in `Over.lean` acceptable?

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,114 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+
+end StructuredArrow
+
+namespace CostructuredArrow
+
+/-- Characterization of the costructured arrow category on the projection functor of any
+costructured arrow category. -/
+def ofCostructuredArrowProjEquivalence {D : Type u₂} [Category.{v₂} D]
+    (F : T ⥤ D) (Y : D) (X : T) :
+    CostructuredArrow (CostructuredArrow.proj F Y) X ≌ CostructuredArrow (Over.forget X ⋙ F) Y :=
```
joelriou, line 721: It would be slightly more readable to use the following syntax:
```suggestion
    CostructuredArrow (CostructuredArrow.proj F Y) X ≌ CostructuredArrow (Over.forget X ⋙ F) Y where
  functor := sorry
  inverse := sorry
  unitIso := sorry
  counitIso := sorry
```

TwoFX: I would say that using `Equivalence.mk` is acceptable because it is a "smart" constructor that changes the unit and counit so that the triangle equality is automatically satisfied.

javra: Yes, that was a conscious choice, will try if proving the triangle equality is cheap, otherwise I'd keep it this way.

javra:
Also, there's versions of `ofDiagEquivalence` with `X.1` and `X.2` swapped, do we also want those? @TwoFX

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,114 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+    (Functor.toStructuredArrow
+      (Functor.toUnder (StructuredArrow.proj X _ ⋙ StructuredArrow.proj Y _) _
+        (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.right.hom) (by simp))
+    (Functor.toStructuredArrow
+      (Functor.toStructuredArrow (StructuredArrow.proj Y _ ⋙ Under.forget X) _ _
+         (fun g => by exact g.hom) (fun m => by have := m.w; aesop_cat)) _ _
+      (fun f => f.right.hom) (by simp))
+    (NatIso.ofComponents (fun x => by
+      simp only [Functor.id_obj, Functor.const_obj_obj, proj_obj, left_eq_id, Functor.const_obj_map,
```
TwoFX, line 676: Wait, isn't this still generating data using tactics? Can the `simp only` be replaced by `dsimp only` at least?

TwoFX:
> Also, there's versions of `ofDiagEquivalence` with `X.1` and `X.2` swapped, do we also want those? @TwoFX

It might be fun to deduce this by specializing your `StructuredArrow X (StructuredArrow.proj Y F) ≌ StructuredArrow Y (Under.forget X ⋙ F)` to `StructuredArrow X (Under.forget Y) ≌ StructuredArrow Y (Under.forget X)` by taking `F` to be the identity.

Mathlib/CategoryTheory/Comma/Over.lean
javra: You're right.... I tried that to make it faster than just to use `Iso.refl _`... Maybe I should try to piece the iso together instead...

javra:
> It might be fun to deduce this by specializing your `StructuredArrow X (StructuredArrow.proj Y F) ≌ StructuredArrow Y (Under.forget X ⋙ F)` to `StructuredArrow X (Under.forget Y) ≌ StructuredArrow Y (Under.forget X)` by taking `F` to be the identity.

Good idea, would have tried to do it with the swap endofunctor otherwise.

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -620,7 +620,7 @@ variable {S : Type u₂} [Category.{v₂} S]
 /-- Given `X : T`, to upgrade a functor `F : S ⥤ T` to a functor `S ⥤ Over X`, it suffices to
     provide maps `F.obj Y ⟶ X` for all `Y` making the obvious triangles involving all `F.map g`
     commute. -/
-@[simps! obj_left map_left]
+@[simps! obj_left map_left obj]
```
joelriou, line 623: This change is unnecessary (the code compile without it) and problematic: one argument is that `obj_left` would no longer be triggered automatically. (Same below.)

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,140 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
 
 end Functor
 
+namespace StructuredArrow
+
+/-- A functor from the structured arrow category on the projection functor for any structured
+arrow category. -/
+@[simps!]
+def ofStructuredArrowProjFunctorFrom {D : Type u₂} [Category.{v₂} D] (F : D ⥤ T) (Y : T) (X : D) :
```
joelriou, line 665: I would prefer names like:
```suggestion
def ofStructuredArrowProjEquivalence.functor {D : Type u₂} [Category.{v₂} D] (F : D ⥤ T) (Y : T) (X : D) :
```
(Also, as `D` is used multiple times, it could be declared as variable by the beginning of the file.)

Mathlib/CategoryTheory/Comma/Over.lean
joelriou: All the triangle equalities are proved automatically (and the file would take a very similar amount of time to compile on my computer). In https://github.com/leanprover-community/mathlib4/pull/15919 I removed almost all the unnecessary uses of `Equivalence.mk` (and surprisingly, it made mathlib faster! I do not know how...). I do not think we should speculate about whether nobody would ever need to unfold the definitions of the (co)unit isomorphisms.

Mathlib/CategoryTheory/Comma/Over.lean
javra: What would you call the inverse functors?

Mathlib/CategoryTheory/Comma/Over.lean
javra: An earlier version of the constructions below needed it, but it was a bad fix anyway since it relied on the `obj` simp having higher priority than the `obj_left` simp

Mathlib/CategoryTheory/Comma/Over.lean
joelriou: `ofStructuredArrowProjEquivalence.inverse`!

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +657,140 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+    (NatIso.ofComponents (fun _ => Iso.refl _) (by simp))
+    (NatIso.ofComponents (fun _ => Iso.refl _) (by aesop_cat))
+
+/-- A version of `StructuredArrow.ofDiagEquivalence` with the roles of the first and second
+projection swapped. -/
+def ofDiagEquivalence' (X : T × T) :
+    StructuredArrow X (Functor.diag _) ≌ StructuredArrow X.1 (Under.forget X.2) :=
+  (ofDiagEquivalence X).trans
+    <| (ofStructuredArrowProjEquivalence (𝟭 T) X.1 X.2).trans
+    <| StructuredArrow.mapNatIso (eqToIso (Functor.comp_id (Under.forget X.2)))
```
joelriou, line 721: As this is an infix operator `<|` should be before the linebreak.

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +644,136 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+  unitIso := NatIso.ofComponents (fun _ => Iso.refl _) (by simp)
+  counitIso := NatIso.ofComponents (fun _ => Iso.refl _) (by aesop_cat)
+
+/-- A version of `StructuredArrow.ofDiagEquivalence` with the roles of the first and second
+projection swapped. -/
+def ofDiagEquivalence' (X : T × T) :
+    StructuredArrow X (Functor.diag _) ≌ StructuredArrow X.1 (Under.forget X.2) :=
+  (ofDiagEquivalence X).trans <|
+    (ofStructuredArrowProjEquivalence (𝟭 T) X.1 X.2).trans <|
+    StructuredArrow.mapNatIso (eqToIso (Functor.comp_id (Under.forget X.2)))
```
joelriou, line 708:
```suggestion
    StructuredArrow.mapNatIso (Under.forget X.2).rightUnitor
```

Mathlib/CategoryTheory/Comma/Over.lean
```diff
@@ -657,4 +644,136 @@ lemma toUnder_comp_forget (F : S ⥤ T) (X : T) (f : (Y : S) → X ⟶ F.obj Y)
...
+  unitIso := NatIso.ofComponents (fun _ => Iso.refl _) (by simp)
+  counitIso := NatIso.ofComponents (fun _ => Iso.refl _) (by aesop_cat)
+
+/-- A version of `CostructuredArrow.ofDiagEquivalence` with the roles of the first and second
+projection swapped. -/
+def ofDiagEquivalence' (X : T × T) :
+    CostructuredArrow (Functor.diag _) X ≌ CostructuredArrow (Over.forget X.2) X.1 :=
+  (ofDiagEquivalence X).trans <|
+    (ofCostructuredArrowProjEquivalence (𝟭 T) X.1 X.2).trans <|
+    CostructuredArrow.mapNatIso (eqToIso (Functor.comp_id (Over.forget X.2)))
```
joelriou, line 775:
```suggestion
    CostructuredArrow.mapNatIso (Over.forget X.2).rightUnitor
```

joelriou:
Thanks!

bors merge

PR #15841, opened 2024-08-15
chore(Order/WellFounded): Use `WellFoundedLT` / `WellFoundedGT`

vihdzp:
This has two main benefits:
- `WellFoundedGT.max` has the correct name
- Typeclass inference allows us to golf a bunch of proofs.

---


vihdzp:
Truth being told, I'm not even sure whether we should have `min` at all. It's nearly always followed by `not_lt_min` with the same parameters, so we should probably just be using `has_min` instead.

PR #15842, opened 2024-08-15
[Merged by Bors] - fix: don't resynthesize type class arguments in `fun_prop`

lecopivo:
`fun_prop` was resynthesizing type class arguments the same way as  `simp` used to do. It seems to cause just troubles and seemingly no benefit so let's remove it from `fun_prop` too.


j-loreaux:
!bench

j-loreaux:
I don't understand at all why this would make anything *slower*. I will note that the places using `fun_prop` the most (continuous functional calculus files) did get marginally faster.

lecopivo:
I also find it baffling. My wild guess is that type classes are cached anyway so resynthesizing them is not such a big deal. However, you might end up with instances that are defeq but not trivially so type checking takes longer.

But I do not know what the "instructions" metric is.

j-loreaux:
Instructions are the best proxy we have for "amount of computation done independent of hardware"

lecopivo:
Do we want to merge this? Is the 16% increase on that one file a blocker? Should that be investigated?

j-loreaux:
I'm not entirely sure about this. It does speed things up a bit in the continuous functional calculus portion of the library, where we use `fun_prop` heavily, but I don't have a good understanding of the slowdown in the Sobolev file.

maintainer merge

mattrobball:
Overall instructions are basically flat. So from that perspective this is fine. I think the slowdown you see is exactly as @lecopivo surmised (and as core would probably chastise us for). But if this works where it needs to right now, I think it is ok to run with and evaluate as we go.

bors merge

sgouezel:
bors r-
Can you merge master, fix the lint, and then send it back to bors?
bors d+

PR #15843, opened 2024-08-15
Relax `IsSemiReal` typeclass constraint

artie2000:
Typeclass constraint on instance of `IsSemireal` relaxed from `LinearOrderedField` to `LinearOrderedRing`. This instance requires `LinearOrderedSemiring` and well as negation, so this is now the most general instance possible.

---

### DIFF
--- a/Mathlib/Algebra/Ring/Semireal/Defs.lean
+++ b/Mathlib/Algebra/Ring/Semireal/Defs.lean
@@ -42,6 +42,6 @@ class IsSemireal [AddMonoid R] [Mul R] [One R] [Neg R] : Prop where
 @[deprecated (since := "2024-08-09")] alias isSemireal.neg_one_not_SumSq :=
   IsSemireal.not_isSumSq_neg_one
 
-instance [LinearOrderedField R] : IsSemireal R where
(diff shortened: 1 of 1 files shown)


ScottCarnahan:
Perhaps this is out of scope for this PR, but it seems the definition of `IsSemireal` uses negation unnecessarily.
If we change the second condition `not_isSumSq_neg_one : ¬IsSumSq (-1 : R)` to something like
`not_exist_isSumSq_and_add_one : ¬∃ (x : R), IsSumSq x ∧ x + 1 = 0`
 then you can change to `[AddZeroClass R] [Mul R] [One R]` and the instance can be
```
instance [LinearOrderedSemiring R] [ExistsAddOfLE R]: IsSemireal R where
  non_trivial := zero_ne_one
  not_exist_isSumSq_and_add_one h := by
    obtain ⟨x, hx⟩ := h
    exact not_le.2 (lt_of_le_of_lt hx.1.nonneg (lt_add_one x)) (le_of_eq hx.2)
```

FM22:
I'll likely be refactoring this more widely, and I'll probably do something like this then yep.

PR #15845, opened 2024-08-15
[Merged by Bors] - chore: import enabled mathlib syntax linters in `Mathlib.Init`

grunweg:
These linters are meant to run on all of mathlib, but they need to be imported in order to run on a given file.
Hence, move them to `Mathlib.Init`, which has precisely this purpose.

All linters moved in this PR are syntax linters (which need to be imported to run) which are enabled by default.
Linters missing from this list include
- Linter/Textbased (which is not syntax-based)
- Linter/MinImports (which is off by default)
- Linter/HaveLet (which is disabled right now)

All linters only have imports in `Lean`, except for `Linter/Lint` (which imports `Batteries.Tactic.Lint` to define an environment linter). This import could be avoided by splitting the environment linters into a different file; there is no need to import these syntax linters this early.

---

`semorrison` [indicated](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Duplication.20in.20style.20linters/near/463325858) that these linters are fine to import for now.

Follow-up clean-ups, out of scope of this PR
- write a linter to ensure Mathlib.Init is actually imported in every file
- remove the text-based line length linter (which is now duplicate)
- adapt the in-flight PR for the file length linter


Mathlib/Init.lean
```diff
@@ -0,0 +1,25 @@
...
+
+-- All these linters are meant to run on all of mathlib, are active by default,
+-- and have no imports beyond `Lean`.
+import Mathlib.Tactic.Linter.OldObtain
+import Mathlib.Tactic.Linter.RefineLinter
+-- xxx: this file imports Batteries.Lean.HashSet; is this acceptable?
+import Mathlib.Tactic.Linter.HashCommandLinter
+import Mathlib.Tactic.Linter.GlobalAttributeIn
+-- xxx: this file imports Batteries.Data.String.Matcher and Batteries.Tactic.Lint; acceptable?
+import Mathlib.Tactic.Linter.Lint
```
adomani, line 22: Similarly here: the string matcher is there since one of the linters currently uses `findSubstring` or something like this.  I think that a previous version of the linter used `splitOn` and counted the pieces: that would allow to reduce the imports.

Mathlib/Init.lean
```diff
@@ -0,0 +1,25 @@
...
+-- xxx: this files imports Batterires.Tactic.Unreachable; is this acceptable?
+import Mathlib.Tactic.Linter.UnusedTactic
```
adomani, line 24: The "unreachable" dependence is just for some name resolution double ticks to work: making those silently not exist using a single tick would allow removing the dependence.  In this case, I am not completely sure how the behaviour of the linter would change.

Mathlib/Init.lean
```diff
@@ -0,0 +1,25 @@
...
+* strong preference for avoiding files that themselves have imports beyond `Lean`, and
+  any exception to this rule should by accompanied by a comment explaining the transitive imports.
+-/
+
+-- All these linters are meant to run on all of mathlib, are active by default,
+-- and have no imports beyond `Lean`.
+import Mathlib.Tactic.Linter.OldObtain
+import Mathlib.Tactic.Linter.RefineLinter
+-- xxx: this file imports Batteries.Lean.HashSet; is this acceptable?
+import Mathlib.Tactic.Linter.HashCommandLinter
```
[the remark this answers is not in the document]
grunweg: Agreed about inlining - thanks for the context!

grunweg:
I pushed a commit fixing almost all tests. `tests/MinImports.lean` needs more careful thought. @adomani Do you have a moment to take a look?

grunweg:
!bench

adomani:
I'm not at a computer, but is the error just some whitespace noise?  I'm only going by what I can see in the CI report, though.

adomani:
Michael, thank you very much for using the file already!

I took the liberty of fixing the failing `MinImports` test and, while I was at it, I also reduced the imports in the remaining linters.

I think that all transitive imports are acceptable now, but let's see what the others think about it!

adomani:
By the way, I am not completely sure about what changed with `MinImports`, but it is something that I had observed before.  I think that there are two paths to make the syntax typecheck and Lean genuinely uses one or the other depending on what is available.  I do not know why Lean now prefers the "shorter" path, but I would not worry about it too much.

Mathlib/Init.lean
adomani: Looking at the code, those names only play a role when they are available and do not cause any concern otherwise, so I removed the import and used single ticks.

Mathlib/Init.lean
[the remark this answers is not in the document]
adomani: I went for `Array` and reduced the imports.

Mathlib/Init.lean
adomani: Sure, if we can avoid increasing imports, I am happy with a different solution.  The one using `splitOn` is very low tech and yet robust, though.

adomani:
Oops, I merged master by mistake, sorry!

By the way, I think that back-porting all the linting changes would be welcome!

grunweg:
Thanks for quickly fixing the test, and tweaking the linters. I just filed #15868 and #15869 for backporting the linter fixes.

grunweg:
I have added a comment inlining Mathlib.Tactic.Linter: it only contained a single import now, and was only meaningfully imported in Mathlib.Tactic.Common. (And, to be honest, I'm not convinced importing it in `Tactic.Basic` is not a better choice... but that's for another thread.)

grunweg:
The first PR has landed; I have just rebased on master.

Mathlib/Tactic/Linter.lean
joneugster, line 1: I don't know what the mathlib standard is for this. My gut tells me that `Mathlib/Tactic/Linter.lean` should keep existing, maybe be autogenerated, still import all linters, and keep being imported in `Mathlib.Tactic.Common` (or `Mathlib.Tactic.Basic`).

Mathlib/Tactic/Common.lean
```diff
@@ -61,7 +61,10 @@ import Mathlib.Tactic.InferParam
 import Mathlib.Tactic.Inhabit
 import Mathlib.Tactic.IrreducibleDef
 import Mathlib.Tactic.Lift
-import Mathlib.Tactic.Linter
+-- Import any linters which are not imported in `Mathlib.Init`, such as
+-- syntax linters which are disabled by default in mathlib.
+import Mathlib.Tactic.Linter.HaveLetLinter
+
```
joneugster, line 67: (see comment below)

Mathlib/Init.lean
```diff
@@ -1,4 +1,4 @@
-/-!
+/-
```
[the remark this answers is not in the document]
grunweg: That is a fair point. There is an annoying detail that imports must come before module docstrings, but I can be convinced that the doc generation page is a better argument.

Mathlib/Init.lean
```diff
@@ -10,3 +10,15 @@ Here are some general guidelines:
 * strong preference for avoiding files that themselves have imports beyond `Lean`, and
   any exception to this rule should by accompanied by a comment explaining the transitive imports.
 -/
+
+-- All these linters are meant to run on all of mathlib,
+-- are syntax linters active by default (no need to be imported as early as possible),
```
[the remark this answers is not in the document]
grunweg: Oops, that is a negation too much. I have reworded this comment.

Mathlib/Tactic/Linter.lean
grunweg: Hm. From a code writing perspective, an essentially empty file feels silly. This can be nice from a documentation perspective, though...

Ultimately, I would like to have a place to put a big comment documenting the linters, what they do and their current state (like [here](https://github.com/leanprover-community/mathlib4/blob/c64a270a650434893fc6da7c434448ff01a344b0/Mathlib/Init.lean)). If you think `Tactic.Linter.lean` is a good place, I can get behind that.
I don't feel strongly about this.

Having an essentially empty file feels silly; I agree

grunweg:
!bench

Mathlib/Init.lean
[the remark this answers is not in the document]
joneugster: What speaks against just doing this:

```lean
import Mathlib.Tactic.Linter.HashCommandLinter
import Mathlib.Tactic.Linter.GlobalAttributeIn
-- This file imports Batteries.Tactic.Lint, where the `env_linter` attribute is defined.
import Mathlib.Tactic.Linter.Lint
import Mathlib.Tactic.Linter.OldObtain
import Mathlib.Tactic.Linter.RefineLinter
import Mathlib.Tactic.Linter.UnusedTactic
import Mathlib.Tactic.Linter.Style

/-!
This is the root file in Mathlib: it is imported by virtually *all* Mathlib files.
For this reason, the imports of this files are carefully curated.
Any modification involving a change in the imports of this file should be discussed beforehand.
Here are some general guidelines:
* no bucket imports (e.g. `Batteries`/`Lean`/etc);
* every import needs to have a comment explaining why the import is there;
* strong preference for avoiding files that themselves have imports beyond `Lean`, and
  any exception to this rule should by accompanied by a comment explaining the transitive imports.

Note: All linters imported here have no bulk imports;
**Not** imported in this file are
- the text-based linters in `Linters/TextBased.lean`, as they can be imported later
- the `minImports` linter, as that is still disabled by default.
It *does* check a useful global coherence property, so perhaps changing this default at some
point can be useful. `mathlib` might need a fair amount of fixes for this to become viable, though.
-- XXX: the `haveLet` linter could also be imported here...
-/

```

Mathlib/Tactic/Linter.lean
joneugster: it's not that uncommon to have a file to support `import Mathlib.Tactic.Linter`, that's analogue to `import Mathlib.Tactic`, and indeed, the module docstring of that file would probably be a good place for a manual documentation.

Anyways, I think if you want to remove `Mathlib/Tactic/Linter.lean` I think you should ask some maintainers or Zulip about their opinion, I don't think I should be the one calling a decision here :)

grunweg: Fair point; I have restored that file now.

Mathlib/Init.lean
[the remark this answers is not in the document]
grunweg: This is a good solution; I've switched to it. Thanks!

grunweg:

joneugster (approved):
Thanks!

maintainer merge

grunweg:
Thanks. The build failure should be fixed now; running CI to confirm.

grunweg:
CI passes now :tada:

semorrison:
bors merge

test/fail_if_no_progress.lean
```diff
@@ -2,6 +2,7 @@ import Mathlib.Tactic.FailIfNoProgress
 import Mathlib.Tactic.Basic
 
 set_option linter.unusedVariables false
+set_option linter.stylesetOption false
```
grunweg, line 5:
```suggestion
set_option linter.style.setOption false

grunweg:
All CI steps pass (except linting, which passed in the previous commit). Thanks for the review!
bors merge

PR #15848, opened 2024-08-15
feat(CI): automatic labeling of PRs

adomani:
This PR adds a CI action that tries to automatically label PRs with the appropriate `t-xxx` labels.

The script contains a "dictionary" to convert modified folders into `t-xxx` labels.
The script finds all possible labels for the files modified by the PR and, if there is exactly one possible label, applies it.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Auto-label.20PR.20topics)

---

In case it helps with reviewing, the action is adapted from the new contributor action.

### DIFF
--- a/.github/workflows/autolabel_action.yaml
+++ b/.github/workflows/autolabel_action.yaml
@@ -20,7 +20,7 @@ jobs:
 
       - run: |
           ./scripts/autolabels.sh long
-          printf 'Just the labels: "%s"\n' "$(./scripts/autolabels.sh)"
+          printf 'About to add the following labels: "%s"\n' "$(./scripts/autolabels.sh)"
           LABELS="$(./scripts/autolabels.sh)"
           printf $'gh issue edit \'%s\' --add-label \'%s\'\n' "$NUMBER" "$LABELS"
           if [ -z "$LABELS" ]
--- a/.github/workflows/autolabel_action.yaml
+++ b/.github/workflows/autolabel_action.yaml
@@ -1,0 +1,39 @@
+name: Apply at most one `t-`label to PRs
+
+on: pull_request
+
+jobs:
+  label-and-report-new-contributor:
+    runs-on: ubuntu-latest
+    # Don't run on forks, where we wouldn't have permissions to add the label anyway.
+    if: github.repository == 'leanprover-community/mathlib4'
+    permissions:
+      issues: write
+      checks: write
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - run: |
+          ./scripts/autolabels.sh long
+          printf 'Just the labels: "%s"\n' "$(./scripts/autolabels.sh)"
+          LABELS="$(./scripts/autolabels.sh)"
+          printf $'gh issue edit \'%s\' --add-label \'%s\'\n' "$NUMBER" "$LABELS"
+          if [ -z "$LABELS" ]
+          then
+            printf 'No labels found\n'
+          elif [[ $LABELS != *,* ]]
+          then
+            printf 'Actually doing it!\n'
+            gh issue edit "$NUMBER" --add-label "$LABELS"
+          else
+            printf 'More than one label found: not adding a label\n'
+          fi
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.number }}
--- a/.github/workflows/autolabel_action.yaml
+++ b/.github/workflows/autolabel_action.yaml
@@ -3,7 +3,7 @@ name: Apply at most one `t-`label to PRs
 on: pull_request
 
 jobs:
-  label-and-report-new-contributor:
+  apply_one_t_label:
     runs-on: ubuntu-latest
     # Don't run on forks, where we wouldn't have permissions to add the label anyway.
     if: github.repository == 'leanprover-community/mathlib4'
--- a/scripts/autolabels.sh
+++ b/scripts/autolabels.sh
@@ -29,7 +29,7 @@ from `Tactic/anything_other_than_linter`.
 |t-linter                | Tactic/Linter                                                                     |
 |t-logic                 | Logic, ModelTheory                                                                |
 |t-measure-probability   | MeasureTheory, Probability, InformationTheory                                     |
-|t-meta                  | Tactic, Lean, Util, Mathport, Control, Testing                              |
+|t-meta                  | Tactic, Lean, Util, Mathport, Control, Testing                                    |
 |t-number-theory         | NumberTheory                                                                      |
 |t-order                 | Order                                                                             |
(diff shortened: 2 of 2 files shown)


.github/workflows/autolabel_action.yaml
```diff
@@ -0,0 +1,39 @@
...
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - run: |
+          ./scripts/autolabels.sh long
+          printf 'Just the labels: "%s"\n' "$(./scripts/autolabels.sh)"
```
grunweg, line 23: Let's make this easier to understand for first-time readers:
```suggestion
          printf 'About to add the following labels: "%s"\n' "$(./scripts/autolabels.sh)"

File: .github/workflows/autolabel_action.yaml [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+          LABELS="$(./scripts/autolabels.sh)"
+          printf $'gh issue edit \'%s\' --add-label \'%s\'\n' "$NUMBER" "$LABELS"
```
grunweg, line 25: How about moving this to the branch actually adding the labels, such as printing
"About to add labels by calling `'gh issue edit \'%s\' --add-label \'%s\'\n' "$NUMBER" "$LABELS"`"

scripts/autolabels.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
+
+ : <<'BASH_MODULE_DOCS'
+
+This file contains the dictionary for converting a modified dir into a github label,
```
grunweg, line 5: What do you think? Optional.
```suggestion
This file contains the dictionary for converting a list of modified directories into a github label,

File: scripts/autolabels.sh [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+as well as the script to perform the conversion.
```
grunweg, line 6:
```suggestion
as well as a script to actually compute which label(s) to apply to the current changes.

File: scripts/autolabels.sh [hunk 5]
```diff
@@ -0,0 +1,94 @@
...
+
+Rules:
```
grunweg, line 8: How about "Dictionary entries must follow the following rules:" or so?

scripts/autolabels.sh
```diff
@@ -0,0 +1,94 @@
...
+* only lines starting with `|t-` are considered;
+* repeated spaces, commas and pipes (|) are ignored;
+* `Mathlib` is added by default to each `Root folder`.
```
grunweg, line 11:
```suggestion
* `Mathlib` is added by default to each "root folder".

File: scripts/autolabels.sh [hunk 7]
```diff
@@ -0,0 +1,94 @@
...
+
+Exception: `Tactic/Linter` is converted into `Tactic-Linter` internally, to simplify separating it from
```
grunweg, line 13: "to easily distinguish it from `Tactic/anything_other_than_linter`"?

scripts/autolabels.sh
```diff
@@ -0,0 +1,94 @@
...
+`Tactic/anything_other_than_linter`.
+
+|Label                   |Root folders                                                                       |
+|-                       |-                                                                                  |
+|t-algebra               | Algebra, FieldTheory, RingTheory, GroupTheory, RepresentationTheory, LinearAlgebra|
+|t-algebraic-geometry    | AlgebraicGeometry, Geometry/RingedSpace                                           |
+|t-euclidean-geometry    | Geometry.Euclidean                                                                |
```
grunweg, line 20: Is this a typo, and should be `Geometry/Euclidean`?

scripts/autolabels.sh
```diff
@@ -0,0 +1,94 @@
...
+|t-category-theory       | CategoryTheory                                                                    |
+|t-combinatorics         | Combinatorics                                                                     |
+|t-computability         | Computability                                                                     |
+|t-condensed             | Condensed                                                                         |
+|t-data                  | Data                                                                              |
+|t-dynamics              | Dynamics                                                                          |
+|t-linter                | Tactic/Linter                                                                     |
+|t-logic                 | Logic, ModelTheory                                                                |
+|t-measure-probability   | MeasureTheory, Probability, InformationTheory                                     |
+|t-meta                  | Tactic, Lean, Init, Util, Mathport, Control, Testing                              |
```
grunweg, line 32: I would rather *not* apply `Init` here: that directory is a random mix of things.

grunweg: Sorry for bringing this up now...

scripts/autolabels.sh
```diff
@@ -0,0 +1,94 @@
...
+|t-number-theory         | NumberTheory                                                                      |
+|t-order                 | Order                                                                             |
+|t-topology              | Topology, AlgebraicTopology                                                       |
+|t-set-theory            | SetTheory                                                                         |
+
+BASH_MODULE_DOCS
+
+# strips replaces repeated `|, ` with a single space, then only keeps lines beginning with `t-`
```
grunweg, line 40: This sentence does not parse. Do you mean to delete the word "strips"?

grunweg (approved):
The CI job looks plausible, the dictionary is well documented. (I have a few wording suggestions as usual.) My awk is still not that good - the printing logic in the end makes sense, but I can only guess what the middle part does.

I do wonder if writing a shell script with heavy awk is a good decision from the maintainance point of view. On the other hand, this script does not strike me as needing much maintenance (and should not be difficult to recreate, in the hopefully never occurring case that is needed).

scripts/autolabels.sh
adomani: I like this, thanks!

scripts/autolabels.sh
adomani: Also an improvement!

scripts/autolabels.sh
adomani: Indeed, it is a typo!  However, it would have probably matched regardless, since the matching is via regex and `.` matches anything!  :smile:

scripts/autolabels.sh
adomani: Removed it!  I should have removed it to begin with, but I was more focused on getting the script work, than actually looking at the labeling.

scripts/autolabels.sh
adomani: Removed `strips`, thanks!

.github/workflows/autolabel_action.yaml
adomani: I need to experiment with this: `$NUMBER` is actually read off via the workflow API and I do not really know if I can use it in a script that the API calls.

Besides, I am planning to rewrite the whole script in Lean, since I think that it will be easier to understand and maintain.

grunweg: Waking up, I was wondering how long rewriting this in Python or Lean would take.
A Lean rewrite would be perfect!

adomani:
Eventually, #17062 got merged.

PR #15850, opened 2024-08-15
[Merged by Bors] - chore: follow the "reducible non-instances" library note

j-loreaux:
This marks some non-instance `def`s as `abbrev`s, per the relevant library note. Without this, we're placing default transparency roadblocks whenever these `def`s get used to create specific `instance`s on types.

---

This recently caused me some unnecessary pain.


j-loreaux:
!bench

dupuisf:
Thanks!

bors r+

mattrobball:
bors delegate+

j-loreaux:
bors merge

PR #15851, opened 2024-08-15
[Merged by Bors] - feat: generators for SL(2, Z)

CBirkbeck:
We show that SL(2, Z) is generated by S and T.

---

- [x] depends on: #16159 
- [x] depends on : #16160 


Mathlib/LinearAlgebra/Matrix/FixedDetMatrices.lean
```diff
@@ -242,6 +244,45 @@ theorem induction_on {C : Δ m → Prop} {A : Δ m} (hm : m ≠ 0)
...
 
 end FixedDetMatrices
+
+open MatrixGroups FixedDetMatrices
+
+section SL2Z_generators
+
+/-- `SL(2, ℤ)` is generated by `S` and `T`. -/
+lemma SpecialLinearGroup.SL2Z_generators : S_T_subgroup = ⊤ := by
+  refine (Subgroup.eq_top_iff' S_T_subgroup).mpr (fun _ => det_one_mem_S_T_subgroup _)
```
jcommelin, line 286: I'm inclined to inline `S_T_subgroup`. And maybe even the entire proof.
Or do you think it will be valuable to have `S_T_subgroup` as a standalone definition later on?

jcommelin (commented):
Thanks :tada:

bors merge

Mathlib/LinearAlgebra/Matrix/FixedDetMatrices.lean
CBirkbeck: So when doing something like `Subgroup.closure_induction` one needs to give it the set, so for that its maybe useful? I'm picking `Subgroup.closure_induction` since I was using it yesteday in this context.

tb65536: Even if one is using `Subgroup.closure_induction`, I don't see why `closure {S, T} = ⊤` should be any harder to use than `S_T_subgroup = ⊤`. I've opened up #20086, but feel free to disagree.

CBirkbeck: No that's fine, I don't have a strong preference either way!

PR #15858, opened 2024-08-15
[Merged by Bors] - chore: remove unnecessary references to Nat.zero_eq

Ruben-VandeVelde:
The induction tactics no longer put a Nat.zero into the context.

---


eric-wieser (approved):
bors d+

Thanks!

PR #15862, opened 2024-08-16
[Merged by Bors] - feat(Filter/Pointwise): add `one_prod` and `prod_one`

urkud:
Similar lemmas about `pure` are already in the default `simp` set.

---

- [x] depends on: #15922 


grunweg (commented):
Looks reasonable, thanks!
maintainer merge

PR #15863, opened 2024-08-16
[Merged by Bors] - feat(Order/SuccPred/CompleteLinearOrder): `ConditionallyCompleteLinearOrder.toSuccOrder`

vihdzp:
Every complete linear order with well-founded `<` is a successor order.

We also change some `LinearOrder α` + `IsWellOrder α (<)` to `LinearOrder α` + `WellFoundedLT α`. Note that these instances are mathematically equivalent, but the former has redundant assumptions.

---

### DIFF
--- a/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
+++ b/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
@@ -60,6 +60,8 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
     ∃ i, f i = x := hf.mem_of_nonempty_of_not_isPredLimit (Set.range_nonempty f) hx
 
 open Classical in
+/-- Every conditionally complete linear order is a successor order, by setting the succesor of an
+element to the infimum of all larger elements. -/
 noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [WellFoundedLT α] :
     SuccOrder α where
   succ a := if IsMax a then a else sInf {b | a < b}
--- a/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
+++ b/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
@@ -61,7 +61,7 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
 
 open Classical in
 /-- Every conditionally complete linear order with well-founded `<` is a successor order, by setting
-the succesor of an element to the infimum of all larger elements. -/
+the successor of an element to be the infimum of all larger elements. -/
 noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [WellFoundedLT α] :
     SuccOrder α where
   succ a := if IsMax a then a else sInf {b | a < b}
--- a/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
+++ b/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
@@ -60,7 +60,7 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
     ∃ i, f i = x := hf.mem_of_nonempty_of_not_isPredLimit (Set.range_nonempty f) hx
 
 open Classical in
-noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [IsWellOrder α (· < ·)] :
+noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [WellFoundedLT α] :
     SuccOrder α where
   succ a := if IsMax a then a else sInf {b | a < b}
   le_succ a := by
--- a/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
+++ b/Mathlib/Order/SuccPred/CompleteLinearOrder.lean
@@ -76,14 +76,12 @@ noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [WellFoundedLT α
     simp [h] at hs
     rw [not_isMax_iff] at h
     exact hs.not_lt (csInf_mem h)
-  le_of_lt_succ := by
-    intro b a hs
+  le_of_lt_succ {b a} hs := by
     by_cases h : IsMax a
     · simpa [h] using le_of_lt hs
     · simp [h] at hs
       simpa using not_mem_of_lt_csInf hs ⟨a, fun _ hc => hc.le⟩
-  succ_le_of_lt := by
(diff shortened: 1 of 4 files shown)


Mathlib/Order/SuccPred/CompleteLinearOrder.lean
```diff
@@ -59,6 +59,34 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
...
+    · simp only [h, ↓reduceIte]
+      rw [not_isMax_iff] at h
+      exact le_csInf h (fun b => le_of_lt)
+  max_of_succ_le hs := by
+    by_contra h
+    simp [h] at hs
+    rw [not_isMax_iff] at h
+    exact hs.not_lt (csInf_mem h)
+  le_of_lt_succ := by
+    intro b a hs
```
YaelDillies, line 80:
```suggestion
  le_of_lt_succ b a hs := by
```

Mathlib/Order/SuccPred/CompleteLinearOrder.lean
```diff
@@ -59,6 +59,34 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
...
+    by_cases h : IsMax a
+    · simpa [h] using le_of_lt hs
+    · simp [h] at hs
+      simpa using not_mem_of_lt_csInf hs ⟨a, fun _ hc => hc.le⟩
+  succ_le_of_lt := by
+    intro a b ha
```
YaelDillies, line 86:
```suggestion
  succ_le_of_lt a b ha := by
```

vihdzp: This doesn't work, unfortunately. I think implicit variables might be the issue.

YaelDillies:
```suggestion
  succ_le_of_lt {a b} ha := by
```
then

Mathlib/Order/SuccPred/CompleteLinearOrder.lean
```diff
@@ -59,6 +59,32 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
     (hf : IsGLB (Set.range f) x) (hx : ¬ IsPredLimit x) :
     ∃ i, f i = x := hf.mem_of_nonempty_of_not_isPredLimit (Set.range_nonempty f) hx
 
+open Classical in
+/-- Every conditionally complete linear order with well-founded `<` is a successor order, by setting
+the succesor of an element to the infimum of all larger elements. -/
```
YaelDillies, line 64:
```suggestion
the successor of an element to be the infimum of all larger elements. -/
```

Mathlib/Order/SuccPred/CompleteLinearOrder.lean
```diff
@@ -59,6 +59,32 @@ lemma IsGLB.exists_of_nonempty_of_not_isPredLimit
...
+noncomputable def ConditionallyCompleteLinearOrder.toSuccOrder [WellFoundedLT α] :
+    SuccOrder α where
+  succ a := if IsMax a then a else sInf {b | a < b}
```
YaelDillies, line 67: Something I don't understand: Why not just
```suggestion
noncomputable def WellFoundedLT.toSuccOrder [WellFoundedLT α] :
    SuccOrder α where
  succ a := if ha : IsMax a then a else WellFounded.min (by simpa ha)
```
so that you don't even have to assume the order is conditionally complete?

vihdzp: The definition with `WellFoundedLT` is a bit weird conceptually. `succ a` would return an entirely arbitrary element covering `a`. But yes, it could be done.

We could even set this definition up in terms of `WellFounded.succ`(which I guess shows it does have some usage). Then I could add a comment on that definition explaining that if your well-order relation is `<`, you should use this instance instead.

Proving `Cardinal.succ_def` would become more annoying though. We would have to add the theorem `WellFounded.succ a = sInf {b | b < a}`, which does require a conditional linear order lattice.

YaelDillies: > The definition with `WellFoundedLT` is a bit weird conceptually. `succ a` would return an entirely arbitrary element covering `a`.

But that element is unique! (at least if your order is complete)

vihdzp: Of course, the definition is well-behaved if you enforce that additional restriction. In an order that isn't complete, this instance would have little computational value. For instance, the successor of `{Ø}` under `ZFSet` membership could be either of `{Ø, {Ø}}` or `{{Ø}}` with no way to tell.

By the way, I'm finding `WellFounded.succ` annoying to work with. The definition cases on whether `∃ b, r a b` which is a different (albeit equivalent) spelling to `¬ IsMax a`. So this might be yet another reason to ditch that definition.

vihdzp: I've just discovered a pretty big issue: this definition doesn't work on partial orders! Consider the partial order `a, b < c`. Then `a < succ b` but `¬ a ≤ b`.

In fact, isn't it the case that a partial SuccOrder has to be the direct sum of linear SuccOrders? Such a relation can't be well-founded unless it's well-ordered, in which case we land on the definition I already have.

YaelDillies: > A linear order with well-founded `<` is always a conditionally complete linear order, so we would gain no generality (and some annoyance)

Which spelling does mathlib use the most?

vihdzp: We have some API for the latter, see e.g. [`csInf_mem`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/ConditionallyCompleteLattice/Basic.html#csInf_mem). I don't think we have much for the former, except for the [IsWellOrderLimitElement](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/IsWellOrderLimitElement.html) file I've complained about for this exact reason.

In any case, we should always prefer the latter spelling, as it gives us access to `Inf` and `Sup` notation and all the lemmas that exist for it.

YaelDillies: Okay, I agree

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #9781, opened 2024-01-16
[Merged by Bors] - feat: bijection between Dyck words and rooted binary trees

Parcly-Taxel:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15864, opened 2024-08-16
[Merged by Bors] - feat: first return of a Dyck word

Parcly-Taxel:
Second part of #9781. This also includes a `PartialOrder` instance on Dyck words directly reflecting the tree bijection, that in turn allows a `positivity` extension for `firstReturn`.

---


Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
 
 end Semilength
 
+section FirstReturn
+
+/-- `p.firstReturn` is 0 if `p = 0` and the index of the `D` matching the initial `U` otherwise. -/
+def firstReturn : ℕ :=
+  (range p.toList.length).findIdx (fun i ↦
+    (p.toList.take (i + 1)).count U = (p.toList.take (i + 1)).count D)
```
YaelDillies, line 205:
```suggestion
  (range p.toList.length).findIdx fun i ↦
    (p.toList.take (i + 1)).count U = (p.toList.take (i + 1)).count D
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+
+@[simp] lemma firstReturn_zero : firstReturn 0 = 0 := rfl
+
+lemma firstReturn_pos (h : p.toList ≠ []) : 0 < p.firstReturn := by
```
YaelDillies, line 209:
```suggestion
lemma firstReturn_pos (h : p ≠ 0) : 0 < p.firstReturn := by
```
right? And everywhere else

YaelDillies, line 209: Once you have applied the other comment, this lemma will be in a form that positivity understands. Could you write the corresponding positivity extension? Feel free to cargo-cult an existing one

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+lemma firstReturn_lt_length (h : p.toList ≠ []) : p.firstReturn < p.toList.length := by
+  have lp := length_pos_of_ne_nil h
+  rw [← length_range p.toList.length]
+  apply findIdx_lt_length_of_exists
+  simp only [mem_range, decide_eq_true_eq]
+  use p.toList.length - 1
+  exact ⟨by omega, by rw [Nat.sub_add_cancel lp, take_of_length_le (le_refl _),
+    p.count_U_eq_count_D]⟩
+
+lemma count_eq_count_of_take_firstReturn_add_one (h : p.toList ≠ []) :
```
YaelDillies, line 226: `take_firstReturn_add_one` is not an assumption, so this is a bad lemma name. What about this instead?
```suggestion
lemma count_take_firstReturn_add_one (h : p.toList ≠ []) :
```
The full name would be something like
```suggestion
lemma count_take_firstReturn_add_one_U_eq_count_take_firstReturn_add_one_D (h : p.toList ≠ []) :
```
but that's too long to be memorable

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+        beq_iff_eq, ite_false, take_append_eq_append_take,
+        show j - p.toList.length = 0 by omega, take_zero, append_nil]
+      have := p.count_D_le_count_U j
+      omega
+  · simp_rw [length_range, u, length_append, length_cons]
+    exact Nat.lt_add_one _
+
+/-- The right part of the Dyck word decomposition. -/
+def rightPart (h : p.toList ≠ []) : DyckWord :=
+  p.drop (p.firstReturn + 1) (p.count_eq_count_of_take_firstReturn_add_one h)
```
YaelDillies, line 274: Can you just return `0` if `p = 0`? In fact, do you think there is a reasonable choice of junk value for `DyckWord.drop` when the subword isn't balanced?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+      simp only [take, length_take, lt_min_iff] at ub ⊢
+      replace ub := ub.1
+      rw [take_take, min_eq_left ub.le]
+      rw [show i = i - 1 + 1 by omega] at ub ⊢
+      rw [Nat.add_lt_add_iff_right] at ub
+      exact p.count_lt_count_of_lt_firstReturn _ ub)
+
+@[simp]
+theorem leftPart_nest_add_rightPart (h : p.toList ≠ []) :
+    (p.leftPart h).nest + p.rightPart h = p := by
```
YaelDillies, line 289: WHy is `leftPart` denested but not `rightPart`? I find this confusing

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+  rw [DyckWord.ext_iff, leftPart, denest_nest]
+  apply take_append_drop
+
+lemma leftPart_length_lt (h : p.toList ≠ []) : (p.leftPart h).toList.length < p.toList.length := by
```
YaelDillies, line 293:
```suggestion
lemma length_leftPart_lt (h : p.toList ≠ []) : (p.leftPart h).toList.length < p.toList.length := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+  nth_rw 2 [← p.leftPart_nest_add_rightPart h]
+  change _ < (U :: (p.leftPart h).toList ++ [D] ++ (p.rightPart h).toList).length
+  simp_rw [length_append, length_singleton, length_cons]; omega
+
+theorem leftPart_semilength_lt (h : p.toList ≠ []) : (p.leftPart h).semilength < p.semilength := by
```
YaelDillies, line 298:
```suggestion
theorem semilength_leftPart_lt (h : p.toList ≠ []) : (p.leftPart h).semilength < p.semilength := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +197,139 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+  change _ < (U :: (p.leftPart h).toList ++ [D] ++ (p.rightPart h).toList).length
+  simp_rw [length_append, length_singleton, length_cons]; omega
+
+theorem rightPart_semilength_lt (h : p.toList ≠ []) :
+    (p.rightPart h).semilength < p.semilength := by
+  rw [← Nat.mul_lt_mul_left zero_lt_two]
+  simp_rw [two_mul_semilength_eq_length]
+  exact p.rightPart_length_lt h
+
+lemma nest_add_ne_empty : (p.nest + q).toList ≠ [] := by change U :: _ ++ _ ≠ []; simp
```
YaelDillies, line 315: This belongs earlier, to the `nest` API

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: This is the exact unambiguous inductive construction of Dyck words (see the last part of #9781).

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: In order to write such an extension I need a partial order on Dyck words. There appears to be no suitable such partial order.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
YaelDillies: I mean I understand this, but naming them `leftPart` and `rightPart` is misleading in that the reader might expect them to be exactly analogous while there is actually a difference between the two. Can you find better, less, symmetrical, names?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -78,46 +78,52 @@ instance : AddMonoid DyckWord where
 
 namespace DyckWord
 
-variable (p q : DyckWord)
+variable {p q : DyckWord}
 
 lemma toList_eq_nil : p.toList = [] ↔ p = 0 := by rw [DyckWord.ext_iff]; rfl
+lemma toList_ne_nil : p.toList ≠ [] ↔ p ≠ 0 := by rw [not_iff_not, toList_eq_nil]
```
YaelDillies, line 84:
```suggestion
lemma toList_ne_nil : p.toList ≠ [] ↔ p ≠ 0 := toList_eq_nil.ne
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -78,46 +78,52 @@ instance : AddMonoid DyckWord where
...
-  nth_rw 2 [← p.toList.dropLast_append_getLast h, ← p.toList.dropLast.take_append_drop 1]
-  rw [p.getLast_eq_D, drop_one, this, p.head_eq_U]
+  nth_rw 2 [← p.toList.dropLast_append_getLast h', ← p.toList.dropLast.take_append_drop 1]
+  rw [getLast_eq_D h, drop_one, this, head_eq_U h]
   rfl
 
+variable (p) in
 /-- Prefix of a Dyck word as a Dyck word, given that the count of `U`s and `D`s in it are equal. -/
-def take (i : ℕ) (hi : (p.toList.take i).count U = (p.toList.take i).count D) : DyckWord where
+def take {i : ℕ} (hi : (p.toList.take i).count U = (p.toList.take i).count D) : DyckWord where
```
YaelDillies, line 118: It's a bad idea to make `i` implicit here.
1. It's much easier to provide the proof of `(p.toList.take i).count U = (p.toList.take i).count D` if you know what `i` is
2. You don't want the user seeing `p.take ...` in the infoview (since proofs are ellipsed) and having to click through to find out what `i` is

Parcly-Taxel: Yeah, I eventually deemed that as a mistake. This will be reverted in the next commit.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: I've done the first modification, but I answer no to the second question; in order to preserve `p.take i + p.drop i = p` for all Dyck words and have both summands also Dyck words, if `p` is of the form `q` then one of `p.take i` and `p.drop i` must be empty and the other must be the whole word, and there is no preference for either choice.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: I should have answered all your queries.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -77,43 +78,49 @@ instance : AddMonoid DyckWord where
...
+variable {p q : DyckWord}
 
 lemma toList_eq_nil : p.toList = [] ↔ p = 0 := by rw [DyckWord.ext_iff]; rfl
+lemma toList_ne_nil : p.toList ≠ [] ↔ p ≠ 0 := toList_eq_nil.ne
+
+variable (h : p ≠ 0)
 
 /-- The first element of a nonempty Dyck word is `U`. -/
-lemma head_eq_U (h : p.toList ≠ []) : p.toList.head h = U := by
+lemma head_eq_U : p.toList.head (toList_ne_nil.mpr h) = U := by
```
YaelDillies, line 89: Having `h : p ≠ 0` instead of `h : p.toList ≠ []` means that you will need to provide the hypothesis when rewriting. In those cases where an hypothesis appears in the LHS, we pretend it's not an argument to the lemma and make other arguments explicit in consequence. Here this gives
```suggestion
lemma head_eq_U (p : DyckWord) (h) : p.toList.head h = U := by
```
Note that the `(h)` is an extra sign that Lean can infer `h` from the LHS

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -77,43 +78,49 @@ instance : AddMonoid DyckWord where
...
   rcases p with - | s; · tauto
   rw [head_cons]
   by_contra f
   rename_i _ nonneg
   simpa [s.dichotomy.resolve_left f] using nonneg 1
 
 /-- The last element of a nonempty Dyck word is `D`. -/
-lemma getLast_eq_D (h : p.toList ≠ []) : p.toList.getLast h = D := by
+lemma getLast_eq_D : p.toList.getLast (toList_ne_nil.mpr h) = D := by
```
YaelDillies, line 97: Same here

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -142,17 +150,21 @@ def nest : DyckWord where
     rw [add_comm]
     exact add_le_add (zero_le _) ((count_le_length _ _).trans (by simp))
 
+lemma nest_add_ne_empty : (p.nest + q) ≠ 0 := by rw [← toList_ne_nil]; change U :: _ ++ _ ≠ []; simp
```
YaelDillies, line 153: Can you have a simp lemma of the form `p + q = 0 ↔ p = 0 ∧ q = 0` and a non-simp lemma of the form `p + q ≠ 0 ↔ p ≠ 0 ∨ q ≠ 0`? Then you wouldn't need `nest_add_ne_empty`
```suggestion
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +209,143 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+  exact ⟨by omega, by rw [Nat.sub_add_cancel lp, take_of_length_le (le_refl _),
+    p.count_U_eq_count_D]⟩
+
+include h in
+lemma count_take_firstReturn_add_one :
+    (p.toList.take (p.firstReturn + 1)).count U = (p.toList.take (p.firstReturn + 1)).count D := by
+  have := findIdx_get (w := (length_range p.toList.length).symm ▸ firstReturn_lt_length h)
+  simpa using this
+
+lemma count_lt_count_of_lt_firstReturn {i : ℕ} (hi : i < p.firstReturn) :
```
YaelDillies, line 248:
```suggestion
lemma count_D_lt_count_U_of_lt_firstReturn {i : ℕ} (hi : i < p.firstReturn) :
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +209,143 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+      simp_rw [cons_append, take_cons, count_cons, beq_self_eq_true, ite_true,
+        beq_iff_eq, ite_false, take_append_eq_append_take,
+        show j - p.toList.length = 0 by omega, take_zero, append_nil]
+      have := p.count_D_le_count_U j
+      omega
+  · simp_rw [length_range, u, length_append, length_cons]
+    exact Nat.lt_add_one _
+
+lemma nest_add_firstReturn : (p.nest + q).firstReturn = p.toList.length + 1 := by
+  rw [firstReturn_add, firstReturn_nest, ite_eq_right_iff, ← toList_eq_nil]; intro; contradiction
```
YaelDillies, line 288:
```suggestion
lemma firstReturn_nest_add : (p.nest + q).firstReturn = p.toList.length + 1 := by simp
```
also aybe we don't need that lemma at all as it's basically `firstReturn_add` + `firstReturn_nest`? Can you make sure simp proves it?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +209,143 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+      simp only [take, length_take, lt_min_iff] at ub ⊢
+      replace ub := ub.1
+      rw [take_take, min_eq_left ub.le]
+      rw [show i = i - 1 + 1 by omega] at ub ⊢
+      rw [Nat.add_lt_add_iff_right] at ub
+      exact count_lt_count_of_lt_firstReturn ub)
+
+variable (p) in
+/-- The right part of the Dyck word decomposition,
+outside the `U,D` pair that `firstReturn` refers to. `outsidePart 0 = 0`. -/
```
YaelDillies, line 307:
```suggestion
outside the `U, D` pair that `firstReturn` refers to. `outsidePart 0 = 0`. -/
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +209,143 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+def outsidePart : DyckWord :=
+  if h : p = 0 then 0 else p.drop (p.firstReturn + 1) (count_take_firstReturn_add_one h)
+
+@[simp] lemma insidePart_zero : insidePart 0 = 0 := by simp [insidePart]
+@[simp] lemma outsidePart_zero : outsidePart 0 = 0 := by simp [outsidePart]
+
+include h in
+@[simp]
+theorem insidePart_nest_add_outsidePart : p.insidePart.nest + p.outsidePart = p := by
```
YaelDillies, line 316:
```suggestion
theorem nest_insidePart_add_outsidePart : p.insidePart.nest + p.outsidePart = p := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +209,143 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+include h in
+theorem semilength_insidePart_lt : p.insidePart.semilength < p.semilength := by
+  rw [← Nat.mul_lt_mul_left zero_lt_two]
+  simp_rw [two_mul_semilength_eq_length]
+  nth_rw 2 [← insidePart_nest_add_outsidePart h]
+  change _ < (U :: p.insidePart.toList ++ [D] ++ p.outsidePart.toList).length
+  simp_rw [length_append, length_singleton, length_cons]; omega
+
+include h in
+theorem semilength_outsidePart_lt : p.outsidePart.semilength < p.semilength := by
```
YaelDillies, line 329: Can you deduce both lemmas from `p.insidePart.semilength + p.outsidePart.semilength + 1 = p.semilength`, which you itself can deduce from `p.insidePart.length + p.outsidePart.length + 2 = p.length`?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
YaelDillies: Okay, maybe you want this lemma because `p.nest + q` is a common idiom, but please make sure simp works

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: All done.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -77,43 +78,56 @@ instance : AddMonoid DyckWord where
...
 
-variable (p q : DyckWord)
+variable {p q : DyckWord}
 
 lemma toList_eq_nil : p.toList = [] ↔ p = 0 := by rw [DyckWord.ext_iff]; rfl
+lemma toList_ne_nil : p.toList ≠ [] ↔ p ≠ 0 := toList_eq_nil.ne
+
+@[simp]
+lemma add_eq_zero_iff : p + q = 0 ↔ p = 0 ∧ q = 0 := by
+  simp_rw [← toList_eq_nil]; apply append_eq_nil
```
YaelDillies, line 88: Actually can you prove `Subsingleton (AddUnits DyckWord)`? Then this lemma is just `add_eq_zero`

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -77,43 +78,56 @@ instance : AddMonoid DyckWord where
...
+
+lemma add_ne_zero_iff : p + q ≠ 0 ↔ p ≠ 0 ∨ q ≠ 0 := by
+  rw [not_iff_comm, add_eq_zero_iff]; tauto
```
YaelDillies, line 91: Can you generalise this lemma to be just after `add_eq_zero`?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -77,43 +78,56 @@ instance : AddMonoid DyckWord where
...
+  have : p.toList.dropLast.take 1 = [p.toList.head h'] := by
     rcases p with - | ⟨s, ⟨- | ⟨t, r⟩⟩⟩
     · tauto
     · rename_i bal _
       cases s <;> simp at bal
     · tauto
-  nth_rw 2 [← p.toList.dropLast_append_getLast h, ← p.toList.dropLast.take_append_drop 1]
-  rw [p.getLast_eq_D, drop_one, this, p.head_eq_U]
+  nth_rw 2 [← p.toList.dropLast_append_getLast h', ← p.toList.dropLast.take_append_drop 1]
+  rw [p.getLast_eq_D h, drop_one, this, p.head_eq_U h]
```
YaelDillies, line 120:
```suggestion
  rw [p.getLast_eq_D, drop_one, this, p.head_eq_U]
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: `add_eq_zero` requires `AddCommMonoid`, and Dyck word addition is not commutative.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: See my reply to your previous comment.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
YaelDillies: I'm sure `add_eq_zero` can be generalise to `AddMonoid`. Can you do that?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: That doesn't work; the `p ≠ 0` hypotheses are then left as side goals.

YaelDillies: That's because you haven't removed the `toList_ne_nil.mpr` in the statement of `getLast_eq_D`

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: That, unfortunately, **cannot be done** – at least left-cancellativity is required (which Dyck words do satisfy). [This MSE answer](https://math.stackexchange.com/a/4535529/357390) gives a counterexample to `add_ne_zero` on arbitrary monoids as the free one on two generators with `ab=1`.

YaelDillies: Mind writing the cancellative version then?

Parcly-Taxel: All done (including your comment below on `head_eq_U`/`getLast_eq_D`).

Mathlib/Algebra/Group/Units.lean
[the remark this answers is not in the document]
Parcly-Taxel: #15921

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +221,142 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+variable (p) in
+/-- `p.firstReturn` is 0 if `p = 0` and the index of the `D` matching the initial `U` otherwise. -/
+def firstReturn : ℕ :=
+  (range p.toList.length).findIdx fun i ↦
+    (p.toList.take (i + 1)).count U = (p.toList.take (i + 1)).count D
+
+@[simp] lemma firstReturn_zero : firstReturn 0 = 0 := rfl
+
+include h in
+lemma firstReturn_pos : 0 < p.firstReturn := by
```
YaelDillies, line 235: Either write the corresponding `positivity` extension or
```suggestion
-- TODO: `positivity` extension
include h in
lemma firstReturn_pos : 0 < p.firstReturn := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +221,142 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+theorem semilength_insidePart_lt : p.insidePart.semilength < p.semilength := by
+  have := semilength_insidePart_add_semilength_outsidePart_add_one h
+  omega
+
+include h in
+theorem semilength_outsidePart_lt : p.outsidePart.semilength < p.semilength := by
+  have := semilength_insidePart_add_semilength_outsidePart_add_one h
+  omega
+
+theorem nest_add_insidePart_eq : (p.nest + q).insidePart = p := by
```
YaelDillies, line 347:
```suggestion
theorem insidePart_nest_add : (p.nest + q).insidePart = p := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -196,4 +221,142 @@ lemma two_mul_semilength_eq_length : 2 * p.semilength = p.toList.length := by
...
+  have : p.toList.length + 1 + 1 = p.nest.toList.length := by simp [nest]
+  simp_rw [insidePart, firstReturn_nest_add, take,
+    show (p.nest + q).toList = p.nest.toList ++ q.toList by rfl, take_append_eq_append_take,
+    this, take_length, tsub_self, take_zero, append_nil, denest, nest, dropLast_concat]
+  rfl
+
+theorem nest_add_outsidePart_eq : (p.nest + q).outsidePart = q := by
```
YaelDillies, line 354:
```suggestion
theorem outsidePart_nest_add : (p.nest + q).outsidePart = q := by
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
Parcly-Taxel: All done.

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -360,3 +368,24 @@ theorem nest_add_outsidePart_eq : (p.nest + q).outsidePart = q := by
...
 end DyckWord
+
+namespace Mathlib.Meta.Positivity
+
+open Lean Meta Qq
+
+/-- Extension for the `positivity` tactic:
+`p.firstReturn` is nonnegative, and is positive if `p` is nonzero. -/
+@[positivity DyckWord.firstReturn _]
+def evalFirstReturn : PositivityExt where eval {u α} _zα _pα e := do
```
YaelDillies, line 432:
```suggestion
def evalDyckWordFirstReturn : PositivityExt where eval {u α} _zα _pα e := do
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -79,6 +79,14 @@ instance : AddCancelMonoid DyckWord where
   add_left_cancel p q r h := by rw [DyckWord.ext_iff] at *; exact append_cancel_left h
   add_right_cancel p q r h := by rw [DyckWord.ext_iff] at *; exact append_cancel_right h
 
+/-- We define a partial order on Dyck words by writing `p ≤ q` if `p` is a suffix of `q`. -/
+instance : PartialOrder DyckWord where
+  le p q := p.toList <:+ q.toList
+  le_refl p := suffix_rfl
+  le_trans p q r := IsSuffix.trans
+  le_antisymm p q pq qp :=
+    DyckWord.ext_iff.mpr <| eq_of_suffix_of_length_eq pq <| pq.length_le.antisymm qp.length_le
```
YaelDillies: Can you provide lemmas like ` p.toList <:+ q.toList ↔ p < q`, `0 < p ↔ p ≠ 0`, ...? I suspect it might have to be a separate PR

Parcly-Taxel: I've probably added – and changed – enough in this PR; we can merge this in its current state and leave the extra ordering lemmas to the next and last PR in my plans, #9781 itself where I will add the tree bijection/relation to Catalan numbers.

YaelDillies: Sure

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

bryangingechen:
bors r+

Just in case...
bors d+

PR #15866, opened 2024-08-16
[Merged by Bors] - feat: A subgroup of a `ZMod`-module is a `ZMod`-module

YaelDillies:
From LeanAPAP

---

### DIFF
--- a/Mathlib/Data/ZMod/Basic.lean
+++ b/Mathlib/Data/ZMod/Basic.lean
@@ -1372,6 +1372,25 @@ end lift
 
 end ZMod
 
+section Module
+variable {S G : Type*} [AddCommGroup G] {n : ℕ} [Module (ZMod n) G] [SetLike S G]
+  [AddSubgroupClass S G] {K : S} {x : G}
+
+lemma zmod_smul_mem (hx : x ∈ K) : ∀ a : ZMod n, a • x ∈ K := by
+  simpa [ZMod.forall, Int.cast_smul_eq_zsmul] using zsmul_mem hx
+
+namespace AddSubgroupClass
+
+instance instZModSMul : SMul (ZMod n) K where smul a x := ⟨a • x, zmod_smul_mem x.2 _⟩
+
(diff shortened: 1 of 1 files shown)


Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1372,6 +1372,25 @@ end lift
 
 end ZMod
 
+section Module
+variable {S G : Type*} [AddCommGroup G] {n : ℕ} [Module (ZMod n) G] [SetLike S G]
+  [AddSubgroupClass S G] {K : S} {x : G}
+
+lemma zmod_smul_mem (hx : x ∈ K) : ∀ a : ZMod n, a • x ∈ K := by
```
eric-wieser, line 1379:
```suggestion
lemma zmod_smul_mem (a : ZMod n) (hx : x ∈ K) : a • x ∈ K := by
```
seems like a better argument order.

Alternatively, can you create a `SMulMemClass` instance?

YaelDillies: I wrote `zmod_smul_mem` to match `nsmul_mem`/`zsmul_mem`. These disagree with your argument order.

The `SMulMemClass` statement cannot be made an instance because of `R := ZMod n` being an out-param and the `Module (ZMod n) G` argument. I added it as a lemma instead

Mathlib/Data/ZMod/Basic.lean
```diff
@@ -1372,6 +1373,30 @@ end lift
...
+  simpa [ZMod.forall, Int.cast_smul_eq_zsmul] using zsmul_mem hx
+
+/-- This cannot be made an instance because of the `[Module (ZMod n) G]` argument and the fact that
+`n` only appears on the right of the colon in `R := ZMod n`, which is an `OutParam` in the
+definition of `SMulMemClass`. -/
```
eric-wieser, line 1385:
```suggestion
`n` only appears in the second argument of `SMulMemClass`, which is an `OutParam`. -/
```

eric-wieser (approved):
bors d+

Thanks!

YaelDillies:
bors merge

PR #15868, opened 2024-08-16
[Merged by Bors] - chore: add missing end

grunweg:
Fix mathlib for running the missingEnd linter everywhere. Prerequiste to #15845.

---

### DIFF
--- a/Mathlib/Algebra/Order/Monoid/Unbundled/OrderDual.lean
+++ b/Mathlib/Algebra/Order/Monoid/Unbundled/OrderDual.lean
@@ -62,3 +62,5 @@ instance covariantClass_swap_mul_lt [LT α] [Mul α]
   ⟨c.1.flip⟩
 
 
+
+end OrderDual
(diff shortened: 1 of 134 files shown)


Mathlib/Algebra/Order/Monoid/Unbundled/OrderDual.lean
```diff
@@ -62,3 +62,5 @@ instance covariantClass_swap_mul_lt [LT α] [Mul α]
   ⟨c.1.flip⟩
 
 
+
```
adomani, line 65:
```suggestion

```
Maybe this is too many empty lines?

adomani (approved):
Looks good to me, thanks!

There is some inconsistency with `end Z end Y` and `end Y.Z`.  However, this is minor and, if there is a clear style preference, it would likely have to be enforced by a linter.

maintainer merge

grunweg:
Good find, thanks! I tweaked some changes of this PR, trying to more closely follow "end namespaces how they were opened".
Previous mathlib style is uneven, so changing this in a follow-up (if we care) seems appropriate.

jcommelin:
Thanks :tada:

bors merge

PR #15869, opened 2024-08-16
chore: fix remaining linters in mathlib and tests

grunweg:
Remove two unused tactics; fix a few long lines, and make sure all tests disable the linters they want to ignore.

In at least one instance, this has to wait for #15845, as the respective linter is not available yet.

---


grunweg:
This PR fails as-is: all disabling of linters needs those linters to be defined already... let's roll this into #15845 instead.
Let me close this in favour of #15872, which has the backportable part of these changes.

PR #15874, opened 2024-08-16
[Merged by Bors] - doc(Order/InitialSeg): Improve documentation

vihdzp:
The file previously conflated initial segments and their embeddings. Our docstrings are now more careful about the distinction.

---


Mathlib/Order/InitialSeg.lean
```diff
@@ -9,16 +9,20 @@ import Mathlib.Order.WellFounded
 /-!
 # Initial and principal segments
 
-This file defines initial and principal segments.
+This file defines initial and principal segment embeddings. Though these definitions make sense for
+arbitrary relations, they're intended for use with well orders.
+
+An initial segment within a well order is any downwards closed subset. A principal segment is an
+initial segment other than `Set.univ`.
```
YaelDillies, line 16: Your sentence is technically true but misleading. Instead of explaining what the definitions are equivalent to in the well-order case, it should give the general definitions, namely
```suggestion
An initial segment is simply a lower set. A principal segment is a set of the form `Set.Iic x` for some `x`.
```

YaelDillies (commented):
Where did you get that "initial segment" vs "initial segment embedding" terminology from? This is certainly not how we call things in combinatorics. cc @b-mehta

b-mehta:
> Where did you get that "initial segment" vs "initial segment embedding" terminology from? This is certainly not how we call things in combinatorics. cc @b-mehta

What do you mean? The corrected changes in this PR look reasonable to me.

YaelDillies:
There is no such thing as an "initial segment embedding" in my combinatorics lecture notes

b-mehta:
Indeed, initial segments are a set theory / ordinals notion that was imported to combinatorics, with the same meaning. This file talks about embeddings which are initial segments, not about initial segments. As such, it shouldn't have the name initialSeg - as Violeta correctly points out in the PR description - and this makes it _more_ consistent with your notes.

vihdzp:
In pen and paper set theory, the way you generally work with initial segments is that you talk about a well-order being isomorphic to an initial segment of another, by which you mean a downwards-closed set under the same relation. 

Mathlib's notion of an "initial segment embedding" might be ad-hoc, but it's nicer to work with since it avoids the intermediate `Subrel`.

Mathlib/Init/Algebra/Classes.lean
```diff
@@ -181,6 +181,18 @@ variable {α : Sort u} {r : α → α → Prop}
...
+  rcases @trichotomous α r _ b c with (lt | rfl | gt)
+  · exact _root_.trans hab lt
+  · exact hab
+  · exact (hcb gt).elim
+
+theorem trans_le_lt [IsStrictTotalOrder α r] {a b c : α} (hab : ¬ b ≺ a) (hcb : b ≺ c) : a ≺ c := by
+  rcases @trichotomous α r _ a b with (lt | rfl | gt)
+  · exact _root_.trans lt hcb
+  · exact hcb
+  · exact (hab gt).elim
```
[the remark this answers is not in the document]
vihdzp: Why not? Admittedly this typeclass is a bit strange and specific, but it's satisfied by `IsWellOrder` which does see use in Mathlib.

Mathlib/Order/InitialSeg.lean
vihdzp: `Set.Iio`, not `Set.Iic`. I've updated the descriptions of `InitialSeg` and `PrincipalSeg` with your suggestion in mind.

Mathlib/Init/Algebra/Classes.lean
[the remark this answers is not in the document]
YaelDillies: `Init.Algebra.Classes` is an ancient file. Almost every declaration in here is deprecated and the file itself will soon be deleted. `IsWellOrder` too is ancient and we don't want to be adding any new lemma for it.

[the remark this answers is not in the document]
vihdzp: This file uses the `IsWellOrder` typeclass extensively, and ordinals themselves are defined using `IsWellOrder`. I'm not opposed to getting rid of it, but when that inevitably happens, there's going to be a lot of fallout around this area. So maybe these lemmas can be cleaned up as part of that?

I can remove the unused `trans_lt_le` and move `trans_le_lt` to this file, with a docstring note explaining that at some point we need to refactor it out.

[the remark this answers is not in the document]
YaelDillies: Well what you replace `IsStrictTotalOrder` by `IsWellOrder` in your new lemmas?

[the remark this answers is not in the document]
vihdzp: Alright, I moved them to the `RelClasses` file and stated them for `IsWellOrder`.

Mathlib/Order/InitialSeg.lean
```diff
@@ -10,16 +10,20 @@ import Mathlib.Data.Sum.Order
 /-!
 # Initial and principal segments
 
-This file defines initial and principal segments.
+This file defines initial and principal segment embeddings. Though these definitions make sense for
+arbitrary relations, they're intended for use with well orders.
+
+An initial segment is simply a lower set. A principal segment is a set of the form `Set.Iio x` for
```
b-mehta, line 16: While "lower set" is a common enough term in order theory, I think we should still have something like the old description giving a quick explanation of the term. This saves the reader looking it up, and more importantly reassures that the term we use is the same as the informal one! (as you've just seen, that's not always the case in mathlib ;) )

Mathlib/Order/InitialSeg.lean
```diff
@@ -29,31 +33,21 @@ These notations belong to the `InitialSeg` locale.
...
 open Function
 
 /-- If `r` is a relation on `α` and `s` in a relation on `β`, then `f : r ≼i s` is an order
-embedding whose range is an initial segment. That is, whenever `b < f a` in `β` then `b` is in the
+embedding whose `Set.range` is a lower set. That is, whenever `b < f a` in `β` then `b` is in the
 range of `f`. -/
 structure InitialSeg {α β : Type*} (r : α → α → Prop) (s : β → β → Prop) extends r ↪r s where
-  /-- The order embedding is an initial segment -/
-  init' : ∀ a b, s b (toRelEmbedding a) → ∃ a', toRelEmbedding a' = b
+  /-- The range of the embedding is a lower set. -/
```
b-mehta, line 47: We write `lower set` here, but this doesn't use the predicate [IsLowerSet](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/UpperLower/Basic.html#IsLowerSet). We should either
a) use the existing definition
b) prove the existing definition is equivalent to this and reference that proof in this docstring
c) give an alternative informal description

The choice is yours :)

Mathlib/Order/InitialSeg.lean
```diff
@@ -194,29 +196,18 @@ protected theorem acc (f : r ≼i s) (a : α) : Acc r a ↔ Acc s (f a) :=
...
-segments.
--/
-
+/-! ### Principal segments -/
 
-/-- If `r` is a relation on `α` and `s` in a relation on `β`, then `f : r ≺i s` is an order
-embedding whose range is an open interval `(-∞, top)` for some element `top` of `β`. Such order
-embeddings are called principal segments -/
+/-- If `r` is a relation on `α` and `s` in a relation on `β`, then `f : r ≺i s` is an initial
+segment embedding whose range is `Set.Iio top` for some element `top`. If `β` is a well-order, this
```
b-mehta, line 202: I don't like the variable name `top` here, especially because this probably won't be the top element of the order

Mathlib/Order/InitialSeg.lean
vihdzp: We can't use `IsLowerSet` since this is for an arbitrary relation, and `IsLowerSet` requires `≤`. But morally, that's what it should be.

Mathlib/Order/InitialSeg.lean
vihdzp: I'm compelled to agree, but I'm not sure what a good replacement might be.

Mathlib/Order/InitialSeg.lean
b-mehta: Good point - that rules out a) but giving the special case for orders would work for b), and c) is my preferred one anyway

Mathlib/Order/InitialSeg.lean
b-mehta: I think just `x` should be okay here - everything else is already a single-letter name in this context

Mathlib/Order/InitialSeg.lean
vihdzp: Working with initial and principal segments with `<` is somewhat annoying, and it often requires writing the types explicitly like `((· < ·) : α → α → Prop) ≼i ((· < ·) : β → β → Prop)`. I think that issue should probably be fixed before we introduce any API for this special case.

Do I just copy the description from the structure docstring? The "whenever `b < f a` in `β` then..." bit.

vihdzp: Actually, I will introduce the `IsLowerSet` version of this in a separate PR. That way we can use "proper" vocabulary and have it be right!

vihdzp:
Drafting this until #17481 goes through, at which point I can spell the fact that the range of `InitialSeg` is in fact an initial segment.

vihdzp:
In fact, I think this PR might have done too much... I'll open separate PRs for most of the other changes and keep only the stuff on documentation here.

Mathlib/Order/InitialSeg.lean
vihdzp: Two months later, and we now have [`InitialSeg.isLowerSet_range`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/InitialSeg.html#InitialSeg.isLowerSet_range).

Mathlib/Order/InitialSeg.lean
```diff
@@ -35,22 +39,15 @@ These notations belong to the `InitialSeg` locale.
...
+/-! ### Initial segment embeddings -/
 
 variable {α : Type*} {β : Type*} {γ : Type*} {r : α → α → Prop} {s : β → β → Prop}
   {t : γ → γ → Prop}
 
 open Function
 
 /-- If `r` is a relation on `α` and `s` in a relation on `β`, then `f : r ≼i s` is an order
-embedding whose range is an initial segment. That is, whenever `b < f a` in `β` then `b` is in the
+embedding whose `Set.range` is a lower set. That is, whenever `b < f a` in `β` then `b` is in the
```
b-mehta, line 50: Didn't we discuss changing the definition to make this literally true? Or am I misremembering the conclusion

Mathlib/SetTheory/Ordinal/Basic.lean
```diff
@@ -449,7 +449,7 @@ the elements of `α`. -/
 -- The explicit typing is required in order for `simp` to work properly.
 @[simps! symm_apply_coe]
 def enum (r : α → α → Prop) [IsWellOrder α r] :
-    @RelIso { o // o < type r } α (Subrel (· < ·) { o | o < type r }) r :=
+    @RelIso { o // o < type r } α (Subrel (· < ·) (Iio (type r))) r :=
```
b-mehta, line 452: Is this change needed in this PR? If not, I'd suggest leaving it out of this one and doing it elsewhere, for a cleaner diff.

vihdzp: This isn't needed here, I'll put it in another PR.

Mathlib/Order/InitialSeg.lean
vihdzp: This can't be done as `IsLowerSet` is defined in terms of an order relation, while `InitialSeg` is stated in terms of an arbitrary relation.

However, we do already have `InitialSeg.isLowerSet_range` for initial segments between `<` in a preorder and `<` in a partial order.

Mathlib/SetTheory/Ordinal/Basic.lean
vihdzp: This is now #20413.

b-mehta (approved):
Thanks!

bors merge

PR #15875, opened 2024-08-16
[Merged by Bors] - chore(SpecialFunctions): re-instate many simp lemmas

grunweg:
Remove some now-addressed porting notes; re-instate a few more lemmas which were formerly marked as simp.

---


j-loreaux:
@mattrobball what does your keying tool say about these `simp` lemmas? If you point me to it, I can check myself.

mattrobball:
`#discr_tree_simp_key my_lemma`

Mathlib/MeasureTheory/Integral/IntervalIntegral.lean
```diff
@@ -620,12 +620,7 @@ section Comp
 
 variable {a b c d : ℝ} (f : ℝ → E)
 
-/-!
-Porting note: some `@[simp]` attributes in this section were removed to make the `simpNF` linter
-happy. TODO: find out if these lemmas are actually good or bad `simp` lemmas.
--/
-
-@[simp] -- Porting note (#10618): was @[simp]
+@[simp]
```
mattrobball, line 623: `@intervalIntegral _ _ _ <other> _ _ (@volume ℝ _)`

Mathlib/MeasureTheory/Integral/IntervalIntegral.lean
```diff
@@ -638,32 +633,32 @@ theorem integral_comp_mul_right (hc : c ≠ 0) :
       Measure.restrict_congr_set (α := ℝ) (μ := volume) Ico_ae_eq_Ioc]
   · simp [h, mul_div_cancel_right₀, hc, abs_of_pos]
 
-@[simp] -- Porting note (#10618): was @[simp]
+@[simp]
```
mattrobball, line 636: `@HSMul.hSMul ℝ _ _ _ _ (@intervalIntegral _ _ _ <other> _ _ (@volume ℝ _))`

Mathlib/MeasureTheory/Integral/IntervalIntegral.lean
```diff
@@ -638,32 +633,32 @@ theorem integral_comp_mul_right (hc : c ≠ 0) :
...
 theorem smul_integral_comp_mul_right (c) :
     (c • ∫ x in a..b, f (x * c)) = ∫ x in a * c..b * c, f x := by
   by_cases hc : c = 0 <;> simp [hc, integral_comp_mul_right]
 
-@[simp] -- Porting note (#10618): was @[simp]
+@[simp]
```
mattrobball, line 641: `@intervalIntegral _ _ _ <other> _ _ (@volume ℝ _)`

Mathlib/MeasureTheory/Integral/IntervalIntegral.lean
```diff
@@ -638,32 +633,32 @@ theorem integral_comp_mul_right (hc : c ≠ 0) :
...
 theorem integral_comp_mul_left (hc : c ≠ 0) :
     (∫ x in a..b, f (c * x)) = c⁻¹ • ∫ x in c * a..c * b, f x := by
   simpa only [mul_comm c] using integral_comp_mul_right f hc
 
-@[simp] -- Porting note (#10618): was @[simp]
+@[simp]
```
mattrobball, line 646: `@HSMul.hSMul ℝ _ _ _ _ (@intervalIntegral _ _ _ <other> _ _ (@volume ℝ _))`

mattrobball (commented):
I went through a couple and the keys aren't great for the context.

mattrobball:
!bench

j-loreaux:
If we wanted to, we could still scope them to `intervalIntegral`, although in that case, it would be nice if there were a comment in the docstring for each lemma.

mattrobball:
Since `intervalIntegral` is the head symbol, the discrimination tree already effectively does the scoping. The main issue is that it doesn't see anything beyond that head symbol for the most part. I think these are ok for now because the domain is pretty small. It will probably be fine indefinitely.

grunweg:
I do not have a strong opinion about the newly marked simp lemmas: I am happy to also just remove the porting note.

mattrobball:
I think just merging is fine.

j-loreaux:
bors merge

j-loreaux:
bors r-

bors d+

bryangingechen:
One of the runners ran out of space (and I've temporarily removed it from the pool), so re-running this should work. I'm not sure if there's another reason this was given `bors r-` though.

PR #15876, opened 2024-08-16
[Merged by Bors] - chore: Fix precedence of `~r` infix for `IsRotated`

YaelDillies:
---

### DIFF
--- a/Mathlib/Data/List/Rotate.lean
+++ b/Mathlib/Data/List/Rotate.lean
@@ -374,7 +374,9 @@ def IsRotated : Prop :=
   ∃ n, l.rotate n = l'
 
 @[inherit_doc List.IsRotated]
-infixr:1000 " ~r " => IsRotated
+infixr:50 " ~r " => IsRotated
+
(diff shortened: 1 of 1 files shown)


Mathlib/Data/List/Rotate.lean
```diff
@@ -374,7 +374,7 @@ def IsRotated : Prop :=
   ∃ n, l.rotate n = l'
 
 @[inherit_doc List.IsRotated]
-infixr:1000 " ~r " => IsRotated
+infixr:50 " ~r " => IsRotated
```
eric-wieser, line 377: Does this match the precedence of `Perm`? If so, can you add a `--` comment to that effect?

bors d+

YaelDillies:
```suggestion
-- This matches the precedence of the infix `~` for `List.Perm`, and more generally of relation infixes
infixr:50 " ~r " => IsRotated
```

YaelDillies:
bors merge

PR #10541, opened 2024-02-14
feat(Algebra/SkewMonoidAlgebra/Basic): add SkewMonoidAlgebra

xgenereux:
# Skew Monoid algebras
This file presents a skewed version of `Mathlib.Algebra.MonoidAlgebra.Basic`.

## Definition
We define `SkewMonoidAlgebra k G := G →₀ k` attached with a skewed convolution product.
Here, the product of two elements `f g : SkewMonoidAlgebra k G` is the finitely supported
function whose value at `a` is the sum of `f x * (x • g y)` over all pairs `x, y`
such that `x * y = a`.

This will be used in a later PR to define skew polynomial rings.

Co-authored-by: María Inés de Frutos Fernández <[mariaines.dff@gmail.com](mailto:mariaines.dff@gmail.com)>

---

- [x] depends on: #15878
- [x] depends on: #19084
- [x] depends on: #22078

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15878, opened 2024-08-16
[Merged by Bors] - feat(SkewMonoidAlgebra): add SkewMonoidAlgebra with irreducible def

xgenereux:
## Skew Monoid algebras
This file presents a skewed version of `Mathlib.Algebra.MonoidAlgebra.Basic` with an **irreducible** definition.

## Context
Hi everyone,
As a part of a bigger project aiming to get [Drinfeld Modules](https://en.wikipedia.org/wiki/Drinfeld_module) into Mathlib, @**María Inés de Frutos Fernández**  and I have been trying to get a generalization of `MonoidAlgebra`, `SkewMonoidAlgebra`, into Mathlib. The first attempt at this was #10541, but progress has been slow as making the definition irreducible caused a huge refactor.

Following @kbuzzard's [suggestion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60SkewMonoidAlgebra.60.20and.20the.20separation.20with.20.60Finsupp.60/near/456861000), I am opening this PR which contains only the first 160 lines of #10541. In particular, this PR is missing most of what makes these objects skewed (Multiplication was defined only at line 558 as it requires stronger hypotheses and there are many lemmas on the additive structure that precede it.) This means that the purpose of this PR is mainly to get the separation from `Finsupp` right.

Please consult this [thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60SkewMonoidAlgebra.60.20and.20the.20separation.20with.20.60Finsupp.60/near/456834154) for more context or in short:

The current separation we have consists of:

1. Wrapping the definition with a structure.
2. Making basic operations such as `add` private and irreducible.
3. Adapting relevant definitions.
4. Copying and adapting relevant theorems from the base construct to obtain statements in terms of our new object.

Is this what everyone has in mind? If so, are there parts of the current PR that come short in respect to any of these?

I am happy to integrate suggestions to improve the separation if there are any!

Thanks a lot for your attention and help!
Xavier

---


Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,159 @@
...
+
+private irreducible_def add :
+    SkewMonoidAlgebra k G → SkewMonoidAlgebra k G → SkewMonoidAlgebra k G
+  | ⟨a⟩, ⟨b⟩ => ⟨a + b⟩
+
+private irreducible_def smul {S : Type*} [SMulZeroClass S k] :
+    S → SkewMonoidAlgebra k G → SkewMonoidAlgebra k G
+  | s, ⟨b⟩ => ⟨s • b⟩
+
+instance zero : Zero (SkewMonoidAlgebra k G) := ⟨⟨0⟩⟩
```
Ruben-VandeVelde, line 54: Nit: (new) named instances should be `instTypeclass`, e.g. `instZero`

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,159 @@
...
+
+instance nontrivial [Nontrivial k] [Nonempty G] :
+    Nontrivial (SkewMonoidAlgebra k G) := Function.Injective.nontrivial ofFinsupp_injective
+
+instance addCommMonoid : AddCommMonoid (SkewMonoidAlgebra k G) where
+    __ := toFinsupp_injective.addCommMonoid _ toFinsupp_zero toFinsupp_add
+      (fun _ _ => toFinsupp_smul _ _)
+    toAdd  := SkewMonoidAlgebra.add'
+    toZero := SkewMonoidAlgebra.zero
+    nsmul  := (· • ·)
```
Ruben-VandeVelde, line 131: These lines should only be indented two spaces

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,159 @@
...
+Copyright (c) 2024 María Inés de Frutos Fernández. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: María Inés de Frutos Fernández, Xavier Généreux
+-/
+import Mathlib.Data.Finsupp.Basic
+
+/-!
+This file presents a skewed version of `Mathlib.Algebra.MonoidAlgebra.Basic`.
+
+We define `SkewMonoidAlgebra k G` attached with a skewed convolution product.
```
Ruben-VandeVelde, line 11: Please note in the docstring which parts exist at the current state of the file

mattrobball:
This is an `irreducible_def` not an `@[irreducible] def`. If there is not kernel performance issues evident, then I much prefer the latter to push maintenance issues to core.

Louddy:
@Ruben-VandeVelde, this has been updated thank you.

@mattrobball, I am a little bit unfamiliar with the [@irreducible] attribute. Where is it defined? In particular, I can't prove `instSMulZeroClass.smul_zero` when I use it.

mattrobball:
It is in core Lean. Current practice is to use `unseal foo in` before the declaration if you need to access the definition of `foo` or to use `with_unfolding_all` in tactic mode for steps that require unfolding

madvorak:
LGTM.

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+theorem toFinsupp_smul {S : Type*} [SMulZeroClass S k] (a : S) (b : SkewMonoidAlgebra k G) :
+    (a • b).toFinsupp = a • b.toFinsupp := by
+  cases b
+  rw [← ofFinsupp_smul]
+
+theorem _root_.IsSMulRegular.skewMonoidAlgebra {S : Type*} [Monoid S] [DistribMulAction S k] {a : S}
+    (ha : IsSMulRegular k a) : IsSMulRegular (SkewMonoidAlgebra k G) a
+  | ⟨_⟩, ⟨_⟩, h => by
+    simp only [← ofFinsupp_smul] at h
+    exact congr_arg _ <| ha.finsupp (ofFinsupp.inj (h))
```
kbuzzard, line 103:
```suggestion
    exact congr_arg _ <| ha.finsupp (ofFinsupp.inj h)
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+  | ⟨a⟩, ⟨b⟩ => ⟨a + b⟩
+
+@[irreducible]
+private def smul {S : Type*} [SMulZeroClass S k] :
+    S → SkewMonoidAlgebra k G → SkewMonoidAlgebra k G
+  | s, ⟨b⟩ => ⟨s • b⟩
+
+instance instZero : Zero (SkewMonoidAlgebra k G) := ⟨⟨0⟩⟩
+
+instance instAdd' : Add (SkewMonoidAlgebra k G) := ⟨add⟩
```
kbuzzard, line 63:
```suggestion
instance instAdd : Add (SkewMonoidAlgebra k G) := ⟨add⟩
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+/-- The skew monoid algebra over a semiring `k` generated by the monoid `G`.
+It is the type of finite formal `k`-linear combinations of terms of `G`,
+endowed with a skewed convolution product. -/
+structure SkewMonoidAlgebra (k : Type*) (G : Type*) [Zero k] where
+  /-- Map **from** `G →₀ k`. -/
+  ofFinsupp ::
+  /-- Map **to** `G →₀ k`. -/
+  toFinsupp : G →₀ k
+
+open Function hiding Commute
```
kbuzzard, line 37:
```suggestion
open Function
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+
+section AddCommMonoid
+
+variable [AddCommMonoid k]
+
+@[simp]
+theorem eta (f : SkewMonoidAlgebra k G) : ofFinsupp f.toFinsupp = f := rfl
+
+@[irreducible]
+private def add :
```
kbuzzard, line 52: Why is this (and `smul`) `private`? And why are they irreducible?

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+  ofFinsupp ::
+  /-- Map **to** `G →₀ k`. -/
+  toFinsupp : G →₀ k
+
+open Function hiding Commute
+namespace SkewMonoidAlgebra
+
+universe u₁ u₂
+
+variable {k : Type u₁} {G : Type u₂} {H : Type*} {R : Type*}
```
kbuzzard, line 42: Why are you naming some universes but not others?

kbuzzard, line 42:
```suggestion
variable {k G H R : Type*}
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,167 @@
...
+@[irreducible]
+private def smul {S : Type*} [SMulZeroClass S k] :
+    S → SkewMonoidAlgebra k G → SkewMonoidAlgebra k G
+  | s, ⟨b⟩ => ⟨s • b⟩
+
+instance instZero : Zero (SkewMonoidAlgebra k G) := ⟨⟨0⟩⟩
+
+instance instAdd' : Add (SkewMonoidAlgebra k G) := ⟨add⟩
+
+unseal smul in
```
kbuzzard, line 65: If you hadn't made it irreducible, you wouldn't have to do this!

kbuzzard (approved):
This looks good to me. I've left some more comments.

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
Louddy: I think the idea is that we eventually don't want to be able to reference these definitions unless we really need to. The point of the separation is to forget about the `Finsupp` structure, so we ultimately want to be in a position where we never need to unfold these. That being said, we need to use it a couple time to get there.

This is my understanding of what the separation we are doing means but I could be wrong as to my knowledge there are no clear guidelines.

Louddy:
Hey everyone, I have requested reviews from the reviewers of #10541.

In particular, @eric-wieser mentioned diamond problems with the irreducibility of `smul` [here](https://github.com/leanprover-community/mathlib4/pull/10541#discussion_r1705399772). Is working with `@[irreducible]` instead of irreducible_def fine? (I assume not?) Should I revert this and only have `add` as irreducible?

kbuzzard:
@eric-wieser What is your feeling about irreducibility of add and smul here?

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
mariainesdff: They are private to hide `Finsupp` (as is done in the current implementation of `Polynomial`); same for the irreducibility of `add`.

I have removed the `irreducible` tag from `smul` since, as @eric-wieser pointed out [here](https://github.com/leanprover-community/mathlib4/pull/10541#discussion_r1705688680), this causes diamonds. In particular, the following example fails if `smul` is irreducible.
```lean
example {S} [Monoid S] [AddCommMonoid k] [DistribMulAction S k] :
    (Units.instDistribMulAction : DistribMulAction Sˣ (SkewMonoidAlgebra k G)) =
    SkewMonoidAlgebra.distribMulAction := by
  with_reducible_and_instances rfl
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,163 @@
...
+need to be a monoid for most of our uses.
+-/
+
+noncomputable section
+
+/-- The skew monoid algebra over a semiring `k` generated by the monoid `G`.
+It is the type of finite formal `k`-linear combinations of terms of `G`,
+endowed with a skewed convolution product. -/
+structure SkewMonoidAlgebra (k : Type*) (G : Type*) [Zero k] where
+  /-- Map **from** `G →₀ k`. -/
```
kbuzzard, line 32:
```suggestion
  /-- The natural map from `G →₀ k` to `SkewMonoidAlgebra k G`. -/
```

Mathlib/Algebra/SkewMonoidAlgebra/Basic.lean
```diff
@@ -0,0 +1,163 @@
...
+  ofFinsupp ::
+  /-- Map **to** `G →₀ k`. -/
```
kbuzzard, line 34:
```suggestion
  /-- The natural map from `SkewMonoidAlgebra k G` to `G →₀ k`. -/
```

kbuzzard (approved):
I am minded to merge this PR (I just left two docstring comments) if we don't hear from Eric within the next few days. My understanding is that we have already discussed all the irreducibility issues.

eric-wieser (approved):
bors d+

Thanks!

PR #15879, opened 2024-08-16
[Merged by Bors] - chore: Let `gcongr` know about `Nat.cast_lt`

YaelDillies:
From LeanAPAP

---

### DIFF
--- a/Mathlib/Data/Nat/Cast/Order/Basic.lean
+++ b/Mathlib/Data/Nat/Cast/Order/Basic.lean
@@ -69,6 +69,9 @@ variable [CharZero α] {m n : ℕ}
 theorem strictMono_cast : StrictMono (Nat.cast : ℕ → α) :=
   mono_cast.strictMono_of_injective cast_injective
 
+@[gcongr]
+lemma _root_.GCongr.natCast_lt_natCast {a b : ℕ} (h : a < b) : (a : α) < b := strictMono_cast h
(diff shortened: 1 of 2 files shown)


Mathlib/Data/Nat/Cast/Order/Basic.lean
```diff
@@ -69,6 +69,9 @@ variable [CharZero α] {m n : ℕ}
 theorem strictMono_cast : StrictMono (Nat.cast : ℕ → α) :=
   mono_cast.strictMono_of_injective cast_injective
 
+@[gcongr]
+lemma _root_.GCongr.natCast_lt_natCast {a b : ℕ} (h : a < b) : (a : α) < b := strictMono_cast h
```
j-loreaux, line 73: Can we just make this an `alias`? And add a docstring to that effect?

j-loreaux (commented):
bors d+

Mathlib/Data/Nat/Cast/Order/Basic.lean
YaelDillies: An alias of what? `strictMono_cast`? `gcongr` doesn't understand `StrictMono`

YaelDillies:
I think your remark isn't crucial, Jireh, so I will march on towards the juicy stuff in APAP.

bors merge

PR #15881, opened 2024-08-16
[Merged by Bors] - refactor(Order/SuccPred): Remove `le_of_lt_succ` and `le_of_pred_lt` from `SuccOrder`/`PredOrder`

Command-Master:
See disccusion on Zulip: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/weaker.20SuccOrder.2FPredOrder

---


vihdzp:
I mentioned this on Zulip too: I think `StrongXOrder` aren't needed. In practice, when we talk about a `SuccOrder` in its current definition, we assume that it's a linear order. There are partial orders satisfying this definition – Yaël gives the example `ℕ ⊕ ℕ`, but these haven't really come up. So instead, we should just have `SuccOrder` without `le_of_lt_succ`, prove it as a theorem in the linear order case, and use those two typeclasses together when needed.

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -235,30 +202,23 @@ theorem wcovBy_succ (a : α) : a ⩿ succ a :=
 theorem covBy_succ_of_not_isMax (h : ¬IsMax a) : a ⋖ succ a :=
   (wcovBy_succ a).covBy_of_lt <| lt_succ_of_not_isMax h
 
-theorem lt_succ_iff_of_not_isMax (ha : ¬IsMax a) : b < succ a ↔ b ≤ a :=
-  ⟨le_of_lt_succ, fun h => h.trans_lt <| lt_succ_of_not_isMax ha⟩
+theorem lt_succ_of_not_isMax_of_le (ha : ¬IsMax a) : b ≤ a → b < succ a :=
```
YaelDillies, line 205: I would prefer the `_of_not_isMax` to always come last in the lemma names, for consistency. Maybe swap the hypotheses around so that it still matches the naming convention
```suggestion
theorem lt_succ_of_le_of_not_isMax (hab : b ≤ a) (ha : ¬IsMax a) : b < succ a :=
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -290,17 +250,21 @@ theorem isMax_iterate_succ_of_eq_of_ne {n m : ℕ} (h_eq : succ^[n] a = succ^[m]
   · rw [h_eq]
     exact isMax_iterate_succ_of_eq_of_lt h_eq.symm (lt_of_le_of_ne h h_ne.symm)
 
-theorem Iio_succ_of_not_isMax (ha : ¬IsMax a) : Iio (succ a) = Iic a :=
-  Set.ext fun _ => lt_succ_iff_of_not_isMax ha
+theorem subset_Iio_succ_of_not_isMax (ha : ¬IsMax a) : Iic a ⊆ Iio (succ a) :=
```
YaelDillies, line 253:
```suggestion
theorem Iic_subset_Iio_succ_of_not_isMax (ha : ¬IsMax a) : Iic a ⊆ Iio (succ a) :=
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -290,17 +250,21 @@ theorem isMax_iterate_succ_of_eq_of_ne {n m : ℕ} (h_eq : succ^[n] a = succ^[m]
...
+  fun _ => lt_succ_of_not_isMax_of_le ha
 
 theorem Ici_succ_of_not_isMax (ha : ¬IsMax a) : Ici (succ a) = Ioi a :=
   Set.ext fun _ => succ_le_iff_of_not_isMax ha
 
-theorem Ico_succ_right_of_not_isMax (hb : ¬IsMax b) : Ico a (succ b) = Icc a b := by
-  rw [← Ici_inter_Iio, Iio_succ_of_not_isMax hb, Ici_inter_Iic]
+theorem subset_Ico_succ_right_of_not_isMax (hb : ¬IsMax b) : Icc a b ⊆ Ico a (succ b) := by
```
YaelDillies, line 259:
```suggestion
theorem Icc_subset_Ico_succ_right_of_not_isMax (hb : ¬IsMax b) : Icc a b ⊆ Ico a (succ b) := by
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -290,17 +250,21 @@ theorem isMax_iterate_succ_of_eq_of_ne {n m : ℕ} (h_eq : succ^[n] a = succ^[m]
...
+  rw [← Ici_inter_Iio, ← Ici_inter_Iic]
+  gcongr
+  apply subset_Iio_succ_of_not_isMax hb
 
-theorem Ioo_succ_right_of_not_isMax (hb : ¬IsMax b) : Ioo a (succ b) = Ioc a b := by
-  rw [← Ioi_inter_Iio, Iio_succ_of_not_isMax hb, Ioi_inter_Iic]
+theorem subset_Ioo_succ_right_of_not_isMax (hb : ¬IsMax b) : Ioc a b ⊆ Ioo a (succ b) := by
```
YaelDillies, line 264:
```suggestion
theorem Ioc_subset_Ioo_succ_right_of_not_isMax (hb : ¬IsMax b) : Ioc a b ⊆ Ioo a (succ b) := by
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -316,41 +280,35 @@ theorem lt_succ (a : α) : a < succ a :=
...
   succ_le_iff_of_not_isMax <| not_isMax a
 
-theorem succ_le_succ_iff : succ a ≤ succ b ↔ a ≤ b := by simp
-
-theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := by simp
-
-alias ⟨le_of_succ_le_succ, _⟩ := succ_le_succ_iff
-
-alias ⟨lt_of_succ_lt_succ, succ_lt_succ⟩ := succ_lt_succ_iff
+theorem succ_lt_succ : a < b → succ a < succ b := by intro; simp_all
```
YaelDillies, line 290:
```suggestion
@[gcongr] theorem succ_lt_succ (hab : a < b) : succ a < succ b := by simp
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -394,6 +347,82 @@ lemma succ_eq_of_covBy (h : a ⋖ b) : succ a = b := (succ_le_of_lt h.lt).antisy
...
 alias _root_.CovBy.succ_eq := succ_eq_of_covBy
 
+section NoMaxOrder
+
+variable [NoMaxOrder α]
+
+theorem succ_eq_iff_covBy : succ a = b ↔ a ⋖ b :=
+  ⟨by
+    rintro rfl
+    exact covBy_succ _, CovBy.succ_eq⟩
```
YaelDillies, line 357: This is really supposed to be a one-liner
```suggestion
  ⟨by rintro rfl; exact covBy_succ _, CovBy.succ_eq⟩
```

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -394,6 +347,82 @@ lemma succ_eq_of_covBy (h : a ⋖ b) : succ a = b := (succ_le_of_lt h.lt).antisy
...
+
+variable [OrderTop α]
+
+@[simp]
+theorem succ_top : succ (⊤ : α) = ⊤ := by
+  rw [succ_eq_iff_isMax, isMax_iff_eq_top]
+
+-- Porting note (#10618): removing @[simp],`simp` can prove it
+theorem succ_le_iff_eq_top : succ a ≤ a ↔ a = ⊤ :=
+  succ_le_iff_isMax.trans isMax_iff_eq_top
```
YaelDillies, line 371: Is this porting note still relevant?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -60,8 +60,6 @@ class SuccOrder (α : Type*) [Preorder α] where
   max_of_succ_le {a} : succ a ≤ a → IsMax a
   /-- Proof that `succ` satisfies ordering invariants between `LT` and `LE`-/
```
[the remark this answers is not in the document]
Command-Master: How do you think of it, mathematically? I mostly thought of it as the parent in an inverted tree

Mathlib/Order/SuccPred/Basic.lean
Command-Master: Yes

Mathlib/Order/SuccPred/Basic.lean
Command-Master: I applied this, but it makes using it a bit awkward. Is there a way to keep the order while having `of_not_isMax` at the end?

Mathlib/Order/SuccPred/Basic.lean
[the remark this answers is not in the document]
YaelDillies: I think of it as the unique element that's minimally bigger than `a`, or the first element that's bigger than `a`

Mathlib/Order/SuccPred/Basic.lean
vihdzp: Do we generally keep porting notes like these around? I'd imagine once they've been read and discussed they can just be removed. After all, it's not like we're going back to Lean 3.

Mathlib/Order/SuccPred/Basic.lean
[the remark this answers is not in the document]
Command-Master: I changed the docstring, is it good?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -57,7 +57,7 @@ class SuccOrder (α : Type*) [Preorder α] where
   le_succ : ∀ a, a ≤ succ a
   /-- Proof of interaction between `succ` and maximal element-/
   max_of_succ_le {a} : succ a ≤ a → IsMax a
-  /-- Proof that `succ` satisfies ordering invariants between `LT` and `LE`-/
+  /-- Proof that `succ a` is the minimal element greater than `a`-/
```
vihdzp, line 60: Is "minimal" the correct word? Isn't the successor in a partial order always unique, meaning we can replace this by "minimum"?

vihdzp (commented):
Initial remarks (I haven't checked the Basic.lean file yet).

Mathlib/Order/SuccPred/Basic.lean
Command-Master: This definition only assumes a preorder, not a partial order

Mathlib/Order/SuccPred/LinearLocallyFinite.lean
```diff
@@ -98,18 +98,11 @@ theorem succFn_le_of_lt (i j : ι) (hij : i < j) : succFn i ≤ j := by
   rw [IsGLB, IsGreatest, mem_lowerBounds] at h
   exact h.1 j hij
 
-theorem le_of_lt_succFn (j i : ι) (hij : j < succFn i) : j ≤ i := by
```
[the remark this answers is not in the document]
Command-Master: I thought this is just `le_of_lt_succ` for the following instance, but I now see it doesn't assume LocalllyFiniteOrder. I'll restore it.

Mathlib/Order/SuccPred/Basic.lean
vihdzp: In a preorder, all minimal elements grater than `a` should be equivalent, I believe? So my statement is still approximately right.

Really, successor orders on preorders are somewhat silly. I don't think we lose anything by defining them for preorders like this, but we should really be thinking about them as either partial or total orders.

Mathlib/Order/SuccPred/Basic.lean
YaelDillies: I tend to agree with Violeta here: This porting note should instead have been a comment in the porting PR, and it's not even keeping any information that we couldn't figure out quickly: Tagging the lemma as simp would give you an error in CI iff the comment is still relevant

Mathlib/Order/SuccPred/Basic.lean
YaelDillies: Informally, "minimum element" is not something we really say, at least much less than we would say "minimal element", so I'm happy with the docstring as it is

Mathlib/Order/SuccPred/Basic.lean
Command-Master: Alright, I'll remove it.

fpvandoorn:
We can generalize the instance 
```
variable [DecidableEq α] [PartialOrder α] [OrderTop α] [SuccOrder α]
instance : SuccOrder (WithTop α) where
```
by removing the `OrderTop` variable.
In the definition of `succ` we then use `| Option.some a => ite (IsMax a) ⊤ (some (succ a))`

this also allows us to remove `succOrderOfNoMaxOrder`

fpvandoorn (approved):
Other than my above comment, LGTM

Command-Master:
> We can generalize the instance
> 
> ```
> variable [DecidableEq α] [PartialOrder α] [OrderTop α] [SuccOrder α]
> instance : SuccOrder (WithTop α) where
> ```
> 
> by removing the `OrderTop` variable (and by making the instance classical and removing `DecidableEq`). In the definition of `succ` we then use `| Option.some a => ite (IsMax a) ⊤ (some (succ a))`
> 
> this also allows us to remove `succOrderOfNoMaxOrder`

Applied. This is slightly weaker than `succOrderOfNoMaxOrder`, as it requires a partial order, but I guess that's ok.

Command-Master:
I need the theorem that a `DecidableEq` `OrderTop` has decidable `IsMax`, as well as a `NoMaxOrder`. What's a good place to put those?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -979,13 +979,13 @@ namespace WithTop
 
 section Succ
 
-variable [DecidableEq α] [PartialOrder α] [OrderTop α] [SuccOrder α]
+variable [Preorder α] [@DecidablePred α IsMax] [SuccOrder α]
```
fpvandoorn, line 982: Please just remove `[@DecidablePred α IsMax]` and use `classical` or `open scoped Classical in`. 
This predicate will only be rarely provable (e.g. only on finite types with decidable le)

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -1009,15 +1009,14 @@ instance : SuccOrder (WithTop α) where
...
     · exact coe_le_coe.2 (succ_le_of_lt h)
 
 @[simp]
-theorem succ_coe_top : succ ↑(⊤ : α) = (⊤ : WithTop α) :=
-  dif_pos rfl
+theorem succ_coe_of_isMax {a : α} (h : IsMax a) : succ ↑a = (⊤ : WithTop α) :=
+  dif_pos h
 
-theorem succ_coe_of_ne_top {a : α} (h : a ≠ ⊤) : succ (↑a : WithTop α) = ↑(succ a) :=
+theorem succ_coe_of_not_isMax {a : α} (h : ¬ IsMax a) : succ (↑a : WithTop α) = ↑(succ a) :=
```
fpvandoorn, line 1019: maybe add a (simp?)-lemma in a `NoMaxOrder`?

Mathlib/Order/SuccPred/Basic.lean
Command-Master: This is actually provable in a partial order with decidable equality, thanks to `succ_eq_iff_isMax`. Should I just use that?

fpvandoorn: Ah, good point. I'm fine with either option.

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -424,6 +451,30 @@ section NoMaxOrder
...
+
+theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := by simp
+
+alias ⟨le_of_succ_le_succ, _⟩ := succ_le_succ_iff
+
+alias ⟨lt_of_succ_lt_succ, _⟩ := succ_lt_succ_iff
+
+@[simp]
+theorem Iio_succ (a : α) : Iio (succ a) = Iic a :=
+  Iio_succ_of_not_isMax <| not_isMax _
```
YaelDillies, line 468: ... and this is true in a succ-archimedean order with bot. Same, please write a TODO

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -424,6 +451,30 @@ section NoMaxOrder
...
+
+@[simp]
+theorem Ico_succ_right (a b : α) : Ico a (succ b) = Icc a b :=
+  Ico_succ_right_of_not_isMax <| not_isMax _
+
+@[simp]
+theorem Ioo_succ_right (a b : α) : Ioo a (succ b) = Ioc a b :=
+  Ioo_succ_right_of_not_isMax <| not_isMax _
```
YaelDillies, line 476: This is still true in a non-linear succ-archimedean order. Mind writing a TODO?

Command-Master: Like so?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -629,41 +648,37 @@ theorem pred_lt (a : α) : pred a < a :=
...
-  pred_lt_iff_of_not_isMin <| not_isMin a
+theorem pred_lt_of_le : a ≤ b → pred a < b :=
+  pred_lt_of_not_isMin_of_le <| not_isMin a
 
 @[simp]
 theorem le_pred_iff : a ≤ pred b ↔ a < b :=
   le_pred_iff_of_not_isMin <| not_isMin b
 
-theorem pred_le_pred_iff : pred a ≤ pred b ↔ a ≤ b := by simp
+theorem pred_le_pred_of_le : a ≤ b → pred a ≤ pred b := by intro; simp_all
```
vihdzp, line 658: Does `simp` not work?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -907,13 +983,13 @@ namespace WithTop
...
 
-variable [DecidableEq α] [PartialOrder α] [OrderTop α] [SuccOrder α]
+variable [DecidableEq α] [PartialOrder α] [SuccOrder α]
 
 instance : SuccOrder (WithTop α) where
   succ a :=
     match a with
     | ⊤ => ⊤
-    | Option.some a => ite (a = ⊤) ⊤ (some (succ a))
+    | Option.some a => ite (succ a = a) ⊤ (some (succ a))
```
vihdzp, line 992: Why doesn't the old definition work anymore?

vihdzp (approved):
Looks good to me, just a few minor questions.

Mathlib/Order/SuccPred/Basic.lean
vihdzp: Sorry, "least" element would be the proper wording.

Can we at least still say that it is unique in the case of partial orders?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -314,41 +280,35 @@ theorem lt_succ (a : α) : a < succ a :=
...
   succ_le_iff_of_not_isMax <| not_isMax a
 
-theorem succ_le_succ_iff : succ a ≤ succ b ↔ a ≤ b := by simp
-
-theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := by simp
-
-alias ⟨le_of_succ_le_succ, _⟩ := succ_le_succ_iff
-
-alias ⟨lt_of_succ_lt_succ, succ_lt_succ⟩ := succ_lt_succ_iff
+@[gcongr] theorem succ_lt_succ (hab : a < b) : succ a < succ b := by simp [hab]
```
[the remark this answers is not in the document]
YaelDillies: Don't think that's an improvement?

Mathlib/Order/SuccPred/Basic.lean
Command-Master: As [mentioned](https://github.com/leanprover-community/mathlib4/pull/15881#issuecomment-2296312137) above, using the new definition the instance can be generalized to avoid the `OrderTop` assumption.

Mathlib/Order/SuccPred/Basic.lean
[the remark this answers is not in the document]
vihdzp: I thought that was just the Lean 4 style now...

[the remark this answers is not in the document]
YaelDillies: That's how everything was mathported. Doesn't mean it's the style

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -424,6 +451,32 @@ section NoMaxOrder
...
+theorem lt_succ_iff : a < succ b ↔ a ≤ b :=
+  lt_succ_iff_of_not_isMax <| not_isMax b
+
+theorem succ_le_succ_iff : succ a ≤ succ b ↔ a ≤ b := by simp
+
+theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := by simp
+
+alias ⟨le_of_succ_le_succ, _⟩ := succ_le_succ_iff
+
+alias ⟨lt_of_succ_lt_succ, _⟩ := succ_lt_succ_iff
```
YaelDillies, line 464: In general, we like to glue together series of similar lemmas. It's a visual cue that they are related
```suggestion
theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := by simp

alias ⟨le_of_succ_le_succ, _⟩ := succ_le_succ_iff
alias ⟨lt_of_succ_lt_succ, _⟩ := succ_lt_succ_iff
```

YaelDillies (commented):
Feel free to address https://github.com/leanprover-community/mathlib4/pull/15881/files#r1728190836 or mark it as resolved

Mathlib/Order/SuccPred/Basic.lean
Command-Master: I changed it to least.

fpvandoorn:
Thanks!

bors merge

PR #15882, opened 2024-08-16
[Merged by Bors] - feat: Doubling and difference constants of two finsets

YaelDillies:
Define the doubling and difference constants of two finsets in a group. The doubling and difference constants being small are one way to measure how close a set is from being a subgroup. This makes it a central concept in additive combinatorics and related fields.

From LeanAPAP

---


b-mehta (approved):
This looks generally pretty good! Lots of things we could add to these files, but this is a good start.

b-mehta (approved):
bors merge

PR #15883, opened 2024-08-16
[Merged by Bors] - feat: Expectation of a function over  a finset

YaelDillies:
Define the expectation of a function `f : ι → M` over a finset `s : Finset ι` as `∑ i ∈ s, f i` divided by `s.card`.

Naïvely, this requires `M` to be a semifield but, since `s.card⁻¹` is rational, we can get away with assuming `M` to be a `NNRat`-module. This generality is important to talk about expectation of functions.

From LeanAPAP

---


Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,400 @@
...
+@[simp] lemma expect_empty (f : ι → M) : expect ∅ f = 0 := by simp [expect]
+@[simp] lemma expect_singleton (f : ι → M) (i : ι) : expect {i} f = f i := by simp [expect]
+@[simp] lemma expect_const_zero (s : Finset ι) : 𝔼 _i ∈ s, (0 : M) = 0 := by simp [expect]
+
+@[congr]
+lemma expect_congr {t : Finset ι} (hst : s = t) (h : ∀ i ∈ t, f i = g i) :
+    𝔼 i ∈ s, f i = 𝔼 i ∈ t, g i := by rw [expect, expect, sum_congr hst h, hst]
+
+lemma expectWith_congr (p : ι → Prop) [DecidablePred p] (h : ∀ i ∈ s, p i → f i = g i) :
+    𝔼 i ∈ s with p i, f i = 𝔼 i ∈ s with p i, g i :=
```
Command-Master, line 110: This is more `expectWith_congrFun`, not a fully general congruence, which could be
```lean
lemma expectWith_congr (p q : ι → Prop) [DecidablePred p] [DecidablePred q] (hst : s = t)
    (hpq : ∀ i ∈ s, p i ↔ q i) (h : ∀ i ∈ s, p i → f i = g i) :
    𝔼 i ∈ s with p i, f i = 𝔼 i ∈ t with q i, g i
```

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,397 @@
...
+## Notation
+
+* `𝔼 i ∈ s, f i` is notation for `Finset.expect s f`. It is the expectation of `f i` where `i`
+  ranges over the finite set `s` (either a `Finset` or a `Set` with a `Fintype` instance).
+* `𝔼 i, f i` is notation for `Finset.expect Finset.univ f`. It is the expectation of `f i` where `i`
+  ranges over the finite domain of `f`.
+* `𝔼 i ∈ s with p i, f i` is notation for `Finset.expect (Finset.filter p s) f`. This is referred to
+  as `expectWith` in lemma names.
+* `𝔼 (i ∈ s) (j ∈ t), f i j` is notation for `Finset.expect (s ×ˢ t) (fun ⟨i, j⟩ ↦ f i j)`.
+-/
```
b-mehta, line 26:
```suggestion
  as `expectWith` in lemma names.
* `𝔼 (i ∈ s) (j ∈ t), f i j` is notation for `Finset.expect (s ×ˢ t) (fun ⟨i, j⟩ ↦ f i j)`.

## TODO

* Connect `Finset.expect` with the expectation over `s` in the probability theory sense.
-/
```

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,397 @@
...
+      if ppDomain then
+        let ty ← withNaryArg 0 delab
+        `(bigOpBinder| $(.mk i):ident : $ty)
+      else
+        `(bigOpBinder| $(.mk i):ident)
+    `(𝔼 $binder:bigOpBinder, $body)
+  else
+    let ss ← withNaryArg 3 $ delab
+    `(𝔼 $(.mk i):ident ∈ $ss, $body)
+
```
eric-wieser, line 87: Can you dedup this with the `Finset.sum` printer somehow?

YaelDillies: Sounds too fragile. I would need to keep the argument orders in sync (note that `expect` has a `Module _ _` argument which `sum` and `prod` don't have OR make the general delaborator parametric over the argument order, but I don't know how to make a parametric match

Command-Master:
This is [Finset.centerMass](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/Convex/Combination.html#Finset.centerMass) with a constant weight of `1`, right?

YaelDillies:
Yes and no:
* Yes, there is a common generalisation of `Finset.centerMass` and `Finset.expect`
* No, this common generalisation is far away, because I have a huge refactor of convexity in the works and I can't ensure `Finset.centerMass` will stay the way it is

b-mehta:
Unrequesting my review for this because I wrote early versions of this API

eric-wieser:
Should I take that as you bring happy with this version, @b-mehta? I agree that means you probably shouldn't merge, but an approval would be a valuable signal!

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,402 @@
...
+namespace Finset
+section AddCommMonoid
+variable [AddCommMonoid M] [Module ℚ≥0 M] [AddCommMonoid N] [Module ℚ≥0 N] {s t : Finset ι}
+  {f g : ι → M} {m : N → M} {p q : ι → Prop} [DecidablePred p] [DecidablePred q]
+
+lemma expect_univ [Fintype ι] : 𝔼 i, f i = (∑ i, f i) /ℚ Fintype.card ι := by
+  rw [expect, card_univ]
+
+@[simp] lemma expect_empty (f : ι → M) : expect ∅ f = 0 := by simp [expect]
```
b-mehta, line 105: Let's use the notation here and below

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,402 @@
...
+    (h : ∀ a ∈ s, f a = g (i a)) : 𝔼 i ∈ s, f i = 𝔼 i ∈ t, g i := by
+  simp_rw [expect, card_nbij' i j hi hj left_inv right_inv,
+    sum_nbij' i j hi hj left_inv right_inv h]
+
+/-- `Finset.expect_equiv` is a specialization of `Finset.expect_bij` that automatically fills in
+most arguments. -/
+lemma expect_equiv (e : ι ≃ κ) (hst : ∀ i, i ∈ s ↔ e i ∈ t) (hfg : ∀ i ∈ s, f i = g (e i)) :
+    𝔼 i ∈ s, f i = 𝔼 i ∈ t, g i := by simp_rw [expect, card_equiv e hst, sum_equiv e hst hfg]
+
+lemma expect_product' (f : ι → κ → M) : 𝔼 i ∈ s ×ˢ t, f i.1 i.2 = 𝔼 i ∈ s, 𝔼 y ∈ t, f i y := by
```
b-mehta, line 230: Let's add expect_product too

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,402 @@
...
+
+end Semiring
+
+section CommSemiring
+variable [CommSemiring M] [Module ℚ≥0 M] [IsScalarTower ℚ≥0 M M] [SMulCommClass ℚ≥0 M M]
+
+local notation:70 s:70 " ^^ " n:71 => Fintype.piFinset fun _ : Fin n ↦ s
+
+lemma expect_pow (s : Finset ι) (f : ι → M) (n : ℕ) :
+    (𝔼 i ∈ s, f i) ^ n = 𝔼 p ∈ s ^^ n, ∏ i, f (p i) := by
```
b-mehta, line 311: Not sure the notation is worth it here

b-mehta (approved):

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,403 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yaël Dillies, Bhavik Mehta
+-/
+import Mathlib.Algebra.Algebra.Rat
+import Mathlib.Algebra.BigOperators.Ring
+import Mathlib.Data.Finset.Pointwise.Basic
+import Mathlib.Data.Fintype.BigOperators
+import Mathlib.GroupTheory.GroupAction.BigOperators
+
+/-!p
```
b-mehta, line 12: typo?

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,403 @@
...
+lemma _root_.Fintype.expect_eq_sum_div_card [Fintype ι] (f : ι → M) :
+    𝔼 i, f i = (∑ i, f i) / Fintype.card ι := Finset.expect_eq_sum_div_card _ _
+
+lemma expect_div (s : Finset ι) (f : ι → M) (a : M) : (𝔼 i ∈ s, f i) / a = 𝔼 i ∈ s, f i / a := by
+  simp_rw [div_eq_mul_inv, expect_mul]
+
+end Semifield
+end Finset
+
+namespace algebraMap
```
b-mehta, line 344: Don't you want AlgebraMap here?

b-mehta (approved):
I'm happy modulo changes mentioned above!

Mathlib/Algebra/BigOperators/Expect.lean
b-mehta: Weird, but okay!

Parcly-Taxel (approved):

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,403 @@
...
+    (h : ∀ a ∈ s, f a = g (i a)) : 𝔼 i ∈ s, f i = 𝔼 i ∈ t, g i := by
+  simp_rw [expect, card_nbij' i j hi hj left_inv right_inv,
+    sum_nbij' i j hi hj left_inv right_inv h]
+
+/-- `Finset.expect_equiv` is a specialization of `Finset.expect_bij` that automatically fills in
+most arguments. -/
+lemma expect_equiv (e : ι ≃ κ) (hst : ∀ i, i ∈ s ↔ e i ∈ t) (hfg : ∀ i ∈ s, f i = g (e i)) :
+    𝔼 i ∈ s, f i = 𝔼 i ∈ t, g i := by simp_rw [expect, card_equiv e hst, sum_equiv e hst hfg]
+
+lemma expect_product (f : ι × κ → M) : 𝔼 x ∈ s ×ˢ t, f x = 𝔼 i ∈ s, 𝔼 j ∈ t, f (i, j) := by
```
b-mehta, line 230: Let's add a comment about the rewrite direction of these, like we have in the finset versions of similar lemmas

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,403 @@
...
+  ranges over the finite set `s` (either a `Finset` or a `Set` with a `Fintype` instance).
+* `𝔼 i, f i` is notation for `Finset.expect Finset.univ f`. It is the expectation of `f i` where `i`
+  ranges over the finite domain of `f`.
+* `𝔼 i ∈ s with p i, f i` is notation for `Finset.expect (Finset.filter p s) f`. This is referred to
+  as `expectWith` in lemma names.
+* `𝔼 (i ∈ s) (j ∈ t), f i j` is notation for `Finset.expect (s ×ˢ t) (fun ⟨i, j⟩ ↦ f i j)`.
+
+## TODO
+
+Connect `Finset.expect` with the expectation over `s` in the probability theory sense.
```
b-mehta, line 29:
```suggestion
Connect `Finset.expect` with the expectation over `s` in the probability theory sense.
Give a formulation of Jensen's inequality and the Cauchy-Schwarz inequality in this language.
```

Mathlib/Algebra/BigOperators/Expect.lean
```diff
@@ -0,0 +1,403 @@
...
+-/
```
b-mehta, line 30: imo we should add an implementation note here explaining why this definition is helpful to have, as it isn't immediately clear otherwise. That is, why is this useful when probability expectation exists, and why is it not just an abbreviation on top of Finset.sum. Of course I'm fully on board with this definition being in mathlib, but to future readers / contributors these factors may not be clear.

b-mehta (approved):
LGTM, but as mentioned above I'd like others to review and merge, not me.

Mathlib/Algebra/BigOperators/Expect.lean
YaelDillies: I'm not sure I addressed exactly your point. Please comment if you want further modifications

jcommelin (commented):
Thanks :tada:

bors merge

riccardobrasca:
bors d+

YaelDillies:
bors merge

PR #15888, opened 2024-08-16
[Merged by Bors] - chore(.github/workflows/nightly_detect_failure.yml): fix the sha hash

jcommelin:
The sha hash was only printed to the github env if the git tag
"nightly-testing-${version}" did not yet exist.

We move this out of the if clause, so that it is always printed.

---


Ruben-VandeVelde (approved):
Seems plausible

maintainer merge

bryangingechen:
Thanks!
bors r+

PR #15890, opened 2024-08-16
[Merged by Bors] - feat: interior, closure and frontier of Euclidean half-spaces

grunweg:
Preparatory lemmas towards proving that the real interval $[x,y]$ has boundary {x,y} as a smooth manifold.

---

### DIFF
--- a/Mathlib/Geometry/Manifold/Instances/Real.lean
+++ b/Mathlib/Geometry/Manifold/Instances/Real.lean
@@ -90,6 +90,26 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
   Subtype.range_val
 @[deprecated (since := "2024-04-05")] alias range_half_space := range_euclideanHalfSpace
 
+-- TODO: generalise these lemmas to other values of `p`
+
+theorem interior_halfspace {a : ℝ} {n : ℕ} (i : Fin n) :
+    interior { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a < y i } := by
+  let f : (EuclideanSpace ℝ (Fin n)) →L[ℝ] ℝ := ContinuousLinearMap.proj i
+  change interior (f ⁻¹' Set.Ici a) = f ⁻¹' Set.Ioi a
+  rw [f.interior_preimage (Function.surjective_eval _), interior_Ici]
+
+theorem closure_halfspace {a : ℝ} {n : ℕ} (i : Fin n) :
+    closure { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a ≤ y i } := by
+  let f : (EuclideanSpace ℝ (Fin n)) →L[ℝ] ℝ := ContinuousLinearMap.proj i
+  change closure (f ⁻¹' Set.Ici a) = f ⁻¹' Set.Ici a
+  rw [f.closure_preimage (Function.surjective_eval _), closure_Ici]
+
+theorem frontier_halfspace {a : ℝ} {n : ℕ} (i : Fin n) :
+    frontier { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a = y i } := by
+  rw [frontier, interior_halfspace, closure_halfspace]
+  ext y
+  simpa only [mem_diff, mem_setOf_eq, not_lt] using antisymm_iff
+
 theorem range_euclideanQuadrant (n : ℕ) :
     (range fun x : EuclideanQuadrant n => x.val) = { y | ∀ i : Fin n, 0 ≤ y i } :=
   Subtype.range_val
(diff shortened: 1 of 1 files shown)


Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -90,6 +90,26 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
   Subtype.range_val
 @[deprecated (since := "2024-04-05")] alias range_half_space := range_euclideanHalfSpace
 
+-- TODO: generalise these lemmas to other values of `p`
+
+theorem interior_halfspace {a : ℝ} {n : ℕ} (i : Fin n) :
+    interior { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a < y i } := by
+  let f : (EuclideanSpace ℝ (Fin n)) →L[ℝ] ℝ := ContinuousLinearMap.proj i
```
eric-wieser, line 97:
```suggestion
  let f : EuclideanSpace ℝ (Fin n) →L[ℝ] ℝ := ContinuousLinearMap.proj i
```
and below

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -90,6 +90,26 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
   Subtype.range_val
 @[deprecated (since := "2024-04-05")] alias range_half_space := range_euclideanHalfSpace
 
+-- TODO: generalise these lemmas to other values of `p`
```
j-loreaux, line 93: Why not do this now? I think it's relatively easy (note: I also don't like the `change` utilizing the defeq there).
```lean
open ENNReal in
theorem interior_halfspace {a : ℝ} {n : ℕ} {p : ℝ≥0∞} (i : Fin n) :
    interior { y : PiLp p (fun _ : Fin n ↦ ℝ) | a ≤ y i } = { y | a < y i } := by
  let f : PiLp p (fun _ : Fin n ↦ ℝ) →L[ℝ] ℝ := ContinuousLinearMap.proj i
  simpa [interior_Ici] using f.interior_preimage (Function.surjective_eval _) (Ici a)
```
This allows your lemma as a corollary, even using `with_reducible` since `EuclideanSpace` is an `abbrev`
```lean
example {a : ℝ} {n : ℕ} (i : Fin n) :
    interior { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a < y i } := by
  with_reducible exact interior_halfspace i
```
Finally, I'm not sure I understand the choices behind your use of explicit / implicit variables here. Why is `i` explicit but `a` and `n` implicit?

grunweg: I was thinking of allowing different component spaces... in which case I'd even need to think about the statement.
I like your generalisation and proof, thanks!

I had simply overlooked that i could be implicit as well, and just made it so.

grunweg:
Thanks for the useful feedback! I have addressed all comments.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -90,6 +90,25 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
...
 
+open ENNReal in
+theorem interior_halfspace {n : ℕ} {p : ℝ≥0∞} {a : ℝ} {i : Fin n} :
+    interior { y : PiLp p (fun _ : Fin n ↦ ℝ) | a ≤ y i } = { y | a < y i } := by
+  let f : PiLp p (fun _ : Fin n ↦ ℝ) →L[ℝ] ℝ := ContinuousLinearMap.proj i
+  simpa [interior_Ici] using f.interior_preimage (Function.surjective_eval _) (Ici a)
+
+open ENNReal in
+theorem closure_halfspace {n : ℕ} {p : ℝ≥0∞} {a : ℝ} {i : Fin n} :
+    closure { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a ≤ y i } := by
```
eric-wieser, line 101:
```suggestion
    closure { y : PiLp p (fun _ : Fin n ↦ ℝ) | a ≤ y i } = { y | a ≤ y i } := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -90,6 +90,25 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
...
+  let f : PiLp p (fun _ : Fin n ↦ ℝ) →L[ℝ] ℝ := ContinuousLinearMap.proj i
+  simpa [closure_Ici] using f.closure_preimage (Function.surjective_eval _) (Ici a)
+
+open ENNReal in
+theorem frontier_halfspace {n : ℕ} {p : ℝ≥0∞} {a : ℝ} {i : Fin n} :
+    frontier { y : EuclideanSpace ℝ (Fin n) | a ≤ y i } = { y | a = y i } := by
```
eric-wieser, line 107:
```suggestion
    frontier { y : PiLp p (fun _ : Fin n ↦ ℝ) | a ≤ y i } = { y | a = y i } := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -90,6 +90,25 @@ theorem range_euclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
...
+  rw [frontier, closure_halfspace (p := p), interior_halfspace]
```
eric-wieser, line 108:
```suggestion
  rw [frontier, closure_halfspace, interior_halfspace]
```
you only needed `p` because it was not part of the statement without my corrections above.

j-loreaux:
Also, I would argue the first 3 lemmas should be `@[simp]` and that the `p`, `a`, and `i` arguments should be explicit.

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: :facepalm: Indeed, thanks for reading properly!

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -157,6 +176,25 @@ scoped[Manifold]
     (modelWithCornersEuclideanHalfSpace n :
       ModelWithCorners ℝ (EuclideanSpace ℝ (Fin n)) (EuclideanHalfSpace n))
 
+lemma range_modelWithCornersEuclideanHalfSpace (n : ℕ) [Zero (Fin n)] :
```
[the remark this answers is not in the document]
grunweg: Done; I have performed the same change to the other instances in this file.

grunweg:
Thanks again; all comments have been addressed.

j-loreaux:
Hmmm... I didn't realize they were littered throughout (my fault!).

@sgouezel can you explain the preference this file has for `[Zero (Fin n)]` over `[NeZero n]`? It seems a strange choice to me.

sgouezel:
I guess it dates back to the time where `NeZero n` wasn't a thing yet. `[NeZero n]` is clearly better.

j-loreaux:
bors merge

grunweg:
Thanks for the fast reviews!

PR #15891, opened 2024-08-16
[Merged by Bors] - feat: lemmas about the left and right boundary point under `Icc{Left,Right}Chart`

grunweg:
Another step towards proving that a non-empty the real interval $[x,y]$ has boundary ${ x,y }$ (as a manifold).

Add missing `Icc.coe_top` and `Icc.coe_bot` lemmas.

---


Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -245,6 +245,33 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+    _ = 0 := rfl
+
+lemma IccLeftChart_extend_interior_pos {p : Set.Icc x y} (hp : x < p.val ∧ p.val < y) :
+    (((IccLeftChart x y).extend (𝓡∂ 1)) p) 0 > 0 := by
+  set lhs := (IccLeftChart x y).extend (𝓡∂ 1) p
+  have : lhs 0 = p.val - x := rfl
+  rw [this]
+  norm_num [hp.1]
+
+lemma IccLeftChart_extend_left_mem_frontier :
```
grunweg, line 270: I'm open to the argument that this lemma (and its right counterpart) is sufficiently trivial to be inlined.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -245,6 +245,33 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
     have B : Continuous fun z : EuclideanSpace ℝ (Fin 1) => z 0 := continuous_apply 0
     exact (A.comp B).comp continuous_subtype_val
 
+variable {x y : ℝ} [hxy : Fact (x < y)]
+
+/-- The endpoint `x ∈ Icc x y`, as a point in `Icc x y` (assuming `x < y`). -/
+abbrev IccManifold.left : Icc x y := ⟨x, ⟨le_refl x, by have := hxy.out; linarith⟩⟩
+
+/-- The endpoint `y ∈ Icc x y`, as a point in `Icc x y` (assuming `x < y`). -/
+abbrev IccManifold.right : Icc x y := ⟨y, ⟨by have := hxy.out; linarith, le_refl y⟩⟩
```
semorrison, line 254: This has nothing to do with the manifold setup. Shouldn't this be available to anyone working with the set `Icc`?

grunweg: Good point: I moved this to the definition of Icc. The names `Icc.left` and `Icc.right` might cause conflicts; if you prefer a different name, I'm open to changing it.

grunweg:
@YaelDillies In this PR, I'm adding definitions for "the left endpoint of `Icc a b`, as a term of type `Icc a b`" (and similar). I'm not perfectly happy with the naming `left` and `right` for the endpoints; there is one place in this PR where I had to fix a name conflict. Do you have ideas for a better name?

YaelDillies:
Isn't that just `⊥` and `⊤`? See `Set.Icc.orderBot` and `Set.Icc.orderTop`.

grunweg:
Thanks for pointing this out! I'll take a look.

grunweg:
I checked: `Order/LatticeIntervals.lean` in fact contains such instances by now, so there was in fact nothing to do for me.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +294,31 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
     have B : Continuous fun z : EuclideanSpace ℝ (Fin 1) => z 0 := continuous_apply 0
     exact (A.comp B).comp continuous_subtype_val
 
+variable {x y : ℝ} [hxy : Fact (x < y)]
+
+-- TODO: how to avoid passing hxy_twice?
```
grunweg, line 299: I'd appreciate help with this question: I'm not sure how to solve this. Removing `hxy_twice`, even if writing `include hxy in` before the theorem, yields an error "failed to synthesize Fact (x ≤ y)".

The same issue appears three more times below.

grunweg: @YaelDillies I know you're not a differential geometer, but since you're looking right now: do you happen to know the answer? This is just a Lean question, I presume.

YaelDillies: Maybe an instance `Fact (a < b) -> Fact (a <= b)` is sensible

YaelDillies: Btw please include more context by making a multiline comment next time. It's really painful to navigate from this comment to the relevant piece of code on mobile

grunweg: Sorry, I'll try to include more context next time. Thanks for the tip, the solution worked (I made it a scoped instance).

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +294,33 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+open Fact.Manifold
+
+lemma IccLeftChart_extend_left_eq :
+    ((IccLeftChart x y).extend (𝓡∂ 1)) (Icc.instOrderBotElem.bot) = 0 := by
+  calc ((IccLeftChart x y).extend (𝓡∂ 1)) (Icc.instOrderBotElem.bot)
+    _ = (𝓡∂ 1) ⟨fun _ ↦ 0, by norm_num⟩ := by norm_num [IccLeftChart, Icc.coe_bot]
+    _ = 0 := rfl
+
+lemma IccLeftChart_extend_interior_pos {p : Set.Icc x y} (hp : x < p.val ∧ p.val < y) :
+    (((IccLeftChart x y).extend (𝓡∂ 1)) p) 0 > 0 := by
```
Ruben-VandeVelde, line 314: This should be `0 <`, right? And do you really need all the parentheses?

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +294,33 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+
+end Fact.Manifold
+
+open Fact.Manifold
+
+lemma IccLeftChart_extend_left_eq :
+    ((IccLeftChart x y).extend (𝓡∂ 1)) (Icc.instOrderBotElem.bot) = 0 := by
+  calc ((IccLeftChart x y).extend (𝓡∂ 1)) (Icc.instOrderBotElem.bot)
+    _ = (𝓡∂ 1) ⟨fun _ ↦ 0, by norm_num⟩ := by norm_num [IccLeftChart, Icc.coe_bot]
+    _ = 0 := rfl
```
Ruben-VandeVelde, line 311: Seems like there's a lot of `rfl`s doing the heavy lifting. Are we missing API?

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +294,33 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+namespace Fact.Manifold
+
+scoped instance : Fact (x ≤ y) := Fact.mk hxy.out.le
+
+end Fact.Manifold
+
+open Fact.Manifold
+
+lemma IccLeftChart_extend_left_eq :
+    ((IccLeftChart x y).extend (𝓡∂ 1)) (Icc.instOrderBotElem.bot) = 0 := by
```
jcommelin, line 308: Why do you need the spelling `Icc.instOrderBotElem.bot`? Is there a way to make that a bit nicer?

grunweg: Thanks for making me check: there is a nicer way. Changed.

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: I do not; changed.

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: I've fixed all but three rfls; those seem to be missing API/ I certainly can't find them. Help finding those is welcome.

grunweg: I have added these lemmas as helper results now.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -229,6 +228,9 @@ scoped[Manifold]
     (modelWithCornersEuclideanHalfSpace n :
       ModelWithCorners ℝ (EuclideanSpace ℝ (Fin n)) (EuclideanHalfSpace n))
 
+lemma modelWithCornersEuclideanHalfSpace_zero {n : ℕ} [NeZero n] :
+    (𝓡∂ n) ⟨fun _ ↦ 0, by norm_num⟩ = 0 := rfl
```
YaelDillies, line 232: Do you need the brackets here?
```suggestion
    𝓡∂ n ⟨fun _ ↦ 0, by norm_num⟩ = 0 := rfl
```

YaelDillies, line 232: Why is this not
```suggestion
    (𝓡∂ n) 0 = 0 := rfl
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +297,32 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+
+namespace Fact.Manifold
+
+scoped instance : Fact (x ≤ y) := Fact.mk hxy.out.le
+
+end Fact.Manifold
+
+open Fact.Manifold
+
+lemma IccLeftChart_extend_left_eq : (IccLeftChart x y).extend (𝓡∂ 1) ⊥ = 0 := by
```
YaelDillies, line 310:
```suggestion
lemma IccLeftChart_extend_bot : (IccLeftChart x y).extend (𝓡∂ 1) ⊥ = 0 := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +297,32 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+  norm_num [IccLeftChart, modelWithCornersEuclideanHalfSpace_zero]
+
+lemma iccLeftChart_extend_zero {p : Set.Icc x y} :
+    (IccLeftChart x y).extend (𝓡∂ 1) p 0 = p.val - x := rfl
+
+lemma IccLeftChart_extend_interior_pos {p : Set.Icc x y} (hp : x < p.val ∧ p.val < y) :
+    0 < ((IccLeftChart x y).extend (𝓡∂ 1)) p 0 := by
```
YaelDillies, line 317:
```suggestion
    0 < (IccLeftChart x y).extend (𝓡∂ 1) p 0 := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -295,6 +297,32 @@ def IccLeftChart (x y : ℝ) [h : Fact (x < y)] :
...
+  simp_rw [iccLeftChart_extend_zero]
+  norm_num [hp.1]
+
+lemma IccLeftChart_extend_left_mem_frontier :
+    (IccLeftChart x y).extend (𝓡∂ 1) ⊥ ∈ frontier (range (𝓡∂ 1)) := by
```
YaelDillies, line 322:
```suggestion
lemma IccLeftChart_extend_bot_mem_frontier :
    (IccLeftChart x y).extend (𝓡∂ 1) ⊥ ∈ frontier (range (𝓡∂ 1)) := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -343,6 +371,15 @@ def IccRightChart (x y : ℝ) [h : Fact (x < y)] :
     have B : Continuous fun z : EuclideanSpace ℝ (Fin 1) => z 0 := continuous_apply 0
     exact (A.comp B).comp continuous_subtype_val
 
+lemma IccRightChart_extend_right_eq :
+    (IccRightChart x y).extend (𝓡∂ 1) ⊤ = 0 := by
```
YaelDillies, line 375:
```suggestion
lemma IccRightChart_extend_top :
    (IccRightChart x y).extend (𝓡∂ 1) ⊤ = 0 := by
```

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -343,6 +371,15 @@ def IccRightChart (x y : ℝ) [h : Fact (x < y)] :
...
+  norm_num [IccRightChart, modelWithCornersEuclideanHalfSpace_zero]
+
+lemma IccRightChart_extend_right_mem_frontier :
+    (IccRightChart x y).extend (𝓡∂ 1) ⊤ ∈ frontier (range (𝓡∂ 1)) := by
```
YaelDillies, line 379:
```suggestion
lemma IccRightChart_extend_top_mem_frontier :
    (IccRightChart x y).extend (𝓡∂ 1) ⊤ ∈ frontier (range (𝓡∂ 1)) := by
```

Mathlib/Order/LatticeIntervals.lean
```diff
@@ -194,10 +194,16 @@ variable [Preorder α] [Fact (a ≤ b)]
 instance : OrderBot (Icc a b) :=
   (isLeast_Icc Fact.out).orderBot
 
+@[simp]
+theorem coe_bot : ↑(⊥ : Icc a b) = a := rfl
```
YaelDillies, line 198:
```suggestion
@[simp, norm_cast]
theorem coe_bot : ↑(⊥ : Icc a b) = a := rfl
```

Mathlib/Order/LatticeIntervals.lean
```diff
@@ -194,10 +194,16 @@ variable [Preorder α] [Fact (a ≤ b)]
...
+
 /-- `Icc a b` has a top element whenever `a ≤ b`. -/
 instance : OrderTop (Icc a b) :=
   (isGreatest_Icc Fact.out).orderTop
 
+@[simp]
+theorem coe_top : ↑(⊤ : Icc a b) = b := rfl
```
YaelDillies, line 205:
```suggestion
@[simp, norm_cast]
theorem coe_top : ↑(⊤ : Icc a b) = b := rfl
```

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: I do, otherwise Lean complains `function expected`. (The notation 𝓡∂ takes exactly one argument.)

grunweg: This would yield `failed to synthesize
  OfNat (EuclideanHalfSpace n) 0`

grunweg:
Thanks for the nits; comments addressed.

Mathlib/Geometry/Manifold/Instances/Real.lean
YaelDillies: That sounds like a missing `withOverApp` or something like that. Can you start a topic on Zulip?

YaelDillies: Wait, the Euclidean half plane doesn't have a zero currently?!

grunweg: Indeed, it didn't. Added a zero instance (also for the euclidean quadrant) and simplified the proof.

grunweg: Asked on zulip: let's see what comes out of it.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -78,6 +77,12 @@ instance [NeZero n] : Inhabited (EuclideanHalfSpace n) :=
 instance : Inhabited (EuclideanQuadrant n) :=
   ⟨⟨0, fun _ => le_rfl⟩⟩
 
+instance {n : ℕ} [NeZero n] : Zero (EuclideanHalfSpace n) where
+  zero :=  ⟨fun _ ↦ 0, by norm_num⟩
```
YaelDillies, line 81:
```suggestion
  zero.val := 0
  zero.prop := by norm_num
```
if you like (and same below)

YaelDillies (approved):
Thanks!

maintainer delegate

jcommelin (commented):
bors d+

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: Why do *you* prefer it this way? (I don't have a strong opinion. I may argue that mathematically, not much is happening, so golfing it into one line feels fine.)

grunweg:
Thanks for the reviews!
bors merge

PR #15892, opened 2024-08-16
[Merged by Bors] - feat: the real manifold [x,y] has boundary {x,y}

grunweg:
Also rename one lemma added in #15891, which I forgot to rename there.

---


Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -382,7 +384,7 @@ lemma IccRightChart_extend_top :
   norm_num [IccRightChart, modelWithCornersEuclideanHalfSpace_zero]
   congr
 
-lemma IccRightChart_extend_right_mem_frontier :
+lemma IccRightChart_extend_top_mem_frontier :
```
grunweg, line 387: This was merged today: if this PR gets reviewed quickly, I think omitting a deprecation is fine.

YaelDillies:
Is it not true in general that the manifold boundary of a closed set `C : Set ℝⁿ` (which is a manifold) is `frontier C`? How much harder would it be to prove that directly?

grunweg:
This is an obvious question. Let's assume we talk about a closed subset of `C : Set R^n` which is the closure of an open set.
(Otherwise, we could be talking about closed submanifolds of R^n... for which this is harder (and mathlib currently doesn't know about submanifolds).)

The mathlib definition is "a point is a boundary point if it is mapped to the frontier via its preferred chart" (i.e., `extChartAt H x x` lies in `frontier (range (extChartAt H x))`). Since closed intervals are defined using two charts, a priori you have to do *some* work to deal with this --- which is what this PR does.

Mathematically, you can say "but 0 lies in the frontier under the obvious identity chart to R".
Formalising this requires the innocuous-looking lemma "if x is a boundary point is independent of the choice of chart".
This lemma has a simple proof for 1-manifolds and a harder proof for 2-manifolds (using the fundamental group of spheres; not sure if mathlib has these already). In general, proving this requires invariance of domain (which uses* e.g. the homology groups of spheres: these have been formalised in Lean 3, but re-doing this in mathlib would take significant time: see zulip for Brouwer's fixed point theorem).
(For smooth manifolds, the general case follows from the inverse function theorem. Yet, long-term that's not a proof I'd like to see in mathlib, as the result also holds in the topological category.)

<details>
<summary>* small detour</summary>
There are proofs of Brouwer/invariance of domain without using algebraic topology, but e.g. Sperner's lemma instead. (See e.g. [Tao's blog](https://terrytao.wordpress.com/2011/06/13/brouwers-fixed-point-and-invariance-of-domain-theorems-and-hilberts-fifth-problem/) or [this stackexchange answer](https://math.stackexchange.com/a/2316534) for some references. Sperner himself proved invariance of domain, for example. I would prefer not to spend my time on this. Homology groups of spheres are extremely basic algebraic topology; mathlib will need them eventually.)
</details>

**Short version**: I don't think you'd get a shorter proof.
I could add a lemma "being interior is independent of the chosen chart" just for dimension 1, with a TODO to improve the proof. I'm not sure if this would be useful. (In practice, everything you need to know about 1-manifolds and boundary is in this PR: you don't gain anything from the generality.)

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -403,6 +403,50 @@ instance IccChartedSpace (x y : ℝ) [h : Fact (x < y)] :
       simpa only [not_lt] using h'
   chart_mem_atlas z := by by_cases h' : (z : ℝ) < y <;> simp [h']
 
+@[simp]
+lemma iccManifold_chartAt {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} :
```
YaelDillies, line 407:
```suggestion
lemma IccManifold_chartAt {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} :
```
Same below

Actually, I'm not sure what the `Manifold` is referring to here.

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -403,6 +403,50 @@ instance IccChartedSpace (x y : ℝ) [h : Fact (x < y)] :
...
+    chartAt _ z = if z.val < y then IccLeftChart x y else IccRightChart x y := rfl
+
+lemma iccManifold_chartAt_of_le_top {x y : ℝ} [h : Fact (x < y)]
+    {z : Set.Icc x y} (h : z.val < y) : chartAt _ z = IccLeftChart x y := by
+  simp [iccManifold_chartAt, h]
+
+lemma iccManifold_chartAt_of_ge_top {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} (h : z.val ≥ y) :
```
YaelDillies, line 414:
```suggestion
lemma iccManifold_chartAt_of_top_le {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} (h : y \le z.val) :
```
Same below

Mathlib/Geometry/Manifold/Instances/Real.lean
```diff
@@ -403,6 +403,50 @@ instance IccChartedSpace (x y : ℝ) [h : Fact (x < y)] :
...
+  rw [ModelWithCorners.isBoundaryPoint_iff, extChartAt, iccManifold_chartAt_of_ge_top (by norm_num)]
+  exact IccRightChart_extend_top_mem_frontier
+
+lemma iccManifold_isInteriorPoint_interior {p : Set.Icc x y} (hp : x < p.val ∧ p.val < y) :
+    (𝓡∂ 1).IsInteriorPoint p := by
+  rw [ModelWithCorners.IsInteriorPoint, extChartAt, iccManifold_chartAt_of_le_top hp.2,
+    interior_range_modelWithCornersEuclideanHalfSpace]
+  exact IccLeftChart_extend_interior_pos hp
+
+lemma boundary_IccManifold : (𝓡∂ 1).boundary (Icc x y) = { ⊥, ⊤ } := by
```
YaelDillies, line 433:
```suggestion
lemma boundary_IccManifold : (𝓡∂ 1).boundary (Icc x y) = {⊥, ⊤} := by
```
We don't write `{ a }` for the singleton

YaelDillies:
I was expecting that the answer would be "no", but I didn't think homology of spheres would enter the picture 😄 Fine by me to leave the result as is

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: Good point; `Manifold` should be `ChartedSpace`. (I'm not sure if that's a great name - better suggestions welcome.)
Would `chartAt_iccChartedSpace` be better?

Just to check: why should it be UpperCamelCase? I would have thought `snake_case` (rules 1 and 5).

YaelDillies: So `Manifold`/`ChartedSpace` comes from `ChartedSpace.chartAt`? If so, the name should rather be
```suggestion
lemma Icc_chartedSpaceChartAt {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} :
```
or
```suggestion
lemma chartedSpaceChartAt_Icc {x y : ℝ} [h : Fact (x < y)] {z : Set.Icc x y} :
```
or some namespaced version thereof.

`Icc` (as well as the other intervals) is an exception to the rules.

Mathlib/Geometry/Manifold/Instances/Real.lean
Ruben-VandeVelde: Too bad it didn't get approved quickly :)

grunweg: Indeed; I've added a deprecation alias now.

Mathlib/Geometry/Manifold/Instances/Real.lean
grunweg: Thanks for explaining. I've gone with `Icc_chartedSpaceChartAt` and similar.

YaelDillies (approved):
Thanks! 🚀

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #13514, opened 2024-06-04
feat(Computability/ContextFreeGrammar): closure under union

madvorak:
- [x] depends on: #15895

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15895, opened 2024-08-16
feat(Computability/ContextFreeGrammar): mapping between two types of nonterminal symbols

madvorak:
---


madvorak:
This is the first half of #13514 as a standalone PR.

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,172 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+closure properties. When combining several grammars together, we usually want to take a sum type of
+their nonterminal types and embed respective nonterminals to this sum type. We subsequently show
+that the resulting grammar preserves derivations of those strings that may contain any terminals but
+only the proper nonterminals. The embedding operation must be injective. The projection operation
+must be injective on those symbols where it is defined. -/
+
+/-- Mapping `Symbol` when it is a nonterminal. -/
+def Symbol.map {N₀ N : Type*} (f : N₀ → N) : Symbol T N₀ → Symbol T N
+  | Symbol.terminal t => Symbol.terminal t
+  | Symbol.nonterminal n => Symbol.nonterminal (f n)
```
YaelDillies, line 269: Why not
```suggestion
/-- Mapping `Symbol` when it is a nonterminal. -/
def Symbol.map {N₀ N T₀ T : Type*} (f : T₀ → T) (g : N₀ → N) : Symbol T₀ N₀ → Symbol T N
  | Symbol.terminal t => Symbol.terminal (f t)
  | Symbol.nonterminal n => Symbol.nonterminal (g n)
```
This matches `Sum.map`

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,172 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+
+/-- Mapping `Symbol` when it is a nonterminal; may return `none`. -/
+def Symbol.filterMap {N₀ N : Type*} (f : N → Option N₀) : Symbol T N → Option (Symbol T N₀)
+  | Symbol.terminal t => some (Symbol.terminal t)
+  | Symbol.nonterminal n => Option.map Symbol.nonterminal (f n)
+
+/-- Mapping `ContextFreeRule` to a another nonterminal type. -/
```
YaelDillies, line 276:
```suggestion
/-- Map the type of nonterminal symbols of a `ContextFreeRule` . -/
```
and similar for other docstrings

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,172 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+def ContextFreeRule.map {N₀ N : Type*} (r : ContextFreeRule T N₀) (f : N₀ → N) :
+    ContextFreeRule T N :=
+  ⟨f r.input, r.output.map (Symbol.map f)⟩
+
+/-- A pair of `ContextFreeGrammar`s that, roughly speaking, work the same way. -/
```
YaelDillies, line 281: Could the docstring speak less roughly?

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: Yes but it would introduce additional useless argument.

YaelDillies: Are you sure you will never need to map the terminal symbols?

madvorak: If it ever happens, I will use a different mapping function that changes terminals only. Throughout the whole file, terminals and nonterminals are treated in "asymmetric" ways. Note that one of them is a type argument and the other is a field of any given grammar.

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: I added something; feel free to write something different.

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
 
 end closure_reversal
+
+section embed_project
+
+/-! This section contains only auxiliary constructions that will shorten upcoming proofs of
+closure properties. When combining several grammars together, we usually want to take a sum type of
+their nonterminal types and embed respective nonterminals to this sum type. We subsequently show
+that the resulting grammar preserves derivations of those strings that may contain any terminals but
+only the proper nonterminals. The embedding operation must be injective. The projection operation
```
YaelDillies, line 263: Using "terminal" and "nonterminal" in place of "terminal symbol" and "nonterminal symbol" is quite idiosyncratic and definitely hampers comprehension
```suggestion
that the resulting grammar preserves derivations of those strings that may contain any terminal symbols but
only the proper nonterminal symbols. The embedding operation must be injective. The projection operation
```
and below

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+def Symbol.filterMap {N₀ N : Type*} (f : N → Option N₀) : Symbol T N → Option (Symbol T N₀)
+  | Symbol.terminal t => some (Symbol.terminal t)
+  | Symbol.nonterminal n => Option.map Symbol.nonterminal (f n)
+
+/-- Map the type of nonterminal symbols of a `ContextFreeRule` . -/
+def ContextFreeRule.map {N₀ N : Type*} (r : ContextFreeRule T N₀) (f : N₀ → N) :
+    ContextFreeRule T N :=
+  ⟨f r.input, r.output.map (Symbol.map f)⟩
+
+/-- A pair of `ContextFreeGrammar`s with maps between their nonterminal types that work,
```
YaelDillies, line 281:
```suggestion
/-- A pair of `ContextFreeGrammar`s with maps between their types of nonterminal symbols that work,
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+structure EmbeddedContextFreeGrammar (T : Type uT) where
+  /-- The smaller grammar. -/
+  g₀ : ContextFreeGrammar.{uN} T
+  /-- The bigger grammar. -/
+  g : ContextFreeGrammar.{uN} T
+  /-- Mapping nonterminals from the smaller type to the bigger type. -/
+  embedNT : g₀.NT → g.NT
+  /-- Mapping nonterminals from the bigger type to the smaller type. -/
+  projectNT : g.NT → Option g₀.NT
+  /-- The former map is injective. -/
```
YaelDillies, line 292: You shouldn't rely on the order of the fields in docstrings
```suggestion
  /-- The function on nonterminal symbols is injective. -/
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  embed_inj : Function.Injective embedNT
+  /-- The latter map is injective where defined. -/
+  project_inj : ∀ x y, projectNT x = projectNT y → x = y ∨ projectNT x = none
```
YaelDillies, line 295: I think this will be more useful to you
```suggestion
  project_inj ⦃x y⦄ : projectNT x ≠ none → projectNT x = projectNT y → x = y
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  embed_mem_rules : ∀ r : ContextFreeRule T g₀.NT, r ∈ g₀.rules → r.map embedNT ∈ g.rules
+  /-- Each rule of the bigger grammar whose input nonterminal the smaller grammar recognizes
+  has a corresponding rule in the smaller grammar. -/
+  preimage_of_rules :
+    ∀ r : ContextFreeRule T g.NT,
+      r ∈ g.rules → ∀ n₀ : g₀.NT,
+        embedNT n₀ = r.input → ∃ r₀ ∈ g₀.rules, r₀.map embedNT = r
+
+lemma EmbeddedContextFreeGrammar.projectNT_inverse_embedNT (G : EmbeddedContextFreeGrammar T) :
+    ∀ x : G.g.NT, (∃ n₀, G.projectNT x = some n₀) → (Option.map G.embedNT (G.projectNT x) = x) := by
```
YaelDillies, line 308: Can you please eliminate the `∃`? In fact, why is the lemma not
```suggestion
lemma EmbeddedContextFreeGrammar.projectNT_inverse_embedNT (G : EmbeddedContextFreeGrammar T) {x n₀} :
    G.projectNT x = some n₀ → G.embedNT n₀ = x := by
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+/-- A `Symbol` is good iff it is one of those nonterminals that result from projecting or it is any
+terminal. -/
+def EmbeddedContextFreeGrammar.Good {G : EmbeddedContextFreeGrammar T} : Symbol T G.g.NT → Prop
+  | Symbol.terminal _ => True
+  | Symbol.nonterminal n => ∃ n₀ : G.g₀.NT, G.projectNT n = n₀
+
+/-- A string is good iff every `Symbol` in it is good. -/
+def EmbeddedContextFreeGrammar.GoodString {G : EmbeddedContextFreeGrammar T}
+    (s : List (Symbol T G.g.NT)) : Prop :=
+  ∀ a ∈ s, Good a
```
YaelDillies, line 347:
```suggestion
  ∀ ⦃a⦄, a ∈ s → Good a
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,176 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+
+lemma EmbeddedContextFreeGrammar.singletonGoodString {G : EmbeddedContextFreeGrammar T}
```
YaelDillies, line 349:
```suggestion
lemma EmbeddedContextFreeGrammar.goodString_singleton {G : EmbeddedContextFreeGrammar T}
```

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: Both are functions non nonterminal symbols.

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: Are you sure that you want to use "terminal symbols" and "nonterminal symbols" only?

In the ITP paper that documents the Lean 3 library, I was naming it as follows:
let `(n : g.NT)`
`n` .. nonterminal
`Symbol.nonterminal n` ... nonterminal symbol

Mathlib/Computability/ContextFreeGrammar.lean
YaelDillies:
```suggestion
  /-- The embedding map is injective. -/
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,178 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- The former map is injective. -/
+  embed_inj : Function.Injective embedNT
+  /-- The latter map is injective where defined. -/
+  project_inj : ∀ x y : g.NT, projectNT x = projectNT y → x = y ∨ projectNT x = none
+  /-- The two mappings are essentially inverses. -/
+  projectNT_embedNT : ∀ n₀ : g₀.NT, projectNT (embedNT n₀) = some n₀
+  /-- Each rule of the smaller grammar has a corresponding rule in the bigger grammar. -/
+  embed_mem_rules : ∀ r : ContextFreeRule T g₀.NT, r ∈ g₀.rules → r.map embedNT ∈ g.rules
+  /-- Each rule of the bigger grammar whose input nonterminal the smaller grammar recognizes
+  has a corresponding rule in the smaller grammar. -/
```
YaelDillies, line 302: I don't understand nonterminal this sentence

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,178 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+    (hn : G.projectNT n = some n₀) :
+    Option.map G.embedNT (G.projectNT n) = n := by
+  rw [hn, Option.map_some']
+  apply congr_arg
+  by_contra hnx
+  cases (G.projectNT_embedNT n₀ ▸ G.project_inj n (G.embedNT n₀)) hn with
+  | inl case_valu => exact hnx case_valu.symm
+  | inr case_none => exact Option.noConfusion (hn ▸ case_none)
+
+/-- Production by `G.g₀` can be mirrored by `G.g` production. -/
```
YaelDillies, line 320:
```suggestion
/-- Production by `G.g₀` can be mirrored by production by `G.g`. -/
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,178 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  rcases hG with ⟨r, rin, hr⟩
+  rcases hr.exists_parts with ⟨u, v, bef, aft⟩
+  refine ⟨r.map G.embedNT, G.embed_mem_rules r rin, ?_⟩
+  rw [ContextFreeRule.rewrites_iff]
+  use u.map (Symbol.map G.embedNT), v.map (Symbol.map G.embedNT)
+  constructor
+  · simpa only [List.map_append] using congr_arg (List.map (Symbol.map G.embedNT)) bef
+  · simpa only [List.map_append] using congr_arg (List.map (Symbol.map G.embedNT)) aft
+
+/-- Derivation by `G.g₀` can be mirrored by `G.g` derivation. -/
```
YaelDillies, line 333: Same here

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,178 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+    G.g.Derives (w₁.map (Symbol.map G.embedNT)) (w₂.map (Symbol.map G.embedNT)) := by
+  induction hG with
+  | refl => rfl
+  | tail _ orig ih => exact ih.trans_produces (produces_map orig)
+
+/-- A `Symbol` is good iff it is one of those nonterminals that result from projecting or it is any
+terminal. -/
+def EmbeddedContextFreeGrammar.Good : Symbol T G.g.NT → Prop
+  | Symbol.terminal _ => True
+  | Symbol.nonterminal n => ∃ n₀ : G.g₀.NT, G.projectNT n = n₀
```
YaelDillies, line 345:
```suggestion
inductive EmbeddedContextFreeGrammar.Good : Symbol T G.g.NT → Prop
  | terminal (t : T) : Good (.terminal t)
  | nonterminal (n : G.g.NT) (n₀ : G.g₀.NT) (hn : G.projectNT n = n₀) : Good (.nonterminal n)
```
will probably be nicer to work with

madvorak:
There is a bug but my VS Code doesn't show any error. What's more, the Infoview is updated on every line as if it was a correct proof. How is it possible?

YaelDillies:
I could reproduce locally the behavior you describe: No error shown. However, your version of mathlib was ancient in this PR, so I merged master and the error appeared. The lack of error locally was probably due to the new variable inclusion mechanism not being available yet on this branch, meaning that you were referring some far out `hn₀` (instead of receiving an error, since your hypothesis is called `hn`).

madvorak:
> I could reproduce locally the behavior you describe: No error shown. However, your version of mathlib was ancient in this PR, so I merged master and the error appeared. The lack of error locally was probably due to the new variable inclusion mechanism not being available yet on this branch, meaning that you were referring some far out `hn₀` (instead of receiving an error, since your hypothesis is called `hn`).

Thank you very much! I will give it a try now.

madvorak:
It seems the only remaining problem was that `List.filterMap` is no longer a `@[simp]` lemma.

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,179 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+
+/-! This section contains only auxiliary constructions that will shorten upcoming proofs of
+closure properties. When combining several grammars together, we usually want to take a sum type of
+their nonterminal types and embed respective nonterminals to this sum type.
+We subsequently show that the resulting grammar preserves derivations of those strings that may
+contain any terminal symbols but only the proper nonterminal symbols.
+The embedding operation must be injective.
+The projection operation must be injective on those symbols where it is defined. -/
+
+/-- Mapping `Symbol` when it is a nonterminal. -/
```
YaelDillies, line 267:
```suggestion
/-- Mapping `Symbol` when it is a nonterminal symbol. -/
```
and below

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,179 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+def Symbol.map {N₀ N : Type*} (f : N₀ → N) : Symbol T N₀ → Symbol T N
+  | Symbol.terminal t => Symbol.terminal t
+  | Symbol.nonterminal n => Symbol.nonterminal (f n)
```
YaelDillies, line 270: I think anonymous dot notation is really useful here to unclutter
```suggestion
  | .terminal t => .terminal t
  | .nonterminal n => .nonterminal (f n)
```
and below

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,179 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- Each rule of the bigger grammar whose input nonterminal is recognized by the smaller grammar
+  has a corresponding rule in the smaller grammar. -/
+  preimage_of_rules :
+    ∀ r : ContextFreeRule T g.NT,
+      r ∈ g.rules → ∀ n₀ : g₀.NT,
+        embedNT n₀ = r.input → ∃ r₀ ∈ g₀.rules, r₀.map embedNT = r
+
+variable {G : EmbeddedContextFreeGrammar T}
+
+lemma EmbeddedContextFreeGrammar.projectNT_inverse_embedNT {n : G.g.NT} {n₀ : G.g₀.NT}
```
YaelDillies, line 310: I would suggest
```suggestion
namespace EmbeddedContextFreeGrammar
variable {G : EmbeddedContextFreeGrammar T}

lemma projectNT_inverse_embedNT {n : G.g.NT} {n₀ : G.g₀.NT}
```

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,179 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  | Symbol.terminal t => some (Symbol.terminal t)
+  | Symbol.nonterminal n => Option.map Symbol.nonterminal (f n)
+
+/-- Map the type of nonterminal symbols of a `ContextFreeRule` . -/
+def ContextFreeRule.map {N₀ N : Type*} (r : ContextFreeRule T N₀) (f : N₀ → N) :
+    ContextFreeRule T N :=
+  ⟨f r.input, r.output.map (Symbol.map f)⟩
+
+/-- A pair of `ContextFreeGrammar`s with maps between their types of nonterminal symbols that work,
+roughly speaking, in a good way. -/
```
YaelDillies, line 283: This is still really uninformative. Please improve.

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,179 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- The two mappings are essentially inverses. -/
+  projectNT_embedNT : ∀ n₀ : g₀.NT, projectNT (embedNT n₀) = some n₀
+  /-- Each rule of the smaller grammar has a corresponding rule in the bigger grammar. -/
+  embed_mem_rules : ∀ r : ContextFreeRule T g₀.NT, r ∈ g₀.rules → r.map embedNT ∈ g.rules
+  /-- Each rule of the bigger grammar whose input nonterminal is recognized by the smaller grammar
+  has a corresponding rule in the smaller grammar. -/
+  preimage_of_rules :
+    ∀ r : ContextFreeRule T g.NT,
+      r ∈ g.rules → ∀ n₀ : g₀.NT,
+        embedNT n₀ = r.input → ∃ r₀ ∈ g₀.rules, r₀.map embedNT = r
```
YaelDillies, line 306: Do you have a reference for this?

madvorak: I have a reference for a similar structure.

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: Please reply to my overall comment about "nonterminal" vs "nonterminal symbol".

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: I don't know how to describe it in words without repeating the entire definition informally in the docstring.

Mathlib/Computability/ContextFreeGrammar.lean
madvorak: https://arxiv.org/pdf/2302.06420 page 6

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,181 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- The smaller grammar. -/
+  g₀ : ContextFreeGrammar.{uN} T
+  /-- The bigger grammar. -/
+  g : ContextFreeGrammar.{uN} T
+  /-- Mapping nonterminals from the smaller type to the bigger type. -/
+  embedNT : g₀.NT → g.NT
+  /-- Mapping nonterminals from the bigger type to the smaller type. -/
+  projectNT : g.NT → Option g₀.NT
+  /-- The embedding map is injective. -/
+  embed_inj : Function.Injective embedNT
```
YaelDillies, line 294: You don't need to assume this: `projectNT_embedNT` already implies it

Mathlib/Computability/ContextFreeGrammar.lean
```diff
@@ -253,3 +253,181 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- The projecting map is injective where defined. -/
+  project_inj : ∀ x y : g.NT, projectNT x = projectNT y → x = y ∨ projectNT x = none
+  /-- The two mappings are essentially inverses. -/
+  projectNT_embedNT : ∀ n₀ : g₀.NT, projectNT (embedNT n₀) = some n₀
```
YaelDillies, line 298: I believe you don't need `project_inj` at all. Here's my reasoning:
1. All you care about is that `projectNT` is a one-sided inverse to `embedNT`.
2. This already implies that `projectNT` is injective on the range of `embedNT`.
3. Hence all `project_inj` is giving you is that `projectNT` is also injective outside of the range of `embedNT`.
4. But `projectNT`might as well always be `none` outside of the range of `embedNT` (you could even enforce that if you wanted).
5. Therefore you don't need `project_inj`.

Mathlib/Computability/ContextFreeGrammar.lean
YaelDillies:
```suggestion
/-- An embedding from a context-free grammar `g₀` to a context-free grammar `g` is an embedding
`embedNT` of the nonterminal symbols from the former to the latter along with a one-sided inverse
`projectNT` such that all rewrite rules of `g` that end in `embedNT n₀` for some `n₀` come from some
rule in `g₀`. -/```

File: Mathlib/Computability/ContextFreeGrammar.lean [hunk 22]
```diff
@@ -253,3 +253,181 @@ theorem Language.IsContextFree.reverse (L : Language T) :
...
+  /-- Each rule of the smaller grammar has a corresponding rule in the bigger grammar. -/
+  embed_mem_rules : ∀ r : ContextFreeRule T g₀.NT, r ∈ g₀.rules → r.map embedNT ∈ g.rules
+  /-- Each rule of the bigger grammar whose input nonterminal is recognized by the smaller grammar
+  has a corresponding rule in the smaller grammar. -/
+  preimage_of_rules :
+    ∀ r : ContextFreeRule T g.NT,
+      r ∈ g.rules → ∀ n₀ : g₀.NT,
+        embedNT n₀ = r.input → ∃ r₀ ∈ g₀.rules, r₀.map embedNT = r
```
YaelDillies, line 306: Something like this
```suggestion
  map_embedNT_surjOn :
    Set.SurjOn (Symbol.map embedNT) {r | r ∈ g.rules ∧ ∃ n₀, embedNT = n₀ = r.input}
      {r₀ | r₀ ∈ g₀.rules}
```

YaelDillies (commented):
I have now read your paper with Jasmin. Here are a few comments in consequence

madvorak:
What is wrong here, please?
```
invalid occurrence of universe level 'u_2' at 'ContextFreeGrammar.Embedding.GoodString', it does not occur at the declaration type, nor it is explicit universe level provided by the user, occurring at expression
  Good.{uT, u_2, u_1} a
at declaration body
  fun {T : Type uT} {g : ContextFreeGrammar T} (s : List (Symbol T g.NT)) ↦ ∀ ⦃a : Symbol T g.NT⦄, a ∈ s → Good a
```

PR #15896, opened 2024-08-16
[Merged by Bors] - feat: the lambdaSyntax linter

grunweg:
The mathlib style guide states clearly that using \lambda for anonymous functions is disallowed, and the fun keyword is preferred. This linter enforces this.

---

I have decided to call the linter option `linter.style.lambdaSyntax`, fareshadowing the analogous renaming of options in #15971.
This linter is actually pretty simple: most of this PR are detailed tests.

### DIFF
--- a/Mathlib/Tactic/Linter/Lint.lean
+++ b/Mathlib/Tactic/Linter/Lint.lean
@@ -30,7 +30,7 @@ Linter that checks whether a structure should be in Prop.
     -- remark: using `Lean.Meta.isProp` doesn't suffice here, because it doesn't (always?)
     -- recognize predicates as propositional.
     let isProp ← forallTelescopeReducing (← inferType (← mkConstWithLevelParams declName))
-      fun _ ty => return ty == .sort .zero
+      fun _ ty ↦ return ty == .sort .zero
     if isProp then return none
     let projs := (getStructureInfo? (← getEnv) declName).get!.fieldNames
     if projs.isEmpty then return none -- don't flag empty structures
@@ -87,18 +87,18 @@ def getIds : Syntax → Array Syntax
   | .node _ `Batteries.Tactic.Alias.alias args => args[2:3]
   | .node _ ``Lean.Parser.Command.export args => (args[3:4] : Array Syntax).map (·[0])
   | stx@(.node _ _ args) =>
-    ((args.attach.map fun ⟨a, _⟩ => getIds a).foldl (· ++ ·) #[stx]).filter (·.getKind == ``declId)
+    ((args.attach.map fun ⟨a, _⟩ ↦ getIds a).foldl (· ++ ·) #[stx]).filter (·.getKind == ``declId)
   | _ => default
 
 @[inherit_doc linter.dupNamespace]
-def dupNamespace : Linter where run := withSetOptionIn fun stx => do
+def dupNamespace : Linter where run := withSetOptionIn fun stx ↦ do
   if Linter.getLinterValue linter.dupNamespace (← getOptions) then
     match getIds stx with
       | #[id] =>
         let ns := (← getScope).currNamespace
         let declName := ns ++ (if id.getKind == ``declId then id[0].getId else id.getId)
         let nm := declName.components
-        let some (dup, _) := nm.zip (nm.tailD []) |>.find? fun (x, y) => x == y
+        let some (dup, _) := nm.zip (nm.tailD []) |>.find? fun (x, y) ↦ x == y
           | return
         Linter.logLint linter.dupNamespace id
           m!"The namespace '{dup}' is duplicated in the declaration '{declName}'"
@@ -182,7 +182,7 @@ def findCDot : Syntax → Array Syntax
   | stx@(.node _ kind args) =>
     let dargs := (args.map findCDot).flatten
     match kind with
-      | ``Lean.Parser.Term.cdot | ``cdotTk=> dargs.push stx
+      | ``Lean.Parser.Term.cdot | ``cdotTk => dargs.push stx
       | _ =>  dargs
   |_ => #[]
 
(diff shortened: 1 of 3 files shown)


mattrobball:
How about test with lambda used in other kinds of syntax nodes?

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,51 @@ initialize addLinter cdotLinter
...
+def lambdaSyntaxLinter : Linter where run := withSetOptionIn fun stx ↦ do
+    unless Linter.getLinterValue linter.lambdaSyntax (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    for s in findLambdaSyntax stx do
+      if let .atom _ "λ" := s.getArgs[0]! then
+        Linter.logLint linter.lambdaSyntax s m!"\
+        Please use 'fun' and not 'λ' to define anonymous functions.\n\
+        The latter syntax has been deprecated in mathlib 4."
```
eric-wieser, line 253: Note that this approach does not allow:
```
example : ℕ → ℕ := set_option linter.lambdaSyntax false in λ _ ↦ 0
```
which would require walking the infotree instead.

grunweg: Good point! Not supporting this seems fine to me, but I have added a test to this effect.

grunweg:
> How about test with lambda used in other kinds of syntax nodes?

I just added a lot of tests - which cannot be exhaustive; let me know if a particular kind of tests is missing!

grunweg:
Rebased on master, and disabled this linter by default.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -247,6 +247,51 @@ initialize addLinter dollarSyntaxLinter
...
+namespace Style.lambdaSyntax
+
+/--
+`findLambdaSyntax stx` extracts from `stx` all syntax nodes of `kind` `Term.fun`. -/
+partial
+def findLambdaSyntax : Syntax → Array Syntax
+  | stx@(.node _ kind args) =>
+    let dargs := (args.map findLambdaSyntax).flatten
+    match kind with
+      | ``Parser.Term.fun => dargs.push stx
```
adomani, line 275:
```suggestion
      | ``Parser.Term.fun => dargs.push stx[0]
```
If from here you extract the `atom` directly...

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -247,6 +247,51 @@ initialize addLinter dollarSyntaxLinter
...
+
+@[inherit_doc linter.style.lambdaSyntax]
+def lambdaSyntaxLinter : Linter where run := withSetOptionIn fun stx ↦ do
+    unless Linter.getLinterValue linter.style.lambdaSyntax (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    for s in findLambdaSyntax stx do
+      if let .atom _ "λ" := s.getArgs[0]! then
+        Linter.logLint linter.style.lambdaSyntax s m!"\
```
adomani, line 287:
```suggestion
      if let .atom _ "λ" := s then
        Linter.logLint linter.style.lambdaSyntax s m!"\
```
... then here you can simply log the warning on the `λ` more easily.

adomani (approved):
Just a very small stylistic suggestion: with the change, just `λ` should be underlined, with the current implementation, the whole `λ x => ...` gets underlined.

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -247,6 +247,51 @@ initialize addLinter dollarSyntaxLinter
...
+# The `lambdaSyntax` linter
+
+The `lambdaSyntax` linter is a syntax linter that flags uses of the symbol `λ` to define anonymous
+functions, as opposed to the `fun` keyword. These are syntactically equivalent; mathlib style
+prefers the latter as it is considered more readable.
+-/
+
+/--
+The `lambdaSyntax` linter flags uses of the symbol `λ` to define anonymous functions.
+This is syntactically equivalent to the `fun` keyword; mathlib style prefers using the latter.-/
```
adomani, line 260:
```suggestion
This is syntactically equivalent to the `fun` keyword; mathlib style prefers using the latter.
-/
```

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -247,6 +247,51 @@ initialize addLinter dollarSyntaxLinter
...
+def lambdaSyntaxLinter : Linter where run := withSetOptionIn fun stx ↦ do
+    unless Linter.getLinterValue linter.style.lambdaSyntax (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    for s in findLambdaSyntax stx do
+      if let .atom _ "λ" := s.getArgs[0]! then
+        Linter.logLint linter.style.lambdaSyntax s m!"\
+        Please use 'fun' and not 'λ' to define anonymous functions.\n\
+        The latter syntax has been deprecated in mathlib 4."
```
adomani, line 289:
```suggestion
        The 'λ'-syntax has been deprecated in mathlib4."
```
Just to be more explicit.

adomani:
I just noticed that the changes in mathlib are `=>` vs `\mapsto`, though this is not what the linter enforces.  This should not necessarily change, but seems a little odd to modify code that the linter does not flag.

grunweg:
Thanks for the suggestions, Damiano! I implemented all of them (tweaking the `stx` one slightly), and updated the tests.

jcommelin (commented):
bors d+

grunweg:
bors merge

PR #15897, opened 2024-08-16
[Merged by Bors] - chore: use `Batteries` test driver directly in the lake file

mattrobball:
With leanprover/lean4#4261, we can specify a test driver from a dependency. We do this with `Batteries` test driver. Previously, the test driver from `Batteries` was called via `scripts/test.lean` which is now removed.

---


mattrobball:
@semorrison I was looking at `lake test` for my class and noticed this todo. Local testing seems good but you would know better. Let me know if anything is amiss.

adomani:
I don't really know how to review this, since I do not know the implications of the changes!  However, I can see that CI was successful and that the tests have been run, so this looks good to me!  :smile:

adomani (approved):

bryangingechen:
@semorrison Could you take a look at this when you have a spare moment?

PR #15898, opened 2024-08-16
lean CI: automatic labeling of PRs

adomani:
---


grunweg:
I was curious about your Lean rewrite of the autolabel script, so I took a look already. I hope you don't mind! (Let me know if you prefer to just hack on it instead.)

Your new script looks really shiny and elaborate. In my mind, I was thinking of something much simpler (basically, a direct port as much as possible), hence I wonder if all this complexity is necessary. (It is really fun to *write* this, I agree --- but as code is read more often, I care about future readers more...) As far as I can see, the drivers of that complexity are
(1) adding exclusions (to handle Tactic.Lint getting labelled as t-linter, but Tactic.Simp as t-meta)
(2) ensuring "Algebra/File.lean" does not get labelled "t-algebraic-geometry"
(Can you handle this by using the `Name` API more? I presume it has all the logic with prefixes already...)
(3) allowing for multi-level inclusions in a general way
(4) separating the core logic and the mapping, which implies an environment extension, macros to manipulate this list etc.

The shell script dealt with these by
- performing one preprocessing step to handle "Tactic/Linter", and
- doing a simple matching of root directory to label name
- disallowing multi-part directories, other than that
- defining both the mapping and the logic in the same script.

I wonder if the new script's complexity is already necessary. (That said, reading this in detail will be a good way for me to learn more about Lean programming.)

I do like the self-tests, these are great. Another thought I had was validating the script's configuration. Because it's all in Lean, you can check for typos (e.g. "." vs "/"), warn if the label rules and the root directories on disk do not match (on both directions), etc. this *is* much nicer than in shell, and definitely useful.

adomani:
This PR contains both the shell script and the lean one: I am closing this, since there are two separate PRs now, one with the lean script (#15938) and one with the shell script (#15848, probably closed in favour of the Lean one).

PR #15899, opened 2024-08-16
[Merged by Bors] - refactor: downgrade `linear_combination`

hrmacbeth:
This PR drops support from several operations from the `linear_combination` tactic:
- multiplication of two hypotheses
- division by a hypothesis
- inversion of a hypothesis
- addition or subtraction of a constant and a hypothesis

The first three operations are removed in order to make `linear_combination` perform only genuinely linear operations; this makes the tactic more predictable and also allows for more flexibility in refactors/reimplementations. The last operation is removed because, after of the removal of the other three, it is redundant (adding a constant to both sides of an equation doesn't change the difference of the two sides).

The old operations are still available using the `congr(...)` term elaborator or using the legacy `linear_combination'` tactic.

A side effect that be seen in `Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean` is that elaboration changes slightly; previously `⁻¹` was one of the supported operations *on hypotheses*, so `⁻¹` got handled differently *in constants* too.  In that example, previously the atom which got pre-elaborated was `C'.u`; now the atom which gets pre-elaborated is `C'.u⁻¹`.

Discussion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Narrowing.20the.20scope.20of.20.60linear_combination.60)

---


Mathlib/Tactic/LinearCombination.lean
```diff
@@ -191,17 +169,6 @@ Note: The left and right sides of all the equalities should have the same
 * `linear_combination (exp := n) e` will take the goal to the `n`th power before subtracting the
   combination `e`. In other words, if the goal is `t1 = t2`, `linear_combination (exp := n) e`
   will change the goal to `(t1 - t2)^n = 0` before proceeding as above.
-  This feature is not supported for `linear_combination2`.
-* `linear_combination2 e` is the same as `linear_combination e` but it produces two
-  subgoals instead of one: rather than proving that `(a - b) - (a' - b') = 0` where
-  `a' = b'` is the linear combination from `e` and `a = b` is the goal,
-  it instead attempts to prove `a = a'` and `b = b'`.
-  Because it does not use subtraction, this form is applicable also to semirings.
```
eric-wieser, line 199: Losing support for semirings feels like a bit of a shame; would it make sense to combine the syntaxes, but fallback to this approach internally if no ring can be found?

hrmacbeth: I really think the tactic is *improved* by removing this from the scope.  I laid out my thoughts here
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Narrowing.20the.20scope.20of.20.60linear_combination.60/near/462845544
(and let's continue the discussion there rather than here), but for the record, let me write precisely what I think that scope should be:

`linear_combination` is a tactic which reduces problems of the form
```
h₁ : a₁ = b₁
h₂ : a₂ = b₂
h₃ : a₃ = b₃
⊢ A = B
```
to the task of showing that 
```
(f a₁ a₂ a₃ - f b₁ b₂ b₃) - (A - B)
```
equals zero (and then tries a standard normalization procedure to check that).  I believe the scope of the tactic should be limited to operations `f` for which this is the same thing as
```
f (a₁ - b₁) (a₂ - b₂) (a₃ - b₃) - (A - B)
```
Namely, **linear** operations `f`.

urkud:
Please describe the change of scope in the commit message (e.g., by copy+pasting relevant paragraphs from the Zulip thread).

hrmacbeth:
This is something that won't be settled until #16103 is decided about.

Mathlib/Tactic/LinearCombination.lean
hrmacbeth: To summarize where we landed after the Zulip discussion and some follow-up PRs:
- the PR #16103 added *native* support for cancellative semirings in `linear_combination`;
- the PR #16166 preserved (near-) backward compatibility (in particular the original syntax and the rudimentary support for non-cancellative semirings) by duplicating `linear_combination` to a tactic `linear_combination'` which will not be further modified.

urkud: I don't think that removing support for `Nat` makes the tactic better. I would prefer the tactic to swap LHS and RHS as needed to avoid `-`.

hrmacbeth: @urkud I'm not sure what you mean here -- as of #16103 Nat is supported natively (and `-` *means* swapping).  Can you say a bit more about your concern?

urkud: I wrote my comment before the description of the scope change was added. Please ignore it.

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -69,14 +61,15 @@ partial def expandLinearCombo (ty : Expr) (stx : Syntax.Term) : TermElabM Expand
   | `($e₁ + $e₂) => do
     match ← expandLinearCombo ty e₁, ← expandLinearCombo ty e₂ with
     | .const c₁, .const c₂ => .const <$> ``($c₁ + $c₂)
-    | .proof p₁, .const c₂ => .proof <$> ``(pf_add_c $p₁ $c₂)
-    | .const c₁, .proof p₂ => .proof <$> ``(c_add_pf $p₂ $c₁)
     | .proof p₁, .proof p₂ => .proof <$> ``(add_pf $p₁ $p₂)
+    | _ , _ => throwError "'linear_combination' is agnostic to the addition of constants"
```
eric-wieser, line 65: nit: I'd argue that if it were agnostic to them, this would be a warning not an error. Throwing an error doesn't sound agnostic to me!

hrmacbeth: OK, let me know if you have alternative wording that still accomplishes the goal of dropping this functionality!

hrmacbeth:
I believe the Zulip discussion reached consensus in favour of this change, so I will remove that tag.

Mathlib/Tactic/LinearCombination.lean
eric-wieser: Either
```suggestion
    | _ , _ => throwError "'linear_combination' does not support addition of constants as it would have no effect"
```
or
```suggestion
    | .proof p, .const _ | .const _, .proof p =>
      logWarning "'linear_combination' is agnostic to the addition of constants, so this has no effect"
      pure (.proof p)
```
where the second one allows users to ignore the warning

eric-wieser:
Just to check: there is no docstring change here because the feature was never documented in the first place? Or was the docstring pre-emptively updated in a previous PR?

hrmacbeth:
> Just to check: there is no docstring change here because the feature was never documented in the first place? Or was the docstring pre-emptively updated in a previous PR?

Right, the docstring dates from the Lean 3 tactic, whereas the features were added in the port.

Mathlib/Tactic/LinearCombination.lean
hrmacbeth: I like the second one and have implemented it.  Let's revisit this if needed when the other planned refactors come.

Mathlib/Tactic/LinearCombination.lean
```diff
@@ -87,17 +87,12 @@ partial def expandLinearCombo (ty : Expr) (stx : Syntax.Term) : TermElabM Expand
     | .const c₁, .const c₂ => .const <$> ``($c₁ * $c₂)
     | .proof p₁, .const c₂ => .proof <$> ``(pf_mul_c $p₁ $c₂)
     | .const c₁, .proof p₂ => .proof <$> ``(c_mul_pf $p₂ $c₁)
-    | .proof p₁, .proof p₂ => .proof <$> ``(mul_pf $p₁ $p₂)
-  | `($e⁻¹) => do
-    match ← expandLinearCombo ty e with
-    | .const c => .const <$> `($c⁻¹)
-    | .proof p => .proof <$> ``(inv_pf $p)
+    | .proof _, .proof _ => throwError "'linear_combination' supports only linear operations"
```
eric-wieser, line 90: Can you add a test for this?

joneugster (changes requested):
I think it would be reasonable if you updated the docstrings of `linear_combination` and `linear_combination'`:

> `both`: If the `normalize` field of the configuration is set to false

I think this should be changed to using `(norm := skip)`, right? (that's not directly relevant here, but I tripped over it when looking at the one situation above you used `linear_combination'`)

> `linear_combination'`: Note: There is also a similar tactic `linear_combination` (no prime); this version is provided for backward compatibility.

At least mention the difference implemented in this PR.

Mathlib/Tactic/LinearCombination.lean
joneugster: I have to say, debugging the example above I found this message not immensely helpful. Is it possible that the error message mentions the term it is upset about?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
```diff
@@ -498,7 +499,7 @@ lemma negY_of_Z_ne_zero {P : Fin 3 → F} (hPz : P z ≠ 0) :
 lemma Y_sub_Y_mul_Y_sub_negY {P Q : Fin 3 → R} (hP : W'.Equation P) (hQ : W'.Equation Q)
     (hx : P x * Q z ^ 2 = Q x * P z ^ 2) :
     (P y * Q z ^ 3 - Q y * P z ^ 3) * (P y * Q z ^ 3 - W'.negY Q * P z ^ 3) = 0 := by
-  linear_combination (norm := (rw [negY]; ring1)) Q z ^ 6 * (equation_iff P).mp hP
+  linear_combination' (norm := (rw [negY]; ring1)) Q z ^ 6 * (equation_iff P).mp hP
     - P z ^ 6 * (equation_iff Q).mp hQ + hx * hx * hx + W'.a₂ * P z ^ 2 * Q z ^ 2 * hx * hx
     + (W'.a₄ * P z ^ 4 * Q z ^ 4 - W'.a₁ * P y * P z * Q z ^ 4) * hx
```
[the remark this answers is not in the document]
hrmacbeth: The most direct translation would be to use the `congr(...)` term elaborator:
```lean
  linear_combination' (norm := (rw [negY]; ring1)) Q z ^ 6 * (equation_iff P).mp hP
    - P z ^ 6 * (equation_iff Q).mp hQ + congr($hx ^ 3) + W'.a₂ * P z ^ 2 * Q z ^ 2 * congr($hx ^ 2)
    + (W'.a₄ * P z ^ 4 * Q z ^ 4 - W'.a₁ * P y * P z * Q z ^ 4) * hx
```
But in general, a "nonlinear" linear combination has a code smell to me.  For example, I would find it more elegant to solve this example in this way, using a purely "linear" linear combination and avoiding the nonstandard discharger:
```lean
  rw [equation_iff, negY] at *
  linear_combination Q z ^ 6 * hP - P z ^ 6 * hQ
    + (Q z ^ 4 * P z ^ 4 * W'.a₄ - P y * Q z ^ 4 * P z * W'.a₁ 
      + Q z ^ 2 *  P z ^ 2 * (P z ^ 2 * Q x + Q z ^ 2 * P x) * W'.a₂ 
      + (P z ^ 4 * Q x ^ 2 + Q z ^ 2 * P z ^ 2 * Q x * P x + Q z ^ 4 * P x ^ 2)) * hx
```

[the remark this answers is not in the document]
hrmacbeth: In any case, did you write these variants to check your understanding of the "story" of this PR, or are they intended as genuine suggested changes?  The approach I have taken on these two examples which are hit by the downgrade -- switching them over to `linear_combination'`, to serve as a record of the scope of functionality which has been removed --  was requested by @digama0:
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Narrowing.20the.20scope.20of.20.60linear_combination.60/near/465256620

Mathlib/Tactic/LinearCombination.lean
hrmacbeth: I added tests, and I localized the error to the offending `*` or `/` symbol -- what do you think?

hrmacbeth:
> I think it would be reasonable if you updated the docstrings of `linear_combination` and `linear_combination'`:
>
> > `linear_combination'`: Note: There is also a similar tactic `linear_combination` (no prime); this version is provided for backward compatibility.
> 
> At least mention the difference implemented in this PR.

I have done this (98579c8d49852d5ae861b3f892635ecb1847e8a5).  Just flagging that I don't necessarily think that it is sustainable in the long term to continue to document the differences between the two: in the future I hope to extend the scope of `linear_combination` quite substantially.

Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean
```diff
@@ -245,18 +245,18 @@ lemma variableChange_comp (C C' : VariableChange R) (W : WeierstrassCurve R) :
...
   ext <;> simp only [mul_inv, Units.val_mul]
-  · linear_combination (norm := ring1) C.u⁻¹ * C.s * 2 * C'.u.inv_mul
+  · linear_combination (norm := ring1) ↑C.u⁻¹ * C.s * 2 * C'.u.inv_mul
   · linear_combination (norm := ring1)
-      C.s * (-C'.s * 2 - W.a₁) * C.u⁻¹ ^ 2 * C'.u⁻¹ * C'.u.inv_mul
+      C.s * (-C'.s * 2 - W.a₁) * C.u⁻¹ ^ 2 * ↑C'.u⁻¹ * C'.u.inv_mul
         + (C.r * 3 - C.s ^ 2) * C.u⁻¹ ^ 2 * pow_mul_pow_eq_one 2 C'.u.inv_mul
   · linear_combination (norm := ring1)
-      C.r * (C'.s * 2 + W.a₁) * C.u⁻¹ ^ 3 * C'.u⁻¹ * pow_mul_pow_eq_one 2 C'.u.inv_mul
+      C.r * (C'.s * 2 + W.a₁) * C.u⁻¹ ^ 3 * ↑C'.u⁻¹ * pow_mul_pow_eq_one 2 C'.u.inv_mul
```
eric-wieser, line 253: Why did this end up changing?

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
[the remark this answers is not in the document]
joneugster: Thanks for the explanation, your snippet is indeed much more elegant!

So my suggestion is to just replace these two uses of `linear_combination'` with the solution you just wrote out, since thats nicer than the original anyways.

But also, I think the PR can be merged regardless if you do that or not :)

Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean
joneugster: I think I looked at this earlier and from top of my head:

Previously the tactic would parse the `\^-1` before the term `C.u` was parsed (and translate it into using one of the theorems that got removed) but now it's actually the normal term parser parsing the entire `C.u\^-1`.

Therefore, now it thinks the inverse should be taken in `R` instead of in `R\^x` without the coersion hint (which is btw also displayed in the goal). This seems like normal Lean behaviour so the coersion hints are in my opinion expected there.

joneugster:
Thank you heather for addressing the comments! I'm conviced now that this is a positive change overall, and I'm looking forward to see more changes to the new `linear_combination`

Maybe Eric can "maintainer merge" once happy :)

Mathlib/AlgebraicGeometry/EllipticCurve/Jacobian.lean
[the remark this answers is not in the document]
hrmacbeth: If you don't mind either way, then let's go with the status quo, since Mario preferred it!

Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean
hrmacbeth: Previously `⁻¹` was one of the supported operations *on hypotheses*, so `⁻¹` got handled differently *in constants* too: the atom which got pre-elaborated was `C'.u`.  Now the atom which gets pre-elaborated is `C'.u⁻¹`.  (Though I don't actually see why it makes a difference here!)

joneugster (approved):

Mathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean
eric-wieser: Thanks, I've copied @hrmacbeth's version into the PR description.

joneugster:
I think since @eric-wieser marked the last conversation as resolved, I can suggest

maintainer merge

Thanks for the PR!

semorrison:
bors merge

PR #15901, opened 2024-08-16
[Merged by Bors] - feat(ModelTheory): More general ways of deriving `DecidableEq` for logic symbols

awainverse:
Replaces two `DecidableEq` instances in `ModelTheory/Ring/Basic` with more general instances.

---

### DIFF
--- a/Mathlib/ModelTheory/Basic.lean
+++ b/Mathlib/ModelTheory/Basic.lean
@@ -250,9 +250,15 @@ theorem card_mk₂ (c f₁ f₂ : Type u) (r₁ r₂ : Type v) :
           Cardinal.lift.{u} #r₁ + Cardinal.lift.{u} #r₂ := by
   simp [card_eq_card_functions_add_card_relations, add_assoc]
 
-instance {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (f n)] :
+/-- Passes a `DecidableEq` instance on a type of function symbols through the  `Language`
+constructor. Despite the fact that this is proven by `inferInstance`, it is still needed -
+see the `example`s in `ModelTheory/Ring/Basic`.  -/
+instance instDecidableEqFunctions {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (f n)] :
     DecidableEq ((⟨f, R⟩ : Language).Functions n) := inferInstance
 
+/-- Passes a `DecidableEq` instance on a type of relation symbols through the  `Language`
+constructor. Despite the fact that this is proven by `inferInstance`, it is still needed -
+see the `example`s in `ModelTheory/Ring/Basic`.  -/
 instance {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (R n)] :
     DecidableEq ((⟨f, R⟩ : Language).Relations n) := inferInstance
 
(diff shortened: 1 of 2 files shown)


Mathlib/ModelTheory/Basic.lean
```diff
@@ -81,6 +81,16 @@ instance inhabited₂ [h : Inhabited a₂] : Inhabited (Sequence₂ a₀ a₁ a
...
 
+instance [DecidableEq a₀] [DecidableEq a₁] [DecidableEq a₂] {n : ℕ} :
+    DecidableEq (Sequence₂ a₀ a₁ a₂ n) := by
+  cases n with
+  | zero => infer_instance
+  | succ n => cases n with
+    | zero => infer_instance
+    | succ n => cases n with
+      | zero => infer_instance
+      | succ n => infer_instance
```
eric-wieser, line 92:
```suggestion
instance [DecidableEq a₀] [DecidableEq a₁] [DecidableEq a₂] {n : ℕ} :
    DecidableEq (Sequence₂ a₀ a₁ a₂ n) :=
  match n with
  | 0 | 1 | 2 => ‹_›
  | _ + 3 => inferInstance
```

Mathlib/ModelTheory/Basic.lean
```diff
@@ -240,6 +250,12 @@ theorem card_mk₂ (c f₁ f₂ : Type u) (r₁ r₂ : Type v) :
           Cardinal.lift.{u} #r₁ + Cardinal.lift.{u} #r₂ := by
   simp [card_eq_card_functions_add_card_relations, add_assoc]
 
+instance {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (f n)] :
+    DecidableEq ((⟨f, R⟩ : Language).Functions n) := inferInstance
+
+instance {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (R n)] :
+    DecidableEq ((⟨f, R⟩ : Language).Relations n) := inferInstance
```
eric-wieser, line 257: If `inferInstance` can find these, they're probably not very useful as `instances`; so maybe just make them `example`s?

awainverse: These are the things that replace the deleted lines in `ModelTheory.Ring.Basic` - somehow these declarations still help it find these instances. I'd love to know why and find a better solution, but for now, something like these is necessary.

eric-wieser: Please add a comment to that effect (and an `example` to replace the deleted lines in `ModelTheory.Ring.Basic` to demonstrate that the instances work)

Mathlib/ModelTheory/Basic.lean
```diff
@@ -246,9 +246,15 @@ theorem card_mk₂ (c f₁ f₂ : Type u) (r₁ r₂ : Type v) :
...
+/-- Passes a `DecidableEq` instance on a type of function symbols through the  `Language`
+constructor. Despite the fact that this is proven by `inferInstance`, it is still needed -
+see the `example`s in `ModelTheory/Ring/Basic`.  -/
+instance instDecidableEqFunctions {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (f n)] :
     DecidableEq ((⟨f, R⟩ : Language).Functions n) := inferInstance
 
+/-- Passes a `DecidableEq` instance on a type of relation symbols through the  `Language`
+constructor. Despite the fact that this is proven by `inferInstance`, it is still needed -
+see the `example`s in `ModelTheory/Ring/Basic`.  -/
 instance {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (R n)] :
```
eric-wieser, line 258:
```suggestion
instance instDecidableEqRelations {f : ℕ → Type*} {R : ℕ → Type*} (n : ℕ) [DecidableEq (R n)] :
```

eric-wieser:
bors d+

Thanks!

PR #15906, opened 2024-08-17
[Merged by Bors] - feat: define singular manifolds

grunweg:
These are used to define bordism groups: the `n`th bordism group of a topological space `X` is the space of cobordism classes of all singular manifolds on `X`.

---

This PR can best be seen in context in #23138 - which also proves that this definition is usable.
(Older branches in this direction are [this branch](https://github.com/leanprover-community/mathlib4/compare/master...MR-bordism-theory) (my first iteration) and [this one](https://github.com/leanprover-community/mathlib4/compare/master...MR-bordism-theory-v2) (second iteration); #23138 has a bit less progress on the boundary data, but pushes bordisms further.


grunweg:
Note to self: merge my improvements to the bordism theory branch here, once the dependent PR has landed.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+  /-- `M` is a smooth manifold with corners -/
+  [smoothMfd : SmoothManifoldWithCorners model M]
+  /-- `M` is compact -/
+  [compactSpace : CompactSpace M]
+  /-- `M` is boundaryless -/
+  [boundaryless: BoundarylessManifold model M]
+  /-- `M` is finite-dimensional, as its model space `E` is -/
+  [findim: FiniteDimensional ℝ E]
+  /-- `M` is `n`-dimensional, as its model space `E` is -/
+  [dimension : finrank ℝ E = n]
```
alreadydone, line 82: I'd suggest that you use `EuclideanSpace ℝ n` and get rid of `E` and `H`.

grunweg: This makes speaking about products more difficult: when talking about products of bordism classes and the (un)oriented bordism ring, that would be needed. Why would you use euclidean space instead?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+open FiniteDimensional Set
+
+noncomputable section
+
+universe u in
+/-- A **singular `n`-manifold** on a topological space `X`, for `n ∈ ℕ`, is a pair `(M,f)`
+of a closed `n`-dimensional smooth manifold `M` together with a continuous map `M → X`. -/
+structure SingularNManifold (X : Type*) [TopologicalSpace X] (n : ℕ) where
+  /-- The normed space on which the manifold `M` is modeled. -/
+  E : Type u
```
sgouezel, line 56: I don't think it can be correct to assume that `E` lives in the same universe as `M`: typically, I expect `M` to live in a generic universe, while `E` would be `R^n` (living in `Type`). So your definition should probably depend on two universes `u` (for `M`) and `v` (for `E` and `H`)?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+a number of prerequisites are missing from mathlib, but a surprising amount of progress
+can already be made today.
+
+Currently, this file only contains the definition of *singular *n*-manifolds*:
+bordism classes are the equivalence classes of singular n-manifolds w.r.t. the (co)bordism relation.
+
+## Main definitions
+
+- **SingularNManifold**: a singular `n`-manifold on a topological space `X`, for `n ∈ ℕ`, is a pair
+`(M,f)` of a closed `n`-dimensional smooth manifold `M` together with a continuous map `M → X`.
```
sgouezel, line 22:
```suggestion
`(M, f)` of a closed `n`-dimensional smooth manifold `M` together with a continuous map `M → X`.
```

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+  /-- The smooth manifold `M` is a topological space -/
+  [topSpaceM : TopologicalSpace M]
+  /-- The topological space on which the manifold `M` is modeled -/
+  H : Type u
+  /-- `H` is a topological space -/
+  [topSpaceH : TopologicalSpace H]
+  /-- The smooth manifold `M` is a charted space over `H` -/
+  [chartedSpace : ChartedSpace H M]
+  /-- The model with corners for the manifold `M` -/
+  model : ModelWithCorners ℝ E H
```
sgouezel, line 72: Maybe call it `I` instead of `model`, to follow the convention in the rest of the library?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+  model := I
+  dimension := hdim
+  f := id
+  hf := continuous_id
+
+/-- If `(N, f)` is a singular `n`-manifold on `X` and `M` another `n`-dimensional smooth manifold,
+a smooth map `φ : M → N` induces a singular `n`-manifold structure `(M, f ∘ φ)` on `X`. -/
+noncomputable def comap [h: Fact (finrank ℝ E = n)]
+    (s : SingularNManifold X n)
+    {φ : M → s.M} (hφ : Smooth I s.model φ) : SingularNManifold X n where
```
sgouezel, line 158: It's a little bit strange to assume that phi is smooth, while continuous would do here, no?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+noncomputable def refl (hdim : finrank ℝ E = n) : SingularNManifold M n where
+  H := H
+  model := I
+  dimension := hdim
+  f := id
+  hf := continuous_id
+
+/-- If `(N, f)` is a singular `n`-manifold on `X` and `M` another `n`-dimensional smooth manifold,
+a smooth map `φ : M → N` induces a singular `n`-manifold structure `(M, f ∘ φ)` on `X`. -/
+noncomputable def comap [h: Fact (finrank ℝ E = n)]
```
sgouezel, line 156:
```suggestion
noncomputable def comap [h : Fact (finrank ℝ E = n)]
```
Same thing below

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+/-- An `n`-dimensional manifold induces a singular `n`-manifold on the one-point space. -/
+def trivial [h: Fact (finrank ℝ E = n)] : SingularNManifold PUnit n where
+  E := E
+  M := M
+  model := I
+  dimension := h.out
+  f := fun _ ↦ PUnit.unit
+  hf := continuous_const
+
+/-- The product of a singular `n`- and a `m`-manifold into a one-point space
```
sgouezel, line 199:
```suggestion
/-- The product of a singular `n`- and a singular `m`-manifold into a one-point space
```

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,211 @@
...
+is a singular `n+m`-manifold. -/
+-- FUTURE: prove that this observation inducess a commutative ring structure
```
sgouezel, line 201:
```suggestion
-- FUTURE: prove that this observation induces a commutative ring structure
```

sgouezel (commented):
Can you sketch a little bit more the maths behind this, which I don't know? (For me, a cobordism between two manifolds M and N is a manifold one dimension higher with two boundary components which are diffeomorphic to M and N respectively, but I don't see how your singular manifolds fit into this picture). Either in the file-level docstring, or in the PR, as you prefer.

I think it would help me a lot to see a theorem which is proved with this formalism (the proof of the pudding is in the eating). Do you have a working branch with a specific theorem you could point me to, so that I get a better grasp of where you want to go? (Ideally you could even open a PR with a work in progress branch, for easier discussion there)

YaelDillies:
> Do you have a working branch with a specific theorem you could point me to, so that I get a better grasp of where you want to go?

The PR description links to https://github.com/leanprover-community/mathlib4/compare/master...MR-bordism-theory

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: Sure! I tend to think one-letter variable names are harder to understand, but if you find `I` better, that is convincing to me.

grunweg:
Thanks for the review! I agree that not everything is self-explanatory; let me expand on the background.

About bordisms: the first time I heard about cobordisms, it was the definition you mentioned. The general definition involves an additional map: two singular n-manifolds (M,f:M\to X) and (N,g:N\to X) are cobordant if there's a cobordism between them --- which is a compact (n+1)-manifold W together with a map F\to X such that
- the boundary of W is diffeomorphic to the disjoint union of M and N, and
- under this diffeomorphism, the map F restricts to f resp. g.

In particular, if X is a one-point space, one arrives at the definition you give.

**References.** I learned this from my advisor's course (see [these notes](https://www.mathematik.hu-berlin.de/~wendl/Sommer2024/Topologie3/comments.pdf), week 14). I'm just formalising parts of the first page... as formalisation takes more time and space so far.
Dan Freed's lecture notes (https://web.ma.utexas.edu/users/dafr/bordism.pdf) are also supposed to be pretty good.
The manifold atlas page http://www.map.mpim-bonn.mpg.de/B-Bordism#Piecewise_linear_and_topological_bordism has much more generality. Mathlib is not ready for that yet, not even close.

**Bigger picture for this PR**. As Yael mentioned, [this branch](https://github.com/leanprover-community/mathlib4/compare/master...MR-bordism-theory) has my overall work so far, leading up to defining the unoriented bordism groups. At the end, I realised I needed to change the design of singular n-manifolds (to the current one); I haven't gone through and adapted all of that branch yet. When I do, it shouldn't be too hard to prover that the unordered bordism group is a group.
There are a number of sorries, most of which are missing pre-requisites like "the disjoint union of two n-dimensional manifolds is an n-manifold". Doable, should be in mathlib, but someone needs to do it. (And I'm supposed to work on e.g. the Carleson project now.)

grunweg:
I have just adapted this PR to latest master, and also
- generalised the definition to `C^n` manifolds: while smooth bordism groups are the most classical ones, and the basis for e.g. the continuous case as well, the definition of singular n-manifolds generalises to the continuous setting, and is also used there.
- thought about the model space question: it would surely be nice if we could just use R^n, but I don't think it's that simple. Medium-term, we'd like to perform two constructions with singular n-manifolds: take the direct sum (to define bordism groups), and take the product (for the bordism *ring*).
Taking products means we'll have to take ModelProd as a model space, and using R^n and R^m becomes painful to impossible to use here: we *have* to do something else. On the other hand, the direct sum construction requires *the same* model space for now. However, that step is more flexible: if we have an equivalence of the model to a common model, we can translate the charted space structures are still reach our goal.
This yields the eventual definition I chose: model singular n-manifolds on some space H, but keep a homeomorphism to R^n around (which is used for the direct sum).

- I am not sure about the universes: the checkUnivs linter still complains if I change them as you say.


There are three sorries remaining, but none of them is mathematically problematic (nor interesting).

grunweg:
Actually, the other sorry *is* interesting. I'll think about it more and update this PR accordingly.

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: I have changed my mind again, to parametrising the definition over `E` and `H` instead: this allows disjoint unions (which require the same model space) and products (take `ModelProd`), but does not close the door to using `EuclideanSpace R n` in applications.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+
+Currently, this file only contains the definition of *singular *n*-manifolds*:
+bordism classes are the equivalence classes of singular n-manifolds w.r.t. the (co)bordism relation
+and will be added in a future PR, as well as the definition of the (unoriented) bordism groups.
+
+## Main definitions
+
+- **SingularNManifold**: a singular `n`-manifold on a topological space `X`, for `n ∈ ℕ`, is a pair
+`(M, f)` of a closed `n`-dimensional smooth manifold `M` together with a continuous map `M → X`.
+We don't assume `M` to be modelled on `ℝ^n` (nor to be using with the standard model),
```
sgouezel, line 24:
```suggestion
We don't assume `M` to be modelled on `ℝ^n` (nor to be using with the standard model),
```
I'm not sure I understand the parenthesis here. Could you clarify?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+but instead add the model topological space `H`, the vector space `E` and the model with corners `I`
+as type parameters.
+- `SingularNManifold.map`: a map `X → Y` of topological spaces induces a map between the spaces
+of singular n-manifolds
+- `SingularNManifold.comap`: if `(N,f)` is a singular n-manifold on `X` and `φ: M → N` is smooth,
+the `comap` of `(N,f)` and `φ` is the induced singular n-manifold `(M, f ∘ φ)` on `X`.
+- `SingularNManifold.empty`: the empty set `M`, viewed as an `n`-manifold,
+as a singular `n`-manifold over any space `X`
```
sgouezel, line 32:
```suggestion
as a singular `n`-manifold over any space `X`.
```

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+- `SingularNManifold.trivial`: an `n`-dimensional manifold induces a singular `n`-manifold
+on the one-point space.
+- `SingularNManifold.prod`: the product of a singular `n`-manifold and a singular `m`-manifold
+on the one-point space, is a singular `n+m`-manifold on the one-point space.
+- `SingularNManifold.sum`: the disjoint union of two singular `n`-manifolds
+is a singular `n`-manifold
```
sgouezel, line 38:
```suggestion
is a singular `n`-manifold.
```

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+## (Unoriented) bordism theory
+
+This file defines the beginnings of (unoriented) bordism theory. For the full definition of
+smooth oriented bordism groups, a number of prerequisites are missing from mathlib. However,
+a significant amount of this work is already possible.
+
+Currently, this file only contains the definition of *singular *n*-manifolds*:
+bordism classes are the equivalence classes of singular n-manifolds w.r.t. the (co)bordism relation
+and will be added in a future PR, as well as the definition of the (unoriented) bordism groups.
+
```
sgouezel, line 19: Could you expand a little bit the docstring here, to explain what singular `n` manifolds are: it's not manifolds where you allow some singularities (which is what the name seems to indicate), it's rather a smooth manifold enriched with a continuous map to a topological space `X`. Also explain that, for `X = a point`, you just get manifolds in the usual sense. And maybe sketch an example of something that can be done with this formalism but which can not be done using manifolds in the usual sense?

Note that wikipedia, for instance, only defines bordism in the usual sense (where two manifolds are cobordant if they are the boundaries of a one-dimensional higher manifold), so it's a little bit surprising to say that you are doing "the beginnings of bordism" and go directly to a slightly exotic setting. So I think this deserves big warnings, and motivation in the docstring.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+noncomputable section
+
+/-- A **singular `n`-manifold** on a topological space `X`, for `n ∈ ℕ`, is a pair `(M, f)`
+of a closed `n`-dimensional `C^k` manifold `M` together with a continuous map `M → X`.
+We assume that `M` is a manifold over the pair `(E, H)` with model `I`.
+
+In practice, one commonly wants to take `k=∞` (as then e.g. the intersection form is a powerful tool
+to compute bordism groups; for the definition, this makes no difference.) -/
+structure SingularNManifold (X : Type*) [TopologicalSpace X] (n : ℕ) (k : ℕ∞)
```
sgouezel, line 68:
```suggestion
structure SingularNManifold (X : Type*) [TopologicalSpace X] (n : ℕ) (k : WithTop ℕ∞)
```
? Or do you want to exclude analytic manifolds for some reason? Same thing below.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+open Module Set
+
+noncomputable section
+
+/-- A **singular `n`-manifold** on a topological space `X`, for `n ∈ ℕ`, is a pair `(M, f)`
+of a closed `n`-dimensional `C^k` manifold `M` together with a continuous map `M → X`.
+We assume that `M` is a manifold over the pair `(E, H)` with model `I`.
+
+In practice, one commonly wants to take `k=∞` (as then e.g. the intersection form is a powerful tool
+to compute bordism groups; for the definition, this makes no difference.) -/
```
sgouezel, line 67: Can you mention in the docstring that you want to bundle everything to construct a group out of it?

Also, once you bundle everything, I don't know if you would be better off using the category theory framework, or if it would just make things more complicated.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+structure SingularNManifold (X : Type*) [TopologicalSpace X] (n : ℕ) (k : ℕ∞)
+  {E H : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] [FiniteDimensional ℝ E]
+  [TopologicalSpace H] (I : ModelWithCorners ℝ E H) where
+  /-- The manifold `M` of a singular `n`-manifold `(M, f)` -/
+  M : Type*
```
sgouezel, line 72: Better put an explicit universe here, because your structure will depend on this universe, which can not be inferred from anything else.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+  /-- The manifold `M` is a topological space. -/
+  [topSpaceM : TopologicalSpace M]
+  /-- The manifold `M` is a charted space over `H`. -/
+  [chartedSpace: ChartedSpace H M]
```
sgouezel, line 76:
```suggestion
  [chartedSpace : ChartedSpace H M]
```
Same thing below

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+    (s.map hφ).f = φ ∘ s.f :=
+  rfl
+
+lemma map_comp (s : SingularNManifold X n k I)
+    {φ : X → Y} {ψ : Y → Z} (hφ : Continuous φ) (hψ : Continuous ψ) :
+    ((s.map hφ).map hψ).f = (ψ ∘ φ) ∘ s.f := by
+  simp [Function.comp_def]
+  rfl
+
+-- Let M' and W be real C^k manifolds.
```
sgouezel, line 125: I don't see `W` below

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+    SingularNManifold M n k I where
+  dimension := hdim
+  f := id
+  hf := continuous_id
+
+/-- If `(N, f)` is a singular `n`-manifold on `X` and `M` another `n`-dimensional manifold,
+a smooth map `φ : M → N` induces a singular `n`-manifold structure `(M, f ∘ φ)` on `X`. -/
+noncomputable def comap [h : Fact (finrank ℝ E' = n)]
+    (s : SingularNManifold X n k I)
+    {φ : M' → s.M} (hφ : ContMDiff I' I n φ) : SingularNManifold X n k I' where
```
sgouezel, line 147: There's a funny confusion here: you are assuming that phi is `C^n`, where `n`  is the dimension. I think you meant `C^k`. But you should just assume continuous, because it's what you need.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+@[simp]
+lemma comap_f [Fact (finrank ℝ E' = n)]
+    (s : SingularNManifold X n k I) {φ : M' → s.M} (hφ : ContMDiff I' I n φ) :
+    (s.comap hφ).f = s.f ∘ φ :=
+  rfl
+
+variable (M I) in
+/-- The canonical singular `n`-manifold associated to the empty set (seen as an `n`-dimensional
+manifold, i.e. modelled on an `n`-dimensional space). -/
+def empty [h: Fact (finrank ℝ E = n)] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
```
sgouezel, line 161:
```suggestion
def empty [h : Fact (finrank ℝ E = n)] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
```
(You could search for the regexp `\S:` in this file...)

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+  M := M
+  dimension := h.out
+  f := fun x ↦ (IsEmpty.false x).elim
+  hf := by
+    rw [continuous_iff_continuousAt]
+    exact fun x ↦ (IsEmpty.false x).elim
+
+variable (M I) in
+/-- An `n`-dimensional manifold induces a singular `n`-manifold on the one-point space. -/
+def trivial [h: Fact (finrank ℝ E = n)] : SingularNManifold PUnit n k I where
```
sgouezel, line 172: It's not really trivial, as all the topology and smooth structure of `M` is in there, as well as the standard bordism theory. `toPUnit`, maybe?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,202 @@
...
+  dimension := by rw [finrank_prod, s.dimension, t.dimension]
+
+variable (s t : SingularNManifold X n k I)
+
+/-- The disjoint union of two singular `n`-manifolds on `X` is a singular `n`-manifold on `X`. -/
+-- We need to choose a model space for the disjoint union (as a priori `s` and `t` could be
+-- modelled on very different spaces: for simplicity, we choose `ℝ^n`; all real work is contained
+-- in the two instances above.
+def sum {n : ℕ} (s t : SingularNManifold X n k I) [finrank: Fact (finrank ℝ E = n)] :
+    SingularNManifold X n k I where
```
sgouezel, line 196: I am confused by the design and the comment here. One you fix `I`, you are fixing the model space, so by definition `s` and `t` have the same model space, and there is nothing special to do: just use `I` also for the disjoint union. On the other hand, it makes me realize that the parameter `n` is completely useless in all the story, as it is also determined by `I` (it's the dimension of the target space of `I`). So why not just remove it completely from the definition of `SingularNManifold` (and from the whole file)?

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: You're right, twice!

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: No, just an oversight: analytic manifolds should work fine.

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: I will expand this doc-string (thanks for providing some hints). However, let me say that your claim about wikipedia is not fully correct: it mentions bordism groups as an extraordinary cohomology theory https://en.wikipedia.org/wiki/Cobordism#Cobordism_as_an_extraordinary_cohomology_theory. There are absolute and relative bordism groups; wikipedia only hints at this fact (and mostly mentions the absolute ones).

sgouezel: You're absolutely right! What got me is that I looked for the word "singular" in the wikipedia page and didn't find it. Is the terminology "singular manifold" for a manifold together with a map to `X` completely standard?

grunweg: Good question - let me check some more sources.

The terminology is certainly intentional, to parallel the definition of singular homology. A singular n-chain on a space X is a continuous map from the n-simplex into X. Singular n-manifolds replace the domain by closed smooth n-manifolds instead, hence the term singular n-manifold.
(I will make sure to mention this in the module doc-string.)

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: Good point. In the latest version, `n` is removed from the discussion, and only `I` is a type parameter.

Mathlib/Geometry/Manifold/Bordisms.lean
grunweg: I've mentioned that everything is bundled. I don't think using category theory would make this easier.

sgouezel (commented):
Looks good!

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+XXX design decisions, model parameters etc.
+
+## Main definitions
+
+- **SingularNManifold**: a singular `n`-manifold on a topological space `X`, for `n ∈ ℕ`, is a pair
+  `(M, f)` of a closed `n`-dimensional smooth manifold `M` together with a continuous map `M → X`.
+  We don't assume `M` to be modelled on `ℝⁿ`, but add the model topological space `H`,
+  the vector space `E` and the model with corners `I` as type parameters.
+  To define a disjoint unions of singular n-manifolds, we require their domains to be manifolds
+  over the same model with corners: the model must be explicit (or M be modelled over `ℝⁿ`)
```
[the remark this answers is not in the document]
grunweg: Removed; the implementation notes cover what I mean to say.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+
+## Implementation notes
+
+* We choose a bundled design for singular n-manifolds (and also for bordisms): to construct a group
+  structure on the set of bordism classes, having that be a type is useful.
+* The underlying model with corners is a type parameter, as defining a disjoint union of singular
+  n-manifolds (which is one part of defining the group operation)
+  requires their domains to be manifolds over the same model with corners. Thus, either we restrict
+  to manifolds modelled over `𝓡n` (which we prefer not to), or the model must be a type parameter.
+* Having `SingularNManifold` contain the type `M` as explicit parameter is not ideal, but the best
```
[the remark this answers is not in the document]
grunweg: Good catch! I meant `M` being a field of `SingularNManifold`: is the new phrasing clearer?

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+  solution we found --- as this adds a universe parameter to the structure. We generally cannot have
+  `M` live in the same universe as `X` (a common case is `X` being `PUnit`),
+  determining the universe of `M` via the universes of `E` and `H` makes `SingularNManifold.map`
+  painful to state (as that would require `ULift`ing `M`).
+
```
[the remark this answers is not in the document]
grunweg: I think I did that now - please say if this is not clear enough!

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+lemma map_comp (s : SingularNManifold X k I)
+    {φ : X → Y} {ψ : Y → Z} (hφ : Continuous φ) (hψ : Continuous ψ) :
+    ((s.map hφ).map hψ).f = (ψ ∘ φ) ∘ s.f := by
+  simp [Function.comp_def]
+
+variable {E' E'' E''' H' H'' H''' : Type*}
+  [NormedAddCommGroup E'] [NormedSpace ℝ E'] [NormedAddCommGroup E'']  [NormedSpace ℝ E'']
+  [NormedAddCommGroup E'''] [NormedSpace ℝ E''']
+  [TopologicalSpace H'] [TopologicalSpace H''] [TopologicalSpace H''']
```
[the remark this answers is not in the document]
grunweg: Indeed, I am not. Removed.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+  f := fun _ ↦ PUnit.unit
+  hf := continuous_const
+
+variable {I' : ModelWithCorners ℝ E' H'} [FiniteDimensional ℝ E']
+
+/-- The product of a singular `n`- and a singular `m`-manifold into a one-point space
+is a singular `n+m`-manifold. -/
+-- FUTURE: prove that this observation induces a commutative ring structure
+-- on the unoriented bordism group `Ω_n^O = Ω_n^O(pt)`.
+def prod (s : SingularNManifold PUnit k I) (t : SingularNManifold PUnit k I') :
```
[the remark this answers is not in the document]
grunweg: Interesting suggestions; I had not considered those! I'm not sure if these are mathematically interesting, though: the main use of the product of singular `n`-manifolds is to define the *bordism ring*: the direct sum of the bordism groups on `PUnit` has a ring structure (given by the product of manifolds; in Lean, you also need to apply some isomorphism of models to make sure you always get the same model).

For that, you need the product of two singular manifolds on `X` to be another singular manifold on `X` --- so additive groups don't seem like a useful generalisation. If you know any other space `X` with the property that `X \times X` is naturally `X`, again, one can probably define a bordism ring on `X`. (I don't think that's widely used at all, though.)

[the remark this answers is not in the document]
grunweg: So, my instinct is to leave this as-is; I'm happy to add a comment.

Mathlib/Geometry/Manifold/Bordisms.lean
```diff
@@ -0,0 +1,238 @@
...
+
+singular n-manifold, bordism, bordism group
+-/
+
+open scoped Manifold
+open Module Set
+
+suppress_compilation
+
+/-- A **singular `n`-manifold** on a topological space `X`, for `n ∈ ℕ`, is a pair `(M, f)`
```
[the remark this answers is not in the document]
grunweg: Good point, and very good question! I'd like to think about that some more - let me sleep over it.

grunweg:
Thanks a lot for the careful review, you made some very good points! I have addressed most comments and responded to the others. I tend to leave products as-is for now, and will think more about the name "singular `I`-manifolds".

Mathlib/Geometry/Manifold/Bordisms.lean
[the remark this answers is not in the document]
sgouezel: When you have an additive group, the product of singular manifolds is indeed a singular manifold over the same space, as you have a canonical map from X x X to X (addition). You don't need a homeo between X x X and X, just a continuous map.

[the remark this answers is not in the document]
grunweg: :bulb: Ah, now I understand. Right - that generalisation would work. I'm not sure at all if it is used in the literature.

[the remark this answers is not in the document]
grunweg: I've done some research - see my main comment below.

grunweg:
I've done a bit of research: mostly, "the bordism ring" is the bordism ring of the the one-point space, i.e. `(P)Unit`. One can define a ring structure more generally, but in a more elaborate way: the [nlab](https://ncatlab.org/nlab/show/bordism+ring) describes it in some detail:
Given two maps `f: M\to X` and `g: N\to X`, their product is the fibre product of M and N (w.r.t. f and g) with the induced map to X. (If f and g are transverse, the fibre product is a smooth manifold and we're done. Otherwise, perturb f or g to make them transverse, and then take the fibre product. You need to prove that the result is independent of the choice of this auxiliary perturbation.)

This is quite different from the simple operation I'm performing --- hence naively generalising the current construction would be the wrong thing to do. (Formalising the real product is out of reach right now: mathlib does not have the notion of transversality, nor the smoothness of fibre products of transverse maps, and of course also not the transversality theorem.)

I'll add a comment about this. What do you think about the following?

NB. This definition as written makes sense more generally, for `SingularNManifold X k I` whenever `X` is a topological (additive) group. However, this would not be correct if `X` is not `PUnit`: the bordism ring can be defined for every `C^k` manifold `X`, but the product of two singular manifolds `(M, f)` and `(N, g)` is the *fibre product* of `M` and `N` w.r.t. `f` and `g`, with its induced map into `X`. (If `f` and `g` intersect transversely, this fibre product is a smooth manifold, of dimension `dim M + dim N - dim X`. Otherwise, the transversality theorem proves that `f` (or `g`) can be perturbed by an arbitrarily small amount, to make `f` and `g` transverse. One can prove that different perturbations yield bordant manifolds.)

grunweg:
I've thought about terminology again: "singular manifold" seems just fine (as is also used by the nlab). I have dropped the `n`, and occasionally mentioned `I`.

And I added the comment about the product operation: this seems ready for another look.

Mathlib.lean
```diff
@@ -3426,6 +3426,7 @@ import Mathlib.Geometry.Manifold.Algebra.Monoid
 import Mathlib.Geometry.Manifold.Algebra.SmoothFunctions
 import Mathlib.Geometry.Manifold.Algebra.Structures
 import Mathlib.Geometry.Manifold.AnalyticManifold
+import Mathlib.Geometry.Manifold.Bordisms
```
sgouezel, line 3429:
```suggestion
import Mathlib.Geometry.Manifold.Bordism
```
Most of the time, we don't use plural in general for filenames, I think.

sgouezel (commented):
bors d+
Thanks!

grunweg:
Thanks for the review!
bors merge

PR #15907, opened 2024-08-17
[Merged by Bors] - feat(Order/SuccPred/Basic): `succ_eq_sInf`

vihdzp:
We prove that the successor of an element is the infimum of those greater than it, both in a complete lattice and a conditionally complete lattice with no maxima.

---

### DIFF
--- a/Mathlib/Order/SuccPred/Basic.lean
+++ b/Mathlib/Order/SuccPred/Basic.lean
@@ -3,7 +3,7 @@ Copyright (c) 2021 Yaël Dillies. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Yaël Dillies
 -/
-import Mathlib.Order.CompleteLattice
+import Mathlib.Order.ConditionallyCompleteLattice.Basic
 import Mathlib.Order.Cover
 import Mathlib.Order.GaloisConnection
 import Mathlib.Order.Iterate
@@ -505,18 +505,22 @@ instance [PartialOrder α] : Subsingleton (SuccOrder α) :=
     · exact (@IsMax.succ_eq _ _ h₀ _ ha).trans ha.succ_eq.symm
     · exact @CovBy.succ_eq _ _ h₀ _ _ (covBy_succ_of_not_isMax ha)⟩
 
-section CompleteLattice
+section Lattice
 
-variable [CompleteLattice α] [SuccOrder α]
-
-theorem succ_eq_iInf (a : α) : succ a = ⨅ (b) (_ : a < b), b := by
-  refine le_antisymm (le_iInf fun b => le_iInf succ_le_of_lt) ?_
+theorem succ_eq_sInf [CompleteLattice α] [SuccOrder α] (a : α) :
+    succ a = sInf (Set.Ioi a) := by
+  apply (le_sInf fun b => succ_le_of_lt).antisymm
   obtain rfl | ha := eq_or_ne a ⊤
   · rw [succ_top]
     exact le_top
-  exact iInf₂_le _ (lt_succ_iff_ne_top.2 ha)
+  · exact sInf_le (lt_succ_iff_ne_top.2 ha)
 
-end CompleteLattice
+theorem succ_eq_csInf [ConditionallyCompleteLattice α] [SuccOrder α] [NoMaxOrder α] (a : α) :
+    succ a = sInf (Set.Ioi a) := by
+  apply (le_csInf nonempty_Ioi fun b => succ_le_of_lt).antisymm
+  exact csInf_le ⟨a, fun b => le_of_lt⟩ <| lt_succ a
+
+end Lattice
 
 /-! ### Predecessor order -/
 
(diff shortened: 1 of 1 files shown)


Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -505,18 +505,18 @@ instance [PartialOrder α] : Subsingleton (SuccOrder α) :=
...
+    succ a = sInf (Set.Ioi a) := by
+  apply (le_sInf fun b => succ_le_of_lt).antisymm
   obtain rfl | ha := eq_or_ne a ⊤
   · rw [succ_top]
     exact le_top
-  exact iInf₂_le _ (lt_succ_iff_ne_top.2 ha)
+  · exact sInf_le (lt_succ_iff_ne_top.2 ha)
 
-end CompleteLattice
+theorem succ_eq_csInf [ConditionallyCompleteLattice α] [SuccOrder α] [NoMaxOrder α] (a : α) :
```
YaelDillies, line 516: If `a` is maximal, then `succ ⊤ = ⊤ = sInf ∅ = sInf (Ioi ⊤) = sInf (Ioi a)`
````suggestion
theorem succ_eq_csInf [ConditionallyCompleteLattice α] [SuccOrder α] (a : α) :
```

vihdzp: There's two theorems in that sequence I can't find
1. `IsMax a → IsTop a` for conditionally complete lattices
2. `IsMax a → sInf Ø = a`
Any idea if we have them? If not, I can add them in this PR.

YaelDillies: `IsMax.isTop`

vihdzp: Actually, I don't believe 2 is provable. All lemmas in the definition of a conditionally complete lattice assume that the set is nonempty.

YaelDillies: Hmm yeah

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -505,18 +505,18 @@ instance [PartialOrder α] : Subsingleton (SuccOrder α) :=
     · exact (@IsMax.succ_eq _ _ h₀ _ ha).trans ha.succ_eq.symm
     · exact @CovBy.succ_eq _ _ h₀ _ _ (covBy_succ_of_not_isMax ha)⟩
 
-section CompleteLattice
-
-variable [CompleteLattice α] [SuccOrder α]
-
-theorem succ_eq_iInf (a : α) : succ a = ⨅ (b) (_ : a < b), b := by
```
YaelDillies, line 512: Can you keep `succ_eq_iInf` around?

vihdzp: Wait, we don't seem to have the conditionally complete version of that? That's probably what we should be fixing instead.

YaelDillies: > Wait, we don't seem to have the conditionally complete version of that? That's probably what we should be fixing instead.

Yeah, because it's false. Nested conditional `iInf`s don't behave the way you expect.

vihdzp: `succ_eq_ciInf` is not true in general, and requires that the infimum be less than `sInf Ø`. Here's what the new proof of `succ_eq_iInf` looks like:
```
theorem succ_eq_iInf [CompleteLattice α] [SuccOrder α] (a : α) : succ a = ⨅ (b) (_ : a < b), b := by
  rw [succ_eq_sInf, iInf_subtype', iInf, Subtype.range_coe_subtype]
  rfl
```
Again, I feel like it's a bit dumb to have this, and if we will, why not `⨅ b ∈ Set.Iio a, b` instead? I'll add it if you really want it, but I personally don't get it.

YaelDillies: I mean, consider lemmas like `le_iInf₂`. `succ_eq_sInf` would be annoying to interface with it.

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -513,6 +513,10 @@ theorem succ_eq_sInf [CompleteLattice α] [SuccOrder α] (a : α) :
     exact le_top
   · exact sInf_le (lt_succ_iff_ne_top.2 ha)
 
+theorem succ_eq_iInf [CompleteLattice α] [SuccOrder α] (a : α) : succ a = ⨅ (b) (_ : a < b), b := by
```
YaelDillies, line 516: Does this work now?
```suggestion
theorem succ_eq_iInf [CompleteLattice α] [SuccOrder α] (a : α) : succ a = ⨅ b > a, b := by
```

vihdzp: It does! That honestly kinda justifies having these lemmas at all for me.

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -820,7 +820,7 @@ theorem pred_eq_sSup [CompleteLattice α] [PredOrder α] :
     ∀ a : α, pred a = sSup (Set.Iio a) :=
   succ_eq_sInf (α := αᵒᵈ)
 
-theorem pred_eq_iSup [CompleteLattice α] [PredOrder α] (a : α) : pred a = ⨆ (b) (_ : b < a), b :=
+theorem pred_eq_iSup [CompleteLattice α] [PredOrder α] : ∀ a : α, pred a = ⨆ b < a, b :=
   succ_eq_iInf (α := αᵒᵈ)
```
YaelDillies, line 824:
```suggestion
theorem pred_eq_iSup [CompleteLattice α] [PredOrder α] (a : α) : pred a = ⨆ b < a, b :=
  succ_eq_iInf (α := αᵒᵈ) _
```

YaelDillies (approved):
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15909, opened 2024-08-17
[Merged by Bors] - feat: the dollarSyntax linter

Parcly-Taxel:
Lint against being `$` used in place of the canonical `<|`: this is documented in the mathlib style guide.

---


grunweg:
Thanks for writing this linter; it is always great to see the style guide automatically enforced. Can you extract the replacements into their own PR? Then they can land independently and need not wait until the linter is reviewed.

grunweg:
The linter looks pretty straightforward. Looking at the style guide, I wonder whether always recommending <| and never `|>` is correct. Can you comment on this?

Parcly-Taxel:
> I wonder whether always recommending <| and never `|>` is correct

`|>` is a different operator.

grunweg:
> > I wonder whether always recommending <| and never `|>` is correct
> 
> `|>` is a different operator.

You are perfectly right; thanks for teaching me this difference!

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,43 @@ initialize addLinter cdotLinter
...
+  |_ => #[]
+
+@[inherit_doc linter.dollarSyntax]
+def dollarSyntaxLinter : Linter where run := withSetOptionIn fun stx ↦ do
+    unless Linter.getLinterValue linter.dollarSyntax (← getOptions) do
+      return
+    if (← MonadState.get).messages.hasErrors then
+      return
+    for s in findDollarSyntax stx do
+      Linter.logLint linter.dollarSyntax s m!"Please use '<|' and not '$' for the pipe operator."
```
grunweg, line 245: I like to think carefully about linter wording: what do you think about the following change?
```suggestion
      Linter.logLint linter.dollarSyntax s m!"Please use '<|' instead of '$' for the pipe operator."

File: Mathlib/Tactic/Linter/Lint.lean [hunk 2]
```diff
@@ -211,6 +211,43 @@ initialize addLinter cdotLinter
 
 end CDotLinter
 
+/-!
+# The `dollarSyntax` linter
+
+The `dollarSyntax` linter flags uses of `<|` that are achieved by typing `$`.
+-/
```
grunweg, line 218: How about adding a sentence of justification, here and to the `register option`?
"These are disallowed by the mathlib style guide, as using `<|` pairs better with `|>`."

Mathlib/Tactic/Linter/Lint.lean
```diff
@@ -211,6 +211,43 @@ initialize addLinter cdotLinter
...
+
+/-- The `dollarSyntax` linter flags uses of `<|` that are achieved by typing `$`. -/
+register_option linter.dollarSyntax : Bool := {
+  defValue := true
```
grunweg, line 222: Note to self: when #15616 lands, make sure to also tweak this default. (At the moment, it's not clear which PR will land first.)

test/Lint.lean
```diff
@@ -87,6 +87,15 @@ note: this linter can be disabled with `set_option linter.cdot false`
...
 example : Add Nat where add := (. + ·)
 
+set_option linter.dollarSyntax false in
+/--
+warning: Please use '<|' and not '$' for the pipe operator.
+note: this linter can be disabled with `set_option linter.dollarSyntax false`
+-/
+#guard_msgs in
+set_option linter.dollarSyntax true in
+example (f g : Nat → Nat) : Nat := f $ g 0
```
grunweg, line 97: Thanks for including a test! Can you extend it slightly, to
- include more than one dollar (writing "f $ g $ 0" suffices perfectly here)
- also include other kinds of syntax nodes? (Some ideas: include a theorem, definition or lemma; include this syntax inside a have, let or nested goal, and nested inside a more complex term proof. Ping me if this is unclear!)

grunweg (approved):
I took a very careful review at the linter and only found four very small comments.

Once the fixes have landed and you have addressed them, this is good to go. (Let me know if you prefer help with the last comment.)

test/Lint.lean
Parcly-Taxel: I've done this now.

grunweg (commented):
Thanks!
maintainer merge

grunweg:
Just as heads-up: you will need to cherry-pick `1fafd71ff90bd9899e24336fc149823f74981aae`; feel free to do so.

grunweg:
(A new file with six dollar symbols was merged in the mean-time.)

jcommelin (commented):
Thanks :tada:

bors merge

PR #15913, opened 2024-08-17
[Merged by Bors] - style: replace `$` with `<|`

Parcly-Taxel:
The mathlib style guide requires using `<|` instead of `$`. The linter in #15909 will enforce this.

---


eric-wieser:
bors merge

(indeed, this is style not a fix)

Thanks!

PR #15915, opened 2024-08-17
chore(SetTheory/Ordinal/CantorNormalForm): `CNF.exponents` and `CNF.coeffs`

vihdzp:
We rewrite theorems about the Cantor Normal Form in terms of their lists of exponents and coefficients, instead of awkwardly talking about their first and second entries.

We also put all theorems about the CNF in its own namespace. This keeps us from cluttering the `Ordinal` namespace, and it also avoids having to prefix everything by `CNF_` constantly.

---

- [x] depends on: #16998


Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -40,45 +41,65 @@ namespace Ordinal
...
 
-We special-case `CNF 0 o = CNF 1 o = [(0, o)]` for `o ≠ 0`.
+We special-case `CNF 0 o = CNF 1 o = [⟨0, o⟩]` for `o ≠ 0`.
 
-`CNF b (b ^ u₁ * v₁ + b ^ u₂ * v₂) = [(u₁, v₁), (u₂, v₂)]` -/
+`CNF b (b ^ u₁ * v₁ + b ^ u₂ * v₂) = [⟨u₁, v₁⟩, ⟨u₂, v₂⟩]` -/
 @[pp_nodot]
-def CNF (b o : Ordinal) : List (Ordinal × Ordinal) :=
-  CNFRec b [] (fun o _ho IH ↦ (log b o, o / b ^ log b o)::IH) o
+def CNF (b o : Ordinal) : List (Σ _ : Ordinal, Ordinal) :=
```
Command-Master, line 66: I don't understand the reason for making this a Sigma type, seeing as it's not dependent. Is it only to be able to use List.keys?

vihdzp: It's partially because of that, but it's also so that I can use [`AList.lookupFinsupp`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Finsupp/AList.html#AList.lookupFinsupp) to define the finitely supported function returning the CNF's coefficients in a subsequent PR (`AList` requires a Sigma type).

In practice, all this change really entails is that we write pairs as `⟨a, b⟩` instead of `(a, b)`.

Command-Master: Perhaps `CNF` could directly return a `Ordinal →₀ Ordinal`, then?

vihdzp: I think both forms are useful. `CNF_foldr` would be much more annoying to state with a finitely supported function.

Command-Master: Wouldn't it be fairly simple with `Finset.sort`? Although maybe with addition not being commutative it does make sense to keep the order

vihdzp: Hm, I suppose we could just use `Finset.sort` for that one.

Here's another example of a theorem that would be more annoying to state: `CNF (b ^ e * x + y) = CNF (b ^ e * x) ++ CNF y` for `y < b ^ e`. You could restate this with finsupps as saying the exponents of `CNF (b ^ e * x)` are all larger than those in `CNF y`, and that the coefficients of `CNF (b ^ e * x + y)` match either one of them depending if the exponent is larger or smaller than `e`, but that's a mouthful.

So yeah, I overall wouldn't want to see our list version of `CNF` go. Even if it's not the most natural for proving theorems, it makes the development of said API smoother.

vihdzp: A refactor I would agree with is renaming `CNF` to something like `CNF_List`, to then use the name `CNF` for the finsupp version.

vihdzp:
I renamed `CNF.coefficients` to `CNF.coeffs` for parity with [`Polynomial.coeffs`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Polynomial/Basic.html#Polynomial.coeffs).

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
digama0: I agree that this is not a good change, I think it would be better to just make sure you have the list functions you need

vihdzp: What do you mean by that?

digama0: What goes wrong if you use a plain product? Fix it so it doesn't go wrong anymore

digama0:
I don't like that this is mixing new theory and unrelated proof cleanup. Please do these things separately.

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
vihdzp: Nothing in this PR goes wrong. But it means I can't as readily use [`AList.lookupFinsupp`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Finsupp/AList.html#AList.lookupFinsupp) to define the finsupp version of the CNF in a subsequent PR.

That being said, I guess I could just use `List.map` to cast `(x, y)` into `⟨x, y⟩` in that definition. Since this change seems to be pretty controversial I think I'll just do that.

vihdzp:
Yeah, looking back at my code cleanup I'm not sure I agree with everything I did. I'll put that in a separate PR and clean up this one afterwards.

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -65,80 +63,132 @@ We special-case `CNF 0 o = CNF 1 o = [(0, o)]` for `o ≠ 0`.
 `CNF b (b ^ u₁ * v₁ + b ^ u₂ * v₂) = [(u₁, v₁), (u₂, v₂)]` -/
 @[pp_nodot]
 def CNF (b o : Ordinal) : List (Ordinal × Ordinal) :=
-  CNFRec b [] (fun o _ho IH ↦ (log b o, o / b ^ log b o)::IH) o
+  CNFRec b [] (fun o _ IH ↦ (log b o, o / b ^ log b o)::IH) o
+
+/-- The exponents of the Cantor normal form are stored in the first entries. -/
+def CNF.exponents (b o : Ordinal) := (CNF b o).map Prod.fst
```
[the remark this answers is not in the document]
vihdzp: Honestly, I think putting everything here in the `CNF` namespace is not a bad idea. Otherwise we end up with `CNF_exponents` and `CNF_coeffs` and later down the line `CNF_coeff` too - write that down across every single theorem name and it gets unwieldy fast.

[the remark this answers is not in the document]
vihdzp: You can think of this as matching `Polynomial` in some sense - we have `Polynomial.coeff` rather than `polynomialCoeff` or anything of the sort. This also comes with the added benefit that results on the CNF won't clutter the ordinal namespace for anyone not interested in the CNF.

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -65,80 +63,132 @@ We special-case `CNF 0 o = CNF 1 o = [(0, o)]` for `o ≠ 0`.
...
+/-- The coefficients of the Cantor normal form are stored in the second entries. -/
+def CNF.coeffs (b o : Ordinal) := (CNF b o).map Prod.snd
```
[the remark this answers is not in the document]
vihdzp: I think this wording might be a bit more confusing, since the Cantor normal form in its standard mathematical description has no such thing as first or second entries. Maybe this could read
```
The exponents of the Cantor normal form are stored in the first entries of elements in `CNF`.
```

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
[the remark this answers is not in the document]
vihdzp: I've pushed a commit that namespaces most of the file within the `CNF` namespace. I think it's a nice improvement to having to type `CNF_` everywhere, but let me know what you think.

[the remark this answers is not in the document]
vihdzp: A funny consequence of this is that basically every single theorem that existed in this file has to be deprecated... well, it's not like anyone was using this API anyways :P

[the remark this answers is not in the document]
vihdzp: If this is a change you agree with, I'll do the namespacing in yet another separate PR. I'll stop rambling now :)

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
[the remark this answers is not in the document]
YnirPaz: Sounds good!

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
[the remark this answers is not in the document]
YnirPaz: That's a good point about cluttering `Ordinal`. This approach looks good!

YnirPaz (approved):

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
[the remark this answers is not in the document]
vihdzp: Oh actually, it doesn't make much sense to do the namespacing in a separate PR, since that would introduce yet another round of deprecations. I'll just keep this PR as is.

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
vihdzp: I've been thinking about the [Veblen Normal Form](https://en.wikipedia.org/wiki/Veblen_function#Fundamental_sequences_for_the_Veblen_hierarchy) recently. That one pretty clearly needs to be a list, and if we ever want to have it in Mathlib, the easiest way is probably going to be by defining it in terms of the list form of the CNF.

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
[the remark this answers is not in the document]
vihdzp: Since I'm inside the `CNF` namespace, this isn't needed.

vihdzp:
PR should be ready for review now! Most of the diff is just due to the deprecations, and all the changes here should hopefully be non-controversial.

YaelDillies:
You have two paragraphs in your PR description. Can you make each paragraph its own PR?

vihdzp:
Moving to #28584.

PR #15916, opened 2024-08-17
feat(SetTheory/Ordinal/CantorNormalForm): Coefficients of CNF as finsupp

vihdzp:
We introduce `CNF_coeff b o e`, which returns the coefficient of `b^e` in the Cantor Normal Form of `o`.

---

- [ ] depends on: #15915

I'm not sure if `CNF_Alist` and `CNF_coeff` adhere to naming conventions. Naming suggestions welcome.


vihdzp:
Moved to #28586.

PR #15917, opened 2024-08-17
[Merged by Bors] - feat: Real-valued double counting

YaelDillies:
Generalise the existing double counting arguments to real numbers. This is useful to avoid dealing with floors and ceils everywhere.

From LeanAPAP

---


b-mehta (approved):
Thanks! Is this immediately useful somewhere, or just an improvement you noticed?

bors merge

YaelDillies:
It's coming from my proof of the very small doubling theorem. Though I think your proof avoids needing it, and your proof is better than mine.

b-mehta:
Okay fair enough, this is a positive change anyway

PR #15918, opened 2024-08-17
feat(CategoryTheory/Opposites) : op for functor category is a functor

nrolland:
---

### DIFF
--- a/Mathlib/CategoryTheory/Opposites.lean
+++ b/Mathlib/CategoryTheory/Opposites.lean
@@ -160,6 +160,14 @@ protected def op (F : C ⥤ D) : Cᵒᵖ ⥤ Dᵒᵖ where
   obj X := op (F.obj (unop X))
   map f := (F.map f.unop).op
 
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Opposites.lean
```diff
@@ -160,6 +160,14 @@ protected def op (F : C ⥤ D) : Cᵒᵖ ⥤ Dᵒᵖ where
   obj X := op (F.obj (unop X))
   map f := (F.map f.unop).op
 
+/-- The functor which assigns to a functor its opposite -/
+def opFunctor  :  (C ⥤ D)ᵒᵖ ⥤ (Cᵒᵖ ⥤ Dᵒᵖ)  where
```
joelriou, line 164: This is a duplicate of `Functor.opHom` that is defined further below in the same file.

nrolland: Ah ! I'll close this then..

PR #12728, opened 2024-05-07
[Merged by Bors] - feat(CategoryTheory): the localized category is monoidal

joelriou:
In this PR, we shall show that if `C` is a monoidal category and `W` is a class of morphism that is preserved by the tensor product, then the localized category is also monoidal.

(This shall be used in order to define a monoidal structure on the category of sheaves of abelian groups, and on categories of sheaves of modules.)

Co-authored-by: Dagur Asgeirsson <dagurtomas@gmail.com>

---

- [x] depends on: #20951
- [x] depends on: #11701
- [x] depends on: #18165 
- [x] depends on: #19894  
- [x] depends on: #20197
- [x] depends on: #20788 

This shall be used in order to obtain a monoidal category structure on sheaves of modules from the monoidal structure on presheaves of modules. This also prepares some of the bifunctor API for the left derived monoidal category structure (when the tensor product does not preserves `W` but can still be left derived).

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15919, opened 2024-08-17
[Merged by Bors] - chore: remove unnecessary uses of CategoryTheory.Equivalence.mk

joelriou:
An equivalence of categories consists of a functor, an inverse, and unit and counit isomorphisms which satisfy the same compatibilities as for adjoint functors. However, certain equivalences are constructed via the constructor `Equivalence.mk` which does not assume these compatibilities: by changing the unit isomorphism, it is possible to satisfy the compatibility. When only the existence of an equivalence matters, we may certainly use `Equivalence.mk` (e.g. inside proofs). However, the more we develop the library, the more we may need to unfold the definitions of unit/counit isomorphisms (as it happened in #12728). Then, I would suggest that we refrain from using `Equivalence.mk` as much as possible. In this PR, most uses of `Equivalence.mk` are removed: in most cases, the compatibilities satisfied by adjunctions were easy to prove.

---


Mathlib/CategoryTheory/Category/PartialFun.lean
```diff
@@ -145,13 +146,19 @@ noncomputable def partialFunEquivPointed : PartialFun.{u} ≌ Pointed :=
...
+              exact ⟨b, ⟨ha, h.symm⟩, rfl⟩
+  counitIso :=
     NatIso.ofComponents
       (fun X ↦ Pointed.Iso.mk (by classical exact Equiv.optionSubtypeNe X.point) (by rfl))
       fun {X Y} f ↦ Pointed.Hom.ext <| funext fun a ↦ by
         obtain _ | ⟨a, ha⟩ := a
         · exact f.map_point.symm
         simp_all [Option.casesOn'_eq_elim, Part.elim_toOption]
+  functor_unitIso_comp X := by
+    ext (_|x)
```
b-mehta, line 158: I'm slightly surprised aesop_cat can't do this?

Mathlib/CategoryTheory/Limits/Fubini.lean
```diff
@@ -507,28 +507,19 @@ noncomputable def limitCurrySwapCompLimIsoLimitCurryCompLim :
...
-  erw [NatTrans.id_app]
-  -- Why can't `simp` do this?
+  dsimp [limitCurrySwapCompLimIsoLimitCurryCompLim, Equivalence.counit]
+  rw [Category.assoc, Category.assoc, limitIsoLimitCurryCompLim_hom_π_π,
+    HasLimit.isoOfEquivalence_hom_π]
   dsimp
-  -- Porting note: the original proof only had `simp`.
-  -- However, now `CategoryTheory.Bifunctor.map_id` does not get used by `simp`
-  rw [CategoryTheory.Bifunctor.map_id]
+  erw [G.map_id]
```
b-mehta, line 514: Is there a way to avoid the erw here?

joelriou: I have managed to remove this one.

Mathlib/CategoryTheory/Category/PartialFun.lean
joelriou: No, it cannot.

dagurtomas:
maintainer merge

mattrobball:
!bench

Mathlib/CategoryTheory/Category/PartialFun.lean
mattrobball, line 158:
```suggestion
    ext (_ | x)
```

mattrobball:
Very nice.

mattrobball:
bors delegate+

Thanks!

Mathlib/CategoryTheory/Sums/Associator.lean
```diff
@@ -112,14 +112,14 @@ theorem inverseAssociator_map_inr_inr {X Y : E} (f : inr (inr X) ⟶ inr (inr Y)
...
-      (by rintro ((_|_)|_) ((_|_)|_) f <;> first | cases f | aesop_cat))
-    (NatIso.ofComponents (fun X => eqToIso
-      (by rcases X with (_|(_|_)) <;> rfl)) -- Porting note: aesop_cat fails
-      (by rintro (_|(_|_)) (_|(_|_)) f <;> first | cases f | aesop_cat))
+@[simps functor inverse]
+def associativity : (C ⊕ D) ⊕ E ≌ C ⊕ (D ⊕ E) where
+  functor := associator C D E
+  inverse := inverseAssociator C D E
+  unitIso := NatIso.ofComponents (by rintro ((_|_)|_) <;> exact Iso.refl _) (by
+    rintro ((_|_)|_) ((_|_)|_) f <;> first | cases f | aesop_cat)
```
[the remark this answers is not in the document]
joelriou: I have fixed this one. (There are many occurrences of `_|_` in other places in mathlib.)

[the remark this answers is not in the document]
mattrobball: Unfortunate, but not a responsibility here.

joelriou:
> Very nice.

Actually, I believe that at some point, some proofs of triangle compatibilities were removed because of performances issues. The situation looks better now :-)

mattrobball:
When it passes CI, please merge

joelriou:
Thanks!

bors merge

PR #15923, opened 2024-08-17
[Merged by Bors] - chore(*): use `rfl` to prove `symm_symm`

urkud:
---


b-mehta:
Nice, thanks!

It looks like the declarations diff here gets very confused, this looks like a bug in that script?

bors merge

PR #15924, opened 2024-08-17
[Merged by Bors] - feat(Filter): add disjoint_comap_iff_map etc

urkud:
These are simple corollaries of `Filter.push_pull`.

---


dupuisf:
Thanks!

bors r+

PR #15926, opened 2024-08-17
[Merged by Bors] - feat(Order): A continuous monotone function `f` maps `sSup s` to `sSup (f '' s)`

dupuisf:
This PR shows several lemmas related to the PR title.

Note that we already have several lemmas in this direction, but they have very strong hypotheses (like `Monotone f` when `MonotoneOn f s` would do). 

We also remove several duplicated lemmas and replace them with deprecated aliases.

---


dupuisf:
> _Any_ function `f` maps `sSup s` to `f (sSup s)` 😁 Mind fixing the title?

I added "continuous" -- I think if people are interested in the details they can have a look at the code :-)

YaelDillies:
That's not really what I meant. Let me fix it

Mathlib/Topology/Order/Monotone.lean
```diff
@@ -28,12 +28,20 @@ variable [ConditionallyCompleteLinearOrder α] [TopologicalSpace α] [OrderTopol
 
 /-- A monotone function continuous at the supremum of a nonempty set sends this supremum to
 the supremum of the image of this set. -/
-theorem Monotone.map_sSup_of_continuousAt' {f : α → β} {A : Set α} (Cf : ContinuousAt f (sSup A))
-    (Mf : Monotone f) (A_nonemp : A.Nonempty) (A_bdd : BddAbove A := by bddDefault) :
+theorem MonotoneOn.map_sSup_of_continuousWithinAt' {f : α → β} {A : Set α}
+    (Cf : ContinuousWithinAt f A (sSup A))
+    (Mf : MonotoneOn f A) (A_nonemp : A.Nonempty) (A_bdd : BddAbove A := by bddDefault) :
     f (sSup A) = sSup (f '' A) :=
```
YaelDillies, line 34: For lemmas like this, stating the property of some set depending on some function property (continuity, monotonicity, convexity, differentiability...), I tend to suggest stating the required function property assumptions on a *superset* of the set in the conclusion. This is useful, because usually the set `A` you're trying to plug in is *not* the domain on which the function respect the property, which would rather be an open half-line, or the complement of a singleton or what not.

Here, concretely, I would do
```suggestion
theorem MonotoneOn.map_sSup_of_continuousWithinAt' {f : α → β} {A B : Set α} (hAB : A ⊆ B)
    (Cf : ContinuousWithinAt f B (sSup A))
    (Mf : MonotoneOn f B) (A_nonemp : A.Nonempty) (A_bdd : BddAbove A := by bddDefault) :
    f (sSup A) = sSup (f '' A) :=
```
or even
```suggestion
theorem MonotoneOn.map_sSup_of_continuousWithinAt' {f : α → β} {A B C : Set α} (hAB : A ⊆ B)
    (hAC : A ⊆ C) (Cf : ContinuousWithinAt f B (sSup A))
    (Mf : MonotoneOn f C) (A_nonemp : A.Nonempty) (A_bdd : BddAbove A := by bddDefault) :
    f (sSup A) = sSup (f '' A) :=
```
but that's probably overengineered

YaelDillies (commented):
Indeed I don't think it's worth duplicating lemmas when the sole difference is auto-params. If auto-params are causing issues, then we should improve the user experience rather than duplicating lemmas

Mathlib/Topology/Order/Monotone.lean
dupuisf: Fair enough, I went with the first option.

dupuisf:
> Indeed I don't think it's worth duplicating lemmas when the sole difference is auto-params. If auto-params are causing issues, then we should improve the user experience rather than duplicating lemmas

I removed the duplicates. I kept the version with the autoparam and added deprecated aliases.

Mathlib/Topology/Order/Monotone.lean
YaelDillies: Actually Bhavik disagrees that this is worth doing except for things like `ConvexOn` (because `ConvexOn.subset` is so annoying to use), and I am somewhat convinced by this argument. Would you mind rolling back?

dupuisf: No problem -- I confess I was a bit skeptical as well, when constructing a proof using `refine` I think it's a bit easier to deal with the subset thing afterwards.

YaelDillies (approved):
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15927, opened 2024-08-18
[Merged by Bors] - feat(SetTheory/Ordinal/Arithmetic): `a * b / (a * c) = b / c` and `(x * y) % (x * z) = x * (y % z)`

vihdzp:
Equalities like these prop up in the lemmas about the Cantor normal form.

---


urkud:
Thanks! :tada:
bors merge

PR #15928, opened 2024-08-18
[Merged by Bors] - feat(Data/List/Sigma): Lemmas relating `List.dlookup` and `List.map`

vihdzp:
---

Conflicts with #15951 but can be merged in either order.

These are admittedly very specific results, but I make use of `dlookup_map₁` in an upcoming PR.

I don't know if dependent versions of these can exist.

### DIFF
--- a/Mathlib/Data/List/Sigma.lean
+++ b/Mathlib/Data/List/Sigma.lean
@@ -204,6 +204,25 @@ theorem lookup_ext {l₀ l₁ : List (Sigma β)} (nd₀ : l₀.NodupKeys) (nd₁
   mem_ext nd₀.nodup nd₁.nodup fun ⟨a, b⟩ => by
     rw [← mem_dlookup_iff, ← mem_dlookup_iff, h] <;> assumption
 
+theorem dlookup_map₁ {γ δ} [DecidableEq γ]
+    {l : List (Σ _ : α, δ)} {f : α → γ} (hf : Function.Injective f) (a : α) :
+    (l.map fun x => ⟨f x.1, x.2⟩ : List (Σ _ : γ, δ)).dlookup (f a) = l.dlookup a := by
+  induction' l with b l IH
+  · rw [map_nil, dlookup_nil, dlookup_nil]
+  · rw [map_cons]
+    obtain h | h := eq_or_ne (f a) (f b.1)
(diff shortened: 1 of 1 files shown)


Mathlib/Data/List/Sigma.lean
```diff
@@ -204,6 +204,25 @@ theorem lookup_ext {l₀ l₁ : List (Sigma β)} (nd₀ : l₀.NodupKeys) (nd₁
...
+    (l.map fun x => ⟨f x.1, x.2⟩ : List (Σ _ : γ, δ)).dlookup (f a) = l.dlookup a := by
+  induction' l with b l IH
+  · rw [map_nil, dlookup_nil, dlookup_nil]
+  · rw [map_cons]
+    obtain h | h := eq_or_ne (f a) (f b.1)
+    · rw [h, hf h, dlookup_cons_eq, dlookup_cons_eq]
+    · rw [dlookup_cons_ne _ _ h, dlookup_cons_ne _ _ (fun he => (he ▸ h) rfl), IH]
+
+theorem dlookup_map₂ {γ δ} {l : List (Σ _ : α, γ)} {f : γ → δ} (a : α) :
+    (l.map fun x => ⟨x.1, f x.2⟩ : List (Σ _ : α, δ)).dlookup a = (l.dlookup a).map f := by
```
eric-wieser, line 218: This one can be dependently-typed, right?

vihdzp: Say the first list has type `List (Sigma δ)`. What would the type of `f` be? I don't think it makes much sense.

eric-wieser: It should be `f : ∀ a, γ a → δ a`, I think?

vihdzp: `List.map` takes in a non-dependent `α → β` argument, so I don't think that works.

Unfortunately I don't have any actual use case for this second lemma to test against. I'm just adding it for symmetry with the first one.

eric-wieser: It works just fine, see the commit I just pushed. While `f` is dependent, the function passed to `List.map` is not.

eric-wieser:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15929, opened 2024-08-18
[Merged by Bors] - feat(CategoryTheory/Monoidal): string diagram command

yuma-mizuno:
Add the `#string_diagram` command. For example,
```lean
#string_diagram MonoidalCategory.whisker_exchange
```
displays the string diagram for the exchange law of the left and right whiskerings, and 
```lean
variable {C : Type u} [Category.{v} C] [MonoidalCategory.{v} C] [BraidedCategory.{v} C] (X Y : C) in
#string_diagram (β_ X Y).hom
```
displays the string diagram for the braiding when importing the braided category file.

---


Mathlib/Tactic/Widget/StringDiagram.lean
```diff
@@ -361,4 +369,36 @@ open ProofWidgets
...
+#string_diagram f
+```
+-/
+syntax (name := stringDiagram) "#string_diagram " term : command
+
+@[command_elab stringDiagram, inherit_doc stringDiagram]
+def elabStringDiagramCmd : CommandElab := fun
+  | stx@`(#string_diagram $t:term) => do
+    let html ← runTermElabM fun _ => do
+      let e ← Term.elabTerm t none
```
Vtec234, line 393: I think you can use something like `realizeGlobalConstNoOverloadWithInfo` instead of `elabTerm` to get the constant without requiring `@`.

Vtec234 (changes requested):
Looks good, let's just try to avoid the `@`.

Vtec234 (approved):

joneugster:
LGTM, thanks!

maintainer merge

test/StringDiagram.lean
```diff
@@ -57,4 +59,30 @@ example {X₁ Y₁ X₂ Y₂ : C} (f : X₁ ⟶ Y₁) (g : X₂ ⟶ Y₂) : f 
...
+#string_diagram MonoidalCategory.whisker_exchange_assoc
+#string_diagram MonoidalCategory.pentagon
+#string_diagram MonoidalCategory.whiskerLeft_id
+#string_diagram left_triangle
+#string_diagram yang_baxter
+#string_diagram yang_baxter'
+#string_diagram yang_baxter''
+#string_diagram Category.assoc
+#string_diagram Functor.map_comp
+#string_diagram NatTrans.naturality
```
jcommelin, line 76: @yuma-mizuno @Vtec234 at the moment these tests might start "failing" without us noticing, because we don't record expected output. Can we adapt the tests somehow? Maybe use `#guard_msg` on a slightly more low-level command?

yuma-mizuno: I think it might be possible and reasonable to check the [penrose substance programs](https://penrose.cs.cmu.edu/docs/ref/substance/overview) the string diagram widget creates is what we expect. I'll try to write some test code like that.

joneugster (commented):
Thanks for adding the tests! I think this addresses @jcommelin 's suggestion fully, so I'll add this to the queue again

maintainer merge

semorrison:
bors merge

bryangingechen:
Please merge master, fix the build error and then put this PR back on the merge queue by commenting with "bors merge"
bors r-
bors d+

yuma-mizuno:
bors merge

PR #15930, opened 2024-08-18
[Merged by Bors] - feat(SetTheory/Ordinal/Exponential): `log b x = y ↔ x ∈ Set.Ico (b ^ y) (b ^ (succ y))`

vihdzp:
We give a nice characterization for `log b x = y`. We then use it to prove a more general version of `log_opow_mul_add`, alongside other lemmas.

---


Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -402,35 +417,40 @@ theorem log_mod_opow_log_lt_log_self {b o : Ordinal} (hb : 1 < b) (ho : o ≠ 0)
...
-    (hw : w < b ^ u) : log b (b ^ u * v + w) = u := by
-  have hne' := (opow_mul_add_pos (zero_lt_one.trans hb).ne' u hv w).ne'
-  by_contra! hne
-  cases' lt_or_gt_of_ne hne with h h
-  · rw [← lt_opow_iff_log_lt hb hne'] at h
-    exact h.not_le ((le_mul_left _ (Ordinal.pos_iff_ne_zero.2 hv)).trans (le_add_right _ _))
-  · conv at h => change u < log b (b ^ u * v + w)
-    rw [← succ_le_iff, ← opow_le_iff_le_log hb hne'] at h
-    exact (not_lt_of_le h) (opow_mul_add_lt_opow_succ hvb hw)
+theorem log_eq {b x : Ordinal} (hb : 1 < b) (hx : x ≠ 0) (y : Ordinal) :
```
b-mehta, line 420:
```suggestion
theorem log_eq_iff {b x : Ordinal} (hb : 1 < b) (hx : x ≠ 0) (y : Ordinal) :
```

Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -402,35 +417,40 @@ theorem log_mod_opow_log_lt_log_self {b o : Ordinal} (hb : 1 < b) (ho : o ≠ 0)
...
+    log b x = y ↔ x ∈ Set.Ico (b ^ y) (b ^ (succ y)) := by
```
b-mehta, line 421:
```suggestion
    log b x = y ↔ x ∈ Set.Ico (b ^ y) (b ^ succ y) := by
```

b-mehta, line 421: I'm inclined to say the rhs here should be two inequalities, which is also simp-normal-form for the rhs. Set.Ico is great when you want to describe a set, but the point here is that `x` satisfies two boundedness conditions. But this is optional

Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -402,35 +417,40 @@ theorem log_mod_opow_log_lt_log_self {b o : Ordinal} (hb : 1 < b) (ho : o ≠ 0)
...
+      apply mul_le_mul_left'
+      rw [succ_le_iff]
+      exact lt_opow_succ_log_self hb _
+  · intro h
+    exact mul_ne_zero (opow_ne_zero u (zero_lt_one.trans hb).ne') hv <|
+      left_eq_zero_of_add_eq_zero h
+
+theorem log_opow_mul {b v : Ordinal} (hb : 1 < b) (u : Ordinal) (hv : v ≠ 0) :
+    log b (b ^ u * v) = u + log b v := by
+  convert log_opow_mul_add hb hv (opow_pos u (zero_lt_one.trans hb)) using 1
```
b-mehta, line 447:
```suggestion
  simpa using log_opow_mul_add hb hv (opow_pos u (zero_lt_one.trans hb))
```
might work here, and I think it looks a little cleaner? 

Also is there some pos_of_gt you can write instead of the zero_lt_one.trans hb

b-mehta (approved):
bors d+

Mathlib/SetTheory/Ordinal/Exponential.lean
vihdzp: Sure, I'll change it to the inequality form.

Mathlib/SetTheory/Ordinal/Exponential.lean
vihdzp: `pos_of_gt` has typeclass hypotheses too strong (hopefully the `CanonicallyOrdered` refactor fixes that). However, we can use `bot_lt_of_lt` here and in other places where we don't care for syntactic equality.

YnirPaz (approved):
Looks good to me!

PR #15931, opened 2024-08-18
chore(SetTheory/Ordinal/Exponential): clean up `Ordinal.log` section

vihdzp:
We rewrite a few term-mode proofs in terms of equivalent but more readable tactics, unfold some trivial `simp`s, remove the very silly `log_of_not_one_lt_left`, and remove a redundant assumption in `log_mod_opow_log_lt_log_self`.

---

- [x] depends on: #15930


vihdzp:
Closing in favor of #17635

PR #15933, opened 2024-08-18
[Merged by Bors] - chore: correct `open` -> `open scoped`

grunweg:
If a file contains two subsequent lines `open X Y Z` and `open P Q R`, most likely one of these is meant to be `open scoped`:
as I understand it, this is a mistake during the port, where `open_locale` ought to have get translated as `open scoped` (and not `open`).

We fix a few of these: sometimes, the other `open`s are used now, in which case we merge these statements.
If some item occurs as both open and open scoped, we delete the latter (as `open` implies the effects of `open scoped`).

Let me be bold and label this as technical debt: as fixing a late (and mostly benign) mis-port.

---

### DIFF
--- a/Mathlib/Algebra/Lie/Weights/Basic.lean
+++ b/Mathlib/Algebra/Lie/Weights/Basic.lean
@@ -50,8 +50,7 @@ variable {K R L M : Type*} [CommRing R] [LieRing L] [LieAlgebra R L]
 
 namespace LieModule
 
-open Set Function LieAlgebra TensorProduct TensorProduct.LieModule
-open scoped TensorProduct
+open Set Function  TensorProduct TensorProduct.LieModule
 
 section notation_genWeightSpaceOf
 
--- a/Mathlib/Algebra/Lie/Weights/Basic.lean
+++ b/Mathlib/Algebra/Lie/Weights/Basic.lean
@@ -50,7 +50,7 @@ variable {K R L M : Type*} [CommRing R] [LieRing L] [LieAlgebra R L]
 
 namespace LieModule
 
-open Set Function  TensorProduct TensorProduct.LieModule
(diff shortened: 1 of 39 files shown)


Mathlib/Algebra/Lie/Weights/Basic.lean
```diff
@@ -50,8 +50,7 @@ variable {K R L M : Type*} [CommRing R] [LieRing L] [LieAlgebra R L]
 
 namespace LieModule
 
-open Set Function LieAlgebra TensorProduct TensorProduct.LieModule
-open scoped TensorProduct
+open Set Function  TensorProduct TensorProduct.LieModule
```
grunweg, line 53:
```suggestion
open Set Function TensorProduct LieModule

File: Mathlib/Analysis/Analytic/CPolynomial.lean [hunk 2]
```diff
@@ -42,10 +42,8 @@ We develop the basic properties of these notions, notably:
 variable {𝕜 E F G : Type*} [NontriviallyNormedField 𝕜] [NormedAddCommGroup E] [NormedSpace 𝕜 E]
   [NormedAddCommGroup F] [NormedSpace 𝕜 F] [NormedAddCommGroup G] [NormedSpace 𝕜 G]
 
-open scoped Classical
-open Topology NNReal Filter ENNReal
-
-open Set Filter Asymptotics
+open scoped Classical Topology
+open Set Filter Asymptotics NNReal ENNReal
```
grunweg, line 46: NNReal and ENNReal are required `open`s now

jcommelin (commented):
LGTM

bors d+

grunweg:
Thanks for the review!
bors merge

PR #15935, opened 2024-08-18
[Merged by Bors] - chore(Data/Nat/Choose/Sum): clean up

Parcly-Taxel:
---

### DIFF
--- a/Mathlib/Data/Nat/Choose/Sum.lean
+++ b/Mathlib/Data/Nat/Choose/Sum.lean
@@ -111,10 +111,10 @@ theorem choose_middle_le_pow (n : ℕ) : (2 * n + 1).choose n ≤ 4 ^ n := by
 theorem four_pow_le_two_mul_add_one_mul_central_binom (n : ℕ) :
     4 ^ n ≤ (2 * n + 1) * (2 * n).choose n :=
   calc
-    _ = (1 + 1) ^ (2 * n) := by norm_num [pow_mul]
+    4 ^ n = (1 + 1) ^ (2 * n) := by norm_num [pow_mul]
     _ = ∑ m ∈ range (2 * n + 1), (2 * n).choose m := by set_option simprocs false in simp [add_pow]
     _ ≤ ∑ m ∈ range (2 * n + 1), (2 * n).choose (2 * n / 2) := by gcongr; apply choose_le_middle
-    _ = _ := by simp
+    _ = (2 * n + 1) * choose (2 * n) n := by simp
 
 /-- **Zhu Shijie's identity** aka hockey-stick identity. -/
 theorem sum_Icc_choose (n k : ℕ) : ∑ m ∈ Icc k n, m.choose k = (n + 1).choose (k + 1) := by
@@ -163,7 +163,8 @@ theorem sum_powerset_neg_one_pow_card {α : Type*} [DecidableEq α] {x : Finset 
 
 theorem sum_powerset_neg_one_pow_card_of_nonempty {α : Type*} {x : Finset α} (h0 : x.Nonempty) :
     (∑ m ∈ x.powerset, (-1 : ℤ) ^ m.card) = 0 := by
-  classical rw [sum_powerset_neg_one_pow_card]
+  classical
+  rw [sum_powerset_neg_one_pow_card]
   exact if_neg (nonempty_iff_ne_empty.mp h0)
 
 variable {M R : Type*} [CommMonoid M] [NonAssocSemiring R]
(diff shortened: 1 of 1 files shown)


sgouezel:
You're mixing two unrelated things here, a cleanup (which is definitely useful), and the addition of a new lemma (which is already in #15270, so not really needed here). Could you remove the new lemma here, and then I will happily merge the cleanup PR?

Parcly-Taxel:
> You're mixing two unrelated things here, a cleanup (which is definitely useful), and the addition of a new lemma (which is already in #15270, so not really needed here). Could you remove the new lemma here, and then I will happily merge the cleanup PR?

Done.

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -33,114 +29,116 @@ variable [Semiring R] {x y : R}
...
 
 theorem four_pow_le_two_mul_add_one_mul_central_binom (n : ℕ) :
-    4 ^ n ≤ (2 * n + 1) * choose (2 * n) n :=
+    4 ^ n ≤ (2 * n + 1) * (2 * n).choose n :=
   calc
-    4 ^ n = (1 + 1) ^ (2 * n) := by norm_num [pow_mul]
-    _ = ∑ m ∈ range (2 * n + 1), choose (2 * n) m := by set_option simprocs false in simp [add_pow]
-    _ ≤ ∑ m ∈ range (2 * n + 1), choose (2 * n) (2 * n / 2) := by gcongr; apply choose_le_middle
-    _ = (2 * n + 1) * choose (2 * n) n := by simp
+    _ = (1 + 1) ^ (2 * n) := by norm_num [pow_mul]
```
sgouezel, line 114: In calc blocks, it's often more readable to keep the first and last entry instead of putting underscores. So
```suggestion
    4 ^ n = (1 + 1) ^ (2 * n) := by norm_num [pow_mul]
```

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -33,114 +29,116 @@ variable [Semiring R] {x y : R}
...
+    _ = ∑ m ∈ range (2 * n + 1), (2 * n).choose m := by set_option simprocs false in simp [add_pow]
+    _ ≤ ∑ m ∈ range (2 * n + 1), (2 * n).choose (2 * n / 2) := by gcongr; apply choose_le_middle
+    _ = _ := by simp
```
sgouezel, line 117: and
```suggestion
    _ = (2 * n + 1) * choose (2 * n) n := by simp
```

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -165,10 +163,8 @@ theorem sum_powerset_neg_one_pow_card {α : Type*} [DecidableEq α] {x : Finset
 
 theorem sum_powerset_neg_one_pow_card_of_nonempty {α : Type*} {x : Finset α} (h0 : x.Nonempty) :
     (∑ m ∈ x.powerset, (-1 : ℤ) ^ m.card) = 0 := by
-  classical
-    rw [sum_powerset_neg_one_pow_card, if_neg]
-    rw [← Ne, ← nonempty_iff_ne_empty]
-    apply h0
+  classical rw [sum_powerset_neg_one_pow_card]
```
sgouezel, line 166:
```suggestion
  classical
  rw [sum_powerset_neg_one_pow_card]
```
`classical` is considered a tactic, so it goes on its own line.

Mathlib/Data/Nat/Choose/Sum.lean
```diff
@@ -215,8 +209,7 @@ theorem sum_antidiagonal_choose_succ_mul (f : ℕ → ℕ → R) (n : ℕ) :
   simpa only [nsmul_eq_mul] using sum_antidiagonal_choose_succ_nsmul f n
 
 theorem sum_antidiagonal_choose_add (d n : ℕ) :
-    (Finset.sum (antidiagonal n) fun ij => (d + ij.2).choose d) =
-    (d + n).choose d + (d + n).choose (succ d) := by
+    (∑ ij ∈ antidiagonal n, (d + ij.2).choose d) = (d + n).choose d + (d + n).choose d.succ := by
```
sgouezel, line 212:
```suggestion
    (∑ ij ∈ antidiagonal n, (d + ij.2).choose d) = (d + n).choose d + (d + n).choose (d + 1) := by
```
We try to get away from `succ` as much as we can

sgouezel (commented):
Thanks!

PR #15936, opened 2024-08-18
[Merged by Bors] - feat: add Dockerfiles for lean and mathlib

zeramorphic:
Add Dockerfiles for lean and mathlib

---

Based off https://github.com/leanprover-community/mathlib3/tree/master/.docker. Also modifies the current gitpod Dockerfile so that leanblueprint works out-of-the-box.


YaelDillies (approved):
I have just updated [our Dockerhub images](https://hub.docker.com/repository/docker/leanprovercommunity/) using the two new scripts in this PR. Everything works as intended when I open the LeanAPAP repo (which points to the `leanprovercommunity/mathlib:gitpod` image).

@eric-wieser, can you make sure you too are happy with the changes?

.docker/gitpod/Dockerfile
```diff
@@ -1,12 +1,16 @@
...
 
-# gitpod doesn't support multiple FROM statements, (or rather, you can't copy from one to another)
-# so we just install everything in one go
 FROM ubuntu:jammy
 
 USER root
 
-RUN apt-get update && apt-get install sudo git curl git bash-completion python3 python3-requests -y && apt-get clean
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install sudo git curl git bash-completion texlive texlive-xetex latexmk graphviz graphviz-dev python3 python3-pip python3-requests -y && apt-get clean
```
eric-wieser, line 13: Why do we want `texlive` for mathlib? This makes the image quite a bit bigger.

zeramorphic: Maybe we should make two tags: one that includes a copy of leanblueprint and its dependencies (such as TeXLive), and one that does not.

YaelDillies: This is because we want it for blueprints. But maybe that should be an extra tag/dockerhub repo

zeramorphic:
I added a -4 suffix to the names of the images, and added automatic date tags.
image

YaelDillies:
I think we should get this in now. Our dockerhub is already using these new images

maintainer merge

scripts/docker_push.sh
```diff
@@ -0,0 +1,15 @@
+# This script attempts to build all the docker images, and then push then to the repository.
+# You'll need to have run `docker login` already;
+# check with scott.morrison@gmail.com for credentials.
```
bryangingechen, line 3: Maybe
```suggestion
# check with kim@tqft.net for credentials.
```
@kim-em Do you want your email here at all? Maybe just pointing to Zulip is enough?

bryangingechen:
Also, please merge `master` and document the two new scripts in `scripts/README.md`, per our new linter in #14539.

scripts/docker_push.sh
kim-em:
```suggestion
# check with Mathlib maintainers on Zulip for credentials.
```

kim-em:
bors merge

kim-em:
bors merge

PR #15938, opened 2024-08-18
feat: add autolabeling of PRs

adomani:
This PR introduces the automatic labeling of PRs.  The script is written in Lean and is used in CI.

Even though the PR appears very long, the actual code is only about 100 lines long.  The rest is taken by
* documentation;
* a whole file that simply adds the heuristic from [this Zulip post](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Auto-label.20PR.20topics/near/462530813) for labeling Mathlib PRs;
* a very extensive and verbose test file;
* a CI workflow that is mostly copied from the `new-contributor` workflow.

A very strong reason, in my opinion, to prefer this code to the shell-based one, is maintainability and extensibility.

Right now, the script only adds a label if exactly one applicable label is available.  This is very conservative, but the expectation is that it may change in the future.  As more elaborate heuristics may be developed, the Lean-script should adapt much more easily than the bash script.

---


grunweg:
@adomani I'm pretty sure the glue code in the .yml file can be rewritten as a Lean script. Would you like me to do so?

(I will try to get to reviewing this PR; I think it's great to have this automation. But life/admin is catching up with me, so I don't want to promise any swift review.)

Mathlib/Tactic/Autolabels/EnvExtension.lean
```diff
@@ -0,0 +1,248 @@
...
+It returns the sorted array of the applicable labels with no repetitions.
+-/
+def produceLabels (env : Environment) (gitDiffs : String) : Array String :=
+  let grouped := labelsToFiles env gitDiffs
+  ((grouped.toArray.map Prod.fst).filter (· != "")).qsort (· < ·)
+
+/--
+`produce_labels "A/B/C.lean⏎D/E.lean"` takes as input a string, assuming that it is a
+line-break-separated list of paths to files.
+It uses the paths to check if any of the labels in the environment is applicable.
```
jcommelin, line 222:
```suggestion
It uses the paths to check if any of the labels in the environment are applicable.
```

Mathlib/Tactic/Autolabels/EnvExtension.lean
```diff
@@ -0,0 +1,248 @@
...
+
+/--
+`produceLabels env gitDiffs` takes as input an `Environment` `env` and a string `gitDiffs`.
+It assumes that `gitDiffs` is a line-break-separate list of paths to files.
+It uses the paths to check if any of the labels in the environment is applicable.
+It returns the sorted array of the applicable labels with no repetitions.
+-/
+def produceLabels (env : Environment) (gitDiffs : String) : Array String :=
+  let grouped := labelsToFiles env gitDiffs
+  ((grouped.toArray.map Prod.fst).filter (· != "")).qsort (· < ·)
```
jcommelin, line 217: Is this sorting alphabetically?

jcommelin (commented):
Should this take the size of the git diff into account, to "weigh" the labels if there are multiple candidates? Or maybe that's better left to a follow-up version.

Mathlib/Tactic/Autolabels/EnvExtension.lean
adomani: Yes, I think so: the keys in `grouped` are `String`s, so `Prod.fst` returns an array of `String`s and then `qsort` uses `<` on `String`s which is alphabetical.

adomani:
> Should this take the size of the git diff into account, to "weigh" the labels if there are multiple candidates? Or maybe that's better left to a follow-up version.

I intentionally went for a minimally invasive heuristic that would only label a PR when it was (hopefully) uncontroversial.  I am slightly more in favour of seeing how this minimalistic setup works.

If it ends up being the case that people feel like the auto-labels should be added more aggressively, it will be easy to add more and more heuristics to the mix.

.github/workflows/autolabel_action.yaml
```diff
@@ -0,0 +1,55 @@
...
+          set -o pipefail
+          curl -sSfL https://github.com/leanprover/elan/releases/download/v3.1.1/elan-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          ./elan-init -y --default-toolchain none
+          echo "$HOME/.elan/bin" >> "${GITHUB_PATH}"
+
+      - name: get cache for `Lean.Elab.Command`
+        run: |
+          git checkout master
+          git checkout -
+          lake exe cache get Mathlib/Tactic/Linter/HashCommandLinter.lean
```
joneugster, line 32: The name does not match the command.

.github/workflows/autolabel_action.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Auto-label script
+        run: |
+          printf '\nproduce_labels! "git"\n\nproduce_labels "git"\n' >> Mathlib/Tactic/Autolabels/Labels.lean
+          lake build Mathlib.Tactic.Autolabels.Labels
+          labels="$(lake build Mathlib.Tactic.Autolabels.Labels |
+            sed -n 's=.*Applicable labels: *\(.*\)=\1=p')"
```
joneugster, line 39: This approach feels a bit hacky.

I have a few questions about this step, to which I think the explanation could be added as comments here:

- the `printf >>` hack is to avoid the noise on building normally, right?
- what's the reason to do 'lake build' twice? Has that also to do with the output?
- do you even need `produce_labels!` or are you just `sed`ing for the output of `produce_labels`?

Mathlib/Tactic/Autolabels/EnvExtension.lean
```diff
@@ -0,0 +1,248 @@
...
+The prefix `t-` in front of `meta` is automatically added by `add_label`.
+
+Look at the documentation for `add_label` for further shortcuts.
+
+## Further commands
+
+The file also defines the commands `check_labels` and `produce_labels(!)? str`.
+These are mostly intended as "debugging commands", as they display what `Label`s are
+currently present in the environment and to test-run the automatic application of the labels
+to user-input/`git diff` output.
```
joneugster, line 58: This seems outdated, given that the CI actually uses `produce_labels`.

adomani (commented):
Jon, thank you so much for the review: I addressed your comments, mostly adding comments to the PR!

joneugster (approved):
LGTM

maintainer merge

.github/workflows/autolabel_action.yaml
```diff
@@ -0,0 +1,57 @@
...
+          curl -sSfL https://github.com/leanprover/elan/releases/download/v3.1.1/elan-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          ./elan-init -y --default-toolchain none
+          echo "$HOME/.elan/bin" >> "${GITHUB_PATH}"
+
+      - name: get cache for `Lean.Elab.Command` # although we fetch `HashCommandLinter` since cache is `Mathlib`-based
+        run: |
+          git checkout master
+          git checkout -
+          lake exe cache get Mathlib/Tactic/Linter/HashCommandLinter.lean
+
```
semorrison, line 33: This makes no sense, there is no notion of "cache" for `Lean.Elab.Command`. It is all in the `lean` binary!

adomani: Ah, great!  So, I'll remove this step!

.github/workflows/autolabel_action.yaml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          # the checkout dance, to avoid a detached head
+          git checkout master
+          git checkout -
+
+          # append the "noisy" commands, so we can read off their values and not fail CI
+          printf '\nproduce_labels! "git"\n\nproduce_labels "git"\n' >> Mathlib/Tactic/Autolabels/Labels.lean
+
+          # a first run to store the values in the CI logs -- useful for debugging
+          lake build Mathlib.Tactic.Autolabels.Labels
```
semorrison, line 38: Could you use `tee` to avoid running twice?

Mathlib/Tactic/Autolabels/EnvExtension.lean
```diff
@@ -0,0 +1,248 @@
...
+Ultimately, this produces a `HashMap String String` that takes a string representing a path
+to a modified file and returns the corresponding GitHub label.
+
+Since the labels are independent of the modified file, we define a `PersistentEnvExtension`
+that keeps track of an array of `Label`s.
+
+## The `Label` structure
+
+A `Label` is a structure containing
+* the GitHub label (as a `String` -- `Label.label`);
```
semorrison, line 26: What does `--` signify here? I don't understand.

semorrison:
Could you explain the design more either here or on zulip? I don't understand why an environment extension would be involved at all here.

Mathlib/Tactic/Autolabels/EnvExtension.lean
joneugster: Maybe the following notation is more intuitive?

```
* `Label.label : String`; the Github label
```

.github/workflows/autolabel_action.yaml
adomani: I used `tee`, but I honestly find it less clear.

Mathlib/Tactic/Autolabels/EnvExtension.lean
adomani: Jon's interpretation is indeed what I had in mind, but it looks like it was confusing.  I simply remove `-- ...`, since the source code is right below anyway.

adomani:
Re the environment extension: I thought that they were quite "light", especially when they are basically not imported anywhere, as should be the case here.  It seemed to give an easy user experience, managing in the background all that is needed: keeping track of what has been defined and producing a well-formatted output.

I'm happy with a different proposal, though.

.github/workflows/autolabel_action.yaml
joneugster: What's less clear about it? Looks good to me.

adomani: I'm not a big fan of these redirections, but I agree that this is almost as straightforward as it gets! 😊

joneugster:
I have a counterproposal not using env. extensions, which might be cleaner. Might only create a PR with it tomorrow, so let me add "WIP" for the moment.

joneugster:
I've created a modified version at

- [ ] #17062

which does not use any env extensions and seems to be simpler. Only one of the two should be merged.

joneugster:
I'm closing this PR in favour of #17062 which is about to be merged :D

PR #15942, opened 2024-08-18
chore: move style linters into their own file; document all current linters

grunweg:
Just to show the overall direction: this PR is just a tracking PR; it is easier to re-do the remaining pieces individually.

- [x] depends on: #19529 (and needs a merge afterwards)

---

The documenting of items should move into Tactic.Linter...


joneugster:
I've got the impression that this is more a tracking PR and not ready for review. Please remove the "WIP" label again if that's incorrect and this should go on the queue.

grunweg:
I might say "different parts have been split into smaller PRs", but I guess that amounts to the same...
I'll merge master; perhaps the remaining PR can land as-is, or I'll split off another change.

joneugster:
I'm a bit confused by the state of this PR, also because I can't tell how much would be addressed by resolving the merge-conflict. Without looking at the queue to see what you already did, all the "stylistic changes" should probably be split into a bunch of `easy` PRs.

And then probably one PR per linter moving it in it's own file and adding documentation for it would seem reasonable

grunweg:
All style linters are documented now: this PR can be closed.

PR #15944, opened 2024-08-18
[Merged by Bors] - feat(Order/SuccPred): `BddAbove.exists_isGreatest_of_nonempty`

pechersky:
when `IsSuccArchimedean`

---


Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -1401,3 +1401,35 @@ lemma SuccOrder.forall_ne_bot_iff
   rw [← Nat.succ_pred_eq_of_pos hj]
   simp only [Function.iterate_succ', Function.comp_apply]
   apply h
+
+section IsLeast
+
+lemma BddAbove.exists_isGreatest_of_nonempty {X : Type*} [LinearOrder X] [SuccOrder X]
+    [IsSuccArchimedean X] {S : Set X} (hS : BddAbove S) (hS' : S.Nonempty) :
+    ∃ x, IsGreatest S x := by
```
urkud, line 1409: @YaelDillies Are these the right typeclass assumptions?

urkud:
LGTM but I'm not sure if it can be generalized to some other `*Order` class we have, so delegating to @YaelDillies .
bors d=@YaelDillies

Mathlib/Order/SuccPred/Basic.lean
YaelDillies: I guess the right assumption is something like "locally cowell-founded", meaning "`Iic x` is cowell-founded for all `x`". We don't have that in mathlib, so the current generality seems okay.

Note however that `LinearOrder` shouldn't be necessary since the recent refactor on `SuccOrder`
```suggestion
lemma BddAbove.exists_isGreatest_of_nonempty {X : Type*} [Preorder X] [SuccOrder X]
    [IsSuccArchimedean X] {S : Set X} (hS : BddAbove S) (hS' : S.Nonempty) :
    ∃ x, IsGreatest S x := by
```

pechersky: I can't generalize to `Preorder` because I use `Order.le_succ_iff_eq_or_le` in the inductive step proof of `Succ.rec`.

YaelDillies: The end result is still true, though (although I've noticed you also need to assume that the set is directed). Can you try using another lemma?

YaelDillies: @Command-Master, didn't you prove that the assumptions above imply `SemilatticeSup X`?

pechersky: Should this now use results from #16559?

YaelDillies: There was no new result in #16559. Do you mean #16272?

pechersky: Yes, I now see the results got split

pechersky: I've left a TODO to come back to this after that PR.

YaelDillies: Sorry, I won't merge this for now because I suspect you don't even need the lemmas in the current PR. Please first coordinate with @Command-Master to get #16272 (or rather the preliminary part) merged.

pechersky: Can you take a look at https://github.com/leanprover-community/mathlib4/pull/16619/files#diff-e9c6f03b09ef6fa37d4bb4f98167c69a8165fcbbd8f189451b15fafbf798f65cR113 please? It's the point of use.

pechersky: I tried to bring in the relevant lemmas from that PR, but I still don't have a proof. I also can't assume that my set is `OrdConnected`. I need help with this.

YaelDillies:
I am not super happy with the generality of this PR, but since @Command-Master's PRs that would unlock the greater generality are currently stalled, let's press on.

bors merge

PR #15945, opened 2024-08-18
[Merged by Bors] - chore(Equiv/Basic): update `simps` lemmas

urkud:
- add lemmas for `Equiv.sigmaProdDistrib`;
- `Equiv.prodCongrLeft'_symm_apply` is no longer a `simp` lemma; I think that we shouldn't introduce `▸`s in `simp`;
- instead, `Equiv.prodCongrLeft'_symm_apply_apply` is a `simp` lemma.

---


mattrobball:
Keys look ok but let's check the impact since this is low level.

mattrobball:
!bench

mattrobball:
bors merge

PR #15922, opened 2024-08-17
[Merged by Bors] - chore(Algebra/Prod): add `Prod.mk_one_one`

urkud:
- Normalize `(1, 1) ` and `(0, 0)` to `1` and `0`, respectively.
- Add `Polynomial.mul_coeff_one`

Moves:
- PerfectClosure.mk_zero_zero -> PefectClosure.mk_zero
- PerfectClosure.mk_zero -> PerfectClosure.mk_zero_right

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15949, opened 2024-08-18
[Merged by Bors] - feat(Topology/ENat): multiplication is continuous

urkud:
Also prove that subtraction is continuous away from `(⊤, ⊤)`.

---

- [ ] depends on: #15862 
- [x] depends on: #15922 


sgouezel:
bors r+
Thanks!

PR #15950, opened 2024-08-18
feat(Order/SuccPred): transfer SuccOrder, IsSuccArchimedean

pechersky:
across monotone equivs

---


vihdzp:
Note that there's currently a large refactor for `SuccOrder` underway in #15881.

pechersky:
> Note that there's currently a large refactor for `SuccOrder` underway in #15881.

I hope that these transfer constructions, especially since they are over linear orders, should be easy to include in the refactor if it proceeds.

Mathlib/Order/Hom/Basic.lean
```diff
@@ -522,6 +522,11 @@ protected def withTopMap (f : α →o β) : WithTop α →o WithTop β :=
 
 end OrderHom
 
+-- See note [lower instance priority]
+instance (priority := 100) OrderHomClass.toOrderHomClassOrderDual [LE α] [LE β]
+    [FunLike F α β] [OrderHomClass F α β] : OrderHomClass F αᵒᵈ βᵒᵈ where
```
YaelDillies, line 527: This can't be an instance because `α` and `β` are out-params in `OrderHomClass`

Mathlib/Order/Hom/Basic.lean
```diff
@@ -1272,6 +1277,11 @@ end BoundedOrder
 
 end LatticeIsos
 
+-- See note [lower instance priority]
+instance (priority := 100) OrderIsoClass.toOrderIsoClassOrderDual [LE α] [LE β]
+    [EquivLike F α β] [OrderIsoClass F α β] : OrderIsoClass F αᵒᵈ βᵒᵈ where
```
YaelDillies, line 1282: Here however `α` and `β` are not out-params in `OrderIsoClass`? @Vierkantor, is this an oversight?

Mathlib/Order/SuccPred/Basic.lean
```diff
@@ -1202,3 +1202,32 @@ instance [hα : Nonempty α] : IsEmpty (SuccOrder (WithBot α)) :=
...
+@[reducible] protected def SuccOrder [SuccOrder X] (f : F) : SuccOrder Y where
+  succ y := f (Order.succ (EquivLike.inv f y))
+  le_succ y := by
+    obtain ⟨x, rfl⟩ := EquivLike.surjective f y
+    simpa using (map_le_map_iff f).mpr (Order.le_succ x)
+  max_of_succ_le {y} hy := by
+    obtain ⟨x, rfl⟩ := EquivLike.surjective f y
+    simp only [EquivLike.inv_apply_apply, map_le_map_iff] at hy
+    have := Order.max_of_succ_le hy
+    exact (OrderIso.isMax_apply (f : X ≃o Y)).mpr this
```
YaelDillies, line 1220: Does this not work?
```suggestion
    exact (OrderIso.isMax_apply (f : X ≃o Y)).mpr (Order.max_of_succ_le hy)
```

Mathlib/Order/Hom/Basic.lean
pechersky: It worked at my site of use here: https://github.com/leanprover-community/mathlib4/pull/15950/files#diff-6c24c5f799f540195c932d072bf980aa91eaf384ac4fecb2d64e56042aa23791R250

pechersky: Ah, I see. You're saying for `OrderHom` it won't work, while I used `OrderIso`

YaelDillies:
I think your new hom class pattern should be discussed on Zulip. I am of the opinion that it is a bad idea, but I would like feedback from more people

pechersky:
Would you prefer the plain `OrderIso` based transfer I did in #16490?

YaelDillies:
Yeah, that would be much better

pechersky:
Closed in favor of #16785.

PR #15951, opened 2024-08-18
[Merged by Bors] - feat(Data/List/Sigma): Lemmas relating `List.dlookup` and `List.append`

vihdzp:
---

Conflicts with #15928 but can be merged in any order.

Both of these lemmas are used in my development of the Cantor normal form.

### DIFF
--- a/Mathlib/Data/List/Sigma.lean
+++ b/Mathlib/Data/List/Sigma.lean
@@ -282,16 +282,15 @@ theorem perm_lookupAll (a : α) {l₁ l₂ : List (Sigma β)} (nd₁ : l₁.Nodu
     (p : l₁ ~ l₂) : lookupAll a l₁ = lookupAll a l₂ := by
   simp [lookupAll_eq_dlookup, nd₁, nd₂, perm_dlookup a nd₁ nd₂ p]
 
-theorem dlookup_append {l₀ : List (Sigma β)} {a : α} (l₁ : List (Sigma β)) :
-    (l₀ ++ l₁).dlookup a = (l₀.dlookup a).or (l₁.dlookup a) := by
-  induction' l₀ with b l₀ IH
-  · rfl
-  · rw [cons_append]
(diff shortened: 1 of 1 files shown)


Mathlib/Data/List/Sigma.lean
```diff
@@ -282,6 +282,24 @@ theorem perm_lookupAll (a : α) {l₁ l₂ : List (Sigma β)} (nd₁ : l₁.Nodu
     (p : l₁ ~ l₂) : lookupAll a l₁ = lookupAll a l₂ := by
   simp [lookupAll_eq_dlookup, nd₁, nd₂, perm_dlookup a nd₁ nd₂ p]
 
+theorem dlookup_append_of_not_mem_left {l₀ : List (Sigma β)} {a : α} (h : a ∉ l₀.keys)
+    (l₁ : List (Sigma β)) : (l₀ ++ l₁).dlookup a = l₁.dlookup a := by
```
eric-wieser, line 286: Is this true more generally as
```suggestion
theorem dlookup_append_of_not_mem_left {l₀ : List (Sigma β)} {a : α}
    (l₁ : List (Sigma β)) : (l₀ ++ l₁).dlookup a = l₀.dlookup a <|> l₁.dlookup a := by
```

vihdzp: Yeah! That's a much nicer way to state it.

eric-wieser (approved):
I pushed some tweaks, from least to most opinionated:
* Arguments to lemmas about `=` are usually explicit
* `induction` is now preferred to `induction'`
* `iterate` makes the proof a little harder to inspect, and `rfl` is usually worse than finding the right lemma if you're already using other lemmas
* `Decidable.eq_or_ne` saves needing `Classical.choice`, but no one really cares

Feel free to revert either of the last two if you disagree with them

bors d+

vihdzp:
I like this statement a lot better.

bors r+

PR #15953, opened 2024-08-19
[Merged by Bors] - chore(SetTheory/Ordinal/Principal): clean up formatting

vihdzp:
This PR does the following:

- Clean up some term mode sorcery.
- Use the `ω` notation throughout.
- Remove unneeded parentheses.
- Indent proofs correctly.

---

### DIFF
--- a/Mathlib/SetTheory/Ordinal/Principal.lean
+++ b/Mathlib/SetTheory/Ordinal/Principal.lean
@@ -24,16 +24,15 @@ We define principal or indecomposable ordinals, and we prove the standard proper
   of `fun x ↦ ω ^ x`.
 -/
 
-universe u v w
-
-noncomputable section
+universe u
 
 open Order
 
 namespace Ordinal
 
--- Porting note: commented out, doesn't seem necessary
---local infixr:0 "^" => @pow Ordinal Ordinal Ordinal.hasPow
+section Arbitrary
+
+variable {op : Ordinal → Ordinal → Ordinal}
 
 /-! ### Principal ordinals -/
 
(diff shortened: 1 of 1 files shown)


Mathlib/SetTheory/Ordinal/Principal.lean
```diff
@@ -42,41 +41,45 @@ namespace Ordinal
...
 almost exclusively used for additive and multiplicative principal ordinals.
 
-For simplicity, we break usual convention and regard 0 as principal. -/
+For simplicity, we break usual convention and regard `0` as principal. -/
 def Principal (op : Ordinal → Ordinal → Ordinal) (o : Ordinal) : Prop :=
   ∀ ⦃a b⦄, a < o → b < o → op a b < o
 
-theorem principal_iff_principal_swap {op : Ordinal → Ordinal → Ordinal} {o : Ordinal} :
-    Principal op o ↔ Principal (Function.swap op) o := by
+theorem principal_swap_iff {o : Ordinal} : Principal (Function.swap op) o ↔ Principal op o := by
```
grunweg, line 48: You're flipping and renaming a lemma. Can you add a deprecated alias for the old lemma?

grunweg (commented):
I am not an expert on this part of the library - but these changes all look reasonable (and much better to my eye). Thanks!
maintainer merge

Mathlib/SetTheory/Ordinal/Principal.lean
vihdzp: The old lemma (and actually this new one too) weren't used anywhere. My reason for flipping it is that iff statements like these should have the LHS be simpler.

But sure, I can still add the label.

semorrison:
bors merge

PR #15934, opened 2024-08-18
test style linters / reviewdog in a clean branch

bryangingechen:
Not to be merged! A cleaner test to demonstrate a possible issue with the linters introduced in #15051 and found [here](https://github.com/leanprover-community/mathlib4/pull/15781#discussion_r1719779686).

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15954, opened 2024-08-19
[Merged by Bors] - fix(lint-style): make sure Python style errors fail the linting step; fix --fix mode

grunweg:
We need to pass any --fix flag down to `lint-style.py`; this detail got lost along the rewrite.

While at it, tweak the formatting of error messages slightly: avoid an extra newline after the Python output,
and pluralise errors more correctly.

---


bryangingechen:
Thanks for looking into this! It looks like reviewdog didn't trigger on 24514a4 ([log](https://github.com/leanprover-community/mathlib4/actions/runs/10445126543/job/28920613811#step:7:232)). Compare [this log](https://github.com/leanprover-community/mathlib4/actions/runs/10403495420/job/28810093271#step:7:232) where reviewdog did run.

grunweg:
Good catch - there was another hidden bug (the --fix option was not percolating all the way down to lint-style.py).
I tested it locally, not it worked. Does it also work for you?

bryangingechen:
Yep, I merged this branch into #15934 and reviewdog is picking up the errors now! Great work!

bors r+

PR #15957, opened 2024-08-19
[Merged by Bors] - chore: deprecate Data/ByteArray

kim-em:
https://github.com/leanprover-community/batteries/pull/851 provides a replacement in Batteries. It is not an exact replacement, so I have not given suggestion in the deprecation statements, and instead added an pointer to the new file in the module doc.


grunweg (commented):
Have you tried to check if there are declarations which match closely? Assuming you have/a maintainer thinks this is not necessary: looks reasonable, thanks!
maintainer merge

Mathlib/Data/ByteArray.lean
```diff
@@ -43,6 +49,7 @@ end ByteSliceT
 def ByteArray.toSliceT (arr : ByteArray) : ByteSliceT := ⟨arr, 0⟩
 
 /-- A byte slice, given by a backing byte array, and an offset and length. -/
+@[deprecated (since := "2024-08-19")]
```
eric-wieser, line 52: I think this one is basically a drop-in replacement, semantically.
```suggestion
@[deprecated Batteries.ByteSubarray (since := "2024-08-19")]
```

Mathlib/Data/ByteArray.lean
```diff
@@ -9,11 +9,16 @@ import Mathlib.Init
...
+
+This entire file has been deprecated on 2024-08-19 in favour of `ByteSubarray` in Batteries.
 -/
 
+set_option linter.deprecated false
+
 namespace Nat
 
 /-- A well-ordered relation for "upwards" induction on the natural numbers up to some bound `ub`. -/
+@[deprecated (since := "2024-08-19")]
```
jcommelin, line 21:
```suggestion
@[deprecated "deprecated in favour of `ByteSubarray` in Batteries"  (since := "2024-08-19")]
```
Can we leave a message like this?

eric-wieser: I don't think that makes any sense here, since `Up` is unrelated to `ByteSubarray`.

Mathlib/Data/ByteArray.lean
```diff
@@ -27,6 +32,7 @@ def upRel (ub : Nat) : WellFoundedRelation Nat := ⟨Up ub, Up.WF ub⟩
 end Nat
```
eric-wieser, line 32: Do we not want to deprecated `upRel` as well?

semorrison: I decided it wasn't necessary it individually deprecate things, once the types they talked about were deprecated... This is all unused code.

eric-wieser: This definition doesn't talk about any deprecated types though

Mathlib/Data/ByteArray.lean
```diff
@@ -80,6 +90,7 @@ def ByteArray.toSlice (arr : ByteArray) : ByteSlice := ⟨arr, 0, arr.size⟩
 
 /-- Convert a string of assumed-ASCII characters into a byte array.
 (If any characters are non-ASCII they will be reduced modulo 256.) -/
+@[deprecated (since := "2024-08-19")]
```
eric-wieser, line 93: @digama0, is there a replacement for this elsewhere that you know of?

digama0: No. AFAIK the only reason it's being deprecated is because the semantics are somewhat weird in the non-ascii case.

eric-wieser: Right, the semantics are pretty strange indeed; but if a `String.toAscii? : Option ByteArray` existed, it would probably be sane to mention in this message.

eric-wieser:
bors d+

CI is probably unhappy for unrelated reasons.

semorrison:
bors merge

PR #15958, opened 2024-08-19
[Merged by Bors] - chore: remove >1 year deprecations

kim-em:
This isn't everything with >1year deprecation ages: unfortunately lots is still actually in use!


vihdzp:
It'd be nice to have a list of the old deprecated theorems in current use, so people can get to cleaning them up.

semorrison:
> It'd be nice to have a list of the old deprecated theorems in current use, so people can get to cleaning them up.

The answer here is basically: "any lemma with `nthLe` in the name".

semorrison:
@Parcly-Taxel, thanks for taking a look here. It would actually make more sense if you made that commit into a separate PR:
1. I could then immediately merge it, instead of us needing to go find another reviewer
2. It will keep this PR simpler, as it is just deleting things, not making significant changes.

Parcly-Taxel:
> @Parcly-Taxel, thanks for taking a look here. It would actually make more sense if you made that commit into a separate PR:
> 
>     1. I could then immediately merge it, instead of us needing to go find another reviewer
> 
>     2. It will keep this PR simpler, as it is just deleting things, not making significant changes.

#15965

Mathlib/Data/List/Cycle.lean
```diff
@@ -273,106 +274,93 @@ theorem next_get : ∀ (l : List α) (_h : Nodup l) (i : Fin l.length),
...
-        refine h _ _ hn Nat.succ_pos' ?_
-        simpa using H
-      · intro H
-        suffices n.succ.succ = 1 by simpa
-        rw [nodup_iff_nthLe_inj] at h
-        refine h _ _ hn (Nat.succ_lt_succ Nat.succ_pos') ?_
-        simpa using H
-
-set_option linter.deprecated false in
+theorem prev_get (l : List α) (_h : Nodup l) (i : Fin l.length) :
```
mattrobball, line 277: I would prefer that we explicitly turn off the linter and comment about how it is failing at `h` than work around the issues.

semorrison: @Parcly-Taxel pushed these changes to my PR ... :-(

semorrison: I've deleted Jeremy's commit from this PR, so this comment now belongs on https://github.com/leanprover-community/mathlib4/pull/15965.

semorrison: (And I've implemented the suggestion there.)

grunweg:
Thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #15959, opened 2024-08-19
[Merged by Bors] - feat(Topology/Order): add `isOpen_sSup_iff`

urkud:
---


eric-wieser (commented):
Can you add the `isClosed` version too?

eric-wieser (approved):
bors merge

Thanks!

PR #15960, opened 2024-08-19
[Merged by Bors] - feat: `Fintype (LTSeries α)` if `Fintype α`

Parcly-Taxel:
From the Carleson project.

---

### DIFF
--- a/Mathlib/Order/RelSeries.lean
+++ b/Mathlib/Order/RelSeries.lean
@@ -8,7 +8,6 @@ import Mathlib.Data.Fintype.Card
 import Mathlib.Data.Fintype.Pi
 import Mathlib.Data.Fintype.Sigma
 import Mathlib.Data.Rel
-import Mathlib.Tactic.Abel
 
 /-!
 # Series of a relation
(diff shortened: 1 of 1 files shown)


Mathlib/Order/RelSeries.lean
```diff
@@ -692,6 +690,21 @@ def mk (length : ℕ) (toFun : Fin (length + 1) → α) (strictMono : StrictMono
   toFun := toFun
   step i := strictMono <| lt_add_one i.1
 
+/-- An injection from the type of strictly monotone functions with limited length to `LTSeries`. -/
+def inj (n : ℕ) : {f : (l : Fin n) × (Fin (l + 1) → α) // StrictMono f.2} ↪ LTSeries α where
```
jcommelin, line 694: This name is maybe a bit too generic? What do you think of `injStrictMono` or something like that?

Parcly-Taxel: I've changed it to `injStrictMono`.

jcommelin (commented):
Thanks :tada:

bors merge

PR #15961, opened 2024-08-19
chore(SetTheory/Ordinal/Artihmetic): clean up `pred` API

vihdzp:
This PR does the following:
- Use `o ∈ range succ` instead of the full (and backwards) spelling `∃ a, o = succ a`
- Add the missing convenience lemmas `pred_eq` and `pred_one`
- Golf proofs, or otherwise remove term mode sorcery
- Remove the unnecessary `Set` prefix in `Set.range` while we're at it

---


vihdzp:
Closing since I plan on generalizing `pred` and deprecating this definition instead.

PR #15965, opened 2024-08-19
[Merged by Bors] - chore: remove `nthLe`

Parcly-Taxel:
From #15958.


grunweg:
Great PR, I look forward to this landing! Let me label this with awaiting-CI, as this seems to on its pass to passing.

semorrison:
Likely whenever the linter complains, the right thing to do is just remove the `simp` attribute. There are overlapping `simp` lemmas in Lean 4.

Mathlib/Data/List/Range.lean
```diff
@@ -27,15 +27,6 @@ namespace List
 
 variable {α : Type u}
 
-set_option linter.deprecated false in
-@[simp]
-theorem nthLe_range' {n m step} (i) (H : i < (range' n m step).length) :
-    nthLe (range' n m step) i H = n + step * i := get_range' i H
```
semorrison, line 33: I think these can't be deleted yet: they need to be `@[deprecated]` first.

Parcly-Taxel: The continued existence of `nthLe` and other lemmas depending on it is an embarrassment to mathlib and the sooner they are purged the better. I don't think we can afford to wait a whole year to deprecate them.

semorrison:
@Parcly-Taxel, I pushed a commit implementing @mattrobball's request from the original PR.

semorrison:
bors merge

PR #15968, opened 2024-08-19
[Merged by Bors] - chore: no need to pass autoParam to simp or dsimp

nomeata:
and with https://github.com/leanprover/lean4/pull/4154 it actually
complains, so let's remove them now.


TwoFX (approved):
Thanks!
bors r+

PR #11909, opened 2024-04-04
[Merged by Bors] - chore(Data/Real/Sqrt): delete commented-out incomplete code

Smaug123:
Delete the currently-unused `sqrtAux`, which is a doomed attempt at a computable definition of `Real.sqrt`.

---

Years-old discussion at https://leanprover.zulipchat.com/#narrow/stream/116395-maths/topic/Computable.20real.2Esqrt . The commented-out theorem is in fact false.

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15973, opened 2024-08-19
[Merged by Bors] - chore: replace last uses of `lambda` instead of `fun`

grunweg:
Split out from #15896.

---

### DIFF
--- a/Mathlib/Data/Real/Sqrt.lean
+++ b/Mathlib/Data/Real/Sqrt.lean
@@ -115,15 +115,14 @@ noncomputable def sqrt (x : ℝ) : ℝ :=
 @[inherit_doc]
 prefix:max "√" => Real.sqrt
 
-/- quotient.lift_on x
-  (λ f, mk ⟨sqrt_aux f, (sqrt_aux_converges f).fst⟩)
-  (λ f g e, begin
-    rcases sqrt_aux_converges f with ⟨hf, x, x0, xf, xs⟩,
-    rcases sqrt_aux_converges g with ⟨hg, y, y0, yg, ys⟩,
-    refine xs.trans (eq.trans _ ys.symm),
-    rw [← @mul_self_inj_of_nonneg ℝ _ x y x0 y0, xf, yg],
-    congr' 1, exact quotient.sound e
-  end)-/
+/- Quotient.liftOn x
+  (fun f ↦ mk ⟨sqrt_aux f, (sqrt_aux_converges f).fst⟩)
+  (fun f g e := by
+    rcases sqrt_aux_converges f with ⟨hf, x, x0, xf, xs⟩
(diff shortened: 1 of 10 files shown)


Mathlib/Data/Real/Sqrt.lean
```diff
@@ -115,15 +115,14 @@ noncomputable def sqrt (x : ℝ) : ℝ :=
...
-/- quotient.lift_on x
-  (λ f, mk ⟨sqrt_aux f, (sqrt_aux_converges f).fst⟩)
-  (λ f g e, begin
-    rcases sqrt_aux_converges f with ⟨hf, x, x0, xf, xs⟩,
-    rcases sqrt_aux_converges g with ⟨hg, y, y0, yg, ys⟩,
-    refine xs.trans (eq.trans _ ys.symm),
-    rw [← @mul_self_inj_of_nonneg ℝ _ x y x0 y0, xf, yg],
-    congr' 1, exact quotient.sound e
-  end)-/
+/- Quotient.liftOn x
```
mattrobball, line 118: What is this comment doing?

eric-wieser: This is some partial code to make `Real.sqrt` computable, which perhaps lost some context that was present in Lean 3.

eric-wieser: The context was a
```
/- TODO(Mario): finish the proof
theorem sqrt_aux_converges (f : cau_seq ℚ abs) : ∃ h x, 0 ≤ x ∧ x * x = max 0 (mk f) ∧
  mk ⟨sqrt_aux f, h⟩ = x :=
begin
  rcases sqrt_exists (le_max_left 0 (mk f)) with ⟨x, x0, hx⟩,
  suffices : ∃ h, mk ⟨sqrt_aux f, h⟩ = x,
  { exact this.imp (λ h e, ⟨x, x0, hx, e⟩) },
  apply of_near,

  rsuffices ⟨δ, δ0, hδ⟩ : ∃ δ > 0, ∀ i, abs (↑(sqrt_aux f i) - x) < δ / 2 ^ i,
  { intros }
end -/
```
comment a few lines above.

mattrobball: Can we add that back?

mattrobball: Or remove and make an issue?

eric-wieser: I think it's left from #11909. Let's remove it, but not in this PR.

mattrobball:
bors merge

Mathlib/Data/Real/Sqrt.lean
mattrobball: #15976

grunweg:
Thanks for the fast review!

PR #15974, opened 2024-08-19
test: add assertNotExists linter and extension

adomani:
---


adomani:
The early attempts here had exponentially growing oleans.

PR #15982, opened 2024-08-19
[Merged by Bors] - feat: disjoint union of topological spaces is commutative and associative

grunweg:
These results are fully analogous to existing results about products, but are missing in mathlib.
While at it, also add the four-way commutativity `sumSumSumComm`, whose product analogue is already in mathlib.

From my bordism theory project (which wants to prove smooth analogues of these results).

---

### DIFF
--- a/Mathlib/Logic/Equiv/Basic.lean
+++ b/Mathlib/Logic/Equiv/Basic.lean
@@ -3,7 +3,6 @@ Copyright (c) 2015 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
-import Aesop
 import Mathlib.Data.Bool.Basic
 import Mathlib.Data.Option.Defs
 import Mathlib.Data.Prod.Basic
@@ -379,8 +378,8 @@ def sumSumSumComm (α β γ δ) : (α ⊕ β) ⊕ γ ⊕ δ ≃ (α ⊕ γ) ⊕ 
       ∘ (Sum.map (Sum.map (@id α) (sumComm β γ).symm) (@id δ))
       ∘ (Sum.map (sumAssoc α γ β) (@id δ))
       ∘ (sumAssoc (α ⊕ γ) β δ).symm
-  left_inv x := by aesop
-  right_inv x := by aesop
+  left_inv x := by simp [Function.comp]
+  right_inv x := by simp [Function.comp]
 
 @[simp]
 theorem sumSumSumComm_symm (α β γ δ) : (sumSumSumComm α β γ δ).symm = sumSumSumComm α γ β δ :=
--- a/Mathlib/Logic/Equiv/Basic.lean
+++ b/Mathlib/Logic/Equiv/Basic.lean
@@ -3,6 +3,7 @@ Copyright (c) 2015 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
+import Aesop
 import Mathlib.Data.Bool.Basic
 import Mathlib.Data.Option.Defs
 import Mathlib.Data.Prod.Basic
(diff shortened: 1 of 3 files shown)


Mathlib/Logic/Equiv/Basic.lean
```diff
@@ -3,6 +3,7 @@ Copyright (c) 2015 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
+import Aesop
```
j-loreaux, line 6: This is too heavy an import for this file I think, and you don't need it with my fix below.

Mathlib/Logic/Equiv/Basic.lean
```diff
@@ -365,6 +366,26 @@ theorem sumAssoc_symm_apply_inr_inl {α β γ} (b) :
...
+      ∘ (Sum.map (Sum.map (@id α) (sumComm β γ)) (@id δ))
+      ∘ (Sum.map (sumAssoc α β γ) (@id δ))
+      ∘ (sumAssoc (α ⊕ β) γ δ).symm
+  invFun :=
+    (sumAssoc (α ⊕ β) γ δ) ∘ (Sum.map (sumAssoc α β γ).symm (@id δ))
+      ∘ (Sum.map (Sum.map (@id α) (sumComm β γ).symm) (@id δ))
+      ∘ (Sum.map (sumAssoc α γ β) (@id δ))
+      ∘ (sumAssoc (α ⊕ γ) β δ).symm
+  left_inv x := by aesop
+  right_inv x := by aesop
```
j-loreaux, line 383: Remember, `∘` is not a reducible transparency, and `Function.comp` is not a `simp` lemma. You need to unfold it manually.
```suggestion
  left_inv x := by simp [Function.comp]
  right_inv x := by simp [Function.comp]
```
In fact, this makes me wonder slightly if your definitions of `toFun` and `invFun` should not be written in point-free style, but rather with a `match`. I'm not sure about the answer to this, maybe you can explore?

Mathlib/Topology/Homeomorph.lean
```diff
@@ -512,6 +512,77 @@ section
 
 variable (X Y Z)
 
+/-- `X ⊕ Y` is homeomorphic to `Y ⊕ X`. -/
+def sumComm : X ⊕ Y ≃ₜ Y ⊕ X where
+  toEquiv := Equiv.sumComm X Y
+  continuous_toFun := continuous_sum_swap
+  continuous_invFun := continuous_sum_swap
```
j-loreaux, line 519: This and the rest of the lemmas here seem to be dumped in the middle of a `Prod` section. There are a few lemmas here about `Sum` (e.g., `sumCongr`). Maybe this fits better there?

Mathlib/Topology/Homeomorph.lean
```diff
@@ -512,6 +512,77 @@ section
...
+  toEquiv := Equiv.sumAssoc X Y Z
+  continuous_toFun := continuous_sumAssoc X Y Z
+  continuous_invFun := continuous_sumAssoc_symm X Y Z
+
+@[simp]
+lemma sumAssoc_toEquiv : (sumAssoc X Y Z).toEquiv = Equiv.sumAssoc X Y Z := rfl
+
+-- FIXME: move W further up...
+variable (X Y W Z : Type*)
+  [TopologicalSpace X] [TopologicalSpace Y] [TopologicalSpace Z] [TopologicalSpace W]
```
j-loreaux, line 548: Probably you should just deal with this now?

Mathlib/Logic/Equiv/Basic.lean
grunweg: Can you explain more why importing aesop is a heavy import? Those proofs can be replaced by the output of aesop?, right? (In any case, your proposed proof is nicer, so I have removed the import again.)

Mathlib/Topology/Homeomorph.lean
grunweg: `sumCongr` is right next to `prodCongr` and a `Congr` definition...
I moved this into its own section, right before the corresponding `Prod` section.

Mathlib/Logic/Equiv/Basic.lean
grunweg: I'm not sure if I understand your suggestion with a match. Can you say more about this?

j-loreaux: Yes, I'll get to it today.

j-loreaux: Sorry, I got distracted. This is what I meant, but I'm not saying it's actually better:
```lean
/-- Four-way commutativity of `sum`. The name matches `add_add_add_comm`. -/
@[simps apply]
def sumSumSumComm (α β γ δ) : (α ⊕ β) ⊕ γ ⊕ δ ≃ (α ⊕ γ) ⊕ β ⊕ δ where
  toFun := fun
    | inl (inl a) => inl (inl a)
    | inl (inr b) => inr (inl b)
    | inr (inl c) => inl (inr c)
    | inr (inr d) => inr (inr d)
  invFun := fun
    | inl (inl a) => inl (inl a)
    | inl (inr c) => inr (inl c)
    | inr (inl b) => inl (inr b)
    | inr (inr d) => inr (inr d)
  left_inv := by rintro ((a | b) | c | d); all_goals rfl
  right_inv := by rintro ((a | c) | b | d); all_goals rfl
```

Mathlib/Topology/Homeomorph.lean
```diff
@@ -508,9 +508,81 @@ theorem prodCongr_symm (h₁ : X ≃ₜ X') (h₂ : Y ≃ₜ Y') :
 theorem coe_prodCongr (h₁ : X ≃ₜ X') (h₂ : Y ≃ₜ Y') : ⇑(h₁.prodCongr h₂) = Prod.map h₁ h₂ :=
   rfl
 
-section
+-- Commutativity and associativity of the disjoing union of topological spaces,
```
j-loreaux, line 511:
```suggestion
-- Commutativity and associativity of the disjoint union of topological spaces,
```

j-loreaux (approved):
bors d+

Mathlib/Logic/Equiv/Basic.lean
grunweg: Yes, this makes sense - thanks for explaining. (I checked: aesop has about 15000 lines of code, in a hundred files. That's definitely a heavy dependency.)

Mathlib/Logic/Equiv/Basic.lean
grunweg: No worries; thanks for elaborating. I also like your definition. I'm not sure either which one is better. (Both would be nice to have.)
I'll merge this PR as-is; if you think this relation would be useful to have, just say so and I can file a follow-up PR to either change the definition (and add the equality with the current definition as lemmas) or vice versa).

grunweg:
Thanks for the review!
bors merge

PR #15985, opened 2024-08-19
chore(CategoryTheory/Sites): remove unused lemmas and simplify some proofs in Sites.Discrete

dagurtomas:
This PR does the following
- Renames `Sheaf.IsDiscrete` to `Sheaf.IsConstant`. 
- Renames the file `Sites/Discrete` to `Sites/IsConstant`
- Removes two lemmas which were essentially duplicates of other ones, and simplifies a proof to be able to remove an auxiliary lemma. 

---

I added these lemmas initially because I used them in #14027, but it turns out they're essentially duplicates of other lemmas, now merged into one lemma `constantSheafAdj_counit_w` which is used in this file and in #14027.


joelriou:
Why not using the name `IsConstant`? Is "discrete" a terminology that is specific to condensed objects?

dagurtomas:
I guess "discrete" is specific to topological objects such as condensed sets, and `IsConstant` is better for general sheaves (note that the constant sheaf is in general not with this terminology though...). I'll rename it.

dagurtomas:
I messed up the merge, closing in favour of #16385

PR #15987, opened 2024-08-19
feat(Order/Basic): product order is subrelation of lexicographic order

vihdzp:
The product order on `α × β` is a subrelation of `Prod.Lex (· < ·) (· < ·)` whenever `α` is a partial order.

---

It'd be really nice if `decreasing_tactic` could be extended with `Lex.lt_of_le_of_lt`. Currently, the following code fails:
```lean 4
import Mathlib.Order.Basic
import Mathlib.Data.Nat.Defs

theorem recursion (a b : Nat) : True := by
  let (c, d) := (0, 0)
  have : c ≤ a := sorry
  have : d < b := sorry
  exact recursion c d -- failed to prove termination
termination_by (a, b)
```

### DIFF
--- a/Mathlib/Order/Basic.lean
+++ b/Mathlib/Order/Basic.lean
@@ -1172,6 +1172,23 @@ lemma mk_lt_mk_of_le_of_lt (h₁ : a₁ ≤ a₂) (h₂ : b₁ < b₂) : (a₁, 
 
 end Preorder
 
+section PartialOrder
+
+variable [PartialOrder α] [Preorder β] {x y : α × β}
+
+lemma toLex (h : x < y) : Prod.Lex (· < ·) (· < ·) x y := by
+  cases y
+  cases' lt_iff.1 h with h₁ h₂
+  · exact Lex.left _ _ h₁.1
+  · obtain h₂ | rfl := lt_or_eq_of_le h₂.1
+    · exact Lex.left _ _ h₂
+    · exact Lex.right _ h₂.2
+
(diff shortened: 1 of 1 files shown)


Mathlib/Order/Basic.lean
```diff
@@ -1172,6 +1172,27 @@ lemma mk_lt_mk_of_le_of_lt (h₁ : a₁ ≤ a₂) (h₂ : b₁ < b₂) : (a₁,
...
+
+variable [PartialOrder α] [Preorder β] {x y : α × β}
+
+lemma toLex (h : x < y) : Prod.Lex (· < ·) (· < ·) x y := by
+  cases y
+  cases' lt_iff.1 h with h₁ h₂
+  · exact Lex.left _ _ h₁.1
+  · obtain h₂ | rfl := lt_or_eq_of_le h₂.1
+    · exact Lex.left _ _ h₂
+    · exact Lex.right _ h₂.2
```
eric-wieser, line 1185: Is this related to

https://github.com/leanprover-community/mathlib4/blob/48fd37e25593c8ea6b65bd3d60a4330cdd12d40d/Mathlib/Data/Prod/Lex.lean#L112-L116

somehow?

vihdzp: I think this was the result I was looking for earlier. I'll edit the PR later to only include the `lt_of_le_of_lt` result.

vihdzp:
Hm, actually I'm not sure anymore if any of this is needed. Closing for now.

PR #15989, opened 2024-08-20
feat(SetTheory/Ordinal/Principal): golf theorems on additively principal ordinals

vihdzp:
We add some new results on additively principal ordinals and use them to simplify existing theorems.

---

- [x] depends on: #15990
- [x] depends on: #17740
- [x] depends on: #17741
- [x] depends on: #17742
- [x] depends on: #17842
- [x] depends on: #18902


Mathlib/SetTheory/Ordinal/Principal.lean
```diff
@@ -8,7 +8,8 @@ import Mathlib.SetTheory.Ordinal.FixedPoint
 /-!
 # Principal ordinals
 
-We define principal or indecomposable ordinals, and we prove the standard properties about them.
+We define principal or indecomposable ordinals, i.e. ordinals closed under a binary operation, and
```
kbuzzard, line 11: What binary operation? Presumably not an arbitrary one?

Mathlib/SetTheory/Ordinal/Principal.lean
```diff
@@ -195,65 +204,63 @@ theorem principal_add_iff_add_lt_ne_self : Principal (· + ·) a ↔ ∀ b < a,
...
-@[deprecated (since := "2024-09-30")]
-alias add_omega_opow := add_omega0_opow
+-- `add_omega0` is proven in the Arithmetic file.
+
+theorem add_of_le_omega0 : a < ω → ω ≤ b → a + b = b :=
+  principal_add_omega0.add_absorp_of_le
 
-theorem principal_add_omega0_opow (o : Ordinal) : Principal (· + ·) (ω ^ o) :=
-  principal_add_iff_add_left_eq_self.2 fun _ => add_omega0_opow
+theorem principal_add_omega0_opow (o : Ordinal) : Principal (· + ·) (ω ^ o) := by
```
kbuzzard, line 219: Why has this proof got a lot longer?

Mathlib/SetTheory/Ordinal/Principal.lean
vihdzp: One can talk about different flavors of principal ordinals, with the most common/useful being principal additive and principal multiplicative. That said, I am interesting in proving the characterization of exponentially principal and Veblen principal ordinals as well (since these just correspond to the epsilon and gamma ordinals, respectively, which I have a PR open for).

Mathlib/SetTheory/Ordinal/Principal.lean
vihdzp: I switched the order of proofs around, proving this before proving `add_omega0_opow`. Note that the two proofs combined are shorter than they were before.

Mathlib/SetTheory/Ordinal/Principal.lean
```diff
@@ -180,12 +181,20 @@ theorem principal_add_iff_add_left_eq_self : Principal (· + ·) o ↔ ∀ a < o
   · rw [← h a hao]
     exact (isNormal_add_right a).strictMono hbo
 
+theorem Principal.add_absorp (ho : Principal (· + ·) o) (ha : a < o) : a + o = o :=
```
YaelDillies, line 184:
```suggestion
theorem Principal.add_eq_right (ho : Principal (· + ·) o) (ha : a < o) : a + o = o :=
```
sounds better than making `absorp` part of our naming convention

Mathlib/SetTheory/Ordinal/Principal.lean
kbuzzard: I think this is still an extremely confusing first line. How about spelling it out: "If `r` is a binary operation on ordinals then we say that an ordinal is r-principal if, when considered as a set of ordinals, it is closed under this binary operation", or something like that? My instinct is that the module docstring of a mathlib file should be as readable as possible for non-experts (e.g. someone like me who knows what an ordinal is but has never heard of principal ordinals).

PR #15990, opened 2024-08-20
[Merged by Bors] - feat(SetTheory/Ordinal/Exponential): more lemmas on `Ordinal.log`

vihdzp:
We show multiple variants of `opow_le_iff_le_log` with slightly different assumptions.

---


Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -86,6 +86,13 @@ theorem opow_pos {a : Ordinal} (b : Ordinal) (a0 : 0 < a) : 0 < a ^ b := by
 theorem opow_ne_zero {a : Ordinal} (b : Ordinal) (a0 : a ≠ 0) : a ^ b ≠ 0 :=
   Ordinal.pos_iff_ne_zero.1 <| opow_pos b <| Ordinal.pos_iff_ne_zero.2 a0
 
+theorem opow_eq_zero {a b : Ordinal} : a ^ b = 0 ↔ a = 0 ∧ b ≠ 0 := by
```
YaelDillies, line 89:
```suggestion
@[simp] theorem opow_eq_zero {a b : Ordinal} : a ^ b = 0 ↔ a = 0 ∧ b ≠ 0 := by
```

YaelDillies (approved):
maintainer merge

Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -287,16 +295,49 @@ theorem opow_log_le_self (b : Ordinal) {x : Ordinal} (hx : x ≠ 0) : b ^ log b
...
+theorem opow_le_of_le_log {b x c : Ordinal} (hc : c ≠ 0) (h : c ≤ log b x) : b ^ c ≤ x := by
+  obtain hb | hb := le_or_lt b 1
+  · rw [log_of_left_le_one hb] at h
+    exact (h.not_lt (Ordinal.pos_iff_ne_zero.2 hc)).elim
+  · rwa [opow_le_iff_le_log' hb hc]
 
 theorem lt_opow_iff_log_lt {b x c : Ordinal} (hb : 1 < b) (hx : x ≠ 0) : x < b ^ c ↔ log b x < c :=
   lt_iff_lt_of_le_iff_le (opow_le_iff_le_log hb hx)
 
+/-- This lemma assumes `c ≠ 0` rather than `x ≠ 0`. -/
```
b-mehta, line 331: Let's reference the one which does assume x != 0 here, and also reference the one-sided ones since they have weaker assumptions. Same for the earlier lemmas.

vihdzp: The version assuming `x ≠ 0` is just the unprimed variant, so I'm not sure if this needs to be stated explicitly. As for the other variants, maybe we can reference them all at once in the docstring for `lt_opow_iff_log_lt`? Something like:
```
/-- `opow b` and `log b` (almost) form a Galois connection.

See `lt_opow_iff_log_lt'` for a variant assuming `c ≠ 0` rather than `x ≠ 0`. See also
`lt_opow_of_log_lt` and `lt_log_of_lt_opow`, which are both separate implications under weaker
assumptions. -/
theorem lt_opow_iff_log_lt {b x c : Ordinal} (hb : 1 < b) (hx : x ≠ 0) : x < b ^ c ↔ log b x < c :=
  lt_iff_lt_of_le_iff_le (opow_le_iff_le_log hb hx)
```
Ditto for `opow_le_iff_le_log`.

b-mehta: Having this comment without referencing the other lemma is the problem - you can remove the comment (bad) or make the comment useful for people hovering (good). 

I think referencing them in only one place defeats the entire point of cross-referencing in docstrings

vihdzp: I guess the question then becomes: should each of these four lemmas reference all other three? Or how do we want to go about this?

b-mehta: `lt_opow_iff_log_lt` and `lt_opow_iff_log_lt'` should reference each other. And then each of `lt_opow_iff_log_lt` and `lt_opow_iff_log_lt'` should reference each of `lt_opow_of_log_lt` and `lt_log_of_lt_opow`

jcommelin:
Removing `maintainer-merge`, since it's been > 1wk.

vihdzp:
Sorry, forgot about this! Will get back to it soon.

YaelDillies (approved):
maintainer merge

Mathlib/SetTheory/Ordinal/Exponential.lean
```diff
@@ -286,17 +294,67 @@ theorem opow_log_le_self (b : Ordinal) {x : Ordinal} (hx : x ≠ 0) : b ^ log b
...
-        ((opow_le_opow_iff_right hb).2 (succ_le_of_lt hn)).trans h,
-    fun h => ((opow_le_opow_iff_right hb).2 h).trans (opow_log_le_self b hx)⟩
-
+/-- `opow b` and `log b` (almost) form a Galois connection.
+
+See `opow_le_iff_le_log'` for a variant assuming `c ≠ 0` rather than `x ≠ 0`. See also
+`le_log_of_opow_le` and `opow_le_of_le_log`, which are both separate implications under weaker
+assumptions. -/
+theorem opow_le_iff_le_log {b x c : Ordinal} (hb : 1 < b) (hx : x ≠ 0) :
+    b ^ c ≤ x ↔ c ≤ log b x := by
```
b-mehta, line 303: I think I prefer this lemma in the opposite direction actually - ordinal exponentiation feels like a more fundamental operation than `log`; so this is a lemma describing how log relates to it, and so having it in the backwards direction feels more natural... Similarly for the other `iff` statements like this one. But this is optional.

bors d+

vihdzp: I agree with this, but I believe this lemma is used at least a couple of times, so it'd have to be deprecated and replaced. I'll do this in a follow-up PR.

PR #15991, opened 2024-08-20
feat(SetTheory/Ordinal/CantorNormalForm): alternate CNF recursion principle

vihdzp:
The idea of this recursion principle is that you can build an ordinal by adding one power of `b` at a time, as long as `1 < b ≤ ω`.

---


Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -37,6 +37,9 @@ open List
 
 namespace Ordinal
 
+/-! ### Recursion principles -/
+
+
```
YaelDillies, line 42:
```suggestion
```
Same below

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -57,6 +60,29 @@ theorem CNFRec_pos (b : Ordinal) {o : Ordinal} {C : Ordinal → Sort*} (ho : o 
     (H : ∀ o, o ≠ 0 → C (o % b ^ log b o) → C o) :
     @CNFRec b C H0 H o = H o ho (@CNFRec b C H0 H _) := by rw [CNFRec, dif_neg ho]
 
+/-- Inducts on the base `ω` expansion of an ordinal.
+
+This differs from `CNFRec` in that every instance of `ω ^ a` is considered separately. -/
```
YaelDillies, line 65: Is it not possible to consider eveery instance of `b ^ a` separately for a general base `b`?

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -57,6 +60,29 @@ theorem CNFRec_pos (b : Ordinal) {o : Ordinal} {C : Ordinal → Sort*} (ho : o 
...
+@[elab_as_elim]
+noncomputable def CNFRec_omega {C : Ordinal → Sort*} (H0 : C 0)
```
YaelDillies, line 67:
```suggestion
noncomputable def CNFRecOmega {C : Ordinal → Sort*} (H0 : C 0)
```
or
```suggestion
noncomputable def CNFOmegaRec {C : Ordinal → Sort*} (H0 : C 0)
```

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
vihdzp: I've revamped this PR with said generalization.

YaelDillies (approved):
Thanks! 🚀

maintainer merge

Mathlib/SetTheory/Ordinal/CantorNormalForm.lean
```diff
@@ -55,6 +57,28 @@ theorem CNFRec_pos (b : Ordinal) {o : Ordinal} {C : Ordinal → Sort*} (ho : o 
     CNFRec b H0 H o = H o ho (@CNFRec b C H0 H _) := by
   rw [CNFRec, dif_neg]
 
+/-- Inducts on the base `b` expansion of an ordinal, for `1 < b ≤ ω`.
+
+This differs from `CNFRec` in that every instance of `b ^ a` is considered separately. -/
+@[elab_as_elim]
+noncomputable def CNFRec' {C : Ordinal → Sort*} {b : Ordinal} (hb₁ : 1 < b) (hb : b ≤ ω) (H0 : C 0)
```
eric-wieser, line 64: Per the naming conventions, this and the def above should be `cnfrec` or perhaps `cnfRec` or `recCNF`. To avoid deprecating your new name as soon as its introduced, can you make a PR to fix the existing name first?

PR #15994, opened 2024-08-20
[Merged by Bors] - feat: allow several identifiers in assert_not_exists

adomani:
Align the `assert_not_exists` syntax to `assert_not_imported`: they now both allow several identifiers.

---

Even though the diff is big, there is not really much to this PR:
* `Mathlib/Util/AssertExists.lean` looping over the various indices, means indenting the whole `assert_not_exists` body,
  plus adding some `let`s for variables used in the loop, but computable once;
* `test/AssertExists.lean` a test file testing the new syntax;
* `*` using the syntax wherever there were consecutive `assert_not_exists` -- done via a script.


mattrobball (commented):
I'm not sure there is precedent for column aligning multiple arguments into a command.

adomani:
I did it because it seems more legible, but I did not think about it being unusual.

Should I put each block on the same line?

joneugster:
I haven't fully read the PR yet, but I like the main change. @adomani if you could resolve merge conflicts and ping me, then I'll see that this PR get's merged promptly

joneugster:
I need to read the PR carefully tomorrow. I got the impression that a) that's already how `assert_not_imported` works and b) there are two equivalent options to write things:

```lean
assert_not_exists Rat Nat List

-- or

assert_not_exists
  Rat
  Nat
  List
```

arguably, I find the former nicer, too.

adomani:
Ah, I did not realize that there was an issue with the syntax!  Both forms are accepted by the command:
```lean
assert_not_exists Nat Rat Real

assert_not_exists
  Nat
  Rat
  Real
```
are both valid.

Mathlib/Data/Fin/Basic.lean
```diff
@@ -67,9 +67,6 @@ This file expands on the development in the core library.
   by `i ↦ n-(i+1)`
 -/
 
-assert_not_exists Monoid
-assert_not_exists Fintype
-universe u v
```
joneugster, line 72: is this by accident?

Mathlib/Util/AssertExists.lean
```diff
@@ -78,8 +78,8 @@ elab "assert_exists " n:ident : command => do
   let _ ← liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
 
 /--
-`assert_not_exists n` is a user command that asserts that a declaration named `n` *does not exist*
-in the current import scope.
+`assert_not_exists d₁ d₂ ... dₙ` is a user command that asserts that the declarations named
+`d₁ d₂ ... dₙ` *does not exist* in the current import scope.
```
joneugster, line 82:
```suggestion
`d₁ d₂ ... dₙ` *do not exist* in the current import scope.
```

Mathlib/Util/AssertExists.lean
joneugster, line 1: For maintainer: This is the only file with changes (besides the test file), the other 94 files simple contain trivial adaptations which I've checked.

Mathlib/Util/AssertExists.lean
```diff
@@ -95,28 +95,32 @@ You should *not* delete the `assert_not_exists` statement without careful discus
...
-        idx := i
-        msg := msg ++ m!"\n  which is imported by {env.header.moduleNames[i]!},"
-    pure <| msg ++ m!"\n  which is imported by this file.")
-  throw <| .error n m!"{msg}\n\n\
-    These invariants are maintained by `assert_not_exists` statements, \
-    and exist in order to ensure that \"complicated\" parts of the library \
-    are not accidentally introduced as dependencies of \"simple\" parts of the library."
+  let modNames := env.header.moduleNames
+  let modData := env.header.moduleData
+  let modDataSize := env.header.moduleData.size
```
joneugster, line 102:
```suggestion
  let modDataSize := modData.size
```
optional

joneugster (approved):
@adomani could you double-check `Mathlib/Data/Fin/Basic.lean`? see comment below.

after that I think this can be merged, thanks for incorporating Matthew's suggestion

joneugster:
Also, @adomani I think you missed several files:

-

joneugster (approved):

Mathlib/Data/Fin/Basic.lean
adomani: Yes, it was certainly not intended: thanks for finding this!

adomani (commented):
Jon, thank you very much for the review, for checking the code and updating the remaining `assert_not_exists`!

Those were probably further `asserts` that appeared after I opened the PR.

joneugster (approved):
Thanks!

maintainer merge

kim-em:
bors merge

PR #15995, opened 2024-08-20
[Merged by Bors] - fix: remove stale comment

kbuzzard:
---

When ported, two lines in this file looked like
```lean
-- Syntactic tautology
#noalign lie_submodule.to_submodule_eq_coe
```

and then a year later we deleted the `#noalign` line, leaving a stale comment about nothing.

### DIFF
--- a/Mathlib/Algebra/Group/Submonoid/Operations.lean
+++ b/Mathlib/Algebra/Group/Submonoid/Operations.lean
@@ -531,8 +531,6 @@ protected theorem pow_mem {M : Type*} [Monoid M] (S : Submonoid M) {x : M} (hx :
     x ^ n ∈ S :=
   pow_mem hx n
 
--- Porting note: coe_pow removed, syntactic tautology
-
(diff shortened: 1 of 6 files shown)


pitmonticone:
maintainer merge

jcommelin:
Good catch! I think there are a few more examples in mathlib. Would you mind hunting them down with `git grep -i "syntactic tautology"` so that we can remove all of them in one go?

pitmonticone:
@kbuzzard I should have caught all of those comments mentioned by Johan. May I push directly here for efficiency?

j-loreaux:
@pitmonticone I'll answer for Kevin, as I don't imagine I'll be stepping on his toes: sure. Thanks.

kbuzzard:
It's a bit weird reviewing my own PR but I just checked the other instances which Pietro has caught and they look good to me!

j-loreaux:
bors merge

PR #15996, opened 2024-08-20
[Merged by Bors] - chore: use "bash strict mode" in all shell scripts

grunweg:
If a script produces an error, it should rather do so loudly than silently.
See http://redsymbol.net/articles/unofficial-bash-strict-mode/ for an explanation of this incantation;
in particular the section on positional parameters (which these scripts do need a few times).

---


scripts/declarations_diff.sh
```diff
@@ -105,6 +109,9 @@ fi |
   }'
 )
 
+set +e ## TODO, bisect the error...
```
bryangingechen, line 112: What's the status of this TODO? 

The other changes LGTM so feel free to merge if it's not a big deal in your opinion.
bors d+

grunweg: I have tried a bit to understand why activiating `set -e` yields an error, but to no success.
It would be great to understand and fix this - but this PR is already an improvement, so I will merge it now.

grunweg: Thanks for asking and the review!

grunweg:
bors merge

PR #16002, opened 2024-08-20
[Merged by Bors] - feat: `assert_not_exists` and `assert_not_imported` checks

adomani:
This PR "completes" the checks implied by `assert_not_exists` and `assert_not_imported`: at the time of using one of those commands, the declarations/imports are not present in the environment.

Now, each time that one of the commands above is used, the information is recorded in the environment, so that running `#check_assertions` reports the status of each one.

If `#check_assertions` finds that something is still missing, then it emits a warning, otherwise it just logs the report.

The `#check_assertions!` variant is silent unless some declaration/import is missing and, in that case, it only prints what is missing.  This is useful for testing.

There is also the extra file `test/AssertImported.lean` consisting of
```lean
import Mathlib

#check_assertions!
```
to make sure that CI fails if there are non-existing imports/declarations.

---

### DIFF
--- a/Mathlib/Util/AssertExists.lean
+++ b/Mathlib/Util/AssertExists.lean
@@ -47,14 +47,11 @@ elab "#check_assertions" tk:("!")?: command => do
   let allMods := env.allImportedModuleNames
   let mut msgs := #[m!""]
   let mut outcome := m!""
-  let mut (allExist?, cond) := (true, false)
+  let mut allExist? := true
   for d in ext.toArray.qsort fun d e => (e.isDecl < d.isDecl) ||
       (e.isDecl == d.isDecl && (d.givenName.toString < e.givenName.toString)) do
     let type := if d.isDecl then "declaration" else "module"
-    if d.isDecl then
-      cond := env.contains d.givenName
-    else
-      cond := allMods.contains d.givenName
+    let cond := if d.isDecl then env.contains d.givenName else allMods.contains d.givenName
     outcome := if cond then m!"{checkEmoji}" else m!"{crossEmoji}"
     allExist? := allExist? && cond
     if tk.isNone || !cond then
--- a/Mathlib/Util/AssertExists.lean
+++ b/Mathlib/Util/AssertExists.lean
@@ -113,7 +113,9 @@ You should *not* delete the `assert_not_exists` statement without careful discus
 `assert_not_exists` statements should generally live at the top of the file, after the module doc.
 -/
 elab "assert_not_exists " n:ident : command => do
-  let decl ← ((liftCoreM <| realizeGlobalConstNoOverloadWithInfo n) <|> return .anonymous)
+  let decl ←
+    try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
+    catch _ => return .anonymous
   if decl == .anonymous then
     Mathlib.AssertNotExist.addDeclEntry true n.getId
   else
--- a/Mathlib/Util/AssertExists.lean
+++ b/Mathlib/Util/AssertExists.lean
@@ -115,10 +115,10 @@ You should *not* delete the `assert_not_exists` statement without careful discus
 elab "assert_not_exists " n:ident : command => do
   let decl := ←
     try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
-    catch _ => return .anonymous
-  if decl == .anonymous then
-    Mathlib.AssertNotExist.addDeclEntry true n.getId
-  else
+    catch _ =>
+      Mathlib.AssertNotExist.addDeclEntry true n.getId
+      return .anonymous
+  if decl != .anonymous then
   let env ← getEnv
   let c ← mkConstWithLevelParams decl
   let msg ← (do
--- a/Mathlib/Util/AssertExists.lean
+++ b/Mathlib/Util/AssertExists.lean
@@ -5,6 +5,7 @@ Authors: Patrick Massot, Scott Morrison
 -/
 import Mathlib.Init
 import Lean.Elab.Command
+import Mathlib.Util.AssertExistsExt
 
 /-!
 # User commands for assert the (non-)existence of declaration or instances.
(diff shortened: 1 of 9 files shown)


adomani:
!bench

Mathlib/Util/AssertExists.lean
```diff
@@ -52,7 +113,10 @@ You should *not* delete the `assert_not_exists` statement without careful discus
 `assert_not_exists` statements should generally live at the top of the file, after the module doc.
 -/
 elab "assert_not_exists " n:ident : command => do
-  let decl ← try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n catch _ => return
+  let decl ← ((liftCoreM <| realizeGlobalConstNoOverloadWithInfo n) <|> return .anonymous)
```
mattrobball, line 116: I think a `try ... catch` block is more natural here.

mattrobball (commented):
I think it looks good! Thanks. 

Is the CI step coming later?

adomani:
Actually, the CI step is already embedded in the tests: one of the files calls the `#check_assertions!` command with `import Mathlib`: the command is noisy if and only if there are non-existent declarations or imports.

Mathlib/Util/AssertExists.lean
adomani, line 116:
```suggestion
  let decl ←
    try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
    catch _ => return .anonymous
```

Mathlib/Util/AssertExists.lean
```diff
@@ -52,7 +113,12 @@ You should *not* delete the `assert_not_exists` statement without careful discus
 `assert_not_exists` statements should generally live at the top of the file, after the module doc.
 -/
 elab "assert_not_exists " n:ident : command => do
-  let decl ← try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n catch _ => return
+  let decl ←
+    try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
+    catch _ => return .anonymous
```
adomani, line 118:
```suggestion
    catch _ => return Name.anonymous
```

Mathlib/Util/AssertExists.lean
```diff
@@ -52,7 +113,10 @@ You should *not* delete the `assert_not_exists` statement without careful discus
 `assert_not_exists` statements should generally live at the top of the file, after the module doc.
 -/
 elab "assert_not_exists " n:ident : command => do
-  let decl ← try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n catch _ => return
+  let decl ← ((liftCoreM <| realizeGlobalConstNoOverloadWithInfo n) <|> return .anonymous)
+  if decl == .anonymous then
+    Mathlib.AssertNotExist.addDeclEntry true n.getId
```
mattrobball:
```suggestion
  let decl ← try liftCoreM <| realizeGlobalConstNoOverloadWithInfo n
    catch _ => 
      Mathlib.AssertNotExist.addDeclEntry true n.getId
      return
```

adomani: :facepalm: Thanks!  I can then remove the awkward `if ... .anonymous ... then ... else`!

adomani: Actually, `return` does not work, since I should return a name.  I still added the declaration to the environment inside the catch, since that allows to only run the rest of the code in the non-`anonymous` branch.

Mathlib/Util/AssertExists.lean
```diff
@@ -22,6 +23,66 @@ Implement `assert_instance` and `assert_no_instance`
...
+  let ext := assertExistsExt.getState env
+  if ext.isEmpty && tk.isNone then logInfo "No assertions made." else
+  let allMods := env.allImportedModuleNames
+  let mut msgs := #[m!""]
+  let mut outcome := m!""
+  let mut (allExist?, cond) := (true, false)
+  for d in ext.toArray.qsort fun d e => (e.isDecl < d.isDecl) ||
+      (e.isDecl == d.isDecl && (d.givenName.toString < e.givenName.toString)) do
+    let type := if d.isDecl then "declaration" else "module"
+    if d.isDecl then
```
mattrobball, line 54: This logic makes me think there should only be one command that handles both declarations and imports.

Mathlib/Util/AssertExists.lean
```diff
@@ -22,6 +23,66 @@ Implement `assert_instance` and `assert_no_instance`
...
+environment.  In particular, it is silent if everything is imported, making it useful for testing.
+-/
+elab "#check_assertions" tk:("!")?: command => do
+  let env ← getEnv
+  let ext := assertExistsExt.getState env
+  if ext.isEmpty && tk.isNone then logInfo "No assertions made." else
+  let allMods := env.allImportedModuleNames
+  let mut msgs := #[m!""]
+  let mut outcome := m!""
+  let mut (allExist?, cond) := (true, false)
```
mattrobball, line 50: Do you use `cond` outside the loop?

Mathlib/Util/AssertExists.lean
mattrobball: Sorry. I don't understand.

Mathlib/Util/AssertExists.lean
mattrobball: Hmm the comment wasn't focused correctly. I mean the `if d.isDecl` branching

Mathlib/Util/AssertExists.lean
adomani: Actually, it was I who did not understand: I had added a `:=` before the `←` and then I had to return a `Name`.  Removing the `:=` allowed for the slicker code that you suggested.

mattrobball:
> Actually, the CI step is already embedded in the tests: one of the files calls the `#check_assertions!` command with `import Mathlib`: the command is noisy if and only if there are non-existent declarations or imports.

I kinda liked the previous incarnation. I think this is good for an initial commit so I don't think we should go back now but the feedback in CI for failure was a bit earlier. 

Aside: why are building the archive and counterexamples so high up in the workflow?

Mathlib/Util/AssertExists.lean
mattrobball: If there happened to be collision, I would make sure both don't exist and let the error disambiguate if that failed. But this is not in scope here.

Mathlib/Util/AssertExists.lean
```diff
@@ -22,6 +23,66 @@ Implement `assert_instance` and `assert_no_instance`
...
+
+If all declarations and imports are available when `#check_assertions` is used,
+then the command logs an info. Otherwise, it emits a warning.
+
+The variant `#check_assertions!` only prints declarations/imports that are not present in the
+environment.  In particular, it is silent if everything is imported, making it useful for testing.
+-/
+elab "#check_assertions" tk:("!")?: command => do
+  let env ← getEnv
+  let ext := assertExistsExt.getState env
```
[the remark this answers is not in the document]
adomani: I like this suggestion and added `getSortedAssertExists` to the extension file.  I also moved there the `addDeclEntry` function: I did not place it there before, since I thought that this was what causing the `lake build` issues.  I had not realized that the issue was the `olean` explosion.

adomani:
> > Actually, the CI step is already embedded in the tests: one of the files calls the `#check_assertions!` command with `import Mathlib`: the command is noisy if and only if there are non-existent declarations or imports.
> 
> I kinda liked the previous incarnation. I think this is good for an initial commit so I don't think we should go back now but the feedback in CI for failure was a bit earlier.

This is very easy: it just suffices to add a step that builds `test/AssertImported`, failing if there is any sort of output.  Let me know if you want me to add it back in this PR or in a follow up.

> Aside: why are we building the archive and counterexamples so high up in the workflow?

I do not know the reason for this, but both steps are relatively short.

In fact, I had it in the back of my mind to run `Archive`, `Counterexamples`, `tests` and `linting` in parallel, as soon as the cache has been checked.  Each would be its own separate workflow that downloads the cache and does its thing.  What stopped me from implementing this was that I could not find a way to make the github actions contingent on the success of a *different* action.  The issue might have been that the logic for this was supposed to be on `master` and could not be tested on a branch.

Mathlib/Util/AssertExistsExt.lean
```diff
@@ -0,0 +1,55 @@
...
+It extends the `AssertExists` environment extension with the data `isDecl, declName, mod`.
+This information is used to capture declarations and modules that are required to not
+exist/be imported at some point, but should eventually exist/be imported.
+-/
+def addDeclEntry {m : Type → Type} [MonadEnv m] (isDecl : Bool) (declName mod : Name) : m Unit :=
+  modifyEnv (assertExistsExt.addEntry · { isDecl := isDecl, givenName := declName, modName := mod })
+
+/-- `getSortedAssertExists env` returns the array of `AssertExists`, placing first all declarations,
+in alphabetical order, and then all modules, also in alphabetical order. -/
+def getSortedAssertExists (env : Environment) : Array AssertExists :=
```
mattrobball, line 51: Do you want this to live in the `Lean.Environment` namespace for access to dot notation? (This seems standard.) 

Optional suggestion: sorting could be a boolean parameter you pass. Either is fine to me.

adomani: I put `getSortedAssertExists` in the `Lean.Environment` namespace.

I did not do the optional sorting, since the `HashSet`s produce random orderings and for both `#guard_msgs` and retrieving information about a specific declaration/import it is convenient to know where to look.

Mathlib/Util/AssertExistsExt.lean
```diff
@@ -0,0 +1,55 @@
...
+It extends the `AssertExists` environment extension with the data `isDecl, declName, mod`.
+This information is used to capture declarations and modules that are required to not
+exist/be imported at some point, but should eventually exist/be imported.
+-/
+def addDeclEntry {m : Type → Type} [MonadEnv m] (isDecl : Bool) (declName mod : Name) : m Unit :=
+  modifyEnv (assertExistsExt.addEntry · { isDecl := isDecl, givenName := declName, modName := mod })
+
+/-- `getSortedAssertExists env` returns the array of `AssertExists`, placing first all declarations,
+in alphabetical order, and then all modules, also in alphabetical order. -/
+def _root_.Lean.Environment.getSortedAssertExists (env : Environment) : Array AssertExists :=
```
mattrobball, line 51: Maybe end the namespace and start a new one? I this `_root_` is frowned upon.

mattrobball:
Thanks for bearing with me!

bors merge

adomani:
Thank you very much for the reviews: I really feel that the code improved a lot!

PR #16008, opened 2024-08-20
[Merged by Bors] - feat(ModelTheory): Universal and existential formulas

awainverse:
Defines universal and existential formulas
Proves preservation results for universal and existential formulas
As an application, shows that substructures of models of the theory of linear orders are also models of the theory of linear orders.

---

- [x] depends on: #16107 [optional extra text]
- [x] depends on: #16108 [optional extra text]


awainverse:
Right now, it seems that showing that a specific formula has a specific complexity level requires manually recreating the definition of the formula.
I'd like that to be more automated, perhaps with instance resolution, and I'd happily take suggestions on that.

awainverse:
I've realized that the preservation results here should probably be proven at the level of, say, `StrongHomClass` - see #16107 .

Mathlib/ModelTheory/Complexity.lean
```diff
@@ -360,10 +386,80 @@ lemma IsQF.realize_embedding {φ : L.BoundedFormula α n} (hQF : φ.IsQF)
...
+
+lemma isAtomic (r : L.Relations l) (ts : Fin l → L.Term (α ⊕ (Fin n))) :
+    IsAtomic (r.boundedFormula ts) := IsAtomic.rel r ts
+
+lemma isQF (r : L.Relations l) (ts : Fin l → L.Term (α ⊕ (Fin n))) :
+    IsQF (r.boundedFormula ts) := (r.isAtomic ts).isQF
+
+variable (r : L.Relations 2)
+
+protected lemma reflexive_isUniversal : r.reflexive.IsUniversal :=
```
YaelDillies, line 443:
```suggestion
protected lemma isUniversal_reflexive : r.reflexive.IsUniversal :=
```
and below

YaelDillies:
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #16011, opened 2024-08-20
[Merged by Bors] - feat(Logic/Equiv): involutions are inverses of themselves

Rida-Hamadani:
This lemma is ported from to lean4 from Kyle Miller's [repo](https://github.com/kmill/lean-graphs). It is used while proving a lemma about combinatorial maps.

---

### DIFF
--- a/Mathlib/Logic/Equiv/Basic.lean
+++ b/Mathlib/Logic/Equiv/Basic.lean
@@ -1532,6 +1532,15 @@ theorem toPerm_symm {f : α → α} (h : Involutive f) : (h.toPerm f).symm = h.t
 theorem toPerm_involutive {f : α → α} (h : Involutive f) : Involutive (h.toPerm f) :=
   h
 
+theorem leftInverse_iff {f g : α → α} (h : Involutive f) :
+    g.LeftInverse f ↔ g = f :=
+  ⟨fun hg ↦ funext fun x ↦ by rw [← h x, hg, h], fun he ↦ he ▸ h.leftInverse⟩
+
+@[simp]
+theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f := by
+  ext x
+  exact congrFun (h.leftInverse_iff.mp f.left_inv) x
+
 end Function.Involutive
 
 theorem PLift.eq_up_iff_down_eq {x : PLift α} {y : α} : x = PLift.up y ↔ x.down = y :=
--- a/Mathlib/Logic/Equiv/Basic.lean
+++ b/Mathlib/Logic/Equiv/Basic.lean
@@ -1533,9 +1533,8 @@ theorem toPerm_involutive {f : α → α} (h : Involutive f) : Involutive (h.toP
   h
 
 @[simp]
-theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f := by
-  ext x
-  exact congrFun (h.leftInverse_iff.mp f.left_inv) x
+theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f :=
+  DFunLike.coe_injective (h.leftInverse_iff.mp f.left_inv)
 
 end Function.Involutive
 
--- a/Mathlib/Logic/Equiv/Basic.lean
+++ b/Mathlib/Logic/Equiv/Basic.lean
@@ -1532,10 +1532,6 @@ theorem toPerm_symm {f : α → α} (h : Involutive f) : (h.toPerm f).symm = h.t
 theorem toPerm_involutive {f : α → α} (h : Involutive f) : Involutive (h.toPerm f) :=
   h
 
-theorem leftInverse_iff {f g : α → α} (h : Involutive f) :
-    g.LeftInverse f ↔ g = f :=
-  ⟨fun hg ↦ funext fun x ↦ by rw [← h x, hg, h], fun he ↦ he ▸ h.leftInverse⟩
-
 @[simp]
 theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f := by
   ext x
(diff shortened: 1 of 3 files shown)


Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -291,6 +291,13 @@ theorem eq_symm_apply {α β} (e : α ≃ β) {x y} : y = e.symm x ↔ e y = x :
 theorem symm_bijective : Function.Bijective (Equiv.symm : (α ≃ β) → β ≃ α) :=
   Function.bijective_iff_has_inverse.mpr ⟨_, symm_symm, symm_symm⟩
 
+@[simp]
+theorem symm_eq_self_of_involution (f : Equiv.Perm α) (h : Function.Involutive f) : f.symm = f := by
+  ext x
+  calc f.symm x = f.symm (f (f x)) := by rw [h x]
```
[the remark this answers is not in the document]
Rida-Hamadani: I've found something similar, is this okay?

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -291,6 +291,11 @@ theorem eq_symm_apply {α β} (e : α ≃ β) {x y} : y = e.symm x ↔ e y = x :
 theorem symm_bijective : Function.Bijective (Equiv.symm : (α ≃ β) → β ≃ α) :=
   Function.bijective_iff_has_inverse.mpr ⟨_, symm_symm, symm_symm⟩
 
+@[simp]
```
eric-wieser, line 294: It's not clear to me that this should be a `simp` lemma; can you find another example of a simp lemma assuming `Involutive`? We had trouble around this kind of thing long ago, but perhaps the reason is no longer relevant.

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -291,6 +291,11 @@ theorem eq_symm_apply {α β} (e : α ≃ β) {x y} : y = e.symm x ↔ e y = x :
...
+theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Function.Involutive f) : f.symm = f := by
```
eric-wieser, line 295: Could you add this related result:
```lean
theorem Function.Involutive.leftInverse_iff {f g : α → α} (h : Function.Involutive f) :
    Function.LeftInverse g f ↔ g = f :=
  ⟨fun hg => funext fun x => by rw [← h x, hg, h], fun he => he ▸ h.leftInverse⟩
```
which can then be used to prove your result as `h.leftInverse_iff.2 f.left_inv` or similar.

Mathlib/Logic/Equiv/Defs.lean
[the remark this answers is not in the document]
grunweg: Thanks - looks good. (But please follow Eric's suggestion below, which is even better :-))

Mathlib/Logic/Equiv/Defs.lean
Rida-Hamadani: I've found similar `simp` lemmas at `Equiv/Basic.lean`, such as `toPerm_symm`. This lemma is still nice to have since `toPerm_symm` isn't exactly what I want, so I will move it to that file.

Mathlib/Logic/Equiv/Defs.lean
Rida-Hamadani: I tried to make `h.leftInverse_iff.2 f.left_inv` work, but I couldn't improve on the version provided in the last commit. In particular, I couldn't write it in term mode (even when using `funext`).

Mathlib/Logic/Equiv/Basic.lean
```diff
@@ -1532,6 +1532,15 @@ theorem toPerm_symm {f : α → α} (h : Involutive f) : (h.toPerm f).symm = h.t
...
   h
 
+theorem leftInverse_iff {f g : α → α} (h : Involutive f) :
+    g.LeftInverse f ↔ g = f :=
+  ⟨fun hg ↦ funext fun x ↦ by rw [← h x, hg, h], fun he ↦ he ▸ h.leftInverse⟩
+
+@[simp]
+theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f := by
+  ext x
+  exact congrFun (h.leftInverse_iff.mp f.left_inv) x
```
eric-wieser, line 1542: untested:
```suggestion
theorem symm_eq_self_of_involutive (f : Equiv.Perm α) (h : Involutive f) : f.symm = f :=
  DFunLike.ext (h.leftInverse_iff.mp f.left_inv)
```

Mathlib/Logic/Equiv/Basic.lean
```diff
@@ -1532,6 +1532,15 @@ theorem toPerm_symm {f : α → α} (h : Involutive f) : (h.toPerm f).symm = h.t
 theorem toPerm_involutive {f : α → α} (h : Involutive f) : Involutive (h.toPerm f) :=
   h
 
+theorem leftInverse_iff {f g : α → α} (h : Involutive f) :
+    g.LeftInverse f ↔ g = f :=
+  ⟨fun hg ↦ funext fun x ↦ by rw [← h x, hg, h], fun he ↦ he ▸ h.leftInverse⟩
```
eric-wieser, line 1537: This one probably belongs next to `Involutive.leftInverse` in an earlier file.

Command-Master:
!bench

Mathlib/Logic/Equiv/Defs.lean
urkud: The simplifier will try to apply `toPerm_symm` only if it has `(h.toPerm f).symm`, while your lemma will match on every `Equiv.symm (e : α ≃ α)`.

Rida-Hamadani: Ah, I thought it will check for the involutive condition too, I will remove the simp.

urkud: It will match on `Equiv.symm _`, then try to prove `Involutive`.

urkud:
Thanks! :tada:
bors merge

PR #16015, opened 2024-08-21
[Merged by Bors] - feat(UpperHalfPlane): weaken assumptions of multiple atImInfty lemmas

grhkm21:
In this PR, I weaken some assumptions so that they can be applied more widely (more specifically, to my use case :) ) I also renamed some lemmas because they sounded weird.

---


Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
```diff
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2022 Chris Birkbeck. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Chris Birkbeck, David Loeffler
+Authors: Chris Birkbeck, David Loeffler, Gareth Ma
```
grhkm21, line 4: By the way, I don't know if making such small changes qualify as a co-author, and I'm happy to remove this.

grhkm21:
Also, what is the reason that `CuspForm` doesn't extend from `ModularForm`? I know the zero_at_infty'` and `bdd_at_infty'` properties duplicate but having CuspForm inherit ModularForm should reduce some code duplication.

Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
CBirkbeck: I usually wouldnt for something this small (otherwise every file would have Yael's name as an author lol). But I dont know what the rules are about this, I essentially would never do it unless substantially changed/expanded the file.

Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
```diff
@@ -53,18 +53,35 @@ def zeroAtImInftySubmodule (α : Type*) [NormedField α] : Submodule α (ℍ →
...
+
+variable {α : Type*} {f g : ℍ → α}
+
+nonrec theorem add [SeminormedAddCommGroup α] (hf : IsBoundedAtImInfty f)
+    (hg : IsBoundedAtImInfty g) : IsBoundedAtImInfty (f + g) := by
+  simpa only using hf.add hg
+
+nonrec theorem sub [SeminormedAddCommGroup α] (hf : IsBoundedAtImInfty f)
+    (hg : IsBoundedAtImInfty g) : IsBoundedAtImInfty (f - g) := by
+  simpa only using hf.sub hg
```
CBirkbeck, line 66: do we need this given we have: `BoundedAtFilter.add` etc?

Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
```diff
@@ -53,18 +53,35 @@ def zeroAtImInftySubmodule (α : Type*) [NormedField α] : Submodule α (ℍ →
...
+end IsBoundedAtImInfty
 
-theorem bounded_mem (f : ℍ → ℂ) :
-    IsBoundedAtImInfty f ↔ ∃ M A : ℝ, ∀ z : ℍ, A ≤ im z → abs (f z) ≤ M := by
+theorem isBoundedAtImInfty_iff {α : Type*} [Norm α] {f : ℍ → α} :
+    IsBoundedAtImInfty f ↔ ∃ M A : ℝ, ∀ z : ℍ, A ≤ im z → ‖f z‖ ≤ M := by
   simp [IsBoundedAtImInfty, BoundedAtFilter, Asymptotics.isBigO_iff, Filter.Eventually,
     atImInfty_mem]
 
-theorem zero_at_im_infty (f : ℍ → ℂ) :
```
CBirkbeck, line 65: this (and the others) looks like an old lean 3 name that got forgotten in the port.

Mathlib/NumberTheory/ModularForms/Basic.lean
```diff
@@ -128,6 +128,11 @@ protected def CuspForm.copy (f : CuspForm Γ k) (f' : ℍ → ℂ) (h : f' = ⇑
   holo' := h.symm ▸ f.holo'
   zero_at_infty' A := h.symm ▸ f.zero_at_infty' A
 
+def CuspForm.toModularForm (f : CuspForm Γ k) : ModularForm Γ k :=
```
CBirkbeck, line 131: I dont thik we should need this since we have `instance (priority := 99) [FunLike F ℍ ℂ] [CuspFormClass F Γ k] : ModularFormClass F Γ k where` further down the file.

CBirkbeck:
> Also, what is the reason that `CuspForm` doesn't extend from `ModularForm`? I know the zero_at_infty'`and`bdd_at_infty'` properties duplicate but having CuspForm inherit ModularForm should reduce some code duplication.

I think that that duplication is one of the reasons why its not an extends. The reason for having the class versions of the definitions is to solve this code duplication.

Mathlib/NumberTheory/ModularForms/Basic.lean
grhkm21: I see, I never looked too closely at the `Class` things, but I should. I've also removed this then

Ruben-VandeVelde:
When renaming existing results, please add a deprecated alias. (If you start typing "deprecated" in vs code, it should suggest an autocompletion.)

grhkm21:
Hope I'm doing this right, I'm using the neovim plugin for lean and it doesn't seem to show any code action for deprecation, but I copied the code from other parts of Mathlib.

grhkm21:
Hi, bump. Can someone review this?

CBirkbeck:
LGTM!

grhkm21:
Oops, I saw the LGTM part in my email inbox but not the "fix build error" part. This should do it.

Mathlib/NumberTheory/ModularForms/Basic.lean
```diff
@@ -230,7 +230,7 @@ def mul {k_1 k_2 : ℤ} {Γ : Subgroup SL(2, ℤ)} (f : ModularForm Γ k_1) (g :
     -- Porting note: was `by simpa using ...`
     -- `mul_slash_SL2` is no longer a `simp` and `simpa only [mul_slash_SL2] using ...` failed
     rw [SlashInvariantForm.coe_mul, mul_slash_SL2]
-    exact (f.bdd_at_infty' A).mul (g.bdd_at_infty' A)
+    exact Filter.BoundedAtFilter.mul (f.bdd_at_infty' A) (g.bdd_at_infty' A)
```
loefflerd, line 233: I looked at that and thought, "why change the dot-notation back to the clumsier older notation"? On checking out the code and playing with it, I realised that the deprecated alias `IsBoundedAtImInfty.mul` is "masking" the more general lemma `BoundedAtFilter.mul` so it doesn't get found by namespace search. This, to me, seems like a good reason _not_ to add a deprecation for the old lemma. Then this proof can become simply
```lean
  bdd_at_infty' A := by simpa only [coe_mul, mul_slash_SL2] 
    using (f.bdd_at_infty' A).mul (g.bdd_at_infty' A)```

File: Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean [hunk 6]
```diff
@@ -53,18 +53,24 @@ def zeroAtImInftySubmodule (α : Type*) [NormedField α] : Submodule α (ℍ →
...
+
+theorem isZeroAtImInfty_iff {α : Type*} [SeminormedAddGroup α] {f : ℍ → α} :
+    IsZeroAtImInfty f ↔ ∀ ε : ℝ, 0 < ε → ∃ A : ℝ, ∀ z : ℍ, A ≤ im z → ‖f z‖ ≤ ε :=
+  (atImInfty_basis.tendsto_iff Metric.nhds_basis_closedBall).trans <| by simp
+
+@[deprecated (since := "2024-08-27")] alias _root_.zero_at_im_infty := isZeroAtImInfty_iff
+
+theorem IsZeroAtImInfty.isBoundedAtImInfty {α : Type*} [SeminormedAddGroup α] {f : ℍ → α}
+    (hf : IsZeroAtImInfty f) : IsBoundedAtImInfty f :=
+  -- hf.boundedAtFilter requires a stronger condition than we do :)
```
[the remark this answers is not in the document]
loefflerd: (the proof is a one-liner, just do
```
theorem ZeroAtFilter.boundedAtFilter [SeminormedAddCommGroup β] {l : Filter α} {f : α → β}
    (hf : ZeroAtFilter l f) : BoundedAtFilter l f := 
  ((Asymptotics.isLittleO_one_iff _ ).mpr hf).isBigO
```
at line 74 of `ZeroAndBoundedAtFilter.lean`.)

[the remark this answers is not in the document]
grhkm21: Okay I *tried* weakning the assumptions of the general library lemma, but that will of course require also modifying the assumptions of the adjacent lemmas (or actually the entire section, at least those that can be weakened - which is majority actually), and it kind of has a cascading effect with 500+ lines of diff. I don't think it's appropriate for this PR. I will use your one-liner.

Mathlib/NumberTheory/ModularForms/Basic.lean
grhkm21: I am unable to get your code working even after un-deprecating the alias, so I tried an alternative, see if it's okay.

loefflerd: I tried it out before posting it, so it definitely does work.

Mathlib/Analysis/Complex/UpperHalfPlane/FunctionsBoundedAtInfty.lean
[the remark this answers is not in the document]
loefflerd: I don’t follow. The other surrounding lemmas in the filter API all work for seminorms, it’s literally only this one which currently requires a norm; so the fix should definitely be one line, not 500. I am away from my computer now, but I can look at this this evening.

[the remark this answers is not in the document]
loefflerd: Looking at this again: I had overlooked that you wanted to remove the **commutativity** assumption on `α`, not just to weaken from "norm" to "seminorm". This is going to be much more difficult, and I am very curious now what your intended application is. 

(My impression is that you can't do very much with a non-commutative (semi)normed group; in particular, you need commutativity in order for the norm to induce a topological-group structure.)

Mathlib/NumberTheory/ModularForms/Basic.lean
loefflerd: I just tried it again, and it works fine (after removing the section `BasicOperations` in the other file which you have subsequently added). 

Perhaps it was not clear from my old message that I meant "remove `IsBoundedAtImInfty.mul` and don't deprecate it", not "leave the lemma there and don't deprecate it".

loefflerd (commented):
I left some further comments. 

I think it's a good compromise to only modify the general asymptotics lemmas that you specifically need, and leave the general problem of removing commutativity elsewhere in the theory for a later PR.

grhkm21:
Hopefully this makes things pass :)

loefflerd:
LGTM now. One last thing: can you create an issue on GitHub for generalising the asymptotics API to drop commutative assumptions, and quote the issue number in the TODO comment in the code? 

maintainer delegate

riccardobrasca:
Thanks!

bors d+

Ruben-VandeVelde:
What happened to `IsBoundedAtImInfty.mul`?

loefflerd:
We concluded it wasn't needed, since `IsBoundedAtImInfty` is a special case of `IsBoundedAtFilter`, and removing `IsBoundedAtImInfty.mul` just un-masks the more general lemma `IsBoundedAtFilter.mul`. That was the reason for not adding a deprecation.

grhkm21:
Oh I misread the diff, ignore me

PR #9687, opened 2024-01-12
[Merged by Bors] - chore: cache: do not try to upload files just downloaded

Kha:
If `put` in CI finds that an archive already exists in the local cache, we can assume it exists in the cloud cache as well. Should provide a good-enough avoidance of repeatedly uploading the same archive on hosts like Cloudflare that do not support no-overwrite and still save a ton of write requests on Azure.

PR #16020, opened 2024-08-21
feat: compare PR `olean`s size with `master`

adomani:
Adds two CI steps:
* `print the sizes of the oleans` that prints the sizes of all the folders containing `Mathlib` `.olean`s;
* `compare oleans` that compares the sizes of the previous step with the corresponding sizes on `master`.

In the test runs, the two steps have taken at most 3 seconds combined (the first is virtually instantaneous, the second one depends on `curl` to find a job id and on `gh` to retrieve the logs of a previous CI run -- everything else appears to be negligible).

This hopefully helps finding out if some PR is bloating up the `.olean`s.

See [this Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Error.3A.20No.20space.20left.20on.20device/near/463792355) for a PR of mine that prompted this check.

---


.github/workflows/build.yml
```diff
@@ -205,6 +205,7 @@ jobs:
             rm -rf .lake/build/lib/Mathlib
             lake exe cache get || (sleep 1; lake exe cache get)
             lake build --no-build
+            printf '.lake size: %s\n' "$(du -d0 -h .lake)"
```
grunweg, line 208: According to the documentation, `du --summarize --human-readable .lake` is equivalent. I think that is slightly more readable; what do you think?

grunweg (commented):
Looks reasonable and presumably doesn't hurt; thanks for adding this! I have one comment about readability.
maintainer merge

.github/workflows/build.yml
adomani: I just pushed a commit with the long names, thanks!

mattrobball:
This just prints the olean size correct? 

I think a small Lean package that compares the size of the oleans locally to the size of the oleans on master would be very welcome.

mattrobball:
Maybe this should be broken out into a script. It would be nice if 
- there is a overall check that fails if the total oleans increase by more than 2% 
- there is a file by file check that fails if an individual olean increases by more than 10%

Percents can be moved a bit.

mattrobball:
The check should run before anything is executed with the oleans.

adomani:
How about I extract the current script in the workflow file, to a script in `scripts/` and then we can discuss percentages?

adomani:
I randomly replaced 2 with 3 and 9 with 8 on the sizes of my oleans and generated this difference summary (using 32% difference as a threshold for printing):

%-32.81%  (-1000)   Mathlib/Logic/Function
-32.64%   (-1000)   Mathlib/Algebra/Order/Ring/Unbundled
-32.72%   (-1000)   Mathlib/Algebra/Group/Pi
-32.64%   (-1000)   Mathlib/Data/Sigma
-32.47%   (-100)    Mathlib/RingTheory/LocalRing/MaximalIdeal
-32.14%   (-99000)  Mathlib/Tactic
-32.51%   (-1000)   Mathlib/CategoryTheory/Endofunctor
-33.20%   (-1000)   Mathlib/LinearAlgebra/Matrix/Determinant
-32.13%   (-1000)   Mathlib/RingTheory/IntegralClosure/IsIntegralClosure
-32.89%   (-100)    Mathlib/RingTheory/Int
-32.12%   (-9900)   Mathlib/RingTheory/Ideal
-32.18%   (-1000)   Mathlib/Init/Data
-33.33%   (-100)    Mathlib/Data/Sym/Sym2
-32.85%   (-1000)   Mathlib/RingTheory/Trace
-32.05%   (-100)    Mathlib/Topology/Algebra/Constructions
-32.13%   (-1000)   Mathlib/Algebra/Order/Antidiag
-32.18%   (-1000)   Mathlib/Algebra/Group/Equiv
-32.89%   (-100)    Mathlib/MeasureTheory/Category
-32.89%   (-1000)   Mathlib/Topology/Algebra/Valued
-32.98%   (-1000)   Mathlib/CategoryTheory/EffectiveEpi
-32.59%   (-1000)   Mathlib/CategoryTheory/Limits/Shapes/NormalMono


Does this seem like useful information?

bryangingechen:
This might be too far afield from the purpose of this PR, but should we add an additional cleanup step at the very end of our CI to delete the oleans from the runner regardless of the size just so that they don't end up accumulating over time? (Is that what `lake exe cache clean!` does?)

mattrobball:
I think the table is nice. The main thing I worry about are that some individual olean got unusually larger. 

@bryangingechen Yes, I think `clean!` deletes the olean cache.

adomani:
I pushed the table, with a 5% cutoff.  I think that this will be fine-tunes better after seeing it in actual examples.

scripts/olean_comparison.sh
```diff
@@ -0,0 +1,51 @@
...
+  rm -rf .lake/build/lib/Mathlib
+  >&2 lake exe cache clean!
+  >&2 lake exe cache get
+  du .lake/build/lib/Mathlib | sed "s=^=${1} =; s=\t= =g"
+}
+
+# Because the `lean-pr-testing-NNNN` branches use toolchains that are "updated in place"
+# the cache mechanism is unreliable, so we don't test it if we are on such a branch.
+if [[ ! $(cat lean-toolchain) =~ ^leanprover/lean4-pr-releases:pr-release-[0-9]+$ ]]; then
+  git fetch origin master --depth 1
```
Vtec234, line 20: Am I reading correctly that this:
- builds the PR
- uploads it to cache
- deletes the built oleans
- fetches ones cached for master
- measures their size
- re-fetches the oleans from the PR
- measures their size

?

That seems a little wasteful compared to the following alternative:
- build the PR
- measure the size of oleans
- upload them to cache (and possibly don't if the measured size exceeds some threshold; 10G seems quite safe as mathlib is at 3.7G now)
- delete the built oleans
- fetch ones cached for master
- measure their size

In fact, the fetch-from-master step could also be skipped as one can already tell what the size there is by looking at the CI logs for the latest build on master.

mattrobball: I would prefer postponing uploading the new cache until the olean sizes are within parameters.

adomani: If I can find the master sizes somewhere, that would be great!

I think that the final oleans that are left should be the ones from the branch, since the later steps (building Archive, Counterexamples, Shake and Lint) use them.

adomani: @Vtec234 where can I retrieve the information about the sizes of `master`'s oleans?

adomani: Ok, renaming the .lake folder out of the way before downloading `master`'s oleans already reduced by half the time of the step: now it takes about 30 seconds.  If I did not have to download `master`'s oleans, the step would be almost instantaneous, since it is just a very easy text processing command, but already as is, I think that it is much better!

adomani: > I would prefer postponing uploading the new cache until the olean sizes are within parameters. 

Matt, do you want this to be part of this PR, or of a follow up one?

mattrobball: As long as it eventually happens, either is fine

Vtec234: > @Vtec234 where can I retrieve the information about the sizes of `master`'s oleans?

The sizes can be retrieved from the olean cache on Azure, but it needs a bit of logic to figure out exactly which sizes to fetch. A command for that could be added to `lake exe cache`.

Or, you could output all the sizes in CI logs and then fetch them from there in other CI builds, but it would be quite a lot of output.

Vtec234: After looking into how Cache works, the former option looks infeasible without larger changes. This is because the cache itself does not store an association between mathlib4 commits and the oleans that correspond to them. `cache commit` can do that, but it [is not currently ran](https://github.com/leanprover-community/mathlib4/blob/25040ca7ee3241a452efbd6b7ee2e64b56a85513/.github/workflows/build.yml.in#L171) in CI. If the command were re-enabled, it would become possible to list the oleans for a specific mathlib commit and add up their sizes. However, I don't understand what the 'integrate with put-unpacked' part is about.

Vtec234: CC #9687 @Kha

Vtec234: Ok, after some discussion with Mario the latter option is the only one that seems to make sense. But it's also an optimization, and you can probably get away with the current strategy of just downloading the cache.

semorrison:
I haven't followed carefully here, but let's be sure before merging this that we're not adding any meaningful time to CI just to do this check. Time in CI is precious. :-)

adomani:
Yes, I am aware of time in CI: currently, the new step takes 30 seconds, mostly due to checking out master and getting the oleans, just to find out about the sizes of the folders in `.lake`.

I am thinking about retrieving this data from *somewhere*, possibly adding a CI step on master that, after `lake build` calls `du .lake` and then using this, instead of all the steps above.

The main issue is the retrieval of the job_id of the latest CI run on master.

scripts/olean_comparison.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+oleansDir=.lake/build/lib/Mathlib
+# should be master
```
mattrobball, line 4: Is there anything preventing us from pointing this at master and then doing a few test runs with existing branches?

adomani: No, my idea is to test this now pointing it at the current branch, but, before merging the PR, point it at master, where the data will be available.

adomani:
I almost succeeded in setting this up, the main difficulty is that the hoskinson servers don't have gh installed (as opposed to the runners for lint style).  Apart from that, I have essentially a functioning workflow.  I hope to have time to finish it soon.

Vtec234:
> the hoskinson servers don't have gh installed 

We can install this if it helps you.

adomani:
Great, thanks a lot!

I've been using a different PR for testing, so I'll try this out there once I'm back at a computer and will let you know if I run into issues!

adomani:
@Vtec234 thank you so much for installing `gh`!  I now have a working prototype.

adomani:
The new workflow includes one step that computes the sizes of the current branch and then runs the comparison with the analogous results for master.

The initial print of the sizes of the *current* branch means that we can retrieve this data on `master` by looking at the logs for the latest master run.

Should this printout be a separate `- name:` entry in the workflow?  The script that I have does not do so.  I find it cleaner to have a separate `- name: olean size` step and then a `- name: compare oleans` step.  This adds more CI steps and I do not know if this is desirable or not.

adomani:
I forgot to mention: I've seen the new CI step take between 1 and 3 seconds.

adomani:
I tried the other option, with two CI steps:
* `print the sizes of the oleans` that just runs `du .lake/build/lib/Mathlib`;
* `compare oleans` that does the comparison.

It looked better, so I switched to this workflow in the PR: let me know if you prefer the alternative.

scripts/olean_comparison.sh
mattrobball: I think running tests with its final configuration is a good idea before merging.

adomani: What kind of tests would you like to see?  I could implement a version of this script in Bryan's fork of mathlib, where I could push this into master and then run it, on demand, on various commits to see the relative outcomes.  However, it would still be a CI on a different repository, even though it could compute information about the "real" mathlib4.

adomani: Or, I could convert this into a script that triggered by a comment (that would download master's `olean`s) and run the comparison.  That would probably still need to be merged into master to work, but would be much safer, since it would not be on the main CI build, so any bugs would just affect the comment workflow.

scripts/olean_comparison.sh
```diff
@@ -0,0 +1,96 @@
...
+# Because the `lean-pr-testing-NNNN` branches use toolchains that are "updated in place"
+# the cache mechanism is unreliable, so we don't test it if we are on such a branch.
+if [[ $(cat lean-toolchain) =~ ^leanprover/lean4-pr-releases:pr-release-[0-9]+$ ]]; then
+  printf 'No summary, since this is a pr testing branch\n'
+fi
+
+oleansDir=.lake/build/lib/Mathlib
+
+# should be master
+branch=adomani/CI_olean_size
```
adomani, line 49:
```suggestion
branch=master
```
This should be changed before merging: it is the branch that is used as a comparison branch *and* it should contain the CI action that logged the sizes of the `olean`s.

scripts/olean_comparison.sh
```diff
@@ -0,0 +1,96 @@
...
+
+Finally, the script runs the comparison between the "`master`" sizes and the sizes
+of the newly created oleans.
+It flags every folder that has a percentage change of at least 5% (either positive or negative).
+The percentage difference for the full `.olean`s folder is always printed, whether or not it
+exceeds the threshold.
+
+## Check
+
+When bors merges a batch into master, running this code show the net effect of the new code over
```
YaelDillies, line 34:
```suggestion
When bors merges a batch into master, running this code shows the net effect of the new code over
```

YaelDillies (commented):
@mattrobball , do you think you'll get to review this one ?

YaelDillies (commented):
I am going to trust you (and the linked Zulip conversation) that this is useful

maintainer delegate

mattrobball:
Do we have any tests run against `master`?

adomani:
I only ran the script manually on my local copy of mathlib.  The script retrieves the data of the oleans from the runs on `master`, so testing this before merging is a little tricky.

I will look again at the action and make sure that whether or not the script finds a value, the script does not error.  In any case, this might only be relevant when trying to merge this PR to master, since after that, each PR will run its tests against master and there it will find the sizes.

So, as much as someone can be confident that they understand how workflows work, I think that either bors will fail, since this workflow will not find the info it wants and complain, or it will work even without the data being present and then it will not be an issue.

adomani:
I am away from a computer for a couple of hours, but will do some more testing on Bryan's fork and report the outcome here.

mattrobball:
If another maintainer with more comfort about the setup feels good, please don't let me stop them :)

adomani:
No, I think that it is good to test this.  In fact, thinking about it, I would rather have a step in the main CI that simply computes the sizes of the oleans and then a separate CI workflow that does the analysis.

This would mean that `du` would have to fail, which is highly unlikely.  Everything else is dispensable.

adomani:
I extracted the CI step that only computes the sizes of the oleans from this PR in #19283.

Hopefully, that addition is uncontroversial and, if this one here is desirable, the one in #19283 is clearly a step forward!

I also hope that the olean comparison can then be moved to a CI workflow separate from the main build one, so that it will be even less of an issue/cause fewer problems to merge.

leanprover-community-bot-assistant:

PR #16022, opened 2024-08-21
[Merged by Bors] - chore: Deprecate unused relation classes

YaelDillies:
This is cruft from Lean 3 core that was never cleaned up

---


j-loreaux:
bors merge

vihdzp:
I agree with everything else, but are we sure `IsLinearOrder α r` is never going to come up?

YaelDillies:
Given that it hasn't in 7 years of mathlib existing...

PR #16024, opened 2024-08-21
[Merged by Bors] - chore: Don't simplify `Odd n` to `¬ Even n`, but the other way around

YaelDillies:
This is simply not simpler

From LeanAPAP

---

### DIFF
--- a/Mathlib/Algebra/Ring/Int.lean
+++ b/Mathlib/Algebra/Ring/Int.lean
@@ -97,19 +97,24 @@ lemma odd_iff : Odd n ↔ n % 2 = 1 where
 
 lemma not_odd_iff : ¬Odd n ↔ n % 2 = 0 := by rw [odd_iff, emod_two_ne_one]
 
+@[simp] lemma not_odd_iff_even : ¬Odd n ↔ Even n := by rw [not_odd_iff, even_iff]
+@[simp] lemma not_even_iff_odd : ¬Even n ↔ Odd n := by rw [not_even_iff, odd_iff]
+
+@[deprecated not_odd_iff_even (since := "2024-08-21")]
 lemma even_iff_not_odd : Even n ↔ ¬Odd n := by rw [not_odd_iff, even_iff]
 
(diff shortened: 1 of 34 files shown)


Mathlib/Algebra/Ring/Int.lean
```diff
@@ -97,19 +97,24 @@ lemma odd_iff : Odd n ↔ n % 2 = 1 where
...
+lemma even_or_odd (n : ℤ) : Even n ∨ Odd n := Or.imp_right not_even_iff_odd.1 <| em <| Even n
 
 lemma even_or_odd' (n : ℤ) : ∃ k, n = 2 * k ∨ n = 2 * k + 1 := by
   simpa only [two_mul, exists_or, Odd, Even] using even_or_odd n
 
 lemma even_xor'_odd (n : ℤ) : Xor' (Even n) (Odd n) := by
   cases even_or_odd n with
-  | inl h => exact Or.inl ⟨h, even_iff_not_odd.mp h⟩
-  | inr h => exact Or.inr ⟨h, odd_iff_not_even.mp h⟩
+  | inl h => exact Or.inl ⟨h, not_odd_iff_even.2 h⟩
```
grunweg, line 116: This is minor, and not something I will insist on, but: I do find using `.mp(r)` clearer in such situations.
If we agree to disagree/not fuss about this as this is too minor, then please do not make such such unrelated changes.

grunweg (commented):
Thanks for doing this; I agree the new direction of the lemmas makes much more sense.
AIUI, this PR just flips the direction of these simp lemmas (but does not un-simp anything); can you please adjust the PR title to match this?

maintainer merge

Mathlib/Algebra/Ring/Int.lean
YaelDillies: Oh that's just a consequence of the grep I've done. It would be a bit complicated now to automatically find all the places where this change was not

jcommelin (commented):
Thanks :tada:

bors merge

PR #16025, opened 2024-08-21
[Merged by Bors] - refactor(LinearAlgebra/Eigenspace): unified definition of (max(gen)?)?Eigenspace

jcommelin:
---


Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,549 @@ open FiniteDimensional Set
 variable {K R : Type v} {V M : Type w} [CommRing R] [AddCommGroup M] [Module R M] [Field K]
   [AddCommGroup V] [Module K V]
 
+lemma ker_pow_le_ker_pow_of_le (f : End R M) (k l : ℕ) (hkl : k ≤ l) :
+    LinearMap.ker (f ^ k) ≤ LinearMap.ker (f ^ l) := by
+  rw [← pow_sub_mul_pow _ hkl]
+  exact LinearMap.ker_le_ker_comp _ _
```
ocfnash, line 64: This already exists as `LinearMap.iterateKer` (more precisely as `f.iterateKer.monotone hkl`).

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,549 @@ open FiniteDimensional Set
...
+
+/-- The submodule `unifEigenspace f μ k` for a linear map `f`, a scalar `μ`,
+and a number `k : ENat` is the kernel of `(f - μ • id) ^ k` if `k` is a natural number,
+or the union of all these kernels if `k = ∞`. -/
+def unifEigenspace (f : End R M) (μ : R) (k : ENat) : Submodule R M :=
+  ⨆ l : ℕ, ⨆ _ : l ≤ k, LinearMap.ker ((f - μ • 1) ^ l)
+
+lemma mem_unifEigenspace {f : End R M} {μ : R} {k : ENat} {x : M} :
```
ocfnash, line 72: I quite like using the notation:
```suggestion
lemma mem_unifEigenspace {f : End R M} {μ : R} {k : ℕ∞} {x : M} :
```
(also elsewhere)

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,549 @@ open FiniteDimensional Set
...
+lemma ker_pow_le_ker_pow_of_le (f : End R M) (k l : ℕ) (hkl : k ≤ l) :
+    LinearMap.ker (f ^ k) ≤ LinearMap.ker (f ^ l) := by
+  rw [← pow_sub_mul_pow _ hkl]
+  exact LinearMap.ker_le_ker_comp _ _
+
+/-- The submodule `unifEigenspace f μ k` for a linear map `f`, a scalar `μ`,
+and a number `k : ENat` is the kernel of `(f - μ • id) ^ k` if `k` is a natural number,
+or the union of all these kernels if `k = ∞`. -/
+def unifEigenspace (f : End R M) (μ : R) (k : ENat) : Submodule R M :=
+  ⨆ l : ℕ, ⨆ _ : l ≤ k, LinearMap.ker ((f - μ • 1) ^ l)
```
ocfnash, line 70: I'd be tempted to bundle in the monotonicity:
```suggestion
def unifEigenspace (f : End R M) (μ : R) : ℕ∞ →o Submodule R M where
  toFun k := ⨆ l : ℕ, ⨆ _ : l ≤ k, LinearMap.ker ((f - μ • 1) ^ l)
  monotone' k l hkl x := by
    sorry
```

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,549 @@ open FiniteDimensional Set
...
+lemma unifEigenspace_eq_iSup_unifEigenspace_nat (f : End R M) (μ : R) (k : ENat) :
+    f.unifEigenspace μ k = ⨆ l : {l : ℕ // l ≤ k}, f.unifEigenspace μ l := by
+  rw [unifEigenspace]
+  simp only [iSup_subtype, unifEigenspace_nat]
+
+lemma unifEigenspace_one {f : End R M} {μ : R} :
+    f.unifEigenspace μ 1 = LinearMap.ker (f - μ • 1) := by
+  rw [← Nat.cast_one, unifEigenspace_nat, pow_one]
+
+lemma mem_unifEigenspace_one {f : End R M} {μ : R} {x : M} :
```
ocfnash, line 119: Can this be `simp`?
```suggestion
@[simp] lemma mem_unifEigenspace_one {f : End R M} {μ : R} {x : M} :
```

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,549 @@ open FiniteDimensional Set
...
+    x ∈ f.unifEigenspace μ 1 ↔ f x = μ • x := by
+  rw [unifEigenspace_one, LinearMap.mem_ker, LinearMap.sub_apply,
+    sub_eq_zero, LinearMap.smul_apply, LinearMap.one_apply]
+
+lemma mem_unifEigenspace_zero {f : End R M} {μ : R} {x : M} :
```
ocfnash, line 124: And this?
```suggestion
@[simp] lemma mem_unifEigenspace_zero {f : End R M} {μ : R} {x : M} :
```

ocfnash (commented):
Just a few superficial remarks for now

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
jcommelin: simp can prove this one using the lemma below it.

Mathlib/Order/Hom/Basic.lean
```diff
@@ -1171,6 +1171,12 @@ theorem coe_toDualTopEquiv_eq [LE α] :
     (WithBot.toDualTopEquiv : WithBot αᵒᵈ → (WithTop α)ᵒᵈ) = toDual ∘ WithBot.ofDual :=
   funext fun _ => rfl
 
+/-- The coercion `α → WithBot α` bundled as monotone map. -/
+@[simps]
+def coeOrderHom {α : Type*} [Preorder α] : α →o WithBot α where
```
b-mehta, line 1176: Can/should we bundle this as a `↪o` instead?

b-mehta (approved):
LGTM but I don't know this part of the library that well!

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,560 @@ open FiniteDimensional Set
...
+  use max m n
+  dsimp
+  constructor <;> apply (f - μ • 1).iterateKer.monotone <;> simp
+
+private lemma unifEigenspace_aux_mono (f : End R M) (μ : R) :
+    Monotone (f.unifEigenspace_aux μ) := by
+  intro k l hkl x
+  rw [mem_unifEigenspace_aux, mem_unifEigenspace_aux]
+  rintro ⟨n, hn, hx⟩
+  use n, hn.trans hkl, hx
```
b-mehta, line 80:
```suggestion
private lemma mem_unifEigenspace_aux {f : End R M} {μ : R} {k : ℕ∞} {x : M} :
    x ∈ f.unifEigenspace_aux μ k ↔ ∃ l : ℕ, l ≤ k ∧ x ∈ LinearMap.ker ((f - μ • 1) ^ l) := by
  have : Nonempty {l : ℕ // l ≤ k} := ⟨⟨0, zero_le _⟩⟩
  rw [unifEigenspace_aux, iSup_subtype', Submodule.mem_iSup_of_directed]
  · simp only [LinearMap.mem_ker, Subtype.exists, exists_prop]
  exact Monotone.directed_le fun m n h => by simpa using (f - μ • 1).iterateKer.monotone h

private lemma unifEigenspace_aux_mono (f : End R M) (μ : R) :
    Monotone (f.unifEigenspace_aux μ) := fun _ _ hkl => biSup_mono fun _ hi => hi.trans hkl
```

I couldn't resist golfing these a bit - entirely optional if you want to use it.

Mathlib/Order/Hom/Basic.lean
jcommelin: Good point. Done

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,543 @@ open FiniteDimensional Set
...
+
+lemma mem_unifEigenspace {f : End R M} {μ : R} {k : ℕ∞} {x : M} :
+    x ∈ f.unifEigenspace μ k ↔ ∃ l : ℕ, l ≤ k ∧ x ∈ LinearMap.ker ((f - μ • 1) ^ l) :=
+  mem_unifEigenspace_aux
+
+lemma unifEigenspace_mono (f : End R M) (μ : R) : Monotone (f.unifEigenspace μ) := by
+  intro k l hkl x
+  rw [mem_unifEigenspace, mem_unifEigenspace]
+  rintro ⟨n, hn, hx⟩
+  use n, hn.trans hkl, hx
```
ocfnash, line 90: This should go as it is bundled into the type.
```suggestion
```
(in other words, this is already available via `OrderHom.monotone`)

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,543 @@ open FiniteDimensional Set
...
-    (hμ : f.HasGenEigenvalue μ k) : f.HasEigenvalue μ := by
-  intro contra; apply hμ
-  erw [LinearMap.ker_eq_bot] at contra ⊢; rw [LinearMap.coe_pow]
-  exact Function.Injective.iterate contra k
+    (hμ : f.HasGenEigenvalue μ k) : f.HasEigenvalue μ :=
+  hμ.lt zero_lt_one
 
 /-- Generalized eigenvalues are actually just eigenvalues. -/
 @[simp]
 theorem hasGenEigenvalue_iff_hasEigenvalue {f : End R M} {μ : R} {k : ℕ} (hk : 0 < k) :
```
ocfnash, line 589: This lemma makes it clear that the predicate `Module.End.HasGenEigenvalue` should not exist. It would be nice to excise it (though not necessarily as part of these changes).

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
```diff
@@ -58,227 +58,543 @@ open FiniteDimensional Set
...
-    _ = LinearMap.ker (f - b⁻¹ • algebraMap K (End K V) a) := rfl
-    _ = LinearMap.ker (b • (f - b⁻¹ • algebraMap K (End K V) a)) := by
-        rw [LinearMap.ker_smul _ b hb]
-    _ = LinearMap.ker (b • f - algebraMap K (End K V) a) := by rw [smul_sub, smul_inv_smul₀ hb]
+  unifEigenspace_div f a b hb
 
 /-- The generalized eigenspace for a linear map `f`, a scalar `μ`, and an exponent `k ∈ ℕ` is the
 kernel of `(f - μ • id) ^ k`. (Def 8.10 of [axler2015]). Furthermore, a generalized eigenspace for
 some exponent `k` is contained in the generalized eigenspace for exponents larger than `k`. -/
 def genEigenspace (f : End R M) (μ : R) : ℕ →o Submodule R M where
```
ocfnash, line 488: Why keep this?

I would have expected that we drop this, and that what is here called `unifEigenspace` would take this name.

ocfnash (approved):
Some minor comments but LGTM.

There is certainly room for further clean up but this is a big step in the right direction.

Thanks.

bors d+

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
jcommelin: I would be happy to do this in a follow-up PR.

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
jcommelin: That is definitely my plan. But I would prefer to do it in multiple steps.
First extend the API, then deprecate the old one, then rename.

Mathlib/LinearAlgebra/Eigenspace/Basic.lean
jcommelin: Yes! Thanks for catching this

Issue #1074, opened 2022-12-16
`to_additive` feature requests


fpvandoorn:
This is a single issue for new features or feature requests for `to_additive`, that I'll try to keep up-to-date. Feel free to comment with other suggestions
* [ ] Reorder arguments using an arbitrary permutation. This can be used to fix some `to_additive` connections in #1056
* [ ] Improve output of `to_additive?`, showing the type of the generated declaration (and not the body for lemmas)
* [ ] Even more reorder features, so that we can reorder `Pow.mk` to `SMul.mk` (requires re-ordering the function arguments in a particular argument)

YaelDillies:
`to_additive` does not seem to support the auxiliary lemmas generated by `norm_cast`: See #8916

fpvandoorn:
Can you please minimize the example?

YaelDillies:
Is this minimal enough?
```
import Mathlib.Data.Finset.Pointwise

open scoped Pointwise

@[to_additive]
lemma mul_foo {α : Type*} [DecidableEq α] [Mul α] (s t : Finset α) :
    (↑(s * t) : Set α) = ↑s * ↑t := by norm_cast
/-
application type mismatch
  Mathlib.Data.Finset.Pointwise._auxLemma.9
argument has type
  Add α
but function has type
  ∀ [inst : Mul α] (s t : Finset α), ↑s * ↑t = ↑(s * t)
-/
```

fpvandoorn:
Yes, thanks!
This was actually already a [todo](https://github.com/leanprover-community/mathlib4/blob/7be3f9939fab1b0a8c8cfe8c4b1cdfcdaafc6c4b/Mathlib/Tactic/ToAdditive.lean#L949) in the code. I will see if I can find a more systematic way to additivize new declarations generated attributes.

fpvandoorn:
`norm_cast` handling is fixed in #14628

YaelDillies:
> * [ ]  noncomputable definitions in a noncomputable sections won't be translated properly unless you explicitly specify `noncomputable` (this is [feat: mark predefinitions that failed to compile as noncomputable leanprover/lean4#2610](https://github.com/leanprover/lean4/pull/2610))

This makes `to_additive` incompatible with `noncomputable section`. This has come up in #26774.

PR #16026, opened 2024-08-21
[Merged by Bors] - feat: toAdditive to transfer MatcherInfo data

nomeata:
so that printing the definitions shows `match … with` properly, and
meta code looking for matchers work.

Fixes an item on #1074

### DIFF
--- a/Mathlib/Tactic/ToAdditive/Frontend.lean
+++ b/Mathlib/Tactic/ToAdditive/Frontend.lean
@@ -13,6 +13,7 @@ import Mathlib.Lean.Name
 import Lean.Elab.Tactic.Ext
 import Lean.Meta.Tactic.Symm
 import Lean.Meta.Tactic.Rfl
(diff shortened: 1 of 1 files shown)


grunweg:
Thanks for fixing this! I guess/hope that a few workarounds for this issue can be removed now (search for #1074, item (8)). Would you like to perform this clean-up also?

nomeata:
> Would you like to perform this clean-up also?

I wouldn’t mind if someone else will do that :-)

grunweg:
Never mind, I was thinking about item 7: I don't see any obvious locations where this would simplify mathlib code.

nomeata:
This is also needed after https://github.com/leanprover/lean4/pull/4154 hits lean master, so if we merge it now then the nightly testing maintainers (@semorrison and @jcommelin) have less work with the next nightly.

semorrison:
Sold!

bors merge

PR #16028, opened 2024-08-21
[Merged by Bors] - chore(CI): skip full `cache get` if cache looks completely cold

nomeata:
if the current toolchain has never been built on the CI, then spending
close to 4mins to look for each mathlib file is wasteful. So this checks
the existence a cache file for some file in `batteries` (I picked
`Batteries.WF`). If that is not present, then very likely the current
toolchain has not been tested yet, and we can skip the full `lake
cache`.

This speeds up mostly new `pr-testing` CI branches, or after a push to
the lean PR.

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -163,9 +163,9 @@ jobs:
         id: get
         run: |
           lake exe cache clean
-          # Fail quickly if the cache is completely cold, by checking for Batteries.WF
-          lake exe cache get Batteries.WF
-          ! test -e ./.lake/packages/batteries/.lake/build/lib/Batteries/WF.olean || lake exe cache get
+          # Fail quickly if the cache is completely cold, by checking for Mathlib.Init
+          lake exe cache get Mathlib.Init
+          ! test -e .lake/build/lib/Mathlib/Init.olean || lake exe cache get
 
       - name: build mathlib
         id: build
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -163,7 +163,9 @@ jobs:
         id: get
         run: |
           lake exe cache clean
-          lake exe cache get
+          # Fail quickly if the cache is completely cold, by checking for Batteries.WF
+          lake exe cache get Batteries.WF
+          ! test -e ./.lake/packages/batteries/.lake/build/lib/Batteries/WF.olean || lake exe cache get
 
       - name: build mathlib
         id: build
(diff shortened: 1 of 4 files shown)


.github/workflows/bors.yml
```diff
@@ -163,7 +163,9 @@ jobs:
         id: get
         run: |
           lake exe cache clean
-          lake exe cache get
+          # Fail quickly if the cache is completely cold, by checking for Batteries.WF
+          lake exe cache get Batteries.WF
+          ! test -e ./.lake/packages/batteries/.lake/build/lib/Batteries/WF.olean || lake exe cache get
```
grunweg, line 168: You could also check for `Mathlib.Init`, now that this file exists. I think that would be even clearer, as that file is supposed to be imported everywhere. Can you test if that also works, and if yes, switch to it? Thanks!

grunweg (approved):
This looks worth having, thanks for doing this! I have left one suggestion below; I will maintainer merge with that addressed.

.github/workflows/bors.yml
adomani: I would be even bolder and look for a file further up the import chain: in 4 minutes, `lake build` has produced `olean`s probably around 100-200 files, so maybe testing for Mathlib.Data.Nat.Defs` is more efficient still.

adomani: Also, looking inside `Mathlib` has the advantage of speeding up PRs to `master` that normally would not have issues with files in dependencies.

nomeata: Hmm, let’s see.

If the toolchain has changed (the use case I am aiming for), any file serves well.

If the toolchain hasn’t changed, but `Mathlib.Data.Nat.Defs` or a file it depends on has changed, then you’d be ignoring possible cache hits for all files that don't depend on `Mathlib.Data.Nat.Defs`, in particular all of `batteries`.

The tradeoffs seem unclear to me, so I'd stick to something simple. `Mathlib.Init` is nice and declarative.

grunweg: I was wondering to this effect, and agree!

nomeata: Right, but that wasn’t the aim – I was aiming for the cold cache case.

If you use a module somewhere in between, then some PRs benefit, other get slower, and it’s basically a game of chance.

Ideally `cache get` would dynamically stop fetching when 404s are returned, but that seems out of scope for it's current architecture.

adomani:
Thanks: I think that this is an improvement!

maintainer merge

mattrobball:
bors merge

PR #16029, opened 2024-08-21
[Merged by Bors] - feat: `a ^ (m - n : ℤ) = a ^ m / a ^ n` for `m n : ℕ`

YaelDillies:
This is a very common special case of `zpow_sub` which is particularly useful for rewriting backwards without having to hit everything with `zpow_natCast`

From LeanAPAP

---


eric-wieser:
bors d+

I found your previous title confusing

YaelDillies:
Sure, I was trying to make it searchable but I guess that's not a huge concern

bors merge

eric-wieser:
I think github searches the PR description too, so perhaps that's a better place for wordy explanation without length limits.

PR #16030, opened 2024-08-21
[Merged by Bors] - feat: Polynomial.eraseLead_degree_lt

ChrisHughes24:
---

### DIFF
--- a/Mathlib/Algebra/Polynomial/EraseLead.lean
+++ b/Mathlib/Algebra/Polynomial/EraseLead.lean
@@ -173,6 +173,9 @@ theorem eraseLead_add_of_natDegree_lt_right {p q : R[X]} (pq : p.natDegree < q.n
 theorem eraseLead_degree_le : (eraseLead f).degree ≤ f.degree :=
   f.degree_erase_le _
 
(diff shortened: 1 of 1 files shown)


Mathlib/Algebra/Polynomial/EraseLead.lean
```diff
@@ -173,6 +173,9 @@ theorem eraseLead_add_of_natDegree_lt_right {p q : R[X]} (pq : p.natDegree < q.n
 theorem eraseLead_degree_le : (eraseLead f).degree ≤ f.degree :=
   f.degree_erase_le _
 
+theorem eraseLead_degree_lt (hf : f ≠ 0) : (eraseLead f).degree < f.degree :=
```
YaelDillies, line 176: I see that the naming convention is also direspected in the rest of the file, but this really should be
```suggestion
theorem degree_eraseLead_lt (hf : f ≠ 0) : (eraseLead f).degree < f.degree :=
```

YaelDillies:
Thanks!

maintainer merge

mattrobball:
bors merge

PR #16032, opened 2024-08-21
[Merged by Bors] - feat(GroupTheory/Perm): composition of permutations is 1 iff they are inverses 

Rida-Hamadani:
---

### DIFF
--- a/Mathlib/Algebra/Group/Basic.lean
+++ b/Mathlib/Algebra/Group/Basic.lean
@@ -645,6 +645,14 @@ theorem mul_eq_one_iff_inv_eq : a * b = 1 ↔ a⁻¹ = b := by
   rw [mul_eq_one_iff_eq_inv, inv_eq_iff_eq_inv]
 
 @[to_additive]
+theorem mul_eq_one_iff_eq_inv' : a * b = 1 ↔ b = a⁻¹ := by
+  rw [mul_eq_one_iff_inv_eq, eq_comm]
+
+@[to_additive]
+theorem mul_eq_one_iff_inv_eq' : a * b = 1 ↔ b⁻¹ = a := by
+  rw [mul_eq_one_iff_eq_inv, eq_comm]
+
+@[to_additive]
 theorem eq_inv_iff_mul_eq_one : a = b⁻¹ ↔ a * b = 1 :=
   mul_eq_one_iff_eq_inv.symm
 
--- a/Mathlib/Algebra/Group/Basic.lean
+++ b/Mathlib/Algebra/Group/Basic.lean
@@ -644,12 +644,12 @@ theorem mul_eq_one_iff_eq_inv : a * b = 1 ↔ a = b⁻¹ :=
 theorem mul_eq_one_iff_inv_eq : a * b = 1 ↔ a⁻¹ = b := by
   rw [mul_eq_one_iff_eq_inv, inv_eq_iff_eq_inv]
 
-/-- Variant of `mul_eq_one_iff_eq_inv` with swapped equality. --/
+/-- Variant of `mul_eq_one_iff_eq_inv` with swapped equality. -/
 @[to_additive]
 theorem mul_eq_one_iff_eq_inv' : a * b = 1 ↔ b = a⁻¹ := by
   rw [mul_eq_one_iff_inv_eq, eq_comm]
 
-/-- Variant of `mul_eq_one_iff_inv_eq` with swapped equality. --/
+/-- Variant of `mul_eq_one_iff_inv_eq` with swapped equality. -/
 @[to_additive]
 theorem mul_eq_one_iff_inv_eq' : a * b = 1 ↔ b⁻¹ = a := by
   rw [mul_eq_one_iff_eq_inv, eq_comm]
--- a/Mathlib/Algebra/Group/Basic.lean
+++ b/Mathlib/Algebra/Group/Basic.lean
@@ -644,10 +644,12 @@ theorem mul_eq_one_iff_eq_inv : a * b = 1 ↔ a = b⁻¹ :=
 theorem mul_eq_one_iff_inv_eq : a * b = 1 ↔ a⁻¹ = b := by
   rw [mul_eq_one_iff_eq_inv, inv_eq_iff_eq_inv]
 
+/-- Variant of `mul_eq_one_iff_eq_inv` with swapped equality. --/
 @[to_additive]
 theorem mul_eq_one_iff_eq_inv' : a * b = 1 ↔ b = a⁻¹ := by
   rw [mul_eq_one_iff_inv_eq, eq_comm]
 
+/-- Variant of `mul_eq_one_iff_inv_eq` with swapped equality. --/
 @[to_additive]
 theorem mul_eq_one_iff_inv_eq' : a * b = 1 ↔ b⁻¹ = a := by
   rw [mul_eq_one_iff_eq_inv, eq_comm]
--- a/Mathlib/GroupTheory/Perm/Basic.lean
+++ b/Mathlib/GroupTheory/Perm/Basic.lean
@@ -157,10 +157,10 @@ theorem symm_mul (e : Perm α) : e.symm * e = 1 :=
 theorem trans_eq_one_iff_eq_symm (e₁ e₂ : Equiv.Perm α) : e₁.trans e₂ = 1 ↔ e₁ = e₂.symm := by
   refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
   · ext x
-    have : e₁.trans e₂ x = x := by rw [h, coe_one, id_eq]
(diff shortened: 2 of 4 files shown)


Mathlib/GroupTheory/Perm/Basic.lean
```diff
@@ -154,6 +154,16 @@ theorem self_trans_inv (e : Perm α) : e.trans e⁻¹ = 1 :=
 theorem symm_mul (e : Perm α) : e.symm * e = 1 :=
   Equiv.self_trans_symm e
 
+theorem trans_eq_one_iff_eq_symm (e₁ e₂ : Equiv.Perm α) : e₁.trans e₂ = 1 ↔ e₁ = e₂.symm := by
```
eric-wieser, line 157: Two requests:

* I think this is really just a variant of `mul_eq_one_iff_eq_inv` (on general groups) with the multiplication the other way around. Can you add that lemma?
* Can you generalize this to `Equiv` rather than `Equiv.perm`?

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -301,6 +301,15 @@ theorem symm_bijective : Function.Bijective (Equiv.symm : (α ≃ β) → β ≃
...
 @[simp] theorem self_trans_symm (e : α ≃ β) : e.trans e.symm = Equiv.refl α := ext <| by simp
 
+theorem trans_eq_refl_iff_eq_symm (f : α ≃ β) (g : β ≃ α) :
+    f.trans g = Equiv.refl α ↔ f = g.symm := by
+  refine ⟨fun h ↦ ?_, fun h ↦ ?_⟩
+  · ext x
+    apply g.injective
+    rw [← trans_apply, h, refl_apply, apply_symm_apply]
+  · rw [h, ← self_trans_symm]
+    rfl
```
tb65536, line 311: Here's a golf if you put this at line 497 after `symm_comp_eq`:
```
theorem trans_eq_refl_iff_eq_symm (f : α ≃ β) (g : β ≃ α) :
    f.trans g = Equiv.refl α ↔ f = g.symm := by
  rw [← Equiv.coe_inj, coe_trans, coe_refl, ← eq_symm_comp, comp_id, Equiv.coe_inj]
```

Rida-Hamadani: This is nice, thank you.

test/RewriteSearch/Polynomial.lean
```diff
@@ -174,7 +174,7 @@ example {R : Type u} [Ring R] [Nontrivial R] (x : R) :
 #guard_msgs(drop info) in
 example {S : Type v} [Ring S] (c : S) :
     Polynomial.nextCoeff (Polynomial.X - Polynomial.C c) = -c := by
-  rw_search [-Polynomial.nextCoeff_X_sub_C]
```
Ruben-VandeVelde, line 177: Why did you make this change?

Rida-Hamadani: The test was failing, I will revert this change and wait for the CI to see if it still fails now.

Rida-Hamadani: The test still fails without this change, [507b3a5](https://github.com/leanprover-community/mathlib4/pull/16032/commits/507b3a5cad74511cfeb084e9b1dab0145654a75c).

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -494,6 +494,10 @@ theorem eq_symm_comp {α β γ} (e : α ≃ β) (f : γ → α) (g : γ → β)
 theorem symm_comp_eq {α β γ} (e : α ≃ β) (f : γ → α) (g : γ → β) : e.symm ∘ g = f ↔ g = e ∘ f :=
   ((Equiv.refl γ).arrowCongr e).symm_apply_eq
 
+theorem trans_eq_refl_iff_eq_symm (f : α ≃ β) (g : β ≃ α) :
```
eric-wieser, line 497: We have four variants of this theorem for `mul_eq_one`, so probably we should have the same thing for this lemma.

Can you also make the `f` and `g` arguments implicit to match?

test/RewriteSearch/Polynomial.lean
urkud: It looks like the test is non-deterministic: it can decide to `rw [sub_eq_neg_add]` instead of `rw [sub_eq_add_neg]`, then fail.

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -494,6 +494,23 @@ theorem eq_symm_comp {α β γ} (e : α ≃ β) (f : γ → α) (g : γ → β)
...
+  rw [← Equiv.coe_inj, coe_trans, coe_refl, ← eq_symm_comp, comp_id, Equiv.coe_inj]
+
+theorem trans_eq_refl_iff_symm_eq {f : α ≃ β} {g : β ≃ α} :
+    f.trans g = Equiv.refl α ↔ f.symm = g := by
+  rw [trans_eq_refl_iff_eq_symm]
+  exact ⟨fun h ↦ h ▸ rfl, fun h ↦ h ▸ rfl⟩
+
+theorem eq_symm_trans_eq_refl {f : α ≃ β} {g : β ≃ α} :
+    f = g.symm ↔ f.trans g = Equiv.refl α :=
+  trans_eq_refl_iff_eq_symm.symm
```
urkud, line 508: I suggest that we don't add these 2 versions. One can always use `rw [← _]` or `_.symm` with the previous theorems.

Mathlib/Algebra/Group/Basic.lean
```diff
@@ -644,6 +644,14 @@ theorem mul_eq_one_iff_eq_inv : a * b = 1 ↔ a = b⁻¹ :=
 theorem mul_eq_one_iff_inv_eq : a * b = 1 ↔ a⁻¹ = b := by
   rw [mul_eq_one_iff_eq_inv, inv_eq_iff_eq_inv]
 
+@[to_additive]
+theorem mul_eq_one_iff_eq_inv' : a * b = 1 ↔ b = a⁻¹ := by
+  rw [mul_eq_one_iff_inv_eq, eq_comm]
```
urkud, line 649: How do you plan to use these lemmas? Can you just use `rw [mul_eq_one_iff_inv_eq, eq_comm]` instead?

Rida-Hamadani: This lemma and the one you ask about in the previous comment were both added due to suggestions by Eric, the context is [here](https://github.com/leanprover-community/mathlib4/pull/16032#discussion_r1725944729) and [here](https://github.com/leanprover-community/mathlib4/pull/16032#discussion_r1737282636).

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -494,6 +494,23 @@ theorem eq_symm_comp {α β γ} (e : α ≃ β) (f : γ → α) (g : γ → β)
...
+theorem trans_eq_refl_iff_eq_symm {f : α ≃ β} {g : β ≃ α} :
+    f.trans g = Equiv.refl α ↔ f = g.symm := by
+  rw [← Equiv.coe_inj, coe_trans, coe_refl, ← eq_symm_comp, comp_id, Equiv.coe_inj]
+
+theorem trans_eq_refl_iff_symm_eq {f : α ≃ β} {g : β ≃ α} :
+    f.trans g = Equiv.refl α ↔ f.symm = g := by
+  rw [trans_eq_refl_iff_eq_symm]
+  exact ⟨fun h ↦ h ▸ rfl, fun h ↦ h ▸ rfl⟩
+
+theorem eq_symm_trans_eq_refl {f : α ≃ β} {g : β ≃ α} :
```
urkud, line 506:
```suggestion
theorem eq_symm_iff_trans_eq_refl {f : α ≃ β} {g : β ≃ α} :
```

Mathlib/Logic/Equiv/Defs.lean
```diff
@@ -494,6 +494,23 @@ theorem eq_symm_comp {α β γ} (e : α ≃ β) (f : γ → α) (g : γ → β)
...
+    f = g.symm ↔ f.trans g = Equiv.refl α :=
+  trans_eq_refl_iff_eq_symm.symm
+
+theorem symm_eq_trans_eq_refl_iff{f : α ≃ β} {g : β ≃ α} :
```
urkud, line 510:
```suggestion
theorem symm_eq_iff_trans_eq_refl {f : α ≃ β} {g : β ≃ α} :
```

Mathlib/Algebra/Group/Basic.lean
urkud: OK, let's leave them.

urkud:
Thanks! Please fix names (see suggestions above), wait for the CI, then merge.
bors d+

PR #16039, opened 2024-08-21
[Merged by Bors] - feat: Register more @[bound] lemmas for floor, ceil, and log

girving:
`Int.ceil_lt_add_one` and `Real.log_pos` came up in https://github.com/girving/ray, and I've registered a variety of nearby lemmas.

---

### DIFF
--- a/Mathlib/Algebra/Order/Floor.lean
+++ b/Mathlib/Algebra/Order/Floor.lean
@@ -151,6 +151,7 @@ theorem floor_le (ha : 0 ≤ a) : (⌊a⌋₊ : α) ≤ a :=
 theorem lt_succ_floor (a : α) : a < ⌊a⌋₊.succ :=
   lt_of_floor_lt <| Nat.lt_succ_self _
 
+@[bound]
 theorem lt_floor_add_one (a : α) : a < ⌊a⌋₊ + 1 := by simpa using lt_succ_floor a
 
 @[simp]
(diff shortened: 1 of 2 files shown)


j-loreaux:
@girving I'm tempted to trust you, and just merge this, but I'm waiting because I have a larger question, which I believe needs answering:

What constitutes a bad `@[bound]` lemma?

My concern is that without this kind of guidance, we'll end up tagging lemmas with `@[bound]` that cause extremely serious performance degradation‌. Moreover, it may not show up immediately with `!bench`. Is there somewhere this is addressed that I have missed?

girving:
The main performance constraint is that expressions on either side of inequality hypotheses should be strict subexpressions of the conclusion (unless they are very simple expressions like 0 or 1), to (1) avoid loops and (2) adhere to the general goal of `bound`. However, I am not sure what the precise formalization of this should be, or whether there is a precise formalization that definitely avoids loops.

Note that this requirement holds vacuously if there are no hypotheses, which is true of a lot of the non-gcongr-like lemmas tagged here. `Real.log_pos` shows why you need "unless they are very simple", as it turns `0 < log x` into `1 < x` where 1 doesn't occur in the goal.

girving:
If we can find a satisfactory version of this which is programmatic I could make a test enforcing it, or enforce it at attribute time.

joneugster:
!bench

joneugster:
if you are able to write some tests to prevent some bad uses of the attribute, I encourage you to do so; that would be well-received! But that seems like it's material for a follow up.

maintainer merge

(@j-loreaux)

j-loreaux:
bors merge

girving:
@joneugster Thank you! I will eventually write the subexpression test, but there may be a delay involved.

PR #16041, opened 2024-08-21
feat(SetTheory/Ordinal/Arithmetic): `Ordinal.toNat`

vihdzp:
A simple function sending ordinals less than `ω` to themselves, and anything else to `0`.

---

- [ ] depends on: #17758


Mathlib/SetTheory/Ordinal/Arithmetic.lean
```diff
@@ -2482,6 +2482,64 @@ theorem sup_mul_nat (o : Ordinal) : (sup fun n : ℕ => o * n) = o * ω := by
     exact sup_eq_zero_iff.2 fun n => zero_mul (n : Ordinal)
   · exact (mul_isNormal ho).apply_omega0
 
+/-- Converts an ordinal less than `ω` into a natural number, sending all infinite ordinals to zero.
+-/
+def toNat (o : Ordinal) : ℕ :=
+  if h : o < ω then Classical.choose (lt_omega0.1 h) else 0
```
urkud, line 2488: The projection `toENat` may have better properties, see https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Cardinal/ENat.html#Cardinal.toENat for `Cardinal` version. Also, what do you think about linking your functions to `Ordinal.card`, or even defining it as `o.card.toENat`?

vihdzp: I like this idea. I'll put these definitions in files `Ordinal.ENat` and `Ordinal.ToNat` the cardinal files.

PR #16053, opened 2024-08-22
chore(SetTheory/Ordinal/Notation): improve documentation

vihdzp:
We improve comment formatting and perform some simple golfing.

---

There's a lot with this file that can be improved. Auxiliary theorems are inconsistently named and poorly documented, the use of `NF` as a typeclass is contentious (at least to myself, see [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/When.20to.20use.20instances.20for.20terms)), the use of `OfNat` instead of the `NatCast` typeclass is somewhat idiosyncratic, and a lot of the proofs can definitely be simplified. But I think it's better to go about this in stages.


Ruben-VandeVelde:
There's still a fair amount of work going on here, and not just documentation. Could you split further or update the summary?

vihdzp:
Yeah, I'll split this PR sometime in the following days. There's a lot of stuff I want to do with this file and I might have gone overboard for an initial PR.

vihdzp:
I've moved the documentation related stuff to #16283

vihdzp:
Closing, the endgame here is probably to just outright replace this file by my [rewritten version](https://github.com/vihdzp/ordinal-notation/blob/master/OrdinalNotation/Cantor.lean).

PR #16054, opened 2024-08-22
chore(SetTheory/Game/Impartial): remove `Impartial` typeclass

vihdzp:
... and use a simple proposition instead. Dot notation gives us back our lost convenience gain.

---

- [x] depends on: #16517 [doing this first makes the diff much smaller]
- [x] depends on: #16349

See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/When.20to.20use.20instances.20for.20terms) for a discussion on this topic.


vihdzp:
Closing. No longer convinced this is a good idea.

PR #16057, opened 2024-08-22
[Merged by Bors] - chore: deprecate unused MLList material, which has moved to a separate repo

kim-em:
This material is unused in Mathlib, and now exists in a separate repo at https://github.com/semorrison/lean-monadic-list


semorrison:
I've simply deleted the test files, as these are already reproduced at the external repo, and deprecated everything else in case it is being used downstream.

Parcly-Taxel (approved):

jcommelin (commented):
Thanks :tada:

bors merge

PR #16061, opened 2024-08-22
[Merged by Bors] - feat(Data/Finsupp/Defs): `induction_on_max`

vihdzp:
A finitely supported function can be built in ascending or descending order of "degree".

---


Mathlib/Data/Finsupp/Defs.lean
```diff
@@ -1034,6 +1034,48 @@ theorem induction_linear {p : (α →₀ M) → Prop} (f : α →₀ M) (h0 : p
     (hadd : ∀ f g : α →₀ M, p f → p g → p (f + g)) (hsingle : ∀ a b, p (single a b)) : p f :=
   induction₂ f h0 fun _a _b _f _ _ w => hadd _ _ w (hsingle _ _)
 
+section LinearOrder
+
+variable [LinearOrder α] {p : (α →₀ M) → Prop}
+
+theorem induction_on_max (f : α →₀ M) (h0 : p 0)
```
eric-wieser, line 1041: Can you add docstrings to the new induction principles?

YaelDillies (approved):
Looks fine by me

maintainer merge

Mathlib/Data/Finsupp/Defs.lean
```diff
@@ -1034,6 +1034,52 @@ theorem induction_linear {p : (α →₀ M) → Prop} (f : α →₀ M) (h0 : p
     (hadd : ∀ f g : α →₀ M, p f → p g → p (f + g)) (hsingle : ∀ a b, p (single a b)) : p f :=
   induction₂ f h0 fun _a _b _f _ _ w => hadd _ _ w (hsingle _ _)
 
+section LinearOrder
+
+variable [LinearOrder α] {p : (α →₀ M) → Prop}
+
+/-- A finitely supported function can be built by adding up `single a b` for increasing `a`. -/
+theorem induction_on_max (f : α →₀ M) (h0 : p 0)
+    (ha : ∀ (a b) (f : α →₀ M), (∀ c ∈ f.support, c < a) → b ≠ 0 → p f → p (single a b + f)) :
```
fpvandoorn, line 1043: Would it make sense to have a version for partial orders with `∀ c ∈ f.support, c < a` replaced by `∀ c ∈ f.support, ¬ a ≤ c`? I think that could also be useful.

fpvandoorn: Another thing is that it would be nice to eliminate to `Sort*`
But maybe both of these fall outside the scope of this PR?

Mathlib/Data/Finsupp/Defs.lean
```diff
@@ -1034,6 +1034,52 @@ theorem induction_linear {p : (α →₀ M) → Prop} (f : α →₀ M) (h0 : p
...
+
+/-- A finitely supported function can be built by adding up `single a b` for decreasing `a`. -/
+theorem induction_on_min (f : α →₀ M) (h0 : p 0)
+    (ha : ∀ (a b) (f : α →₀ M), (∀ c ∈ f.support, a < c) → b ≠ 0 → p f → p (single a b + f)) :
+    p f :=
+  induction_on_max (α := αᵒᵈ) f h0 ha
+
+/-- A finitely supported function can be built by adding up `single a b` for increasing `a`. -/
+theorem induction_on_max₂ (f : α →₀ M) (h0 : p 0)
+    (ha : ∀ (a b) (f : α →₀ M), (∀ c ∈ f.support, c < a) → b ≠ 0 → p f → p (f + single a b)) :
```
fpvandoorn, line 1063: Is the only difference the order of argument of `+`? Can you mention that in the doc-string? 
I don't like the subscript `₂` to distinguish this version. Maybe `_right` instead of `2`?
Isn't the proof just 1 line from `induction_on_max`? (Though if this is generalized to `Sort* then the definition might matter.)

Mathlib/Data/Finsupp/Defs.lean
vihdzp: This proof relies mostly on `Finset.induction_on_max`, so these changes would have to be implemented there first.

Note that we need a linear order for the `Sort*` version to be useful, as otherwise whichever of two incomparable elements we build first would be entirely arbitrary.

Mathlib/Data/Finsupp/Defs.lean
vihdzp: This naming scheme matches that of [`Finsupp.induction₂`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Finsupp/Defs.html#Finsupp.induction%E2%82%82)

fpvandoorn: Oh, you're right, my plan doesn't work without commutativity. I guess you can try to apply it in `AddOpposite M`, but the current situation is fine as well. Can you please mention the difference in the docstring?

Mathlib/Data/Finsupp/Defs.lean
fpvandoorn: Ok fair. Let's keep it like this.

fpvandoorn:
bors d+

PR #16064, opened 2024-08-22
[Merged by Bors] - feat(Category/Cat): add API lemmas

callesonne:
This PR adds more API lemmas about the bicategory of categories `Cat`. These are added as `simp` is not able to use the corresponding `NatTrans` lemmas when working with 2-morphisms in the bicategory `Cat`.

Due to this some proofs in `CategoryTheory/Grothendieck` can be golfed.

---


joelriou:
Thanks!

bors merge

PR #16065, opened 2024-08-22
[Merged by Bors] - feat: stability of `List.insertionSort`

marcusrossel:
The statements of stability follow those used for `mergeSort` in leanprover/lean4#5092.

---

This PR is a result of [this discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Insertion.20Sort/near/464335927), where @semorrison recommended we verify that `List.insertionSort` is stable.


Mathlib/Data/List/Sort.lean
```diff
@@ -356,6 +356,52 @@ theorem sorted_insertionSort : ∀ l, Sorted r (insertionSort r l)
 
 end TotalAndTransitive
 
+theorem orderedInsert_stable {l c : List α} (a : α) (hl : c <+ l) : c <+ orderedInsert r a l := by
+  induction l generalizing c
+  case nil => simp_all only [sublist_nil, orderedInsert, sublist_cons_self]
```
mattrobball, line 361: Can you try to minimize the use of `case`? Thanks.

marcusrossel: Sure :) Out of interest, what is the reasoning behind this?

mattrobball: It is just stylistic for the library. If you have to use `case`, then that is an indication that the tactic's syntax needs to be extended.

semorrison:
Otherwise LGTM!

marcusrossel:
I'll be able to get back to this on Monday.

Mathlib/Data/List/Sort.lean
```diff
@@ -356,6 +356,52 @@ theorem sorted_insertionSort : ∀ l, Sorted r (insertionSort r l)
 
 end TotalAndTransitive
 
+theorem orderedInsert_stable {l c : List α} (a : α) (hl : c <+ l) : c <+ orderedInsert r a l := by
```
eric-wieser, line 359: Here's a shorter proof
```suggestion
theorem orderedInsert_stable {l c : List α} (a : α) (hl : c <+ l) : c <+ orderedInsert r a l  :=
  hl.trans <| sublist_orderedInsert a l
```
which maybe suggests you don't need this lemma at all!

Mathlib/Data/List/Sort.lean
```diff
@@ -356,6 +368,27 @@ theorem sorted_insertionSort : ∀ l, Sorted r (insertionSort r l)
 
 end TotalAndTransitive
 
+/--
+If `c` is a sorted sublist of `l`, then `c` is still a sublist of `insertionSort r l`.
+-/
+theorem insertionSort_stable {l c : List α} (hr : c.Pairwise r) (hc : c <+ l) :
```
eric-wieser, line 374: This only needs the weaker `c <+~ l`, right?

marcusrossel: Presumably. The reason I chose `<+`, is because it's what is used for the statement of stability in https://github.com/leanprover/lean4/pull/5092.

marcusrossel: Probably, yes. I chose `<+` because that's what's used to state stability in https://github.com/leanprover/lean4/pull/5092. Would it be desirable to generalize to `<+~`?

marcusrossel: I have a proof of the more general version, using the following assumed lemmas:

```lean
theorem THM₁ [DecidableEq α] {l : List α} {p} (a : α) (h : l.Pairwise p) :
    (l.erase a).Pairwise p := by
  sorry

theorem THM₂ [DecidableEq α] {a : α} {as bs} (h : a :: as ~ bs) : as ~ bs.erase a := by
  sorry

theorem THM₃ {as bs} (a : α) (ha : as.Sorted r) (hs : as <+ bs) :
    orderedInsert r a as <+ orderedInsert r a bs := by
  sorry
```

eric-wieser: > If it's necessary, then perhaps it would be good to also keep the <+ version of the theorem.

Agreed.

marcusrossel:
I've added the stability theorem based on `<+~`, as well as the additional lemma `orderedInsert_sublist`.

marcusrossel:
Also, https://github.com/leanprover/lean4-nightly/commit/6d0b00885ed79c0910080613f33e1189bb496ce8 adds `Pairwise.erase`, which I would use on Line 425. Should I add a TODO to change this once Mathlib uses the next release, or is this not important enough?

Mathlib/Data/List/Sort.lean
```diff
@@ -327,6 +327,37 @@ theorem sublist_orderedInsert (x : α) (xs : List α) : xs <+ xs.orderedInsert r
...
+theorem cons_sublist_orderedInsert {l c : List α} {a : α} (hl : c <+ l) (ha : ∀ a' ∈ c, a ≼ a') :
+    a :: c <+ orderedInsert r a l := by
+  induction l with
+  | nil         => simp_all only [sublist_nil, orderedInsert, Sublist.refl]
+  | cons _ _ ih =>
+    unfold orderedInsert
+    split
+    · exact .cons₂ _ hl
+    · cases hl
+      · exact .cons _ <| ih ‹_›
```
Ruben-VandeVelde, line 339: You're using `‹_›` a lot, which isn't great for readability. I'd suggest switching to `cases ... with` and `split_ifs`, if that helps, and otherwise at least fill in the underscores.

Ruben-VandeVelde (commented):
And please add the TODO, yes

Mathlib/Data/List/Sort.lean
marcusrossel: Fixed

marcusrossel:
I've changed the theorem names as to align with https://github.com/leanprover/lean4/pull/5242

Mathlib/Data/List/Sort.lean
```diff
@@ -345,6 +345,40 @@ theorem sublist_orderedInsert (x : α) (xs : List α) : xs <+ xs.orderedInsert r
...
+  | nil         => simp_all only [sublist_nil, orderedInsert, Sublist.refl]
+  | cons _ _ ih =>
+    unfold orderedInsert
+    split_ifs with hr
+    · exact .cons₂ _ hl
+    · cases hl with
+      | cons _ h => exact .cons _ <| ih h
+      | cons₂    => exact absurd (ha _ <| mem_cons_self ..) hr
+
+theorem sublist_orderedInsert_sublist [IsTrans α r] {as bs} (x) (hs : as <+ bs) (hb : bs.Sorted r) :
```
jcommelin, line 360: I think this should be called
```suggestion
theorem Sublist.orderedInsert_sublist [IsTrans α r] {as bs} (x) (hs : as <+ bs) (hb : bs.Sorted r) :
```
or
```suggestion
theorem sublist_orderedInsert_of_sublist [IsTrans α r] {as bs} (x) (hs : as <+ bs) (hb : bs.Sorted r) :
```

kim-em:
bors merge

PR #16066, opened 2024-08-22
[Merged by Bors] - feat(CategoryTheory/Adjunction): dualize adjoint lifting theorem

dagurtomas:
This PR renames the file `CategoryTheory.Adjunction.Lifting` to `CategoryTheory.Adjunction.Lifting.Left` and dualizes it in a new file `CategoryTheory.Adjunction.Lifting.Right` to lift right adjoints through comonads.

It also changes some names in the original file which didn't follow naming conventions, as well as renaming the implementation namespace.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #15894, opened 2024-08-16
[Merged by Bors] - feat(CategoryTheory/Monoidal): Day's reflection theorem

dagurtomas:
---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16067, opened 2024-08-22
[Merged by Bors] - feat(CategoryTheory/Closed): functors into a closed monoidal category form a closed monoidal category

dagurtomas:
...assuming the target category has certain limits

---

- [x] depends on: #16066 


Mathlib/CategoryTheory/Closed/FunctorCategory/Complete.lean
```diff
@@ -0,0 +1,75 @@
...
+import Mathlib.CategoryTheory.Closed.FunctorCategory.Groupoid
+import Mathlib.CategoryTheory.Groupoid.Discrete
+import Mathlib.CategoryTheory.Limits.Preserves.FunctorCategory
+import Mathlib.CategoryTheory.Monad.Comonadicity
+/-!
+
+# Functors into a complete monoidal closed category form a monoidal closed category.
+
+TODO (in progress by Joël Riou): make a more explicit construction of the internal hom in functor
+categories.
```
joelriou, line 16: Indeed, in order to obtain the monoidal structure on categories of sheaves with values in a closed symmetric monoidal category (with suitable limits), using #12728, we will need to know that if `Y` is any presheaf, and `f : X₁ ⟶ X₂` is a morphism of presheaves which becomes an isomorphism after sheafification, then it is also the case of `f ▷ Y`. This is equivalent to saying that if `Z` is any sheaf, the induced map `(X₂ ⊗ Y ⟶ Z) → (X₁ ⊗ Y ⟶ Z)` is a bijection, which is the same as showing there is a bijection `(X₂ ⟶ Y ⟹ Z) → (X₁ ⟶ Y ⟹ Z)` (if I use the notation `⟹` for the internal hom of presheaves), and we may conclude if we know that `Y ⟹ Z` is a sheaf when `Z` is a sheaf (see `Sites.SheafHom` for a similar result). Finally, using your result in PR #15894, we may also get that the monoidal category structure on sheaves is also closed.

In order to get the tensor product of sheaves (with its monoidal compatibilites), we not only need to know that the monoidal structure on presheaves is closed, but we also need a certain understanding of what the sections of `Y ⟹ Z` look like. Then, I am not sure what to do with this PR, because on the one hand, your proof is really great, but we may not extract enough information out of it for our future applications. Another approach would be to develop the more explicit construction in a way that would allow both a construction of a `MonoidalClosed` instance, and working with a preexistent `MonoidalClosed` instance, e.g. the one given by this PR? In which case, the API would provide isomorphisms with the given `ihom`.

joelriou: I have done the first step in https://github.com/leanprover-community/mathlib4/pull/18009 which together with #17326 gives an enrichment of a functor category `J ⥤ C` in `C`. In a certain sense, this is the "global sections" of the expected internal hom. Then, in order to construct the actual internal hom, I will study the functoriality of this enrichement of `J ⥤ C` with respect to the category `J`. Then, all the enrichements of `Under j ⥤ C` for `j : J` over `C` should give the expected enrichment of `J ⥤ C` in itself, and hopefully it will be possible to deduce that `J ⥤ C` is closed monoidal.

joelriou: I feel the best solution is to merge this one.

joelriou:
Thanks!

bors merge

PR #16070, opened 2024-08-22
[Merged by Bors] - feature(Analysis/Normed/Module/Dual): Polar of the open ball

mans0954:
The polar of an open ball at the origin is also the closed dual ball at the origin.

I think one would normally use the bipolar theorem to prove this, which I don't think we have yet.

---


urkud:
Thanks! :tada:
bors merge

PR #16071, opened 2024-08-22
[Merged by Bors] - chore(CategoryTheory/Adjunction): clean up file `Adjunction.Basic`

dagurtomas:
Removes a bunch of porting notes and simplifies some proofs.

Also removes the `simp` attributes from the `homEquiv_naturality` lemmas. These were bad `simp` lemmas because they can interfere with the `simp` lemmas `homEquiv_unit` and `homEquiv_counit`. Removing them doesn't break anything throughout all of mathlib.

---

### DIFF
--- a/Mathlib/CategoryTheory/Adjunction/Basic.lean
+++ b/Mathlib/CategoryTheory/Adjunction/Basic.lean
@@ -251,11 +251,6 @@ structure CoreHomEquiv (F : C ⥤ D) (G : D ⥤ C) where
 
 namespace CoreHomEquiv
 
--- Porting note: Workaround not needed in Lean 4.
--- restate_axiom homEquiv_naturality_left_symm'
-
--- restate_axiom homEquiv_naturality_right'
-
 attribute [simp] homEquiv_naturality_left_symm homEquiv_naturality_right
 
 variable {F : C ⥤ D} {G : D ⥤ C} (adj : CoreHomEquiv F G) {X' X : C} {Y Y' : D}
--- a/Mathlib/CategoryTheory/Adjunction/Basic.lean
+++ b/Mathlib/CategoryTheory/Adjunction/Basic.lean
@@ -174,13 +174,13 @@ theorem homEquiv_naturality_right_square_iff (f : X' ⟶ X) (g : X ⟶ G.obj Y')
 @[simp]
 theorem left_triangle : whiskerRight adj.unit F ≫ whiskerLeft F adj.counit = 𝟙 _ := by
   ext; dsimp
-  erw [← adj.homEquiv_counit, Equiv.symm_apply_eq, adj.homEquiv_unit]
+  rw [← adj.homEquiv_counit, Equiv.symm_apply_eq, adj.homEquiv_unit]
   simp
 
 @[simp]
 theorem right_triangle : whiskerLeft G adj.unit ≫ whiskerRight adj.counit G = 𝟙 _ := by
   ext; dsimp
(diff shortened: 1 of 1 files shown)


Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -63,11 +63,9 @@ structure Adjunction (F : C ⥤ D) (G : D ⥤ C) where
   unit : 𝟭 C ⟶ F.comp G
   /-- The counit of an adjunction -/
   counit : G.comp F ⟶ 𝟭 D
-  -- Porting note: It's strange that this `Prop` is being flagged by the `docBlame` linter
```
dagurtomas, line 66: It's not anymore, but having a docstring doesn't hurt. Even better if the docstring is correct :-)

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -105,11 +103,6 @@ noncomputable def Adjunction.ofIsRightAdjoint (right : C ⥤ D) [right.IsRightAd
 
 namespace Adjunction
 
--- Porting note: Workaround not needed in Lean 4
--- restate_axiom homEquiv_unit'
-
--- restate_axiom homEquiv_counit'
```
dagurtomas, line 111: I guess this means the porting note can be safely removed.

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -132,35 +125,23 @@ theorem homEquiv_id (X : C) : adj.homEquiv X _ (𝟙 _) = adj.unit.app X := by s
 
 theorem homEquiv_symm_id (X : D) : (adj.homEquiv _ X).symm (𝟙 _) = adj.counit.app X := by simp
 
-/-
-Porting note: `nolint simpNF` as the linter was complaining that this was provable using `simp`
-but it is in fact not. Also the `docBlame` linter expects a docstring even though this is `Prop`
-valued
--/
```
dagurtomas, line 139: It can in fact be proved by `simp`, and in any case it's probably a bad simp lemma because it interferes with `homEquiv_counit`.

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -381,11 +341,6 @@ def mkOfUnitCounit (adj : CoreUnitCounit F G) : F ⊣ G :=
           simp only [id_comp] at t
           exact t } }
 
-/- Porting note: simpNF linter claims these are solved by simp but that
-is not true -/
```
dagurtomas, line 385: The linter doesn't complain anymore.

Mathlib/CategoryTheory/Adjunction/Basic.lean
```diff
@@ -63,11 +63,9 @@ structure Adjunction (F : C ⥤ D) (G : D ⥤ C) where
   unit : 𝟭 C ⟶ F.comp G
   /-- The counit of an adjunction -/
   counit : G.comp F ⟶ 𝟭 D
-  -- Porting note: It's strange that this `Prop` is being flagged by the `docBlame` linter
-  /-- Naturality of the unit of an adjunction -/
+  /-- The relationship between the unit and hom set equivalence of an adjunction -/
```
dagurtomas, line 66: It's not anymore, but having a docstring doesn't hurt. Even better if the docstring is correct :-)

jcommelin (commented):
Thanks :tada: -- very nice! Getting rid of some `erw`s as well!

bors merge

PR #16072, opened 2024-08-22
[Merged by Bors] - fix: error positioning in `linear_combination`

hrmacbeth:
A tweak to `linear_combination` so that when the default normalization tactic fails, the "red underlines" appear on the `linear_combination` call itself, rather than just on the "by" at the start of the proof.  For example, compare before/after on this example:
```lean
import Mathlib

example {a : ℤ} (h : a = 1) : a = 2 := by
  linear_combination h
```

Zulip discussion:
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Propagating.20errors.20through.20Unhygienic.2Erun

---

### DIFF
--- a/Mathlib/Tactic/LinearCombination.lean
+++ b/Mathlib/Tactic/LinearCombination.lean
@@ -126,7 +126,7 @@ theorem eq_of_add_pow [Ring α] [NoZeroDivisors α] (n : ℕ) (p : (a:α) = b)
   rw [← sub_eq_zero] at p ⊢; apply pow_eq_zero (n := n); rwa [sub_eq_zero, p] at H
 
 /-- Implementation of `linear_combination` and `linear_combination2`. -/
-def elabLinearCombination
+def elabLinearCombination (tk : Syntax)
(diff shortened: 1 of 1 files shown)


Mathlib/Tactic/LinearCombination.lean
```diff
@@ -137,7 +137,7 @@ def elabLinearCombination
     match ← expandLinearCombo ty e with
     | .const c => `(Eq.refl $c)
     | .proof p => pure p
-  let norm := norm?.getD (Unhygienic.run `(tactic| ring1))
+  let norm := norm?.getD (Unhygienic.run <| withRef (← getRef) `(tactic| ring1))
```
eric-wieser, line 140: This puts the error message on the whole tactic, not just the `linear_combination`, right?

If so, presumably Kyle's suggestion [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Propagating.20errors.20through.20Unhygienic.2Erun/near/464484715) would resolve it.

eric-wieser (approved):
bors d+

We can cleanup the `Unhygienic.run` in a later PR after hearing from Mario on Zulip.

PR #16074, opened 2024-08-22
feat: combinatorial maps and planar graphs

Rida-Hamadani:
We define combinatorial maps, then we define planar graph using combinatorial maps.

---


Parcly-Taxel:
1. In Gonthier's 4CT Coq proof combinatorial maps are called hypermaps and their three permutations are called `edge, node, face` in that order. I believe this is a more appropriate name as _none_ of the three permutations has to be an involution.
2. Don't use `.trans`, `.comp` and the like; use `*!

Rida-Hamadani:
@Parcly-Taxel Thank you for you comments.
1.  hypermaps are a generalization of combinatorial maps. I can define hypermaps and then let combinatorial maps extend them. What do you think?
2. I agree, I will do the change.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+  Fintype.ofFinite M.vertices
+
+noncomputable instance [Fintype D] : Fintype M.edges :=
+  Fintype.ofFinite M.edges
+
+noncomputable instance [Fintype D] : Fintype M.faces :=
+  Fintype.ofFinite M.faces
+
+/-- The Euler characteristic of a `CombinatorialMap`. -/
+noncomputable def EulerCharacteristic [Fintype D] : ℤ :=
```
lambda-fairy, line 180: eulerCharacteristic

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+  around an edge.
+  -/
+  σ : Equiv.Perm D
+  /-- Permutation `α` whose orbits correspond to edges. `α` gives the opposite dart of an edge. -/
+  α : Equiv.Perm D
+  /--
+  Permutation `φ` whose orbits correspond to faces. `φ` gives the next dart counter-clockwise of the
+  same face.
+  -/
+  φ : Equiv.Perm D
```
lambda-fairy, line 36: Can we call these fields `vertex`, `edge`, and `face`? Or some similar word, like `vertices` or `vertexPerm` etc.

Usually we use single Greek letters for type variables.

lambda-fairy, line 36: IIRC, in the Coq version, the three perms are declared as normal functions, and the fact that they are permutations is derived from `composition`. Can we do that here?

Alternatively, since any one perm is fixed by the other two, we can _define_ one of them that way, and then the `composition` property is redundant.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+  composition : φ * σ * α = 1
+  involutive : Function.Involutive α
```
lambda-fairy, line 38: This should be named `edgePerm_involutive` or similar, because it's not the whole map that is involutive, just the edge permutation.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+  fixedPoints_isEmpty : IsEmpty (Function.fixedPoints α)
```
lambda-fairy, line 39: isEmpty_fixedPoints_vertexPerm

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+  rw [← M.involutive.symm_eq_self_of_involutive, h]
+  rfl
+
+lemma composition_apply {x : D} : M.φ (M.σ (M.α x)) = x := by
+  nth_rw 2 [← Equiv.Perm.one_apply x]
+  rw [← M.composition]
+  rfl
+
+/-- This order of permutations also results in the identity map. -/
+lemma composition' : M.α * M.φ * M.σ = 1 :=
```
lambda-fairy, line 59: Instead of defining `composition`, `composition'`, etc., consider using a more literal name like `vertex_mul_edge_mul_face`.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+
+lemma composition'_apply {x : D} : M.α (M.φ (M.σ x)) = x := by
+  convert_to (M.α * M.φ * M.σ) x = x
+  simp [M.composition']
+
+/--
+A homomorphism of `CombinatorialMap`s is a function that commutes with each of the respective
+permutations.
+-/
+def Hom (M : CombinatorialMap D) (M' : CombinatorialMap D') (f : D → D') : Prop :=
```
lambda-fairy, line 73: This feels unidiomatic to me. Can we bundle the `f` into the structure, like we do for `SimpleGraph.Hom`?

We can define a separate `Iso` that is bijective.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+    simp [composition'_apply]
+  involutive := by
+    have : M.α⁻¹ = M.α := M.involutive.symm_eq_self_of_involutive
+    simp [this, M.involutive]
+  fixedPoints_isEmpty := by
+    have : M.α⁻¹ = M.α := M.involutive.symm_eq_self_of_involutive
+    simp [this, M.fixedPoints_isEmpty]
+
+/-- The double dual of a `CombinatorialMap` is equal to the original map. -/
+lemma double_dual_eq (M : CombinatorialMap D) : M.dual.dual = M := rfl
```
lambda-fairy, line 125: dual_dual

And I don't think this needs a doc comment.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+
+/-- The vertices of a `CombinatorialMap`. -/
+@[reducible] def vertices (M : CombinatorialMap D) :=
```
lambda-fairy, line 128: Names of types should be uppercase and singular.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+
+noncomputable instance [Fintype D] : Fintype M.edges :=
+  Fintype.ofFinite M.edges
+
+noncomputable instance [Fintype D] : Fintype M.faces :=
+  Fintype.ofFinite M.faces
+
+/-- The Euler characteristic of a `CombinatorialMap`. -/
+noncomputable def EulerCharacteristic [Fintype D] : ℤ :=
+  (Fintype.card M.vertices) - (Fintype.card M.edges) + (Fintype.card M.faces)
```
lambda-fairy, line 181: Function application binds tightly, so we don't need the parentheses here.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+/-- The edges of a `CombinatorialMap`. -/
+@[reducible] def edges (M : CombinatorialMap D) :=
+  Quotient (Equiv.Perm.SameCycle.setoid M.α)
+
+/-- The faces of a `CombinatorialMap`. -/
+@[reducible] def faces (M : CombinatorialMap D) :=
+  Quotient (Equiv.Perm.SameCycle.setoid M.φ)
+
+/-- The vertex corresponding to a dart `d`. -/
+@[reducible] def dartVertex (M : CombinatorialMap D) (d : D) : M.vertices :=
```
lambda-fairy, line 140: I think it makes more sense to call this `Vertex.mk`. Similarly for the edges and faces.

Mathlib/Combinatorics/CombinatorialMap.lean
```diff
@@ -0,0 +1,187 @@
...
+        · intro
+          ext
+          exact ⟨h.symm.trans, h.trans⟩
+        · simp_all
+      ext
+      exact ⟨h.symm.trans, h.trans⟩) v
+
+/-- Set of darts incident to a vertex. -/
+def vertices.darts (v : M.vertices) : Set D :=
+  {d : D | M.dartVertex d = v}
```
lambda-fairy, line 168: Use `Set.preimage`.

Or, since nothing uses this definition yet, we can delete it and have users write `M.dartVertex ⁻¹' {v}` themselves.

lambda-fairy (commented):
Good start, thanks for working on this!

I wonder if it's worth trying to prove something non-trivial with this API first? Right now I don't have an idea how usable this is, and having an example would help with that.

Mathlib/Combinatorics/CombinatorialMap.lean
Rida-Hamadani: Thank you for the valuable review and sorry for taking so long to address it.
Did you mean `isEmpty_fixedPoints_edgePerm` here or is it a part of the definition that the vertex permutation has no fixed points?

Mathlib/Combinatorics/CombinatorialMap.lean
Rida-Hamadani: Great idea, but how should we implement it? Permutations are not a class, I cannot say something such as `instance : Perm D vertexPerm` where `vertexPerm : D → D` right?
One idea is to define `CombinatorialMap` as following:
```lean
/--
A two-dimension combinatorial map is a collection `D` of darts, and maps `vertexMap`, `edgeMap`, and
`edgeMap`. `edgeMap` is a fixed-point-free involution.

Note: this definition excludes disjoint isolated vertices.
-/
structure CombinatorialMap (D : Type*) where
  /--
  Map whose orbits correspond to vertices, it gives the next dart counter-clockwise around an edge.
  -/
  vertexMap : D → D
  /-- Map whose orbits correspond to edges, it gives the opposite dart of an edge. -/
  edgeMap : D → D
  /--
  Map whose orbits correspond to faces, it gives the next dart counter-clockwise of the same face.
  -/
  faceMap : D → D
  edgeMap_vertexMap_faceMap_eq_id : edgeMap ∘ vertexMap ∘ faceMap = id
  edgeMap_involutive : edgeMap.Involutive
  isEmpty_fixedPoints_edgeMap : IsEmpty (Function.fixedPoints edgeMap)
  ```
  And then have separate permutations in the `CombinatorialMap` namespace as follows:
  ```
  def edgePerm : Equiv.Perm D where
  toFun := M.edgeMap
  invFun := M.edgeMap
  left_inv := M.edgeMap_involutive
  right_inv := M.edgeMap_involutive
  ```
  And so on for the rest of the maps. Is this solution satisfactory or do you have other ideas?

Rida-Hamadani: Another idea is to keep the permutations definition and have a `CombinatorialMap.mk` function that returns a combinatorial map given 3 maps and the required assumptions.

leanprover-community-bot-assistant:

Mathlib/Combinatorics/CombinatorialMap.lean
Rida-Hamadani: I have thought about this and I think the best way to do this is by letting the maps in `CombinatorialMap` be permutations but provide a way to construct a `CombinatorialMap` from maps. I've left this as a `TODO` item, I'll work on it later since I care about planar graphs for now, but let me know if you prefer me to do it now instead (or if you have a different method of implementing this).

Mathlib/Combinatorics/CombinatorialMap.lean
Rida-Hamadani: I've checked and I am almost sure you meant `_edgePerm` instead of `_vertexPerm`, so I'll resolve this.

leanprover-community-bot-assistant:

MrBrain295:
> Good start, thanks for working on this!
> 
> I wonder if it's worth trying to prove something non-trivial with this API first? Right now I don't have an idea how usable this is, and having an example would help with that.

I'm new to this so idk if this counts as making it usable but this pr is blocked by `isPlanar` being implemented google-deepmind/formal-conjectures#1413.

Rida-Hamadani:
@JMR-Independent That's great! Would you like to open a dependent PR with these additions?

joneugster:

Rida-Hamadani:
I will not work on this in the foreseeable future, so I will close it with hopes of someone else doing it.

PR #16075, opened 2024-08-22
feat(Data/Finsupp/Basic): `equivMapDomain f.symm (equivMapDomain f l) = l`

vihdzp:
---

Used within my work with the Cantor normal form.


eric-wieser (commented):
I understand you want this lemma, but I really don't think this scales. We have many functorial equivalences in mathlib, and for now the convention is to have only the `refl`, `symm`, and `trans` lemmas about them. I don't think we want to add `symm_comp` and `comp_symm` variants everywhere.

If we could generate these lemmas automatically with an attribute then I think my stance would change.

PR #16083, opened 2024-08-23
[Merged by Bors] - feat(SetTheory/Ordinal/NaturalOps): add missing instances

vihdzp:
---

### DIFF
--- a/Mathlib/SetTheory/Ordinal/NaturalOps.lean
+++ b/Mathlib/SetTheory/Ordinal/NaturalOps.lean
@@ -56,6 +56,8 @@ instance NatOrdinal.linearOrder : LinearOrder NatOrdinal := {Ordinal.linearOrder
 
 instance NatOrdinal.succOrder : SuccOrder NatOrdinal := {Ordinal.succOrder with}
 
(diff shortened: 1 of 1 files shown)


Mathlib/SetTheory/Ordinal/NaturalOps.lean
```diff
@@ -86,7 +87,14 @@ instance : WellFoundedLT NatOrdinal :=
   Ordinal.wellFoundedLT
 
 instance : IsWellOrder NatOrdinal (· < ·) :=
-  Ordinal.isWellOrder
+  { }
```
eric-wieser, line 90: What's going on here?

eric-wieser (approved):

Mathlib/SetTheory/Ordinal/NaturalOps.lean
vihdzp: An `IsWellOrder α (· < ·)` instance can always be inferred from `WellFoundedLT α` + `LinearOrder α`. Unfortunately, it doesn't seem like I can remove this instance altogether.

I can revert this golf if there's any issues with it.

urkud:
Thanks! :tada:
bors merge

PR #16085, opened 2024-08-23
feat: change map on measures to make sure that probability measures map to probability measures

sgouezel:
---

### DIFF
--- a/Mathlib/MeasureTheory/Measure/MeasureSpace.lean
+++ b/Mathlib/MeasureTheory/Measure/MeasureSpace.lean
@@ -1070,26 +1070,50 @@ theorem le_liftLinear_apply {f : OuterMeasure α →ₗ[ℝ≥0∞] OuterMeasure
     f μ.toOuterMeasure s ≤ liftLinear f hf μ s :=
   le_toMeasure_apply _ (hf μ) s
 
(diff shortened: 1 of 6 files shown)


Mathlib/MeasureTheory/Measure/MeasureSpace.lean
```diff
@@ -1070,26 +1070,50 @@ theorem le_liftLinear_apply {f : OuterMeasure α →ₗ[ℝ≥0∞] OuterMeasure
     f μ.toOuterMeasure s ≤ liftLinear f hf μ s :=
   le_toMeasure_apply _ (hf μ) s
 
+/-- When a function is not measurable, the target space is not empty. Therefore, we can choose
+a point there. Useful for junk values when some maps are not measurable. -/
+def default_of_not_measurable [MeasurableSpace α] (f : α → β) (hf : ¬(Measurable f)) : β := by
```
eric-wieser, line 1075:
```suggestion
def defaultOfNotMeasurable [MeasurableSpace α] (f : α → β) (hf : ¬(Measurable f)) : β := by
```
since this is data, and below.

sgouezel:
Sorry @eric-wieser, this is definitely not ready for review!

PR #16088, opened 2024-08-23
[Merged by Bors] - feat(SetTheory/Ordinal/Nimber): Initial development of nimbers

vihdzp:
For the moment, we only prove that they form an additive commutative group.

---

I've proven that they form a field in my branch [vi.nimber_mul](https://github.com/leanprover-community/mathlib4/tree/vi.nimber_mul).


Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+theorem add_def (a b : Nimber) :
+    a + b = sInf {x | (∃ a' < a, a' + b = x) ∨ ∃ b' < b, a + b' = x}ᶜ := by
+  change Nimber.add a b = _
+  rw [Nimber.add]
+  simp_rw [exists_prop]
+  rfl
+
+/-- The set in the definition of `Nimber.add` is nonempty. -/
+theorem add_nonempty (a b : Nimber) :
+    {x | (∃ a' < a, a' + b = x) ∨ ∃ b' < b, a + b' = x}ᶜ.Nonempty := by
```
Command-Master, line 208: This could also be proved by copying `small_Iio`, `small_Iic` and `not_small_ordinal` to nimbers and then stating that this set is too small,
```lean
theorem add_nonempty (a b : Nimber.{u}) :
    {x | (∃ a' < a, a' + b = x) ∨ ∃ b' < b, a + b' = x}ᶜ.Nonempty := by
  simp_rw [Set.nonempty_compl, Set.setOf_or, ← Set.mem_Iio, ← Set.image.eq_1]
  apply_fun (fun a : Set Nimber ↦ Small.{u} a)
  have : Small.{u} ↑((· + b) '' Set.Iio a ∪ (a + ·) '' Set.Iio b) := inferInstance
  simp [this, small_congr (Equiv.Set.univ _)]
  exact not_small_nimber
```
I'm not sure what's cleaner.

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+protected theorem add_comm (a b : Nimber) : a + b = b + a := by
+  apply le_antisymm <;>
+  apply add_le_of_forall_ne <;>
+  intro x hx
+  on_goal 1 => rw [Nimber.add_comm x, add_ne_add_right]
+  on_goal 2 => rw [Nimber.add_comm a, add_ne_add_left]
+  on_goal 3 => rw [← Nimber.add_comm a, add_ne_add_right]
+  on_goal 4 => rw [← Nimber.add_comm x, add_ne_add_left]
+  all_goals exact hx.ne
+termination_by (a, b)
```
Command-Master, line 261:
```suggestion
protected theorem add_comm (a b : Nimber) : a + b = b + a := by
  rw [add_def, add_def]
  simp_rw [or_comm]
  congr! 7 <;>
  (rw [and_congr_right_iff]; intro; rw [Nimber.add_comm])
termination_by (a, b)
```

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+
+-- TODO: add CharP 2 instance
+
+theorem add_cancel_right (a b : Nimber) : a + b + b = a := by
+  rw [add_assoc, add_self, add_zero]
+
+theorem add_cancel_left (a b : Nimber) : a + (a + b) = b := by
+  rw [← add_assoc, add_self, zero_add]
+
+theorem add_trichotomy {a b c : Nimber} (h : a ≠ b + c) :
```
Command-Master, line 347: Perhaps it's more symmetrical to take `a + b + c ≠ 0` instead of `a ≠ b + c`?

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+  neg_add_cancel := add_self
+  add_comm := Nimber.add_comm
+
+-- TODO: add CharP 2 instance
+
+theorem add_cancel_right (a b : Nimber) : a + b + b = a := by
+  rw [add_assoc, add_self, add_zero]
+
+theorem add_cancel_left (a b : Nimber) : a + (a + b) = b := by
+  rw [← add_assoc, add_self, zero_add]
```
Command-Master, line 345: Shouldn't `add_cancel_right` and `add_cancel_left` be `@[simp]`?

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+  apply add_le_of_forall_ne <;>
+  intro x hx
+  on_goal 1 => rw [Nimber.add_comm x, add_ne_add_right]
+  on_goal 2 => rw [Nimber.add_comm a, add_ne_add_left]
+  on_goal 3 => rw [← Nimber.add_comm a, add_ne_add_right]
+  on_goal 4 => rw [← Nimber.add_comm x, add_ne_add_left]
+  all_goals exact hx.ne
+termination_by (a, b)
+
+@[simp]
```
Command-Master, line 263: I don't think this should be a simp lemma. For comparison, `sub_eq_zero` isn't `@[simp]`.

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,361 @@
...
+theorem add_eq_zero_iff {a b : Nimber} : a + b = 0 ↔ a = b := by
```
Command-Master, line 264: Perhaps this should be called `add_eq_zero`, to match `sub_eq_zero`?

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: I agree, but this lemma is meant to match [`Nat.xor_trichotomy`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Bitwise.html#Nat.xor_trichotomy) which uses this same assumption.

What do you think about refactoring that theorem too?

Command-Master: That sounds like a good idea

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: `blsub₂` is some code stink that should eventually be removed, all it really does is codify the fact that `Iio o × Iio o` is a small type. So I'm glad to replace it by an argument using `Small`.

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: Sure thing.

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,359 @@
...
+protected theorem not_lt_zero (a : Nimber) : ¬ a < 0 :=
+  Ordinal.not_lt_zero a
+
+protected theorem pos_iff_ne_zero {a : Nimber} : 0 < a ↔ a ≠ 0 :=
+  Ordinal.pos_iff_ne_zero
+
+instance (a : Nimber.{u}) : Small.{u} (Set.Iio a) :=
+  Ordinal.small_Iio a
+
+theorem not_small_nimber : ¬ Small.{u} Nimber.{u} :=
```
Command-Master, line 143:
```suggestion
theorem not_small_nimber : ¬ Small.{u} Nimber.{max u v} :=
```
This is `Nimber.not_small_nimber` compared to `_root_.not_small_ordinal`, I think this should be outside the namespace.

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,359 @@
...
+  Ordinal.le_zero
+
+protected theorem not_lt_zero (a : Nimber) : ¬ a < 0 :=
+  Ordinal.not_lt_zero a
+
+protected theorem pos_iff_ne_zero {a : Nimber} : 0 < a ↔ a ≠ 0 :=
+  Ordinal.pos_iff_ne_zero
+
+instance (a : Nimber.{u}) : Small.{u} (Set.Iio a) :=
+  Ordinal.small_Iio a
```
Command-Master, line 141: Could you also transfer `Ordinal.small_Iic`?

Command-Master (approved):

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: In fact, I could make the statement even more symmetrical by changing it to
```
b + c < a ∨ c + a < b ∨ a + b < c
```

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: Sure. It'd be nice to be able to infer `Small (Set.Iic a)` (and the other intervals) from `Small (Set.Iio a)` and `NoMaxOrder`.

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,362 @@
...
+instance : AddCommGroupWithOne Nimber where
+  add_assoc := Nimber.add_assoc
+  add_zero := Nimber.add_zero
+  zero_add := Nimber.zero_add
+  nsmul := nsmulRec
+  zsmul := zsmulRec
+  neg_add_cancel := add_self
+  add_comm := Nimber.add_comm
+
+-- TODO: add CharP 2 instance
```
b-mehta, line 339: Let's move this TODO to the list of others at the top

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,362 @@
...
+
+To reduce API duplication, we opt not to implement operations on `Nimber` on `Ordinal`. The order
+isomorphisms `Ordinal.toNimber` and `Nimber.toOrdinal` allow us to cast between them whenever
+needed.
+
+## Todo
+
+- Define nim multiplication and prove nimbers are a commutative ring.
+- Define nim division and prove nimbers are a field.
+- Show the nimbers are algebraically closed.
```
b-mehta, line 36: It would also be nice to relate the bitwise `xor` definition on naturals to the addition you've defined here. Something like, doing `xor` on naturals then embedding into nimbers is the same as embedding into nimbers then adding

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,362 @@
...
+  rfl
+
+/-- The set in the definition of `Nimber.add` is nonempty. -/
+private theorem add_nonempty (a b : Nimber.{u}) :
+    {x | (∃ a' < a, a' + b = x) ∨ ∃ b' < b, a + b' = x}ᶜ.Nonempty := by
+  simp_rw [Set.nonempty_compl, Set.setOf_or, ← Set.mem_Iio, ← Set.image.eq_1]
+  apply_fun (fun a : Set Nimber ↦ Small.{u} a)
+  have : Small.{u} ↑((· + b) '' Set.Iio a ∪ (a + ·) '' Set.Iio b) := inferInstance
+  simp [this, small_congr (Equiv.Set.univ _)]
+  exact not_small_nimber
```
b-mehta, line 222:
```suggestion
  simpa [this, small_congr (Equiv.Set.univ _)] using not_small_nimber
```
looks like it might work, and eliminates a non-terminal simp.
If it doesn't work, then try to eliminate the non-terminal simp in some other way

Mathlib/SetTheory/Ordinal/Nimber.lean
```diff
@@ -0,0 +1,362 @@
...
+
+protected theorem add_zero (a : Nimber) : a + 0 = a := by
+  apply le_antisymm
+  · apply add_le_of_forall_ne
+    · intro a' ha
+      rw [Nimber.add_zero]
+      exact ha.ne
+    · intro _ h
+      exact (Nimber.not_lt_zero _ h).elim
+  · -- by_contra! doesn't work for whatever reason.
```
b-mehta, line 312: This seems worth investigating further - can leave that to another pr though

b-mehta (approved):
maintainer merge

Mathlib/SetTheory/Ordinal/Nimber.lean
vihdzp: This shouldn't be too difficult to prove using induction plus [`Nat.lt_xor_cases`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Bitwise.html#Nat.lt_xor_cases), but it does require porting some of the API of `Ordinal.natCast` over, so I'll do it in a separate PR.

Something that might also be good to do in said PR is introduce the notation `∗x` for `Ordinal.toNimber x`.

jcommelin (commented):
Thanks :tada:

bors merge

PR #16092, opened 2024-08-23
[Merged by Bors] - chore(CategoryTheory/Sites): address a bunch of porting notes

dagurtomas:
---


Mathlib/CategoryTheory/Sites/Grothendieck.lean
```diff
@@ -111,6 +111,10 @@ theorem mem_sieves_iff_coe : S ∈ J.sieves X ↔ S ∈ J X :=
   Iff.rfl
 -/
 
+@[simp]
+theorem mem_sieves_iff_coe : S ∈ J.sieves X ↔ S ∈ J X :=
+  Iff.rfl
+
```
dagurtomas, line 117:
```suggestion
```

Mathlib/CategoryTheory/Sites/Grothendieck.lean
```diff
@@ -409,10 +404,11 @@ instance : CoeFun (J.Cover X) fun _ => ∀ ⦃Y⦄ (_ : Y ⟶ X), Prop :=
   ⟨fun S => S.sieve⟩
 
 /-
-Porting note: This is now a syntactic tautology.
+Porting note: This lemma was in mathlib 3, but doesn't compile anymore because of the removed
```
mattrobball, line 407: Do you want to restore the `Coe (J.Cover X) (Sieve X)` instance? It is fine as a `CoeOut` (ie we start with covering and try to fill in as opposed as the reverse). All the commented instances work with that.

mattrobball: The `CoeFun` could just be 
```lean
instance : CoeFun (J.Cover X) fun _ => ∀ ⦃Y⦄ (_ : Y ⟶ X), Prop :=
  ⟨fun S => (S : Sieve X)⟩
```

mattrobball: Either way. I think that some decision should be made and these comments purged.

mattrobball: Otherwise this looks good to me.

dagurtomas: I did this, and the `coe_fun_coe` lemma became a syntactic tautology again. 

If I try to give a `DFunLike` instance for `J.Cover X` instead of the `CoeFun`, I cannot avoid the annoying notation `S _ f` described in the porting note above. Should we just remove the `coe_fun_coe` comment? What about the one about the notation in the `CoeFun` instance?

mattrobball: I would guess because it's `CoeFun`'s all the way down (eg `Sieve` has a `CoeFun` not a `FunLike`).

mattrobball:
bors merge

PR #16093, opened 2024-08-23
[Merged by Bors] - chore(CategoryTheory/Limits): remove two porting notes

dagurtomas:
This appears to have been a bug in the `simpNF` linter which is now fixed

---


joelriou:
Thanks!

bors merge

PR #16094, opened 2024-08-23
[Merged by Bors] - feat(Algebra/Ring): generalise and extend material about sums of squares and semireal rings

artie2000:
* Generalise definition of a semireal ring, as discussed at https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Formally.20real.20fields
* Add facts about multiplicative structure of sums of squares -- preserved under multiplication, form a semiring, set of sums of squares is the semiring generated by the squares.
* Add some lemmas to simpset
* Update documentation

Deletions:
* `mem_sumSq_of_isSquare` (statement changed to `IsSquare.isSumSq`)
---

This is work towards formalising some real algebra (see https://github.com/artie2000/real_closed_field).

- [x] depends on: #20269
- [x] depends on: #20271
- [x] depends on: #20272
- [x] depends on: #20449
- [x] depends on: #20451
- [x] depends on: #20526
- [x] depends on: #20528


Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -28,19 +27,19 @@ semireal.
...
+A semireal ring is a commutative ring (with unit) in which `-1` is *not* a sum of
+squares. We define the class `IsSemireal R` for all structures equipped with
+a multiplication, an addition, a multiplicative unit and an additive unit.
 -/
 @[mk_iff]
-class IsSemireal [AddMonoid R] [Mul R] [One R] [Neg R] : Prop where
-  non_trivial         : (0 : R) ≠ 1
-  not_isSumSq_neg_one : ¬IsSumSq (-1 : R)
+class IsSemireal [Add R] [Mul R] [One R] [Zero R] : Prop where
+  not_IsSumSq_neg_one (a : R) (ssa : IsSumSq a): ¬ (a + 1 = 0)
```
jcommelin, line 36:
```suggestion
  not_IsSumSq_neg_one (a : R) (ssa : IsSumSq a) : ¬ (a + 1 = 0)
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
matematiflo, line 1: Would it also make sense to correct and get rid of the deprecated declarations? These were due to an error in the convention used (`isSumSq` is preferable over `IsSumSq` *etc*).

matematiflo: I would think so! But if you prefer to wait for @YaelDillies's opinion, that's also fine (there's no rush).

FM22: I was actually worried about some project *you* might have, so

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -28,19 +27,16 @@ semireal.
...
-
-@[deprecated (since := "2024-08-09")] alias isSemireal := IsSemireal
-@[deprecated (since := "2024-08-09")] alias isSemireal.neg_one_not_SumSq :=
-  IsSemireal.not_isSumSq_neg_one
-
-instance [LinearOrderedRing R] : IsSemireal R where
-  non_trivial := zero_ne_one
-  not_isSumSq_neg_one := fun h ↦ (not_le (α := R)).2 neg_one_lt_zero h.nonneg
+class IsSemireal [Add R] [Mul R] [One R] [Zero R] : Prop where
+  not_IsSumSq_neg_one (a : R) (ssa : IsSumSq a) : ¬ (1 + a = 0)
```
YaelDillies, line 36:
```suggestion
  not_isSumSq_neg_one (a : R) (ssa : IsSumSq a) : 1 + a \ne 0
```
or maybe
```suggestion
  add_one_ne_zero_of_isSumSq (a : R) (ssa : IsSumSq a) : a + 1 \ne 0
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -22,90 +23,150 @@ sum of squares in `R`.
 
 ## Auxiliary declarations
 
-- The type `sumSqIn R` : in an additive monoid with multiplication `R`, we introduce the type
-`sumSqIn R` as the submonoid of `R` whose carrier is the subset `{S : R | IsSumSq S}`.
+- The terms `AddMonoid.sumSqIn R` and `Subsemiring.sumSqIn R` :
+in an additive monoid with multiplication `R` and a semiring `R`, we introduce the terms
+`AddMonoid.sumSqIn R` and `Subsemiring.sumSqIn R` as the submonoid and subsemiring, respectively,
+of `R` whose carrier is the subset `{S : R | IsSumSq S}`.
```
YaelDillies, line 29: Are they really auxiliary definitions?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -22,90 +23,150 @@ sum of squares in `R`.
...
+if `S1` and `S2` are sums of squares, then `S1 + S2` is a sum of squares. -/
+theorem IsSumSq.add [AddMonoid R] [Mul R] {S1 S2 : R}
+    (h1 : IsSumSq S1) (h2 : IsSumSq S2) : IsSumSq (S1 + S2) := by
+  induction h1 with
+  | zero             => rw [zero_add]; exact h2
   | sq_add a S pS ih => rw [add_assoc]; exact IsSumSq.sq_add a (S + S2) ih
 
-@[deprecated (since := "2024-08-09")] alias isSumSq.add := IsSumSq.add
+/-- In an additive unital magma with multiplication, `x * x` is a sum of squares for all `x`. -/
+theorem isSumSq_mul_self [AddZeroClass R] [Mul R] (a : R) : IsSumSq (a * a) := by
```
YaelDillies, line 65:
```suggestion
theorem IsSumSq.mul_self [AddZeroClass R] [Mul R] {a : R} : IsSumSq (a * a) := by
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -22,90 +23,150 @@ sum of squares in `R`.
...
+  rw [← add_zero (a * a)]
+  exact IsSumSq.sq_add _ _ IsSumSq.zero
 
-variable (R) in
+/-- In an additive unital magma with multiplication `R`, squares in `R` are sums of squares.
+By definition, a square in `R` is a term `x : R` such that `x = y * y` for some `y : R`
+and in Mathlib this is known as `IsSquare R` (see Mathlib.Algebra.Group.Even). -/
+theorem isSumSq_isSquare [AddZeroClass R] [Mul R] {x : R} (hx : IsSquare x) : IsSumSq x := by
```
YaelDillies, line 72:
```suggestion
theorem IsSquare.isSumSq [AddZeroClass R] [Mul R] {x : R} (hx : IsSquare x) : IsSumSq x := by
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -22,90 +23,150 @@ sum of squares in `R`.
...
+    {p : (S : R) → (h : IsSumSq S) → Prop} (S : R) (hS : IsSumSq S)
+    (zero : p 0 zero)
+    (sq_add : ∀ x S, (hS : IsSumSq S) → (hx : IsSquare x) → p S hS →
+      p (x + S) (by rcases hx with ⟨y, hy⟩; rw [hy]; exact sq_add y S hS)) : p S hS := by
+  induction hS with
+  | zero => exact zero
+  | sq_add a S hS hS_ih => apply sq_add (a * a) S hS _ hS_ih; use a
+
+/-- Helper lemma for `IsSumSq.mul`. -/
+theorem IsSumSq.mul_isSquare [NonUnitalCommSemiring R] {S x : R}
```
YaelDillies, line 88:
```suggestion
private theorem IsSumSq.mul_isSquare [NonUnitalCommSemiring R] {S x : R}
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -22,90 +23,150 @@ sum of squares in `R`.
...
+                              exact IsSumSq.add (isSumSq_isSquare (IsSquare.mul hy hx)) ih
+
+/-- In a (not necessarily unital) commutative semiring,
+if `S1` and `S2` are sums of squares, then `S1 * S2` is a sum of squares. -/
+theorem IsSumSq.mul [NonUnitalCommSemiring R] {S1 S2 : R}
+    (h1 : IsSumSq S1) (h2 : IsSumSq S2) : IsSumSq (S1 * S2) := by
+  induction' S2, h2 using IsSumSq.induction_alt
+  case zero => simpa using zero
+  case sq_add y T hT hy ih => rw [left_distrib]
+                              exact IsSumSq.add (IsSumSq.mul_isSquare h1 hy) ih
```
YaelDillies, line 102: I don't understand why you are doing induction here. Can't you first prove the obvious theorem that `IsSumSq a` iff `a` is a sum of squares (in terms of `Finset
sum`) then use `Finset.sum_mul_sum` to express the product of sums of squares as a sum of squares.

FM22: The predicate is defined inductively, so an induction proof is easy. I think the fact that a sum of squares can be written as.. as sum of squares is somehow implicit in the AddSubmonoid closure theorem?

Mathlib/Algebra/Order/Ring/Basic.lean
```diff
@@ -121,6 +121,13 @@ lemma Bound.pow_le_pow_right_of_le_one_or_one_le (h : 1 ≤ a ∧ n ≤ m ∨ 0
 
 end OrderedSemiring
 
+/-- Turn an ordered domain into a strict ordered ring. -/
```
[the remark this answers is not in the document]
FM22: Ops I forgot to revert these!
Can you suggest the same changes in the other PR?

Mathlib/Algebra/Ring/SumsOfSquares.lean
FM22: was following existing docs, changed

Mathlib/Algebra/Ring/SumsOfSquares.lean
FM22: done

Mathlib/Algebra/Ring/SumsOfSquares.lean
YaelDillies: As mentioned on Xena, the predicate is defined inductively because it would be annoying to state it using an indexed sum, but that doesn't forbid you from stating the indexed sum. And yes it is somehow implicit in the closure theorem, but you shouldn't expect your users from knowing that when they search for the fact that a sum of squares is a sum of squares!

Mathlib/Algebra/Ring/Semireal/Defs.lean
YaelDillies: Looks like you forgot to apply this?

YaelDillies (commented):
You are missing `IsSumSq.sum` and `IsSumSq.prod`

Mathlib/Algebra/Ring/SumsOfSquares.lean
FM22: Have found this theorem now (`Submonoid.exists_multiset_of_mem_closure`); still unsure about using `Finset.sum_mul_sum`. In particular, is there a lemma flattening two sums into one?

FM22: Basically I think `Finset.sum` is really fiddly and so the inducive definition is easier to work with. What I will do though is prove the 'obvious theorem' to characterise `IsSumSq` in terms of `Finset.sum`.

YaelDillies: > Basically I think `Finset.sum` is really fiddly

That's an issue, not an issme. I can teach you how to use it. It really shouldn't be that hard

matematiflo: > As mentioned on Xena, the predicate is defined inductively because it would be annoying to state it using an indexed sum, but that doesn't forbid you from stating the indexed sum. And yes it is somehow implicit in the closure theorem, but you shouldn't expect your users from knowing that when they search for the fact that a sum of squares is a sum of squares!

Hi @YaelDillies ! May I ask what you mean when say 'as mentioned on Xena'?

YaelDillies: Artie brought up my comment above on Kevin's Xena Discord server, and I replied there

Mathlib/Algebra/Ring/Semireal/Defs.lean
FM22: I just checked and I'm pretty sure it's correct as of now.

YaelDillies: Then maybe you forgot to push. I still see `¬ (1 + a = 0)` instead of `1 + a ≠ 0`

FM22: Oh sorry, I only saw the name change, my bad

FM22: Fixed

Mathlib/Algebra/Ring/SumsOfSquares.lean
FM22: Done

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -16,96 +18,175 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
-## Auxiliary declarations
-
-- The type `sumSqIn R` : in an additive monoid with multiplication `R`, we introduce the type
-`sumSqIn R` as the submonoid of `R` whose carrier is the subset `{S : R | IsSumSq S}`.
+- The terms `AddMonoid.sumSqIn R` and `Subsemiring.sumSqIn R` :
+in an additive monoid with multiplication `R` and a semiring `R`, we introduce the terms
+`AddMonoid.sumSqIn R` and `Subsemiring.sumSqIn R` as the submonoid and subsemiring, respectively,
+of `R` whose carrier is the subset `{S : R | IsSumSq S}`.
 
 ## Theorems
```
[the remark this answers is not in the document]
FM22: Should I keep any? Or just delete the whole section? The previous version just listed everything, but I don't understand why.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -16,96 +18,175 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
 -/
 @[mk_iff]
-inductive IsSumSq [Add R] [Zero R] : R → Prop
+inductive IsSumSq [Mul R] [Add R] [Zero R] : R → Prop
   | zero                              : IsSumSq 0
-  | sq_add (a S : R) (pS : IsSumSq S) : IsSumSq (a * a + S)
-
-@[deprecated (since := "2024-08-09")] alias isSumSq := IsSumSq
-
+  | sq_add (a S : R) (hS : IsSumSq S) : IsSumSq (a * a + S)
```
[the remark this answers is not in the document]
FM22: hm
won't this cause a problem with requiring inference of `a`?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -16,96 +18,175 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
 squares, then for all `a : R`, `a * a + S` is a sum of squares in `R`.
 -/
 @[mk_iff]
-inductive IsSumSq [Add R] [Zero R] : R → Prop
+inductive IsSumSq [Mul R] [Add R] [Zero R] : R → Prop
   | zero                              : IsSumSq 0
-  | sq_add (a S : R) (pS : IsSumSq S) : IsSumSq (a * a + S)
-
-@[deprecated (since := "2024-08-09")] alias isSumSq := IsSumSq
-
```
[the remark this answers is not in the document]
FM22: Flo and I thought it would be fine, here
https://github.com/leanprover-community/mathlib4/pull/16094#discussion_r1729199213
Should I reinstate deprecated names for those declarations which I haven't rewritten? I can understand why that would be sensible.

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
FM22: OK, it doesn't cause any problems within the current file, but I've been forced to use inference in a few places, and I'm worried that will cause problems down the line. Maybe I don't understand something about how this works in practice, though.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -117,7 +117,7 @@ theorem AddSubmonoid.closure_isSquare [AddMonoid R] [Mul R] :
   | sq_add a S _ ih  => exact AddSubmonoid.add_mem _ (subset_closure ⟨a, rfl⟩) ih
 
 open AddSubmonoid in
-/-- A term of `R` satisfies `IsSumSq` if and only if it can be written as `∑ i ∈ I, x i`. -/
+/-- A term of `R` satisfies `IsSumSq` if and only if it can be written as `∑ i ∈ I, x i * x i`. -/
 theorem isSumSq_iff_finsum [AddCommMonoid R] [Mul R] (a : R) :
```
YaelDillies, line 118: `finsum` already means something else
```suggestion
theorem isSumSq_iff_exists_finsetSum [AddCommMonoid R] [Mul R] (a : R) :
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
YaelDillies: Yes, it looks like the policy is to keep the deprecations for 3-6 months and definitely at least one month so that they make it to at least one `stable` release

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
YaelDillies: Please make `S` implicit, at least

Mathlib/Algebra/Ring/SumsOfSquares.lean
FM22: Reverted as per https://github.com/leanprover-community/mathlib4/pull/16094#discussion_r1732221358

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
YaelDillies: I think just delete

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
FM22: I've made them both implicit; I think `a` should always be clear from the context, and you can always specify the type for the elaborator if needed

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -61,7 +61,7 @@ theorem IsSumSq.induction_alt [Mul R] [Add R] [Zero R]
       p (x + S) (by rcases hx with ⟨y, hy⟩; rw [hy]; exact sq_add hS)) : p S hS := by
   induction hS with
   | zero            => exact zero
-  | sq_add hS hS_ih => exact sq_add hS (isSquare_mul_self _) hS_ih
+  | sq_add hS hS_ih => exact sq_add hS (IsSquare.mul_self _) hS_ih
```
YaelDillies, line 64:
```suggestion
  | sq_add hS hS_ih => exact sq_add hS (.mul_self _) hS_ih
```

FM22: done

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
FM22: OK, done

artie2000:
I'm back! Updated and cleaned up everything.

matematiflo (approved):
The main changes in this PR appear in the module `Mathlib.Algebra.Ring.SumsOfSquares`. 

A new induction principle for sums of squares has been added (though it has not found a use yet) and several results have been added to treat inductively-defined sums of squares using a sum operator instead. Several results that were previously proved using induction have now been refactored to use this approach. 

The added results are all useful, so my opinion is that we should merge these changes. My only concern lies with the `@[deprecated (since := "0204-08-09")]` attributes, which I wish we could get rid of on the occasion of the current refactoring.

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -15,105 +16,204 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
+    (∃ (ι : Type u) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a) := by
+  rcases exists_sum ha with ⟨ι, I, x, _⟩
+  exact ⟨ULift.{u} ι, .map (Equiv.ulift.symm.toEmbedding) I, x ∘ (Equiv.ulift.toEmbedding),
+    by simpa⟩
+
+/-- A term of `R` satisfies `IsSumSq` if and only if it can be written as `∑ i ∈ I, x i`
+where each `x i` is a square in `R`. -/
+theorem isSumSq_iff_exists_sum [AddCommMonoid R] [Mul R] (a : R) :
+    IsSumSq a ↔
+    (∃ (ι : Type) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a) :=
```
YaelDillies, line 139:
```suggestion
      ∃ (ι : Type) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a :=
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -15,105 +16,204 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
 
+/-- A term of `R` satisfying `IsSumSq` can be written as `∑ i ∈ I, x i`
+where each `x i` is a square in `R`. -/
+theorem IsSumSq.exists_sum [AddCommMonoid R] [Mul R] {a : R} (ha : IsSumSq a) :
+    (∃ (ι : Type) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a) :=
+  AddSubmonoid.exists_finset_sum_of_mem_closure (s := {x : R | IsSquare x}) (by simpa)
+
+/-- Universe-polymorphic version of `exists_sum`. -/
+theorem IsSumSq.exists_sum' [AddCommMonoid R] [Mul R] {a : R} (ha : IsSumSq a) :
+    (∃ (ι : Type u) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a) := by
```
YaelDillies, line 130:
```suggestion
    ∃ (ι : Type u) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a := by
```

artie2000:
> The main changes in this PR appear in the module `Mathlib.Algebra.Ring.SumsOfSquares`.
> 
> A new induction principle for sums of squares has been added (though it has not found a use yet) and several results have been added to treat inductively-defined sums of squares using a sum operator instead. Several results that were previously proved using induction have now been refactored to use this approach.
> 
> The added results are all useful, so my opinion is that we should merge these changes. My only concern lies with the `@[deprecated (since := "0204-08-09")]` attributes, which I wish we could get rid of on the occasion of the current refactoring.

The new material is useful generally, but specifically is required to prove that a field is semireal if and only if it can be ordered -- PR on the way ;)

And yeah not too sure about the alternative induction principle -- the motivation for it is that `IsSquare x` is easier to work with than `x * x` in practice, but I haven't actually found an application where it matters.

Mathlib/Algebra/Ring/Semireal/Defs.lean
```diff
@@ -28,19 +26,22 @@ semireal.
...
 @[deprecated (since := "2024-08-09")] alias isSemireal := IsSemireal
 @[deprecated (since := "2024-08-09")] alias isSemireal.neg_one_not_SumSq :=
-  IsSemireal.not_isSumSq_neg_one
+  IsSemireal.add_one_ne_zero_of_isSumSq
 
-instance [LinearOrderedRing R] : IsSemireal R where
-  non_trivial := zero_ne_one
-  not_isSumSq_neg_one := fun h ↦ (not_le (α := R)).2 neg_one_lt_zero h.nonneg
+theorem IsSemireal.not_isSumSq_neg_one {R : Type*} [AddGroup R] [One R] [Mul R] [IsSemireal R]:
+    ¬ IsSumSq (-1 : R) := (by simpa using add_one_ne_zero_of_isSumSq ·)
```
YaelDillies, line 42: Doesn't this work?
```suggestion
    ¬ IsSumSq (-1 : R) := by simpa using add_one_ne_zero_of_isSumSq
```

Mathlib/Algebra/Group/Submonoid/Membership.lean
```diff
@@ -274,6 +275,18 @@ theorem induction_of_closure_eq_top_right {s : Set M} {p : M → Prop} (hs : clo
   | one => exact H1
   | mul_right x _ y hy ih => exact Hmul x y hy ih
 
+@[to_additive]
+theorem exists_finset_prod_of_mem_closure {M : Type*} [CommMonoid M] {s : Set M} {x : M}
```
YaelDillies, line 279:
```suggestion
theorem exists_finsetProd_of_mem_closure {M : Type*} [CommMonoid M] {s : Set M} {x : M}
```

Mathlib/Algebra/Group/Submonoid/Membership.lean
```diff
@@ -274,6 +275,18 @@ theorem induction_of_closure_eq_top_right {s : Set M} {p : M → Prop} (hs : clo
...
+    (hx : x ∈ closure s) :
+    ∃ (ι : Type) (I : Finset ι) (f : ι → M), (∀ i ∈ I, f i ∈ s) ∧ ∏ i ∈ I, f i = x := by
+  induction hx using Submonoid.closure_induction_left
+  case one                      => exact ⟨Empty, ∅, Empty.elim, by simp⟩
+  case mul_left y _ _ _ h_prod =>
+    rcases h_prod with ⟨_, I, f, _, rfl⟩;
+    exact ⟨Option I, Finset.univ, (Option.elim · y (f ·)),
+          ⟨by intro i; induction i <;> simp [*],
+          by simp [Finset.prod_attach]⟩⟩
```
YaelDillies, line 288: It usually is a bad idea to prove lemmas of the form "All elements in the closure of `s` are of form blah". Wherever you use those lemmas, there is a better proof using the closure induction principle, which is why those lemmas don't exist. Can you try rewriting your proofs without using this lemma?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -15,105 +16,204 @@ instances. Sums of squares in `R` are defined by an inductive predicate `IsSumSq
...
-  | sq_add a S _ ih  => exact AddSubmonoid.add_mem _ (subset_closure ⟨a, rfl⟩) ih
+  | zero         => exact zero_mem _
+  | sq_add _ ih  => exact add_mem (subset_closure (.mul_self _)) ih
 
 @[deprecated (since := "2024-08-09")] alias SquaresAddClosure := AddSubmonoid.closure_isSquare
 
+/-- A term of `R` satisfying `IsSumSq` can be written as `∑ i ∈ I, x i`
+where each `x i` is a square in `R`. -/
+theorem IsSumSq.exists_sum [AddCommMonoid R] [Mul R] {a : R} (ha : IsSumSq a) :
+    (∃ (ι : Type) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a) :=
```
YaelDillies, line 125:
```suggestion
    ∃ (ι : Type) (I : Finset ι) (x : ι → R), (∀ i ∈ I, IsSquare (x i)) ∧ ∑ i ∈ I, x i = a :=
```

Mathlib/Algebra/Ring/Semireal/Defs.lean
artie2000: Typeclass inference on `Semireal ?_` fails

Mathlib/Algebra/Ring/SumsOfSquares.lean
artie2000: oops cheers

Mathlib/Algebra/Group/Submonoid/Membership.lean
artie2000: This lemma already exists in the list and multiset versions, and I'm applying it in this PR to prove directly the specialisation where `s := SumSqIn R`.
Would you like me to effectively move this proof into the proof of `IsSumSq.exists_sum`?
Alternatively, we could remove all 6 explicit sum lemmas from `SumsOfSquares.lean`. In that case, lemmas such as `IsSumSq.mul` would have to be proved by induction. However, I'm confused as I thought you previously indicated that you prefer the explicit sum-rewriting proof of `IsSumSq.mul` to the inductive one.

artie2000: I've just pushed a commit testing out the version without use of explicit sums (i.e., all done by induction). I'm happy with either; what do you (and others) think?

YaelDillies: Sorry, I guess I'm not internally consistent on the matter 🙈 Could you bring it up on Zulip to see what people think?

artie2000: RFC here:
https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/RFC.3A.20using.20vs.20avoiding.20explicit.20sum.20representations

artie2000: @urkud has pointed out to me that the issue is not relevant as the result I want (`IsSumSq.mul`) has already been proved in more generality. The discussion is still worth having because the same decision will need to be made regarding material I intend to PR very soon.

I'm rewriting the PR according to @urkud 's suggestion; this involves filling in some missing general lemmas and also making a new file. Will re-request a review in the coming days. Thanks for all the comments and discussion!

Mathlib.lean
```diff
@@ -272,7 +272,8 @@ import Mathlib.Algebra.Group.Defs
 import Mathlib.Algebra.Group.Embedding
 import Mathlib.Algebra.Group.Equiv.Basic
 import Mathlib.Algebra.Group.Equiv.TypeTags
-import Mathlib.Algebra.Group.Even
+import Mathlib.Algebra.Group.Even.Basic
```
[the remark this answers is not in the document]
artie2000: I don't understand this suggestion. Could someone please explain it to me?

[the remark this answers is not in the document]
YaelDillies: `Basic` comes after `Advanced` in alphabetical order

Mathlib/Algebra/Algebra/Subalgebra/Unitization.lean
```diff
@@ -208,10 +208,36 @@ variable {R : Type*} [NonAssocSemiring R]
...
+theorem Subsemiring.coe_toNonUnitalSubsemiring (S : Subsemiring R) :
+  (S.toNonUnitalSubsemiring : Set R) = S := rfl
+
 theorem Subsemiring.one_mem_toNonUnitalSubsemiring (S : Subsemiring R) :
     (1 : R) ∈ S.toNonUnitalSubsemiring :=
   S.one_mem
 
+@[simp]
+theorem Submonoid.subsemiringClosure_toNonUnitalSubsemiring {M : Submonoid R} :
+    (M.subsemiringClosure).toNonUnitalSubsemiring = NonUnitalSubsemiring.closure M := by
```
YaelDillies, line 235:
```suggestion
    M.subsemiringClosure.toNonUnitalSubsemiring = NonUnitalSubsemiring.closure M := by
```

Mathlib/Algebra/Group/Even/Advanced.lean
```diff
@@ -0,0 +1,93 @@
...
+-/
+@[to_additive evenIn
+"In an abelian additive group `G`, the type `Submonoid.sqIn G` is
+the subgroup of even elements in `G`."]
+def squareIn : Subgroup G where
+  __ := Submonoid.squareIn G
+  inv_mem' := IsSquare.inv
+
+@[to_additive (attr := simp)]
+theorem squareIn_toSubmonoid : (squareIn G).toSubmonoid = Submonoid.squareIn G := rfl
```
YaelDillies, line 79:
```suggestion
theorem squareIn_toSubmonoid : (squareIn G).toSubmonoid = .squareIn G := rfl
```

Mathlib/Algebra/Group/Even/Basic.lean
```diff
@@ -45,6 +45,8 @@ for some `r : α`. -/
 for some `r : α`."]
 def IsSquare (a : α) : Prop := ∃ r, a = r * r
 
+@[to_additive] lemma isSquare_iff (a : α) : IsSquare a ↔ ∃ r, a = r * r := Iff.rfl
```
YaelDillies, line 48: Do we really need this? You can already do `rw [IsSquare`

YaelDillies (commented):
For reviewing, it would help if you split that PR in two or three

Mathlib/Algebra/Group/Even/Basic.lean
artie2000: My intention was to future-proof against definition changes and avoid possible defeq abuse in my code. This is can be used in more places eg `rw [IsSquare]` eg `(hx : IsSquare x) ... rcases hx with ...` becomes `rcases (isSquare_iff x).mp hx`.

I'm unaware of the existing conventions around definitions, though.

artie2000:
> For reviewing, it would help if you split that PR in two or three

Is there some way to link them together, so that eg the changes to `Group.Even` file structure make sense?

Mathlib/Algebra/Group/Even/Basic.lean
YaelDillies: I doubt this definition will change ever again, so I don't think it's worth the trouble of avoiding defeq abuse

YaelDillies:
> Is there some way to link them together, so that eg the changes to `Group.Even` file structure make sense?

Do you mean the `- [ ] depends on:` syntax in the PR description?

Mathlib/Algebra/Group/Even/Basic.lean
artie2000: OK, will revert.
tbh I don't really understand when defeq is "bad" -- do you know of any resources documenting it, or is it just vibes?

YaelDillies: It's mostly vibes, but I'm sure we could extract some nice guideline from all the examples we have in mathlib

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,123 +1,206 @@
...
-- `IsSumSq.add`: if `S1` and `S2` are sums of squares in `R`, then so is `S1 + S2`.
-- `IsSumSq.nonneg`: in a linearly ordered semiring `R` with an `[ExistsAddOfLE R]` instance, sums of
-squares are non-negative.
-- `mem_sumSq_of_isSquare`: a square is a sum of squares.
-- `AddSubmonoid.closure_isSquare`: the submonoid of `R` generated by the squares in `R` is the set
-of sums of squares in `R`.
+- The predicate `IsSumSq : R → Prop`: for a type `R` with addition, multiplication and a zero.,
+an inductive predicate defining the property of being a sum of squares in `R`.
+`0 : R` is a sum of squares and if `S` is a sum of squares, then, for all `a : R`, `a * a + s` is a
+sum of squares.
```
YaelDillies, line 21:
```suggestion
sum of squares.
```
```suggestion
  an inductive predicate defining the property of being a sum of squares in `R`.
  `0 : R` is a sum of squares and if `S` is a sum of squares, then, for all `a : R`, `a * a + s` is
  a sum of squares.
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,123 +1,206 @@
...
+- The terms `AddMonoid.sumSq R` and `Subsemiring.sumSq R`: respectively
+the submonoid and subsemiring of sums of squares in the corresponding structure `R`.
```
YaelDillies, line 23:
```suggestion
  the submonoid and subsemiring of sums of squares in the corresponding structure `R`.
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,123 +1,206 @@
...
+namespace AddSubmonoid
+variable {T : Type*} [AddMonoid T] [Mul T] {s : T}
 
-variable (R) in
+variable (T) in
 /--
-In an additive monoid with multiplication `R`, `AddSubmonoid.sumSq R` is the submonoid of sums of
-squares in `R`.
+In an additive monoid with multiplication `R`, `AddSubmonoid.sumSq R`
+is the submonoid of sums of squares in `R`.
```
YaelDillies, line 70: Why are you rewrapping this?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,123 +1,206 @@
...
 -/
-def sumSq [AddMonoid R] : AddSubmonoid R where
-  carrier   := {S : R | IsSumSq S}
+def sumSq : AddSubmonoid T where
+  carrier := {s : T | IsSumSq s}
   zero_mem' := IsSumSq.zero
-  add_mem'  := IsSumSq.add
+  add_mem' := IsSumSq.add
```
YaelDillies, line 75: If you are going to fiddle with this, you could golf it even further
```suggestion
  add_mem' := .add
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,123 +1,206 @@
...
 
-@[deprecated (since := "2024-08-09")] alias SumSqIn := sumSq
-@[deprecated (since := "2025-01-03")] alias sumSqIn := sumSq
+@[simp] theorem mem_sumSq : s ∈ sumSq T ↔ IsSumSq s := Iff.rfl
+@[simp, norm_cast] theorem coe_sumSq : sumSq T = {s : T | IsSumSq s} := rfl
```
YaelDillies, line 78: Why are the deprecations gone?

YaelDillies (commented):
There are still a lot of cleanups for a "feat" PR. Can you possibly split out all the cleanups and refactors to another PR? It's currently hard to tell what's new from what's not

artie2000:
@YaelDillies the file changed in the time I made this, to add some of the features I am adding, but not in a unified way. I'll do my best to separate out refactors, but essentially I've rewritten the file from scratch so I don't know how easy it will be.

Mathlib/Algebra/Ring/SumsOfSquares.lean
artie2000: oops, messed up the merge

Mathlib/Algebra/Ring/SumsOfSquares.lean
artie2000: another merge issue

Mathlib/Algebra/Ring/SumsOfSquares.lean
artie2000: I think I've done this now?

YaelDillies:
> I've rewritten the file from scratch

I understand the sentiment but I think you can actually tell the story in a way that makes the existing file being partly modified, rather than rewritten from scratch

> the documentation changes I've made depend on the refactors I've done.

I would be fine with leaving out documentation updates until this PR

Mathlib/Algebra/Ring/SumsOfSquares.lean
artie2000: oh no, sorry, the deprecations are gone because this lemma has been deleted - see my updated PR description

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -1,11 +1,13 @@
...
-Authors: Florent Schaffhauser
+Authors: Florent Schaffhauser, Artie Khovanov
 -/
-import Mathlib.Algebra.BigOperators.Group.Finset.Basic
-import Mathlib.Algebra.Group.Submonoid.Basic
-import Mathlib.Algebra.Order.Ring.Defs
+import Mathlib.Algebra.Order.Ring.Basic
+import Mathlib.Algebra.Ring.Subsemiring.Basic
+import Mathlib.Algebra.Group.Subgroup.Even
+import Mathlib.Algebra.Ring.Parity
```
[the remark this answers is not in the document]
artie2000: #min_imports doesn't work here because it doesn't think the last 2 imports are necessary (the last is, the second last isn't but not importing it causes issues downstream for the automation)
Should these be sorted alphabetically?

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -37,10 +40,22 @@ inductive IsSumSq [Mul R] [Add R] [Zero R] : R → Prop
...
+    {s : R} (h : IsSumSq s) : motive s h :=
+  match h with
+  | .zero        => zero
+  | .sq_add _ hs => sq_add (.mul_self _) hs (rec' zero sq_add _)
+
 /--
 In an additive monoid with multiplication,
 if `s₁` and `s₂` are sums of squares, then `s₁ + s₂` is a sum of squares.
 -/
+@[aesop unsafe 90% apply]
```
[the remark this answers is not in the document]
artie2000: I think it fits better in #20558, but yes I can defer to that. Strictly speaking the IsSquare and IsSumSq automation isn't covered by that explanation, but it's analogous.

Mathlib/Algebra/Ring/SumsOfSquares.lean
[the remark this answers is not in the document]
YaelDillies: Yes, alphabetically

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -128,14 +125,15 @@ In an additive commutative monoid with multiplication,
 `∑ i ∈ I, x i`, where each `x i` is a square, is a sum of squares.
 -/
 theorem IsSumSq.sum_isSquare [AddCommMonoid R] [Mul R] {ι : Type*} (I : Finset ι) {x : ι → R}
-    (ha : ∀ i ∈ I, IsSquare <| x i) : IsSumSq (∑ i ∈ I, x i) := by aesop
+    (hx : ∀ i ∈ I, IsSquare <| x i) : IsSumSq (∑ i ∈ I, x i) :=
+  IsSumSq.sum (fun _ hi => IsSquare.isSumSq (hx _ hi))
```
YaelDillies, line 126:
```suggestion
  .sum fun _ hi => IsSquare.isSumSq (hx _ hi)
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -128,14 +125,15 @@ In an additive commutative monoid with multiplication,
...
 
 /--
 In an additive commutative monoid with multiplication,
 `∑ i ∈ I, a i * a i` is a sum of squares.
 -/
 theorem IsSumSq.sum_mul_self [AddCommMonoid R] [Mul R] {ι : Type*} (I : Finset ι) (a : ι → R) :
-    IsSumSq (∑ i ∈ I, a i * a i) := by aesop
+    IsSumSq (∑ i ∈ I, a i * a i) := IsSumSq.sum (fun _ _ => IsSumSq.mul_self _)
```
YaelDillies, line 133:
```suggestion
    IsSumSq (∑ i ∈ I, a i * a i) := .sum fun _ _ => .mul_self _
```

Mathlib/Algebra/Ring/SumsOfSquares.lean
```diff
@@ -188,7 +185,7 @@ In a commutative semiring `R`, `Subsemiring.sumSq R` is the subsemiring of sums
 -/
 def sumSq : Subsemiring T where
   __ := NonUnitalSubsemiring.sumSq T
-  one_mem' := by aesop
+  one_mem' := by simpa using IsSquare.isSumSq IsSquare.one
```
YaelDillies, line 185:
```suggestion
  one_mem' := by simpa using IsSquare.one.isSumSq
```

YaelDillies:
Thanks!

maintainer delegate

jcommelin (commented):
Thanks :tada:

bors merge

PR #16099, opened 2024-08-23
chore(Nat/GCD): split off `GCD/Defs` from `GCD/Basic`

rwst:
Move definition of `Coprime` into the new `GCD/Defs`. Reduces what is imported by `Prime/Defs` and other files.

---

### DIFF
--- a/Mathlib/Data/Nat/GCD/Defs.lean
+++ b/Mathlib/Data/Nat/GCD/Defs.lean
@@ -5,6 +5,10 @@ Authors: Jeremy Avigad, Leonardo de Moura
 -/
 import Mathlib.Order.Basic
 
+/-!
+# Definition of `Nat.Coprime`
+-/
+
 namespace Nat
 
 /-!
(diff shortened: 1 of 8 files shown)


rwst:
!bench

Mathlib/Data/Nat/GCD/Defs.lean
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Nat
+
+/-!
+### `Coprime`
+
+See also `Nat.coprime_of_dvd` and `Nat.coprime_of_dvd'` to prove `Nat.Coprime m n`.
+-/
+
+instance (m n : ℕ) : Decidable (Coprime m n) := inferInstanceAs (Decidable (gcd m n = 1))
```
Ruben-VandeVelde, line 20: Please remove this entirely, it already exists in `Batteries/Data/Nat/Gcd.lean` immediately after the definition of `Nat.Coprime`.

Ruben-VandeVelde (changes requested):
I note that the summary doesn't match what you're actually doing, and does not explain the need for the changes either. Moving the `Decidable` instance is not needed to reduce the imports in `Mathlib/Data/Nat/GCD/Basic.lean` anyway.

PR #16100, opened 2024-08-23
[Merged by Bors] - doc(Data/Finset/Lattice): fix typos in docstrings

MichaelStollBayreuth:
In the docstrings of `Finset.min'` and `Finset.max'` a hypothesis named `H` was referred to as `h`. Even though the text was consistent, I think it is confusing when the docstring uses a different name compared to the definition.

---


MichaelStollBayreuth:
I'm idly wondering why `s` is explicit in `Finset.min'` and `Finset.max'`...

urkud:
I didn't double check that this is still the case, but when the decision was made, making `s` implicit would mean pretty printing the expression as `Finset.min' _`, completely hiding the `Finset` we're talking about.

urkud:
Thanks! :tada:
bors merge

MichaelStollBayreuth:
I guess having `s` explicit also enables dot notation.

PR #16113, opened 2024-08-24
[Merged by Bors] - chore: Namespace `UniformSpace.ball` API

YaelDillies:
All these lemmas have pretty generic names

---


Mathlib/Topology/UniformSpace/UniformEmbedding.lean
```diff
@@ -158,7 +158,7 @@ theorem uniformEmbedding_subtype_val {p : α → Prop} :
 
 theorem uniformEmbedding_set_inclusion {s t : Set α} (hst : s ⊆ t) :
     UniformEmbedding (inclusion hst) where
-  comap_uniformity := by rw [uniformity_subtype, uniformity_subtype, comap_comap]; rfl
+  comap_uniformity := by rw [uniformity_subtype, uniformity_subtype, Filter.comap_comap]; rfl
```
grunweg, line 161: Flagging to other reviewers: this is necessary as there is also `UniformSpace.comap_comap`, which comes in scope with the `open UniformSpace` above.

grunweg (commented):
`ball` is used in many contexts, and this one clashes with `Metric.ball` --- seems reasonable to namespaces it.
maintainer merge

Mathlib/Topology/UniformSpace/UniformEmbedding.lean
```diff
@@ -14,7 +14,7 @@ Extension of uniform continuous functions.
 -/
 
 
-open Filter Function Set Uniformity Topology
+open Filter Function Set Uniformity UniformSpace Topology
```
urkud, line 17: Would it be easier to not open this NS? Or open it hiding `comap_comap`? If you open it, then you can drop `UniformSpace.` here and there.

urkud:
Also, could you please add deprecated aliases?

YaelDillies:
Actually I don't think I should add deprecated aliases as they would conflict with the namespaced lemmas, forbidding everyone to `open UniformSpace`

urkud:
Thanks! :tada:
bors merge

PR #16114, opened 2024-08-24
[Merged by Bors] - chore: Rename `Real.RCLike` to `Real.instRCLike`

YaelDillies:
... to avoid the obvious name clash

---


grunweg (commented):
Wasn't the convention to start instance names with `inst` anyway? Looks reasonable, thanks!
maintainer merge

YaelDillies:
Yes it is. I guess it was never renamed since the port

urkud:
Thanks! :tada:
bors merge

PR #16117, opened 2024-08-24
[Merged by Bors] - feat: A function is eventually bounded above/below iff it's eventually `BddAbove`/`BddBelow`

YaelDillies:
Translate between two statements that are "obviously the same"

---


jcommelin (commented):
Thanks :tada:

bors merge

PR #16118, opened 2024-08-24
[Merged by Bors] - feat: A convex function bounded on a set is bounded on its convex hull

YaelDillies:
This is a version of the maximum principle for convex functions. Also make the existing maximum principles more syntactically general for ease of use.

This syntactic generalisation is useful because often the domain on which a function is continuous is a ball, a half-space, an interval... which is merely a superset of the set on which we want to apply the maximum principle. Before this PR, one therefore has to restrict the convexity of a function from its natural domain to the set on which we want to apply the maximum principle. This is unnecessarily cumbersome.

---

### DIFF
--- a/Mathlib/Analysis/Convex/Jensen.lean
+++ b/Mathlib/Analysis/Convex/Jensen.lean
@@ -254,10 +254,16 @@ theorem ConvexOn.le_sup_of_mem_convexHull {t : Finset E} (hf : ConvexOn 𝕜 s f
   exact (hf.map_centerMass_le hw₀ (by positivity) hts).trans
     (centerMass_le_sup hw₀ <| by positivity)
 
-theorem inf_le_of_mem_convexHull {t : Finset E} (hf : ConcaveOn 𝕜 s f) (hts : ↑t ⊆ s)
+theorem ConvexOn.inf_le_of_mem_convexHull {t : Finset E} (hf : ConcaveOn 𝕜 s f) (hts : ↑t ⊆ s)
     (hx : x ∈ convexHull 𝕜 (t : Set E)) :
     t.inf' (coe_nonempty.1 <| convexHull_nonempty_iff.1 ⟨x, hx⟩) f ≤ f x :=
   hf.dual.le_sup_of_mem_convexHull hts hx
+
+@[deprecated (since := "2024-08-25")]
+alias le_sup_of_mem_convexHull := ConvexOn.le_sup_of_mem_convexHull
+
+@[deprecated (since := "2024-08-25")]
+alias inf_le_of_mem_convexHull := ConvexOn.inf_le_of_mem_convexHull
 
 /-- If a function `f` is convex on `s`, then the value it takes at some center of mass of points of
 `s` is less than the value it takes on one of those points. -/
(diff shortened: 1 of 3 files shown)


Mathlib/Analysis/Convex/Jensen.lean
```diff
@@ -244,25 +244,25 @@ end Jensen
...
 
 variable [LinearOrderedField 𝕜] [AddCommGroup E] [LinearOrderedAddCommGroup β] [Module 𝕜 E]
-  [Module 𝕜 β] [OrderedSMul 𝕜 β] {s : Set E} {f : E → β} {t : Finset ι} {w : ι → 𝕜} {p : ι → E}
+  [Module 𝕜 β] [OrderedSMul 𝕜 β] {s : Set E} {f : E → β} {w : ι → 𝕜} {p : ι → E}
   {x y z : E}
 
-theorem le_sup_of_mem_convexHull {s : Finset E} (hf : ConvexOn 𝕜 (convexHull 𝕜 (s : Set E)) f)
-    (hx : x ∈ convexHull 𝕜 (s : Set E)) :
-    f x ≤ s.sup' (coe_nonempty.1 <| convexHull_nonempty_iff.1 ⟨x, hx⟩) f := by
+theorem ConvexOn.le_sup_of_mem_convexHull {t : Finset E} (hf : ConvexOn 𝕜 s f) (hts : ↑t ⊆ s)
```
grunweg, line 250: Can you add deprecated aliasses for this and the `inf` version, please?

Mathlib/Analysis/Convex/Jensen.lean
```diff
@@ -244,25 +244,25 @@ end Jensen
...
   obtain ⟨w, hw₀, hw₁, rfl⟩ := mem_convexHull.1 hx
-  exact (hf.map_centerMass_le hw₀ (by positivity) <| subset_convexHull _ _).trans
+  exact (hf.map_centerMass_le hw₀ (by positivity) hts).trans
     (centerMass_le_sup hw₀ <| by positivity)
 
-theorem inf_le_of_mem_convexHull {s : Finset E} (hf : ConcaveOn 𝕜 (convexHull 𝕜 (s : Set E)) f)
-    (hx : x ∈ convexHull 𝕜 (s : Set E)) :
-    s.inf' (coe_nonempty.1 <| convexHull_nonempty_iff.1 ⟨x, hx⟩) f ≤ f x :=
-  le_sup_of_mem_convexHull hf.dual hx
+theorem inf_le_of_mem_convexHull {t : Finset E} (hf : ConcaveOn 𝕜 s f) (hts : ↑t ⊆ s)
```
grunweg, line 257: So, you renaming the dual of this lemma, but not this one - please be consistent. I see a point in favour of namespacing this under `ConvexOn`, as well as one against (with your syntactic generalisation, `s` and `t` are only slightly related).

grunweg (approved):
I'm not an expert of this part of the library, but these changes make sense.
I have two small comments, see below.
Would you like to say a few more words in the commit message why you're changing the maximum principles syntactically/do you have a particular reason? (I'd agree these look useful regardless, but if you really need them, this seems worth documenting for others to come here.)

maintainer delegate

YaelDillies:
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #16119, opened 2024-08-24
[Merged by Bors] - feat(Tactic/CategoryTheory/ToApp): Add `to_app` attribute

callesonne:
This PR adds the attribute `to_app`, which generates new declarations similar to how `reassoc` works.

The point is the following: given a lemma about an equality of 2-morphisms in some bicategory, `to_app` specializes this lemma so that the equality is taking place in `Cat`, then it applies `NatTrans.congr_app` and simplifies the resulting expression.

Lemmas of this form are useful to apply when working with a (simplified) expression of 1-morphisms in a category where parts (but not all) of this expression consists of components of natural transformations. So this attribute makes it easier to apply purely bicategorical lemmas in such situations.

A situation where I think that this is useful is when working with pseudofunctors valued in `Cat`. See the branch `bicategory-grothendieck-functoriality` for an example of where these automatically generated declarations are used. I think it might also be useful for 2-yoneda, as the pseudofunctors there are valued in `Cat` also.

Much of the code has been copied from `reassoc`. However, a new definition `toCatExpr` is added to specialize the lemma so that the equality is taking place in `Cat`. One also needs some amount of extra preprocessing to properly deal with universes (since `Cat` has a different dependence on universes than an arbitrary bicategory).

---


dagurtomas:
This seems very useful. I would review it if I knew anything about metaprogramming. Maybe @adamtopaz can take a look?

Mathlib/Tactic/CategoryTheory/ToApp.lean
```diff
@@ -0,0 +1,152 @@
...
+initialize registerBuiltinAttribute {
+  name := `to_app
+  descr := ""
+  applicationTime := .afterCompilation
+  add := fun src ref kind => match ref with
+  | `(attr| to_app $[(attr := $stx?,*)]?) => MetaM.run' do
+    if (kind != AttributeKind.global) then
+      throwError "`to_app` can only be used as a global attribute"
+    addRelatedDecl src "_app" ref stx? fun type value levels => do
+      let levelMVars ← levels.mapM λ _ => mkFreshLevelMVar
```
yuma-mizuno, line 133: I think it's common to use `fun` instead of `λ`.

yuma-mizuno (commented):
I think this is a useful attribute to obtain `Cat` theorems!

I've left several comments.

callesonne:
Thanks for the review @yuma-mizuno! All the suggestions were great, I have implemented them now + improved the error catching of the function.

Mathlib/Tactic/CategoryTheory/ToApp.lean
```diff
@@ -0,0 +1,149 @@
...
+        | _ => throwError "The conclusion {conclusion} is not an equality of 2-morphisms!"
+      | _ => throwError "The conclusion {conclusion} is not an equality of 2-morphisms!"
+    | _ => throwError "The conclusion {conclusion} is not an equality!"
+  let some BIdx := args.findIdx? (· == B)
+    | throwError "Can not find the bicategory {B} in the arguments this declaration"
+  -- Create level metavariables to be used for `Cat.{v u}`
+  let u ← mkFreshLevelMVar
+  let v ← mkFreshLevelMVar
+  -- Assign `B` to `Cat.{v, u}`
+  let _ ← isDefEq B (.const ``Cat [u, v])
```
yuma-mizuno, line 67: I think `[v, u]` is a consistent notation here.

Mathlib/Tactic/CategoryTheory/ToApp.lean
```diff
@@ -0,0 +1,149 @@
...
+      return (← inferType x).getAppFnArgs == (`CategoryTheory.Bicategory, #[B])
+    | throwError "Can not find the argument for the bicategory instance of the bicategory in which \
+      the equality is taking place."
+  let _ ← isDefEq inst (.const ``CategoryTheory.Cat.bicategory [u, v])
+  -- Construct the new expression
+  let value := mkAppN e args
+  let mvars := (← getMVars value).map (Expr.mvar)
+  -- Erease the binderinfos for the bicategory and the instance
+  let binderInfos := binderInfos.eraseIdx BIdx
+  let binderInfos := binderInfos.eraseIdx BIdx
```
yuma-mizuno, line 79: It seems that we still need `Bicategory B` immediately after `B`. I attempted to remove this limitation by using `args` instead of `mvars`.
```lean
  let rec
  /-- Recursive function which applies `mkLambdaFVars` stepwise
  (so that each step can have different binderinfos) -/
    apprec (i : Nat) (e : Expr) : MetaM Expr := do
      if i < args.size then
        let arg := args[i]!
        let bi := binderInfos[i]!
        let e' ← apprec (i + 1) e
        unless arg != B && arg != inst do return e'
        mkLambdaFVars #[arg] e' (binderInfoForMVars := bi)
      else
        return e
```

Mathlib/Tactic/CategoryTheory/ToApp.lean
```diff
@@ -0,0 +1,149 @@
...
+
+This is useful for automatically generating lemmas that can be applied to expressions of 1-morphisms
+in `Cat` which contain components of 2-morphisms.
+
+There is also a term elaborator `to_app_of% t` for use within proofs.
+-/
+
+open Lean Meta Elab Tactic
+open Mathlib.Tactic
+open Qq
```
yuma-mizuno, line 30:
```suggestion
```

yuma-mizuno (commented):
lgtm after addressing additional comments.

Mathlib/Tactic/CategoryTheory/ToApp.lean
callesonne: Good catch, thanks a lot!

Mathlib/Tactic/CategoryTheory/ToApp.lean
callesonne: Right, thanks!

yuma-mizuno (approved):

semorrison:
Could you add a small test file? All this needs to do is import some file with a `to_app` attribute on some unlikely-to-change lemma, and verifies that the type of the generated lemma is what it is now.

semorrison:
Besides that, let's try it out.

bors d+

PR #16120, opened 2024-08-24
feat(ModelTheory/Algebra/Ring/Basic): Ring homomorphisms are a `StrongHomClass` for the language of rings

awainverse:
Adds an `IsAlgebraic` instance to the language of rings
Adds a `StrongHomClass` instance to the type of ring homomorphisms between rings with `CompatibleRing` structures

---

### DIFF
--- a/Mathlib/ModelTheory/Algebra/Ring/Basic.lean
+++ b/Mathlib/ModelTheory/Algebra/Ring/Basic.lean
@@ -236,29 +236,10 @@ def compatibleRingOfRing (R : Type*) [Add R] [Mul R] [Neg R] [One R] [Zero R] :
     funMap_zero := fun _ => rfl,
     funMap_one := fun _ => rfl }
 
-/-- An isomorphism in the language of rings is a ring isomorphism -/
-def languageEquivEquivRingEquiv {R S : Type*}
-    [NonAssocRing R] [NonAssocRing S]
-    [CompatibleRing R] [CompatibleRing S] :
-    (Language.ring.Equiv R S) ≃ (R ≃+* S) :=
-  { toFun := fun f =>
-    { f with
-      map_add' := by
-        intro x y
-        simpa using f.map_fun addFunc ![x, y]
-      map_mul' := by
-        intro x y
-        simpa using f.map_fun mulFunc ![x, y] }
-    invFun := fun f =>
-    { f with
-      map_fun' := fun {n} f => by
-        cases f <;> simp
-      map_rel' := fun {n} f => by cases f },
-    left_inv := fun f => by ext; rfl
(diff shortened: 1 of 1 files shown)


ChrisHughes24:
Looks okay to me based on my out of date knowledge of mathlib. I guess `languageEquivEquivRingEquiv` is obsolete now that we have `StrongHomClass`. Maybe it was already obsolete when it was added.

awainverse:
I completely missed that that `languageEquivEquivRingEquiv` was there, despite it being right above where I put this instance. I think it's kind of obsolete, but I should make sure it's easily replaced using this `StrongHomClass` instance.

awainverse:
I've made the proof of `languageEquivEquivRingEquiv` really simple with various homclass instances, but I'm never confident about determining whether I've created a bad instance loop, so somebody please check whether I should make some of these instances into `def`s or change priority.

awainverse:
Yeah - as I suspected, this loop is problematic. Which direction do we want to keep as an instance?

Mathlib/ModelTheory/Algebra/Ring/Basic.lean
```diff
@@ -64,6 +64,10 @@ namespace Ring
 
 open ringFunc Language
 
+instance : IsAlgebraic (Language.ring) := by
```
YaelDillies, line 67:
```suggestion
instance : IsAlgebraic Language.ring := by
```

Mathlib/ModelTheory/Algebra/Ring/Basic.lean
```diff
@@ -232,24 +236,41 @@ def compatibleRingOfRing (R : Type*) [Add R] [Mul R] [Neg R] [One R] [Zero R] :
     funMap_zero := fun _ => rfl,
     funMap_one := fun _ => rfl }
 
+instance _root_.RingHomClass.strongHomClass {R : Type*} [NonAssocRing R] [CompatibleRing R]
```
YaelDillies, line 239:
```suggestion
instance _root_.RingHomClass.toStrongHomClassRing {R : Type*} [NonAssocRing R] [CompatibleRing R]
```
or something like that

YaelDillies:
Do you want to implement [the two locales approach I alluded to](https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/.2316120.20Ring.20homomorphisms.20as.20a.20StrongHomClass/near/464883938)?

awainverse:
I think that'd be good - worst case, one locale will end up getting used more, and we'll choose that one later.

jcommelin:
This does seem to cause some slowdowns in TC synth. Maybe these could be defs, and locally you make them into instances?

joneugster:

PR #16121, opened 2024-08-24
[Merged by Bors] - fix: make `mkRichHCongr` reduce at default transparency

kmill:
Without this, `with_reducible congr!` could fail because `getFunInfoNArgs` always operates at default transparency, and the assumption is that `mkRichHCongr` will see the same arity as recorded in the resulting `FunInfo`. The fix is to use `whnfD` instead of `whnf` when creating the double telescope.

---


joneugster:
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #16122, opened 2024-08-24
feat: tensoring with R^n, n finite.

kbuzzard:
We prove various results about tensoring over `R` with `Rⁿ` or `Mⁿ`, if `n` is a finite type and `M` is an `R`-module (here `R` is a commutative (semi)ring).

---

- [x] depends on: #16123

This PR (and in particular the naming of results) is heavily inspired by `Mathlib/LinearAlgebra/DirectSum/Finsupp.lean`, although the results in that file cannot always be used directly to prove the ones in this file. For example in the finsupp file it's proved that `(ι →₀ R) ⊗[R] (κ →₀ R) ≃ₗ[R] ι × κ →₀ R`, whereas we prove that `(ι → R) ⊗[R] (κ → R) ≃ₗ[R] ι × κ → R` if either `ι` or `κ` are finite (allowing the other to be infinite, so it's not covered by the Finsupp result). 

I am sad that we can't use `calc` to make the defs in this PR (which are all of the form "compose this isomorphism with this one (and then possibly with this one)") so I have written out the definitions in "calc style" to make them easier to understand. In particular they can certainly be golfed by removing all the intermediate types and replacing many explicit types with `_`, and I would regard this as a step backwards.


Ruben-VandeVelde (approved):
lgtm

Mathlib/LinearAlgebra/TensorProduct/Finite.lean
```diff
@@ -0,0 +1,170 @@
...
+* `TensorProduct.finiteTensorPiLid R N n m` : if `n` is a finite type, `Rⁿ ⊗[R] Nᵐ ≃ₗ[R] Nⁿˣᵐ`.
+* `TensorProduct.piTensorFiniteRid R M m n` : if `n` is a finite type, `Mᵐ ⊗[R] Rⁿ ≃ₗ[R] Mᵐˣⁿ`.
+
+Note that in the last two results, the type `m` is arbitrary and does not have to be finite.
+-/
+
+namespace TensorProduct
+
+variable (R : Type*) [CommSemiring R] (M : Type*) [AddCommMonoid M] [Module R M]
+  (N : Type*) [AddCommMonoid N] [Module R N] (ι : Type*) [DecidableEq ι] [Finite ι]
```
sgouezel, line 32:
```suggestion
  (N : Type*) [AddCommMonoid N] [Module R N] (ι : Type*) [Finite ι]
```
You don't need decidability for the statements to make sense, so better `open scoped Classical in` in the first two definitions.

erdOne:
I think the results should go in `Mathlib/LinearAlgebra/TensorProduct/Pi.lean` and probably be merged with the contents already inside?

kbuzzard:
Sorry, I've been travelling. I don't need this PR any more, this was from back when I was using a different definition of the "canonical" topology on a module over a topological ring. I've done as suggested, moving the results into LinearAlgebra.TensorProduct.Pi, but that is all done for `Fintype` which in the application which no longer exists, I didn't have (but could have had).

kbuzzard:
I don't need these results any more so I'm closing this for now.

PR #16123, opened 2024-08-24
[Merged by Bors] - feat: generalise LinearEquiv.curry to R-modules

kbuzzard:
Generalise statement of `LinearEquiv.curry` from `(ι × κ) → R ≃ₗ[R] ι → κ → R` to `(ι × κ) → M ≃ₗ[R] ι → κ → M` with `M` an `R`-module.

---

(I also remove some spurious variables.)

This PR changes the number of inputs to the declaration, as now both `R` and `M` need to be given. The fact that mathlib compiles without the change is evidence that this function is not currently used anywhere in mathlib. I need the more general version for a forthcoming PR #16122 .

### DIFF
--- a/Mathlib/Algebra/Module/Equiv/Basic.lean
+++ b/Mathlib/Algebra/Module/Equiv/Basic.lean
@@ -403,23 +403,23 @@ end Subsingleton
 
 section Uncurry
 
-variable [Semiring R] [Semiring R₂] [Semiring R₃]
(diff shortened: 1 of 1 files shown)


grunweg (approved):
Looks reasonable, thanks!
maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #16130, opened 2024-08-25
[Merged by Bors] - feat: add a new class `IsZeroOrProbabilityMeasure`

sgouezel:
The need for this class has shown up in PFR: when conditioning on a set, one gets a measure which is either zero or a probability measure, and many results that hold for probability measures also hold trivially for the zero measure.

---


RemyDegenne:
Looks good!
bors r+

PR #16131, opened 2024-08-25
[Merged by Bors] - feat: remove the assumption that the measure is a probability in the strong law of large numbers

sgouezel:
Indeed, there is an independence assumption around, which implies that either the functions we're considering are zero (and the result is trivial) or the measure is a probability.

---


RemyDegenne:
Thanks!
bors r+

PR #16136, opened 2024-08-25
[Merged by Bors] - chore: Make measurability of sets/functions in discrete spaces available with anonymous dot notation

YaelDillies:
From LeanAPAP

---


RemyDegenne:
Can you explain why that kind of name change is desirable? (or point to some zulip discussion)
I know that a similar change was done for `eventually_of_forall`, and I don't understand it. To me, that use of anonymous dot notation simply looks like code obfuscation.

YaelDillies:
I am doing a lot of discrete analysis in LeanAPAP and as such I need to invoke general measure theory lemmas in statements where all side conditions are trivially satisfied. It's much nicer to do `general_lemma .of_discrete .of_discrete .of_finite` rather than `general_lemma (measurableSet_discrete _) (measurable_discrete _) (integrable_of_finite _ _)`.

RemyDegenne:
That sounds sensible. Let's merge this.
bors r+

b-mehta:
Just wanted to add here that I've been doing discrete probability recently, and this exact change is one of the things making it feel really nice

PR #16139, opened 2024-08-25
[Merged by Bors] - feat: The counting measure is nonzero

YaelDillies:
... and other simple measure theory lemmas

From LeanAPAP

---


RemyDegenne:
Thanks
bors r+

PR #16141, opened 2024-08-25
[Merged by Bors] - feat(MorphismProperty/Presheaf): Add basic properties of relatively representable morphisms

callesonne:
This PR adds some basic properties of the morphism property "relatively representable".

Co-authored-by: Joël Riou <rioujoel@gmail.com>

---

- [x] depends on: #16140

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16140, opened 2024-08-25
[Merged by Bors] - feat(MorphismProperty/Presheaf): Define relatively representable morphisms

callesonne:
This PR defines relatively representable morphisms, and adds some basic API.

Co-authored-by: Joël Riou <rioujoel@gmail.com>

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16142, opened 2024-08-25
[Merged by Bors] - feat(MorphismProperty/Presheaf): Add properties of the `relative` morphism property

callesonne:
This PR adds some basic properties about the "presheaf" morphism property. For example, that it is stable under base change.

Co-authored-by: Joël Riou <rioujoel@gmail.com>

---

- [x] depends on: #16140
- [x] depends on: #16141
- [x] depends on: #16143


jcommelin (commented):
Thanks :tada:

bors merge

PR #16146, opened 2024-08-25
[Merged by Bors] - feat(Measure/Typeclasses): prove `≪` both ways

urkud:
- Rename `exists_absolutelyContinuous_isFiniteMeasure` to `exists_isFiniteMeasure_absolutelyContinuous` to match the statement.
- Prove `ν ≪ μ`, not only `μ ≪ ν`.
- Turn the old lemma into a deprecated alias.

---

### DIFF
--- a/Mathlib/MeasureTheory/Measure/Typeclasses.lean
+++ b/Mathlib/MeasureTheory/Measure/Typeclasses.lean
@@ -536,22 +536,24 @@ instance [SFinite μ] (s : Set α) : SFinite (μ.restrict s) :=
     by rw [← restrict_sum_of_countable, sum_sFiniteSeq]⟩
 
 variable (μ) in
-/-- For An s-finite measure, there exists a finite measure with the same null sets. -/
-theorem exists_isFiniteMeasure_null_iff [SFinite μ] :
-    ∃ ν : Measure α, IsFiniteMeasure ν ∧ ∀ s, ν s = 0 ↔ μ s = 0 := by
+/-- For an s-finite measure `μ`, there exists a finite measure `ν`
+such that each of `μ` and `ν` is absolutely continuous with respect to the other.
+-/
+theorem exists_isFiniteMeasure_absolutelyContinuous [SFinite μ] :
+    ∃ ν : Measure α, IsFiniteMeasure ν ∧ μ ≪ ν ∧ ν ≪ μ := by
   rcases ENNReal.exists_pos_tsum_mul_lt_of_countable top_ne_zero (sFiniteSeq μ · univ)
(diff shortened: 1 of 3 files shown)


Mathlib/MeasureTheory/Measure/Typeclasses.lean
```diff
@@ -536,17 +536,22 @@ instance [SFinite μ] (s : Set α) : SFinite (μ.restrict s) :=
...
-    have : ∀ n, (sFiniteSeq μ n) s = 0 := by simpa [hsm, (hc₀ _).ne'] using hs
-    rw [← sum_sFiniteSeq μ, sum_apply _ hsm]
-    simp [this]
+  refine ⟨.sum fun n ↦ c n • sFiniteSeq μ n, ⟨by simpa [mul_comm] using hc⟩, fun s ↦ ?_⟩
+  conv_rhs => rw [← sum_sFiniteSeq μ, sum_apply_of_countable]
+  simp [(hc₀ _).ne']
+
+variable (μ) in
+/-- An s-finite measure is absolutely continuous with respect to some finite measure. -/
+@[deprecated exists_isFiniteMeasure_null_iff (since := "2024-08-25")]
```
sgouezel, line 550: I don't think we should deprecate this one: often, it is the one you need for the applications.

Mathlib/MeasureTheory/Measure/Typeclasses.lean
```diff
@@ -536,17 +536,22 @@ instance [SFinite μ] (s : Set α) : SFinite (μ.restrict s) :=
     by rw [← restrict_sum_of_countable, sum_sFiniteSeq]⟩
 
 variable (μ) in
-/-- An s-finite measure is absolutely continuous with respect to some finite measure. -/
-theorem exists_absolutelyContinuous_isFiniteMeasure [SFinite μ] :
-    ∃ ν : Measure α, IsFiniteMeasure ν ∧ μ ≪ ν := by
+/-- For An s-finite measure, there exists a finite measure with the same null sets. -/
+theorem exists_isFiniteMeasure_null_iff [SFinite μ] :
```
sgouezel, line 540: The name is a little bit misleading (I expect that a lemma ending with `iff` is an equivalence). Maybe `exists_isFiniteMeasure_null_iff_null` to lift the ambiguity?

Mathlib/MeasureTheory/Measure/Typeclasses.lean
urkud: What do you think about the new version?

sgouezel:
bors r+
Thanks!

PR #16161, opened 2024-08-26
[Merged by Bors] - chore(SetTheory/Ordinal/*): `o.out.α` → `o.toType`

vihdzp:
It's often desirable to get a concrete type in universe `Type u` that's order isomorphic to `o : Ordinal.{u}`. The way we've done this is by using the non-documented `(Quotient.out o).α`. We rename this to `Ordinal.toType o` for discoverability, and to hide what frankly should remain an implementation detail.

This also entails renaming a bunch of theorems that made reference to `out` in some form.

---

This change reveals a lot of `typein ((· < ·) : o.toType → o.toType → Prop)` in the code. A future refactor will replace this by the much more convenient `enumIsoToType`.


Mathlib/SetTheory/Ordinal/Basic.lean
```diff
@@ -987,51 +1013,57 @@ theorem enum_le_enum (r : α → α → Prop) [IsWellOrder α r] {o₁ o₂ : {o
...
 
 theorem enum_inj {r : α → α → Prop} [IsWellOrder α r] {o₁ o₂ : {o // o < type r}} :
     enum r o₁ = enum r o₂ ↔ o₁ = o₂ := by
   rw [EmbeddingLike.apply_eq_iff_eq, Subtype.mk.injEq]
 
-/-- The order isomorphism between ordinals less than `o` and `o.out.α`. -/
+/-- The order isomorphism between ordinals less than `o` and `o.toType`. -/
 @[simps!]
-noncomputable def enumIsoOut (o : Ordinal) : Set.Iio o ≃o o.out.α where
+noncomputable def enumIsoToType (o : Ordinal) : Set.Iio o ≃o o.toType where
```
vihdzp, line 1041: If the plan is to use this correspondence rather than `typein` and `enum` moving forward, I think we should come up for a better name for it. What about `IioIsoToType`? Or even just `IsoToType`?

YaelDillies:
Why do we even have `o.toType`? Why don't we just use `Iio o` everywhere?

vihdzp:
> Why do we even have `o.toType`? Why don't we just use `Iio o` everywhere?

There are some definitions that outright don't work without it, like `Ordinal.toPGame`, `PGame.nim`, or the future `Ordinal.toZFSet`. They all require an indexing type that lives within the same universe `u`. 

That said, `o.toType` is certainly overused in the current library, which is something I plan to address in a future refactor as I mention in my comment about `typein`.

YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/MeasureTheory/MeasurableSpace/Card.lean
```diff
@@ -31,8 +31,7 @@ variable {α : Type u}
 
 open Cardinal Set
 
--- Porting note: fix universe below, not here
```
b-mehta, line 34: This porting note still applies, correct?

b-mehta (approved):

Mathlib/MeasureTheory/MeasurableSpace/Card.lean
vihdzp: Yes, the universe has to be specified separately from the notation.

YaelDillies: So shouldn't you keep the porting note?

vihdzp: Ah, my bad. Must have deleted it out of instinct.

b-mehta (approved):
bors merge

PR #16162, opened 2024-08-26
[Merged by Bors] - feat(Dynamics.TopologicalEntropy): add definition of topological entropy by nets

D-Thomine:
Third PR in a sequence to implement the notion of topological entropy for maps using Bowen-Dinaburg's formalism. This file introduces a definition of topological entropy using nets.

- [x] Dynamical uniformities
- [x] Topological entropy via covers
- [x] Topological entropy via nets
- [ ] Behaviour under semiconjugacy
- [ ] Behaviour for subsets and unions
- [ ] Behaviour under products
- [ ] Full shift

The main file is `Dynamics.TopologicalEntropy.NetEntropy`. It is slightly shorter and less technical than the previous one. There are also a few modifications to `Dynamics.TopologicalEntropy.CoverEntropy` (some golf, and renamed a lemma to enable dot notation -- which is quite convenient here).

Paging @pitmonticone @sgouezel

---


Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+
+We relate in this file `CoverEntropy` and `NetEntropy`. This file is downstream of
+`Mathlib.Dynamics.TopologicalEntropy.CoverEntropy` since the submultiplicative argument there
+(specifically `IsDynCoverOf.iterate_le_pow`) is more natural for covers.
+
+## Main definitions
+- `IsDynNetOf`: property that dynamical balls centered on a subset `s` of `F` are disjoint.
+- `netMaxcard`: maximal cardinal of a dynamical net. Takes values in `ℕ∞`.
+- `netEntropyInfEnt`/`netEntropyEnt`: exponential growth of `netMaxcard`. The former is
+defined with a `liminf`, the later with a `limsup`. Take values in `EReal`.
```
sgouezel, line 35:
```suggestion
defined with a `liminf`, the latter with a `limsup`. Take values in `EReal`.
```

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Damien Thomine, Pietro Monticone
+-/
+import Mathlib.Dynamics.TopologicalEntropy.CoverEntropy
+
+/-!
+# Topological entropy via nets
+We implement Bowen-Dinaburg's definitions of the topological entropy, via nets.
+
+All is stated in the vocabulary of uniform spaces. For compact spaces, the uniform structure
```
sgouezel, line 12: I don't think it's worth having the same text as in `CoverEntropy`. Maybe just give a pointer? What would be more useful is describing what the definition via nets is, for people who have never seen topological entropy.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+
+open Set Uniformity UniformSpace
+
+variable {X : Type*}
+
+/-! ### Dynamical nets -/
+
+/-- Given a subset `F`, an entourage `U` and an integer `n`, a subset `s` of `F` is a
+`(U, n)`-dynamical net of `F` if no two orbits of length `n` of points in `s` shadow each other.-/
+def IsDynNetOf (T : X → X) (F : Set X) (U : Set (X × X)) (n : ℕ) (s : Set X) : Prop :=
```
sgouezel, line 65: The `Of` seems weird: for the cover, you have a cover of `s`, so it feels right, but there the point is more that you're inside `s`, so `IsDynNetIn` would seem more natural.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+lemma isDynNetOf_empty {T : X → X} {F : Set X} {U : Set (X × X)} {n : ℕ} :
+    IsDynNetOf T F U n ∅ :=
+  ⟨empty_subset F, pairwise_empty _⟩
+
+lemma isDynNetOf_singleton (T : X → X) {F : Set X} (U : Set (X × X)) (n : ℕ) {x : X} (h : x ∈ F) :
+    IsDynNetOf T F U n {x} :=
+  ⟨singleton_subset_iff.2 h, pairwise_singleton x _⟩
+
+/-- Given an entourage `U` and a time `n`, a dynamical net has a smaller cardinal than a dynamical
+  cover. This lemma is the first of two key results to compare two versions topological entropy:
```
sgouezel, line 87:
```suggestion
  cover. This lemma is the first of two key results to compare two versions of  topological entropy:
```

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+    specialize ht (hs.1 x_s)
+    simp only [Finset.coe_sort_coe, mem_iUnion, Subtype.exists, exists_prop] at ht
+    exact ht
+  choose! F s_t using this
+  simp only [mem_ball_symmetry (U_symm.dynEntourage T n)] at s_t
+  apply Finset.card_le_card_of_injOn F (fun x x_s ↦ (s_t x x_s).1)
+  exact fun x x_s y y_s Fx_Fy ↦
+    PairwiseDisjoint.elim_set hs.2 x_s y_s (F x) (s_t x x_s).2 (Fx_Fy ▸ (s_t y y_s).2)
+
+/-! ### Maximal cardinal of dynamical nets -/
```
sgouezel, line 103:
```suggestion
/-! ### Maximal cardinality of dynamical nets -/
```
here and below

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+  exact (one_le_netMaxcard_iff T F U n).2 h
+
+/-! ### Net entropy of entourages -/
+
+open Filter
+
+/-- The entropy of an entourage `U`, defined as the exponential rate of growth of the size of the
+largest `(U, n)`-dynamical net of `F`. Takes values in the space of extended real numbers
+`[-∞,+∞]`. This version uses a `limsup`, and is chosen as the default definition.-/
+noncomputable def netEntropyEnt (T : X → X) (F : Set X) (U : Set (X × X)) :=
```
sgouezel, line 273: I'm just realizing that the `Ent` for entourage collides with the `Ent` for entropy. Given this, you should use longer names, like `netEntropyEntour` or `netEntropyEntourage`. And do the same in the other file...

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+- `netMaxcard`: maximal cardinal of a dynamical net. Takes values in `ℕ∞`.
+- `netEntropyInfEnt`/`netEntropyEnt`: exponential growth of `netMaxcard`. The former is
+defined with a `liminf`, the later with a `limsup`. Take values in `EReal`.
+
+## Implementation notes
+As when using covers, there are two competing definitions `netEntropyInfEnt` and
+`netEntropyEnt` in this file: one uses a `liminf`, the other a `limsup`. When using covers, we chose
+the `limsup` definition as the default. Because of lemmas `coverEntropyInf_eq_iSup_netEntropyInfEnt`
+and `coverEntropy_eq_iSup_netEntropyEnt`, we make the same choice here. Theorems about the
+topological entropy of invariant subsets will be stated using only `coverEntropy`.
```
sgouezel, line 42: Could you emphasize that you are *not* defining `netEntropy` and `netEntropyInf` (as the supremum over all entourages of ...) because it coincides with the cover version, so there's no point in having two names for the same object?

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+    coverEntropyEnt T F (U ○ U) ≤ netEntropyEnt T F U := by
+  refine (limsup_le_limsup) (Eventually.of_forall fun n ↦ ?_)
+  apply div_le_div_right_of_nonneg (Nat.cast_nonneg' n) (log_monotone _)
+  exact ENat.toENNReal_le.2 (coverMincard_le_netMaxcard T F U_rfl U_symm n)
+
+/-! ### Relationship with entropy via covers -/
+
+variable [UniformSpace X] (T : X → X) (F : Set X)
+
+lemma coverEntropyInf_eq_iSup_netEntropyInfEnt :
```
sgouezel, line 357: This one and the next one would deserve to be called theorems, and would deserve longer docstrings.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+noncomputable def netEntropyInfEnt (T : X → X) (F : Set X) (U : Set (X × X)) :=
+  atTop.liminf fun n : ℕ ↦ log (netMaxcard T F U n) / n
+
+lemma netEntropyInfEnt_antitone (T : X → X) (F : Set X) :
+    Antitone (fun U : Set (X × X) ↦ netEntropyInfEnt T F U) :=
+  fun _ _ U_V ↦ (liminf_le_liminf) (Eventually.of_forall
+    fun n ↦ monotone_div_right_of_nonneg (Nat.cast_nonneg' n)
+      (log_monotone (ENat.toENNReal_mono (netMaxcard_antitone T F n U_V))))
+
+
```
sgouezel, line 288:
```suggestion
```

sgouezel (commented):
Nice!

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
pitmonticone: There was "cardinal" in other places too, but now they're all fixed. Thanks!

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
pitmonticone: Done, thanks.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
D-Thomine: Done in both files. I've chosen `netEntropyEntourage`, because there is not much to gain by deleting 3 letters of an already long name.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
D-Thomine: Nice point. Done.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
D-Thomine: Done.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
D-Thomine: Done.

Mathlib/Dynamics/TopologicalEntropy/NetEntropy.lean
```diff
@@ -0,0 +1,408 @@
...
+open Set Uniformity UniformSpace
+
+variable {X : Type*}
+
+/-! ### Dynamical nets -/
+
+/-- Given a subset `F`, an entourage `U` and an integer `n`, a subset `s` of `F` is a
+`(U, n)`-dynamical net of `F` if no two orbits of length `n` of points in `s` shadow each other.-/
+def IsDynNetIn (T : X → X) (F : Set X) (U : Set (X × X)) (n : ℕ) (s : Set X) : Prop :=
+    s ⊆ F ∧ s.PairwiseDisjoint (fun x : X ↦ ball x (dynEntourage T U n))
```
sgouezel, line 62:
```suggestion
  s ⊆ F ∧ s.PairwiseDisjoint (fun x : X ↦ ball x (dynEntourage T U n))
```

sgouezel (commented):
bors d+
Thanks!

PR #16163, opened 2024-08-26
chore: adaptations for nightly-2024-08-26

jcommelin:


Mathlib/Algebra/Homology/ExactSequence.lean
```diff
@@ -159,7 +159,7 @@ lemma exact_iff_of_iso {S₁ S₂ : ComposableArrows C n} (e : S₁ ≅ S₂) :
 lemma exact₀ (S : ComposableArrows C 0) : S.Exact where
   toIsComplex := S.isComplex₀
   -- See https://github.com/leanprover/lean4/issues/2862
```
semorrison, line 161:
```suggestion
```

Mathlib/Combinatorics/Enumerative/DyckWord.lean
```diff
@@ -259,11 +259,18 @@ include h in
...
   by_contra! f
   rw [Nat.le_zero, firstReturn, findIdx_eq] at f
-  · simp only [get_eq_getElem, getElem_range] at f
-    rw [← p.cons_tail_dropLast_concat h] at f
-    simp at f
+  #adaptation_note
+  /--
+  If we don't swap, then the second goal is dropped after completing the first goal.
+  What's going on?
+  -/
```
jcommelin, line 266: Is this related to the membership mem swap that is now fixed again? So this will be very transient?
If not, shall we add a date string?

Mathlib/Tactic/ToAdditive/Frontend.lean
```diff
@@ -818,9 +818,10 @@ partial def transformDeclAux
   if isProtected (← getEnv) src then
     setEnv <| addProtected (← getEnv) tgt
   if let some matcherInfo ← getMatcherInfo? src then
-    -- Use
-    --   Match.addMatcherInfo tgt matcherInfo
-    -- once on lean 4.13.
+    /-
```
jcommelin, line 821:
```suggestion
    #adaptation_note
    /--
```
Or should we use some other marker, so that it will be recorded as tech debt?

Mathlib/Combinatorics/Enumerative/DyckWord.lean
semorrison: I'm not sure, unfortunately.

Mathlib/Tactic/ToAdditive/Frontend.lean
```diff
@@ -818,9 +818,11 @@ partial def transformDeclAux
...
     setEnv <| addProtected (← getEnv) tgt
   if let some matcherInfo ← getMatcherInfo? src then
-    -- Use
-    --   Match.addMatcherInfo tgt matcherInfo
-    -- once on lean 4.13.
+    #adaptation_note
+    /--
+    Use `Match.addMatcherInfo tgt matcherInfo`
+    once https://github.com/leanprover/lean4/pull/5068 is in
+    -/
```
semorrison: @jcommelin, we can't use #adaptation note in term mode.

Mathlib/Tactic/ToAdditive/Frontend.lean
semorrison: We can't used adaptation note in term mode.

PR #16164, opened 2024-08-26
[Merged by Bors] - feat (RingTheory/HahnSeries) : coefficient-wise map

ScottCarnahan:
This PR introduces `HahnSeries.map` which applies a zero-preserving map to each coefficient. We more or less copy the `Polynomial.map` API.

---

### DIFF
--- a/Mathlib/RingTheory/HahnSeries/Basic.lean
+++ b/Mathlib/RingTheory/HahnSeries/Basic.lean
@@ -25,6 +25,8 @@ in the file `RingTheory/LaurentSeries`.
   coefficient if `x ≠ 0`, and is `⊤` when `x = 0`.
 * `order x` is a minimal element of `Γ` where `x` has a nonzero coefficient if `x ≠ 0`, and is zero
   when `x = 0`.
+* `map` takes each coefficient of a Hahn series to its target under a zero-preserving map.
+* `embDomain` preserves coefficients, but embeds the index set `Γ` in a larger poset.
 
 ## References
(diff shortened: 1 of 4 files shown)


Mathlib/RingTheory/HahnSeries/Basic.lean
```diff
@@ -108,8 +108,19 @@ nonrec theorem support_nonempty_iff {x : HahnSeries Γ R} : x.support.Nonempty 
 theorem support_eq_empty_iff {x : HahnSeries Γ R} : x.support = ∅ ↔ x = 0 :=
   Function.support_eq_empty_iff.trans coeff_fun_eq_zero_iff
 
+/-- The map of Hahn series induced by applying a zero-preserving map to each coefficient. -/
+@[simps]
+def map [Zero S] (x : HahnSeries Γ R) (f : ZeroHom R S) : HahnSeries Γ S where
```
awainverse, line 113: Should this allow for `f` to come from an arbitrary `ZeroHomClass`? I see that it works fine for monoid and ring homs elsewhere, presumably using a coercion, so I'm not sure which implementation is better.

It's probably worth mentioning in the module doc that this maps coefficients, where `HahnSeries.embDomain` maps indices.

ScottCarnahan: Thank you, that had not occurred to me.

jcommelin (commented):
Sorry for the massive delay in reviewing! This looks absolutely fine.

bors merge

PR #16167, opened 2024-08-26
[Merged by Bors] - feat(CategoryTheory/Adjunction): left adjoint in adjoint triple is fully faithful iff right adjoint is

dagurtomas:
---


Mathlib/CategoryTheory/Adjunction/Triple.lean
```diff
@@ -0,0 +1,55 @@
...
+    let adjId : F ⋙ G ⊣ 𝟭 C := adj.ofNatIsoRight (asIso adj₂.counit)
+    exact adj₁.isIso_unit_of_iso (adjId.leftAdjointUniq id)
+
+/--
+Given an adjoint triple `F ⊣ G ⊣ H`, the left adjoint `F` is fully faithful if and only if the
+right adjoint `H` is fully faithful.
+-/
+noncomputable def fullyFaithfulEquiv : F.FullyFaithful ≃ H.FullyFaithful where
+  toFun h :=
+    haveI := h.full
```
b-mehta, line 39: Does it work to say `haveI := h` instead? I'd hope we have the instance from FullyFaithful to Full. And if so, does it work to omit the `haveI` entirely?

dagurtomas: That doesn't work, and an instance from `FullyFaithful` to `Full` and to `Faithful` wouldn't work either, because `FullyFaithful` is not a class

dagurtomas: When `Full` was made a `Prop`, the `FullyFaithful` structure was added, which contains the data of a preimage to every morphism in the image of the functor

joelriou:
Thanks!

bors merge

joelriou:
Retrying...

bors merge

PR #16172, opened 2024-08-27
[Merged by Bors] - chore: remove `autoImplicit` from `Logic.Equiv.Basic`

Parcly-Taxel:
Split from #16154.

---


TwoFX (changes requested):
Your changes specialize `Equiv.subtypePreimage` from `Sort` to `Type`. It is possible that there are more declarations where this is the case. Please go over all declarations in the file and make sure the statements didn't change.

TwoFX (approved):
bors r+

PR #16175, opened 2024-08-27
[Merged by Bors] - feat: turn off bad simp lemmas

kim-em:
Removes some bad `simp` lemmas as identified by @mattrobball at https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Infrastructure.20for.20tracking.20frequently.20applied.20simp.20theorems/near/459926416

### DIFF
--- a/Mathlib/CategoryTheory/Limits/Shapes/Biproducts.lean
+++ b/Mathlib/CategoryTheory/Limits/Shapes/Biproducts.lean
@@ -640,6 +640,7 @@ lemma biproduct.whiskerEquiv_inv_eq_lift {f : J → C} {g : K → C} (e : J ≃ 
     · rintro rfl
       simp at h
 
+attribute [local simp] Sigma.forall in
 instance {ι} (f : ι → Type*) (g : (i : ι) → (f i) → C)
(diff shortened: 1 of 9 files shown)


TwoFX (approved):
bors r+

Mathlib/CategoryTheory/Limits/Shapes/Biproducts.lean
```diff
@@ -640,6 +640,7 @@ lemma biproduct.whiskerEquiv_inv_eq_lift {f : J → C} {g : K → C} (e : J ≃
     · rintro rfl
       simp at h
 
+attribute [local simp] Sigma.forall in
```
mattrobball, line 643: Can this be inlined into the `simp` calls in the proof?

semorrison: I was too slow, but: this was used by an autoparam `aesop` call.

mattrobball: I guess I didn't read carefully

PR #16178, opened 2024-08-27
[Merged by Bors] - feat: generalize universes in `preservesFiniteProductsOfPreservesBinaryAndTerminal`

callesonne:
---
I noticed some theorems that hold with more general universe parameters than given. Since this is just a generalization of a lemma, I assume there is no harm in doing this?


dagurtomas:
In this direction, this is an obvious improvement. Thanks!

maintainer merge

PR #16179, opened 2024-08-27
chore: add the stacks attribute

adomani:
This PR contains a subset of the code of #16191 and is probably not going to be merged.

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/stacks.20tags)

---

### DIFF
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -1,7 +1,7 @@
 /-
-Copyright (c) 2024 Dmaiano Testa. All rights reserved.
+Copyright (c) 2024 Damiano Testa. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Dmaiano Testa
+Authors: Damiano Testa
 -/
 import Lean.Parser.Syntax
 
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -37,7 +37,8 @@ initialize Lean.registerBuiltinAttribute {
   name := `stacks
   descr := "Apply a Stacks project tag to a theorem."
   add := fun _decl stx _attrKind => Lean.withRef stx do
-    -- check that there are no spaces in the tag
+    -- check that the tag consists of 4 characters and
+    -- that only digits and uppercase letter are present
     let tag := stx[1]
     if let some s := tag.getSubstring? then
       let str := s.toString.trimRight
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -1,0 +1,41 @@
+/-
+Copyright (c) 2024 Dmaiano Testa. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dmaiano Testa
+-/
+import Lean.Parser.Syntax
+
+/-!
+The `stacks` attribute.
+-/
+
+open Lean
+
+namespace Mathlib.Stacks
+
+/--
+The syntax for a Stacks tag: it is an optional number followed by an optional identifier.
+This allows `044Q3` and `GH3F6` as possibilities.
+-/
+declare_syntax_cat stackTags
+
+@[inherit_doc Parser.Category.stackTags]
+syntax (num)? (ident)? : stackTags
+
+/-- The `stacks` attribute.
+Use it as `@[stacks TAG "Optional comment"]`.
+The `TAG` is mandatory.
+-/
+syntax (name := stacks) "stacks " (stackTags)? (str)? : attr
+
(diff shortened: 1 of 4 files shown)


Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,39 @@
+/-
+Copyright (c) 2024 Dmaiano Testa. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dmaiano Testa
```
bryangingechen, line 4:
```suggestion
Authors: Damiano Testa
```

adomani: :facepalm: thanks!

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,39 @@
...
+The `TAG` is mandatory.
+-/
+syntax (name := stacks) "stacks " (stackTags)? (str)? : attr
+
+initialize Lean.registerBuiltinAttribute {
+  name := `stacks
+  descr := "Apply a Stacks project tag to a theorem."
+  add := fun _decl stx _attrKind => Lean.withRef stx do
+    match stx with
+      | `(attr| stacks $_:stackTags $_:str) => return
```
mattrobball, line 36: Can we validate the `stackTag` here?

adomani: I just added a check that there are no spaces in the tag itself.  By "validate" you mean "make sure that such a tag exists"?  That should probably be done by a linter, right?

mattrobball: Sorry. What something like
```lean
syntax num | lit : stacksTagAtom
syntax stacksAtom stacksAtom stacksAtom stacksAtom stacksAtom : stacksTag
```
Is there some uppercase Latin character syntax category? Do we really have to write out `stacksAtom` 5 times?

adomani: Oh, the length 5 check can be done with the substring and also the uppercase check.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,44 @@
...
+The `stacks` attribute.
+-/
+
+open Lean
+
+namespace Mathlib.Stacks
+
+/--
+The syntax for a Stacks tag: it is an optional number followed by an optional identifier.
+This allows `044Q3` and `GH3F6` as possibilities.
```
bryangingechen, line 18: Maybe worth linking to the Stacks project page on tags: https://stacks.math.columbia.edu/tags

mattrobball:
For the environment extension, do we want to store anything that map from the declaration name to the stacks tag as a string?

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,44 @@
...
+-/
+declare_syntax_cat stackTags
+
+@[inherit_doc Parser.Category.stackTags]
+syntax (num)? (ident)? : stackTags
```
mattrobball, line 23:
```suggestion
syntax (num)? (ident)? : stackTag
```

mattrobball: This is a single tag right?

Mathlib/Tactic/StacksAttribute.lean
adomani: I have a slight preference for not using the double quotes, which a `str` would want.  If there is a way of parsing a string without quotes, then I would probably prefer that!

Mathlib/Tactic/StacksAttribute.lean
adomani: Added, thanks!

adomani:
> For the environment extension, do we want to store anything more than a map from the declaration name to the stacks tag as a string?

For the moment, I think probably not.  Maybe in the future we may want to add something else.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,53 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Damiano Testa
+-/
+import Lean.Attributes
+
+/-!
+The `stacks` attribute.
+
+This allows tagging of mathlib lemmas with the corresponding
+[Tags](stacks.math.columbia.edu/tags) from the Stacks Project.
```
bryangingechen, line 12: I'm not sure if doc-gen4 is smart enough to link to the right place without https, but probably safest to include it anyways (also below):
```suggestion
[Tags](https://stacks.math.columbia.edu/tags) from the Stacks Project.
```

adomani:
The "other" PR is getting merged, so I'll close this one.

Thank you all for your comments: they are part of the other PR anyway!

PR #15820, opened 2024-08-14
[Merged by Bors] - refactor(SetTheory/Ordinal/Arithmetic): Ditch `Ordinal.sup`

vihdzp:
`Ordinal.sup` is just `iSup` with more restricted universes. This is the first part in a series of refactors aimed at putting the ordinal API in line with the API of other lattices.

Theorems like `le_iSup` were kept as protected theorems within the `Ordinal` namespace, as they can be useful when you do know that the input type lives in a smaller universe. That said, using them indiscriminately will make the universes in theorems stricter than they need to be.

I kept `sup` within the API for `bsup`, `lsub`, `blsub`, and `mex`, as I plan to refactor those individually in the near future.

---

Here's a list of other planned subsequent PRs:
- Write theorems for `sSup` when appropriate
- Remove `bsup` in favor of `iSup`
- Remove or perhaps rework `lsub` (that should probably be discussed beforehand)
- Generalize universes whenever possible (e.g. write `Small.{u} ι` instead of `ι : Type u`)

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16180, opened 2024-08-27
[Merged by Bors] - chore(SetTheory/Ordinal/Exponential): use `pow` instead of `opow` 

vihdzp:
The theorem `sup_opow_nat` gets renamed in #15820, so I won't rename it here.

---


jcommelin (commented):
Thanks :tada:

bors merge

vihdzp:
Wait a second, I don't think this currently builds! Let me fix that real quick.

vihdzp:
CI now passes, can someone please re-send this to bors?

PR #16181, opened 2024-08-27
[Merged by Bors] - feat: Conjugation-negation operator

YaelDillies:
This is a fundamental operator in discrete Fourier analysis. For example, it is the operator relating convolution and difference convolution: https://yaeldillies.github.io/LeanAPAP/docs/LeanAPAP/Prereqs/Convolution/Discrete/Defs.html#conv_conjneg

From LeanAPAP

---


eric-wieser:
Can you link to an example of where this is useful in the PR description?

TFBloom:
> Can you link to an example of where this is useful in the PR description?

One reason this operator comes up a lot is because the conj (Fourier transform of f)  = Fourier transform of (conjneg f). So whenever you are using both Fourier analysis and the conjugation operator conjneg is around.

For example, the absolute value of (Fourier transform of f)^2 is equal to the Fourier transform of f * conjneg f.

Mathlib/Algebra/Star/Conjneg.lean
```diff
@@ -0,0 +1,75 @@
...
+Authors: Yaël Dillies
+-/
+import Mathlib.Algebra.BigOperators.Pi
+import Mathlib.Algebra.Star.Pi
+import Mathlib.Data.Set.Pointwise.Basic
+
+/-!
+# Conjugation-negation operator
+
+This file defines the conjugation-negation operator, useful in Fourier analysis.
```
b-mehta, line 13: Could be nice to additionally say here the way in which it's useful. eg, that convolution with conjneg f is adjoint to convolution with f

eric-wieser:
Sorry, what I meant is not "when does this come up mathematically", but more "why not write `fun x ↦ conj (f (-x))`"? Would having notation for `star` make this more bearable?

YaelDillies:
`conj (f <| -∙)` is invalid syntax (gives `expected token` at the `·`). `fun x ↦ conj (f (-x))` is pretty verbose (and notation for `conj` (as a reminder this is already notation!) won't fix it) and is inconvenient for stating lemmas like `conjneg (f + g) = conjneg f + conjneg g`, `conjneg (f ∗ g) = conjneg f ∗ conjneg g` or `f ∗ conjneg g = f ○ g` (where `∗` is convolution and `○` difference convolution).

There are two particular points of concern for me:
* There are at least two different spellings of `conjneg`, namely `fun x ↦ conj (f (-x))` and `fun x ↦ conj f (-x)`.
* `conjneg f` is used unapplied. `conjneg f x` never appears in LeanAPAP.

Having a single named function solves both issues.

Mathlib/Algebra/Star/Conjneg.lean
```diff
@@ -0,0 +1,78 @@
...
+open Function
+open scoped ComplexConjugate
+
+variable {ι G R : Type*} [AddGroup G]
+
+section CommSemiring
+variable [CommSemiring R] [StarRing R] {f g : G → R}
+
+/-- Conjugation-negation. Sends `f` to `fun x ↦ conj (f (-x))`. -/
+def conjneg (f : G → R) : G → R := conj fun x ↦ f (-x)
```
jcommelin, line 28: Does it make sense to make this a bundled ring hom?

YaelDillies: I don't like making functions bundled by default because it makes them always be prepended with `⇑`, have their API dispersed across the general hom APIs and their own, and is generally little to not useful. I have instead added a bundled version `conjnegRingHom`

Mathlib/Algebra/Star/Conjneg.lean
```diff
@@ -0,0 +1,87 @@
...
+@[simp] lemma support_conjneg (f : G → R) : support (conjneg f) = -support f := by
+  ext; simp [starRingEnd_apply]
+
+/-- `conjneg` bundled as a ring homomorphism. -/
+@[simps] def conjnegRingHom : (G → R) →+* (G → R) where
+  toFun := conjneg
+  map_zero' := conjneg_zero
+  map_one' := conjneg_one
+  map_add' := conjneg_add
+  map_mul' := conjneg_mul
```
jcommelin, line 77: If you move this to line 47, then you can very slightly golf the proofs of `conjneg_{sum,prod}`.

jcommelin (commented):
bors d+

Mathlib/Algebra/Star/Conjneg.lean
YaelDillies: I've golfed them even further (and that doesn't even use `conjnegRingHom` 😁)

YaelDillies:
bors merge

Mathlib/Algebra/Star/Conjneg.lean
```diff
@@ -0,0 +1,84 @@
...
+
+@[simp] lemma conjneg_zero : conjneg (0 : G → R) = 0 := by ext; simp
+@[simp] lemma conjneg_one : conjneg (1 : G → R) = 1 := by ext; simp
+@[simp] lemma conjneg_add (f g : G → R) : conjneg (f + g) = conjneg f + conjneg g := by ext; simp
+@[simp] lemma conjneg_mul (f g : G → R) : conjneg (f * g) = conjneg f * conjneg g := by ext; simp
+
+@[simp] lemma conjneg_sum (s : Finset ι) (f : ι → G → R) :
+    conjneg (∑ i ∈ s, f i) = ∑ i ∈ s, conjneg (f i) := by ext; simp
+
+@[simp] lemma conjneg_prod (s : Finset ι) (f : ι → G → R) :conjneg (∏ i ∈ s, f i) = ∏ i ∈ s, conjneg (f i) := by ext; simp
```
Parcly-Taxel, line 51:
```suggestion
@[simp] lemma conjneg_prod (s : Finset ι) (f : ι → G → R) :
    conjneg (∏ i ∈ s, f i) = ∏ i ∈ s, conjneg (f i) := by ext; simp
```

YaelDillies:
bors merge

PR #16183, opened 2024-08-27
[Merged by Bors] - feat: the translation operator

YaelDillies:
Define translation of a function by an element of its domain

From LeanAPAP

---

I don't really like this code, but it is foundational for discrete analysis and I have found no nice way around using it. I would like to know @urkud's opinion as I am aware he needs translation for dynamical systems. Here are a few points of concern I am aware of:
* There is no multiplicative version. That's because I do not need the multiplicative version and wouldn't know how to call it
* This is basically an action by `DomMulAct`. The trouble is that it would be inconvenient to state it this way because I would also need to stick some `op` in there. And generally this action is wide-spread in LeanAPAP and I want custom notation for it.

### DIFF
--- a/Mathlib/Algebra/Group/Translate.lean
+++ b/Mathlib/Algebra/Group/Translate.lean
@@ -13,9 +13,17 @@ import Mathlib.GroupTheory.GroupAction.DomAct.Basic
 
 This file defines the translation of a function from a group by an element of that group.
 
-# Notation
+## Notation
 
 `τ a f` is notation for `translate a f`.
+
+## See also
+
+Generally, translation is the same as acting on the domain by subtraction. This setting is
+abstracted by `DomAddAct` in such a way that `τ a f = DomAddAct.mk (-a) +ᵥ f` (see
+`translate_eq_domAddActMk_vadd`). Using `DomAddAct` is irritating in applications because of this
+negation appearing inside `DomAddAct.mk`. Although mathematically equivalent, the pen and paper
+convention is that translating is an action by subtraction, not by addition.
 -/
 
 open Function Set
--- a/Mathlib/Algebra/Group/Translate.lean
+++ b/Mathlib/Algebra/Group/Translate.lean
@@ -1,5 +1,5 @@
 /-
-Copyright (c) 2024 Yaël Dillies. All rights reserved.
+Copyright (c) 2023 Yaël Dillies. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Yaël Dillies
 -/
--- a/Mathlib/Algebra/Group/Translate.lean
+++ b/Mathlib/Algebra/Group/Translate.lean
@@ -1,0 +1,79 @@
+/-
+Copyright (c) 2024 Yaël Dillies. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Yaël Dillies
+-/
+import Mathlib.Algebra.BigOperators.Pi
+import Mathlib.Data.Set.Pointwise.SMul
+
+/-!
+# Translation operator
+
+This file defines the translation of a function from a group by an element of that group.
+
+# Notation
+
+`τ a f` is notation for `translate a f`.
+-/
+
+open Function Set
+open scoped Pointwise
+
+variable {ι α β M G H : Type*} [AddCommGroup G]
+
+/-- Translation of a function in a group by an element of that group.
+`τ a f` is defined as `x ↦ f (x - a)`.  -/
+def translate (a : G) (f : G → α) : G → α := fun x ↦ f (x - a)
+
+@[inherit_doc] scoped[translate] notation "τ " => translate
+
+open scoped translate
+
+@[simp] lemma translate_apply (a : G) (f : G → α) (x : G) : τ a f x = f (x - a) := rfl
+@[simp] lemma translate_zero (f : G → α) : τ 0 f = f := by ext; simp
+
+lemma translate_translate (a b : G) (f : G → α) : τ a (τ b f) = τ (a + b) f := by
+  ext; simp [sub_sub]
+
+lemma translate_add (a b : G) (f : G → α) : τ (a + b) f = τ a (τ b f) := by ext; simp [sub_sub]
+
+lemma translate_add' (a b : G) (f : G → α) : τ (a + b) f = τ b (τ a f) := by
+  rw [add_comm, translate_add]
+
+lemma translate_comm (a b : G) (f : G → α) : τ a (τ b f) = τ b (τ a f) := by
+  rw [← translate_add, translate_add']
+
+-- We make `simp` push the `τ` outside
+@[simp] lemma comp_translate (a : G) (f : G → α) (g : α → β) : g ∘ τ a f = τ a (g ∘ f) := rfl
+
+@[simp]
+lemma translate_smul_right [SMul H α] (a : G) (f : G → α) (c : H) : τ a (c • f) = c • τ a f := rfl
+
+@[simp] lemma translate_zero_right [Zero α] (a : G) : τ a (0 : G → α) = 0 := rfl
+lemma translate_add_right [Add α] (a : G) (f g : G → α) : τ a (f + g) = τ a f + τ a g := rfl
+lemma translate_sub_right [Sub α] (a : G) (f g : G → α) : τ a (f - g) = τ a f - τ a g := rfl
+lemma translate_neg_right [Neg α] (a : G) (f : G → α) : τ a (-f) = -τ a f := rfl
+
+section AddCommMonoid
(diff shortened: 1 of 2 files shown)


eric-wieser:
Just to confirm, the `DomAddAct` spelling is `DomAddAct.mk (-a) +ᵥ f`

YaelDillies:
Precisely. And you see that I can't make `τ a f` notation for `DomAddAct.mk (-a) +ᵥ f` because then `τ (-a) f` would get "simplified" to `DomAddAct.mk a +ᵥ f`.

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,78 @@
...
+open Function Set
+open scoped Pointwise
+
+variable {ι α β M G H R : Type*} [AddCommGroup G]
+
+/-- Translation of a function in a group by an element of that group.
+`τ a f` is defined as `x ↦ f (x - a)`.  -/
+def translate (a : G) (f : G → α) : G → α := fun x ↦ f (x - a)
+
+@[inherit_doc] notation "τ " => translate
```
b-mehta, line 28: This probably shouldn't be global notation

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,78 @@
...
+
+lemma translate_add (a b : G) (f : G → α) : τ (a + b) f = τ a (τ b f) := by ext; simp [sub_sub]
+
+lemma translate_add' (a b : G) (f : G → α) : τ (a + b) f = τ b (τ a f) := by
+  rw [add_comm, translate_add]
+
+lemma translate_comm (a b : G) (f : G → α) : τ a (τ b f) = τ b (τ a f) := by
+  rw [← translate_add, translate_add']
+
+@[simp] lemma comp_translate (a : G) (f : G → α) (g : α → β) : g ∘ τ a f = τ a (g ∘ f) := rfl
```
b-mehta, line 44: This is asserting that simp-normal-form for τ is as far out as possible - let's add a comment to this effect.

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,78 @@
...
+
+lemma translate_sub_right (a : G) (f g : G → H) : τ a (f - g) = τ a f - τ a g := rfl
+lemma translate_neg_right (a : G) (f : G → H) : τ a (-f) = -τ a f := rfl
+
+@[simp] lemma support_translate (a : G) (f : G → H) : support (τ a f) = a +ᵥ support f := by
+  ext; simp [mem_vadd_set_iff_neg_vadd_mem, sub_eq_neg_add]
+
+end AddCommGroup
+
+variable [CommRing R]
```
b-mehta, line 75: This looks far too strong: I think `CommMonoid R` is enough

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,78 @@
...
+
+@[simp] lemma comp_translate (a : G) (f : G → α) (g : α → β) : g ∘ τ a f = τ a (g ∘ f) := rfl
+
+@[simp]
+lemma translate_smul_right [SMul H α] (a : G) (f : G → α) (c : H) : τ a (c • f) = c • τ a f := rfl
+
+section AddCommMonoid
+variable [AddCommMonoid M]
+
+@[simp] lemma translate_zero_right (a : G) : τ a (0 : G → M) = 0 := rfl
```
b-mehta, line 52: It looks like AddCommMonoid M is too strong for this one and the one directly below: just having zero (resp add) seems enough?

b-mehta (approved):
I agree this definition is valuable for the application Yaël and I have in mind, but I'd also like to hear others' opinions on its general value.
Other than that, this looks good to me.

Mathlib/Algebra/Group/Translate.lean
YaelDillies: Do you have a good idea of what locale it should be in?

b-mehta: I don't :/

urkud:
I'm OK with adding this spelling. In fact, it turned out that I don't need `DomMulAct` for the goal I was aiming to use it for, but it looks like people found other uses. Please add a lemma linking your definition to `DomAddAct` action and mention `DomAddAct` somewhere in the module docs, explaining that `translate` spelling is more convenient in many applications.

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,93 @@
...
+abstracted by `DomAddAct` in such a way that `τ a f = DomAddAct.mk (-a) +ᵥ f` (see
+`translate_eq_domAddActMk_vadd`). Using `DomAddAct` is irritating in applications because of this
+negation appearing inside `DomAddAct.mk`. Although mathematically equivalent, the pen and paper
+convention is that translating is an action by subtraction, not by addition.
+-/
+
+open Function Set
+open scoped Pointwise
+
+variable {ι α β M G H : Type*} [AddCommGroup G]
```
jcommelin, line 32: I think this can be useful in noncommutative contexts as well. So please generalize the definition.

Mathlib/Algebra/Group/Translate.lean
```diff
@@ -0,0 +1,93 @@
...
+
+/-- Translation of a function in a group by an element of that group.
+`τ a f` is defined as `x ↦ f (x - a)`.  -/
+def translate (a : G) (f : G → α) : G → α := fun x ↦ f (x - a)
```
jcommelin, line 36: Should we multiplicitivize this already? I wouldn't be surprised if we want to use this for `Units Complex` or `GL n` or something like that.

YaelDillies: What name would you like me to use?

Mathlib/Algebra/Group/Translate.lean
YaelDillies: But then do you want left or right translation?

YaelDillies:
@jcommelin, both your comments are pushing towards `translate a f` being notation for `DomAddAct.mk (-a) +ᵥ f` (or maybe `DomAddAct.mk a +ᵥ f` against mathematical convention to account for the unwanted `DomAddAct.mk (-(-a)) +ᵥ f = DomAddAct.mk a +ᵥ f` simplification). Is this what you want?

I would be happy to switch over but, as previously pointed out, it goes against mathematical convention.

Mathlib/Algebra/Group/Translate.lean
jcommelin: right (pun intended), that's the question :grinning: 
We might want to have both. I always forget what left/right means: is the constant or the variable on the left/right?

Mathlib/Algebra/Group/Translate.lean
jcommelin: `leftMulTranslate`? Would that work?

jcommelin:
When you say "notation", do you mean `notation` or `abbrev`?
I don't think we should diverge from mathematical convention, but I also don't think I'm pushing for *that*.

YaelDillies:
Yes, I mean using `notation`.

Here are the four options I see:
1. Define `τ a f` as notation for `DomAddAct.mk (-a) +ᵥ f`. Follows mathematical convention, can be multiplicativised (eg `τₘ a f` as notation for `DomMulAct.mk a⁻¹ • f`) and also works for other actions (eg of a subgroup, on the right by using `τ (op a) f`). The not great thing is that simp would turn `τ (-a) f` into `DomAddAct.mk a +ᵥ f`.
2. Define `τ a f` as notation for `DomAddAct.mk a +ᵥ f`. Can be multiplicativised, also works for other actions (same as above), doesn't have an issue with simp removing the notation. The not so great thing is that it doesn't follow mathematical convention, but only to the extent that the action by `DomMulAct` doesn't*.
3. Define `τ a f` as a `def` which is equal to `DomAddAct.mk (-a) +ᵥ f`. Follows mathematical convention, but doesn't work for other actions and is not multiplicativised.
4. Define four kinds of `τ`-like notations as `def`s. Follows mathematical convention, but doesn't work for other actions.

*One could redefine the action by `DomMulAct` so that `(DomMulAct.mk c • f) a = f (c⁻¹ • a)`, instead of `(DomMulAct.mk c • f) a = f (c • a)` currently. What do you think, @urkud ?

Solution 1 won't be viable. Solution 3 is the one I currently have. Solution 4 is what you are pushing for, Johan. Solution 2 is what I see as the mathlib philosophy one, and it can be made to follow mathematical convention if we redefine the `DomMulAct` action.

jcommelin (commented):
Since there are some ope questions that don't have good answers right now, I'm going to merge this as is. When answers arrive, we can refactor.

bors merge

PR #16185, opened 2024-08-27
chore(SetTheory/Ordinal/Basic): ditch `Ordinal.inductionOn`

vihdzp:
`Ordinal.inductionOn` was a specialized version of `Quotient.inductionOn` for `Ordinal`. This PR removes it.

There's three things I don't like about this theorem:

1. It has a misleading name - this has nothing to do with the much more common transfinite induction.
2. At best, it's barely more useful than `Quotient.inductionOn`. I only had to insert `type_def` in a couple of places.
3. When you don't actually want to name the arguments, there's more blank spaces you need to insert. I imagine this same reason is why we didn't have `Ordinal.inductionOn₂` or `Ordinal.inductionOn₃`, even though the quotient versions are used all over the place - that would be nine variables to introduce, most of which you won't need!

Deletions: 
- `Ordinal.inductionOn`

---

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ordinal.2EinductionOn)

The large diff is largely due to changed proof indentation.

### DIFF
--- a/Mathlib/SetTheory/Cardinal/Cofinality.lean
+++ b/Mathlib/SetTheory/Cardinal/Cofinality.lean
@@ -236,8 +236,8 @@ theorem cof_eq_sInf_lsub (o : Ordinal.{u}) : cof o =
 
 @[simp]
 theorem lift_cof (o) : Cardinal.lift.{u, v} (cof o) = cof (Ordinal.lift.{u, v} o) := by
-  refine inductionOn o ?_
-  intro α r _
+  refine Quotient.inductionOn o ?_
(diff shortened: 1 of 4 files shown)


Command-Master:
Could you add
Deletions:
- Ordinal.inductionOn
to the bottom of the commit message?

vihdzp:
I've reconsidered, and I've come to the conclusion that this lemma is definitely useful, and that requiring the use of `Quotient.inductionOn` really only makes things more confusing. I'll do the opposite refactor where I clean up instances of `Quotient.inductionOn` soon enough.

PR #16186, opened 2024-08-27
chore: use emoji-variant-selector `\uFE0F` for emojis ✅️,❌️,💥️,🟡️

joneugster:
Append the emoji-variant selector `\uFE0F` to these 4 unicode symbols which should always be rendered as emojis.

---

This has been done in core, so it makes sense to do the same in Mathlib.

- [x] depends on: leanprover/lean4#5173
- [ ] depends on: #16215


urkud:
Since github diff viewer doesn't show what changed, could you please add more details to the commit message? E.g., write Unicode sequences before and after.

PR #16188, opened 2024-08-27
chore: purge `autoImplicit` from tests

Parcly-Taxel:
Surprisingly no adaptations were needed anywhere.

---

### DIFF
--- a/test/LintStyle.lean
+++ b/test/LintStyle.lean
@@ -76,7 +76,7 @@ set_option linter.setOption true in
 set_option debug.moduleNameAtTimeout false
 
 -- The lint does not fire on arbitrary options.
-set_option autoImplicit false
+set_option linter.dollarSyntax true
 
 -- We also cover set_option tactics.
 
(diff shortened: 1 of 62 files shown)


test/LintStyle.lean
```diff
@@ -76,7 +76,7 @@ set_option linter.setOption true in
 set_option debug.moduleNameAtTimeout false
 
 -- The lint does not fire on arbitrary options.
-set_option autoImplicit false
+set_option linter.dollarSyntax true
```
grunweg, line 79: I agree that changing to a different option is a good idea; I'm not too psyched about this one.
At the end of the day, I don't feel strongly about *which* new option you choose.

grunweg (commented):
If I remember correctly, mathlib's build flags don't take effect in `tests`: in other words, is `autoImplicit true` (the default from core) implicitly active in these files? Put differently, does everything still work if you change this to set_option autoImplicit false instead?

PR #16190, opened 2024-08-27
[Merged by Bors] - feat(Archive/Imo): formalize IMO 1982q3

AlexBrodbelt:
Formalize problem 3 of 1982 IMO.

Co-authored by: Violeta Hernández Palacios <vi.hdz.p@gmail.com>

---


Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,426 @@
...
+  ∀ N,
+  ∑ n : Fin N, x n = ∑ n ∈ Ico 0 N, x n
+  := by
+  intro N
+  rw [Fin.sum_univ_eq_sum_range, Nat.Ico_zero_eq_range]
+
+/-
+Specialization of Cauchy-Schwarz inequality with the sequences x n / √(y n) and √(y n)
+-/
+lemma Sedrakyan's_lemma
```
vihdzp, line 51: Admittedly I've never seen this lemma used outside of the math olympiad, but even then I think it's useful enough that it could be added as its own PR to Mathlib.

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,426 @@
...
+
+namespace Imo1982Q3
+
+
+lemma sum_Fin_eq_sum_Ico
+    {x : ℕ → ℝ}
+  :
+  ∀ N,
+  ∑ n : Fin N, x n = ∑ n ∈ Ico 0 N, x n
+  := by
```
vihdzp, line 44:
```suggestion
lemma sum_Fin_eq_sum_Ico {x : ℕ → ℝ} :
  ∀ N, ∑ n : Fin N, x n = ∑ n ∈ Ico 0 N, x n := by
```
Please don't write arguments like this, or at the very least, keep the semicolon in the same line as the last argument previous to it.

AlexBrodbelt: Ok, I'll keep this in mind. I'll correct the style of other statements as well

vihdzp: The latter is more standard throughout Mathlib, but the former should be ok too.

AlexBrodbelt: Ok, thank you :)

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,381 @@
...
+      rw [one_rpow, one_mul, ← Real.sqrt_eq_rpow, sqrt_sq _]
+      apply le_of_lt sum_fin_pos
+    _ ≤ 2 * ((1/2 : ℝ) * 1 + (1/2 : ℝ) * (∑ n : Fin N, x (n  + 1))^2) := by
+      rw [mul_le_mul_left (by norm_num)]
+      apply Real.geom_mean_le_arith_mean2_weighted
+        (by norm_num) (by norm_num) (by norm_num) (sq_nonneg _) (by norm_num)
+    _ ≤ 1 + (∑ n : Fin N, x (n  + 1))^2 := by field_simp
+
+lemma Ico_sdiff_zero_eq_Ico : ∀ N, Ico 0 N \ {0} = Ico 1 N := by
+    intro N
```
jsm28, line 211: General comments applicable both to this lemma and elsewhere in this PR:

* Normally rather than having a lemma statement start with `∀` after the colon and then the proof starting with `intro`, we use named hypotheses before the colon, so `(N : ℕ)` in this case, and so don't need to start the proof with `intro`. Much the same applies if the statement is of the form `h → p`: move the hypotheses `h` before the colon rather than having an implication on the right hand side of the colon.
* Standard indentation of the proof body is two columns, not four.

Specific remark about this lemma: it's essentially a combination of the existing lemmas `sdiff_singleton_eq_erase`, `Ico_erase_left` and `Ico_succ_left`. That should allow a much shorter proof. It might then be reasonable to inline that at the point where this lemma is needed.

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,365 @@
...
+  rw [mul_div_cancel₀ (hb := sqrt_ne_zero'.mpr (yi_pos _))]
+  have RHS_nonneg : 0 ≤ (∑ n : Fin n, y n) * (∑ n : Fin n, x n ^ 2 / y n):= by
+    rw [mul_nonneg_iff_of_pos_left]
+    apply Finset.sum_nonneg (fun i _ => div_nonneg (sq_nonneg _) (yi_nonneg i))
+    apply sum_yi_pos
+  have LHS_nonneg : 0 ≤ ∑ n : Fin n, √(y n) * (x n / √(y n)) := by
+    apply Finset.sum_nonneg
+    intro i _hi
+    apply mul_nonneg (le_of_lt (sqrt_yi_pos i))
+    apply div_nonneg (xi_nonneg _) (le_of_lt (sqrt_yi_pos _))
```
YaelDillies, line 72:
```suggestion
  have LHS_nonneg : 0 ≤ ∑ n : Fin n, √(y n) * (x n / √(y n)) := by positivity
```

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,365 @@
...
+
+lemma sum_Fin_eq_sum_Ico {x : ℕ → ℝ} {N : ℕ} : ∑ n : Fin N, x n = ∑ n ∈ Ico 0 N, x n := by
+  rw [Fin.sum_univ_eq_sum_range, Nat.Ico_zero_eq_range]
+
+/-
+Specialization of Cauchy-Schwarz inequality with the sequences x n / √(y n) and √(y n)
+-/
+lemma Sedrakyan's_lemma {n : ℕ} {x y: EuclideanSpace ℝ (Fin n)}
+    (hN : 0 < n) (xi_pos : ∀ i, 0 < x i) (yi_pos : ∀ i, 0 < y i) :
+  (∑ n : Fin n, x n)^2 / (∑ n : Fin n, y n) ≤ (∑ n : Fin n, ((x n)^2 / (y n))) := by
```
YaelDillies, line 48: This is a general fact about sequences of reals. Can you generalise? The lemma you want to use is `Finset.sum_mul_sq_le_sq_mul_sq` instead of `real_inner_le_norm`

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,365 @@
...
+    (hN : 1 < N) (hx : ∀ i , x (i + 1) ≤ x i) (x_pos : ∀ i, x i > (0 : ℝ)) :
+  (N - 1) / N * (1 / ∑ n : Fin (N - 1), x (n + 1)) ≤ 1 / (∑ n : Fin N, x (n + 1)) := by
+  have ne_zero : N - 1 ≠ 0 := by
+    intro h
+    rw [Nat.sub_eq_iff_eq_add (le_of_lt hN), zero_add] at h
+    rw [h] at hN; apply lt_irrefl _ hN
+  have ne_zero' : (N : ℝ) - 1 ≠ 0 :=  by
+    rw [ne_eq]; intro h
+    rw [sub_eq_iff_eq_add, zero_add] at h
+    rw [@Nat.cast_eq_one] at h
```
YaelDillies, line 130: These `@` are usually useless
```suggestion
    rw [Nat.cast_eq_one] at h
```

Archive/Imo/Imo1982Q3.lean
vihdzp: I'd also like to reiterate my petition for this lemma to be added within Mathlib proper. It will definitely be useful for other IMO problems!

hrmacbeth:
@AlexBrodbelt, I have the impression that you have quite a few tricks yet to learn about writing analysis arguments in Lean.  As an example, I have pushed an efficient proof of your first lemma.

The "Archive" is for high quality proofs that, while "leaves", show off the automation and libraries developed in Mathlib.  At the moment this proof is quite far from being this kind of showcase.  If you want to work on getting this proof to that level, it will take a lot of back-and-forth, and this may proceed rather slowly, depending on reviewers' availability to point out inefficiencies.

If you just want to make this proof available in a public repository for the community to use, you might consider instead contributing it to https://github.com/dwrensha/compfiles.  Are you aware of that project?

dwrensha:
FYI I added this solution to Compfiles: https://github.com/dwrensha/compfiles/commit/e7f00a521a0e1b5c0a80b4e759ff26e2c61ffc4d

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,329 @@
...
+
+namespace Imo1982Q3
+
+lemma sum_Fin_eq_sum_Ico {x : ℕ → ℝ} {N : ℕ} : ∑ n : Fin N, x n = ∑ n ∈ Ico 0 N, x n := by
+  rw [Fin.sum_univ_eq_sum_range, Nat.Ico_zero_eq_range]
+
+/-
+Specialization of Cauchy-Schwarz inequality with the sequences x n / √(y n) and √(y n)
+-/
+theorem Sedrakyan's_lemma {ι : Type*} {s : Finset ι} {x y : ι → ℝ}
```
vihdzp, line 45: This lemma actually holds without the `0 < #s` assumption, and it should also hold if the variables are nonnegative rather than positive (recall Lean's convention that `x / 0 = 0`).

I'm currently working on proving this more general version and I'll PR it into core Mathlib (crediting both you and Heather Macbeth, of course).

vihdzp: Correction: `g` does need to be a sequence of positive reals.

Archive/Imo/Imo1982Q3.lean
vihdzp: #19311

vihdzp:
@AlexBrodbelt I did a complete rewrite of the proof on [my own branch](https://github.com/leanprover-community/mathlib4/tree/vi.imo1982_p3). It ends up being only about 100 lines long, and I think it better showcases Mathlib's automation. If you think what I did is sensible, you can merge the changes in that branch with this branch. Note that my proof depends on PR #19311.

vihdzp:
Since there's been no activity from OP since September, I've decided to take over this PR. The new proof is an almost complete rewrite from the original one, and makes use of the version of [Sedrayan](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/BigOperators/Ring/Finset.html#Finset.sq_sum_div_le_sum_sq_div) I had previously PR'd onto Mathlib.

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,96 @@
...
+-/
+import Mathlib.Algebra.Order.BigOperators.Ring.Finset
+import Mathlib.Algebra.GeomSum
+import Mathlib.Data.NNReal.Basic
+
+/-!
+# IMO 1982 Q3
+
+Consider infinite sequences $\{x_n\}$ of positive reals such that $x_0 = 0$ and
+$x_0 \ge x_1 \ge x_2 \ge \ldots$
```
YaelDillies, line 14: ... so `x_0 = x_1 = x_2 = ... = 0`?

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,96 @@
...
+  convert ineq₂ hn hx h0 hp using 1
+  field_simp
+
+/-- Finally, AM-GM gives us the main lower bound for the RHS. -/
+lemma ineq₄ : 4 * n / (n + 1) ≤ ∑ k ∈ range (n + 1), x k ^ 2 / x (k + 1) := by
+  apply le_trans _ (ineq₃ hn hx h0 hp)
+  gcongr
+  rw [le_div_iff₀]
+  · simpa using four_mul_le_sq_add (∑ k ∈ range n, x (k + 1)) 1
+  · exact sum_pos (fun k _ ↦ hp _) (nonempty_range_iff.2 hn)
```
YaelDillies, line 71: Can you please turn this into a single `calc` block showcasing the powerful combination of `calc` + `gcongr` + `field_simp` + `ring` + `linarith` + `omega`?

vihdzp: I've merged `ineq₂` through `ineq₄` into a single calc block. I found it cleaner to keep `ineq₁` as an auxiliary lemma, since I instantiate it with `x ⚬ succ` rather than just `x`.

YaelDillies (approved):
Thanks! 🚀

maintainer merge

Archive/Imo/Imo1982Q3.lean
```diff
@@ -0,0 +1,91 @@
...
+-/
+
+open Finset NNReal
+
+variable {x : ℕ → ℝ} {n : ℕ} (hn : n ≠ 0) (hx : Antitone x)
+
+namespace Imo1982Q3
+include hn hx
+
+/-- `x (n + 1)` is at most the average of `x 1`, `x 2`, ..., `x n`. -/
```
kbuzzard, line 38: This docstring does not seem to correspond to the lemma it's documenting.

vihdzp: The actual lemma is just this plus `∑ k ∈ range n, x k` added to both sides.

kbuzzard: That's not true either because of various errors involving bounds (`range n` is 0<i<n, right?).

vihdzp: You're right, I've updated the docstring.

kbuzzard: So now the docstring does claim a result which is mathematically equivalent to the one being proved, but it still feels very weird to me that the docstring doesn't actually say what the theorem says. Everything else in this PR looks fine but can't you at least explain in the docstring that the theorem doesn't actually say what you say it says? i.e. incorporate your comment above about what the lemma actually says, into the docstring? Or just actually prove the result you claim in the docstring as another lemma? Why prove X and then claim that you're proving X'? It just seems illogical to me.

vihdzp: I've added my comment on `∑ k ∈ range n, x k` to the docstring.

kbuzzard:
Thanks!

bors merge

PR #16191, opened 2024-08-27
[Merged by Bors] - chore: add the stacks attribute and extension

adomani:
This extends the simple attribute of #16179 with an environment extension that tracks also the declaration names and tags.

The `#stacks_tags` command retrieves all declarations that have the `stacks` attribute.

For each found declaration, it prints a line
```
'declaration_name' corresponds to tag 'declaration_tag'.
```
The variant `#stacks_tags!` also adds the theorem statement after each summary line.

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/stacks.20tags)

---

### DIFF
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -58,9 +58,6 @@ See the [Tags page](https://stacks.math.columbia.edu/tags) in the Stacks project
 -/
 syntax (name := stacks) "stacks " (stackTag)? (ppSpace str)? : attr
 
-initialize stacksTagAttribute : TagAttribute ←
-  registerTagAttribute `stacks1 "hel"
-
 initialize Lean.registerBuiltinAttribute {
   name := `stacks
   descr := "Apply a Stacks project tag to a theorem."
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -1,7 +1,7 @@
 /-
-Copyright (c) 2024 Dmaiano Testa. All rights reserved.
+Copyright (c) 2024 Damiano Testa. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Dmaiano Testa
+Authors: Damiano Testa
 -/
 import Lean.Parser.Syntax
 
--- a/Mathlib/Tactic/StacksAttribute.lean
+++ b/Mathlib/Tactic/StacksAttribute.lean
@@ -84,10 +84,19 @@ end Mathlib.Stacks
 
 open Mathlib.Stacks
 /--
-`getSortedTags env` returns the array of `Tags`, sorted by alphabetical order of declaration name.
+`getSortedStackProjectTags env` returns the array of `Tags`, sorted by alphabetical order of tag.
 -/
-def Lean.Environment.getSortedTags (env : Environment) : Array Tag :=
-  tagExt.getState env |>.toArray.qsort (·.declName.toString < ·.declName.toString)
+def Lean.Environment.getSortedStackProjectTags (env : Environment) : Array Tag :=
+  tagExt.getState env |>.toArray.qsort (·.tag < ·.tag)
+
+/--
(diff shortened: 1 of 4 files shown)


mattrobball:
Looks great! Can you fix the failing `docBlame`?

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,117 @@
...
+
+/-- The `stacks` attribute.
+Use it as `@[stacks TAG "Optional comment"]`.
+The `TAG` is mandatory.
+
+See the [Tags page](https://stacks.math.columbia.edu/tags) in the Stacks project for more details.
+-/
+syntax (name := stacks) "stacks " (stackTag)? (ppSpace str)? : attr
+
+initialize stacksTagAttribute : TagAttribute ←
```
mattrobball, line 61: Do we need this?

adomani: No, this is a left-over from a previous attempt: I removed it!

adomani (commented):
Matt, thanks for the review!  I should have taken care of cleaning up the left-overs!

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,113 @@
...
+          | _ => throwUnsupportedSyntax
+}
+
+end Mathlib.Stacks
+
+open Mathlib.Stacks
+/--
+`getSortedTags env` returns the array of `Tags`, sorted by alphabetical order of declaration name.
+-/
+def Lean.Environment.getSortedTags (env : Environment) : Array Tag :=
```
mattrobball, line 89: This is big ambiguous from the name. I think most people would guess it returns the tags sorted by the stack project tag, like the master list on the stacks project repo. 

I'm guessing there will be more querying using the stacks project tag than the name of the registered declaration so adding a function here to do that would nice also.

adomani: Ok, I'll rename to `Lean.Environment.getSortedStackProjectTags`.

The second comment is about creating a function that takes a (string) "tag" as input and returns an `Option Name`?

mattrobball: I think you should just switch to returning the `Stacks.Tag`s sorted by the `tag` rather than `declName`. 

The second comment was about a `Lean.Environment.findStacksTag` function which indeed returned an `Option Name` and consumed the string Stacks project tag.

adomani: Ok, great!  I'll return an `Array Name`, since conceivably the same Stacks tag could be applied to multiple lemmas, if they each prove a part of the Stacks retult.

adomani: I just pushed both changes

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -0,0 +1,113 @@
...
+      | none => logWarning "Please, enter a Tag after `stacks`."
+      | some str =>
+        let str := str.toString.trimRight
+        if str.length != 4 then
+          logWarningAt tag
+            m!"Tag '{str}' is {str.length} characters long, but it should be 4 characters long"
+        else if 2 ≤ (str.split (fun c => (!c.isUpper) && !c.isDigit)).length then
+          logWarningAt tag m!"Tag '{str}' should only consist of digits and uppercase letters"
+        else match stx with
+          | `(attr| stacks $_:stackTag $_:str) => addTagEntry str decl
```
mattrobball, line 78: Uh oh. Are we keeping track of the comments at all?

adomani: I thought that the comments would be discarded: should the `Stacks.Tag` record them as well?

mattrobball: Yeah, I think that was why Kim asked for them. To record discrepancies between the two statements.

adomani: Comments added!

mattrobball:
Ok. I am ready to merge when CI is green.

mattrobball:
bors merge

PR #16192, opened 2024-08-27
[Merged by Bors] - chore(ModelTheory): Tweaks involving the empty language

awainverse:
Makes `FirstOrder.Language.emptyStructure` a `def` instead of an `instance`
Removes `def`s that could be accomplished with `StrongHomClass.toEmbedding` and `StrongHomClass.toEquiv`

---


YaelDillies (approved):
Thanks!

maintainer merge

jcommelin (commented):
Thanks :tada:

bors merge

PR #16193, opened 2024-08-27
[Merged by Bors] - chore(ModelTheory): Add @[simps] to functions creating maps from `HomClass` instances

awainverse:
Adds `@[simps]` tag to `HomClass.toHom`, `StrongHomClass.toEmbedding`, `StrongHomClass.toEquiv`

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16194, opened 2024-08-27
[Merged by Bors] - feat(ModelTheory/Fraisse): Shows that any countably infinite structure in the empty language is a Fraïssé limit

awainverse:
Shows that any countably infinite structure in the empty language is the Fraïssé limit of the class of finite structures, which is thus a Fraïssé class

---

- [x] depends on: #16192 
- [x] depends on: #16193 


YaelDillies (approved):
Thanks!

maintainer merge

Mathlib/ModelTheory/Fraisse.lean
```diff
@@ -403,6 +407,42 @@ theorem nonempty_equiv : Nonempty (M ≃[L] N) := by
...
+    (M : Type*) [Countable M] [Infinite M] [Language.empty.Structure M] :
+    IsFraisseLimit { S : Bundled Language.empty.Structure | Finite S } M where
+  age := by
+    ext S
+    simp only [age, Structure.fg_iff_finite, empty.nonempty_embedding_iff, Cardinal.lift_id,
+      Cardinal.mk_eq_aleph0, mem_setOf_eq, and_iff_left_iff_imp]
+    exact fun hS => Cardinal.mk_le_aleph0
+  ultrahomogeneous := by
+    intro S hS f
+    haveI : Finite S := hS.finite
```
b-mehta, line 424: Is there a reason these need to be `haveI` rather than `have`?

awainverse: The only reason is that I haven't fully converted from Lean3 syntax. Should I also change this elsewhere in the file, or will that make the diff too weird?

Mathlib/ModelTheory/Fraisse.lean
```diff
@@ -403,6 +407,42 @@ theorem nonempty_equiv : Nonempty (M ≃[L] N) := by
...
+    have : Infinite { x // x ∉ f.toHom.range } :=
+      (Set.Finite.infinite_compl (Set.toFinite _)).to_subtype
+    refine ⟨StrongHomClass.toEquiv (@Equiv.subtypeCongr _ _ _ (Classical.decPred _)
+      (Classical.decPred _) f.equivRange.toEquiv (nonempty_equiv_of_countable).some), ?_⟩
+    ext x
+    simp only [Equiv.subtypeCongr, Embedding.comp_apply, coeSubtype, Equiv.coe_toEmbedding,
+      StrongHomClass.toEquiv_toFun, Equiv.trans_apply, SetLike.coe_mem,
+      Equiv.sumCompl_apply_symm_of_pos, Subtype.coe_eta, Equiv.sumCongr_apply, Sum.map_inl,
+      Equiv.sumCompl_apply_inl]
+    rfl
```
b-mehta, line 437:
```suggestion
theorem isFraisseLimit_of_countable_infinite
    (M : Type*) [Countable M] [Infinite M] [Language.empty.Structure M] :
    IsFraisseLimit { S : Bundled Language.empty.Structure | Finite S } M where
  age := by
    ext S
    simp only [age, Structure.fg_iff_finite, mem_setOf_eq, and_iff_left_iff_imp]
    intro hS
    simp
  ultrahomogeneous S hS f := by
    classical
    have : Finite S := hS.finite
    have : Infinite { x // x ∉ S } := ((Set.toFinite _).infinite_compl).to_subtype
    have : Finite f.toHom.range := (((Substructure.fg_iff_structure_fg S).1 hS).range _).finite
    have : Infinite { x // x ∉ f.toHom.range } := ((Set.toFinite _).infinite_compl ).to_subtype
    refine ⟨StrongHomClass.toEquiv (f.equivRange.subtypeCongr nonempty_equiv_of_countable.some), ?_⟩
    ext x
    simp [Equiv.subtypeCongr]
```
if you'd like, you can golf this quite a bit! 

The last `simp` does require 
`@[simps toEquiv_apply]
noncomputable def equivRange`
in Substructures.lean though. I think the `simps` setup for some of these structures doesn't look quite optimal, but the general fixes can be a later PR.

Mathlib/ModelTheory/Fraisse.lean
b-mehta: I'd leave it for now!

Mathlib/ModelTheory/Fraisse.lean
awainverse: Thanks for the golf - I did try doing something with `classical`, but I don't know why it didn't work.

Could you point out to me what kind of changes you'd suggest to the `simps` setup?
(Probably a conversation for Zulip, because I agree it's better to save them for a later PR)

jcommelin (commented):
Thanks :tada:

bors merge

PR #16198, opened 2024-08-27
[Merged by Bors] - chore: do not directly import `Tactic.Cases`

Parcly-Taxel:
This entails replacing not only `induction'` by its unprimed variant, but also `cases'` by `cases/rcases/obtain`. This reduces the number of imports for some files.

---


Mathlib/Data/Nat/Size.lean
```diff
@@ -57,8 +57,9 @@ end
 @[simp]
 theorem size_shiftLeft' {b m n} (h : shiftLeft' b m n ≠ 0) :
     size (shiftLeft' b m n) = size m + n := by
-  induction' n with n IH
-  · simp [shiftLeft']
+  induction n with
+  | zero => simp [shiftLeft']
+  | succ n IH =>
```
semorrison, line 62: Indent the remaining proof?

Mathlib/Order/WithBot.lean
```diff
@@ -338,16 +338,18 @@ instance preorder [Preorder α] : Preorder (WithBot α) where
     let ⟨c, hc, bc⟩ := h₂ b hb
     ⟨c, hc, le_trans ab bc⟩
 
-instance partialOrder [PartialOrder α] : PartialOrder (WithBot α) :=
-  { WithBot.preorder with
```
semorrison, line 342: For simplicity, best not to make random change like the `__` change here. (Potentially subtle consequences, and orthogonal to everything else, so make an easy to review PR not quite as easy.)

semorrison:
bors d+

Parcly-Taxel:
bors merge

PR #16203, opened 2024-08-27
[Merged by Bors] - chore(Linter/TextBased): namespace declarations

grunweg:
Currently, all declarations of the text-based linters live in the root namespace; this can cause conflicts (and is to be used sparingly anyway).

---


semorrison:
bors merge

PR #16209, opened 2024-08-28
[Merged by Bors] - feat: `1 - x⁻¹ ≤ log x`

YaelDillies:
From LeanAPAP

Co-authored-by: Bhavik Mehta <bhavik.mehta8@gmail.com>

---

### DIFF
--- a/Mathlib/Analysis/SpecialFunctions/Log/Basic.lean
+++ b/Mathlib/Analysis/SpecialFunctions/Log/Basic.lean
@@ -275,10 +275,8 @@ theorem log_le_sub_one_of_pos {x : ℝ} (hx : 0 < x) : log x ≤ x - 1 := by
   convert add_one_le_exp (log x)
   rw [exp_log hx]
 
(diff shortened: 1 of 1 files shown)


b-mehta:
I'm not really sure this deserves to be a lemma in this file (or even a public lemma at all) - it's a trivial consequence of `log_le_sub_one_of_pos` and should be used as such in applications

YaelDillies:
Given that we spent several hours trying to prove it on Xena, I'm not sure I would call that "trivial"...

b-mehta:
I disagree pretty strongly on that - you can see from the proof that it's literally an application of that lemma combined with basic algebra - variants like this show up really often in practice, clearly not all of them should be lemmas.

The one variant I wouldn't mind having is 1 - 1/x <= log(x), since then we have a lower and upper bound on log. Again your lemma is a trivial consequence of this.

b-mehta:
Okay now I'm happy :)

bors merge

PR #16211, opened 2024-08-28
fix: hack for binary numbers in Stacks Tags

adomani:
While I find the time to actually fix the issue with parsing, this is a temporary solution for being able to use `@[stacks 0BR8]`.

[Reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/stacks.20tags/near/465660576)

---

### DIFF
--- a/test/StacksAttribute.lean
+++ b/test/StacksAttribute.lean
@@ -24,7 +24,7 @@ warning: Please, enter a Tag after `stacks`.
 warning: Please, enter a Tag after `stacks`.
 -/
 #guard_msgs in
-@[stacks "", stacks]
+@[stacks "", stacks,stacks «0B81»]
 example : True := .intro
 
 /--
--- a/test/StacksAttribute.lean
+++ b/test/StacksAttribute.lean
@@ -24,7 +24,7 @@ warning: Please, enter a Tag after `stacks`.
 warning: Please, enter a Tag after `stacks`.
 -/
 #guard_msgs in
-@[stacks "", stacks,stacks «0B81»]
+@[stacks "", stacks, stacks «0B81»]
 example : True := .intro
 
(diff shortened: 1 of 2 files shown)


TwoFX:
Could we get a test for this or would the test interfere with the data extraction we are planning to do in the future?

adomani:
Ah, I meant to include a test.  It would not interfere with harvesting data, since it would be in the `test` folder, but I'm no longer able to modify the PR for several hours.

If someone else wants to add a test, something as simple as
```lean
@[stacks «08B1»]
example : True := .intro
```
would do.

If someone else wants to push this through, feel free to do so!

test/StacksAttribute.lean
```diff
@@ -24,7 +24,7 @@ warning: Please, enter a Tag after `stacks`.
 warning: Please, enter a Tag after `stacks`.
 -/
 #guard_msgs in
-@[stacks "", stacks]
+@[stacks "", stacks,stacks «0B81»]
```
grunweg, line 27:
```suggestion
@[stacks "", stacks, stacks «0B81»]

grunweg (commented):
Would you like to add a comment that this is a hack, and should be fixed properly?
maintainer merge

adomani:
> Would you like to add a comment that this is a hack, and should be fixed properly? maintainer merge

Added a comment on the temporary fix.

adomani:
#16230 is a "competing" PR that tries to implement a custom parser for the tags, rather than building it up from available parsers.  At most one of these two PRs should be merged, I think.  My preference would be to merge #16230 rather than this PR, assuming that #16230 does not have issues.

semorrison:
Closed in favour of #16230

PR #16214, opened 2024-08-28
[Merged by Bors] - chore(Pow/Real): fix lemma names about `Nat.cast`/`Int.cast`

YaelDillies:
From LeanAPAP

---


eric-wieser (approved):
bors merge

Thanks!

PR #16215, opened 2024-08-28
feat(Tactic/Linter): lint unwanted unicode

adomasbaliuka:
Add a text-based style linter that checks all unicode characters. 

- Only allow whitelisted characters, with a suggestion to add the flagged character to the whitelist
- Some characters are specified to use emoji-variant or text-variant. The linter ensures these variant-selectors only appear in the correct places and provides an auto-fix to clean them up.

---

Discussed at [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Whitelist.20for.20Unicode.3F/near/448721083)

The current proposal checks each character against a whitelist. There is also the question of what the whitelist should contain. 

A reasonable whitelist might be "all unicode currently in Mathlib" ∪ [Wikipedia/Mathematical_operators_and_symbols_in_Unicode](https://en.wikipedia.org/wiki/Mathematical_operators_and_symbols_in_Unicode) ∪ [EdAyers/mathlib3/docs/unicode.md](https://github.com/EdAyers/mathlib/blob/fb5c2ffe917049959597b0a93b872e91f4004922/docs/unicode.md). Currently implemented is a subset of this list closer to what's currently present in mathlib.

- [x] depends on: #17068
- [x] depends on: #17116
- [x] depends on: #17129 (part 1 of this PR)
- ~depends on #34719 (part 1 of this PR)~
-   ~depends on: #17131 (part 2 of this PR)~ (obsolete)
- [x]  depends on: #34022 (part 2 of this PR)
- Future note: @adomasbaliuka says: "we might add something to the Mathlib 4 docs about this as well?"


Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -103,6 +107,8 @@ def StyleError.errorMessage (err : StyleError) (style : ErrorFormat) : String :=
     | ErrorFormat.exceptionsFile =>
         s!"{sizeLimit} file contains {currentSize} lines, try to split it up"
     | ErrorFormat.humanReadable => s!"file contains {currentSize} lines, try to split it up"
+  | StyleError.newUnicode c =>
+      s!"Unicode character {c} is not recommended. Consider adding it to the whitelist."
```
grunweg, line 111:
```suggestion
      s!"unicode character '{c}' is not recommended. Consider adding it to the whitelist."

File: Mathlib/Tactic/Linter/TextBased.lean [hunk 2]
```diff
@@ -114,6 +120,7 @@ def StyleError.errorCode (err : StyleError) : String := match err with
   | StyleError.broadImport _ => "ERR_IMP"
   | StyleError.lineLength _ => "ERR_LIN"
   | StyleError.fileTooLong _ _ _ => "ERR_NUM_LIN"
+  | StyleError.newUnicode _ => "NO_ERR_CODE_AVAILABLE"
```
grunweg, line 123: You can just make one up, perhaps "ERR_UNICODE".

Can you also update the logic for `parse?_errorContext` below, please? (You can follow how e.g. the `ERR_AUT` code is treated.)

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -380,9 +387,44 @@ def checkFileLength (lines : Array String) (existingLimit : Option ℕ) : Option
 
 end
 
+
+section unicodeLinter
+
+/- Hashable instance for use with the unicodeLinter -/
```
grunweg, line 393: Does this fix it? A /- Comment -/ is just a multi-line comment in Lean, a /-- Comment -/ is a documentation comment.
```suggestion
/- Hashable instance for use with the unicodeLinter -/

File: Mathlib/Tactic/Linter/TextBased.lean [hunk 4]
```diff
@@ -380,9 +387,44 @@ def checkFileLength (lines : Array String) (existingLimit : Option ℕ) : Option
...
+local instance : Hashable Char where
+  hash c := c.val.toUInt64
+
+/-- TODO make complete and order nicely -/
```
grunweg, line 397: Yes, ordering this would be great. I presume this contains numbers, all ASCII letters, space etc., but it's very hard for me to check (short of using the search function ~60 times).

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -380,9 +387,44 @@ def checkFileLength (lines : Array String) (existingLimit : Option ℕ) : Option
...
+
+/-- Lint a collection of input strings if one of them contains unwanted unicode. -/
+def unicodeLinter : TextbasedLinter := fun lines ↦ Id.run do
+  let mut errors : Array (StyleError × ℕ) := Array.mkEmpty 0
+  let mut lineNumber := 1
+  for line in lines do
+    let badChars := line.toList.filter (fun c => !unicodeWhitelist.contains c)
+    if badChars.length > 0 then
+      let newErrors : Array (StyleError × ℕ) :=
+        badChars.toArray.map fun c ↦ (StyleError.newUnicode c, lineNumber)
```
grunweg, line 418: Here's a small Lean trick, just so you know.
You can also use \centerdot to write this more compactly as `badChars.toArray.map (StyleError.newUnicode ·, lineNumber)`; same for the definition of `badChars` three lines higher.

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -6,6 +6,7 @@ Authors: Michael Rothgang
 
 import Batteries.Data.String.Matcher
 import Mathlib.Data.Nat.Notation
+import Batteries.Lean.HashSet
```
grunweg, line 9: Can you sort the imports alphabetically, if possible? Thanks!

grunweg (commented):
Some quick comments. Overall, this linter
- flags interesting things
- does not seem to turn up something worrying,
both of which is great.

grunweg:
Would you like to take a look at the "spaces" the linter flags? (I haven't checked what this is about exactly; just looking at it in the text editor is probably not useful...)

adomasbaliuka:
> Would you like to take a look at the "spaces" the linter flags? (I haven't checked what this is about exactly; just looking at it in the text editor is probably not useful...)

That seems to be

 ' ': Unicode U+202F (category Zs: Separator, space)

I think we'll remove them from Mathlib but let's see...

adomasbaliuka:
I have removed all non-breaking spaces (U+00a0) from Mathlib. This might be excessive:
- I can PR it separately if the change is too big or off topic in this PR
- We may decide to *allow* non-breaking spaces

grunweg:
Thank you! I think splitting the removal of non-breaking spaces into a separate PR (do you know git cherry-pick?) makes sense, and would be happy to nominate it for the maintainers to look. Reading the zulip discussion, changing the spaces does not appear controversial to me.

adomasbaliuka:
I split up the whitelist strings into `Array Char` because this way it's more clear what the individual characters are.

I also removed some more weird characters (invisible, or ones that my VSCode font does not support). We can add them back if needed once most of the non-breaking whitespace is gone...

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: I put the "test" I made for this `parse?_errorContext` into ` test/LintStyle.lean`.

joneugster:
@adomasbaliuka I've added the logic to validate the variant-selector characters here: #16368

Maybe you could have a look, especially at the changes in [`TextBased.lean`](https://github.com/leanprover-community/mathlib4/pull/16368/files#diff-c973e59ceee698f73d927eed559746d5b761304155603aac9f1f41f7375b7e85) and if you think they are reasonable, I could directly push them to this PR.

In particular it showed me that there is already one place in mathlib where there are characters followed by both selectors `\uFE0E\FE0F`, which is not valid Unicode, so having a linter for this would certainly be good.

Mathlib/Init/Data/String.lean
joneugster, line 1: I didn't know where to put these, so I put them here for now.

I'll make a PR to Lean core to modify `String.get` now, so hopefully `get₂` isn't required by the time this is up for review.

joneugster:
@adomasbaliuka I've pushed my suggestions from #16368 here, feel free to take it from here and make changes as you see fit!

I left the lint errors unfixed so you have a chance to look at them, I figured it's only worth fixing them once your happy with the linter

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -371,9 +431,231 @@ def checkFileLength (lines : Array String) (existingLimit : Option ℕ) : Option
...
+
+    for (selector, symbols) in #[
+        (UnicodeVariant.emoji, mathlibEmojiSymbols),
+        (UnicodeVariant.text, mathlibTextSymbols) ] do
+      -- Ensure specified emojis/text-symbols have the correct variant-selector
+      let charPos := line.findAll (symbols.contains ·)
+      for pos in charPos do
+        let nextC := line.get₂ (line.next pos)
+        let previousC := line.get₂ (line.prev pos) -- may be the char. itself if `pos == 0`
+        unless (previousC == '\'' ∧ nextC == '\'') ∨ nextC == selector do
```
adomasbaliuka, line 640: What is the logic here? What do the ' characters signify?

adomasbaliuka:
@joneugster  I completely changed the way it's implemented (but I think it should do the same thing as your version, except about the ' characters where I left a comment). 

The whitelist is now only used to ignore characters. Special cases such as variant selectors and emojis are checked further and then allowed despite not being in the whitelist if they are used correctly. This way we have only one lookup in the whitelist for each whitelisted character, which should be faster (not sure how relevant speed is but I hope this code is clearer).

There may be an issue with the usage of the `next` and `prev` functions. They operate on UTF8 bytes and give unspecified results if given an invalid position (which I'm not sure can happen with the current code but I think yes?!).

I'm also not sure about which characters exactly should have which selectors if any. Some arrows in Mathlib have text selectors, for example. For several characters in the emoji-list, the selectors don't seem to be doing anything. Can we check this once more?

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +377,507 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+        else
+          errors := errors.push (StyleError.unicodeVariant ⟨[thisC]⟩ none pos, lineNumber)
+      else  -- End of "special circumstances".
+        -- Linter rejects generic non-whitelisted unicode.
+        errors := errors.push (StyleError.unwantedUnicode thisC, lineNumber)
+    lineNumber := lineNumber + 1
+  return errors
+
+
+namespace unicodeLinter
```
adomasbaliuka, line 630: This stuff is duplicated here

Mathlib/Tactic/Linter/TextBased.lean
joneugster: outdated now

Mathlib/Init/Data/String.lean
joneugster: (we dont actually need this to be changed in core. Replaced the one remaining occurence with `s.get? _ |>.getD '\uFFFD')`

Mathlib/Tactic/Linter/TextBased.lean
joneugster: I believe this is fixed

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +379,261 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+  || othersInMathlib.contains c
+
+/-- Creates `StyleError`s for non-whitelisted unicode characters as well as
+bad usage of (emoji/text)-variant-selectors. -/
+def findBadUnicodeAux (s : String) (pos : String.Pos) (c : Char)
+    (err : Array StyleError := #[]) : Array StyleError :=
+  if h : pos < s.endPos then
+    have := Nat.sub_lt_sub_left h (String.lt_next s pos)
+    let posₙ := s.next pos
+    let cₙ := s.get? posₙ |>.getD '\uFFFD' -- �
```
joneugster, line 581: This is the only occurence of `String.next`. @adomasbaliuka I guess this is fine, isn't it?

joneugster:
@adomasbaliuka What do you think of the version I now pushed? Maybe you can test for speed again. This way should be even less looping and it's overall a bit shorter.

Also, maybe you know better how to write the `parse?_errorContext` entry?

Let me know what you think, and if you're happy I could clean up the text and tests and fix the linter warnings

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +378,260 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+def othersInMathlib := #[
+'▼', 'c', 'ō', '⏩', '❓',
+'🆕', 'š', 'ř', '⚬', '│', '├', '┌', 'ő',
+'⟍', '̂', 'ᘁ', 'ń', 'ć', '⟋', 'ỳ', 'ầ', '⥥',
+'ł', '◿', '◹', '－', '＼', '◥', '／', '◢', 'Ž', 'ă',
+'И', 'в', 'а', 'н', 'о', 'и', 'ч', 'Š', 'ᴜ', 'ᵧ', '´',
+'ᴄ', 'ꜰ', 'ß', 'ᴢ', 'ᴏ', 'ᴀ', 'ꜱ', 'ɴ', 'ꞯ', 'ʟ',
+'ʜ', 'ᵟ', 'ʙ', 'ᵪ', 'ᵩ', 'ᵦ', 'ᴊ', 'ᴛ', 'ᴡ', 'ᴠ',
+'ɪ', '̀', 'ᴇ', 'ᴍ', 'ʀ', 'ᴅ', 'ɢ', 'ʏ', 'ᴘ', 'ĝ', 'ᵨ',
+'ᴋ', 'ś', '꙳', '𝓡', '𝕝', '𝖣', '⨳' ]
```
joneugster, line 553: I think it's fine to remain conservative with the non-emoji characters, and even leave it empty for now.

And regarding your question further up, I think it's fine if you add whatever whitelist you like best to the `othersInMathlib`. Although, it's also desirable that this list here eventually decreases in size in favour of adding VSCode abbreviations for the characters.

adomasbaliuka: Leaving the non-emojis empty is fine for me.

I don't understand the other point, sorry. Add what whitelist to what?

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +379,261 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+
+/-- Hashable instance for use with the unicodeLinter -/
+local instance instHashableChar : Hashable Char where
+  hash c := c.val.toUInt64
+
+/-- Allowed whitespace characters -/
+def ASCII.whitespace (c : Char) := #[' ', '\n'].contains c
+
+/-- Printable ASCII characters, not including whitespace. -/
+def ASCII.printable (c : Char) := 21 ≤ c.toNat || c.toNat ≤ 126
```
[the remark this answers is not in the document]
joneugster: ofc thanks for spotting🤦🏻‍♂️🤦🏻‍♂️

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +378,260 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
 
+
+namespace unicodeLinter
+
+/-- Hashable instance for use with the unicodeLinter -/
+local instance instHashableChar : Hashable Char where
+  hash c := c.val.toUInt64
+
+/-- Allowed whitespace characters -/
+def ASCII.whitespace (c : Char) := #[' ', '\n'].contains c
```
adomasbaliuka, line 389: I would call this e.g. `allowedWhitespace` so as to avoid confusion with the concept "all ASCII whitespace".

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -190,6 +246,15 @@ def parse?_errorContext (line : String) : Option ErrorContext := Id.run do
...
             some (StyleError.broadImport BroadImports.Lake)
+        | "ERR_UNICODE" =>
+            if let some str := errorMessage.get? 2 then
+              if let some c := str.get? ⟨1⟩ then
+                some (StyleError.unwantedUnicode c)
+              else none
+            else none
+        | "ERR_UNICODE_VARIANT" =>
+          -- some (StyleError.unicodeVariant _ _ _)
+          none --TODO
```
adomasbaliuka, line 257: I think the idea here is to parse back the text representation of the linter error into an `ErrorContext` object. I don't actually know what it's used for. If I understand it correctly, it should be the inverse of `outputMessage`, which actually isn't strictly possible the way that we print the error (we ignore the second character, for example). 

Probably this doesn't matter and we just "best-effort" parse the error message.  I could do it if you like...

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: Seems fine, assuming the function is never called with an invalid position. Maybe put that in the docstring? (we inherit the fact that the result is "unspecified" for invalid `pos`)

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -314,9 +378,260 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+      findBadUnicodeAux s posₙ cₙ errₙ
+    else if ! isAllowedCharacter c then
+      -- bad: character not allowed
+      let errₙ := err.push (.unwantedUnicode c)
+      findBadUnicodeAux s posₙ cₙ errₙ
+    else
+      -- okay
+      findBadUnicodeAux s posₙ cₙ err
+  -- emojis/non-emojis should not be the last character in the line
+  -- in practise, the last char is `\n`, so this is in theory superfluous.
```
adomasbaliuka, line 604: I believe with the way this linter is called, the final \n is not part of the "line" we process.

test/LintStyle.lean
```diff
@@ -135,3 +135,18 @@ lemma foo' : True := trivial
...
+open Mathlib.Linter.TextBased
+
+#guard let errContext : ErrorContext := {
+    error := .unwantedUnicode 'Z', lineNumber := 4, path:="./MYFILE.lean"}
+  (parse?_errorContext <| outputMessage errContext .exceptionsFile) == some errContext
+
+-- The list `othersInMathlib` should only contain characters not present in another list.
+#guard unicodeLinter.othersInMathlib.toList ∩ unicodeLinter.withVSCodeAbbrev.toList = ∅
+#guard unicodeLinter.othersInMathlib.toList ∩ unicodeLinter.emojis.toList = ∅
+#guard unicodeLinter.othersInMathlib.toList ∩ unicodeLinter.nonEmojis.toList = ∅
```
adomasbaliuka, line 150: Maybe we should add some more here? Intersection of `emojis` and `nonEmojis` for example?

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: I did it.

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: I renamed it.

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: I removed the comment and kept the check. Please double-check if you agree.

test/LintStyle.lean
adomasbaliuka: I added some.

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: Did this.

adomasbaliuka:
@joneugster LGTM

Mathlib/GroupTheory/GroupExtension/Defs.lean
```diff
@@ -22,14 +22,14 @@ such as splittings and equivalences.
...
+      ↗ E  ↘
 1 → N   ↓    G → 1
-      ↘︎ E' ↗︎️
+      ↘ E' ↗
 
 For additive groups:
-      ↗︎ E  ↘
+      ↗ E  ↘
 0 → N   ↓    G → 0
-      ↘︎ E' ↗︎️
```
eric-wieser, line 32: I think these are genuinely intended to be non-breaking; the right solution is to put this in a code block.

Mathlib/Tactic/Widget/Calc.lean
```diff
@@ -116,7 +116,7 @@ def suggestSteps (pos : Array Lean.SubExpr.GoalsLocation) (goalType : Expr) (par
 @[server_rpc_method]
 def CalcPanel.rpc := mkSelectionPanelRPC suggestSteps
   "Please select subterms."
-  "Calc 🔍"
+  "Calc 🔍️"
```
eric-wieser, line 119: These are also intended to be non-breaking; I think using the unicode escape would be appropriate here.

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -349,12 +427,290 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
...
+/--
+Symbols with VSCode extension abbreviation as of Aug. 28, 2024
+
+Taken from abbreviations.json in
+github.com/leanprover/vscode-lean4/blob/97d7d8c382d1549c18b66cd99ab4df0b6634c8f1
+
+Obtained using Julia code:
+```julia
+filter(!isascii, unique( <all text in JSON file> )) |> repr
+```
```
eric-wieser, line 451: Let's do this with Lean!

Mathlib/Tactic/Linter/TextBased.lean
```diff
@@ -349,12 +427,290 @@ def isImportsOnlyFile (lines : Array String) : Bool :=
 
 end
 
+
+namespace unicodeLinter
```
eric-wieser, line 431: This is really quite long; can it go in its own file?

Mathlib/GroupTheory/GroupExtension/Defs.lean
joneugster: Sorry these are about the text-variant-selector of the arrows characters, i.e. there were invisible `\uFE0E` following the diagonal arrows to prevent them form being displayed as emojis.

There are no spaces changed here.

If they are supposed to have the variant-selector (which is a fair choice) then, they can be added to the list of text-symbols defined here and the selectors will be added back automatically by `lake exe lint-style --fix`

Mathlib/Tactic/Widget/Calc.lean
joneugster: same here, there is no space change, but the magnifying glass received a emoji-variant-selector because its specified in the list of emojis.

Mathlib/Tactic/Linter/TextBased.lean
joneugster: absolutely! I'll move it to `Mathlib/Tactic/Linter/Unicode.lean`

Mathlib/Tactic/Widget/Calc.lean
joneugster: For context, there are currently no non-breaking spaces in any parts of mathlib except in docstrings about condensed mathematics & cat. theory because Dagur's IDE automatically adds them after each closing \`

And the dependency PRs fixing whitespace are mainly because Dagur's quite busy and commiting a lot :D

Mathlib/Tactic/Linter/TextBased.lean
adomasbaliuka: a script that does this was added in #36773

PR #16216, opened 2024-08-28
feat(RingTheory): isotypic components

jxjwan:
Added isotypic components to RingTheory, added lemmas in SimpleModule that aid in some proofs

---

### DIFF
--- a/Mathlib/Order/CompactlyGenerated/Basic.lean
+++ b/Mathlib/Order/CompactlyGenerated/Basic.lean
@@ -592,6 +592,12 @@ theorem exists_setIndependent_isCompl_sSup_atoms (h : sSup { a : α | IsAtom a }
     · exact s_atoms x hx
     · exact ha
 
+proof_wanted exists_setIndependent_isCompl_sSup_sub_atoms {α : Type*} [CompleteLattice α]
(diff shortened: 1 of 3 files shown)


loefflerd:
I fear that this PR has gone un-noticed by mathlib reviewers because it's made from a fork of Mathlib, not from a branch within the main mathlib repository. This unfortunately means our build-test script won't run on it, so it won't get flagged as "passes tests", hence it won't show up in the review queue.

Can you make a new PR with the same content from a branch inside the mathlib repo? Anyone who wants to contribute can get push access to non-master mathlib branches by asking on the zulip chat.

alreadydone:
The new PR is at #18646. My general impression of the PR(s) is that there's a lot of duplicated definitions, i.e. equivalent or similar concepts that become multiple definitions. I think we should fix one definition for each concept and prove equivalences as lemmas. It would be helpful to actually prove Wedderburn--Artin to see what defs/API lead to the shortest path.

grunweg:
Since you have submitted a new PR from a mathlib branch, let me go ahead and close this PR, to keep the review queue useful. Thanks for your PR, I hope the new version can make it into mathlib!

PR #16217, opened 2024-08-28
[Merged by Bors] - feat(Tactic/Algebraize): add algebraize tactic

callesonne:
This PR adds the `algebraize` tactic, which automatically adds `Algebra`, and if possible `IsScalarTower` instances corresponding to `RingHom`s given as arguments to the tactic.

It will also search through the local context for `RingHom` properties of those `RingHom`s, and add the corresponding `Algebra` properties. The `RingHom` properties which have a corresponding `Algebra` property should be tagged using the `algebraize` attribute, which is also given in this PR. That attribute takes a name of a the corresponding `Algebra` property as an attribute, or of a constructor for such a property (this is needed when the proof of the corresponding `Algebra` property is not just given by a term the `RingHom` property).

For examples of its usage, see the files in `RingTheory/` which are also modified in this PR.

Most of this tactic was created during the AIM workshop "Formalizing algebraic geometry" in June 2024.

Co-authored by: Johan Commelin, Nick Kuhn, Arend Mellendijk, Christian Merten, Adam Topaz
(there might have been more, please let me know if you are not in this list!)

---


joneugster (commented):
Superficially this looks very neat! I'll have a closer look in the next few days.

Meanwhile, could you add a test file `test/algebraize.lean` where you use some `#guard_hyps` to test the tactic.

callesonne:
Thanks for the suggestion! Because of this, I found an error in tagging `IsIntegral` with the algebraize attribute which I fixed now :)

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+this function adds the instance `Algebra A B` to the context (if it does not already exist).
+
+This function also requries the type of `f`, given by the parameter `ft`. The reason this is done
+(even though `ft` can be inferred from `f`) is to avoid recomputing `ft` in the `algebraize` tactic,
+as when `algebraize` calls `addAlgebraInstanceFromRingHom` it has already computed `ft`. -/
+def addAlgebraInstanceFromRingHom (f ft : Expr) : TacticM Unit := withMainContext do
+  let (_, l) := ft.getAppFnArgs
+  -- The type of the corresponding algebra instance
+  let alg ← mkAppOptM `Algebra #[l[0]!, l[1]!, none, none]
+  try let _ ← synthInstance alg -- If the instance already exists, we do not do anything
```
callesonne, line 136: @joneugster I just had the following thought: Is this too expensive? Should I maybe instead loop through the local context and check that nothing already has this type?

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+Copyright (c) 2024 Calle Sönne. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Johan Commelin, Nick Kuhn, Arend Mellendijk, Christian Merten, Calle Sönne, Adam Topaz
+-/
+
+import Mathlib.Algebra.Algebra.Tower
+
+/-!
+
+## Algebraize tactic
```
joneugster, line 11:
```suggestion
# Algebraize tactic
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+
+This file defines the `algebraize` tactic. The basic functionality of this tactic is to
+automatically add `Algebra` instances given `RingHom`s. For example, `algebraize f g` where
+`f : A →+* B` and `g : B →+* C` are `RinHom`'s, will add the instances `Algebra A B` and
+`Algebra B C` corresponding to these `RingHom`s.
+
+# Further functionality
```
joneugster, line 18:
```suggestion
## Further functionality
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+add the instance `IsScalarTower A B C` if possible.
+
+After having added suitable `Algebra` and `IsScalarTower` instances, the tactic will search through
+the local context for `RingHom` properties that can be converted to properties of the corresponding
+`Algebra`. For example, given `f : A →+* B` and `hf : f.FiniteType`, then `algebraize f` will add
+the instance `Algebra A B` and the corresponding property `Algebra.FiniteType A B`. The tactic knows
+which `RingHom` properties have a corresponding `Algebra` property through the `algebraize`
+attribute.
+
+# Algebraize attribute
```
joneugster, line 30:
```suggestion
## Algebraize attribute
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+  Module.Finite A B
+```
+An example with a constructor as parameter:
+```
+@[algebraize Algebra.Flat.out]
+class RingHom.Flat {R : Type u} {S : Type v} [CommRing R] [CommRing S] (f : R →+* S) : Prop where
+  out : f.toAlgebra.Flat := by infer_instance
+```
+
+# Algebraize'
```
joneugster, line 69:
```suggestion
## Algebraize'
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+then constructs the `Algebra` property by applying the constructor to the `RingHom` property.
+
+Finally, if no argument is provided to the `algebraize` attribute, it is assumed that the tagged
+declaration has name `RingHom.Property` and that the corresponding `Algebra` property has name
+`Algebra.Property`. The attribute then returns `Algebra.Property` (so assume case 1 above).
+
+Here are three examples of properties tagged with the `algebraize` attribute:
+```
+@[algebraize]
+def FiniteType (f : A →+* B) : Prop :=
```
joneugster, line 52:
```suggestion
def RingHom.FiniteType (f : A →+* B) : Prop :=
```

Maybe worth writing out the fully-qualified name in the example for readability.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+  @Algebra.FiniteType A B _ _ f.toAlgebra
+```
+An example with a parameter (where the `Algebra` does not have the expected name):
+```
+@[algebraize Module.Finite]
+def Finite (f : A →+* B) : Prop :=
```
joneugster, line 58:
```suggestion
def RingHom.Finite (f : A →+* B) : Prop :=
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+`RingHom` and the `Algebra` property are definitionally the same, and the tactic will construct the
+`Algebra` property by giving the `RingHom` property as a term.
+
+2. A constructor for the `Algebra` property. In this case it is assumed that the `RingHom` property
+is the last argument of the constructor, and that no other explicit argument is needed. The tactic
+then constructs the `Algebra` property by applying the constructor to the `RingHom` property.
+
+Finally, if no argument is provided to the `algebraize` attribute, it is assumed that the tagged
+declaration has name `RingHom.Property` and that the corresponding `Algebra` property has name
+`Algebra.Property`. The attribute then returns `Algebra.Property` (so assume case 1 above).
```
joneugster, line 47: > [...] Using an (optional) parameter, it will also generate a `Name` of a declaration which
> will help the tactic to access the corresponding `Algebra` property.
> 
> There are two cases for what declaration corresponding to this `Name` can be.
> 1. An inductive type (i.e. the `Algebra` property itself), in this case it is assumed that the
> `RingHom` and the `Algebra` property are definitionally the same, and the tactic will construct the
> `Algebra` property by giving the `RingHom` property as a term.
> 
> 2. A constructor for the `Algebra` property. In this case it is assumed that the `RingHom` property
> is the last argument of the constructor, and that no other explicit argument is needed. The tactic
> then constructs the `Algebra` property by applying the constructor to the `RingHom` property.
> 
> Finally, if no argument is provided to the `algebraize` attribute, it is assumed that the tagged
> declaration has name `RingHom.Property` and that the corresponding `Algebra` property has name
> `Algebra.Property`. The attribute then returns `Algebra.Property` (so assume case 1 above).


I find this a bit hard to read, especially the first sentense.

Maybe

```suggestion
The `algebraize` attribute is used to tag `RingHom` properties that can be converted to `Algebra`
properties. It assumes that the tagged
declaration has a name of the form `RingHom.Property` and that the corresponding `Algebra` property has the name `Algebra.Property`.

If not, the `Name` of the corresponding algebra property can be provided as optional argument. The specified declaration should be one of the following:

1. An inductive type (i.e. the `Algebra` property itself), in this case it is assumed that the
`RingHom` and the `Algebra` property are definitionally the same, and the tactic will construct the
`Algebra` property by giving the `RingHom` property as a term.

2. A constructor for the `Algebra` property. In this case it is assumed that the `RingHom` property
is the last argument of the constructor, and that no other explicit argument is needed. The tactic
then constructs the `Algebra` property by applying the constructor to the `RingHom` property.
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+```
+@[algebraize Algebra.Flat.out]
+class RingHom.Flat {R : Type u} {S : Type v} [CommRing R] [CommRing S] (f : R →+* S) : Prop where
+  out : f.toAlgebra.Flat := by infer_instance
+```
+
+# Algebraize'
+
+To avoid searching through the local context and adding corresponding `Algebra` properties, use
+`algebraize'` which only adds `Algebra` and `IsScalarTower` instances.
```
joneugster, line 72: Might also reformulate this, did not understand it immediately.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+    return [mvar]
+
+/-- Given an expression `g.comp f` which is the composition of two `RingHom`s, this function adds
+the instance `IsScalarTower A B C` to the context (if it does not already exist). -/
+def addIsScalarTowerInstanceFromRingHomComp (fn : Expr) : TacticM Unit := withMainContext do
+  let (_, l) := fn.getAppFnArgs
+  let tower ← mkAppOptM `IsScalarTower #[l[0]!, l[1]!, l[2]!, none, none, none]
+  try let _ ← synthInstance tower -- If the instance already exists, we do not do anything
+  catch _ => liftMetaTactic fun mvarid => do
+    let nm ← mkFreshUserName `scalarTowerInst
```
joneugster, line 150:
```suggestion
    let nm ← mkFreshBinderNameForTactic `scalarTowerInst
```

This ensures that one can use `set_option tactic.hygienic false`. (and also in the other 3 places this is used)

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,264 @@
...
+/-- Given an expression `f` of type `RingHom A B` where `A` and `B` are commutative semirings,
+this function adds the instance `Algebra A B` to the context (if it does not already exist).
+
+This function also requries the type of `f`, given by the parameter `ft`. The reason this is done
+(even though `ft` can be inferred from `f`) is to avoid recomputing `ft` in the `algebraize` tactic,
+as when `algebraize` calls `addAlgebraInstanceFromRingHom` it has already computed `ft`. -/
+def addAlgebraInstanceFromRingHom (f ft : Expr) : TacticM Unit := withMainContext do
+  let (_, l) := ft.getAppFnArgs
+  -- The type of the corresponding algebra instance
+  let alg ← mkAppOptM `Algebra #[l[0]!, l[1]!, none, none]
```
joneugster, line 135:
```suggestion
  let alg ← mkAppOptM ``Algebra #[l[0]!, l[1]!, none, none]
```
Any declaration which already exists can be marked with double ticks: ` ``Algebra `. This way Lean checks it exists and complains if somebody renamed it.

joneugster (commented):
I'll have to finish the review after lunch, but here are already some comments :)

Mathlib/Tactic/Algebraize.lean
callesonne: This is great, thanks a lot!

callesonne:
Thanks a lot for the review so far! I've added your suggestions :)

Mathlib/Tactic/Algebraize.lean
callesonne: I meant that the typical case is that this will fail to synthesize, which might still be expensive? Although I guess that it could be synthesized whilst not being a local hypothesis, so I guess the above check has to be done.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,262 @@
...
+    let (_, mvar) ← mvar.intro1P
+    return [mvar]
+
+/-- This function takes an array of expressions `t`, all of which are assumed to be `RingHom`'s,
+and searches through the local context to find any additional properties of these `RingHoms`, after
+which it tries to add the corresponding `Algebra` properties to the context. It only looks for
+properties that have been tagged with the `algebraize` attribute, and uses this tag to find the
+corresponding `Algebra` property. -/
+def searchContext (t : Array Expr) : TacticM Unit := withMainContext do
+  let ctx ← MonadLCtx.getLCtx
```
joneugster, line 168:
```suggestion
  let ctx ← getLCtx
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,262 @@
...
+  ctx.forM fun decl => do
+    if decl.isImplementationDetail then return
+    let (nm, args) := decl.type.getAppFnArgs
+    -- Check if the type of the current hypothesis has been tagged with the `algebraize` attribute
+    match Attr.algebraizeAttr.getParam? (← getEnv) nm with
+    -- If it has, `p` will be the name of the corresponding `Algebra` property (or a constructor)
+    | some p =>
+      -- The last argument is the `RingHom` property is assumed to be `f`
```
joneugster, line 176:
```suggestion
      -- The last argument of the `RingHom` property is assumed to be `f`
```

joneugster:
!bench

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,262 @@
...
+      | _ => continue
+
+    -- Search through the local context to find other instances of algebraize
+    searchContext t
+
+/-- Version of `algebraize`, which only adds `Algebra` instances and `IsScalarTower` instances. -/
+syntax "algebraize'" (ppSpace colGt term:max)* : tactic
+
+elab_rules : tactic
+  | `(tactic| algebraize' $[$t:term]*) => do
```
joneugster, line 245: I think after fixing a few stylistic things this would be ready to merge. Thank you for the cool contribution!

One final question, since you end up using `algebraize'` quite often, I wonder if the two versions should be named `algebraize!` and `algebraize` (instead of `algebraize` and `algebraize'`).
Maybe that could be a question for Zulip or for @adamtopaz / @jcommelin?

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,262 @@
...
+    return [mvar]
+
+/-- This function takes an array of expressions `t`, all of which are assumed to be `RingHom`'s,
+and searches through the local context to find any additional properties of these `RingHoms`, after
+which it tries to add the corresponding `Algebra` properties to the context. It only looks for
+properties that have been tagged with the `algebraize` attribute, and uses this tag to find the
+corresponding `Algebra` property. -/
+def searchContext (t : Array Expr) : TacticM Unit := withMainContext do
+  let ctx ← MonadLCtx.getLCtx
+  ctx.forM fun decl => do
```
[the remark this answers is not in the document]
callesonne: It seems not! I think for this part I just copied the metaprogramming book, but the code seems to break when I change this. Here `.forM` is `LocalContext.forM` so I guess this is more specific than what one would get with `for decl in ctx`? I'm not sure though.

Mathlib/Tactic/Algebraize.lean
callesonne: I think that's a good idea, but I can ask in Zulip also.

jcommelin: Could you please explain the difference between the two in a bit more detail?

callesonne: @jcommelin the only difference is that `algebraize'` does not try to add algebra properties corresponding to ringhom properties that are available as hypothesis. I noticed that there were some statements where such properties where available, but they didn't need to be algebraized, and of course when there are no ringhom properties available one does not need to search for them.

I guess peformance wise it might be a neglegible improvement, but I still thought it seemed unecessary to peform this search when it was not needed.

callesonne: and thank you @joneugster for the review & suggestions!

jcommelin: I see. I do wonder whether it's worth it from a UX point of view.
Would it be possible to turn this into some boolean flag that can be passed as config option to `algebraize`?

I'm also fine with having them both as is, and then we can evaluate in a few months to see whether `algebraize'` is used much in practice.

joneugster: I actually like the config idea. Here is a snippet how other tactics like `simp`, `rw`, etc. do this:

```lean
namespace Algebraize

[...]

structure Config where
  searchContext : Bool := true
deriving Inhabited

declare_config_elab elabAlgebraizeConfig Algebraize.Config

end Algebraize

open Algebraize Lean.Parser.Tactic

/-- Tactic that, given `RingHom`s, adds the corresponding `Algebra` and (if possible)
`IsScalarTower` instances, as well as `Algebra` corresponding to `RingHom` properties available
as hypotheses.

Example: given `f : A →+* B` and `g : B →+* C`, and `hf : f.FiniteType`, `algebraize f g` will add
the instances `Algebra A B`, `Algebra B C`, and `Algebra.FiniteType A B`.

See the `algebraize` tag for instructions on what properties can be added. -/
syntax "algebraize" (config)? (ppSpace colGt term:max)* : tactic

elab_rules : tactic
  | `(tactic| algebraize $[$config]? $[$t:term]*) => do
    let cfg ← match config with
      | some conf => elabAlgebraizeConfig conf
      | none => pure default
    let t ← t.mapM fun i => Term.elabTerm i none
    -- We loop through the given terms and add algebra instances
    for f in t do
      let ft ← inferType f
      match ft.getAppFn with
      | Expr.const ``RingHom _ => addAlgebraInstanceFromRingHom f ft
      | _ => throwError m!"{f} is not of type `RingHom`"
    -- After having added the algebra instances we try to add scalar tower instances
    for f in t do
      match f.getAppFn with
      | Expr.const ``RingHom.comp _ =>
        try addIsScalarTowerInstanceFromRingHomComp f
        catch _ => continue
      | _ => continue

    -- Search through the local context to find other instances of algebraize
    if cfg.searchContext then
      searchContext t

```

and then it's used with `algebraize (config := { searchContext := false })` (or a less verbose version thereof). Maybe you can think about what the best name for the config option is.

test/algebraize.lean
joneugster, line 1: Sorry for keep adding remarks. Could you modify the tests maybe so that it imports not much besides `Mathlib.Tactic.Algebraize`? I just realised I couldn't use them to test the modified tactic because `algebraize` is already used somewhere in the imports of the test file.

Maybe you can just add `@[algebraize] def RingHom.testProperty` to test the functionality instead of testing it with real properties?

joneugster: (but that's not strictly necessary if you don't feel like it)

Mathlib/Tactic/Algebraize.lean
joneugster: > Maybe what I can do, is use the config setup that you gave above, and then define algebraize' to be just a macro / abbrevation for algebraize (config := { searchContext := false })?

Let's to that, that sounds like a good suggestion!

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,261 @@
...
+`IsScalarTower` instances, as well as `Algebra` corresponding to `RingHom` properties available
+as hypotheses.
+
+Example: given `f : A →+* B` and `g : B →+* C`, and `hf : f.FiniteType`, `algebraize f g` will add
+the instances `Algebra A B`, `Algebra B C`, and `Algebra.FiniteType A B`.
+
+See the `algebraize` tag for instructions on what properties can be added.
+
+The tactic also comes with a configuration option `searchContext` which, if set to `true` (default).
+This determines whether the tactic should search through the local context for `RingHom` properties
```
joneugster, line 227:
```suggestion
The tactic also comes with a configuration option `searchContext`. If set to `true` (default), the tactic searches through the local context for `RingHom` properties
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,261 @@
...
+that can be converted to `Algebra` properties. The macro `algebraize'` calls `algebraize` with
+`searchContext := false`, so in other words it only adds `Algebra` and `IsScalarTower` instances. -/
```
joneugster, line 229:
```suggestion
that can be converted to `Algebra` properties. The macro `algebraize'` calls `algebraize (config := {searchContext := false})`, so in other words it only adds `Algebra` and `IsScalarTower` instances. -/
```

joneugster (approved):
Thank you for this contribution!

I left two wording comments, but I think I can directly delegate this

~~maintainer merge~~ (I'll do this again after CI passes)

callesonne:
@joneugster this is ready again I think, and thanks again for the review!

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+        ← mkAppOptM ``algebraMap #[l[0]!, l[1]!, none, none, none]]])
+    -- Note: this could fail, but then `algebraize` will just continue, and won't add this instance
+    h.mvarId!.refl
+    let mvar ← mvarid.define nm tower
+      (← mkAppOptM ``IsScalarTower.of_algebraMap_eq'
+        #[l[0]!, l[1]!, l[2]!, none, none, none, none, none, none, h])
+    let (_, mvar) ← mvar.intro1P
+    return [mvar]
+
+/-- This function takes an array of expressions `t`, all of which are assumed to be `RingHom`'s,
```
joneugster, line 162:
```suggestion
/-- This function takes an array of expressions `t`, all of which are assumed to be `RingHom`s,
```

Mathlib/Tactic/Algebraize.lean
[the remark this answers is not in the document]
joneugster: ah interesting, thanks for checking

test/algebraize.lean
```diff
@@ -0,0 +1,97 @@
...
+  algebraize f
+  guard_hyp algebraizeInst : Module.testProperty2 A B
+  trivial
+
+example (A B : Type*) [CommRing A] [CommRing B] (f : A →+* B) (hf : f.testProperty3) : True := by
+  algebraize f
+  guard_hyp algebraizeInst : Algebra.testProperty3 A B
+  trivial
+
+set_option tactic.hygienic false in
```
joneugster, line 86:
```suggestion
```
superfluous, right?

joneugster (approved):
Thanks for the PR!

joneugster:
maintainer merge

Mathlib/RingTheory/FinitePresentation.lean
```diff
@@ -415,11 +416,9 @@ theorem id : FinitePresentation (RingHom.id A) :=
 variable {A}
 
 theorem comp_surjective {f : A →+* B} {g : B →+* C} (hf : f.FinitePresentation) (hg : Surjective g)
-    (hker : g.ker.FG) : (g.comp f).FinitePresentation :=
-  letI := f.toAlgebra
-  letI := (g.comp f).toAlgebra
-  letI : Algebra.FinitePresentation A B := hf
-  Algebra.FinitePresentation.of_surjective
+    (hker : g.ker.FG) : (g.comp f).FinitePresentation := by
+  algebraize f (g.comp f)
```
semorrison, line 420: I'm not super excited about it taking a sequence of terms, with no separators. Seems like an invitation for confusing parsing. Would it be a problem to *either* take a single terms, or to take a list `[f, g, (g.comp f)]`?

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+To avoid searching through the local context and adding corresponding `Algebra` properties, use
+`algebraize'` which only adds `Algebra` and `IsScalarTower` instances.
+
+-/
+
+open Lean Elab Tactic Term Meta
+
+namespace Lean.Attr
+
+/-- Function that extracts the name of the corresponding `Algebra` property from the a `RingHom`
```
semorrison, line 78:
```suggestion
/-- Function that extracts the name of the corresponding `Algebra` property from a `RingHom`
```

Mathlib/RingTheory/Finiteness.lean
```diff
@@ -776,20 +778,11 @@ theorem of_surjective (f : A →+* B) (hf : Surjective f) : f.Finite :=
...
+  algebraize f g (g.comp f)
   exact Module.Finite.trans B C
 
 theorem of_comp_finite {f : A →+* B} {g : B →+* C} (h : (g.comp f).Finite) : g.Finite := by
-  letI := f.toAlgebra
-  letI := g.toAlgebra
-  letI := (g.comp f).toAlgebra
-  letI : IsScalarTower A B C := RestrictScalars.isScalarTower A B C
-  letI : Module.Finite A C := h
+  algebraize f g (g.comp f)
```
semorrison, line 785: Why do we need to write `g.comp f` separately? Surely the tactic could identify composable pairs of specified RingHoms automatically?

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+
+import Mathlib.Algebra.Algebra.Tower
+
+/-!
+
+## Algebraize tactic
+
+This file defines the `algebraize` tactic. The basic functionality of this tactic is to
+automatically add `Algebra` instances given `RingHom`s. For example, `algebraize f g` where
+`f : A →+* B` and `g : B →+* C` are `RinHom`s, will add the instances `Algebra A B` and
```
semorrison, line 15:
```suggestion
`f : A →+* B` and `g : B →+* C` are `RingHom`s, will add the instances `Algebra A B` and
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+      | Expr.const ``RingHom.comp _ =>
+        try addIsScalarTowerInstanceFromRingHomComp f
+        catch _ => continue
+      | _ => continue
+
+    -- Search through the local context to find other instances of algebraize
+    if cfg.searchContext then
+      searchContext t
+
+/-- Version of `algebraize`, which only adds `Algebra` instances and `IsScalarTower` instances. -/
```
semorrison, line 259:
```suggestion
/-- Version of `algebraize`, which only adds `Algebra` instances and `IsScalarTower` instances,
but not ...... -/
```

semorrison:
It's reasonable to reject my suggestions above. :-) But I think they are good!

Mathlib/RingTheory/FinitePresentation.lean
joneugster: I don't know if there is a standard on this.

We could definitely do `algebraize [f, g]` following the example of [`Lean.Parser.Tactic.rwRuleSeq`](https://leanprover-community.github.io/mathlib4_docs/Init/Tactics.html#Lean.Parser.Tactic.rwRuleSeq)

`simp` even does this directly in-place:

```lean
syntax (name := simp) "simp" (config)? (discharger)? (&" only")?
  (" [" withoutPosition((simpStar <|> simpErase <|> simpLemma),*,?) "]")? (location)? : tactic
```

It seems reasonable to do it the same way in that case.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+```
+@[algebraize Algebra.Flat.out]
+class RingHom.Flat {R : Type u} {S : Type v} [CommRing R] [CommRing S] (f : R →+* S) : Prop where
+  out : f.toAlgebra.Flat := by infer_instance
+```
+
+## Algebraize'
+
+To avoid searching through the local context and adding corresponding `Algebra` properties, use
+`algebraize'` which only adds `Algebra` and `IsScalarTower` instances.
```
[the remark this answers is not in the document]
joneugster:
```
Version of algebraize, which only adds Algebra instances and IsScalarTower instances.

macro_rules
  | `(tactic| algebraize' $[$t:term]*) =>
    `(tactic| algebraize (config := {searchContext := false}) $t*)
```

Since `algebraize'` is literally just an abbreviation for `algebraize (config := {searchContext := false})`  I would argue it would be fine to leave the macro in as is. If it's any more verbose, one could directly use the version with the `config` instead.

To throw in another suggestion, one could rename `searchContext` to one of `properties`, `findProperties`, or `searchProperties`. Especially `algebraize (config := {properties := false})` isn't that verbose anymore.

In any case, I think @callesonne it's reasonable if you have another think about the naming of macro+config option (maybe with your co-authors) and then decide on what you like best :)

Mathlib/RingTheory/Finiteness.lean
joneugster: This would be nice to implement, but @callesonne I think it is also fine if you add this as

```
## Future work

* If source/target coincide, automatically add `(g.comp f)` when calling `algebraize f g`
```

to the module docstring if you don't feel like looking at how to implement it now.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+property.",
+    getParam := algebraizeGetParam }
+
+end Lean.Attr
+
+namespace Mathlib.Tactic
+
+namespace Algebraize
+
+/-- Given an expression `f` of type `RingHom A B` where `A` and `B` are commutative semirings,
```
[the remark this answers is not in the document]
joneugster: I believe the tactic mainly uses [RingHom.toAlgebra](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Algebra/Defs.html#RingHom.toAlgebra):

```lean
def RingHom.toAlgebra {A B} [CommSemiring A] [CommSemiring B] (i : A →+* B) : Algebra A B
```

which assumes that both `A` and `B` are commutative.

Mathlib/RingTheory/Finiteness.lean
callesonne: @semorrison we discussed this during the workshop, but an issue that comes up is that say you have three morphisms `f : A -> B` and `g, h : B -> C`. Then, if adding the algebra instance corresponding to the compositions is automatic, you would add two algebra instances `Algebra A C` which becomes a problem. So that's why we decided to force the user to supply the compositions themselves.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+
+This file defines the `algebraize` tactic. The basic functionality of this tactic is to
+automatically add `Algebra` instances given `RingHom`s. For example, `algebraize f g` where
+`f : A →+* B` and `g : B →+* C` are `RingHom`s, will add the instances `Algebra A B` and
+`Algebra B C` corresponding to these `RingHom`s.
+
+## Further functionality
+
+When given a composition of `RingHom`s, e.g. `algebraize g.comp f`, the tactic will also try to
+add the instance `IsScalarTower A B C` if possible.
```
chrisflav, line 21:
```suggestion
When given a composition of `RingHom`s, e.g. `algebraize (g.comp f)`, the tactic will also try to
add the instance `IsScalarTower A B C` if possible.
```

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,263 @@
...
+@[algebraize]
+def RingHom.FiniteType (f : A →+* B) : Prop :=
+  @Algebra.FiniteType A B _ _ f.toAlgebra
+```
+An example when the `Name` is provided (as the `Algebra` does not have the expected name):
+```
+@[algebraize Module.Finite]
+def Module.Finite (f : A →+* B) : Prop :=
+  letI : Algebra A B := f.toAlgebra
+  Module.Finite A B
```
chrisflav, line 58:
```suggestion
@[algebraize Module.Finite]
def RingHom.Finite (f : A →+* B) : Prop :=
  letI : Algebra A B := f.toAlgebra
  Module.Finite A B
```

Mathlib/Tactic/Algebraize.lean
jcommelin: @callesonne Please expand this docstring to explain what `algebraize'` does *not* do compared to `algebraize`.

Mathlib/RingTheory/FinitePresentation.lean
jcommelin: This seems like a good idea.

Mathlib/Tactic/Algebraize.lean
callesonne: Yes will do, sorry I was away (and needed to prepare a talk) for a conference!

Mathlib/Tactic/Algebraize.lean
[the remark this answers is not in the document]
callesonne: @joneugster I really like the config name option suggestion!

How about `algebraize_only` instead of `algebraize'`, or somehow making the syntax `algebraize only` work?

Mathlib/RingTheory/FinitePresentation.lean
callesonne: I agree, I like the list suggestion! Will try to incorporate it soon.

Mathlib/Tactic/Algebraize.lean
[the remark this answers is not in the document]
joneugster: My suggestion is to use the suggestion `algebraize_only`. That sounds reasonable.

With `algebraize only` I would argue that `algebraize only [f, g, f.comp g]` somehow looks unexpected. For me the `only`-keyword would suggest that whatever follows is a subset of all tagged declarations, i.e. I'd rather expect `algebraize only [RingHom.Finite]` to work, which might not integrate well: `algebraize only [RingHom.Finite] at f g (f.comp g)` would be uniform with other tactics, but then we're back at the sequence of terms... (on a techincal level, it wouldn't be difficult: one could directly [copy the syntax from `simp`](https://github.com/leanprover/lean4/blob/e9e858a4484905a0bfe97c4f05c3924ead02eed8/src/Init/Tactics.lean#L570-L571))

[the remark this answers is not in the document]
callesonne: That would be very cool to implement! @kim-em what do you think about this `algebraize only [...] at ...` syntax as in Jon's new suggestion? Or maybe I just call the macro `algebraize_only` for now, so that some version of this reaches mathlib, and then work on `algebraize only` after.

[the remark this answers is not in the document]
joneugster: I think the problem is that usually the `at ...` is essentially a sequence of `ident`s, but here it would be an undeterminated sequence of `term`s, which is much more general and the reason Kim was worried about parsing errors (if I understand correctly). So this would require thinking again about the way `f.comp g` is handled. This might be out-of-scope for this PR.

For the sake of preserving your own efforts, I would say just use `algebraize_only` here and work this out in a follow-up PR if you have the motivation. We already kept adding more and more, and I think it would be better to get a first version merged and then have improvements in follow-up PRs.

joneugster:
@callesonne I took the liberty to push the few changes discussed above. This include

- rename `algebraize'` to `algebraize_only`
- rename config option from `searchConfig` to `properties`
- change to `algebraize [f, g, g.comp f]` syntax.

The last one is implemented in this commit: [1a2673e](https://github.com/leanprover-community/mathlib4/pull/16217/commits/1a2673e824059a5bc8f9455fe17ce0997f154ef7). Would you have time to look at this commit and decide whether you like this implementation?

And also have a look if there is anything you would like to cleanup before we try to get this merged. (note: follow-up PRs improving your tactic will always be welcome of course!)

Mathlib/RingTheory/FinitePresentation.lean
joneugster: Implemented this :)

Mathlib/Tactic/Algebraize.lean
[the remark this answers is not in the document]
joneugster: switched to `algebraize_only` and config options`properties`. for now

joneugster:
!bench

Mathlib/Tactic/Algebraize.lean
joneugster: done

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,283 @@
...
+      we assume that the `RingHom` property is the last argument of the constructor (and that
+      this is all we need to supply explicitly). -/
+      else
+        let pargs := pargs.set! (n - 1) decl.toExpr
+        let val ← mkAppOptM p pargs
+        liftMetaTactic fun mvarid => do
+          let nm ← mkFreshBinderNameForTactic `algebraizeInst
+          -- let mvar ← mvarid.define nm _ val -- TODO: How to get the type of the target of `p`?
+          -- let (_, mvar) ← mvar.intro1P
+          let (_, mvar) ← mvarid.note nm val
```
joneugster, line 206: I have a hunch that some of the regression could come from the fact that we are using `MVarId.note` (`have` decl) instead of `MVarId.define` (`let` decl) here... Does anybody know how to construct the expected type of `val`?(added template as comment)

joneugster:
!bench

Mathlib/Tactic/Algebraize.lean
joneugster: I moved my debugging commits to

- leanprover-community/mathlib4#17284

callesonne:
@joneugster thanks a lot for taking care of the final fixes & working on debugging the peformance! I am in the middle of moving flats, but should be able to assist with what remains from next week.

Mathlib/Tactic/Algebraize.lean
callesonne: Very odd that it has regressed, given that an earlier bench showed no significant changes.

Mathlib/Tactic/Algebraize.lean
```diff
@@ -0,0 +1,298 @@
...
+        let mut argsType := Array.mkArray (ctor.numParams) (none : Option Expr)
+        argsType := argsType.set! 0 args[0]!
+        argsType := argsType.set! 1 args[1]!
+        argsType := argsType.set! (ctor.numParams - 1) algInst
+        let tp := ← mkAppOptM ctor.induct argsType
+
+        unless (← synthInstance? tp).isSome do
+        liftMetaTactic fun mvarid => do
+          let nm ← mkFreshBinderNameForTactic `algebraizeInst
+          let mvar ← mvarid.define nm tp val
```
joneugster, line 218: I realised now that this change I made wouldn't be necessary: since all ring properties are always `Prop` (at least so far), I think it doesn't matter if one uses `mvarid.define` or `mvarid.note` in `addProperties` due to proof-irrelevance.

So this construction of the expected type above here could be removed again

Mathlib/Tactic/Algebraize.lean
joneugster: Somehow I think the regression boils down to the import change (which shake suggested). Undoing that everything is fine again.

joneugster:
!bench

Mathlib/Tactic/Algebraize.lean
jcommelin: I don't think this is a critical issue. So let's fix it in a follow-up PR. Because this tactic is going to be very helpful in the immediate future.

jcommelin (commented):
Thanks :tada:

bors merge

PR #16220, opened 2024-08-28
[Merged by Bors] - refactor(SetTheory/Cardinal/Ordinal): make `Cardinal.aleph'` an order iso

vihdzp:
We deprecate  `alephIdx.initialSeg`, `alephIdx`, `alephIdx.relIso`, `Aleph'.relIso`, and `aleph'Equiv`, which were just stating this idea in slightly different ways.

We also make `Cardinal.aleph` an order embedding.

---

- [x] depends on: #16222


Mathlib/SetTheory/Cardinal/Ordinal.lean
```diff
@@ -209,41 +225,46 @@ theorem aleph'_omega : aleph' ω = ℵ₀ :=
...
-  uncountable cardinal, and so on. -/
-def aleph (o : Ordinal) : Cardinal :=
-  aleph' (ω + o)
+/-- The `aleph` function gives the infinite cardinals listed by their ordinal index. `aleph 0 = ℵ₀`,
+`aleph 1 = succ ℵ₀` is the first uncountable cardinal, and so on.
+
+For a version including finite cardinals, see `Cardinal.aleph'`. -/
+def aleph : Ordinal ↪o Cardinal :=
+  OrderEmbedding.ofMapLEIff
+    (fun o => aleph' (ω + o)) fun _ _=> aleph'_le.trans (add_le_add_iff_left ω)
```
eric-wieser, line 240: Can you build this by composing `aleph'.toOrderEmbedding` with the order embedding that adds `ω`?

vihdzp: What's the name of the latter?

eric-wieser (approved):

Mathlib/SetTheory/Cardinal/Ordinal.lean
eric-wieser: I don't know, but you can loogle for it using the lemma that characterizes its application.

If it doesn't exist, you could build it inline in this proof.

vihdzp: Oh I found it, it's [`OrderEmbedding.addLeft`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/Monoid/Basic.html#OrderEmbedding.addLeft).

Mathlib/SetTheory/Cardinal/Ordinal.lean
eric-wieser, line 1: Can you fix the module docstring to not refer to `Cardinal.aleph/idx`, which look like a typo of the deprecated name?

eric-wieser (commented):
Since you requested their review,

bors d=@YaelDillies

Feel free to merge without review if you don't care much about this PR, Yael.

Mathlib/SetTheory/Cardinal/Ordinal.lean
vihdzp: I've updated the docstring, and I've also removed the obsolete part about `bit0` and `bit1`.

Mathlib/SetTheory/Cardinal/Ordinal.lean
```diff
@@ -73,6 +70,26 @@ theorem noMaxOrder {c} (h : ℵ₀ ≤ c) : NoMaxOrder c.ord.toType :=
 
 section aleph
 
+/-- The `aleph'` function gives the cardinals listed by their ordinal index. `aleph' n = n`,
+`aleph' ω = ω`, `aleph' (ω + 1) = succ ℵ₀`, etc.
```
YaelDillies, line 74:
```suggestion
`aleph' ω = ℵ₀`, `aleph' (ω + 1) = succ ℵ₀`, etc.
```
no?

YaelDillies:
The rest looks sensible. Thanks Eric for simplifying my job!

bors merge

PR #16222, opened 2024-08-28
[Merged by Bors] - feat(Order/*): `CovBy` and `OrderIso`

vihdzp:
We show that order isomorphisms map maximum elements, the covering relation, and successors.

---

### DIFF
--- a/Mathlib/Order/Cover.lean
+++ b/Mathlib/Order/Cover.lean
@@ -127,6 +127,13 @@ alias ⟨_, WCovBy.toDual⟩ := toDual_wcovBy_toDual_iff
 
 alias ⟨_, WCovBy.ofDual⟩ := ofDual_wcovBy_ofDual_iff
 
+theorem OrderIso.map_wcovBy {α β : Type*} [Preorder α] [Preorder β]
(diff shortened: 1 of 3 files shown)


Mathlib/Order/Cover.lean
```diff
@@ -127,6 +127,13 @@ alias ⟨_, WCovBy.toDual⟩ := toDual_wcovBy_toDual_iff
 
 alias ⟨_, WCovBy.ofDual⟩ := ofDual_wcovBy_ofDual_iff
 
+theorem OrderIso.map_wcovBy {α β : Type*} [Preorder α] [Preorder β]
+    (f : α ≃o β) {x y : α} (h : x ⩿ y) : f x ⩿ f y := by
```
YaelDillies, line 131: Should be an iff. Can you make the reverse implication a lemma for order embeddings (it might already exist)?

vihdzp: Done. Is the name `map_covBy` still appropriate for the iff version though?

YaelDillies (approved):
Thanks!

maintainer merge

mattrobball:
bors merge

PR #16230, opened 2024-08-28
[Merged by Bors] - feat: parser for Stacks Project Tags

adomani:
Introduce a parser for Stacks Projects Tags: it accepts 4-character sequences of digits and uppercase letter, giving meaningful errors for shorter/empty sequences.

See [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/stacks.20tags/near/465662710) for the issues with the previous implementation of the attribute and, in particular, parsing `0BR2` as a tag.

---

See [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Parser.20for.20digits.20and.20uppercase.20letters) for some of the parser.


eric-wieser:
Here's an untested simpler version:
```
initialize Lean.registerBuiltinAttribute {
  name := `stacks
  descr := "Apply a Stacks project tag to a theorem."
  add := fun decl stx _attrKind => Lean.withRef stx do
    let `(attr| stacks $[$tag]? $[$comment]?) := stx | throwUnsupportedSyntax
    let .some tag@⟨Syntax.atom _ str⟩ := tag | logWarning "Please, enter a Tag after `stacks`."
    if str.length != 4 then
      logWarningAt tag
        m!"Tag '{str}' is {str.length} characters long, but it should be 4 characters long"
    else if !str.all (fun c => c.isUpper || c.isDigit) then
      logWarningAt tag m!"Tag '{str}' should only consist of digits and uppercase letters"
    else
      addTagEntry decl str <| (comment.map (·.getString)).getD ""
}

```

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -43,22 +43,59 @@ It extends the `Tag` environment extension with the data `declName, tag, comment
...
 
 /-- The `stacks` attribute.
 Use it as `@[stacks TAG "Optional comment"]`.
-The `TAG` is mandatory.
+The `TAG` is mandatory and should be a sequence of 4 digits or uppercase letters.
 
 See the [Tags page](https://stacks.math.columbia.edu/tags) in the Stacks project for more details.
 -/
-syntax (name := stacks) "stacks " (stackTag)? (ppSpace str)? : attr
+syntax (name := stacks) "stacks " (stacksTag)? (ppSpace str)? : attr
```
mattrobball, line 98:
```suggestion
syntax (name := stacks) "stacks " stacksTag (ppSpace str)? : attr
```
I think we should require the stacks tag.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -43,22 +43,59 @@ It extends the `Tag` environment extension with the data `declName, tag, comment
...
-The syntax for a Stacks tag: it is an optional number followed by an optional identifier.
-This allows `044Q3` and `GH3F6` as possibilities.
--/
-declare_syntax_cat stackTag
+open Parser
+
+/-- `stacksTag` is the node kind of Stacks Project Tags: a sequence of digits and
+uppercase letters. -/
+abbrev stacksTagKind : SyntaxNodeKind := `stacksTag
+
```
mattrobball, line 51:
```suggestion

/-- Extract the tag if it well-formed -/
def Lean.TSyntax.validateStacksTag (stx : TSyntax stacksTagKind) : Option String :=
  match Syntax.isLit? stacksTagKind stx with
  | some val =>
    if val.all (fun c => c.isDigit || c.isUpper) && val.length == 4 then val
    else none
  | none => none
```

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -68,17 +105,18 @@ initialize Lean.registerBuiltinAttribute {
     -- that only digits and uppercase letter are present
     let tag := stx[1]
     match tag.getSubstring? with
-      | none => logWarning "Please, enter a Tag after `stacks`."
+      | none => logWarning "Please, enter a Tag after `stacks` ." -- this may be unreachable
```
mattrobball, line 108: Following Eric's lead, the following seems to work 
```lean
    let `(attr| stacks $tag $[$comment]?) := stx | throwUnsupportedSyntax
    let some str := tag.validateStacksTag |
      logWarning "Stacks Project tags should be 4 characters long and only consist of digits and uppercase letters."
    addTagEntry decl str <| (comment.map (·.getString)).getD "" 
```

Mathlib/Tactic/StacksAttribute.lean
mattrobball: You could probably also just throw the error here.

mattrobball:
```lean
def Lean.TSyntax.validateStacksTag (stx : TSyntax stacksTagKind) : CoreM String := do
  let missing := "Please, enter a Tag after `stacks`."
  let malformed :=
    "Stacks Project tags should be 4 characters and only consist of digits and uppercase letters."
  let some val := Syntax.isLit? stacksTagKind stx | throwError malformed
  if val.length == 0 then throwError missing
  else if val.all (fun c => c.isDigit || c.isUpper) && val.length == 4 then return val
  else throwError malformed
```

Mathlib/Tactic/StacksAttribute.lean
mattrobball: Actually, I am not sure if this change does anything?

Mathlib/Tactic/StacksAttribute.lean
mattrobball: With the `CoreM` version this is just 
```lean
    let `(attr| stacks $tag $[$comment]?) := stx | throwUnsupportedSyntax
    let str ← tag.validateStacksTag 
    addTagEntry decl str <| (comment.map (·.getString)).getD "" 
```

Mathlib/Tactic/StacksAttribute.lean
adomani: I *think* that it does nothing, since the empty stacksTag is allowed.  This probably also explains the unreachable code below.

mattrobball: You can get a slightly better parser error with 
```lean
syntax (name := stacks) "stacks " stacksTag !(ident) (ppSpace str)? : attr
```
which replaces "unexpected identifier" with "unexpected element" if you only use digits and uppercase letters in `stacksTag`. 

Parsers themselves have errors built and aesthetically validation *should* be built into the parser but I am not sure of the recipe to get custom error here.

mattrobball:
Here is a hacky(?) way to build a reasonable error into the parser 
```lean
def stacksTagFn : ParserFn := fun c s =>
  let i := s.pos
  let firstHalf := Substring.mk c.input 0 i |>.takeRightWhile (fun c => c.isDigit || c.isUpper)
  let secondHalf := Substring.mk c.input i c.input.endPos |>.takeWhile (fun c => c.isDigit || c.isUpper)
  let txt := firstHalf.toString ++ secondHalf.toString
  if txt.length != 4 then
    ParserState.mkError s "Stacks Project tags should be 4 characters and only consist of digits and uppercase letters."
  else
  let st := takeWhileFn (fun c => c.isDigit || c.isUpper) c s
  mkNodeToken stacksTagKind i c st
```

adomani:
Ok, so I merged Eric's suggestions for the attribute, with Matt's suggestions for the parser (with small modifications).

Now, all the checking is done by the parser, which seems to be better, and the error messages are also more aware of what went wrong with the parsing.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -95,29 +101,15 @@ The `TAG` is mandatory and should be a sequence of 4 digits or uppercase letters
...
-            m!"Tag '{str}' is {str.length} characters long, but it should be 4 characters long"
-        else if 2 ≤ (str.split (fun c => !(c.isUpper || c.isDigit))).length then
-          logWarningAt tag m!"Tag '{str}' should only consist of digits and uppercase letters"
-        else match stx with
-          | `(attr| stacks $_:stacksTag $comment:str) => addTagEntry decl str comment.getString
-          | `(attr| stacks $_:stacksTag) => addTagEntry decl str ""
-          | _ => throwUnsupportedSyntax
+  add := fun decl stx _attrKind => match stx with
+    | `(attr| stacks $tag $[$comment]?) =>
+      addTagEntry decl tag.raw[0].getAtomVal <| (comment.map (·.getString)).getD ""
```
mattrobball, line 111:
```suggestion
      addTagEntry decl (← tag.getStacksTag)  <| (comment.map (·.getString)).getD ""
```
This seems more type safe for the syntax.

mattrobball:
There is `runParserCategory` to capture parsing errors but perhaps that can be left for later

mattrobball:
Actually this works and can probably be refined. 
```lean
/-- error: <input>:1:0: Stacks tags are 4 characters of digits and uppercase letters. -/
#guard_msgs in
run_cmd do 
  let env ← getEnv
  let input := "A05"
  let p := stacksTagFn
  let ictx := mkInputContext input "<input>" 
  let s := p.run ictx { env, options := {} } (getTokenTable env) (mkParserState input)
  let err := if !s.allErrors.isEmpty then
      Except.error (s.toErrorMsg ictx)
    else if ictx.input.atEnd s.pos then
      Except.ok s.stxStack.back
    else
      Except.error ((s.mkError "end of input").toErrorMsg ictx)
  let env ← getEnv
  let att ← Lean.ofExcept <| err
```

mattrobball:
But still not sure if we need it here

mattrobball (approved):
Looks good to me. Maybe we wait a bit to see if Mario responds on Zulip before merging

mattrobball:
Can you update the PR description?

adomani:
Matt, thanks a lot!

I updated the PR description.

I do not think that it is necessary to actually check the boundaries of the parsing: maybe we can revisit this if someone finds ways of misusing the parser.  I have tested it locally and got the "correct" errors for everything that I tried.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -43,48 +43,83 @@ It extends the `Tag` environment extension with the data `declName, tag, comment
...
--/
-declare_syntax_cat stackTag
+open Parser
+
+/-- `stacksTag` is the node kind of Stacks Project Tags: a sequence of digits and
+uppercase letters. -/
+abbrev stacksTagKind : SyntaxNodeKind := `stacksTag
+
+/-- The main parser for Stacks Project Tags: it accepts any sequence of 4 digits or
+uppercase letters. -/
```
mattrobball, line 53: Oops. This should be updated.

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -43,48 +43,83 @@ It extends the `Tag` environment extension with the data `declName, tag, comment
...
+def stacksTagFn : ParserFn := fun c s =>
+  let i := s.pos
+  let tag :=
+    Substring.mk c.input i c.input.endPos |>.takeWhile (fun c => c.isDigit || c.isUpper) |>.toString
```
eric-wieser, line 57: I think you should take here while the character is alphanumeric, then do a second check for this condition, so that you can distinguish 4 invalid characters from 3 valid ones.

mattrobball: @eric-wieser can you spell out the branching logic you want here? Thanks.

mattrobball:
Instead of preparing for class tomorrow... (NB: I just recording it here not advocating for its inclusion)
```lean
def captureException (env : Environment) (s : ParserFn) (input : String) : Except String Syntax := 
  let ictx := mkInputContext input "<input>"
  let s := s.run ictx { env, options := {} } (getTokenTable env) (mkParserState input)
  if !s.allErrors.isEmpty then
    .error (s.toErrorMsg ictx)
  else if ictx.input.atEnd s.pos then
    .ok s.stxStack.back
  else
    .error ((s.mkError "end of input").toErrorMsg ictx)

/-- error: <input>:1:0: Stacks tags are 4 characters of digits and uppercase letters. -/
#guard_msgs in
run_cmd do 
  let _ ← Lean.ofExcept <| captureException (← getEnv) stacksTagFn "A05"

/-- error: <input>:1:0: No Stacks tag after `stacks`. -/
#guard_msgs in 
run_cmd do 
  let env ← getEnv 
  let _ ← Lean.ofExcept <| captureException env stacksTagFn "\"A04Q\""
```

Mathlib/Tactic/StacksAttribute.lean
adomani: The `alphanum` was how it was in a previous iteration and I'm happy to restore it!

eric-wieser: I pushed my suggestion

Mathlib/Tactic/StacksAttribute.lean
```diff
@@ -43,48 +43,87 @@ It extends the `Tag` environment extension with the data `declName, tag, comment
...
+uppercase letters. -/
+def stacksTagFn : ParserFn := fun c s =>
+  let i := s.pos
+  let s := takeWhileFn (fun c => c.isAlphanum) c s
+  if s.hasError then
+    s
+  else if s.pos == i then
+    ParserState.mkError s "stacks tag"
+  else
+    let tag := Substring.mk c.input i s.pos |>.toString
```
eric-wieser, line 62: I think this causes a caching issue: if I edit the file in vscode, the errors become outdated.

Mathlib/Tactic/StacksAttribute.lean
mattrobball: Yeah, I think that's an improvement.

Mathlib/Tactic/StacksAttribute.lean
eric-wieser: @david-christiansen, perhaps you can advise if this is something that is sensible to do in a parser, since you wrote many for Verso!

Mathlib/Tactic/StacksAttribute.lean
adomani: I like this and I am happy to see that the parser consumes any `alphanum` and then continues with further checks: I think that it really makes for a better user experience!

Mathlib/Tactic/StacksAttribute.lean
adomani: I tried this in VSCode and I was not able to get the caching issue: the parsing seemed to always be correct.

Regardless of whether the saved file was in a "correct" or "incorrect" state, editing (and not saving) would produce the correct parsing outcome.

semorrison:
Let's include Matthew's `captureException` testing framework. It's good to have, and great to have a model that can be used elsewhere.

mattrobball:
@eric-wieser threw it in with his changes (in case that is the reason for the awaiting author). Also it is just a minor change to `runParserCategory` and the existing test `string_gaps` in core. 

@semorrison if it looks good to you, I think it is fine to merge. We are waiting on some opinions but the parser seems reasonable compared to the ones in core and the tests fairly robust.

adomani:
I interpreted the `awaiting-author` as "let's add the guard_msgs for parsers in this PR.  However, I think that the parser may benefit from a more thorough review, so I opened #16305 that, on top of this PR, adds Matt's command and also the tests for the Stacks Tags.

This way, maybe the current PR can be merged and the tags used quickly and then the tests will follow.

Does that seem reasonable?

Mathlib/Tactic/StacksAttribute.lean
eric-wieser: This is already up to date, right?

mattrobball: Despite predating your change, I still think my comment holds. Since you push errors to the parser state with lower case letters, it won't make a syntax node. 

Thought perhaps this is quibbling over little. 

I think saying to the user that it accepts any alphanumeric is confusing to the user who sees an error from the parser with lower case letters

mattrobball: Ugh. I need to read. Never mind

mattrobball:
Been long enough

bors merge

PR #16239, opened 2024-08-29
feat(Geometry/Manifold): orientable manifolds

Rida-Hamadani:
This module defines orientable manifolds. [This](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Orientable.20Surface) Zulip topic is related.

---


Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,80 @@
...
+variable {H : Type*} [NormedAddCommGroup H] [NormedSpace ℝ H]
+
+section orientationPreservingGroupoid
+
+/--
+A map is orientation-preserving on a given set if it is differentiable and the determinant of its
+Jacobian is strictly positive on that set.
+-/
+def OrientationPreserving (f : H → H) (s : Set H) : Prop :=
+  ∀ x ∈ s,
```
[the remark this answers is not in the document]
grunweg: Also: I think you can remove the differentiability requirement, as `fderiv` has junk value `0` whenever the map is not differentiable. (Perhaps add a comment to that effect, if you find one helpful.)
That ought to simplify later proof, right?

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,80 @@
...
+    intro x _
+    simp only [Set.mem_univ, differentiableAt_id, fderiv_id, true_and]
+    rw [ContinuousLinearMap.id, ContinuousLinearMap.det, LinearMap.det_id]
+    exact zero_lt_one
+  locality {f u} _ h x hxu :=
+    have ⟨v, _, hxv, h⟩ := h x hxu
+    h x <| Set.mem_inter hxu hxv
+  congr {f g u} hu fg hf := by
+    intro x hx
+    have := Filter.eventuallyEq_of_mem (IsOpen.mem_nhds hu hx) fg
```
grunweg, line 61: This line and the next ones look like they could use dot notation, right?

grunweg (commented):
Thanks for this PR, I am excited to see somebody work on it! I have one comment, which could simplify your definition and proofs, and one follow-up request.

Right now, this is just a definition, without any API. Do you mind adding some lemmas about this definition, such as
- the identity is orientation-preserving
- the composition of two orientation-preserving maps is orientation-preserving
- (optional) define orientation-reversing maps; show that (over a connected set), each diffeomorphism is either orientation-preserving or reversing
- (optional) the composition of or.-preserving and -reversing is orientation-reversing; composing two orientation-reversing maps yields something orientation-preserving

Another future possibility (perhaps not for this PR) could be to define the groupoid of smooth orientation-preserving manifolds. You need not do all of these! But the first two items above would be nice, to confirm this definition is usable.

Mathlib/Geometry/Manifold/Orientable.lean
Rida-Hamadani: Dot notation works for the next lines, but not for this one (the `have` proof terms).

Mathlib/Geometry/Manifold/Orientable.lean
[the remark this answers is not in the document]
Rida-Hamadani: I've considered doing that, and it turns out that I will need to add another condition to get differentiability from the positivity of `fderiv`, one possible condition to obtain that is `FiniteDimensional.finrank ℝ H ≠ 0` (cf. `LinearMap.det_zero`). Since I need to add a condition anyways, I decided to go with the one that sounded to me closer to spoken mathematics. Do you think that the `FiniteDimensional.finrank ℝ H ≠ 0` condition is better in our case?

grunweg:
Another thought just came to my mind: there is a notion of orientability of *topological* manifolds, using local orientations. (This is not an issue for surfaces, as topological and smooth manifolds are the same in dimension less than four.)
Having this definition is useful now, but it's a definition only for differentiable manifolds. Can you add a comment clarifying this assumption?

grunweg:
Another interesting follow-up lemma could be to define orientation-preserving (or reversing) map between smooth manifolds. Up to you if you'd like to do this!

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+import Mathlib.Geometry.Manifold.SmoothManifoldWithCorners
+
+/-!
+# Orientable Manifolds
+
+This module defines orientable manifolds.
+
+## Main Definitions
+
+- `OrientationPreserving` : a map is orientation-preserving on a given set if the determinant of
```
grunweg, line 15: Let's be more precise: "a map between normed spaces ..."
(This notion also exists for maps between oriented manifolds, which this PR is a step towards.)

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+  its Jacobian is strictly positive on that set.
+- `orientationPreservingGroupoid` : the groupoid of partial homeos of `H` which are
```
grunweg, line 17: Can you mention the corresponding declarations for orientation-reversing as well?

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+  orientation-preserving.
+- `OrientableManifold`: a type class saying that the charted space `M`, modelled on the space `H`,
+  admis an orientation.
```
grunweg, line 20:
```suggestion
  admits an orientation.

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 6]
```diff
@@ -0,0 +1,135 @@
...
+- Define orientable `0`-manifolds.
+
+-/
+
+variable {H : Type*} [NormedAddCommGroup H] [NormedSpace ℝ H]
+
+section OrientationPreserving
+
+/--
+A map is orientation-preserving on a given set if it is differentiable and the determinant of its
```
grunweg, line 33: Similarly as above, can you clarify this to "A map between normed spaces"? Same comment for the other declarations below.

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+Jacobian is strictly positive on that set.
+-/
+def OrientationPreserving (f : H → H) (s : Set H) : Prop :=
+  ∀ x ∈ s, 0 < (fderiv ℝ f x).det
+
```
grunweg, line 38: If I understand Lean's convention (0^0 = 1) correctly, this means 0-dimensional manifolds are orientable. (This is mathematically correct: these are simply points; an orientation of a zero-manifold simply assigns each point a value $$\pm 1$$.)
Can you add a lemma stating this, that `H` being 0-dimensional implies f is orientation-preserving?

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+/--
+A map is orientation-reversing on a given set if it is differentiable and the determinant of its
+Jacobian is strictly negative on that set.
+-/
+def OrientationReversing (f : H → H) (s : Set H) : Prop :=
+  ∀ x ∈ s, (fderiv ℝ f x).det < 0
+
+lemma OrientationPreserving.DifferentiableAt [Nontrivial H] [FiniteDimensional ℝ H] {f : H → H}
```
grunweg, line 46: Three small comments: all of them also apply to the other lemmas below.
- I think you don't need the `Nontrivial H` assumption: if `H` is zero-dimensional, `f` is always differentiable. I think mathlib has this already; ping me if you get stuck finding it!
- By default, I would put the `x` "left of the colon", i.e. write "(x : s) : DifferentiableAt \R f x" instead. Do you have a reason to do this differently?
- In terms of naming, I'd write "OrientationPreserving.differentiableAt".

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+lemma orientationPreserving_comp [Nontrivial H] [FiniteDimensional ℝ H] {f g : H → H} {u v : Set H}
+    (hf : OrientationPreserving f u) (hg : OrientationPreserving g v) :
+    OrientationPreserving (g ∘ f) (u ∩ f ⁻¹' v) := by
+  intro x ⟨hxu, hxv⟩
+  rw [fderiv.comp x (hg.DifferentiableAt (f x) hxv) (hf.DifferentiableAt x hxu)]
+  unfold ContinuousLinearMap.det ContinuousLinearMap.comp
+  rw [(fderiv ℝ g (f x)).toLinearMap.det_comp (fderiv ℝ f x).toLinearMap]
+  exact mul_pos (hg (f x) hxv) (hf x hxu)
+
+lemma orientationReversing_comp_OrientationPreserving [Nontrivial H] [FiniteDimensional ℝ H]
```
grunweg, line 77: `orientationReversing_comp_orientationPreserving`, please - thanks!

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+class OrientableManifold (H : Type*) [Nontrivial H] [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M] extends
+  HasGroupoid M (@orientationPreservingGroupoid H _ _ _ _) : Prop
+
+/-- Typeclass defining orientable smooth manifolds. -/
+class OrientableSmoothManifold (H : Type*) [Nontrivial H] [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
+    [OrientableManifold H M] (I : ModelWithCorners ℝ H M) [SmoothManifoldWithCorners I M] extends
+  SmoothManifoldWithCorners I M ∧ OrientableManifold H M : Prop
+
```
grunweg, line 134: Can you add the corollary that 0-dimensional manifolds are always orientable?

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+- `OrientationPreserving` : a map is orientation-preserving on a given set if the determinant of
+  its Jacobian is strictly positive on that set.
+- `orientationPreservingGroupoid` : the groupoid of partial homeos of `H` which are
+  orientation-preserving.
+- `OrientableManifold`: a type class saying that the charted space `M`, modelled on the space `H`,
+  admis an orientation.
+
+## TODO
+
+- Define orientable `0`-manifolds.
```
grunweg, line 24: IIUC, this PR already handles the 0-dimensional case (and correctly). Instead, I would extend this TODO slightly differently: I'll leave this blank for now and try to return here after dinner.

grunweg (commented):
Thanks for addressing these comments so quickly! I think this PR is converging quickly. I took a quick look; all comments I have are pretty minor. (I didn't look at the proofs in detail yet.) It's time for dinner now; I'll try my best to return to this afterwards.

Mathlib/Geometry/Manifold/Orientable.lean
grunweg: So, here's an initial list of what I'd write.
- try to generalise this discussion to any non-trivially normed field
- on a given connected set, a diffeomorphism is either orientation-preserving or orientation-reversing
- a normed space (with the trivial model) is orientable
- the n-sphere is orientable
- products of orientable manifolds are orientable

- define orientations of a smooth manifold, and show that a manifold is orientable if and only if it admits an orientation
- define orientation-preserving and reserving maps between manifolds

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,135 @@
...
+Copyright (c) 2024 Rida Hamadani. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Rida Hamadani
+-/
+import Mathlib.Geometry.Manifold.SmoothManifoldWithCorners
+
+/-!
+# Orientable Manifolds
+
+This module defines orientable manifolds.
```
grunweg, line 11: Here's a slightly longer version, if you prefer:
```This file defines orientable manifolds: a differentiable manifold is orientable if and only if it admits an orientable atlas, i.e. an atlas whose transition functions have positive Jacobian.```

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 13]
```diff
@@ -0,0 +1,135 @@
...
+  its Jacobian is strictly positive on that set.
+- `orientationPreservingGroupoid` : the groupoid of partial homeos of `H` which are
+  orientation-preserving.
+- `OrientableManifold`: a type class saying that the charted space `M`, modelled on the space `H`,
+  admis an orientation.
+
+## TODO
+
+- Define orientable `0`-manifolds.
+
```
grunweg, line 25: Thinking about it more, it would be nice to mention the choice of definition also. What do you think about this (rewrapping lines as needed)?
```suggestion
## Implementation notes

The current definitions work for differentiable manifolds (and is simpler to define this way).
For topological manifolds, orientability can be defined using *local* orientations
(which mathlib cannot even state yet, as there is no e.g. singular homology).
In the future, it would be nice to generalise these definitions to allow for topological manifolds also, and relate them to the current definition.

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 14]
```diff
@@ -0,0 +1,135 @@
...
+/-- Typeclass defining orientable manifolds. -/
+class OrientableManifold (H : Type*) [Nontrivial H] [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M] extends
+  HasGroupoid M (@orientationPreservingGroupoid H _ _ _ _) : Prop
+
+/-- Typeclass defining orientable smooth manifolds. -/
+class OrientableSmoothManifold (H : Type*) [Nontrivial H] [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
+    [OrientableManifold H M] (I : ModelWithCorners ℝ H M) [SmoothManifoldWithCorners I M] extends
+  SmoothManifoldWithCorners I M ∧ OrientableManifold H M : Prop
```
grunweg, line 133: Hm, I'd need to think if this does what we'd want: this definition asks for "M has a smooth atlas, and M has an orientable atlas" - whereas the obviously correct definition would be "M has a smooth orientable atlas" (i.e., one atlas which is both smooth and orientable).

grunweg (commented):
Here are my remaining comments for today.

Mathlib/Geometry/Manifold/Orientable.lean
Rida-Hamadani: Thank you for pointing these out! Regarding point two: instead of writing `(x : s)`, I decided to look at similar lemmas in mathlib and imitate how it is done there (e.g. `DifferentiableWithinAt.differentiableAt`).

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+  if the determinant of its Jacobian is strictly negative on that set.
+- `orientationPreservingGroupoid` : the groupoid of partial homeos of `H` which are
+  orientation-preserving.
+- `OrientableManifold` : a type class saying that the charted space `M`, modelled on the space
+  `H`, admits an orientation.
+- `OrientableSmoothManifold` : a type class representing a manifold that is both orientable
+  and smooth.
+
+## Main Results
+
```
grunweg, line 29: What do you think about mentioning that 0-dimensional manifolds are always orientable?

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+- `orientationReversing_comp_orientationPreserving` : a composition between an orientation
+  reversing map and an orientation preserving map is orientation reversing.
+- `orientationPreserving_comp_orientationReversing` : a composition between an orientation
+  preserving map and an orientation reversing map is orientation reversing.
+- `orientationReversing_comp` : a composition between two orientation reversing maps is
+  orientation preserving.
+
+## TODO
+
+- Generalize this discussion to any non-trivially normed field.
```
grunweg, line 43: I tried this and got stuck somewhere; let's tone down the wording.
```suggestion
- Generalize this discussion to other fields; it certainly makes sense over ℚ

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 17]
```diff
@@ -0,0 +1,192 @@
...
+- On a given connected set, a diffeomorphism is either orientation preserving or orientation
+  reversing.
+- A normed space (with the trivial model) is orientable.
+- The `n`-sphere is orientable.
+- Products of orientable manifolds are orientable.
+- Define orientations of a smooth manifold, and show that a manifold is orientable if and only if it
+  admits an orientation.
+- Define orientation preserving and reserving maps between manifolds.
```
grunweg, line 51:
```suggestion
- Define orientation preserving and reversing maps between manifolds.

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 18]
```diff
@@ -0,0 +1,192 @@
...
+/--
+A map between normed spaces is orientation-reversing on a given set if it is differentiable and the
+determinant of its Jacobian is strictly negative on that set.
+-/
+def OrientationReversing (f : H → H) (s : Set H) : Prop :=
+  ∀ x ∈ s, (fderiv ℝ f x).det < 0
+
+lemma orientationPreserving_of_zero_dim (f : H → H) (s : Set H) (h : Module.rank ℝ H = 0) :
+    OrientationPreserving f s :=
+  fun _ _ ↦
```
grunweg, line 82: You can golf this proof, using `LinearMap.det_eq_one_of_finrank_eq_zero.`
(I am not sure which way of saying "is zero-dimension" is the best one; a maintainer can certainly weigh in on this.)
```
lemma orientationPreserving_of_zero_dim (f : H → H) (s : Set H)
    (h : FiniteDimensional.finrank ℝ H = 0) : OrientationPreserving f s := by
  intro _ _
  simp [LinearMap.det_eq_one_of_finrank_eq_zero h]
```

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+    rw [(Filter.eventuallyEq_of_mem (hu.mem_nhds hx) fg).fderiv_eq]
+    exact hf x hx
+
+/-- The groupoid of orientation-preserving maps. -/
+def orientationPreservingGroupoid [FiniteDimensional ℝ H] : StructureGroupoid H :=
+  orientationPreservingPregroupoid.groupoid
+
+end OrientationPreserving
+
+section OrientableManifold
```
grunweg, line 169: I personally prefer the comment on top of the section name. I wonder if the current behaviour is also incidental due to the port. Not a strong opinion, in any way.
```
 /-! ### Orientable manifolds -/

section OrientableManifold
```

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+/-- Typeclass defining orientable manifolds. -/
+class OrientableManifold (H : Type*) [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M] extends
+  HasGroupoid M (@orientationPreservingGroupoid H _ _ _) : Prop
+
+/-- `0`-dimensional manifolds are always orientable. -/
+lemma orientableManifold_of_zero_dim (H : Type*) [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
+    (h : Module.rank ℝ H = 0) : OrientableManifold H M where
+  compatible := fun {_ _} _ _ ↦
```
grunweg, line 183:
```suggestion
  compatible := fun _ _ ↦

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 21]
```diff
@@ -0,0 +1,192 @@
...
+  have det_eq_one : (fderiv ℝ f _).det = 1 := by
+    have b : Basis (Fin 0) ℝ H := Basis.ofEquivFun (finDimVectorspaceEquiv 0 h)
+    rw [ContinuousLinearMap.det, ← (fderiv ℝ f _).det_toMatrix b]
+    exact Matrix.det_fin_zero
+  det_eq_one ▸ Real.zero_lt_one
+
+lemma OrientationPreserving.differentiableAt [FiniteDimensional ℝ H] {f : H → H} {s : Set H}
+    (h : OrientationPreserving f s) {x : H} (hs : x ∈ s) : DifferentiableAt ℝ f x := by
+  cases subsingleton_or_nontrivial H
+  · apply DifferentiableWithinAt.differentiableAt
```
grunweg, line 92: The first case can be golfed a little, to
```
apply ((s.subsingleton_of_subsingleton).differentiableOn _ hs).differentiableAt
exact mem_nhds_discrete.mpr hs
```
I would argue this is mathematically "trivial", hence fine.

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+    apply Set.Subsingleton.differentiableOn
+    exact s.subsingleton_of_subsingleton
+    exact hs
+    exact mem_nhds_discrete.mpr hs
+  · unfold OrientationPreserving at h
```
grunweg, line 97: In general, `unfold` is a bit of a "code smell": a sign that something is not ideal (such as, a missing simple lemma about the definition to use). In this case, you can just `rw` by the definition instead.
```suggestion
  · rw [OrientationPreserving] at h

File: Mathlib/Geometry/Manifold/Orientable.lean [hunk 23]
```diff
@@ -0,0 +1,192 @@
...
+    contrapose! h
+    use x, hs
+    rw [fderiv_zero_of_not_differentiableAt h, ContinuousLinearMap.det]
+    simp [ne_of_gt FiniteDimensional.finrank_pos]
+
+lemma OrientationReversing.differentiableAt {f : H → H} {s : Set H} (h : OrientationReversing f s)
+    {x : H} (hs : x ∈ s) : DifferentiableAt ℝ f x := by
+  unfold OrientationReversing at h
```
grunweg, line 105: Please also `rw` here.

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+
+lemma orientationPreserving_id (s : Set H) : OrientationPreserving id s := by
+  intro
+  simp [ContinuousLinearMap.det]
+
+lemma orientationPreserving_comp [FiniteDimensional ℝ H] {f g : H → H} {u v : Set H}
+    (hf : OrientationPreserving f u) (hg : OrientationPreserving g v) :
+    OrientationPreserving (g ∘ f) (u ∩ f ⁻¹' v) := by
+  intro x ⟨hxu, hxv⟩
+  rw [fderiv.comp x (hg.differentiableAt hxv) (hf.differentiableAt hxu)]
```
grunweg, line 119: Here is a golf of the last three lines ("unfold" to "exact").
- avoiding `unfold` in favour of `rw`
- in fact, `simp` can also be used (which makes this even shorter)
- in the last step, `simp` followed by `exact` can often be converged to `simpa [...] using`: I think that is more readable here.
Overall, I'd suggest
```
simpa [ContinuousLinearMap.det] using mul_pos (hg (f x) hxv) (hf x hxu)
```

Please make the same change to the other three lemmas below.

Mathlib/Geometry/Manifold/Orientable.lean
```diff
@@ -0,0 +1,192 @@
...
+    [FiniteDimensional ℝ H] (M : Type*) [TopologicalSpace M] [ChartedSpace H M]
+    (h : Module.rank ℝ H = 0) : OrientableManifold H M where
+  compatible := fun {_ _} _ _ ↦
+    ⟨orientationPreserving_of_zero_dim _ _ h, orientationPreserving_of_zero_dim _ _ h⟩
+
+/-- Typeclass defining orientable smooth manifolds. -/
+class OrientableSmoothManifold (H : Type*) [NormedAddCommGroup H] [NormedSpace ℝ H]
+    [FiniteDimensional ℝ H] {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E] (M : Type*)
+    [TopologicalSpace M] [ChartedSpace H M] (I : ModelWithCorners ℝ E H) extends
+  HasGroupoid M ((contDiffGroupoid ⊤ I) ⊓ orientationPreservingGroupoid) : Prop
```
grunweg, line 190: You *could* extract that groupoid and give it a name. Up to you!

grunweg (commented):
Wow, you're quick! I took a close look at the proofs and have a few golfs to suggest. Should be quick to address. After that, it's probably ready to be maintainer merged.

Mathlib/Geometry/Manifold/Orientable.lean
grunweg: I think this is fine as-is; if necessary, it can always be extracted later.

grunweg:
Thanks! Let's have another pair of eyes on this.
maintainer merge

ocfnash:
I looks like this only deals with the case of manifolds without boundary. I think this approach to orientability can be made to work for boundary (though it's a bit fiddly).

What if I wanted to use this definition state Stokes's theorem?

For the sake of a concrete goal, I'd be much happier merging something that included an instance of orientability for `(Icc (0 : ℝ) 1)`.

PR #15781, opened 2024-08-13
[Merged by Bors] - feat: run style linters in fix mode and push a commit when "bot fix style" is posted in a PR comment

bryangingechen:
The bot is triggered when writes "bot fix style" on a PR, either in an ordinary comment, a review comment, or a review. It then checks if the author has "write" privileges (or above) and if so, will react with a 🚀 emoji on the triggering comment and then lint and push a commit to the PR branch.

Note that this uses the NIGHTLY_TESTING secret.

See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/reviewdog/near/460017984).

---

- [ ] depends on: #15051

[Open in Gitpod](https://gitpod.io/from-referrer/)

This can be tested in [my fork of mathlib4](https://github.com/bryangingechen/mathlib4). The main things left to test are permissions and other edge cases. So feel free to open weird issues / PRs to my fork and comment on them to try and break things. I think only people with write access should be able to trigger the bot now, so if you want write access to my fork let me know here or on Zulip!

For starters, if someone would like to try commenting `bot fix style` on https://github.com/bryangingechen/mathlib4/pull/2 (also in a PR review / PR review comment) I can check if the permissions check is working properly. (Since no one other than me has write access at this time, hopefully the bot will ignore your comment.)

PR #16241, opened 2024-08-29
[Merged by Bors] - fix(*_comment.yml): fix uses of github.event.issue.pull_request

bryangingechen:
In our Github actions workflows which react to comments, there were several instances of `github.event.issue.pull_request != null` to check whether the comment was posted on a PR. However, this always returns `true`, so these workflows could in principle be triggered to run by someone writing comments on issues rather than PRs.

The [Github actions documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only) suggests just using `github.event.issue.pull_request` in conditionals, so I've replaced the incorrect tests with that.

There was also a usage of `toJSON(github.event.issue.pull_request) != null` which even if correct is overcomplicated, so I fixed that as well.

---

Follow-up to https://github.com/leanprover-community/mathlib4/pull/15781#discussion_r1727744412


grunweg:
I checked the documentation: this change makes sense. Thanks for correcting this!
maintainer merge

semorrison:
bors merge

PR #16253, opened 2024-08-29
feat: Basics of Discrete Fair Division in Mathlib

Shreyas4991:
This PR adds the basics of discrete fair division in mathlib. We include the definitions of allocations, valuations, properties of valuations, and the following notions : Envy-Freeness (EF), Envy-Freeness upto 1 good (EF1), and envy freeness upto any good (EFX). We also include the various implications between these notions.

---

### DIFF
--- a/Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
+++ b/Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
@@ -1,0 +1,144 @@
+/-
+Copyright (c) 2024 Shreyas Srinivas. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Shreyas Srinivas
+-/
+
+import Mathlib.Algebra.BigOperators.Group.Finset
+import Mathlib.Data.Real.Basic
+import Mathlib.Tactic.Linarith
+import Mathlib.Tactic.Ring.Basic
+
+
+namespace Mathlib
+
+namespace DiscreteFairDivision
+
+variable {Agents Goods : Type}
+variable [Fintype Agents] [Fintype Goods]
+variable [DecidableEq Agents]
+variable [DecidableEq Goods] -- Indivisible goods only
+
+
+
+structure Alloc where
+  assign : Agents → Finset Goods
+  disjoint : ∀ a b : Agents, a ≠ b → assign a ∩ assign b = ∅
+
+structure Val where
+  val : Agents → Finset Goods → ℝ
+  empty : ∀ a : Agents, val a ∅ = 0
+  pos : ∀ a : Agents, ∀ s : Finset Goods, val a s ≥ 0
+  mono : ∀ a : Agents, ∀ s t : Finset Goods,
+    s ⊆ t → val a s ≤ val a t
+
+
+variable (alloc : @Alloc Agents Goods _)
+variable (v : @Val Agents Goods)
+
+def valSingle (a : Agents) (g : Goods) :=
+  v.val a {g}
+
+def isEmpty (a : Agents):=
+  alloc.assign a = ∅
+
+abbrev unAllocated (g : Goods) := ∀ a, g ∉ alloc.assign a
+
+def allocate (g : Goods) (a : Agents) (h : unAllocated alloc g): @Alloc Agents Goods _ := {
+  assign := fun (x : Agents) => if x = a then (alloc.assign x) ∪ {g} else alloc.assign x
+  disjoint := by
+    intro arb_a arb_b hneq
+    simp
+    by_cases ha : arb_a = a
+    case pos =>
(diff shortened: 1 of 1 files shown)


Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
```diff
@@ -0,0 +1,144 @@
...
+variable {Agents Goods : Type}
+variable [Fintype Agents] [Fintype Goods]
+variable [DecidableEq Agents]
+variable [DecidableEq Goods] -- Indivisible goods only
+
+
+
+structure Alloc where
+  assign : Agents → Finset Goods
+  disjoint : ∀ a b : Agents, a ≠ b → assign a ∩ assign b = ∅
```
eric-wieser, line 26:
```suggestion
  disjoint : Pairwise (Disjoint on assign)
```

Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
```diff
@@ -0,0 +1,144 @@
...
+
+structure Val where
+  val : Agents → Finset Goods → ℝ
+  empty : ∀ a : Agents, val a ∅ = 0
+  pos : ∀ a : Agents, ∀ s : Finset Goods, val a s ≥ 0
+  mono : ∀ a : Agents, ∀ s t : Finset Goods,
+    s ⊆ t → val a s ≤ val a t
```
eric-wieser, line 33: I think this should be
```
abbrev Val := Agents → OrderBotHom Goods ℝ≥0
```
where `OrderBotHom` doesn't currently exist.

Shreyas4991: Two things:
1. What is `OrderBotHom`
2. The way I have stated these definitions is basically how researchers in this field would understand these valuation functions. A lot of arguments use these individual properties both implicitly and explicitly.

Shreyas4991: > I think this should be
> ```
> abbrev Val := Agents → OrderBotHom Goods ℝ≥0
> ```
> where `OrderBotHom` doesn't currently exist.

A third point: One can define additional properties on these valuation functions like submodularity, additivity, subadditivity etc.

eric-wieser: > What is OrderBotHom?

`structure OrderBotHom (A B) extends OrderHom A B, BotHom A B` or similar

> One can define additional properties on these valuation functions like submodularity, additivity, subadditivity etc.

This is perhaps an argument for writing your theorem assuming `[OrderHomClass F Goods ℝ≥0] [BotHomClass F Goods ℝ≥0]`, so that you can then apply it with `abbrev Val := Agents → F`, and still use it on other types of valuation.

Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
```diff
@@ -0,0 +1,144 @@
...
+    case neg =>
+      by_cases hb : arb_b = a <;>
+        simp[hb, ha, alloc.disjoint, hneq]
+      rw[←hb]
+      simp[Finset.inter_union_distrib_left, hneq, alloc.disjoint]
+      simp [h]
+}
+
+def isEmptyAlloc  :=
+  ∀ a : Agents, alloc.assign a = ∅
```
eric-wieser, line 70: This is the same as `alloc.assign = ⊥`, which saves you a definition. You could quite reasonably even implement a `Bot Alloc` instance to make this `alloc = ⊥`.

Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
```diff
@@ -0,0 +1,144 @@
...
+
+def complete : Prop :=
+  ∀ g : Goods, ∃ a : Agents, g ∈ alloc.assign a
+
+def additive : Prop :=
```
eric-wieser, line 75: Propositions should be `PascalCase`.

Mathlib/GameTheory/FairDivision/Discrete/Basic.lean
```diff
@@ -0,0 +1,144 @@
...
+
+structure Val where
+  val : Agents → Finset Goods → ℝ
+  empty : ∀ a : Agents, val a ∅ = 0
+  pos : ∀ a : Agents, ∀ s : Finset Goods, val a s ≥ 0
+  mono : ∀ a : Agents, ∀ s t : Finset Goods,
+    s ⊆ t → val a s ≤ val a t
+
+
+variable (alloc : @Alloc Agents Goods _)
```
eric-wieser, line 36: This is good evidence that `Agents` and `Goods` should be explicit arguments to `alloc`.

PR #16013, opened 2024-08-20
[Merged by Bors] - feat(ContinuousFunctionalCalculus): Show that integration commutes with the CFC

dupuisf:
This PR shows that `cfc (fun r => ∫ x, f x r ∂μ) a = ∫ x, cfc (f x) a ∂μ` under appropriate conditions.

The continuity condition given here is enough for most applications (I think) and reasonably easy to apply for the integral representations this is meant for, but I would be open to suggestions.

---

[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #16255, opened 2024-08-29
[Merged by Bors] - feat(ContinuousFunctionalCalculus): show that the non-unital CFC commutes with integration

dupuisf:
This PR shows that `cfcₙ (fun r => ∫ x, f x r ∂μ) a = ∫ x, cfcₙ(f x) a ∂μ` under appropriate conditions.

Note that #16013 (already merged) showed the corresponding result for the unital CFC.

---

### DIFF
--- a/Mathlib/Analysis/CStarAlgebra/ContinuousFunctionalCalculus/Integral.lean
+++ b/Mathlib/Analysis/CStarAlgebra/ContinuousFunctionalCalculus/Integral.lean
@@ -6,6 +6,7 @@ Authors: Frédéric Dupuis
 
 import Mathlib.Analysis.Normed.Algebra.Spectrum
 import Mathlib.Analysis.CStarAlgebra.ContinuousFunctionalCalculus.Unital
+import Mathlib.Analysis.CStarAlgebra.ContinuousFunctionalCalculus.NonUnital
 import Mathlib.MeasureTheory.Integral.SetIntegral
 
 /-!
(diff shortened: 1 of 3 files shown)


Mathlib/Analysis/CStarAlgebra/ContinuousFunctionalCalculus/Integral.lean
```diff
@@ -19,29 +20,45 @@ that the integral commutes with the continuous functional calculus under appropr
...
-+ Prove a similar result for the non-unital case
 + Lift this to the case where the CFC is over `ℝ≥0`
 + Use this to prove operator monotonicity and concavity/convexity of `rpow` and `log`
 -/
 
 open MeasureTheory
 
+open scoped ContinuousMapZero
+
+theorem ContinuousMapZero.integral_apply {X Y R : Type*} [MeasurableSpace X]
```
j-loreaux, line 37: I guess the point here is that `ContinuousMapZero` isn't imported by anything outside of the continuous functional calculus, so you can't put this earlier?

I *think* I would be in favor instead importing that into the file where `ContinuousMap.integral_apply` appears, as I don't believe it will increase the imports there too substantially (but maybe I'm wrong).

j-loreaux (commented):
LGTM modulo moving that. If moving it increases transitive imports too much, then you can just leave it here.

bors d+

Mathlib/Analysis/CStarAlgebra/ContinuousFunctionalCalculus/Integral.lean
dupuisf: Indeed that's why I had left it there, but now I checked and the extra transitive imports are literally just that one `ContinuousMapZero` file, so I moved the lemma.

PR #16270, opened 2024-08-29
[Merged by Bors] - feat(Order/Hom/Lattice): add `Subtype.val` as lattice homs

Command-Master:
---

### DIFF
--- a/Mathlib/Order/Hom/Lattice.lean
+++ b/Mathlib/Order/Hom/Lattice.lean
@@ -474,6 +474,24 @@ theorem bot_apply [Bot β] (a : α) : (⊥ : SupHom α β) a = ⊥ :=
 theorem top_apply [Top β] (a : α) : (⊤ : SupHom α β) a = ⊤ :=
   rfl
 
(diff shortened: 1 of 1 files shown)


Mathlib/Order/Hom/Lattice.lean
```diff
@@ -474,6 +474,24 @@ theorem bot_apply [Bot β] (a : α) : (⊥ : SupHom α β) a = ⊥ :=
 theorem top_apply [Top β] (a : α) : (⊤ : SupHom α β) a = ⊤ :=
   rfl
 
+/-- `Subtype.val` as a `SupHom`. -/
+def subtype_val {P : β → Prop}
```
YaelDillies, line 478:
```suggestion
def subtypeVal {P : β → Prop}
```

YaelDillies (approved):
Thanks!

maintainer merge

(I am wondering exactly how you are going to use these, since there are so many non-instance arguments, but I guess that's a question for later PRs)

urkud:
Thanks! :tada:
bors merge

PR #16277, opened 2024-08-29
[Merged by Bors] - chore: rename Finsupp.total

PatrickMassot:
to `Finsupp.linearCombination`.

---
See https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Finsupp.2Etotal for discussions.


PatrickMassot:
Note the PR summary bot complaining about loosing  `KaehlerDifferential.derivationQuotKerTotal_lift_comp_total` is a bot mistake. It comes from this name being so long that it can’t fit on the same name as `alias`.

Mathlib/LinearAlgebra/LinearIndependent.lean
```diff
@@ -423,36 +423,36 @@ theorem linearIndependent_comp_subtype {s : Set ι} :
...
+      Disjoint (Finsupp.supported R R s) (LinearMap.ker <| Finsupp.linearCombination R v) := by
   rw [linearIndependent_comp_subtype, LinearMap.disjoint_ker]
 
 theorem linearIndependent_subtype_disjoint {s : Set M} :
     LinearIndependent R (fun x => x : s → M) ↔
-      Disjoint (Finsupp.supported R R s) (LinearMap.ker <| Finsupp.total R id) := by
+      Disjoint (Finsupp.supported R R s) (LinearMap.ker <| Finsupp.linearCombination R id) := by
   apply linearIndependent_comp_subtype_disjoint (v := id)
 
 theorem linearIndependent_iff_totalOn {s : Set M} :
```
urkud, line 450: This theorem should be renamed. Also, please run `git grep totalOn` to make sure you didn't miss other similar theorems.

urkud:
Otherwise LGTM. Thanks!
bors d+

Mathlib/LinearAlgebra/LinearIndependent.lean
PatrickMassot: Good catch. I wasn’t systematic with `totalOn` like I was with `total` because it is almost never used. And indeed that particular lemma was the only missing rename and it is never used in Mathlib.

PatrickMassot:
bores merge

semorrison:
bors merge

PR #16282, opened 2024-08-30
feat(Algebra/Group/Action): add definition of equidecomposition

Felix-Weilacher:
add definitions about equidecompositions and begin to develop some basic theory. 

TODO:
1. Prove that if two sets embeddidecompose into eachother, they are equidecomposable (Schroeder-Bernstein type theorem)
2. Define embeddidecomposability as a Preorder on sets and prove that its induced equivalence relation is equidecomposability. 
3. Prove the definition of equidecomposition used here is equivalent to the more familiar one using partitions. This equivalence is trivial mathematically, but partitions do not seem well-developed to me in mathlib. 

---

### DIFF
--- a/Mathlib.lean
+++ b/Mathlib.lean
@@ -216,6 +216,7 @@ import Mathlib.Algebra.GradedMonoid
 import Mathlib.Algebra.GradedMulAction
 import Mathlib.Algebra.Group.Action.Basic
 import Mathlib.Algebra.Group.Action.Defs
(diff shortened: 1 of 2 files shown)


Felix-Weilacher:
Going to refactor some definitions to use `Rel`...

Felix-Weilacher:
I just found out about `PartialEquiv`, which may be a better way to set this up. I'm going to leave this open in case people want to take a look, but I may refactor it in the coming days.

PR #16287, opened 2024-08-30
[Merged by Bors] - feat(ContinuousFunctionalCalculus): The restriction of a non-unital CFC is equal to the original one

dupuisf:
Using the non-unital CFC over `ℂ` for a function `f : ℝ → ℝ` is the same as using the non-unital CFC over `ℝ` for the same function, and likewise for `ℝ` vs `ℝ≥0`. We already had this for the unital CFC.

---


j-loreaux:
bors merge

PR #16289, opened 2024-08-30
chore(scripts): update nolints.json and style-exceptions.txt

bryangingechen:
I am happy to remove some nolints for you!


bryangingechen:
This was generated by the script in #13498. Though this PR is based off that branch rather than `master`, I think it should work properly once the script is updated in `master` and run from a cron schedule...?

It also looks like I need to get a token from a bot account, rather than my personal account!

PR #16293, opened 2024-08-30
feat(MeasureTheory/MeasurableSpace/Basic): sigma-algebra generated by an element

vihdzp:
We give an explicit characterization of the sets in a σ-algebra generated by a singleton.

---

If there were a way to directly rewrite `∃ x : Set Prop, p x` as `p Ø ∨ p {True} ∨ p {False} ∨ p {True, False}`, the proof could be simplified a bit.


Mathlib/MeasureTheory/MeasurableSpace/Basic.lean
```diff
@@ -1108,6 +1109,26 @@ namespace MeasurableSpace
   rintro _ ⟨u, -, rfl⟩
   exact (show MeasurableSet s from GenerateMeasurable.basic _ <| mem_singleton s).mem trivial
 
+theorem measurableSet_singleton_iff {s t : Set α} :
```
urkud, line 1112: Please mention `generateFrom` in the name. Otherwise it looks like the lemma is about `MeasurableSet {a}`.

vihdzp: `measurableSet_generateFrom_singleton_iff`? That's quite the mouthful.

vihdzp: I updated the theorem name.

Mathlib/Data/Fintype/Basic.lean
```diff
@@ -940,10 +940,12 @@ instance PLift.fintypeProp (p : Prop) [Decidable p] : Fintype (PLift p) :=
 instance Prop.fintype : Fintype Prop :=
   ⟨⟨{True, False}, by simp [true_ne_false]⟩, by simpa using em⟩
 
-@[simp]
```
RemyDegenne, line 943: Why did you remove this `@[simp]`?
I am not saying you shouldn't remove it, but I'd like to have more details about the reasons.

RemyDegenne (commented):
Let's merge this. It is simple and has waited for a review long enough.

I only have one question about a simp attribute, but apart from that I think we could merge.

Mathlib/Data/Fintype/Basic.lean
vihdzp: This was causing problems with simp confluence, where lemmas on `Set.univ` wouldn't fire when written in this form. My proof breaks with the simp attribute active.

RemyDegenne:
Thanks!
bors r+

Mathlib/MeasureTheory/MeasurableSpace/Basic.lean
```diff
@@ -1133,6 +1134,26 @@ namespace MeasurableSpace
   rintro _ ⟨u, -, rfl⟩
   exact (show MeasurableSet s from GenerateMeasurable.basic _ <| mem_singleton s).mem trivial
 
+theorem measurableSet_generateFrom_singleton_iff {s t : Set α} :
+    MeasurableSet[generateFrom {s}] t ↔ t = ∅ ∨ t = s ∨ t = sᶜ ∨ t = Set.univ := by
```
YaelDillies, line 1138: I don't think we want this lemma. Wherever you want to use it, you should be using `generateFrom_singleton` directly.

This is of the same kind as people on Zulip asking how to show that all elements of some subgroup are of a certain form, rather than using the induction principles.

I too wanted it originally, and then I proved `generateFrom_singleton` and my problem was solved.

Mathlib/Data/Finset/Basic.lean
YaelDillies, line 1: FYI this is making the file too long, but you need to merge master to see it

Mathlib/MeasureTheory/MeasurableSpace/Basic.lean
vihdzp: If I correctly recall, my actual application for this lemma was showing that the cardinality of this set is 4, which appeared as a special case in a more general formula for the cardinality of the generated sigma algebra. An induction principle can't be directly used to prove that.

YaelDillies: Can we see the application?

vihdzp: It's in a draft PR I have, I'll link it in an hour or two when I'm on my computer.

vihdzp:
Thinking about it. If we really want to prove results about cardinalities of finite sigma-algebras, it'd be better to prove that the sigma-algebra generated by n elements has cardinality at most 2^(n + 1), instead of only having the case n = 1.

